PR #11804, opened 2018-10-03
HPCC-20684 Minor improvements to the simplified expression flags

shamser:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/ecl/hql/hqlgram2.cpp
+++ b/ecl/hql/hqlgram2.cpp
@@ -6171,7 +6171,7 @@ void HqlGram::reportErrorUnexpectedX(const attribute& errpos, IAtom * unexpected
 
 static bool includeError(HqlLookupContext & ctx, WarnErrorCategory category)
 {
-    if (ctx.syntaxChecking() && !ctx.ignoreCache())
+    if (ctx.syntaxChecking() && !ctx.ignoreSimplified())
     {
         if (category==CategorySyntax || category==CategoryError)
             return true;
@@ -12287,18 +12287,25 @@ bool parseForwardModuleMember(HqlGramCtx & _parent, IHqlScope *scope, IHqlExpres
     return (prevErrors == ctx.errs->errCount());
 }
 
+// If ctx.hasCacheLocation() then the cache entry is always updated with the dependency information (unless it is unavailable i.e. a forward scope)
+// If ctx.ignoreSimplified() then the expressions are never simplified and the parsing behaves the same way as 6.x
+// if ctx.ignoreCache() then the contents of the cache are not used, but the cache is still updated
+// if ctx.regenerateCache() then the contents of the cache are not used and the cache is forced to be updated
+// if ctx.checkSimpleDef() then simplified definitions are created and checked - even if there is no cache configured.
 void parseAttribute(IHqlScope * scope, IFileContents * contents, HqlLookupContext & ctx, IIdAtom * name, const char * fullName)
 {
     bool alreadySimplified = false;
+    bool cacheUptoDate = false;
     HqlLookupContext attrCtx(ctx);
     attrCtx.noteBeginAttribute(scope, contents, name);
 
-    // Use simplified definition if syntax checking
-    if (!ctx.regenerateCache() && ctx.syntaxChecking() && ctx.hasCacheLocation() && !ctx.ignoreCache())
+    // Check if cache is up to date and obtain simplified definition from cache
+    if (ctx.hasCacheLocation() && !ctx.regenerateCache())
     {
         HqlParseContext & parseContext = ctx.queryParseContext();
         Owned<IEclCachedDefinition> cached = parseContext.cache->getDefinition(fullName);
-        if (cached->isUpToDate(parseContext.optionHash))
+        cacheUptoDate = cached->isUpToDate(parseContext.optionHash);
+        if (cacheUptoDate && ctx.syntaxChecking() && !ctx.ignoreSimplified() && !ctx.ignoreCache())
         {
             IFileContents * cachecontents = cached->querySimplifiedEcl();
             if (cachecontents)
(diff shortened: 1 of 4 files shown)


ghalliday:
@shamser can I clarify which combination of options I used for the following scenarios:

i) ignore the contents of the cache, but updated it if it isn't up to date.  (Needed when meta generation is enabled.)
ii) backward compatibility - don't ever create a simplified expression (even for testing)
iii) regression suite - process all these archives and use the simplified expressions when they would normally be used.

I think the options cover (ii) and (iii) but not sure about (i)

ecl/hql/hqlgram2.cpp
```diff
@@ -12339,7 +12340,7 @@ void parseAttribute(IHqlScope * scope, IFileContents * contents, HqlLookupContex
             const bool isMacro = parsed->isMacro();
             const bool neverSimplify = isMacro || ctx.neverSimplify(fullName);
             OwnedHqlExpr simplified = neverSimplify ? nullptr : createSimplifiedDefinition(parsed);
-            if (ctx.hasCacheLocation() && !ctx.ignoreSimplified())
+            if (ctx.hasCacheLocation() && !ctx.ignoreCache())
```
ghalliday: I'm not sure that this is correct - it should mean ignore the contents of the cache, but not prevent it being updated.

Should it be just testing ctx.hasCacheLocation, or is there a reason for checking something else?

shamser: @ghalliday I thought ignoreCache implies that it doesn't update the cache either but I understand your point that metaCache set to blank does the same thing.  I'll update the code to allow updates when ignoreCache=true

ecl/hql/hqlgram2.cpp
```diff
@@ -12329,30 +12330,28 @@ void parseAttribute(IHqlScope * scope, IFileContents * contents, HqlLookupContex
         throw;
     }
     OwnedHqlExpr parsed = scope->lookupSymbol(name, LSFsharedOK|LSFnoreport, ctx);
```
ghalliday: I think it would be useful to have a comment at the start of parseAttribute which details how the flags are meant to work:
E.g.

// If ctx.hasCacheLocation() then the cache entry is always updated with the dependency information (unless it is unavailable i.e. a forward scope)
// If ctx.ignoreSimplified() then the expressions are never simplified and the parsing behaves the same way as 6.x
// if ctx.ignoreCache() then the contents of the cache are not used, but the cache is still updated
// if ctx.regenerateCache() then the contents of the cache are not used and the cache is forced to be updated
// if ctx.checkSimpleDef() then simplified definitions are created and checked - even if there is no cache configured.

I think that covers all the cases.  That was helpful for me to spell it out - I can see one conflict.  If ignoreSimplified is true should the cache information be updated?  If it is not the dependencies will not be correct, if it is the simplified will be forced to be blank.  I the latter is probably better since it means the dependency cache will still work if the user has disabled the simplified expressions.

Also, it highlights a significant optimization.  If the cache was up to date but was not simplified, then there is no benefit in creating a simplified expression - it failed last time, and will fail again.

There is a related issue for the expression too complex case, but I'll comment on that PR separately.

shamser: When ignoreCache() is enabled, should expressions be simplified and used?

ghalliday (commented):
closer

ecl/hql/hqlgram2.cpp
```diff
@@ -12329,30 +12335,35 @@ void parseAttribute(IHqlScope * scope, IFileContents * contents, HqlLookupContex
...
+        const bool canCache = parsed->getOperator() != no_forwardscope;
+        if (canCache && (ctx.hasCacheLocation() || ctx.syntaxChecking()))
         {
-            ctx.incrementAttribsSimplified();
+            const bool isMacro = parsed->isMacro();
+            const bool neverSimplify = isMacro || !ctx.syntaxChecking() ||
+                                       (ctx.ignoreSimplified() && !ctx.checkSimpleDef()) ||
+                                       ctx.neverSimplify(fullName);
+            OwnedHqlExpr simplified = neverSimplify ? nullptr : createSimplifiedDefinition(parsed);
+            if (ctx.hasCacheLocation())
```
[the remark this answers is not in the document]
shamser: In line 12340, it checks "alreadySimplified".  That will be true if cache is up-to date.  It's always false if forceupdate option is used.

ecl/hql/hqlgram2.cpp
```diff
@@ -12329,30 +12335,34 @@ void parseAttribute(IHqlScope * scope, IFileContents * contents, HqlLookupContex
...
-        if (ctx.hasCacheLocation())
-            attrCtx.createCache(simplified, isMacro);
-        if (simplified)
+        const bool canCache = parsed->getOperator() != no_forwardscope;
+        if (canCache && (ctx.hasCacheLocation() || ctx.syntaxChecking()))
         {
-            ctx.incrementAttribsSimplified();
+            const bool isMacro = parsed->isMacro();
+            OwnedHqlExpr simplified;
+            if (!isMacro && (!ctx.ignoreSimplified() || ctx.checkSimpleDef()) && !ctx.neverSimplify(fullName))
```
ghalliday: I'm still not convinced this matches the code below.  I would imagine

if (!isMacro &&
   ((ctx.hasCacheLocation() && !ctx.ignoreSimplified) ||
    ctx.checkSimpleDef() ||
    (ctx.syntaxChecking() && !ctx.ignoreSimplified()) && ctx.neverSimplify(fullName))

You could also argue that !ignoreSimplified() is also checked for the middle test.  If so it would become:

if (!isMacro && !ctx.ignoreSimplified() && 
   ((ctx.hasCacheLocation() ||
    ctx.checkSimpleDef() ||
    ctx.syntaxChecking())  && ctx.neverSimplify(fullName))

I need to think about whether there is an issue with the cache already being up to date, but I think that is only relevant to your other PR.

shamser: The second test looks neat and easily comprehensible.  But just to be clear we did previously agree that "If ctx.ignoreSimplified() is true and checkSimpleDef is true, then it needs to generate a simplified expression" so that it can check the simplified expression.  As long as we happy that these 2 flags set true means that the simplified expression will not be verified, we can use the second test.

ecl/hql/hqlgram2.cpp
```diff
@@ -12329,30 +12335,30 @@ void parseAttribute(IHqlScope * scope, IFileContents * contents, HqlLookupContex
...
-            attrCtx.createCache(simplified, isMacro);
-        if (simplified)
+        const bool canCache = parsed->getOperator() != no_forwardscope;
+        if (canCache && (ctx.hasCacheLocation() || ctx.syntaxChecking()))
         {
-            ctx.incrementAttribsSimplified();
+            const bool isMacro = parsed->isMacro();
+            OwnedHqlExpr simplified;
+            if (!isMacro && !ctx.ignoreSimplified() &&
+                (ctx.hasCacheLocation() || ctx.checkSimpleDef() || ctx.syntaxChecking()) && ctx.neverSimplify(fullName))
```
ghalliday: The test ctx.hasCacheLocation is not quite right.

It should probably be

(ctx.hasCacheLocation() && (!cacheIsUpTodate() || ctx.regenerateCache))

The difference is whether a new simplified definition is generated if the cache is already up to date.  I assume the cost of calculating the cache is up to date is cheaper than creating the simplified definition.

shamser: If the cache is up to date, alreadySimplified will be true.  So, the test in line 12,340 should mean that this section of code will not be executed.

ecl/hql/hqlgram2.cpp
```diff
@@ -12295,16 +12295,17 @@ bool parseForwardModuleMember(HqlGramCtx & _parent, IHqlScope *scope, IHqlExpres
...
     bool alreadySimplified = false;
+    bool cacheUptoDate = false;
     HqlLookupContext attrCtx(ctx);
     attrCtx.noteBeginAttribute(scope, contents, name);
 
-    // Use simplified definition if syntax checking
-    if (!ctx.regenerateCache() && ctx.syntaxChecking() && ctx.hasCacheLocation() &&
-        !ctx.ignoreSimplified() && !ctx.ignoreCache())
+    // Check if cache is up to date and obtain simplified definition from cache
+    if (ctx.hasCacheLocation())
```
ghalliday: Could also check !regenerateCache() here since it will always be overwritten.

ecl/hql/hqlgram2.cpp
```diff
@@ -12337,15 +12338,15 @@ void parseAttribute(IHqlScope * scope, IFileContents * contents, HqlLookupContex
     OwnedHqlExpr parsed = scope->lookupSymbol(name, LSFsharedOK|LSFnoreport, ctx);
     ctx.incrementAttribsProcessed();
 
-    if (parsed && !alreadySimplified)
+    if (parsed && (!alreadySimplified || !cacheUptoDate || ctx.regenerateCache()))
```
ghalliday: I don't think this condition needs to change.  if alreadySimplified is true then cacheUpToDate must be true and regenerate cache must be false.

ecl/hql/hqlgram2.cpp
```diff
@@ -12337,15 +12338,15 @@ void parseAttribute(IHqlScope * scope, IFileContents * contents, HqlLookupContex
...
     {
         const bool canCache = parsed->getOperator() != no_forwardscope;
         if (canCache && (ctx.hasCacheLocation() || ctx.syntaxChecking()))
         {
             const bool isMacro = parsed->isMacro();
             OwnedHqlExpr simplified;
             if (!isMacro && !ctx.ignoreSimplified() &&
-                (ctx.hasCacheLocation() || ctx.checkSimpleDef() || ctx.syntaxChecking()) && ctx.neverSimplify(fullName))
+                (ctx.hasCacheLocation() || ctx.checkSimpleDef() || ctx.syntaxChecking()) && !ctx.neverSimplify(fullName))
```
ghalliday: ctx.hasCacheLocation() is the test that needs modifying.

shamser:
@ghalliday

PR #11860, opened 2018-10-23
HPCC-20397 Return better message when preflight failed

wangkx:
1. When a node is not responsive to preflight SSH command, SSH
returns 255 error code. The WsMachine service may return
"SSH failed to connect 'ip'";
2. Run SSH command in foreground so that exception 127 can be
thrown when preflight script is not found.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/esp/services/ws_machine/ws_machineService.cpp
+++ b/esp/services/ws_machine/ws_machineService.cpp
@@ -1183,7 +1183,7 @@ int Cws_machineEx::runCommand(IEspContext& context, const char* sAddress, const 
         IFRunSSH * connection = createFRunSSH();
         connection->init(command.str(),NULL,NULL,NULL,m_SSHConnectTimeoutSeconds,0);
         // executed as single connection
-        connection->exec(sAddress,NULL,true);
+        connection->exec(sAddress, NULL, false);
         response.append(connection->getReplyText()[0]);
         exitCode = connection->getReply()[0];
         int len = response.length();
--- a/common/remote/rmtssh.cpp
+++ b/common/remote/rmtssh.cpp
@@ -453,6 +453,8 @@ public:
         }
         CriticalBlock block(sect);
         done.append(i);
+        if ((retcode == 255) && outbuf.isEmpty())
+            outbuf.setf("SSH failed to connect %s: %s", slaves.item(i), cmd.str());
         replytext.append(outbuf.str());
         reply.append((unsigned)retcode);
     }


wangkx:
@mckellyln Please review.

esp/services/ws_machine/ws_machineService.cpp
```diff
@@ -1189,6 +1189,9 @@ int Cws_machineEx::runCommand(IEspContext& context, const char* sAddress, const
         int len = response.length();
         if (len > 0 && response.charAt(--len) == '\n') // strip newline
           response.setLength(len);
+        if (len > 0 && !strnicmp(response.str(), "stderr:", 7))
```
mckellyln: Should you strip off the leading "stderr:" part from the response here ?

esp/services/ws_machine/ws_machineService.cpp
```diff
@@ -1199,6 +1202,12 @@ int Cws_machineEx::runCommand(IEspContext& context, const char* sAddress, const
     catch(IException* e)
     {
         exitCode = e->errorCode();
+        if (exitCode == 255)
```
mckellyln: Does it only throw to get here ?  Could it return ok with an exit code of 255 ?

mckellyln:
@wangkx a few questions.

esp/services/ws_machine/ws_machineService.cpp
wangkx: How about changing it to "Error:"?

esp/services/ws_machine/ws_machineService.cpp
wangkx: If it returns ok, the exit code will be set by CFRunSSH. I haven't found 255 exit code in the existing CFRunSSH. If a 255 exit code is added, the code may have different meaning. We may report it with the new meaning.

esp/services/ws_machine/ws_machineService.cpp
mckellyln: I think we should figure out why wait() returns 0 when cmd is not found (instead of 127) and if we can do that then use that logic to know if it failed, otherwise assume it ran ok.

wangkx: Yes, if IFRunSSH.exec() can send back 127, I should check that.

mckellyln: OK - so we know now it always returns 0 because it runs in the background.  If this runs in a new, separate thread then I think I would remove this code that checks for stderr: and just run it in the foreground (ie set background to false in exec() and then no -f flag on ssh cmdline) and it should automatically handle a return code of not zero.  But test it with a sleep in preflight to make sure that doesn't affect the GUI behavior if it takes a long time to complete.

wangkx: I did the test with a sleep in preflight and it doesn't affect the GUI behavior. I also found that, if I set background to false in exec() and the preflight cannot be found, the exec() will throw an exception with code 127.

esp/services/ws_machine/ws_machineService.cpp
```diff
@@ -1199,6 +1199,12 @@ int Cws_machineEx::runCommand(IEspContext& context, const char* sAddress, const
...
+        {
+            response.setf("SSH failed to connect %s", sAddress);
+            return exitCode;
```
mckellyln: Need to release e here ?

mckellyln:
@wangkx if -o LogLevel=QUIET is changed to -o LogLevel=ERROR then we would get output about failling to connect or whatever the error msg is.
But I don't know if its safe to change QUIET to ERROR always, it may be a problem on very old ssh/linux distros ???
If we do change it to ERROR then I would not special case the 255 exitCode.
If we don't change it to ERROR then I would append the "failed to connect" text to the response instead of just setting it, because the cmd is good to know if it fails to connect.
So:
1). research if its ok to change QUIET to ERROR
2). if so then remove the 255 exitCode case, if not then append "failed to connect" to response.
@Michael-Gardner fyi
-m

esp/services/ws_machine/ws_machineService.cpp
wangkx: Sorry, I missed that.

richardkchapman:
Should this target 7.0.2?

wangkx:
Yes, I will change the target.

wangkx:
@mckellyln I am not sure if it is ok to change QUIET to ERROR. Why not add an option?

mckellyln:
@wangkx this feels much better.
I would just add a comment to the code where LogLevel=QUIET is, something like:
```
// NB: -o LogLevel=ERROR might be better choice because it would echo ssh error text
// whereas QUIET does not, but leaving it as QUIET for now for compatibility
```

richardkchapman:
Should this target 7.0.2 ? 7.2.0 ?

wangkx:
A new PR will be created to target 7.0.2.

PR #11880, opened 2018-10-30
HPCC-20397 Return better message when preflight failed

wangkx:
1. When a node is not responsive to preflight SSH command, SSH
returns 255 error code. CFRunSSH should report the "SSH failed
to connect 'ip'";
2. Run SSH command in foreground so that exception 127 can be
thrown when preflight script is not found.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/esp/services/ws_machine/ws_machineService.cpp
+++ b/esp/services/ws_machine/ws_machineService.cpp
@@ -1183,7 +1183,7 @@ int Cws_machineEx::runCommand(IEspContext& context, const char* sAddress, const 
         IFRunSSH * connection = createFRunSSH();
         connection->init(command.str(),NULL,NULL,NULL,m_SSHConnectTimeoutSeconds,0);
         // executed as single connection
-        connection->exec(sAddress,NULL,true);
+        connection->exec(sAddress, NULL, false);
         response.append(connection->getReplyText()[0]);
         exitCode = connection->getReply()[0];
         int len = response.length();
--- a/common/remote/rmtssh.cpp
+++ b/common/remote/rmtssh.cpp
@@ -393,6 +393,8 @@ public:
             }
             if (cmdline.length()==0) {
                 // ssh
+                // NB: -o LogLevel=ERROR might be better choice because it would echo ssh error text
+                // whereas QUIET does not, but leaving it as QUIET for now for compatibility.
                 cmdline.appendf("%s -n -o LogLevel=QUIET -o StrictHostKeyChecking=%s ",usepssh?"pssh":"ssh",strict?"yes":"no -o UserKnownHostsFile=/dev/null");
                 if (!usepssh)
                     cmdline.append("-o BatchMode=yes ");
@@ -453,6 +455,8 @@ public:
         }
         CriticalBlock block(sect);
         done.append(i);
+        if ((retcode == 255) && outbuf.isEmpty())
+            outbuf.setf("SSH failed to connect %s: %s", slaves.item(i), cmd.str());
         replytext.append(outbuf.str());
         reply.append((unsigned)retcode);
     }


wangkx:
I create this PR to replace https://github.com/hpcc-systems/HPCC-Platform/pull/11860 which is targeted to 7.0.0. Please see the review comments on that PR. @mckellyln please review.

esp/services/ws_machine/ws_machineService.cpp
```diff
@@ -1183,7 +1183,7 @@ int Cws_machineEx::runCommand(IEspContext& context, const char* sAddress, const
         IFRunSSH * connection = createFRunSSH();
         connection->init(command.str(),NULL,NULL,NULL,m_SSHConnectTimeoutSeconds,0);
         // executed as single connection
-        connection->exec(sAddress,NULL,true);
+        connection->exec(sAddress, NULL, false);
```
mckellyln: sAddress here is a const char *, but it appears the definition is:
```
    void exec(
              const IpAddress &ip,
              const char *workdirname,
              bool _background)
```
Where the first arg is a const IpAddress &

mckellyln:
@wangkx looks good.  But have one question about exec() arg types.

esp/services/ws_machine/ws_machineService.cpp
mckellyln: Talked with @jakesmith about this - we should add an 'explicit' to the IpAddress() ctor:
```
    explicit IpAddress(const char *text)                         { ipset(text); }
```
In jsocket.hpp to make this more correct.  And then we can no longer pass in sAddress here, but need to first create an IpAddress type and init with the sAddress text and pass that in.

jakesmith: Thanks @mckellyln 

I think this PR is good to merge now.

mckellyln (approved):

mckellyln:
yes, ok - @richardkchapman ok to merge.

esp/services/ws_machine/ws_machineService.cpp
richardkchapman: What will be the impact of changing the "background" parameter to false? Does it mean multiple opens do not happen in parallel, for example? That could be an issue on a large cluster

richardkchapman:
@wangkx One question regarding performance

esp/services/ws_machine/ws_machineService.cpp
wangkx: If a user selects HPCC components from multiple IPs, the ws_machine code will create one thread for each IP and call this runCommand() (and this exec()) in each thread. So, the multiple exec()'s will be run in parallel.

mckellyln:
@richardkchapman you ok with answer ?

esp/services/ws_machine/ws_machineService.cpp
richardkchapman: So why did it say true before?

richardkchapman:
@wangkx One more question, just to check

esp/services/ws_machine/ws_machineService.cpp
wangkx: This is my guessing: the original code (https://github.com/hpcc-systems/HPCC-Platform/commit/993519e6030832340948813816945401fedfc2a5) does not realize that there is already multi-thread implementation.

mckellyln: I think invokeProgram() would run in the foreground ??

mckellyln: I think? the exec() code runs the command and waits for it to finish and collects the output, so not really understanding why it was originally set to background when exit code and output of command is wanted.
@wangkx could we test with >1 IP and a long running (ie adding sleep 30) preflight to make sure it does not take longer than with 1 IP ?

mckellyln: @richardkchapman fyi

richardkchapman:
@mckellyln Are you happy with this?

mckellyln:
@richardkchapman yes, I am ok to merge.

mckellyln (approved):

PR #11883, opened 2018-10-30
HPCC-20864 Avoid race condition fetching loop input dataset

jakesmith:
If a child query activity asked for the entire global loop
input dataset, a race could happen where the result on some
slaves was not prepared at the time the result was requested,
causing an assert to be hit.

Fix is to move the synchronization point, until after the
input is prepared, to do this required moving some of the
code out of a couple of 'execute' helper functions.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/thorlcr/activities/loop/thloop.cpp
+++ b/thorlcr/activities/loop/thloop.cpp
@@ -248,7 +248,8 @@ public:
                 if (sync(loopCounter))
                     break;
 
-                boundGraph->execute(*this, condLoopCounter, ownedResults, (IRowWriterMultiReader *)NULL, 0, extractBuilder.size(), extractBuilder.getbytes());
+                boundGraph->queryGraph()->executeChild(extractBuilder.size(), extractBuilder.getbytes(), ownedResults, NULL);
+
                 ++loopCounter;
                 if (barrier) // barrier passed once all slave graphs have completed
                 {
@@ -320,9 +321,17 @@ public:
         unsigned loopCounter = 1;
         for (;;)
         {
+            Owned<IThorGraphResults> results = queryGraph().createThorGraphResults(1);
+            unsigned condLoopCounter = (helper->getFlags() & IHThorGraphLoopArg::GLFcounter) ? loopCounter : 0;
+            IThorBoundLoopGraph *boundGraph = queryContainer().queryLoopGraph();
+            if (condLoopCounter)
+                boundGraph->prepareCounterResult(*this, results, condLoopCounter, 0);
+
             if (sync(loopCounter))
                 break;
-            queryContainer().queryLoopGraph()->execute(*this, (helper->getFlags() & IHThorGraphLoopArg::GLFcounter)?loopCounter:0, loopResults.get(), extractBuilder.size(), extractBuilder.getbytes());
+
+            boundGraph->queryGraph()->executeChild(extractBuilder.size(), extractBuilder.getbytes(), results, loopResults);
+
             ++loopCounter;
         }
     }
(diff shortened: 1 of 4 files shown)


jakesmith:
@ghalliday - please review

richardkchapman:
Will this have a performance impact?

jakesmith:
>Will this have a performance impact?

I would not expect any difference, it is only moving the synchronization point slightly later.

thorlcr/activities/loop/thloop.cpp
```diff
@@ -320,9 +321,17 @@ class CGraphLoopActivityMaster : public CLoopActivityMasterBase
...
+            unsigned condLoopCounter = (helper->getFlags() & IHThorGraphLoopArg::GLFcounter) ? loopCounter : 0;
+            IThorBoundLoopGraph *boundGraph = queryContainer().queryLoopGraph();
+            if (condLoopCounter)
+                boundGraph->prepareCounterResult(*this, results, condLoopCounter, 0);
+
             if (sync(loopCounter))
                 break;
-            queryContainer().queryLoopGraph()->execute(*this, (helper->getFlags() & IHThorGraphLoopArg::GLFcounter)?loopCounter:0, loopResults.get(), extractBuilder.size(), extractBuilder.getbytes());
+
+            boundGraph->queryGraph()->executeChild(extractBuilder.size(), extractBuilder.getbytes(), results, loopResults);
```
ghalliday: Previously this would have called graph->setLoopCounter(counter), now it is setting a counter result.  Was it previously wrong?

thorlcr/activities/loop/thloop.cpp
```diff
@@ -320,9 +321,17 @@ class CGraphLoopActivityMaster : public CLoopActivityMasterBase
         unsigned loopCounter = 1;
         for (;;)
         {
+            Owned<IThorGraphResults> results = queryGraph().createThorGraphResults(1);
+            unsigned condLoopCounter = (helper->getFlags() & IHThorGraphLoopArg::GLFcounter) ? loopCounter : 0;
+            IThorBoundLoopGraph *boundGraph = queryContainer().queryLoopGraph();
+            if (condLoopCounter)
+                boundGraph->prepareCounterResult(*this, results, condLoopCounter, 0);
```
ghalliday: Should the last parameter be a 2 to match process()?

ghalliday (commented):
@jakesmith a couple of questions

thorlcr/activities/loop/thloop.cpp
jakesmith: Should be same as before.
It set the loop counter result via prepareCounterResult() and still does (line 243 above).
It used to call setLoopCounter inside the now removed IThorBoundLoopGraph::execute() method, now it is called as part of prepareCounterResult()

thorlcr/activities/loop/thloop.cpp
jakesmith: This is as it was before but it was a bit confusing.
It is 2 in the LOOP case, and 0 in the GRAPH case.

And I'm hoping that regression suite tests are validating it is correct.
Having had a quick look at roxie code, I think 0 for GRAPH and 2 for LOOP is correct or at least same in Roxie. Can you verify?

The code above is for GRAPH and before this change it used to call one of the 2nd form of IThorBoundLoopGraph::execute which used to set the counter result as result 0.

jakesmith:
@ghalliday - can you look at my answers please?

PR #11907, opened 2018-11-08
HPCC 20752 Build and run HPCC in VS 2015

gaojx:
## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization


dali/dafilesrv/dafilesrv.cpp
```diff
@@ -386,6 +386,7 @@ int main(int argc,char **argv)
     unsigned throttleSlowQueueLimit = DEFAULT_SLOWCMD_THROTTLEQUEUELIMIT;
 
     unsigned dedicatedRowServicePort = DEFAULT_ROWSERVICE_PORT;
+	StringAttr rowServiceConfiguration = defaultRowSericeConfiguration;
```
ghalliday: more tabs

ghalliday (commented):
A few comments from a brief initial scan.  I will look again in a bit.
Do we have details on how to build the dependent libraries in a wiki?

cmake_modules/FindLIBXML2.cmake
```diff
@@ -48,7 +48,7 @@ if (NOT LIBXML2_FOUND)
   ENDIF()
 
   if (USE_NATIVE_LIBRARIES)
-    # if we didn't find in externals, look in system include path
+    # if we didn't find in extals, look in system include path
```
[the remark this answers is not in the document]
gaojx: Yes, will fix

common/remote/sockfile.cpp
```diff
@@ -4508,7 +4508,7 @@ void checkExpiryTime(IPropertyTree &metaInfo)
     {
         expiryTimeDt.setString(expiryTime);
     }
-    catch (IException *e)
+    catch (IException * /*e*/)
     {
         throw createDafsException(DAFSERR_cmdstream_invalidexpiry, "createRemoteActivity: invalid expiry specification");
```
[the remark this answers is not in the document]
gaojx: Agree.

configuration/configmgr/configmgrlib/EnvSupportLib.cpp
```diff
@@ -29,21 +29,21 @@ EnvSupportLib::EnvSupportLib(const std::string &libName, EnvironmentMgr *pEnvMgr
         throw (ParseException(msg));
     }
 
-    m_libHandle = dlopen(m_libName.c_str(), RTLD_LAZY);
-    if (m_libHandle == nullptr)
+	m_libHandle = LoadSharedObject(m_libName.c_str(),true,false);
```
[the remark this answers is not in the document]
gaojx: This is one issue with VS tab setting is not system wide: that is each time a new solution is generated, we needs to set the tab setting.

dali/dafilesrv/dafilesrv.cpp
gaojx: fixed

cmake_modules/commonSetup.cmake
```diff
@@ -378,6 +384,20 @@ IF ("${COMMONSETUP_DONE}" STREQUAL "")
            add_definitions(/J)
         endif ()
     endif ()
+
+
+    #===  WARNING ======
+    # Temporary disable warnings. Reenable them one by one to reexamine and fix them.
+    IF (NOT ALL_WARNINGS_ON)
+      set (WARNINGS_IGNORE "/wd4267 /wd4244 /wd6340 /wd6297 /wd4018 /wd4302 /wd4311 /wd4320 /wd4800")     # data conversion warnings
+       set (WARNINGS_IGNORE  "${WARNINGS_IGNORE} /wd4251 /wd4275")      # dll-interface for used by clients
```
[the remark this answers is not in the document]
gaojx: fixed

configuration/configmgr/configmgrlib/EnvSupportLib.cpp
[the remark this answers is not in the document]
ghalliday: It doesn't align perfectly when I look at it in chrome.  Also, you can double check by trying to select a single character in the indentation.  If it is a tab it selects a block of characters instead.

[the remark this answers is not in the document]
ghalliday: A pull request that replaces tabs in files that are predominately indented with spaces would be fine.  Quite a few of the tabs that I found with a similar search are in 3rd party code (e.g., cassandra/kafka) or in files submitted by external contributors where the whole file uses tabs.  I wouldn't replace those.

ghalliday:
I tried building the branch on my machine, and came across the following problems:

a) The title of the commit message isn't quite the right format it should start HPCC-20752

b) Problems with references to the SSL ,library:
system\security\cryptohelper\cryptocommon.hpp(43): error C2065: 'EVP_PKEY_CTX': undeclared identifier
security\cryptohelper\cryptocommon.hpp(43): error C2065: 'EVP_PKEY_CTX_free': undeclared identifier

What is the mechanism for getting a compatible version of the SSL libraries.  (These will need to be installed by anyone using the system on windows, and the build box that creates the client tools.)

c) Various other errors.  I think these are all follow on errors because jlib cannot be built because of (b).

system/security/cryptohelper/pke.hpp
```diff
@@ -40,8 +40,10 @@ class jlib_decl CLoadedKey : public CSimpleInterfaceOf<IInterface>
     void loadKeyFromMem(const char *key);
     bool loadKeyFromFile(const char *keyFile);
     void finalize(RSA *rsaKey, const char *keyName);
+
 public:
     CLoadedKey() { }
+	CLoadedKey(const CLoadedKey&) = delete;
```
ghalliday: indent: tabs

ghalliday (commented):
I think this is very close to being ready to take.  I think they are good changes, and will also make my life easier.
The main question is what to do about the external dependencies.  If we can come up with a list of packages, and how to download them/build them so that that platform can be built that would be great.

ghalliday:
@gaojx a few more comments above.  I would also be tempted to target this to candidate-7.2.0, rather than master.  The earlier it is merged, the better.  It might possibly be a candidate for 7.0.2, but I will need to check it carefully and check with Richard before definitely deciding that.  I suspect the change in external dependencies will mean it should be 7.2.

cmake_modules/FindLIBXML2.cmake
```diff
@@ -24,7 +24,7 @@
 
 if (NOT LIBXML2_FOUND)
   IF (WIN32)
-    SET (libxml2_libs "xml2 libxml2")
+    SET (libxml2_libs "libxml2-2")
```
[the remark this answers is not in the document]
gaojx: I don't have a good solution for this. Right now, we use the network share directory provided by Ming Wang (the build master), where all pre-compiled lib/dll are stored. Most of the libraries are already there. I made some change to it: add openssl 1.0.2h which build with current HPCC code, and add libxml2.

CMakeLists.txt
```diff
@@ -181,9 +181,9 @@ elseif ( NOT MAKE_DOCS_ONLY )
     HPCC_ADD_SUBDIRECTORY (services "PLATFORM")
     HPCC_ADD_SUBDIRECTORY (thorlcr "PLATFORM")
     HPCC_ADD_SUBDIRECTORY (testing)
-    if ( NOT WIN32 )
+ #   if ( NOT WIN32 )
```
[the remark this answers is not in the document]
gaojx: Agree.

gaojx:
This is address all the questions about getting the dependencies libraries on Windows.

Right now, Ming created a network shared directory in RELX internal network with all the prebuild lib/dll. 
You can get the URI and user credential to the shared directory. Just call the whole directory locally and point to in in you cmake command (-DEXTERNALS_DIRECTORY=...). 

The think the issue is more of legal question: if we are allowed to make it public (e.g., in a git repository), and/or distribute it as a package? I see websites where you can download prebuilt openssl libraries for various platform (include Windows x86 and x64).  I don't see why we can't do similarly legally? Maybe all we need is to have an approval from legal department,  add some legal disclaimer etc (whatever  the legal feels we need to do).

ghalliday:
I am working on a solution for the dependencies using something called vcpkg.  So far it is looking promising, but requires a bit more work.

ghalliday:
I am working on a solution for the dependencies using something called vcpkg.  So far it is looking promising, but requires a bit more work.

gaojx:
I think we should merge the change now, so the other change #11910, #11926 can be merged.

richardkchapman:
Should target 7.2.0, not earlier

gaojx:
7.2.0 is fine. Is there anything I need to do to make the happens (i.e., set target to 7.2.0)?

richardkchapman:
Edit the PR to set the target to be 7.2.0. You may also need to do

 rebase --onto candidate-7.2.0 master

and force-push your branch

gaojx:
Rebased to candidate-7.2.0. Ready to merge.

CMakeLists.txt
```diff
@@ -237,8 +235,8 @@ if(TOP_LEVEL_PROJECT)
         set(PACKAGE_FILE_NAME_PREFIX  "hpccsystems-clienttools-${projname}")
     endif()
     if(NOT "${CUSTOM_PACKAGE_SUFFIX}" STREQUAL "")
-	set(CPACK_PACKAGE_NAME "${CPACK_PACKAGE_NAME}-${CUSTOM_PACKAGE_SUFFIX}")
-	set(PACKAGE_FILE_NAME_PREFIX "${PACKAGE_FILE_NAME_PREFIX}-${CUSTOM_PACKAGE_SUFFIX}")
+    set(CPACK_PACKAGE_NAME "${CPACK_PACKAGE_NAME}-${CUSTOM_PACKAGE_SUFFIX}")
```
richardkchapman: Why this change? Looks like indentation was correct before and is now wrong?

cmake_modules/commonSetup.cmake
```diff
@@ -77,7 +77,13 @@ IF ("${COMMONSETUP_DONE}" STREQUAL "")
   else()
     option(USE_NUMA "Configure use of numa" ON)
   endif()
-  option(USE_NATIVE_LIBRARIES "Search standard OS locations for thirdparty libraries" ON)
+
+  IF (WIN32)
+     option(USE_NATIVE_LIBRARIES "thirdparty libraries are in EXTERNALS_DIRECTORY" OFF)
```
richardkchapman: The text of this option says the opposite of what the option does, I think. Seems to me the option's text should remain the same for both platforms, it's just the default value you are wanting to change.

cmake_modules/commonSetup.cmake
```diff
@@ -378,6 +384,20 @@ IF ("${COMMONSETUP_DONE}" STREQUAL "")
...
+
+    #===  WARNING ======
+    # Temporary disable warnings. Reenable them one by one to reexamine and fix them.
+    IF (NOT ALL_WARNINGS_ON)
+      set (WARNINGS_IGNORE "/wd4267 /wd4244 /wd6340 /wd6297 /wd4018 /wd4302 /wd4311 /wd4320 /wd4800") # data conversion warnings
+      set (WARNINGS_IGNORE "${WARNINGS_IGNORE} /wd4251 /wd4275") # dll-interface for used by clients
+      set (WARNINGS_IGNORE "${WARNINGS_IGNORE} /wd6246")   # local variable hidden by outter scope
+      set (WARNINGS_IGNORE "${WARNINGS_IGNORE} /wd6031")   # Return value ignored
+      set (WARNINGS_IGNORE "${WARNINGS_IGNORE} /wd4005")   # MACRO redef: same value
+      set (CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} ${WARNINGS_IGNORE}")
```
richardkchapman: What about release builds?

Some of these warnings seem useful - what is the plan for re-enabling them?

common/remote/sockfile.cpp
```diff
@@ -1,4 +1,4 @@
-/*##############################################################################
+﻿/*##############################################################################
```
richardkchapman: Any idea what changed here??

cmake_modules/commonSetup.cmake
```diff
@@ -378,6 +384,20 @@ IF ("${COMMONSETUP_DONE}" STREQUAL "")
...
         endif ()
     endif ()
+
+
+    #===  WARNING ======
+    # Temporary disable warnings. Reenable them one by one to reexamine and fix them.
+    IF (NOT ALL_WARNINGS_ON)
+      set (WARNINGS_IGNORE "/wd4267 /wd4244 /wd6340 /wd6297 /wd4018 /wd4302 /wd4311 /wd4320 /wd4800") # data conversion warnings
+      set (WARNINGS_IGNORE "${WARNINGS_IGNORE} /wd4251 /wd4275") # dll-interface for used by clients
+      set (WARNINGS_IGNORE "${WARNINGS_IGNORE} /wd6246")   # local variable hidden by outter scope
```
richardkchapman: typo outter should say outer

dali/dafilesrv/dafilesrv.cpp
```diff
@@ -1,4 +1,4 @@
-/*##############################################################################
+﻿/*##############################################################################
```
richardkchapman: ?

dali/dafilesrv/dafilesrv.cpp
```diff
@@ -34,7 +34,7 @@
 
 static const bool defaultRowServiceOnStdPort = true;
 static const bool defaultDedicatedRowServiceSSL = false;
-
+static const char* defaultRowSericeConfiguration = "RowSvc";
```
richardkchapman: this change seems somewhat unrelated

system/include/platform.h
```diff
@@ -171,6 +171,8 @@ typedef memsize_t rowsize_t;
 #define GetSharedProcedure(h,name) GetProcAddress(h,(char *)name)
 #define LoadSucceeded(h)           ((memsize_t)h >= 32)
 #define GetSharedObjectError()     GetLastError()
+#define GetSharedObjectErrorString() strerror(errno)
+#define CloseSharedObject(h)        FreeLibrary(h)
```
richardkchapman: This is not a good idea - should be calling FreeSharedObject (in jutil)

richardkchapman (changes requested):
Mostly minor comments

richardkchapman:
@gaojx

CMakeLists.txt
gaojx: It was a tab. Replaced with 4 spaces, should have been 8 spaces.

cmake_modules/commonSetup.cmake
gaojx: The message tells where the 3rd party library is when the USE_NATIVE_LIBRARIES is set to default (which is OFF on Windows, and ON on other platform). I agree it is confusion, and will change it to be the same.

cmake_modules/commonSetup.cmake
gaojx: Good point. Using CMAKE_CXX_FLAGS instead will work for all build types.
We can open one ticket per warning to re-enable them.

common/remote/sockfile.cpp
gaojx: The new version adds an EF BB BF header (means UTF8 encoding). It was probably added by some editor. Will remove it since no other cpp/hpp file has it.

cmake_modules/commonSetup.cmake
gaojx: Fixed

dali/dafilesrv/dafilesrv.cpp
gaojx: UTF8 BAM (byte order mark) header. Will remove.

dali/dafilesrv/dafilesrv.cpp
gaojx: yes. will remove

system/include/platform.h
gaojx: Good suggestion. Will do

dali/dafilesrv/dafilesrv.cpp
richardkchapman: doesn't look like you removed it

cmake_modules/commonSetup.cmake
```diff
@@ -1,4 +1,4 @@
-################################################################################
+﻿################################################################################
```
richardkchapman: I think these changes are caused by BOM markers - probably best to avoid them

richardkchapman:
A couple of minor comments. @ghalliday Are you happy with this, particularly regarding the external components part?

cmake_modules/commonSetup.cmake
gaojx: Yes, this change remove the BOM markers.

dali/dafilesrv/dafilesrv.cpp
gaojx: The variable is used by the code in the the source file, so we can't remove it.

ghalliday:
@gaojx 
The external components need to be sensibly accessible.  I tried downloading them from the shared build directory, but it takes forever.

There seems to be a repository 
https://github.com/hpcc-systems/HPCC-Windows-Dependencies
which has some of the dependencies.  That should be updated with the new directory, so it is at least easy for anyone within the company to get the latest dependencies.  (Without this I will not be able to build the system any more.)

I'll review the rest in the meantime.

ghalliday:
I think it is ready to merge once the dependencies are available in a repository.

PR #11910, opened 2018-11-09
HPCC-20926 Remove the rest of warnings

gaojx:
## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization


gaojx:
This is the continue of HPCC-20752 to remove the rest of compile warnings in VS2015.  

Most of the changes is related to the isdigit, isalpha etc function family. They were defined as macro in platform.h, but the same functions are also defined in boost as member function of a class. So then the platform.h is included before the boost header file, the following boost code:
       obj->isdigit(param1, param2) 
The preprocessor  will try expand isdigit() as macro, but isdigit() macro in platform.h can only take one parameter, thus gives compiling error. 

The solution is to use inline function, instead of macro.

@timothyklemm @asselitx please code review.

system/mp/mpbuff.hpp
```diff
@@ -78,7 +78,7 @@ class mp_decl CMessageBuffer: public MemoryBuffer
                 data.append(", ");
             data.append((unsigned)(byte)toByteArray()[i]);
         }
-        return buf.appendf("CMessageBuffer(%8X) tag=%d, sender=%s, replytag=%d, size== %d, data head = %s", (int)(long)toByteArray(), (int)tag, sender.getUrlStr(ep).str(), (int)replytag, length(), data.str());
+        return buf.appendf("CMessageBuffer(%P) tag=%d, sender=%s, replytag=%d, size== %d, data head = %s", toByteArray(), (int)tag, sender.getUrlStr(ep).str(), (int)replytag, length(), data.str());
```
asselitx: Is this supposed to be lower-case 'p' for 'pointer' type? I see there is also a warning thrown by smoketest on this line.

gaojx: yes, fixed.

system/include/platform.h
```diff
@@ -502,6 +481,20 @@ typedef unsigned __int64 timestamp_type;
...
+inline bool isalpha_char(char c) { return isalpha((unsigned char)c) != 0; }
+inline bool isprint_char(char c) { return isprint((unsigned char)c) != 0; }
+inline bool isblank_char(char c) { return isblank((unsigned char)c) != 0; }
+inline bool isupper_char(char c) { return isupper((unsigned char)c) != 0; }
+inline bool islower_char(char c) { return islower((unsigned char)c) != 0; }
+inline bool isalnum_char(char c) { return isalnum((unsigned char)c) != 0; }
+inline bool iscntrl_char(char c) { return iscntrl((unsigned char)c) != 0; }
+inline bool isxdigit_char(char c) { return isxdigit((unsigned char)c) != 0; }
+inline int toupper_char(char c) { return toupper((unsigned char)c); }
+inline int tolower_char(char c) { return tolower((unsigned char)c); }
```
asselitx: I don't have any examples, but have you considered any unintended issues that may arise due to some subtle differences between these new definitions and the old? It appears to be functionally the same but these are so widely used there could be interesting corner cases.

1. Previously the macros were defined `if (_MSC_VER>=1400)`. Could there be any code previously using the macro that may now be using the implementation in ctype.h? Could that make a difference in any case? (I don't _think_ so...)
2. Why cast now to `(unsigned char)` instead of `(const unsigned char)` as before?
3. Why return bool now instead of int?

cmake_modules/commonSetup.cmake
```diff
@@ -378,6 +384,22 @@ IF ("${COMMONSETUP_DONE}" STREQUAL "")
...
+    IF (NOT ALL_WARNINGS_ON)
+      set (WARNINGS_IGNORE "/wd4267 /wd4244 /wd6340 /wd6297 /wd4018 /wd4302 /wd4311 /wd4320 /wd4800")     # data conversion warnings
+      set (WARNINGS_IGNORE  "${WARNINGS_IGNORE} /wd4251 /wd4275")      # dll-interface for used by clients
+      set (WARNINGS_IGNORE  "${WARNINGS_IGNORE} /wd6246")   # local variable hidden by outter scope
+      set (WARNINGS_IGNORE  "${WARNINGS_IGNORE} /wd6031")   # Return value ignored
+      set (WARNINGS_IGNORE  "${WARNINGS_IGNORE} /wd4005")   # MACRO redef: same value
+      set (WARNINGS_IGNORE  "${WARNINGS_IGNORE} /wd4503")   # decorated name length exceeded, name was truncated
+      
+      set (CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} ${WARNINGS_IGNORE}")
+    ENDIF()
```
asselitx: Are we setting ourselves up for ignoring these forever, or do we have a ticket open to address them?

dali/treeview/CMakeLists.txt
```diff
@@ -44,6 +44,9 @@ if (WIN32)
         )
 
     add_definitions("-D_AFXDLL")
+    set (CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS} /wd4996" )
+	set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /wd4996" )
```
asselitx: Is this another case where we should open a ticket to address the warning - 'deprecated method in use' ?

ecl/hql/hqlgram.y
```diff
@@ -13010,7 +13010,7 @@ yybackup:
   if (yyn == YYPACT_NINF)
     goto yydefault;
 
-yyprocesschar:
+// yyprocesschar: // unsed label
```
asselitx: This was throwing a warning?

system/include/platform.h
gaojx: All good questions. These inline functions are taken from scapps, and they have been used for +10 years.

From https://en.cppreference.com/w/cpp/string/byte/toupper:
Like all other functions from <cctype>, the behavior of std::toupper is undefined if the argument's value is neither representable as unsigned char nor equal to EOF. To use these functions safely with plain chars (or signed chars), the argument should first be converted to unsigned char.

cmake_modules/commonSetup.cmake
gaojx: Definitely not forever. Ideally we don't need to disable any warnings, but it just too much to deal with as one bug (over 40K warnings). 

Yes, we will have ticket to address them. We may have one ticket to enable one or few warnings in a manageable way.

dali/treeview/CMakeLists.txt
gaojx: We could, but as I far as I know, this is used by ConfigEnv only and ConfigEnv is going away so this may not be on our priority.

CMakeLists.txt
```diff
@@ -181,9 +181,9 @@ elseif ( NOT MAKE_DOCS_ONLY )
     HPCC_ADD_SUBDIRECTORY (services "PLATFORM")
     HPCC_ADD_SUBDIRECTORY (thorlcr "PLATFORM")
     HPCC_ADD_SUBDIRECTORY (testing)
-    if ( NOT WIN32 )
+ #   if ( NOT WIN32 )
```
richardkchapman: Please don't comment out lines - remove them

cmake_modules/FindLIBXML2.cmake
```diff
@@ -48,7 +48,7 @@ if (NOT LIBXML2_FOUND)
   ENDIF()
 
   if (USE_NATIVE_LIBRARIES)
-    # if we didn't find in externals, look in system include path
+    # if we didn't find in extals, look in system include path
```
richardkchapman: Was this a deliberate change? Looked more correct before to me

common/fileview2/fvtransform.cpp
```diff
@@ -637,11 +637,11 @@ unsigned MappingParser::lexToken()
...
             tokenType = TokId;
         }
-        else if (isdigit(next) ||
-                 ((next == '-') && (cur+1 < lenInput) && isdigit(buffer[cur+1])))
+        else if (isdigit_char(next) ||
+                 ((next == '-') && (cur+1 < lenInput) && isdigit_char(buffer[cur+1])))
         {
             cur++;
-            while ((cur < lenInput) && (isdigit(buffer[cur])))
+            while ((cur < lenInput) && (isdigit_char(buffer[cur])))
```
richardkchapman: In the interests of making it possible to review this change and get it merged without overwhelming reviewers, I would recommend breaking it into smaller changes each of which can be reviewed and verified independently. For example, a commit that contained nothing but a global rename of these isXXX functions would be very easy to get accepted and merged

common/remote/sockfile.cpp
```diff
@@ -4508,7 +4508,7 @@ void checkExpiryTime(IPropertyTree &metaInfo)
     {
         expiryTimeDt.setString(expiryTime);
     }
-    catch (IException *e)
+    catch (IException * /*e*/)
```
richardkchapman: No. This is an example of why I am very wary of huge "just fix warnings so it must be safe" pull requests.

The warning that you have suppressed here (e unused) is actually showing up a bug (that e is leaked).

I cannot review a change this large and expect to spot all such issues. It needs to be broken up into manageable chunks, each one fixing a single category of warning that can be verified by inspection or automated compare.

common/remote/sockfile.cpp
```diff
@@ -4537,7 +4537,7 @@ void verifyMetaInfo(IPropertyTree &actNode, bool authorizedOnly, const IProperty
     Owned<IPropertyTree> metaInfoEnvelope = createPTree(decompressedMetaInfoMb);
 
     Owned<IPropertyTree> metaInfo;
-#if defined(_USE_OPENSSL) && !defined(_WIN32)
```
richardkchapman: This change is not a warning fix

CMakeLists.txt
gaojx: agree. fixed

common/fileview2/fvtransform.cpp
gaojx: Good suggestion. Will try

common/remote/sockfile.cpp
gaojx: Good point. Add the release of e.

common/remote/sockfile.cpp
gaojx: Right, this change is actually in another ticket, so this will not be here where rebase to the version with that ticket.

ghalliday:
@gaojx this looks like it has got stuck.  Many of the changes look useful, is it possible to go back and reapply them, but in smaller units so they can be merged?

richardkchapman:
Closing as stalled

PR #11920, opened 2018-11-14
HPCC-20957 Allow DESDL gateway transformations

timothyklemm:
- Recognize Gateways[@transformTarget]/Gateway as a meaningful xpath in the
  dynamic method configuration.
- Transform Gateway elements from <Gateway name="my_name" password=
  "my_encrypted_password" url="the_scheme://the_url" username="my_username"/> to
  <row><ServiceName>my_name</ServiceName><URL>the_scheme://my_username:
  my_decrypted_password@the_url</URL></row>.
- Place the transformed elements at the path indicated by @transformTarget.
- Leave dynamic method configuration unchanged, so existing deployments are
  unaffected.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/esp/services/esdl_svc_engine/esdl_binding.cpp
+++ b/esp/services/esdl_svc_engine/esdl_binding.cpp
@@ -843,7 +843,7 @@ void EsdlServiceImpl::handleFinalRequest(IEspContext &context,
     if(isroxie)
     {
         const char *tgtQueryName = tgtcfg->queryProp("@queryname");
-        if (tgtQueryName && *tgtQueryName)
+        if (!isEmptyString(tgtQueryName))
         {
             soapmsg.append("<soap:Body><").append(tgtQueryName).append(">");
             soapmsg.appendf("<_TransactionId>%s</_TransactionId>", context.queryTransactionID());
@@ -853,6 +853,28 @@ void EsdlServiceImpl::handleFinalRequest(IEspContext &context,
 
             soapmsg.append(req.str());
             soapmsg.append("</").append(tgtQueryName).append("></soap:Body>");
+
+            auto cfgGateways = tgtcfg->queryBranch("Gateways");
+            if (cfgGateways)
+            {
+                auto baseXpath = cfgGateways->queryProp("@legacyTransformTarget");
+                if (!isEmptyString(baseXpath))
+                {
+                    Owned<IPTree> gws = createPTree("gateways", 0);
+                    Owned<IPTree> soapTree = createPTreeFromXMLString(soapmsg.append("</soap:Envelope>"), ipt_ordered);
+                    StringBuffer xpath(baseXpath);
+
+                    EsdlBindingImpl::transformGatewaysConfig(tgtcfg, gws, "row");
+                    xpath.replaceString("{$query}", tgtQueryName);
+                    xpath.replaceString("{$method}", mthdef.queryMethodName());
+                    xpath.replaceString("{$service}", srvdef.queryName());
+                    xpath.replaceString("{$request}", mthdef.queryRequestType());
+                    mergePTree(ensurePTree(soapTree, xpath), gws);
+                    toXML(soapTree, soapmsg.clear());
+                    soapmsg.setLength(strstr(soapmsg, "</soap:Envelope>") - soapmsg.str());
+                    soapmsg.trim();
+                }
+            }
         }
         else
             throw makeWsException( ERR_ESDL_BINDING_INTERNERR, WSERR_SERVER, "ESP",
@@ -871,7 +893,7 @@ void EsdlServiceImpl::handleFinalRequest(IEspContext &context,
     soapmsg.append("</soap:Envelope>");
 
     const char *tgtUrl = tgtcfg->queryProp("@url");
-    if (tgtUrl && *tgtUrl)
+    if (!isEmptyString(tgtUrl))
     {
         if (crt)
         {
@@ -3119,7 +3141,7 @@ int EsdlBindingImpl::onGetSampleXml(bool isRequest, IEspContext &ctx, CHttpReque
     return 0;
 }
 
-void EsdlBindingImpl::transformGatewaysConfig( IPropertyTree* srvcfg, IPropertyTree* forRoxie )
+void EsdlBindingImpl::transformGatewaysConfig( IPropertyTree* srvcfg, IPropertyTree* forRoxie, const char* altElementName )
 {
     // Do we need to handle 'local FQDN'? It doesn't appear to be in the
     // Gateway element. Not sure where it's set but the RemoteNSClient
@@ -3131,6 +3153,8 @@ void EsdlBindingImpl::transformGatewaysConfig( IPropertyTree* srvcfg, IPropertyT
         cfgIter->first();
         if( cfgIter->isValid() )
         {
+            const char* treeName = (!isEmptyString(altElementName) ? altElementName : "Gateway");
+
             while( cfgIter->isValid() )
             {
                 IPropertyTree& cfgGateway = cfgIter->query();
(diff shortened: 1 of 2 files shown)


timothyklemm:
@asselitx and @RussWhitehead Please review. @afishbeck is aware this change is being made.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -3095,6 +3118,8 @@ void EsdlBindingImpl::transformGatewaysConfig( IPropertyTree* srvcfg, IPropertyT
         cfgIter->first();
         if( cfgIter->isValid() )
         {
+            const char* treeName = (altElementName && *altElementName ? altElementName : "Gateway");
```
RussWhitehead: I think if altElementName is null, treeName will not be set. Regardeless, I recommend using isEmptyString()

RussWhitehead (changes requested):

esp/services/esdl_svc_engine/esdl_binding.cpp
timothyklemm: I believe the condition as written is equivalent to !isEmptyString(altElementName), but agree that the function is clearer. Will also change three similar occurrences (1 new, 2 existing) in handleFinalRequest.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -860,6 +860,29 @@ void EsdlServiceImpl::handleFinalRequest(IEspContext &context,
 
             soapmsg.append(req.str());
             soapmsg.append("</").append(tgtQueryName).append("></soap:Body>");
+
+            auto cfgGateways = tgtcfg->queryBranch("Gateways");
+            if (cfgGateways)
+            {
+              auto baseXpath = cfgGateways->queryProp("@transformTarget");
```
timothyklemm: Whitespace in this block is incorrect. Will fix.

RussWhitehead:
@timothyklemm Looks good, but you need to fill out the checklist at the top of this PR

asselitx:
@timothyklemm Reviewed, looks good.

timothyklemm:
@afishbeck If this change is acceptable, can you request a merge. It currently targets 7.0.2, but 7.2.0 will be OK if the change is not suitable for a point release.

asselitx:
@timothyklemm Reviewed, looks good.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -860,6 +860,29 @@ void EsdlServiceImpl::handleFinalRequest(IEspContext &context,
...
+
+            auto cfgGateways = tgtcfg->queryBranch("Gateways");
+            if (cfgGateways)
+            {
+                auto baseXpath = cfgGateways->queryProp("@transformTarget");
+                if (!isEmptyString(baseXpath))
+                {
+                    Owned<IPTree> gws = createPTree("gateways", 0);
+                    VStringBuffer tmpSoap("%s</soap:Envelope>", soapmsg.str());
+                    Owned<IPTree> soapTree = createPTreeFromXMLString(tmpSoap);
```
afishbeck: I think you should preserve order using the ipt_ordered  option.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -860,6 +860,29 @@ void EsdlServiceImpl::handleFinalRequest(IEspContext &context,
...
             soapmsg.append("</").append(tgtQueryName).append("></soap:Body>");
+
+            auto cfgGateways = tgtcfg->queryBranch("Gateways");
+            if (cfgGateways)
+            {
+                auto baseXpath = cfgGateways->queryProp("@transformTarget");
+                if (!isEmptyString(baseXpath))
+                {
+                    Owned<IPTree> gws = createPTree("gateways", 0);
+                    VStringBuffer tmpSoap("%s</soap:Envelope>", soapmsg.str());
```
afishbeck: I don't think you need a temporary "tmpSoap" variable.  Seems like you can do all the work in soapmsg, since you always swapWith anyway.  Just add the Envelope, transform, toXML, and then remove the Envelope all with soapmsg, no?  I may be missing something.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -860,6 +860,29 @@ void EsdlServiceImpl::handleFinalRequest(IEspContext &context,
...
+                    StringBuffer xpath(baseXpath);
+
+                    EsdlBindingImpl::transformGatewaysConfig(tgtcfg, gws, "row");
+                    xpath.replaceString("{$query}", tgtQueryName);
+                    xpath.replaceString("{$method}", mthdef.queryMethodName());
+                    xpath.replaceString("{$service}", srvdef.queryName());
+                    xpath.replaceString("{$request}", mthdef.queryRequestType());
+                    mergePTree(ensurePTree(soapTree, xpath), gws);
+                    toXML(soapTree, tmpSoap.clear());
+                    tmpSoap.replaceString("</soap:Envelope>", "").trim();
```
afishbeck: replaceString seems kind of expensive here.  Can't you just remove from the end?  toXML is pretty predictable in output format.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -860,6 +860,29 @@ void EsdlServiceImpl::handleFinalRequest(IEspContext &context,
...
+                    Owned<IPTree> gws = createPTree("gateways", 0);
+                    VStringBuffer tmpSoap("%s</soap:Envelope>", soapmsg.str());
+                    Owned<IPTree> soapTree = createPTreeFromXMLString(tmpSoap);
+                    StringBuffer xpath(baseXpath);
+
+                    EsdlBindingImpl::transformGatewaysConfig(tgtcfg, gws, "row");
+                    xpath.replaceString("{$query}", tgtQueryName);
+                    xpath.replaceString("{$method}", mthdef.queryMethodName());
+                    xpath.replaceString("{$service}", srvdef.queryName());
+                    xpath.replaceString("{$request}", mthdef.queryRequestType());
```
afishbeck: FUTURE: Just a note that in the future we should create a common routine to do these replacements all in one call,  more efficiently.  For example pass it a IPropertyTree of the variables and replace them all in one pass.  We can then create the property object once for the entire transaction and use it all over the place.  The routine would be shared by this code and the ESDL transform code.  Not a big fan of tons of replaceString operations.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -860,6 +860,29 @@ void EsdlServiceImpl::handleFinalRequest(IEspContext &context,
 
             soapmsg.append(req.str());
             soapmsg.append("</").append(tgtQueryName).append("></soap:Body>");
+
+            auto cfgGateways = tgtcfg->queryBranch("Gateways");
+            if (cfgGateways)
+            {
+                auto baseXpath = cfgGateways->queryProp("@transformTarget");
```
afishbeck: It might be a better reminder of what this attribute does if we call it the "legacyTransformTarget"?  Or if you can think of another good name that is more indicative of the exact format we are producing.  We are calling a very specific transform here, not a script-able one.

afishbeck (changes requested):
A few comments, and one note for the future.

afishbeck:
@timothyklemm a few suggestions.  If we need to move some of these suggestions post RIN we can discuss that.

richardkchapman:
Needs rebasing, and smoketest issue needs investigation

esp/services/esdl_svc_engine/esdl_binding.cpp
timothyklemm: Agree with the addition of a common routine. There is a version of StringBuffer with an efficient multiple string replacement function that could be a good addition. It works on a std::vector of pairs, where the first pair field is the text to be found and the second is the replacement text. In this context, we could populate the vector once and apply the replacements where needed.

esp/services/esdl_svc_engine/esdl_binding.cpp
timothyklemm: Will change to "legacyTransformTarget" since I don't have a better idea.

timothyklemm:
@afishbeck I've updated based on your comments.

timothyklemm:
@richardkchapman The smoketest issue was a conflict with an unrelated change by @afishbeck that merged after this PR was created.

afishbeck (approved):
@timothyklemm looks good.  You can squash the two commits.  As discussed we can optimize the variable replacement in future.

timothyklemm:
@richardkchapman please merge

PR #11922, opened 2018-11-16
HPCC-20960 Echo the transformed DESDL request in the response

timothyklemm:
Embed the full request, with minor modifications, into the response to be
used by log agents.
- Replace the roxie query name with a generic value, so service level agent
  configurations can work with multiple queries. Retain the original name
  as an attribute in case it is needed.
- Replace the ESP request name with a generic value, so service level agent
  configurations can work with multiple queries. Retain the original name
  as an attribute in case it is needed.
- Replace the colon delimiter separating the custom request transform's xsdl
  namespace prefix with an underscore. XML parsing fails with the prefix.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/esp/services/esdl_svc_engine/esdl_binding.cpp
+++ b/esp/services/esdl_svc_engine/esdl_binding.cpp
@@ -889,6 +889,96 @@ void EsdlServiceImpl::handleFinalRequest(IEspContext &context,
                    "No target URL configured for %s!", mthdef.queryMethodName());
     }
 
+    auto echoInsertionPoint = strstr(out, "</Result></Results>");
+
+    if (echoInsertionPoint)
+    {
+        try
+        {
+            auto queryname = (isroxie ? tgtcfg->queryProp("@queryname") : nullptr);
+            auto requestname = mthdef.queryRequestType();
+            const char* qname = nullptr;
+            const char* content = nullptr;
+            StringBuffer encodedContent;
+            StringBuffer echoDataset;
+
+            XmlPullParser xpp(soapmsg.str(), soapmsg.length());
+            StartTag stag;
+            EndTag etag;
+            int type = XmlPullParser::END_DOCUMENT;
+
+            echoDataset.ensureCapacity(soapmsg.length());
+            xpp.setSupportNamespaces(false); // parser throws exceptions when it encounters undefined namespace prefixes, such as CRT's 'xsdl'
+            while ((type = xpp.next()) != XmlPullParser::END_DOCUMENT)
+            {
+                switch (type)
+                {
+                case XmlPullParser::START_TAG:
+                    xpp.readStartTag(stag);
+                    qname = stag.getQName();
+                    if (streq(qname, "soap:Envelope"))
+                        echoDataset << "<Dataset name=\"DesdlSoapRequestEcho\">";
+                    else if (streq(qname, "soap:Body"))
+                        echoDataset << "<Row>";
+                    else if (queryname && streq(qname, queryname))
+                        echoDataset << "<roxierequest name=\"" << queryname << "\">";
+                    else if (streq(qname, requestname))
+                        echoDataset << "<esprequest name=\"" << requestname << "\">";
+                    else
+                    {
+                        echoDataset << '<';
+                        if (strncmp(qname, "xsdl:", 5) == 0)
+                            echoDataset << "xsdl_" << qname + 5; // eliminate problematic namespace prefix
+                        else
+                            echoDataset << qname;
+                        for (int idx = 0; idx < stag.getLength(); idx++)
+                        {
+                            encodeXML(stag.getValue(idx), encodedContent.clear());
+                            echoDataset << ' ' << stag.getRawName(idx) << "=\"" << encodedContent << '"';
+                        }
+                        echoDataset << '>';
+                    }
+                    break;
+
+                case XmlPullParser::END_TAG:
+                    xpp.readEndTag(etag);
+                    qname = etag.getQName();
+                    if (streq(qname, "soap:Envelope"))
+                        echoDataset << "</Dataset>";
+                    else if (streq(qname, "soap:Body"))
+                        echoDataset << "</Row>";
+                    else if (queryname && streq(qname, queryname))
+                        echoDataset << "</roxierequest>";
+                    else if (streq(qname, requestname))
+                        echoDataset << "</esprequest>";
+                    else if (strncmp(qname, "xsdl:", 5) == 0)
+                        echoDataset << "</xsdl_" << qname + 5 << '>'; // eliminate problematic namespace prefix
+                    else
+                        echoDataset << "</" << qname << '>';
+                    break;
+
+                case XmlPullParser::CONTENT:
+                    content = xpp.readContent();
+                    if (!isEmptyString(content))
+                    {
+                        encodeXML(content, encodedContent.clear());
+                        echoDataset << encodedContent;
+                    }
+                    break;
+                }
+            }
+            out.insert(echoInsertionPoint - out.str(), echoDataset);
+        }
+        catch (XmlPullParserException& xppe)
+        {
+            ERRLOG("Unable to echo transformed request to response: %s", xppe.what());
+        }
+        catch (...)
+        {
+            ERRLOG("Unable to echo transformed request to response");
+        }
+    }
+
     processResponse(context,srvdef,mthdef,ns,out);
 }
 
--- a/esp/esdllib/esdl_transformer2.cpp
+++ b/esp/esdllib/esdl_transformer2.cpp
@@ -1855,6 +1855,8 @@ void Esdl2Transformer::processHPCCResult(IEspContext &ctx, IEsdlDefMethod &mthde
                 xppToXmlString(*xpp, stag, logdata);
             else if (strieq(dataset, "royaltyset"))
                 xppToXmlString(*xpp, stag, logdata);
+            else if (strieq(dataset, "desdlsoaprequestecho"))
+                xppToXmlString(*xpp, stag, logdata);
             else
             {
                 WARNLOG("ESDL processing HPCC Result: Dataset ignored: %s", dataset);


timothyklemm:
@afishbeck and @wangkx Please review.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -887,6 +887,52 @@ void EsdlServiceImpl::handleFinalRequest(IEspContext &context,
                    "No target URL configured for %s!", mthdef.queryMethodName());
     }
 
+    try
+    {
+        Owned<IPTree> responseTree = createPTreeFromXMLString(out);
```
timothyklemm: Will create this using ipt_ordered

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -887,6 +887,52 @@ void EsdlServiceImpl::handleFinalRequest(IEspContext &context,
...
+
+                xml.replaceString(replaceThis, withThat);
+            };
+
+            if (isroxie && tgtcfg->hasProp("@queryname"))
+                replaceTag(soapmsg, tgtcfg->queryProp("@queryname"), "roxiequery");
+            replaceTag(soapmsg, mthdef.queryRequestType(), "esprequest");
+            soapmsg.replaceString("xsdl:", "xsdl_");
+
+            Owned<IPTree> requestTree = createPTreeFromXMLString(soapmsg);
```
timothyklemm: Will create with ipt_ordered

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -887,6 +887,52 @@ void EsdlServiceImpl::handleFinalRequest(IEspContext &context,
...
+            auto bodyTree = requestTree->queryBranch("soap:Body");
+
+            if (bodyTree)
+            {
+                auto dsTree = responseResult->addPropTree("Dataset");
+                auto rowTree = dsTree->addPropTree("Row", LINK(bodyTree));
+
+                dsTree->addProp("@name", "DESDLRequestEcho");
+                toXML(responseTree, out.clear());
+                DBGLOG("%s", out.str());
```
timothyklemm: Will remove.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -889,6 +889,52 @@ void EsdlServiceImpl::handleFinalRequest(IEspContext &context,
...
+
+            Owned<IPTree> requestTree = createPTreeFromXMLString(soapmsg);
+            auto bodyTree = requestTree->queryBranch("soap:Body");
+
+            if (bodyTree)
+            {
+                auto dsTree = responseResult->addPropTree("Dataset");
+                auto rowTree = dsTree->addPropTree("Row", LINK(bodyTree));
+
+                dsTree->addProp("@name", "DESDLRequestEcho");
```
timothyklemm: HPCC-21061 will add another "request echo" dataset. Make this name more specific to which request is being echoed. Also change the transformer code looking for this name.

afishbeck:
@timothyklemm I wish there were an easier way to just hand the Transformed SOAP request to the logging agent rather then appending it to the result.  Is there now?  If not, perhaps something longer term that is more efficient that we can plan for in the future?

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -889,6 +889,51 @@ void EsdlServiceImpl::handleFinalRequest(IEspContext &context,
                    "No target URL configured for %s!", mthdef.queryMethodName());
     }
 
+    try
+    {
+        Owned<IPTree> responseTree = createPTreeFromXMLString(out, ipt_ordered);
```
afishbeck: Can you avoid loading the result into a ptree?  Seems they can get quite large.  

I think you could just find the right position in the string to 1. insert the new dataset into, or 2. overwrite adding the new dataset and then re-add the closing tags?

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -889,6 +889,51 @@ void EsdlServiceImpl::handleFinalRequest(IEspContext &context,
...
+        auto responseResult = responseTree->queryBranch("//Results/Result");
+
+        if (responseResult)
+        {
+            auto replaceTag = [](StringBuffer& xml, const char* tag, const char* var)
```
afishbeck: Is the benefit from using this syntax worth the fact that it might confuse some people?  That is, rather than just creating a normal function?

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -889,6 +889,51 @@ void EsdlServiceImpl::handleFinalRequest(IEspContext &context,
...
+
+                xml.replaceString(replaceThis, withThat);
+            };
+
+            if (isroxie && tgtcfg->hasProp("@queryname"))
+                replaceTag(soapmsg, tgtcfg->queryProp("@queryname"), "roxiequery");
+            replaceTag(soapmsg, mthdef.queryRequestType(), "esprequest");
+            soapmsg.replaceString("xsdl:", "xsdl_");
+
+            Owned<IPTree> requestTree = createPTreeFromXMLString(soapmsg, ipt_ordered);
```
afishbeck: You could probably do all of this more efficiently with a pull parser.  Even replace the tags that need replacing and copy accross the body tree into the out stringbuffer directly.

Unless we think the fully transformed requests will never be big enough for the efficiency savings to matter.

afishbeck (commented):
@timothyklemm some comments.

timothyklemm:
@afishbeck I've reworked to address your comments.

As to whether there is an easier way without adding to the response, which my changes still do, there might be a way. I can get the response to the point of preparing the IEspLogEntry instance, but there is no place for me to put this new value. If this interface can be changed now, I can give myself a place to put it. @wangkx was not in favor of making this change now.

afishbeck:
@timothyklemm as to the interface change, if Kevin doesn't want it now, that's fine.  Let's consider it in the future.  Of course we will be evaluating long term options for transaction logging anyway.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -889,49 +889,96 @@ void EsdlServiceImpl::handleFinalRequest(IEspContext &context,
...
+                {
+                case XmlPullParser::START_TAG:
+                    xpp.readStartTag(stag);
+                    qname = stag.getQName();
+                    if (strcmp(qname, "soap:Envelope") == 0)
+                        tmpOut << "<Dataset name=\"DesdlSoapRequestEcho\">";
+                    else if (strcmp(qname, "soap:Body") == 0)
+                        tmpOut << "<Row>";
+                    else if (queryname && strcmp(qname, queryname) == 0)
+                        tmpOut << "<roxierequest name=\"" << queryname << "\">";
```
afishbeck: There may be a common routine already to serialize all the children beneath the current tag.  If not we should add one in the future.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -889,49 +889,96 @@ void EsdlServiceImpl::handleFinalRequest(IEspContext &context,
...
+                    if (!isEmptyString(content))
+                    {
+                        encodeXML(content, encodedContent.clear());
+                        tmpOut << encodedContent;
+                    }
+                    break;
+                }
             }
+            tmpOut.append(echoInsertionPoint);
+            out.swapWith(tmpOut);
```
afishbeck: Do you think this could have been done without tmpOut at all?  I.e. append the info directly after echoInsertionPoint and then add the trailing close tags?

afishbeck (changes requested):
@timothyklemm I do think this will be more efficient especially for large outputs.  Some comments, especially wondering if you can get rid of the extra tmpOut buffer copy.

esp/services/esdl_svc_engine/esdl_binding.cpp
timothyklemm: I don't want an unexpected parser exception to leave `out` in an invalid state, because I'm not convinced an error parsing this should be allowed to break the response. If there are no other exceptions possible from the parsing, then yes, `out` could be manipulated in place. I would probably copy the trailing close tags into a second string rather than trying to recreate them, because I think that would be cleaner than finding what comes between `</Results>` and `</soap:Body>`.

timothyklemm: If I keep `tmpOut`, but only use it for the parsed dataset to be inserted, I could then manipulate `out` by inserting `tmpOut` at the insertion point only when `tmpOut` is assembled successfully.

esp/services/esdl_svc_engine/esdl_binding.cpp
timothyklemm: Unfortunately I'm modifying many children of the roxie request, so serializing all children at that level wouldn't work. If the custom request transform namespace issue gets resolved (there's an open issue about the namespace preventing use of the roxie test UI, and the logging code has the same issue), then we could benefit from optimization.

esp/services/esdl_svc_engine/esdl_binding.cpp
afishbeck: ok

esp/services/esdl_svc_engine/esdl_binding.cpp
afishbeck: ok.  Not sure where namespace you are talking about comes from, but it should either be removed, or properly defined at any layer that creates, uses, or passes it on.

timothyklemm: Every statement in the CRT includes `xsdl` as a namespace prefix. The binding requires the prefix to be present, without expecting a namespace declaration. Any pull parser supporting namespaces will throw an exception because of it.

I tried adding a namespace declaration to my transform, without success. I might have done it incorrectly, but I had to delete and recreate the binding after the attempt and didn't have time to keep fighting with it. That's when I opened https://track.hpccsystems.com/browse/HPCC-20883.

timothyklemm:
@afishbeck I've made more updates based on your feedback.

afishbeck (approved):
@timothyklemm looks fine.

afishbeck:
@timothyklemm please squash your commits.

afishbeck:
@richardkchapman this PR can be merged.

PR #11927, opened 2018-11-20
HPCC-20974 Optimize the code that resolves dataset selectors

ghalliday:
The previous code reported the issues when the no_select expression was traversed
but that required expressions to be walked multiple times in different scopes.
Moving the check to the outer expressions allows more expressions to be traversed
only once.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/ecl/hqlcpp/hqlttcpp.cpp
+++ b/ecl/hqlcpp/hqlttcpp.cpp
@@ -9922,7 +9922,7 @@ IHqlExpression * HqlLinkedChildRowTransformer::createTransformedBody(IHqlExpress
 
 HqlScopeTaggerInfo::HqlScopeTaggerInfo(IHqlExpression * _expr) : MergingTransformInfo(_expr)
 {
-    if (!onlyTransformOnce() && isIndependentOfScope(_expr))
+    if (!onlyTransformOnce() && (!containsImplicitNormalize(_expr) || _expr->isIndependentOfScope()))
     {
         //If the node doesn't have any active selectors then it isn't going to be context dependent
         setOnlyTransformOnce(true);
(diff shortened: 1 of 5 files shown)


ghalliday:
I'm not sure if this should go in 7.0.2.  It is only really worth it if subsequent changes in the same area produce significant benefits.
@shamser please review.

ecl/hqlcpp/hqlttcpp.cpp
```diff
@@ -10482,6 +10561,37 @@ void HqlScopeTagger::reportError(WarnErrorCategory category, const char * msg)
...
+
+    //Recursively search for an expression which refers to the first selector that is unresolved.
+    IHqlExpression * selector = &inScope.item(0);
+    IHqlExpression * location = nullptr;
+    IHqlExpression * original = findOriginalReference(&location, expr, transformed, selector);
+    if (original)
+    {
+        StringBuffer exprText;
+        getECL(original->queryBody(), exprText);
+
```
shamser: elideString isn't used here.  Are we sure that this regenerated Ecl not going to be too long?

ecl/hqlcpp/hqlttcpp.cpp
```diff
@@ -9970,6 +9983,52 @@ Datasets
...
+        if (transformOp != expr->getOperator())
+            return nullptr;
+
+        //Keep track of the best location that we have seen so far
+        IHqlExpression * symbol = queryLocation(expr);
+        if (symbol && location && (symbol->getStartLine() != 0))
+            *location = symbol;
+
+        //Find the first child expression that is also dependent on that selector
+        unsigned match = NotFound;
```
shamser: minor: NotFound is a constant from jarray header file, so should really be used for purposes other than related to jarray. (Also, NotFound is of type aindex_t.  It works but mixing types isn't ideal)

shamser (commented):
@ghalliday A couple of minor comments.

ecl/hqlcpp/hqlttcpp.cpp
ghalliday: I will change to use elideString, but it would have to be very obscure for this to be too long - the left hand side is almost certainly a named symbol, which will mean the string remains relatively short.  That isn't true for the other output.

ecl/hqlcpp/hqlttcpp.cpp
ghalliday: picky, but fair.  I'll look at changing.

ghalliday:
@shamser updated

shamser (approved):
@richardkchapman Code reviewed. Looks good.

richardkchapman:
@ghalliday Do you want to sqash before I merge?

PR #11929, opened 2018-11-21
HPCC-21039 Add a variant of CPersistentThreaded that uses tbb tasks

richardkchapman:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [ ] Security
  - [x] Thread-safety
  - [ ] Premature optimization

## Testing:
See unit tests

### DIFF
--- a/common/thorhelper/thorcommon.cpp
+++ b/common/thorhelper/thorcommon.cpp
@@ -2154,20 +2154,27 @@ void CPersistentTask::start()
     class RunTask : public tbb::task
     {
     public:
-        RunTask(IThreaded * _owner, tbb::task * _next) : owner(_owner), next(_next)
+        RunTask(CPersistentTask * _owner, tbb::task * _next) : owner(_owner), next(_next)
         {
         }
         virtual tbb::task * execute()
         {
-            owner->threadmain();
+            try
+            {
+                owner->owner->threadmain();
+            }
+            catch (IException *e)
+            {
+                owner->exception.setown(e);
+            }
             return next;
         }
     protected:
-        IThreaded * owner;
+        CPersistentTask * owner;
         tbb::task * next;
     };
     end = new (tbb::task::allocate_root()) tbb::empty_task();
-    tbb::task * task = new (end->allocate_child()) RunTask(owner, nullptr);
+    tbb::task * task = new (end->allocate_child()) RunTask(this, nullptr);
     end->set_ref_count(1+1);
     tbb::task::spawn(*task);
 }
@@ -2177,6 +2184,8 @@ bool CPersistentTask::join(unsigned timeout)
     end->wait_for_all();
     end->destroy(*end);
     end = nullptr;
+    if (exception.get())
+        throw exception.getClear();
     return true;
 }
 #endif
--- a/common/thorhelper/thorcommon.cpp
+++ b/common/thorhelper/thorcommon.cpp
@@ -2143,3 +2143,41 @@ ITranslator *getTranslators(const char *tracing, unsigned expectedCrc, IOutputMe
     else
         return nullptr;
 }
+
+#ifdef _USE_TBB
+#include "tbb/task.h"
+
+CPersistentTask::CPersistentTask(const char *name, IThreaded *_owner) : owner(_owner) {}
+
+void CPersistentTask::start()
+{
+    class RunTask : public tbb::task
+    {
+    public:
+        RunTask(IThreaded * _owner, tbb::task * _next) : owner(_owner), next(_next)
+        {
+        }
+        virtual tbb::task * execute()
+        {
+            owner->threadmain();
+            return next;
+        }
+    protected:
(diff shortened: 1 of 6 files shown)


richardkchapman:
@ghalliday Please review

It may be appropriate to split this into the test code (for 7.0.x) and the improvement (for 7.2)

system/jlib/jthread.cpp
```diff
@@ -702,22 +702,44 @@ void CAsyncFor::For(unsigned num,unsigned maxatonce,bool abortFollowingException
...
-            ready.signal();
-        IArrayOf<Thread> started;
-        started.ensure(num);
-        for (i=0;i<num;i++) {
-            ready.wait();
-            if (abortFollowingException && e) break;
-            Owned<Thread> thread = new cdothread(this,shuffled?shuffler->lookup(i):i,ready,&errmutex,e);
-            thread->start();
-            started.append(*thread.getClear());
+        if (maxatonce < num || abortFollowingException || shuffled)
```
ghalliday: abortFollowingException isn't really supported if maxatonce == num, so could use the new code for that case, and shuffled would also be a trivial change to support.

testing/unittests/unittests.cpp
```diff
@@ -674,4 +674,136 @@ class Base58Test : public CppUnit::TestFixture
...
+        public:
+            Thread(unsigned _count) : count(_count) {}
+            virtual void threadmain() override
+            {
+                ret = call_from_thread(count);
+            }
+            unsigned count;
+            unsigned ret = 0;
+        } t1(count), t2(count), t3(count);
+        CThreadedPersistent thread1("1", &t1), thread2("2", &t2), thread3("3", &t3);
```
ghalliday: Should be inside the switch, case 0, otherwise it distorts the timings for the rest

roxie/ccd/ccdserver.cpp
```diff
@@ -27239,10 +27239,43 @@ class CActivityGraph : implements IActivityGraph, implements IThorChildGraph, im
         CActivityGraph * container;
     } graphCodeContext;
 
+    class SinkThread : public CThreadedPersistent, implements IThreaded
```
ghalliday: CThreadedPersistent shoulod be a member of this class rather than a base class (its functions are never called from the outside).

Also, CThreadedPersistent is currently derived from CInterface, but there is no need for it to be - I suspect it is historical - that should be removed as a base class.

ghalliday: Although because you are adding them to an array SinkThread should be derived from CInterface.

ghalliday (commented):
You should split the PR into two.  I think the async changes may be suitable for a 7.0.2.  Roxie changes should probably wait until 7.2.0.

common/thorhelper/thorcommon.cpp
```diff
@@ -2143,3 +2143,41 @@ ITranslator *getTranslators(const char *tracing, unsigned expectedCrc, IOutputMe
...
+{
+    class RunTask : public tbb::task
+    {
+    public:
+        RunTask(IThreaded * _owner, tbb::task * _next) : owner(_owner), next(_next)
+        {
+        }
+        virtual tbb::task * execute()
+        {
+            owner->threadmain();
```
ghalliday: catch exceptions and set them in the owner so they can then be retrieved.

common/thorhelper/thorhelper.hpp
```diff
@@ -40,4 +41,30 @@ interface ICsvToRawTransformer : extends IInterface
...
 
+
+// Similar to CThreadedPersistent, but using tbb
+#ifdef _USE_TBB
+namespace tbb
+{
+class task;
+}
+
+class THORHELPER_API CPersistentTask : public CInterface
```
ghalliday: No need to be be derived from CInterface

ghalliday:
@richardkchapman some commments, and could do with splitting into 3 different PRs.

system/jlib/jthread.cpp
```diff
@@ -702,22 +702,44 @@ void CAsyncFor::For(unsigned num,unsigned maxatonce,bool abortFollowingException
...
+            IArrayOf<Thread> started;
+            started.ensure(num);
+            for (i=0;i<num-1;i++)
+            {
+                Owned<Thread> thread = new cdothread(this,i,ready,&errmutex,e);
+                thread->start();
+                started.append(*thread.getClear());
+            }
+            Do(num-1);
+            ready.wait();
```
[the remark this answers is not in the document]
ghalliday: I think you're right.  would be worth commenting.

system/jlib/jthread.cpp
ghalliday: I agree it is fairly meaningless, so could ignore it.

richardkchapman:
@ghalliday

richardkchapman:
Closing to split into 3

PR #11935, opened 2018-11-23
HPCC-21029 Improve AsyncFor code

richardkchapman:
Executes the last cycle on the calling thread, in the common simple case.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Premature optimization

## Testing:
Added unit tests, regression suite

### DIFF
--- a/system/jlib/jthread.cpp
+++ b/system/jlib/jthread.cpp
@@ -635,7 +635,6 @@ void CAsyncFor::For(unsigned num,unsigned maxatonce,bool abortFollowingException
         return;
     }
     Mutex errmutex;
-    Semaphore ready;
     IException *e=NULL;
     Owned<IShuffledIterator> shuffler;
     if (shuffled) {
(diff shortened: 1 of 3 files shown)


system/jlib/jthread.cpp
```diff
@@ -702,22 +702,47 @@ void CAsyncFor::For(unsigned num,unsigned maxatonce,bool abortFollowingException
...
+                started.item(idx).join();
+            }
         }
-        ForEachItemIn(idx, started)
+        else
         {
-            started.item(idx).join();
+            // Common case of execute all at once can be optimized a little
+            // Note that shuffle and abortFollowingException are meaningless when executing all at once
+            dbgassertex(!shuffled);
```
ghalliday: not sure this is a good idea.  Some general purpose code e.g.

    afor.For(eps.ordinality(), 10, false, true);

will trigger the assert if the list is too short

richardkchapman: True. Unlikely, but I will remove the asserts

system/jlib/jthread.cpp
```diff
@@ -702,22 +702,45 @@ void CAsyncFor::For(unsigned num,unsigned maxatonce,bool abortFollowingException
...
+            IArrayOf<Thread> started;
+            started.ensure(num);
+            for (i=0;i<num-1;i++)
+            {
+                Owned<Thread> thread = new cdothread(this,i,ready,&errmutex,e);
+                thread->start();
+                started.append(*thread.getClear());
+            }
+            Do(num-1);
+            ready.wait(); // Waiting on a single semaphore should help reduce context switches, even though we still need to join below
```
ghalliday: This isn't doing what you hope for - the reinit will have set it to a number > 0, so the wait will return immediately.

You cannot supply a -ve value to sem_init, so I'm not sure you can do what you are aiming for.

richardkchapman:
@ghalliday

PR #11941, opened 2018-11-26
HPCC-21049 Thor implementation of DISTRIBUTE,ALL

jakesmith:
Still needs some code generator work to finalize

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/thorlcr/activities/hashdistrib/thhashdistribslave.cpp
+++ b/thorlcr/activities/hashdistrib/thhashdistribslave.cpp
@@ -89,7 +89,7 @@ class CDistributorBase : implements IHashDistributor, implements IExceptionHandl
     Semaphore distribDoneSem, localFinishedSem;
     ICompare *iCompare, *keepBestCompare;
     size32_t bucketSendSize;
-    bool doDedup, allowSpill, connected, selfstopped;
+    bool doDedup, allowSpill, connected, selfstopped, isAll;
     Owned<IException> sendException, recvException;
     IStopInput *istop;
     size32_t fixedEstSize;
@@ -98,19 +98,20 @@ class CDistributorBase : implements IHashDistributor, implements IExceptionHandl
 
 protected:
     /*
-     * CSendBucket - a collection of rows destined for a particular destination target(slave)
+     * CSendBucket - a collection of rows destined for a particular target. A target can be a slave, or ALL
      */
+    class CTarget;
     class CSendBucket : implements IRowStream, public CSimpleInterface
     {
         CDistributorBase &owner;
         size32_t total;
         ThorRowQueue rows;
-        unsigned destination;
+        CTarget *target;
         CThorExpandingRowArray dedupList;
     public:
         IMPLEMENT_IINTERFACE_USING(CSimpleInterface);
 
-        CSendBucket(CDistributorBase &_owner, unsigned _destination) : owner(_owner), destination(_destination),
+        CSendBucket(CDistributorBase &_owner, CTarget *_target): owner(_owner), target(_target),
                 dedupList(*owner.activity, owner.rowIf)
         {
             total = 0;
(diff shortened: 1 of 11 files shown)


jakesmith:
@ghalliday - please review

jakesmith:
Sounds good, btw the setdistribute test currently fails because the key file is what the result should be, but the query is altered to make sure DISTRIBUTE,ALL runs.
Will change when codegen. changes are done.

jakesmith:
Now rebased, based on syntax/codegen changes.

One of the regression test results mismatched (result2 of distriball1.ecl), because the codegenerator removed on of the 2 DISTRIBUTE,ALL's. Have opened new issue : https://track.hpccsystems.com/browse/HPCC-21131. 
Consequently, I have left the //nothor in this test - it can be removed as part of HPCC-21131

@ghalliday - please review

thorlcr/activities/hashdistrib/thhashdistribslave.cpp
```diff
@@ -674,9 +712,9 @@ class CDistributorBase : implements IHashDistributor, implements IExceptionHandl
                  */
                 ForEachItemIn(t, targets)
                 {
-                    CTarget *target = targets.item(t);
-                    if (target->getSenderFinished())
+                    if (getSenderFinished(t))
```
ghalliday: Is this the correct test for the all case?  the parameter is a node number, but this will only be called with 0.  I suspect it works.

thorlcr/activities/hashdistrib/thhashdistribslave.cpp
```diff
@@ -816,11 +853,19 @@ class CDistributorBase : implements IHashDistributor, implements IExceptionHandl
...
                     if (!row)
                         break;
-                    unsigned dest = owner.ihash->hash(row)%owner.numnodes;
-                    CTarget *target = targets.item(dest);
-                    if (target->getSenderFinished())
-                        ReleaseThorRow(row);
+
+                    CTarget *target = nullptr;
+                    if (owner.isAll)
+                        target = targets.item(0);
```
ghalliday: Should this be checking getSenderFinished(0)?

ghalliday:
@jakesmith Generally looks good.
The code is still a bit confusing how the term "target" is used - sometimes it is a node, sometimes it is one of the distributors.  For instance t is used as a loop variable for both.  It also isn't completely clear what getSenderFinished() means for the ALL case.  I suspect it should still mean the nodes, in which case one of the comments above should be checked.

jakesmith:
>The code is still a bit confusing how the term "target" is used - sometimes it is a node, sometimes it is one of the distributors. For instance t is used as a loop variable for both.

"target" and "t" were always referring to a node and still do after this change, except in the special cased CTarget that can now handle being an 'all target'.
In that case it is a little confusing, I'll have a think how it could be changed to clarify.

thorlcr/activities/hashdistrib/thhashdistribslave.cpp
jakesmith: No, intentionally was not checking any finished status here.
In the single standard target case, it checks so it can throw the row away early if target has flagged that it's done.
Instead it checks to see if individual targets are finished during the broadcast (sendToOthers)

thorlcr/activities/hashdistrib/thhashdistribslave.cpp
jakesmith: No looks wrong.
It probably shouldn't be entering checkSendersFinished at all in the All case. Because it can't clear the buckets until all the node targets are done.
I will change.

jakesmith:
@ghalliday - please see fix+changes to clarify target vs destination.

ghalliday:
@jakesmith looks good, please can you squash

jakesmith:
@ghalliday - squashed

PR #11943, opened 2018-11-27
HPCC-21062 Add ECL syntax for DISTRIBUTE(ds, ALL)

ghalliday:
Also implement much of the underlying code to support

    DISTRIBUTE(ds, f(LEFT, TARGETNODE))

to allow distribution to a set of nodes

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/ecl/hqlcpp/hqlhtcpp.cpp
+++ b/ecl/hqlcpp/hqlhtcpp.cpp
@@ -6146,6 +6146,7 @@ double HqlCppTranslator::getComplexity(IHqlExpression * expr, ClusterType cluste
             complexity = 5;
         break;
     case no_distribute:
+    case no_nwaydistribute:
         if (isThorCluster(cluster))
             complexity = 5;
         break;
@@ -6514,6 +6515,9 @@ ABoundActivity * HqlCppTranslator::buildActivity(BuildCtx & ctx, IHqlExpression 
             case no_distribute:
             case no_assertdistributed:
                 result = doBuildActivityDistribute(ctx, expr);
+                break;
+            case no_nwaydistribute:
+                result = doBuildActivityNWayDistribute(ctx, expr);
                 break;
             case no_keyeddistribute:
                 result = doBuildActivityKeyedDistribute(ctx, expr);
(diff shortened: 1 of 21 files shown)


ghalliday:
I'm not sure Set Distribute is the best name.  I have used NWay Distribute within the code generator.  What do we want the activity to be called?

@jakesmith here is the code generator work - check it does what you need
@shamser please review.

jakesmith:
>I'm not sure Set Distribute is the best name. I have used NWay Distribute within the code generator. What do we want the activity to be called?

I think NWay Distribute is an improvement.
"distribute" is a bit misleading, it is more of a copy, clone or propagate.

shamser (approved):
@ghalliday Code Reiewed.  It would be useful to have the documentation Jira linked in this ticket. As far as  whether it's called DISTRIBUTE or NWAYDISTRIBUTE - not really fussed.

ghalliday:
@richardkchapman renamed to nway distribute and repushed.

ecl/hqlcpp/hqlhtcpp.cpp
```diff
@@ -6146,6 +6146,7 @@ double HqlCppTranslator::getComplexity(IHqlExpression * expr, ClusterType cluste
             complexity = 5;
         break;
     case no_distribute:
+    case no_nwaydistribute:
```
richardkchapman: Not sure what the complexity implies, but is it really appropriate to be the same as a normal distribute?

richardkchapman:
@ghalliday one trivial question

ecl/hqlcpp/hqlhtcpp.cpp
ghalliday: This code is used to calculate the "complexity" of a query.  I'm not convinced it is very useful, but treating it the same as a distribute is a reasonable assumption.

ghalliday:
@richardkchapman replied

PR #11979, opened 2018-12-06
HPCC-21059 Enhance custom request transform targeting

afishbeck:
Support target xpath at the xsdl:choose level.  This allows the
choose to reposition the target position in the original xml that
all child setValue and appendValue actions will affect.

Also push the transform level target xpath down to it's child
xsdl:choose level before merging the transforms.  This means the
uniqueness of the target xpaths will be correct after the merge
is done.  Otherwise the transform level xpath can be over ridden
during the merge.

If the choose already has it's own target it will be preserved.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/esp/services/esdl_svc_engine/esdl_binding.cpp
+++ b/esp/services/esdl_svc_engine/esdl_binding.cpp
@@ -350,13 +350,35 @@ void EsdlServiceImpl::addMethodLevelRequestTransform(const char *method, IProper
     }
 }
 
-static void ensureMergeOrderedPTree(Owned<IPropertyTree> &target, IPropertyTree *src)
+static void ensureMergeOrderedEsdlTransform(Owned<IPropertyTree> &dest, IPropertyTree *src)
 {
     if (!src)
         return;
-    if (!target)
-        target.setown(createPTree(ipt_ordered));
-    mergePTree(target, src);
+    if (!dest)
+        dest.setown(createPTree(ipt_ordered));
+    //copy so we can make changes, like adding calculated targets
+    Owned<IPropertyTree> copy = createPTreeFromIPT(src, ipt_ordered);
+    const char *target = copy->queryProp("@target");
+    if (target && *target)
+    {
+        Owned<IPropertyTreeIterator> children = copy->getElements("*");
+        ForEach(*children)
+        {
+            const char *existing = children->query().queryProp("@target");
+            if (existing && *existing)
+            {
+                if (*existing=='.') //support in future so error now so we don't have backward compatability issues
+                {
+                    VStringBuffer errorXpath("unsupported choose xpath*>>%s", existing); //will cause request time xpath error
+                    children->query().setProp("@target", errorXpath);
+                    DBGLOG("ESDL Binding: request transform choose xpath error: %s", existing);
+                }
+                continue;
+            }
+            children->query().setProp("@target", target);
+        }
+    }
+    mergePTree(dest, copy);
 }
 
 void EsdlServiceImpl::configureTargets(IPropertyTree *cfg, const char *service)
@@ -370,10 +392,10 @@ void EsdlServiceImpl::configureTargets(IPropertyTree *cfg, const char *service)
         Owned<IPropertyTree> serviceCrt;
         try
         {
-            ensureMergeOrderedPTree(serviceCrt, target_cfg->queryPropTree("xsdl:CustomRequestTransform"));
+            ensureMergeOrderedEsdlTransform(serviceCrt, target_cfg->queryPropTree("xsdl:CustomRequestTransform"));
             Owned<IPropertyTreeIterator> transforms =  target_cfg->getElements("Transforms/xsdl:CustomRequestTransform");
             ForEach(*transforms)
-                ensureMergeOrderedPTree(serviceCrt, &transforms->query());
+                ensureMergeOrderedEsdlTransform(serviceCrt, &transforms->query());
         }
         catch (IPTreeException *e)
         {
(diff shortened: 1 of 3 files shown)


afishbeck:
@timothyklemm please review.

esp/services/esdl_svc_engine/esdl_svc_custom.cpp
```diff
@@ -208,14 +210,35 @@ bool CEsdlCustomTransformChoose::evaluate(IXpathContext * xpathContext)
...
+        xpath.set(defaultTarget);
+
+    IPropertyTree *reqTarget = request;
+
+    if (xpath.length())
+    {
+        //we can use real xpath processing in the future, for now simple substitution is fine
+        xpath.replaceString("{$query}", xpathContext->getVariable("query"));
+        xpath.replaceString("{$method}", xpathContext->getVariable("method"));
+        xpath.replaceString("{$service}", xpathContext->getVariable("service"));
```
timothyklemm: Can we add a variable for the ESP request type (handleFinalRequest defines `request` in addition to these three values)? The default, backward compatible, target includes this value. If a transform element changes it's target from the default, there is no generic way to get back to the default from a descendant.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -350,13 +350,35 @@ void EsdlServiceImpl::addMethodLevelRequestTransform(const char *method, IProper
...
+            {
+                if (*existing=='.') //support in future so error now so we don't have backward compatability issues
+                {
+                    VStringBuffer errorXpath("unsupported choose xpath*>>%s", existing); //will cause request time xpath error
+                    children->query().setProp("@target", errorXpath);
+                    DBGLOG("ESDL Binding: request transform choose xpath error: %s", existing);
+                }
+                continue;
+            }
+            children->query().setProp("@target", target);
```
timothyklemm: Is this the best name for the attribute?

While not needed now, I can see a scenario where we would extend this mechanism to the transform rules themselves. I have use cases where the transform will set an in-band and an out-of-band value based on the same test result. By only setting targets on the `choose` elements it is necessary to replicate, and repeat the execution of, the test logic for each value set.

Maybe `root` or `base` would be something that makes sense now and could eventually be added to the rules that already include `target` attributes.

timothyklemm (changes requested):

esp/services/esdl_svc_engine/esdl_binding.cpp
afishbeck: I thought about something like “baseTarget” but my current idea is to later support both relative and absolute targets.  A relative target is relative to the parent target.  An absolute target is self contained.  That’s why i error when the path starts with “.” Because in the future that would mean relative.

But we could do the base target thing if we don’t think support for relative paths is enough.

timothyklemm: The rule's `target` is always relative to whatever root node is used in the evaluation. I'm not questioning how this root is selected. I'm just questioning the name of the attribute used to select it, and only because I can see that attribute being added to the rule some day, and two `target` attributes won't work.

I like the idea of supporting relative path references for the attribute, whatever it may be named.

afishbeck: I'm changing this to an attribute intended only for internal use.  CRT/@target is added to the xsdl:choose as xsdl:choose/@_crtTarget.  But not meant to be overridden by the user at all.  We can add support for use defined relative targets at different levels later.  Main reason is I want to get the relative path syntax exactly right.

esp/services/esdl_svc_engine/esdl_svc_custom.cpp
afishbeck: Done

afishbeck:
@timothyklemm please review.  I will test.  I will squash after review.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -363,20 +363,7 @@ static void ensureMergeOrderedEsdlTransform(Owned<IPropertyTree> &dest, IPropert
...
-                {
-                    VStringBuffer errorXpath("unsupported choose xpath*>>%s", existing); //will cause request time xpath error
-                    children->query().setProp("@target", errorXpath);
-                    DBGLOG("ESDL Binding: request transform choose xpath error: %s", existing);
-                }
-                continue;
-            }
-            children->query().setProp("@target", target);
-        }
+            children->query().setProp("@_crtTarget", target); //internal use only attribute
```
timothyklemm: I think the check for existing is still necessary. A top level choose that specifies a custom target will be overwritten, but descendant instances will be honored (whether or not the attribute should be there).

My only issue with the using `@target` was the future possibility adding this value to an element that uses that name for something else.

afishbeck: This happens at load time, the usage and passing the target tree to descendants happens at run time.  This just puts the target in the right place.  The reason I checked for existing before was because I was letting the user set a target at "choose".  I'm not supporting that now.  That's a future enhancement.

afishbeck: So this only pushes an internal attribute down.  There will be no existing because the user isn't supposed to add a _crtTarget attribute on there own.  If they do it will be overwritten as expected.

timothyklemm (approved):

afishbeck:

PR #11989, opened 2018-12-07
HPCC-20951 Add Std.Date.FromStringToSeconds() function

dcamper:
Parse strings containing both date and time values (e.g. '1999-02-01T12:34:56') and convert the result into a Std.Date.Seconds_t value (number of seconds since epoch).

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Any new interfaces or exported functions are appropriately commented.
  - [x] Scalability
  - [x] Performance
  - [ ] Security
  - [x] Thread-safety
  - [x] Premature optimization

## Testing:

Manually and with new unit tests.


richardkchapman:
@ghalliday Please review

plugins/common/simple_strptime.hpp
```diff
@@ -0,0 +1,11 @@
+#ifndef SIMPLE_STRPTIME_INCL
+#define SIMPLE_STRPTIME_INCL
+
+#include <time.h>
+#include "platform.h"
+
+extern "C" {
```
ghalliday: If this does remain compiled into both shared objects (see other comment), there isn't any need for this to be extern "C" and it should also have DECL_LOCAL on the definition to prevent it being exported.

plugins/timelib/timelib.cpp
```diff
@@ -657,6 +659,23 @@ TIMELIB_API void TIMELIB_CALL tlSecondsToString(size32_t &__lenResult, char* &__
 
 //------------------------------------------------------------------------------
 
+TIMELIB_API __int64 TIMELIB_CALL tlStringToSeconds(size32_t lenS, const char* s, const char* fmtin, bool is_local_time)
```
ghalliday: I think this would probably be better to leave the strptime code where it was and add this function into stringlib.cpp, otherwise the code will be compiled into two shared objects.  The function can still be available from Std.Date.

ghalliday (commented):
Generally looks good.  Minor questions to do with the location of the code.

plugins/timelib/timelib.cpp
dcamper: The problem I was dealing with was that tlStringToSeconds relies on non-trivial code from both stringlib and timelib.  strptime and tlMkTime (which contains Linux and Windows versions), respectively.  I originally wanted to migrate strptime and all functions relying it to timelib, but Richard talked me out of it.

plugins/common/simple_strptime.hpp
richardkchapman: Might also be a good idea to use a namespace to ensure that the two copies are regarded as different by tools like debuggers etc.

plugins/timelib/timelib.cpp
richardkchapman: Looks to me as though the code you want to share (simple_strptime) is not currently exported from stringlib, meaning that no-one can be calling it from existing ECL code. So it would be ok to move that code into a shared location (e.g. eclrtl) and to have both these plugins call it there.

dcamper: I will move strptime and its related functions to eclrt.  That will remove the common code currently in place.

dcamper:
@ghalliday simple_strptime moved to eclrtl.  Please review again.  Thanks!

rtl/eclrtl/eclrtl.cpp
```diff
@@ -4213,6 +4213,192 @@ unsigned rtlDelayReturn(unsigned value, unsigned sleepTime)
...
+            strOffset = offset;
+            return true;
+        }
+    }
+    return false;
+}
+
+//This implements a subset of the specifiers allowed for strptime
+//Another difference is it works on a string with a separate length
+ECLRTL_API const char * rltSimpleStrPTime(size32_t lenStr, const char * str, const char * format, struct tm * tm)
```
ghalliday: minor: rlt should be rtl (**r**un**t**ime**l**ibrary).  You could have kept the original name since this function cannot be called from generated code.

ghalliday (commented):
@dcamper looks fine.  Please rename the function though - it was confusing that I couldn't find it because I was searching for the wrong prefix.

dcamper:
Dyslexics untie!

Renamed rltSimpleStrPTime to simple_strptime to help indicate that the function doesn't support generated code.

@ghalliday Please review again.  Thanks!

ghalliday:
Looks good to me.  I have also checked it builds on windows.

PR #11997, opened 2018-12-13
HPCC-21192 Changed confusing "Cluster" in Keys to "ClusterGroup"

jakesmith:
The Keys section mapped a group to a key name pair, but the
entry settings were under a "Cluster" tag which was confusing.
Keep backward compatibility, so with this change picks up new
"ClusterGroup" and legacy "Cluster" settings.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/common/environment/environment.cpp
+++ b/common/environment/environment.cpp
@@ -181,27 +181,51 @@ private:
 
                 keyPairMap[name] = { absPublicKeyPath.str(), absPrivateKeyPath.str() };
             }
-            Owned<IPropertyTreeIterator> clusterIter = p->getElements("EnvSettings/Keys/Cluster");
-            ForEach(*clusterIter)
+
+/* From 7.0.0 until 7.0.6, the <Keys> section of the environment required
+ * the mappings to be defined as "Cluster" instead of "ClusterGroup" - See: HPCC-21192
+ */
+#define BKWRDCOMPAT_CLUSTER_VS_CLUSTERGROUP
+
+            const char *groupKeysPath = "EnvSettings/Keys/ClusterGroup";
+
+#ifdef BKWRDCOMPAT_CLUSTER_VS_CLUSTERGROUP
+            for (unsigned i=0; i<2; i++) // once for std. "ClusterGroup", 2nd time for legacy "Cluster"
             {
-                IPropertyTree &cluster = clusterIter->query();
-                const char *clusterName = cluster.queryProp("@name");
-                if (isEmptyString(clusterName))
+#endif
+                Owned<IPropertyTreeIterator> clusterIter = p->getElements(groupKeysPath);
+
+#ifdef BKWRDCOMPAT_CLUSTER_VS_CLUSTERGROUP
+                if (clusterIter->first() && keyClusterMap.size()) // NB: always 0 1st time around.
                 {
-                    WARNLOG("skipping EnvSettings/Keys/Cluster entry with no name");
-                    continue;
+                    WARNLOG("Invalid configuration: mixed 'Keys/ClusterGroup' definitions and legacy 'Keys/Cluster' definitions found, legacy 'Keys/Cluster' definition will be ignored.");
+                    break;
                 }
-                if (cluster.hasProp("@keyPairName"))
+#endif
+                ForEach(*clusterIter)
                 {
-                    const char *keyPairName = cluster.queryProp("@keyPairName");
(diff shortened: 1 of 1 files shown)


common/environment/environment.cpp
```diff
@@ -181,27 +181,43 @@ class CLocalEnvironment : implements IConstEnvironment, public CInterface
...
+                        if (isEmptyString(keyPairName))
+                        {
+                            WARNLOG("skipping invalid %s entry, name=%s", groupKeysPath, clusterName);
+                            continue;
+                        }
+                        keyClusterMap[clusterName] = keyPairName;
+                    }
                 }
+#ifdef BKWRDCOMPAT_CLUSTER_VS_CLUSTERGROUP
+                groupKeysPath = "EnvSettings/Keys/Cluster";
```
mckellyln: Should we always do this ?  Or at least until we build with BKWRDCOMPAT_CLUSTER_VS_CLUSTERGROUP undefined ?
I mean if we find keys in ClusterGroup - should we still continue to on and overwrite them with what is found in Cluster ?

mckellyln:
@jakesmith one question.

common/environment/environment.cpp
jakesmith: This backward compat. is to allow either, but in particular to allow for old environments that just have old "Cluster" definitions in, to continue to work.

Anyone who has mixed definitions, neither has a purely legacy definition, nor a new corrected definition, so it should probably just error at that point, either way they are asking for trouble.

There's also nothing to stop them adding multiple definitions that override one another, e.g. later "ClusterGroup"'s will override earlier "ClusterGroup"'s. 

I think it's fine as is and hopefully we can delete it in near future.

jakesmith:
@mckellyln - replied to comment. I don't think worth performing any extra checks here.

common/environment/environment.cpp
mckellyln: I would reverse it then - read the old first and then overwrite with what is new.  Seems safer.

jakesmith: I agree if mixed were allowed, newer should override older definitions
But whilst it's legitimate for there to just be old or just new, there should never be mixed entries..

I think the only change worth making, it at all, is add some to detect both and fire an error.

jakesmith: I'll add a check for mixed and explicitly throw a warning and ignore legacy definitions if somehow someone has an environment with both new and legacy mixed together.

jakesmith:
@mckellyln - added a check for invalid mixed definitions, that will issue a warning and ignore legacy definitions if new exists.

common/environment/environment.cpp
```diff
@@ -181,27 +181,48 @@ class CLocalEnvironment : implements IConstEnvironment, public CInterface
...
             {
-                IPropertyTree &cluster = clusterIter->query();
-                const char *clusterName = cluster.queryProp("@name");
-                if (isEmptyString(clusterName))
+                if (keyClusterMap.size()) // NB: always 0 1st time around.
                 {
-                    WARNLOG("skipping EnvSettings/Keys/Cluster entry with no name");
-                    continue;
+                    WARNLOG("Invalid configuration: mixed 'Keys/ClusterGroup' definitions and legacy 'Keys/Cluster' definitions found, legacy 'Keys/Cluster' definition will be ignored.");
+                    break;
```
mckellyln: It is not a big deal - but this will report old keys found even if there are not any there.
This logic could be moved into the ForEach() right before updating the KeyClusterMap[].

jakesmith: yeah you're right, not good, I'll change.

jakesmith: @mckellyln - have changed so only checks if any in iterator. Some might have not have get as far as actually qualifying for keyClusterMap[] addition, but they shouldn't be there at all.
This involves an extra iter->first() call, that could be avoided if really cared, but we don't here point.

mckellyln: looks good

mckellyln (approved):
Approved for merge

mckellyln:
@jakesmith, @richardkchapman - good to merge

jakesmith:
@richardkchapman - not sure why it's on my gitbot PR list - thought it would only attribute to last mention in message.
Anyway, ready to merge.

PR #12006, opened 2018-12-14
HPCC-21215 Detect CSV/XML/FLAT/INDEX mismatches.

jakesmith:
Only throw exception in case of variable diskread reading
mismatched type, or flat vs index (as before).
Otherwise issue workunit warning only.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/dali/base/dadfs.hpp
+++ b/dali/base/dadfs.hpp
@@ -814,6 +814,10 @@ inline bool isPartTLK(IPropertyTree &pt) { const char *kind = pt.queryProp("@kin
 inline bool isPartTLK(IDistributedFilePart *p) { return isPartTLK(p->queryAttributes()); }
 inline bool isPartTLK(IPartDescriptor *p) { return isPartTLK(p->queryProperties()); }
 
+inline const char *queryFileKind(IPropertyTree &pt) { return pt.queryProp("@kind"); }
+inline const char *queryFileKind(IDistributedFile *f) { return queryFileKind(f->queryAttributes()); }
+inline const char *queryFileKind(IFileDescriptor *f) { return queryFileKind(f->queryProperties()); }
+
 extern da_decl void ensureFileScope(const CDfsLogicalFileName &dlfn, unsigned timeoutms=INFINITE);
 
 extern da_decl bool checkLogicalName(const char *lfn,IUserDescriptor *user,bool readreq,bool createreq,bool allowquery,const char *specialnotallowedmsg);
--- a/thorlcr/activities/csvread/thcsvread.cpp
+++ b/thorlcr/activities/csvread/thcsvread.cpp
@@ -35,7 +35,11 @@ public:
         if (headerLines)
             mpTag = container.queryJob().allocateMPTag();
     }
-    void serializeSlaveData(MemoryBuffer &dst, unsigned slave)
+    virtual void validateFile(IDistributedFile *file) override
+    {
+        // NB: CSV can be used to read any format
+    }
+    virtual void serializeSlaveData(MemoryBuffer &dst, unsigned slave) override
     {
         CDiskReadMasterBase::serializeSlaveData(dst, slave);
         if (mapping && mapping->queryMapWidth(slave)>=1)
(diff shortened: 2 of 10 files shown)


jakesmith:
targeting 7.0.6 for now, but perhaps should move to 7.2

@ghalliday - please review

richardkchapman:
Nervous about the potential for false-positives here. Definitely 7.2 rather than 7.06, I would think, but are we sure it's a good idea at all? At least make sure we can disable if necessary.

thorlcr/activities/csvread/thcsvread.cpp
```diff
@@ -35,7 +35,21 @@ class CCsvReadActivityMaster : public CDiskReadMasterBase
         if (headerLines)
             mpTag = container.queryJob().allocateMPTag();
     }
-    void serializeSlaveData(MemoryBuffer &dst, unsigned slave)
+    virtual void validateFile(IDistributedFile *file) override
+    {
+        if (!isFileCsv(file))
```
ghalliday: I don' think there is any harm in allowing csv to read any file format.  The main problem with mismatched files is when it causes thor to crash, and csv should never crash, regardless of the input.

thorlcr/activities/diskread/thdiskread.cpp
```diff
@@ -31,9 +31,11 @@ class CDiskReadMasterVF : public CDiskReadMasterBase
 {
 public:
     CDiskReadMasterVF(CMasterGraphElement *info) : CDiskReadMasterBase(info) { }
-    virtual void validateFile(IDistributedFile *file)
+    virtual void validateFile(IDistributedFile *file) override
     {
         IHThorDiskReadBaseArg *helper = (IHThorDiskReadBaseArg *)queryHelper();
+        if (!isFileFlat(file))
```
ghalliday: I thought this test was only going to be for variable size files - since they are the only files that can crash when reading mismatching files.  It is reasonable to read a fixed with csv file as binary for example.

ghalliday (commented):
@jakesmith a couple of comments.  I also agree with Richard about the target version.

thorlcr/activities/diskread/thdiskread.cpp
jakesmith: I will change to only check in variable case.

thorlcr/activities/csvread/thcsvread.cpp
jakesmith: right, I'll allow CSV to read any type, will change to comment.

jakesmith:
>Nervous about the potential for false-positives here. Definitely 7.2 rather than 7.06, I would think, but are we sure it's a good idea at all? At least make sure we can disable if necessary.

The regression suite mismatches were mostly due to a missing JSON type check (in common XML master read class).
But there was also case of a CSV file being read as XML.

Have changed to target 7.2
Added an option to disable the check at activity/workunit/global level if required.
Only made it a hard error in case of variable flat with mismatched file type, or flat vs index (same prior to this PR). Otherwise issue workunit warning only.

jakesmith:
@ghalliday - please re-review

ghalliday:
@jakesmith I think that now looks good.  Please squash and reassign to richard to sanity check/merge

dali/base/dadfs.hpp
```diff
@@ -814,6 +814,10 @@ inline bool isPartTLK(IPropertyTree &pt) { const char *kind = pt.queryProp("@kin
 inline bool isPartTLK(IDistributedFilePart *p) { return isPartTLK(p->queryAttributes()); }
 inline bool isPartTLK(IPartDescriptor *p) { return isPartTLK(p->queryProperties()); }
 
+inline const char *getFileKind(IPropertyTree &pt) { return pt.queryProp("@kind"); }
```
richardkchapman: picky, but queryFileKind might be more appropriate since the caller does not take ownership of the returned string.

richardkchapman:
@jakesmith One picky comment, I will merge as-is if you don't feel like addressing it

dali/base/dadfs.hpp
jakesmith: fair enough, I will change it.

jakesmith:
@richardkchapman - changed getFileKind to queryFileKind (and squashed change in).

PR #12007, opened 2018-12-18
HPCC-21220 Allow explicit selection of config content sent to roxie

afishbeck:
With the introduction of request transforms the dynamic configuration of
methods gets quite large.  What gets added to the request context config
section should be much more explicit so that we are not wasting bandwidth.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/esp/services/esdl_svc_engine/esdl_svc_engine.cpp
+++ b/esp/services/esdl_svc_engine/esdl_svc_engine.cpp
@@ -39,6 +39,70 @@ CEsdlSvcEngineSoapBindingEx::CEsdlSvcEngineSoapBindingEx(IPropertyTree* cfg, con
 {
 }
 
+IPropertyTree *createContextMethodConfig(IPropertyTree *methodConfig)
+{
+    const char *include = "*";
+    if (methodConfig->hasProp("@contextInclude"))
+        include = methodConfig->queryProp("@contextInclude");
+    const char *remove = "Transforms|xsdl:CustomRequestTransform";
+    if (methodConfig->hasProp("@contextRemove"))
+        remove = methodConfig->queryProp("@contextRemove");
+    const char *removeAttrs = "@contextInclude|@contextRemove|@contextAttRemove";
+    if (methodConfig->hasProp("@contextAttRemove"))
+        removeAttrs = methodConfig->queryProp("@contextAttRemove");
+
+    Owned<IPropertyTree> contextConfig;
+    if (include && !streq(include, "*"))
+    {
+        contextConfig.setown(createPTree("Method", ipt_ordered));
+        Owned<IAttributeIterator> aiter = methodConfig->getAttributes(); //for now include all attributes
+        ForEach (*aiter)
+            contextConfig->addProp(aiter->queryName(), aiter->queryValue());
+
+        if (*include) //if *include==0 we should have an empty tree
+        {
+            StringArray xpaths;
+            xpaths.appendListUniq(include, "|", true); //don't supported quoted | for now
+            ForEachItemIn(pos, xpaths)
+            {
+                Owned<IPropertyTreeIterator> toInclude = methodConfig->getElements(xpaths.item(pos));
+                ForEach(*toInclude)
+                    contextConfig->addPropTree(toInclude->query().queryName(), LINK(&toInclude->query()));
+            }
+        }
+    }
+    if (remove && *remove)
+    {
+        if (!contextConfig)
+            contextConfig.setown(createPTreeFromIPT(methodConfig, ipt_ordered));
+        if (contextConfig->hasChildren())
+        {
+            StringArray xpaths;
+            xpaths.appendListUniq(remove, "|", true); //don't supported quoted | for now
+            ForEachItemIn(pos, xpaths)
+            {
+                Owned<IPropertyTreeIterator> toRemove = contextConfig->getElements(xpaths.item(pos));
+                ForEach(*toRemove)
+                    contextConfig->removeTree(&toRemove->query());
+            }
+        }
+    }
+
+    if (removeAttrs && *removeAttrs)
+    {
+        if (!contextConfig)
+            contextConfig.setown(createPTreeFromIPT(methodConfig, ipt_ordered));
+        StringArray xpaths;
+        xpaths.appendListUniq(removeAttrs, "|", true); //don't supported quoted | for now
+        ForEachItemIn(pos, xpaths)
+            contextConfig->removeProp(xpaths.item(pos));
+    }
+
+    if (contextConfig)
+        return contextConfig.getClear();
+    return LINK(methodConfig); //no copy, nothing changed
+}
+
 IPropertyTree *CEsdlSvcEngine::createTargetContext(IEspContext &context, IPropertyTree *tgtcfg, IEsdlDefService &srvdef, IEsdlDefMethod &mthdef, IPropertyTree *req_pt)
 {
     Owned<IPropertyTree> localCtx(createPTreeFromIPT(m_service_ctx, ipt_none));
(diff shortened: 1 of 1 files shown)


afishbeck:
@timothyklemm please review.

afishbeck:
@timothyklemm please review

afishbeck:
@timothyklemm the information that gets added to the context from the config was formatted differently then I thought, being placed in the Context/ESP/Config tag.  Therefore I decided to take a different approach.
By default everything is grabbed for the context, but then the transforms are filtered out.
The default behavior already helps our current issue because the transforms will be removed.

But besides the default behavior you can customize either the selection of what gets added, or what gets filtered out, or both.
Here's an example:
 <Methods>
       <xsdl:CustomRequestTransform>
          <xsdl:choose>
             <xsdl:when test="$clientversion=1.9">
                <xsdl:SetValue target="Row/Name/First"  value="'service'" />
             </xsdl:when>
             <xsdl:otherwise>
                <xsdl:SetValue target="Row/Name/Last" value="'zz'"/>
             </xsdl:otherwise>
          </xsdl:choose>
       </xsdl:CustomRequestTransform>
    <Method name="JavaEchoPersonInfo" querytype="java" javamethod="EsdlExample.EsdlExampleService.JavaEchoPersonInfo"/>
    <Method name="RoxieEchoPersonInfo" querytype="roxie" url="http://localhost:9876/roxie" queryname="RoxieEchoPersonInfo" contextInclude="xsdl:CustomRequestTransform|MyCustomeValue1|MyCustomeValue1" contextRemove="xsdl:CustomRequestTransform|MyCustomeValue1">
         <xsdl:CustomRequestTransform target="soap:Body/{$query}">
            <xsdl:choose>
               <xsdl:when test="$clientversion=1.9">
                  <xsdl:SetValue target="vertest"  value="'v1.9'"/>
                  <xsdl:SetValue target="RoxieEchoPersonInfoRequest/Row/Name/First"  value="'v1.9'"/>
               </xsdl:when>
               <xsdl:otherwise>
                  <xsdl:SetValue target="vertest"  value="concat('v', $clientversion)"/>
                  <xsdl:SetValue target="RoxieEchoPersonInfoRequest/Row/Name/Last" value="concat('v', $clientversion)"/>
               </xsdl:otherwise>
            </xsdl:choose>
         </xsdl:CustomRequestTransform>
         <Gateways>xxx</Gateways>
         <MyCustomValue1>xxx</MyCustomValue1>
         <MyCustomValue2>xxx</MyCustomValue2>
   </Method>
</Methods>

esp/services/esdl_svc_engine/esdl_svc_engine.cpp
```diff
@@ -48,8 +112,12 @@ IPropertyTree *CEsdlSvcEngine::createTargetContext(IEspContext &context, IProper
     localCtx->setProp("Row/Common/ESP/ServiceName", context.queryServiceName(""));
     //removing this entry since the Row/Common/ESP/Config/Method tree should have an attribute @name
     //localCtx->setProp("Row/Common/ESP/MethodName", mthdef.queryMethodName());
-    ensurePTree(localCtx, "Row/Common/ESP/Config");
-    localCtx->addPropTree("Row/Common/ESP/Config/Method", LINK(tgtcfg));
+    Owned<IPropertyTree> config = createContextMethodConfig(tgtcfg);
```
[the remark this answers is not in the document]
timothyklemm: This comment was premature. `handleFinalRequest` receives the original and config and the output of this method. The original tree is used for gateway transformations, and this method's output is only used for SOAP request generation. It should be OK.

esp/services/esdl_svc_engine/esdl_svc_engine.cpp
```diff
@@ -39,6 +39,70 @@ CEsdlSvcEngineSoapBindingEx::CEsdlSvcEngineSoapBindingEx(IPropertyTree* cfg, con
...
+            xpaths.appendListUniq(include, "|", true); //don't supported quoted | for now
+            ForEachItemIn(pos, xpaths)
+            {
+                Owned<IPropertyTreeIterator> toInclude = methodConfig->getElements(xpaths.item(pos));
+                ForEach(*toInclude)
+                    contextConfig->addPropTree(toInclude->query().queryName(), LINK(&toInclude->query()));
+            }
+        }
+    }
+    if (remove && *remove)
```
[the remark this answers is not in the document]
timothyklemm: I would expect that removing the method element would also remove the method element's attributes. Attributes have no meaning outside the context of their element. It would be incredibly counter intuitive to omit the entire branch but have it included because the attributes were not explicitly marked for removal.

That doesn't mean a contextAttRemove="*" doesn't have value on its own.

afishbeck:
@timothyklemm updated, please review.

esp/services/esdl_svc_engine/esdl_svc_engine.cpp
```diff
@@ -103,6 +103,22 @@ IPropertyTree *createContextMethodConfig(IPropertyTree *methodConfig)
...
+
+    const char *elInclude = cfg->queryProp("@contextInclude");
+    const char *elRemove = cfg->queryProp("@contextRemove");
+    const char *attRemove = cfg->queryProp("@contextAttRemove");
+
+    bool noElIncluded = elInclude && !*elInclude; //empty xpath
+    bool allElRemoved = (elRemove && streq(elRemove, "*"));
+    bool allAttRemoved = (attRemove && streq(attRemove, "*"));
+
+    return (allAttRemoved && (noElIncluded || allElRemoved));
```
timothyklemm: I can't see a logical use case for an explicit request to omit all elements to be completely ignored because I did not also include a separate explicit request to omit all parts of all elements. That's essentially what this is saying: contextRemove can omit everything but only if contextAttRemove explicitly confirms the intent to omit everything, and absent this confirmation nothing is omitted.

Maybe if attRemove is not empty and not "*", but even that seems like a stretch to override elRemove.

In what scenario would someone specify contextRemove="*" and not expect everything to be removed?

afishbeck: @timothyklemm this isn't about what are the likely scenarios, it's about making sure that there is a proper syntax for anything the user wants to do AND is backward compatible by default.

contextAttRemove doesn't affect attributes at every level, it only affects top level attributes, and contextRemove only affects child elements.

I am treating them as ptree xpaths with additional support for "|".

It's done this way to make sure I can support "remove all child elements" but am able keep top level attributes.  
Defaulting contextAttRemove to "*" would not be backward compatible, and it would be hard to then say "keep all".

When users want to build highly custom contexts they can use a request transform.

afishbeck: If contextAttRemove ="*" were the default, keep all could be contextAttRemove="".  But that still doesn't help with backward compatibility.

afishbeck: I suppose if we don't care about logical consistency I could say that contextRemove="*" changes the default for contextAttRemove to "*" also.  Then to keep all the user would have to set contextAttRemove="".  It's weird but would be convenient.

esp/services/esdl_svc_engine/esdl_svc_engine.cpp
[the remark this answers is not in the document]
afishbeck: As we discussed offline, contextRemove="*" affects the children of our root node which in this case is "Method".  So the Method tag and its attributes still exist.  I remove method only if it is completely empty.

But as we also discussed offline I will use contextRemove="." to mean remove the entire current (root) node.  Which will include the attributes.

esp/services/esdl_svc_engine/esdl_svc_engine.cpp
afishbeck: As we discussed offline I will use contextRemove="." to mean remove the entire current (root) node. Which will include the attributes.  I will remove contextConfig="true/false" since that was meant as a similar simplification.

afishbeck:
@timothyklemm updated please review.  I will squash when ready.

timothyklemm (approved):
Reviewed and tested multiple common scenarios.

PR #12009, opened 2018-12-18
HPCC-19177 Implement createSimplifiedDefinition for datasets

shamser:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/ecl/hql/hqlcache.cpp
+++ b/ecl/hql/hqlcache.cpp
@@ -20,6 +20,8 @@
 #include "hqlcache.hpp"
 #include "hqlcollect.hpp"
 #include "hqlexpr.hpp"
+#include "hqltrans.ipp"
+#include "hqlattr.hpp"
 #include "hqlutil.hpp"
 #include "hqlerrors.hpp"
 #include "junicode.hpp"
@@ -340,6 +342,134 @@ void convertSelectsToPath(StringBuffer & filename, const char * eclPath)
 }
 
 //---------------------------------------------------------------------------------------------------------------------
+static HqlTransformerInfo HqlSimplifiedRecordTransformerInfo("HqlSimplifiedRecordTransformer");
+class HqlSimplifiedRecordTransformer : public QuickHqlTransformer
+{
+public:
+    HqlSimplifiedRecordTransformer() : QuickHqlTransformer(HqlSimplifiedRecordTransformerInfo, nullptr)
+    {
+    }
+
+    virtual IHqlExpression * createTransformed(IHqlExpression * expr)
+    {
+        IHqlExpression * body = expr->queryBody(false);
+        if (expr != body)
+            return transform(body);
+
+        node_operator op = expr->getOperator();
+        switch (op)
+        {
+        case no_record:
+            {
+                HqlExprArray children;
+                bool same = true;
+                ForEachChild(idx, expr)
+                {
+                    IHqlExpression * cur = expr->queryChild(idx);
+                    if (!cur->isAttribute())
+                    {
+                        OwnedHqlExpr mapped = transform(cur);
+                        if (mapped != cur)
+                            same = false;
+                        children.append(*(mapped.getClear()));
+                    }
+                    else
+                        same = false;
+                }
+                if (same)
+                    return LINK(expr);
+                return expr->clone(children);
+            }
+        case no_field:
+            {
+                ITypeInfo * type = expr->queryType();
+                if (type->getTypeCode()==type_alien)
+                    throw MakeStringExceptionDirect(0, "Cannot simplify record");
+                HqlExprArray children;
+                bool same = true;
+                ForEachChild(idx, expr)
+                {
+                    IHqlExpression * cur = expr->queryChild(idx);
+                    if (cur->isAttribute())
+                    {
+                        //Ignore _implicitFpos_Atom's as it cannot be written to the cache
+                        if (cur->queryName()==_implicitFpos_Atom)
+                            same = false;
+                        else
+                        {
+                            HqlExprArray attrs;
+                            bool same2 = true;
+                            ForEachChild(idx2, cur)
+                            {
+                                IHqlExpression * attr = cur->queryChild(idx2)->queryBody();
+                                if (!attr->isConstant())
+                                    throw MakeStringExceptionDirect(0, "Cannot simplify record");
+                                if (attr->isAttribute())
+                                {
+                                    attrs.append(*(LINK(attr)));
+                                }
+                                else
+                                {
+                                    IValue * val = attr->queryValue();
+                                    if (!val)
+                                        throw MakeStringExceptionDirect(0, "Cannot simplify record");
+                                    //For numeric constants values of attributes, it is not possible to specify the type of the number
+                                    //in the generated cache.  The parser treats numbers as INTEGER8 by default.  Here, regardless of
+                                    //the exact numeric type of an attribute's value, generate it as the default integer type (INTEGER8)
+                                    if (val->getTypeCode()==type_int)
+                                        attrs.append(*createConstant(val->getIntValue()));
+                                    else
+                                        attrs.append(*createConstant(LINK(attr)));
+                                    same2 = false;
+                                }
+                            }
+                            if (same2)
+                            {
+                                children.append(*(LINK(cur)));
+                            }
+                            else
+                            {
+                                OwnedHqlExpr newattr = cur->queryBody()->clone(attrs);
+                                children.append(*(newattr.getClear()));
+                                same = false;
+                            }
+                        }
+                    }
+                    else
+                        same = false;
+                }
+                OwnedITypeInfo newType = transformType(type);
+
+                if (type != newType)
+                    return createField(expr->queryId(), newType.getClear(), children);
+
+                if (same)
+                    return LINK(expr);
+                return expr->clone(children);
+            }
+        case no_param:
+            {
+                ITypeInfo * type = expr->queryType();
+                OwnedITypeInfo newType = transformType(type);
+                HqlExprArray children;
+                transformChildren(expr, children);      // could just unwind
+                return createParameter(expr->queryId(), UnadornedParameterIndex, newType.getClear(), children);
+            }
+        }
+
+        return QuickHqlTransformer::createTransformed(expr);
+    }
+
+protected:
+    virtual ITypeInfo * transformType(ITypeInfo * type) override
+    {
+        if (type->queryModifier() == typemod_original)
+            return transformType(type->queryTypeBase()); //Strip all original annotations
+
+        return QuickHqlTransformer::transformType(type);
+    }
+
+};
 
 static IHqlExpression * createSimplifiedDefinitionFromType(ITypeInfo * type)
 {
@@ -359,6 +489,23 @@ static IHqlExpression * createSimplifiedDefinitionFromType(ITypeInfo * type)
     case type_int:
         {
             return createValue(no_simplified, LINK(type));
+        }
+    case type_table:
+        {
+            IHqlExpression * origRecord = queryOriginalRecord(type);
+            // TODO: add support for nested child tables (current not generated correctly in cache)
+            if (getNumChildTables(origRecord)>0)
+                return nullptr;
+            try
+            {
+                HqlSimplifiedRecordTransformer simplifiedRecTransformer;
+                return createDataset(no_simplified, simplifiedRecTransformer.createTransformed(origRecord)); 
+            }
+            catch (IException *e)
+            {
+                e->Release();
+                return nullptr;
+            }
         }
     }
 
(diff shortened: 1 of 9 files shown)


ecl/hql/hqlcache.cpp
```diff
@@ -340,6 +342,161 @@ void convertSelectsToPath(StringBuffer & filename, const char * eclPath)
...
+            {
+                HqlExprArray children;
+                bool same = true;
+                ForEachChild(idx, expr)
+                {
+                    IHqlExpression * cur = expr->queryChild(idx);
+                    if (!cur->isAttribute())
+                    {
+                        OwnedHqlExpr mapped = transform(cur);
+                        if (!mapped)
```
ghalliday: transform should never return null, so this should never occur.

ecl/hql/hqlcache.cpp
```diff
@@ -340,6 +342,161 @@ void convertSelectsToPath(StringBuffer & filename, const char * eclPath)
...
+                            throw MakeStringExceptionDirect(0, "Cannot simplify record");
+                        if (mapped != cur)
+                            same = false;
+                        children.append(*(mapped.getClear()));
+
```
ghalliday: whitespace: extra line

ecl/hql/hqlcache.cpp
```diff
@@ -340,6 +342,161 @@ void convertSelectsToPath(StringBuffer & filename, const char * eclPath)
...
+                if (type->getTypeCode()==type_alien)
+                    throw MakeStringExceptionDirect(0, "Cannot simplify record");
+                HqlExprArray children;
+                bool same = true;
+                ForEachChild(idx, expr)
+                {
+                    IHqlExpression * cur = expr->queryChild(idx);
+                    if (cur->isAttribute())
+                    {
+                        if (cur->queryName()==_implicitFpos_Atom)
```
ghalliday: Should have a comment explaining why this is special-cased.

ecl/hql/hqlcache.cpp
```diff
@@ -340,6 +342,161 @@ void convertSelectsToPath(StringBuffer & filename, const char * eclPath)
...
+                                if (attr->isAttribute())
+                                {
+                                    attrs.append(*(LINK(attr)));
+                                }
+                                else
+                                {
+                                    IValue * val = attr->queryValue();
+                                    if (!val)
+                                        throw MakeStringExceptionDirect(0, "Cannot simplify record");
+                                    if (val->getTypeCode()==type_int)
```
ghalliday: Not clear what this is doing either - this function could do with some comments to explain the logic.

ecl/hql/hqlcache.cpp
```diff
@@ -340,6 +342,161 @@ void convertSelectsToPath(StringBuffer & filename, const char * eclPath)
...
+                                    same2 = false;
+                                }
+                            }
+                            if (same2)
+                            {
+                                children.append(*(LINK(cur)));
+                            }
+                            else
+                            {
+                                OwnedHqlExpr newattr = createExprAttribute(cur->queryName(),attrs);
```
ghalliday: cur->queryBody()->clone(attrs) would be safer - otherwise this might change a no_attr to a no_attr_expr.  Although I suspect that will not cause any problems.

ecl/hql/hqlcache.cpp
```diff
@@ -340,6 +342,161 @@ void convertSelectsToPath(StringBuffer & filename, const char * eclPath)
...
+        }
+
+        return QuickHqlTransformer::createTransformed(expr);
+    }
+
+protected:
+    IHqlExpression * createTransformedField(IHqlExpression * expr)
+    {
+        Owned<ITypeInfo> type = getFullyUnqualifiedType(expr->queryType());
+        OwnedITypeInfo newType = transformType(type.getClear());
```
ghalliday: type.getClear() will leak, should just be type

ecl/hql/hqlcache.cpp
```diff
@@ -340,6 +342,161 @@ void convertSelectsToPath(StringBuffer & filename, const char * eclPath)
...
+    {
+        switch (type->queryModifier())
+        {
+        case typemod_original:
+            {
+                return transformType(type->queryTypeBase()); //Strip all original annotations
+            }
+        case typemod_none:
+            {
+                //Ensure all records with the same format get the same original modifier
```
ghalliday: Why do you want to do this?  Would it be better to not add any modifier.  This looks like it might cause problems when regenerating.

ecl/hql/hqlcache.cpp
```diff
@@ -360,6 +517,21 @@ static IHqlExpression * createSimplifiedDefinitionFromType(ITypeInfo * type)
         {
             return createValue(no_simplified, LINK(type));
         }
+    case type_table:
+        {
+            OwnedHqlExpr record;
+            try {
+                HqlSimplifiedRecordTransformer simplifiedRecordTransformer;
+                record.setown(simplifiedRecordTransformer.createTransformed(queryOriginalRecord(LINK(type))));
```
ghalliday: LINK(type) almost certainly leaks.

ecl/hql/hqlgram2.cpp
```diff
@@ -1931,6 +1931,7 @@ void HqlGram::addAssignall(IHqlExpression *tgt, IHqlExpression *src, const attri
     }
 
     doAddAssignCompoundOwn(assignall, tgt, src, NULL, errpos);
+    //tgt->setTransformExtra(tgt);
```
ghalliday: minor: remove commented out code

ecl/hql/hqlgram2.cpp
```diff
@@ -12279,6 +12281,21 @@ void HqlGram::setTemplateAttribute()
 #endif
 }
 
+bool HqlGram::recordTypesMatch(ITypeInfo * left, ITypeInfo * right)
+{
+    if(lookupCtx.syntaxChecking())
```
ghalliday: trivial: consistency of spaces between if  and bracket.  Also function below.

ecl/hql/hqlcache.cpp
ghalliday: I guess that without it the recordTypeMatch test that prevents simplified expression being created may trigger more often.  Is this true?

ghalliday (commented):
@shamser a few comments.

ecl/hql/hqlcache.cpp
shamser: @ghalliday Yes but this doesn't seem to be making any difference in the regression.  I'm removing this section.

shamser: @ghalliday Please see comment below

shamser:
@ghalliday

ghalliday (commented):
@shamser a couple of comments.  Very close to ready.

shamser:
@ghalliday

ecl/hql/hqlcache.cpp
```diff
@@ -340,6 +342,141 @@ void convertSelectsToPath(StringBuffer & filename, const char * eclPath)
...
+                                    IValue * val = attr->queryValue();
+                                    if (!val)
+                                        throw MakeStringExceptionDirect(0, "Cannot simplify record");
+                                    //For numeric constants values of attributes, it is not possible to specify the type of the number
+                                    //in the generated cache.  The parser treats numbers as INTEGER8 by default.  Here, regardless of
+                                    //the exact numeric type of an attribute's value, generate it as the default integer type (INTEGER8)
+                                    if (val->getTypeCode()==type_int)
+                                        attrs.append(*createConstant(val->getIntValue()));
+                                    else
+                                        attrs.append(*createConstant(LINK(val)));
```
ghalliday: Why not LINK(attr)?

ecl/hql/hqlgram2.cpp
```diff
@@ -1915,7 +1915,7 @@ void HqlGram::addAssignall(IHqlExpression *tgt, IHqlExpression *src, const attri
     IHqlExpression * srcRecord = src->queryRecord();
     IHqlExpression * tgtRecord = tgt->queryRecord();
 
-    if (srcRecord && tgtRecord && recordTypesMatch(srcRecord, tgtRecord) && (tgtOp != no_self) && !haveAssignedToChildren(tgt))
+    if (!lookupCtx.syntaxChecking() && srcRecord && tgtRecord && recordTypesMatch(srcRecord, tgtRecord) && (tgtOp != no_self) && !haveAssignedToChildren(tgt))
```
ghalliday: Slightly hesitant that this will kick in if syntax checking - even if simplified expressions not being used.  (Not going through this code could slow parsing down.)
Probably better would be to check ::recordTypesMatch() since this is should just be an optimization.

ecl/hql/hqlthql.cpp
```diff
@@ -1963,7 +1963,7 @@ void HqltHql::toECL(IHqlExpression *expr, StringBuffer &s, bool paren, bool inTy
             }
             if (child1)
             {
-                toECL(child1, s.append(", "), child1->getPrecedence() < 0, inType);
+                toECL(child1, s.append(", "), child1->getPrecedence() < 0, inType, false, true);
```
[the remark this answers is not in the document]
ghalliday: This should probably be reverted - the false parameter is actually the wrong type.

ecl/hql/hqlcache.cpp
ghalliday: Still hesitant about this.  I would merge it with the code included, but there should be a jira to revisit it - and investigate how many attributes it affects.
Currently if the records do not match then a simplified expression is not created - which means that almost all of the other code is not actually needed - but would be if that condition was relaxed.

ghalliday (commented):
@shamser a few more final comments.

ecl/hql/hqlcache.cpp
shamser: @ghalliday That would work too

ghalliday (commented):
Sorry forgot to submit comments.

ghalliday:
@shamser

ecl/hql/hqlcache.cpp
shamser: @ghalliday Comment added above to explain this line.

shamser:
@ghalliday

ecl/hql/hqlcache.cpp
```diff
@@ -340,6 +342,141 @@ void convertSelectsToPath(StringBuffer & filename, const char * eclPath)
...
+                transformChildren(expr, children);      // could just unwind
+                return createParameter(expr->queryId(), UnadornedParameterIndex, newType.getClear(), children);
+            }
+        }
+
+        return QuickHqlTransformer::createTransformed(expr);
+    }
+
+protected:
+    IHqlExpression * createTransformedField(IHqlExpression * expr)
```
ghalliday: This function is unused

ecl/hql/hqlcache.cpp
```diff
@@ -340,6 +342,141 @@ void convertSelectsToPath(StringBuffer & filename, const char * eclPath)
...
+    {
+        Owned<ITypeInfo> type = getFullyUnqualifiedType(expr->queryType());
+        OwnedITypeInfo newType = transformType(type);
+        IIdAtom * id = expr->queryId();
+        return createField(id, newType.getClear(), nullptr);
+    }
+    ITypeInfo * transformType(ITypeInfo * type)
```
ghalliday: should be marked as virtual, and these days we should really use override on all the overridden functions (this and above).

ecl/hql/hqlcache.cpp
```diff
@@ -360,6 +497,23 @@ static IHqlExpression * createSimplifiedDefinitionFromType(ITypeInfo * type)
...
+            {
+                HqlSimplifiedRecordTransformer simplifiedRecordTransformer;
+                record.setown(simplifiedRecordTransformer.createTransformed(queryOriginalRecord(type)));
+            }
+            catch (IException *e)
+            {
+                e->Release();
+                return nullptr;
+            }
+            if (!record || getNumChildTables(record)>0)
```
ghalliday: with the final code record cannot be null.
The check for getNumChildTables() could be done before the record is transformed - avoiding the transform.

ghalliday: Please add a comment why child tables are not allowed...

ghalliday (commented):
@shamser some final minor comments
Please can you also squash.
I'll run a full regression test and see if I spot any other issues, but I think it is probably close to ready to merge.

shamser:
@ghalliday

ghalliday:
I am still in two minds about this change.  It has a few effects when syntax checking:

- It changes the meta information that is generated for definitions that are simplified.  E.g., records become unnamed rather than references to other records.  (Unless ignore simplified is specified.)
- It removes various warnings, even if simplified definitions are being ignored
- It seems to add a few warnings (e.g., for allnodes1)

However painful the decision I am seriously questioning whether the code to process simplified expressions should be removed instead.

AttilaVamos:
@AttilaVamos Those spray errors has been caused by my merged PR-https://github.com/hpcc-systems/HPCC-Platform/pull/12206 . I'm working on the fix. I am sorry about inconvenience caused.

AttilaVamos:
@shamser  back to you.

ghalliday:
Not going to merge.  Likely to disable the functionality,

PR #12010, opened 2018-12-18
HPCC-21223 Roxie may terminate on exceptions in 7.2 branch

richardkchapman:
Can't throw exceptions out of a destructor

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Testing:
Observed in modellingWithR.ecl test when a syntax error was introduced into the R code.

### DIFF
--- a/system/jlib/jthread.cpp
+++ b/system/jlib/jthread.cpp
@@ -547,7 +547,7 @@ CThreadedPersistent::CThreadedPersistent(const char *name, IThreaded *_owner) : 
 
 CThreadedPersistent::~CThreadedPersistent()
 {
-    join(INFINITE);
+    join(INFINITE, false);
     halt = true;
     sem.signal();
     athread.join();
@@ -596,7 +596,7 @@ void CThreadedPersistent::start()
     sem.signal();
 }
 
-bool CThreadedPersistent::join(unsigned timeout)
+bool CThreadedPersistent::join(unsigned timeout, bool throwException)
 {
     unsigned expected = s_running;
     if (state.compare_exchange_strong(expected, s_joining))
@@ -607,11 +607,11 @@ bool CThreadedPersistent::join(unsigned timeout)
             if (state.compare_exchange_strong(expected, s_running)) // if still joining, restore running state
                 return false;
             // if here, threadmain() set s_ready after timeout and has or will signal
-            if (!joinSem.wait(60000)) // should be instant
+            if (!joinSem.wait(60000) && throwException) // should be instant
                 throwUnexpected();
             return true;
         }
-        if (exception.get())
+        if (throwException && exception.get())
         {
             // switch back to ready state and throw
             Owned<IException> e = exception.getClear();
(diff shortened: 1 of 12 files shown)


richardkchapman:
@ghalliday Please review

system/jlib/jthread.cpp
```diff
@@ -547,10 +547,26 @@ CThreadedPersistent::CThreadedPersistent(const char *name, IThreaded *_owner) :
 
 CThreadedPersistent::~CThreadedPersistent()
 {
-    join(INFINITE);
+    try
+    {
+        join(INFINITE);
```
ghalliday: Better would be a parameter on join() to prevent throwing an exception - otherwise if this is cleaning up as a result of an exception being thrown it will terminate anyway.

system/jlib/jthread.cpp
```diff
@@ -547,10 +547,26 @@ CThreadedPersistent::CThreadedPersistent(const char *name, IThreaded *_owner) :
...
+    {
+        EXCLOG(E);
+        E->Release();
+    }
     halt = true;
     sem.signal();
-    athread.join();
+    try
+    {
+        athread.join();
```
ghalliday: I'm not sure this can thrown an exception.

ghalliday (commented):
@richardkchapman

richardkchapman:
@jakesmith There were some instances of similar potential problem in thor. Some I could easily fix, some I wasn't quite sure - please take a look.

thorlcr/activities/keyedjoin/thkeyedjoinslave-legacy.cpp
```diff
@@ -1038,7 +1038,7 @@ class CKeyedJoinSlave : public CSlaveActivity, implements IJoinProcessor, implem
                     pendingSendsSem.signal();
                 }
                 crit.leave();
-                threaded.join();
+                threaded.join(INFINITE, false);
```
jakesmith: I don't think this one is necessary, i.e. ok to throw exception here in CKeyedFetchHandler::stop()

jakesmith:
@richardkchapman - looked through changes and looked through all of Thor's join cases.

There's one change I think should allow throw (see comment), but otherwise looks good.

thorlcr/activities/keyedjoin/thkeyedjoinslave-legacy.cpp
richardkchapman: Ok, but stop() is called from the destructor I think

richardkchapman:
@jakesmith See reply

thorlcr/activities/keyedjoin/thkeyedjoinslave-legacy.cpp
jakesmith: @richardkchapman - This stop() isn't. Destructor in question is ~CKeyedFetchHandler()

richardkchapman: Ok, fixed

richardkchapman:
@jakesmith Are you happy now?

jakesmith:
@richardkchapman - yep looks good to merge.

richardkchapman:

ghalliday:
Looks ok.  One concern is how do we ensure that the correct parameter is supplied on new code?  Could remove the default value from the join?  probably ok as it is though.

PR #12016, opened 2018-12-19
HPCC-21231 Fix KJ fetch race

jakesmith:
Also fix potential deadlock if read ahead creates too many ready
CJoinGroup's after KJ is stopped.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/thorlcr/activities/keyedjoin/thkeyedjoinslave.cpp
+++ b/thorlcr/activities/keyedjoin/thkeyedjoinslave.cpp
@@ -422,6 +422,13 @@ class CKeyedJoinSlave : public CSlaveActivity, implements IJoinProcessor
         unsigned blocked = 0; // number of callers blocked due to exceeding max+leeway.
         Semaphore sem;
         CriticalSection crit;
+        bool enabled = true;
+
+        void unblock()
+        {
+            sem.signal(blocked);
+            blocked = 0;
+        }
     public:
         CLimiter()
         {
@@ -438,6 +445,8 @@ class CKeyedJoinSlave : public CSlaveActivity, implements IJoinProcessor
         {
             {
                 CriticalBlock b(crit);
+                if (!enabled)
+                    return false;
                 if (count++ < max+leeway)
                     return false;
                 ++blocked;
@@ -448,6 +457,8 @@ class CKeyedJoinSlave : public CSlaveActivity, implements IJoinProcessor
         {
             {
                 CriticalBlock b(crit);
+                if (!enabled)
+                    return false;
                 if (count < max+leeway)
                 {
                     ++count;
@@ -465,16 +476,30 @@ class CKeyedJoinSlave : public CSlaveActivity, implements IJoinProcessor
         void dec()
         {
             CriticalBlock b(crit);
+            if (!enabled)
+                return;
             --count;
             if (blocked && (count < max))
-            {
-                sem.signal(blocked);
-                blocked = 0;
-            }
+                unblock();
         }
         void block()
         {
             sem.wait();
+        }
+        void disable()
+        {
+            CriticalBlock b(crit);
+            enabled = false;
+            if (blocked)
+                unblock();
+        }
+        void reset()
+        {
+            CriticalBlock b(crit);
+            enabled = true;
+            if (blocked)
+                unblock();
+            count = 0;
         }
     };
     // There is 1 of these per part, but # running is limited
@@ -1987,10 +2012,7 @@ class CKeyedJoinSlave : public CSlaveActivity, implements IJoinProcessor
                 }
                 lhsRow.setown(inputStream->nextRow());
                 if (!lhsRow)
-                {
-                    stopReadAhead();
                     break;
-                }
             }
             Linked<CJoinGroup> jg;
             if (helper->leftCanMatch(lhsRow))
@@ -2043,6 +2065,7 @@ class CKeyedJoinSlave : public CSlaveActivity, implements IJoinProcessor
             }
         }
         while (!endOfInput);
+        stopReadAhead();
     }
     const void *doDenormTransform(RtlDynamicRowBuilder &target, CJoinGroup &group)
     {
@@ -2484,6 +2507,8 @@ public:
         currentAdded = 0;
         eos = false;
         endOfInput = false;
+        lookupThreadLimiter.reset();
+        fetchThreadLimiter.reset();
         keyLookupHandlers.init();
         fetchLookupHandlers.init();
         readAheadThread.start();
@@ -2638,6 +2663,11 @@ public:
     virtual void stop() override
     {
         endOfInput = true; // signals to readAhead which is reading input, that is should stop asap.
+
+        // could be blocked in readAhead(), because CJoinGroup's are no longer being processed
+        pendingKeyLookupLimiter.disable();
+        doneListLimiter.disable();
+
         readAheadThread.join();
         keyLookupHandlers.stop();
         fetchLookupHandlers.stop();
@@ -2659,6 +2689,8 @@ public:
     // IJoinProcessor
     virtual void onComplete(CJoinGroup *joinGroup) override
     {
+        if (hasStopped())
+            return;
         bool doneListMaxHit = false;
         // moves complete CJoinGroup's from pending list to done list
         {
--- a/thorlcr/slave/slavmain.cpp
+++ b/thorlcr/slave/slavmain.cpp
@@ -103,45 +103,36 @@ void disableThorSlaveAsDaliClient()
 class CKJService : public CSimpleInterfaceOf<IKJService>, implements IThreaded, implements IExceptionHandler
 {
     const unsigned defaultMaxCachedKJManagers = 1000;
+    const unsigned defaultMaxCachedFetchContexts = 1000;
     const unsigned defaultKeyLookupMaxProcessThreads = 16;
 
     unsigned uniqueId = 0;
     CThreadedPersistent threaded;
     mptag_t keyLookupMpTag = TAG_NULL;
     bool aborted = false;
-    unsigned numCached = 0;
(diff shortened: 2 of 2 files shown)


thorlcr/activities/keyedjoin/thkeyedjoinslave.cpp
```diff
@@ -2043,6 +2065,7 @@ class CKeyedJoinSlave : public CSlaveActivity, implements IJoinProcessor
             }
         }
         while (!endOfInput);
+        stopReadAhead();
```
jakesmith: Note to reviewer: stopReadAhead() might not have been called before is KJ::stop() was called and thus set endOfInput=true.

thorlcr/slave/slavmain.cpp
```diff
@@ -1010,13 +969,14 @@ class CKJService : public CSimpleInterfaceOf<IKJService>, implements IThreaded,
...
-    CKMContextHT cachedKMs; // NB: HT by {actId, fname, crc} links IKeyManager's within containers
-    OwningSimpleHashTableOf<CKMContainer, unsigned> cachedKMsByHandle; // HT by { handle }, links IKeyManager's within containers.
-    OwningSimpleHashTableOf<CFetchContext, FetchKey> fetchContextsHT;
-    std::unordered_map<unsigned, CFetchContext *> fetchContextsByHandleHT;
+    std::unordered_map<unsigned, CActivityContext *> activityContextsHT;
+    std::unordered_map<CLookupKey, CKeyLookupContext *, CLookupKey> keyLookupContextsHT;
+    std::unordered_map<CLookupKey, Owned<CKMKeyEntry>, CLookupKey> cachedKMs;
+    std::unordered_map<CFetchKey, Owned<CFCKeyEntry>, CFetchKey> cachedFetchContexts;
+    std::unordered_map<unsigned, Owned<CKMContainer>> activeKManagersByHandle;
+    std::unordered_map<unsigned, Owned<CFetchContext>> activeFetchContextsByHandle;
```
jakesmith: Note to reviewer: the change to use std::unordered_map's here, isn't directly related, but I would have need yet more boiler plate code for SuperHashTable's as a result of this change, so standardizing on STL here made the code cleaner.

NB: I tested for performance differences (none/tiny -/+ variance) and leaks with valrgrind.

thorlcr/slave/slavmain.cpp
```diff
@@ -1049,116 +1009,107 @@ class CKJService : public CSimpleInterfaceOf<IKJService>, implements IThreaded,
...
-        activity_id id = key.id; // JCSMORE annoying, needed because IMPLEMENT_SUPERHASHTABLEOF_REF_FIND doesn't define const (should look at)
-        Linked<CActivityContext> activityCtx = activityContextsHT.find(id);
-        if (!activityCtx)
-        {
-            activityCtx.setown(createActivityContext(job, key.id, createCtxMb));
-            activityContextsHT.replace(*activityCtx); // NB: does not link/take ownership
-        }
-        keyLookupContext = createLookupContext(activityCtx, key);
-        keyLookupContextsHT.replace(*keyLookupContext); // NB: does not link/take ownership
+        Owned<CActivityContext> activityCtx = ensureActivityContext(job, key.id, createCtxMb);
```
jakesmith: Note to reviewer: lines 1061-1068 were equivalent to the pre-existing method, ensureActivityContext() - so this is just clearing up.

thorlcr/slave/slavmain.cpp
```diff
@@ -1049,116 +1008,107 @@ class CKJService : public CSimpleInterfaceOf<IKJService>, implements IThreaded,
...
         {
             CriticalBlock b(lCCrit);
-            activityContextsHT.removeExact(activityContext);
+            activityContextsHT.erase(activityContext->queryId());
         }
     }
     void freeLookupContext(CKeyLookupContext *_lookupContext)
     {
         CriticalBlock b(lCCrit);
         Owned<CKeyLookupContext> keyLookupContext = _lookupContext;
```
ghalliday: trivial: can be before the CriticialBlock.

thorlcr/slave/slavmain.cpp
```diff
@@ -1049,116 +1008,107 @@ class CKJService : public CSimpleInterfaceOf<IKJService>, implements IThreaded,
...
     void removeActivityContext(CActivityContext *activityContext)
     {
         CriticalBlock b(lCCrit);
-        activityContextsHT.removeExact(activityContext);
+        activityContextsHT.erase(activityContext->queryId());
     }
     void freeActivityContext(CActivityContext *_activityContext)
     {
         Owned<CActivityContext> activityContext = _activityContext;
         if (!activityContext->IsShared())
```
ghalliday: I think this test needs to be inside the critical block - otherwise another thread could hit it in the cache before it is removed.

thorlcr/slave/slavmain.cpp
```diff
@@ -1214,21 +1175,60 @@ class CKJService : public CSimpleInterfaceOf<IKJService>, implements IThreaded,
...
+        if (numFCCached == maxCachedFetchContexts)
+        {
+            CFetchContext &oldest = cachedFCsMRU.item(0);
+            auto it = cachedFetchContexts.find(oldest.queryKey());
+            assertex(it != cachedFetchContexts.end());
+            CFCKeyEntry *fce = it->second;
+            verifyex(fce->remove(&oldest));
+            if (0 == fce->count())
+                cachedFetchContexts.erase(it);
+            verifyex(cachedFCsMRU.zap(oldest));
```
ghalliday: .remove(0) would probably be better.  This is also potentially painful - since if there are 1000 items you will be copying 8K of memory on each remove.  Each time one is removed from the middle it will copy half that.

Could do with a LRU class which solved this once.

ghalliday: Hopefully the length is generally much shorter than that - what would you expect it to be?

ghalliday (commented):
@jakesmith some initial comments.  I haven't really looked at any changes in the logic - although it doesn't look like there are too many.

thorlcr/activities/keyedjoin/thkeyedjoinslave.cpp
```diff
@@ -2659,6 +2689,8 @@ class CKeyedJoinSlave : public CSlaveActivity, implements IJoinProcessor
     // IJoinProcessor
     virtual void onComplete(CJoinGroup *joinGroup) override
     {
+        if (endOfInput) // once actiivty stopped.
```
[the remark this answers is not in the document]
jakesmith: hm, well spotted, stop() sets endOfInput, but it is also set as you say, when input is exhausted.
The intent was to bail out after stop() when you know no more are wanted.
But it is incorrect to bail out and ignore groups once input is exhausted.

I mulled over whether to bother putting this shortcut in at all, without it, it just means that there will be a few more groups added (in the event of stop()) that aren't wanted. The main thing is to ensure that those trailing groups can't cause deadlock, which is what the disable() calls prevent.

I'll make a change.

thorlcr/slave/slavmain.cpp
jakesmith: agreed

thorlcr/slave/slavmain.cpp
jakesmith: agree, will change.

thorlcr/slave/slavmain.cpp
jakesmith: I do expect typical cases to be quite small, but I also I don't expect these MRU movement to be frequent.
It is on key part open / fetch part open, they should remain open and out of cache for as long as the subgraph is running, and re-enter the cache when done.

If there were high numbers, and parts were opened and closed v. quickly, then that would be more of a concern. NB: that should even happen in a child query case - i.e. they should remain open.

I did think about using a QueueOf here instead of an array which would be a bit more efficient, but   wasn't that concerned.

jakesmith:
@ghalliday - please review changes.

thorlcr/slave/slavmain.cpp
```diff
@@ -1214,21 +1173,60 @@ class CKJService : public CSimpleInterfaceOf<IKJService>, implements IThreaded,
...
         {
             CKMContainer &oldest = cachedKMsMRU.item(0);
-            verifyex(removeKeyManager(oldest.queryHandle())); // also removes from cachedKMsMRU
+            auto it = cachedKMs.find(oldest.queryCtx().queryKey());
+            assertex(it != cachedKMs.end());
+            CKMKeyEntry *kme = it->second;
+            verifyex(kme->remove(kmc));
+            if (0 == kme->count())
+                cachedKMs.erase(it);
+            verifyex(cachedKMsMRU.zap(oldest));
```
ghalliday: could use
    cachedKMsMRU.remove(0)
but not much more efficient since it matches the first element

thorlcr/slave/slavmain.cpp
```diff
@@ -1214,21 +1173,60 @@ class CKJService : public CSimpleInterfaceOf<IKJService>, implements IThreaded,
     void addToKeyManagerCache(CKMContainer *kmc)
     {
         CriticalBlock b(kMCrit);
```
ghalliday: This, and the other cache code, should cope with maxCachedKJManagers == 0, since it can be set from an option, and that would crash the following code.
It might also be useful so you can compare the effect of having no cache...

ghalliday (commented):
@jakesmith a final couple of comments.  Looks very close to ready.

jakesmith:
@ghalliday - squashed

ghalliday:
@richardkchapman I think this is ready to merge.  You may want to sanity check it before merging.

PR #12022, opened 2019-01-02
HPCC-20633 Soapplus stress test should keep track of failed transactions

mayx:
- Add the number of failed transactions to the stats struct
- Increment the number inside sendStressRequest for failed transations

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Premature optimization

### DIFF
--- a/esp/tools/soapplus/http.cpp
+++ b/esp/tools/soapplus/http.cpp
@@ -84,6 +84,7 @@ HttpStat::HttpStat()
     totalreqlen = 0;
     totalresplen = 0;
     numrequests = 0;
+    numfails = 0;
 }
 
 void HttpStat::printStat(FILE* ofile)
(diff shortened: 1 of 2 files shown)


esp/tools/soapplus/http.cpp
```diff
@@ -1313,7 +1318,15 @@ int HttpClient::sendStressRequest(StringBuffer& request, HttpStat* stat, Owned<C
                         fprintf(m_logfile, "%s", recvbuf);
                 }
                 else
+                {
+                    if (total_len == 0)
+                    {
```
mckellyln: What happened before this change, if len < 0 and total_len == 0 ?  DId it return -1 also ??

mckellyln:
@mayx just one question

esp/tools/soapplus/http.cpp
mayx: No it didn't return -1. But I think total_len == 0 means no response from the server so should be treated as an error. What do you think?

mckellyln: Is a reply ALWAYS expected ?  If so then yes, if len <= 0 and total_len == 0 I would say its an error.
If a reply is not ALWAYS guaranteed then if len == 0 and total_len == 0 it would be ok.
Also, should the socket be closed in this case ?

mayx: Yes a reply is always expected. This code is for non-persistent connection, in which case the socket will be automatically closed in the destructor of CSimpleSocket.

mayx:
@mckellyln question replied...

mckellyln:
@mayx ok, then looks good to merge

mckellyln (approved):
Looks good to merge

PR #12025, opened 2019-01-05
H3 plugin

GordonSmith:
PR only for reference (I need help) When submitting the following query I get the error below:

ECL:
```
h3 := SERVICE : plugin('h3'), library('h3'), namespace('h3'), CPP, PURE
  STRING latLngToH3(CONST REAL4 lat, CONST REAL4 lng, CONST UNSIGNED4 resolution) : cpp,action,context,fold,entrypoint='latLngToH3';
END;

h3.latLngToH3(0, 0, 10);
```

Error:
```
/var/lib/HPCCSystems/myeclccserver/libW20190105-082335.so: undefined symbol: _ZN2h310latLngToH3EP12ICodeContextRjRPcffj
```


GordonSmith:
I figured it out - I had the wrong return type in the ECL declaration!

PR #12046, opened 2019-01-10
HPCC-19155 Multiple problems with lazy caching.

jakesmith:
When the limit is exceeded, it could cause indexes that used
that lazy cache to fail dealing with a file that had been closed.
Resulting in a checked_pread error (if local), or a
RFSERR_InvalidFileIOHandle if dealing with a remote file.

It could also cause a crash, if multiple uses of the same
physical file raced in and out of the cache and were then removed
after use, due to a bug in the handling of the objects during
removal from the cache.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Testing:

Tested extensively with fileCacheLimit set to 1, with a custom query that runs lots of KJ's for a while that force the part in the cache to constantly swap.
Also ran valgrind with 1 of these queries to check for issues and/or leaks.

### DIFF
--- a/thorlcr/activities/fetch/thfetchslave.cpp
+++ b/thorlcr/activities/fetch/thfetchslave.cpp
@@ -43,10 +43,11 @@
 
 struct FPosTableEntryIFileIO : public FPosTableEntry
 {
-    FPosTableEntryIFileIO() { file = NULL; }
-    ~FPosTableEntryIFileIO() { ::Release(file); }
-    unsigned location;
-    IDelayedFile *file;
+    ~FPosTableEntryIFileIO()
+    {
+        ::Release(file);
+    }
+    IFileIO *file = nullptr;
 };
 
 class CFetchStream : public IRowStream, implements IStopInput, implements IFetchStream, public CSimpleInterface
@@ -57,7 +58,6 @@ class CFetchStream : public IRowStream, implements IStopInput, implements IFetch
     Linked<IExpander> eexp;
 
     FPosTableEntryIFileIO *fPosMultiPartTable;
-    unsigned tableSize;
     unsigned files, offsetCount;
     CriticalSection stopsect;
     CPartDescriptorArray parts;
@@ -157,9 +157,7 @@ public:
                 e->base = part.queryProperties().getPropInt64("@offset");
                 e->top = e->base + part.queryProperties().getPropInt64("@size");
                 e->index = f;
-
-                Owned<IDelayedFile> lfile = queryThor().queryFileCache().lookup(owner, logicalFilename, part);
-                e->file = lfile.getClear();
+                e->file = queryThor().queryFileCache().lookupIFileIO(owner, logicalFilename, part); // NB: freed by FPosTableEntryIFileIO dtor
             }
         }
     }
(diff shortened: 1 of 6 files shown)


thorlcr/activities/fetch/thfetchslave.cpp
```diff
@@ -43,10 +43,11 @@
 
 struct FPosTableEntryIFileIO : public FPosTableEntry
 {
-    FPosTableEntryIFileIO() { file = NULL; }
-    ~FPosTableEntryIFileIO() { ::Release(file); }
-    unsigned location;
```
jakesmith: NB: not directly related to the issue, but noticed whilst reviewing changed.
'location' was uninitialized, but only ever used in an error report.

thorlcr/activities/fetch/thfetchslave.cpp
```diff
@@ -57,7 +58,6 @@ class CFetchStream : public IRowStream, implements IStopInput, implements IFetch
     Linked<IExpander> eexp;
 
     FPosTableEntryIFileIO *fPosMultiPartTable;
-    unsigned tableSize;
```
jakesmith: NB: not directly related to the issue, but noticed whilst reviewing changed.

unused member var.

jakesmith:
@mckellyln - please review

thorlcr/activities/fetch/thfetchslave.cpp
```diff
@@ -181,8 +179,8 @@ class CFetchStream : public IRowStream, implements IStopInput, implements IFetch
         keyOutStream.setown(distributor->connect(keyRowIf, keyIn, fposHash, NULL, NULL));
     }
     virtual IRowStream *queryOutput() override { return this; }
-    virtual IFileIO *getPartIO(unsigned part) override { assertex(part<files); return fPosMultiPartTable[part].file->getFileIO(); }
-    virtual StringBuffer &getPartName(unsigned part, StringBuffer &out) override { return getPartFilename(parts.item(part), fPosMultiPartTable[part].location, out, true); }
+    virtual IFileIO *getPartIO(unsigned part) override { assertex(part<files); return LINK(fPosMultiPartTable[part].file); }
+    virtual StringBuffer &getPartName(unsigned part, StringBuffer &out) override { return getPartFilename(parts.item(part), 0, out, true); }
```
mckellyln: Why change second arg of getPartFilename() from fPosMultiPartTable[part].location to 0 ?

thorlcr/graph/thgraph.hpp
```diff
@@ -1165,9 +1165,11 @@ interface IExpander;
 interface IThorFileCache : extends IInterface
 {
     virtual bool remove(const char *filename) = 0;
-    virtual IDelayedFile *lookup(CActivityBase &activity, const char *logicalFilenae, IPartDescriptor &partDesc, IExpander *expander=NULL) = 0;
+    virtual IFileIO *lookupIFileIO(CActivityBase &activity, const char *logicalFilenae, IPartDescriptor &partDesc, IExpander *expander=nullptr) = 0;
```
mckellyln: trviial, *logicalFilenae not *logicalFilename ??  But it was that way before, so ok.

richardkchapman:
Is this something that should be fixed in 7.0.x series?

jakesmith:
>Is this something that should be fixed in 7.0.x series?

It's a long standing issue and not a regression, and there's no great pressure from users afaik.
But it is an issue that bites them once in a while, which is usually mitigated by upping the cache size as a workaround.
So I'm relaxed about it not going into until 7.2

mckellyln:
I am about done with reivew.  It looks very good.  A work-around exists to increase the fileCacheLimit setting, but I am fine with it going into 7.0.x series.

mckellyln (approved):

mckellyln:
@jakesmith nice code, looks good.
I tested several kj cases with a fileCacheLimit = 2 and fileCachePurgeN = 1 all good.
One question above about the arg to getPartFilename().

thorlcr/activities/fetch/thfetchslave.cpp
jakesmith: @mckellyln - because it was never used (only in error), see my comment in PR I added before near top.

jakesmith:
@mckellyln - replied to comment.

PR #12057, opened 2019-01-11
HPCC-20944 Hthor exe cannot create WLOCAL_XXXXX temp dir

mckellyln:
If cannot create temp dir then fall back to an appropriate dir.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Testing:

hthor regression and hthor compiled standalone executable tests as a user other than hpcc and with sudo.

### DIFF
--- a/ecl/eclagent/eclagent.cpp
+++ b/ecl/eclagent/eclagent.cpp
@@ -595,22 +595,7 @@ const char *EclAgent::queryTempfilePath()
     if (agentTempDir.isEmpty()) 
     {
         StringBuffer dir;
-        if (!getConfigurationDirectory(agentTopology->queryPropTree("Directories"),"temp","eclagent",agentTopology->queryProp("@name"),dir))
-        {
-            dir.clear();
-#ifdef _WIN32
-            char path[_MAX_PATH+1];
-            DWORD len = GetEnvironmentVariable("TEMP",path,sizeof(path));
-            if (len)
-                dir.append(path);
-            else
-                dir.append("c:");
-            dir.append("\\HPCCSystems\\hthortemp");
-#else
-            dir.append("/tmp/HPCCSystems/hthortemp");
-#endif
-        }
-        recursiveCreateDirectory(dir.str());
+        getTempFilePath(dir, "eclagent", agentTopology);
         agentTempDir.set(dir.str());
     }
     return agentTempDir.str();
--- a/system/jlib/jutil.cpp
+++ b/system/jlib/jutil.cpp
@@ -2868,29 +2868,103 @@ int parseCommandLine(const char * cmdline, MemoryBuffer &mb, const char** &argvo
     return 0;
 }
 
+#ifndef _WIN32
+# define ALL_PERMS (S_IRUSR|S_IWUSR|S_IXUSR|S_IRGRP|S_IWGRP|S_IXGRP|S_IROTH|S_IWOTH|S_IXOTH)
+#endif
+
 jlib_decl StringBuffer &getTempFilePath(StringBuffer & target, const char * component, IPropertyTree * pTree)
 {
     StringBuffer dir;
     if (pTree)
         getConfigurationDirectory(pTree->queryPropTree("Directories"),"temp",component,pTree->queryProp("@name"),dir);
-    if (!dir.length())
+    bool ok = false;
+    if (!dir.isEmpty())
+    {
+        ok = recursiveCreateDirectory(dir.str());
+        if (ok)
+        {
+#ifdef _WIN32
+            return target.set(dir);
+#else
+            int srtn = access(dir.str(), R_OK|W_OK|X_OK);
+            if (!srtn)
+            {
+                chmod(dir.str(), ALL_PERMS);
+                return target.set(dir);
+            }
+#endif
+        }
+    }
+
+    // runtime dir
+    dir.clear();
+    dir.append(RUNTIME_DIR);
+    dir.append(PATHSEPCHAR).append("hpcc-data").append(PATHSEPCHAR).append(component);
+    dir.append(PATHSEPCHAR).append("temp");
+    ok = recursiveCreateDirectory(dir.str());
+    if (ok)
     {
 #ifdef _WIN32
-        char path[_MAX_PATH+1];
-        if(GetTempPath(sizeof(path),path))
-            dir.append(path).append("HPCCSystems\\hpcc-data");
-        else
-            dir.append("c:\\HPCCSystems\\hpcc-data\\temp");
+        return target.set(dir);
 #else
-        dir.append(getenv("TMPDIR"));
-        if (!dir.length())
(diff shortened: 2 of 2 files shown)


mckellyln:
@ghalliday or @richardkchapman please review.

mckellyln:
Corrected target branch from master to candidate-7.0.8

ecl/eclagent/eclagent.cpp
```diff
@@ -605,12 +614,20 @@ const char *EclAgent::queryTempfilePath()
                 dir.append(path);
             else
                 dir.append("c:");
-            dir.append("\\HPCCSystems\\hthortemp");
+            dir.append("\\HPCCSystems\\").append(instance.str());
 #else
-            dir.append("/tmp/HPCCSystems/hthortemp");
+            dir.append(getenv("TMPDIR"));
+            if (!dir.length())
+                dir.append("/tmp");
```
ghalliday: Change looks ok.  A couple of questions:
- Why not use RUNTIME_DIR as you mentioned in the jira.
- Why the switch to instance.str() instead of hthortemp
Might there be a problem if the temp files get very large?

ghalliday:
@richardkchapman

ecl/eclagent/eclagent.cpp
mckellyln: RUNTIME_DIR suffers the same issue, that it is owned by hpcc, and so if TMPDIR is not set then it results in the same permission failure.
In this use case the instance is empty and I thought it a bit more consistent to use the instance if present and use hthortemp if it is empty.

mckellyln:
ok I can move stuff around a bit.

richardkchapman:
Applying the fix in getTempDirectory will also fix the ongoing issue in the regression suite when trying to syntax check / create archive for code that includes embedded Java.

mckellyln:
Working on it today.

mckellyln:
@richardkchapman - moved logic into getTempFilePath() in jutil.cpp

system/jlib/jutil.cpp
```diff
@@ -2868,29 +2868,103 @@ int parseCommandLine(const char * cmdline, MemoryBuffer &mb, const char** &argvo
...
+            return target.set(dir);
 #else
-        dir.append(getenv("TMPDIR"));
-        if (!dir.length())
-            dir.append(RUNTIME_DIR);
-        dir.append("/hpcc-data/temp");
+            int srtn = access(dir.str(), R_OK|W_OK|X_OK);
+            if (!srtn)
+            {
+                chmod(dir.str(), ALL_PERMS);
```
mckellyln: Should we open up perms on these temp dirs ?

system/jlib/jutil.cpp
```diff
@@ -2868,29 +2868,103 @@ int parseCommandLine(const char * cmdline, MemoryBuffer &mb, const char** &argvo
...
+#else
+    // uniq tmp - perhaps previous dirs above were owned by others ...
+    dir.clear();
+    dir.append(tmpdir);
+    dir.append(PATHSEPCHAR).append("HPCCSystems-XXXXXX");
+    char templt[256];
+    strcpy(templt, dir.str());
+    char *td = mkdtemp(templt);
+    if (!td)
+        throw MakeStringException(-1, "Unable to create temp directory");
```
mckellyln: Should we throw if cannot succeed ?  Or just return and let parent throw when trying to open/create a file ?

richardkchapman:
@mckellyln

system/jlib/jutil.cpp
mckellyln: Agree, but it goes to the source of this issue.  If, when running jobs as the hpcc user, the temp dir was opened up, then running as a different user would have worked ok.  Without this we fall through to the final logic block which creates a unique temp dir for every user.

mckellyln: @richardkchapman

richardkchapman: Allowing other users access to the hpcc temp directory is definitely NOT what we want. We want users to use temp directories that they have access to, not to give them access to whatever directory they happen to feel like using.

You could argue that we should not try to use the hpcc temp directory at all if the user is not hpcc (where by "hpcc" I mean the value of RUNTIME_USER macro), even if by some misconfiguration we did have access to it.

richardkchapman: @mckellyln

mckellyln: Updated with second commit to remove chmod() calls.

mckellyln:
@richardkchapman updated

system/jlib/jutil.cpp
```diff
@@ -2873,24 +2873,84 @@ jlib_decl StringBuffer &getTempFilePath(StringBuffer & target, const char * comp
...
+    // runtime dir
+    dir.clear();
+    dir.append(RUNTIME_DIR);
+    dir.append(PATHSEPCHAR).append("hpcc-data").append(PATHSEPCHAR).append(component);
+    dir.append(PATHSEPCHAR).append("temp");
+    ok = recursiveCreateDirectory(dir.str());
+    if (ok)
+    {
+#ifdef _WIN32
+        return target.set(dir);
```
richardkchapman: Why no access check on Windows?

richardkchapman:
@mckellyln Couple of questions

system/jlib/jutil.cpp
mckellyln: From MSDN documentation: "in Windows 2000 and later operating systems, all directories have read and write access" - so I didn't feel a specific check was required.
To check directory permissions accurately may require much more Windows specific Access Token code.

system/jlib/jutil.cpp
```diff
@@ -2873,24 +2873,84 @@ jlib_decl StringBuffer &getTempFilePath(StringBuffer & target, const char * comp
...
+        {
 #ifdef _WIN32
-        char path[_MAX_PATH+1];
-        if(GetTempPath(sizeof(path),path))
-            dir.append(path).append("HPCCSystems\\hpcc-data");
-        else
-            dir.append("c:\\HPCCSystems\\hpcc-data\\temp");
+            return target.set(dir);
+#else
+            int srtn = access(dir.str(), R_OK|W_OK|X_OK);
```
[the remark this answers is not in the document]
mckellyln: Yes.  There is a:
```
    if (checkDirExists(name))
```
in the recursiveCreateDirectory() code path, so that if it exists and its a dir then its ok, and if it is a file or any part of it is a file then a mkdir would fail and fall through to the next attempt.

system/jlib/jutil.cpp
mckellyln: I can add a call to access(R_OK|W_OK) for Windows to be more similar.

mckellyln:
@richardkchapman

PR #12066, opened 2019-01-16
HPCC-21301 Fix some test differences seen on larger clusters.

jakesmith:
Larger cluster configurations exposed issues with a couple of
the regression tests, that produced inconsistent results,
depending on the size of the cluster (e.g. due to ordering issues).

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/testing/regress/ecl/prefetch2.ecl
+++ b/testing/regress/ecl/prefetch2.ecl
@@ -26,6 +26,7 @@ useTranslation := #IFDEFINED(root.useTranslation, false);
 //--- end of version configuration ---
 
 #option ('layoutTranslation', useTranslation);
+#onwarning(4523, ignore);
 
 import $.setup;
 import setup.TS;
@@ -40,12 +41,12 @@ END;
 
 outrec trans1(in L) := TRANSFORM
   SELF.word := L.word;
-  SELF.doc := SORTED(Files.getSearchSuperIndex()(KEYED(kind = TS.kindType.TextEntry),keyed(word = l.word)), doc)[1].doc;
+  SELF.doc := SORT(Files.getSearchSuperIndex()(KEYED(kind = TS.kindType.TextEntry),keyed(word = l.word)), doc)[1].doc;
 END;
 
 outrec trans2(in L) := TRANSFORM
   SELF.word := L.word;
-  SELF.doc := SORTED(Files.getSearchSuperIndex()(KEYED(kind = TS.kindType.TextEntry),keyed(word = 'gobbledegook'+l.word)), doc)[1].doc;
+  SELF.doc := SORT(Files.getSearchSuperIndex()(KEYED(kind = TS.kindType.TextEntry),keyed(word = 'gobbledegook'+l.word)), doc)[1].doc;
 END;
 
 outrec trans3(in L) := TRANSFORM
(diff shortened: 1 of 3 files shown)


jakesmith:
@ghalliday - please review

testing/regress/ecl/prefetch2.ecl
```diff
@@ -40,12 +40,12 @@ END;
 
 outrec trans1(in L) := TRANSFORM
   SELF.word := L.word;
-  SELF.doc := SORTED(Files.getSearchSuperIndex()(KEYED(kind = TS.kindType.TextEntry),keyed(word = l.word)), doc)[1].doc;
+  SELF.doc := SORT(Files.getSearchSuperIndex()(KEYED(kind = TS.kindType.TextEntry),keyed(word = l.word)), doc)[1].doc;
```
ghalliday: This is reading a result from a superkey of two indexes.  The original result is showing a match from the first of the subfiles in the superkey.  I suspect Thor was hitting a match from the second index which occurred on an earlier node.
So
i) the original SORTED() is wrong
ii) Is it well defined which order matches come in from subindexes of a superindex?

@richardkchapman what is your opinion?

richardkchapman: I think subindexes are supposed to interleave - i.e. you should see matches from part 1 of subindex 1 then from part 1 of subindex 2 then from part 2 of subindex 1

jakesmith: > I suspect Thor was hitting a match from the second index which occurred on an earlier node.

Confirmed that is what's happening.
Thor sees a match in the 2nd index 1st (on an earlier node).

jakesmith: So Thor is interleaving. In this case the width the Thor is the same as the indexes.
So the 1st slave, is processing the 1st parts of the index.
The 2nd slave is processing the 2nd parts, etc.

Thor sees a match from the 2nd index first, the 1st index has matches, but they come in partitions read on later nodes, so are not part of the i[1].doc result.

Which I think is correct.

jakesmith: Btw, hthor result is same as roxie. So it looks like both read all parts of subfile1, before reading subfile2 ..

richardkchapman:
What happens if you add a SORTED,ASSERT?

If there's a SORTED on there but the data is not, in fact, sorted, then the result is undefined, I would have thought

jakesmith:
SORTED,ASSERT doesn't have any effect.
But that's because it is sorted to the point it has read up to, because of the SELECTNTH(1).

If you change the query to try to output all matches, so it reads whole index with SORTED,ASSERT,
e.g.:
```
SORTED(Files.getSearchSuperIndex()(KEYED(kind = TS.kindType.TextEntry),keyed(word = 'boy')), doc, ASSERT);
```

then all engines fail (as they should) with 'detected incorrectly sorted rows'

richardkchapman:
typo in commit message
@ghalliday

ghalliday:
@jakesmith ok, seems reasonable to merge it.  Can you add a #onwarning to suppress the difference in roxie.

jakesmith:
@ghalliday - please merge.

PR #12104, opened 2019-01-28
HPCC-21377 Specify CPP sources via manifest file

richardkchapman:
Also fixes
HPCC-21371 Add option to disallow code in manifest
HPCC-21364 Manifest support for signed code

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [x] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/ecl/hql/hqlmanifest.cpp
+++ b/ecl/hql/hqlmanifest.cpp
@@ -16,6 +16,7 @@
 ############################################################################## */
 #include "jliball.hpp"
 #include "hql.hpp"
+#include "hqlutil.hpp"
 #include "hqlmanifest.hpp"
 
 //-------------------------------------------------------------------------------------------------------------------
@@ -29,12 +30,28 @@ public:
     {
         try
         {
-            manifest.setown(createPTreeFromXMLFile(filename));
+            fileContents.loadFile(filename, false);
+            const char *xml = fileContents.str();
+            StringBuffer body;
+            // Check for signature
+            if (startsWith(fileContents, "-----BEGIN PGP SIGNED MESSAGE-----"))
+            {
+                // Note - we do not check the signature here - we are creating an archive, and typically that means we
+                // are on the client machine, while the signature can only be checked on the server where the keys are installed.
+                stripSignature(body, fileContents);
+                xml = body.str();
+                isSigned = true;
+            }
+            manifest.setown(createPTreeFromXMLString(xml));
         }
-        catch (IException * e)
+        catch (IException * E)
         {
-            e->Release();
-            throw makeStringExceptionV(0, "Invalid manifest file '%s'", filename);
+            StringBuffer msg;
+            E->errorMessage(msg);
+            auto code = E->errorCode();
+            auto aud = E->errorAudience();
+            E->Release();
+            throw makeStringExceptionV(aud, code, "While loading manifest file %s: %s", filename, msg.str());
         }
         makeAbsolutePath(filename, absFilename);
         splitDirTail(absFilename, dir);
@@ -52,8 +69,10 @@ private:
 
 public:
     Owned<IPropertyTree> manifest;
+    StringBuffer fileContents;
     StringBuffer absFilename;
     StringBuffer dir;
+    bool isSigned = false;
 };
 
 void updateResourcePaths(IPropertyTree &resource, const char *dir)
@@ -181,7 +200,7 @@ void ResourceManifest::addToArchive(IPropertyTree *archive)
 {
     IPropertyTree *additionalFiles = ensurePTree(archive, "AdditionalFiles");
 
-    //xsi namespace required for proper representaion after PTree::setPropBin()
+    //xsi namespace required for proper representation after PTree::setPropBin()
     if (!additionalFiles->hasProp("@xmlns:xsi"))
         additionalFiles->setProp("@xmlns:xsi", "http://www.w3.org/2001/XMLSchema-instance");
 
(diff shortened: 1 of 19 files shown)


richardkchapman:
@afishbeck @ghalliday For initial review - some cleanup still needed

richardkchapman:
@ghalliday I think this is now ready to look at

ecl/hql/hqlmanifest.cpp
```diff
@@ -52,8 +68,10 @@ class ResourceManifest : public CInterface
 
 public:
     Owned<IPropertyTree> manifest;
+    StringBuffer fileContents;
```
ghalliday: Should this be a local variable rather than a member?

ecl/hql/hqlmanifest.cpp
```diff
@@ -29,12 +30,27 @@ class ResourceManifest : public CInterface
...
         try
         {
-            manifest.setown(createPTreeFromXMLFile(filename));
+            fileContents.loadFile(filename, false);
+            const char *xml = fileContents.str();
+            StringBuffer body;
+            // Check for signature
+            if (startsWith(fileContents, "-----BEGIN PGP SIGNED MESSAGE-----"))
+            {
+                // OwnedHqlExpr sig = checkSignature(fileContents.length(), fileContents.str());
```
ghalliday: Shouldn't it be checking this?

ecl/hqlcpp/hqlres.cpp
```diff
@@ -40,7 +41,7 @@ class ResourceItem : public CInterface
 };
 
 
-ResourceManager::ResourceManager()
+ResourceManager::ResourceManager(IHqlCppInstance &_cppInstance) : cppInstance(_cppInstance)
```
ghalliday: I'll read some more, but hesitant about adding dependencies from child to parent.

ecl/hqlcpp/hqlres.cpp
```diff
@@ -288,10 +360,54 @@ void ResourceManager::addManifestsFromArchive(IPropertyTree *archive)
...
+                            tempDir.append(PATHSEPCHAR).append("tmp.XXXXXX"); // Note - we share same temp dir for all from this manifest
+                            if (!mkdtemp((char *) tempDir.str()))
+                                throw makeStringExceptionV(0, "Failed to create temporary directory %s (error %d)", tempDir.str(), errno);
+                            cppInstance.addTemporaryDir(tempDir.str());
+                        }
+                        StringBuffer tempFileName;
+                        tempFileName.append(tempDir).append(PATHSEPCHAR).append(item.queryProp("@filename"));
+                        if (!recursiveCreateDirectoryForFile(tempFileName))
+                            throw makeStringExceptionV(0, "Failed to create temporary file %s (error %d)", tempFileName.str(), errno);
+                        FILE *source = fopen(tempFileName.str(), "wt");
```
ghalliday: I'm not sure you need/want the "t" here.

system/jlib/jcomp.cpp
```diff
@@ -59,7 +59,8 @@
 #define BASE_ADDRESS "0x00480000"
 //#define BASE_ADDRESS        "0x10000000"
 
-static const char * CC_NAME[] =   { "\"#" PATHSEPSTR "bin" PATHSEPSTR "cl.bat\"",   "\"#" PATHSEPSTR "bin" PATHSEPSTR "g++\"" };
+static const char * CC_NAME_CPP[] =   { "\"#" PATHSEPSTR "bin" PATHSEPSTR "cl.bat\"",   "\"#" PATHSEPSTR "bin" PATHSEPSTR "g++\"" };
+static const char * CC_NAME_C[] =   { "\"#" PATHSEPSTR "bin" PATHSEPSTR "cl.bat\"",   "\"#" PATHSEPSTR "bin" PATHSEPSTR "gcc\"" };
```
ghalliday: Do you also need to check in a deploy/gcc3/bin/gcc?

ghalliday (commented):
Had one quick look through.  Will read through again later.

ecl/hql/hqlmanifest.cpp
richardkchapman: No - you don't want to check when creating an archive. That would require all users to install the keys required to validate signatures. In "local repository" mode the keys are only on the server.

ecl/hqlcpp/hqlres.cpp
richardkchapman: I think I needed it for adding sources. It's a bit messy as lists of sources (and flags) are maintained in quite a few places, but I was keeping the existing structure for that.

I can probably pass an interface down to the relevant calls instead if you think that is cleaner, but I don't think it's ever going to be especially clean without some major refactoring.

system/jlib/jcomp.cpp
richardkchapman: I believe we always rely on the system gcc/g++ rather than providing one.

ecl/hql/hqlmanifest.cpp
```diff
@@ -29,12 +30,27 @@ class ResourceManifest : public CInterface
...
+                stripSignature(body, fileContents);
+                xml = body.str();
+                isSigned = true;
+            }
+            manifest.setown(createPTreeFromXMLString(xml));
         }
-        catch (IException * e)
+        catch (IException * E)
```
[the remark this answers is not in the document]
richardkchapman: Just because that is the more common convention

ecl/hqlcpp/hqlres.cpp
richardkchapman: It's a c++ source file, which I would have thought was text. I very much doubt any compiler will care, but...

ecl/hql/hqlmanifest.cpp
richardkchapman: No, because I output the contents verbatim to the archive.

ecl/hql/hqlmanifest.cpp
richardkchapman: Should probably remove the commented-out code though (and replace with a comment about why I don't check)

ecl/hql/hqlmanifest.cpp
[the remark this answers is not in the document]
ghalliday: Just did a quick search - 313 matches for upper case E, 1750 matches for lower case (in a catch).
An upper case E doesn't match our standard of variables starting in lower case...

ghalliday (commented):
A few fairly minor comments

ghalliday:
@richardkchapman

richardkchapman:
@ghalliday Addressed your comments

PR #12105, opened 2019-01-28
HPCC-19923 Implement copy/paste support in Config Manager 2.0

kenrowland:
Add serialize method to serialize environment subtree
Add method to insert serialized subtree into an environment
Add ESP service methods to copy a node and insert a node copy

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Premature optimization

### DIFF
--- a/configuration/configmgr/configmgrlib/EnvironmentNode.cpp
+++ b/configuration/configmgr/configmgrlib/EnvironmentNode.cpp
@@ -217,66 +217,53 @@ std::string EnvironmentNode::getLocalValue() const
 }
 
 
-void EnvironmentNode::validate(Status &status, bool includeChildren, bool includeHiddenNodes) const
+void EnvironmentNode::validate(Status &status) const
 {
-    if (!m_pSchemaItem->isHidden() || includeHiddenNodes)
+    //
+    // Check node value
+    if (m_pLocalValue)
     {
-        //
-        // Check node value
-        if (m_pLocalValue)
-        {
-            m_pLocalValue->validate(status, m_id);
-        }
+        m_pLocalValue->validate(status, m_id);
+    }
+
+    //
+    // Check any attributes
+    for (auto &attrIt: m_attributes)
+    {
+        attrIt.second->validate(status, m_id);
 
         //
-        // Check any attributes
-        for (auto attrIt = m_attributes.begin(); attrIt != m_attributes.end(); ++attrIt)
+        // If this value must be unique, make sure it is
+        if (attrIt.second->getSchemaValue()->isUniqueValue())
         {
-            attrIt->second->validate(status, m_id);
-
-            //
-            // If this value must be unique, make sure it is
-            if (attrIt->second->getSchemaValue()->isUniqueValue())
+            bool found = false;
+            std::vector<std::string> allValues;
+            attrIt.second->getAllValuesForSiblings(allValues);
+            std::set<std::string> unquieValues;
+            for (auto it = allValues.begin(); it != allValues.end() && !found; ++it)
             {
-                bool found = false;
-                std::vector<std::string> allValues;
-                attrIt->second->getAllValuesForSiblings(allValues);
-                std::set<std::string> unquieValues;
-                for (auto it = allValues.begin(); it != allValues.end() && !found; ++it)
-                {
-                    auto ret = unquieValues.insert(*it);
-                    found = !ret.second;
-                }
-
-                if (found)
-                {
-                    status.addUniqueMsg(statusMsg::error, m_id, attrIt->second->getName(), "Attribute value must be unique");
-                }
+                auto ret = unquieValues.insert(*it);
+                found = !ret.second;
             }
 
-            //
-            // Does this value need to be from another set of values?
-            if (attrIt->second->getSchemaValue()->isFromUniqueValueSet())
+            if (found)
             {
-                bool found = false;
-                std::vector<std::string> allValues;
-                attrIt->second->getSchemaValue()->getAllKeyRefValues(allValues);
-                for (auto it = allValues.begin(); it != allValues.end() && !found; ++it)
-                    found = *it == attrIt->second->getValue();
-                if (!found)
-                {
-                    status.addMsg(statusMsg::error, m_id, attrIt->second->getName(), "Attribute value must be from a unique set");
-                }
+                status.addUniqueMsg(statusMsg::error, m_id, attrIt.second->getName(), "Attribute value must be unique");
             }
         }
 
         //
-        // Now check all children
-        if (includeChildren)
+        // Does this value need to be from another set of values?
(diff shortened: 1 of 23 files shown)


kenrowland:
@jpmcmu Please review. Thanks.

kenrowland:
@AttilaVamos I know not why this would fail. The file in question is a generated file from the ecm file. This builds w/o error in my dev environment. Unless there is additional logging that I cannot see, could this be something else causing it?

AttilaVamos (commented):
In  esp/services/ws_configmgr/ws_configmgrService.cpp there is an include:

```
#include "../../../../release/generated/ws_configmgr.hpp"
```
I don't think this path is correct because the ws_configmgr directory is in the third level down from the source root and this path references to the 4th level up. If it is then (I think) it means you already have a "release" directory above the source root ("HPCC-Platform") with generated header file.

The other way, why I think it is not correct because the build directory can be anywhere (same or different level from the source "HPCC-Platform" directory) and can called by any name e.g. "build" or "release" and from that build directory can have a "generated" directory.

(And you have an unresolved conflict in this commit)

kenrowland:
@AttilaVamos Yes I saw that. Not sure how that line got in there. Perhaps the tools I use thought it needed to be there. I was especially odd because it was referencing a release build directory directly. I have removed it.

AttilaVamos:
@kenrowland back to you.

kenrowland:
@richardkchapman Please merge

richardkchapman:
Has this been reviewed? @jpmcmu ?

kenrowland:
@richardkchapman sorry confused with another PR. James will review.

configuration/configmgr/configmgrlib/EnvironmentNode.cpp
```diff
@@ -217,66 +217,53 @@ std::string EnvironmentNode::getLocalValue() const
...
+        // If this value must be unique, make sure it is
+        if (attrIt.second->getSchemaValue()->isUniqueValue())
         {
-            attrIt->second->validate(status, m_id);
-
-            //
-            // If this value must be unique, make sure it is
-            if (attrIt->second->getSchemaValue()->isUniqueValue())
+            bool found = false;
+            std::vector<std::string> allValues;
```
jpmcmu: Not required but allValues an uniqueValues could be hoisted outside of the for loop and cleared to avoid memory churn a bit.

configuration/configmgr/configmgrlib/SchemaItem.cpp
```diff
@@ -646,3 +647,93 @@ void SchemaItem::setRequiredInstanceComponents(const std::string list)
...
+
+
+void SchemaItem::validate(Status &status, bool includeChildren, bool includeHiddenNodes) const
+{
+    //
+    // Check the number of environment nodes against the allowed min/max range. Note that since the schema does not
+    // maintain a parent child relationship like the environment, the environment nodes for this schema item need to
+    // be separated by each environment node's parent. Then the number of nodes, by parent, is compared agains the
+    // allowed min/max range
+    std::map<std::string, std::vector<std::shared_ptr<EnvironmentNode>>> envNodesByParent;
```
jpmcmu: It looks like this is used to track the number of children for a given parent. Could this become ``std::map<std::string,uint32_t>`` to avoid having a std::vector in the map? Granted this is probably less of a concern as vector should have move semantics

esp/services/ws_configmgr/ws_configmgrService.cpp
```diff
@@ -352,7 +355,7 @@ bool Cws_configMgrEx::onGetNode(IEspContext &context, IEspGetNodeRequest &req, I
     }
 
     getNodeResponse(pNode, resp);
-    pNode->validate(status, false);  // validate this node only
+    pNode->getSchemaItem()->validate(status, false);  // validate this node only
```
jpmcmu: Can schemaItem be a nullptr?

esp/services/ws_configmgr/ws_configmgrService.cpp
```diff
@@ -593,6 +596,103 @@ bool Cws_configMgrEx::onFetchNodes(IEspContext &context, IEspFetchNodesRequest &
 }
 
 
+bool Cws_configMgrEx::onGetNodeCopy(IEspContext &context, IEspGetNodeCopyRequest &req, IEspGetNodeCopyResponse &resp)
+{
+    std::string nodeId = req.getNodeId();
+    std::string sessionId = req.getSessionId();
+    Status status;
+
+    ConfigMgrSession *pSession = getConfigSession(sessionId, true);
```
jpmcmu: Do we need a nullptr check on pSession? It looks like sessionId is coming from the user

esp/services/ws_configmgr/ws_configmgrService.cpp
```diff
@@ -593,6 +596,103 @@ bool Cws_configMgrEx::onFetchNodes(IEspContext &context, IEspFetchNodesRequest &
...
+
+    if (nodeId.empty())
+        nodeId = pSession->m_pEnvMgr->getRootNodeId();
```
jpmcmu: Can m_pEnvMgr be nullptr?

jpmcmu (commented):
@kenrowland a couple of comments / questions

configuration/configmgr/configmgrlib/SchemaItem.cpp
kenrowland: The observation is a good one. When writing the code, there was the potential need to inspect each of the nodes by type. While that has not materialized, I would prefer to leave the construct the way it is in case that need arises again.

esp/services/ws_configmgr/ws_configmgrService.cpp
kenrowland: No. I ensure that all nodes have a schema item. This is even the case if no schema is defined. A default schema item is constructed for this case.

esp/services/ws_configmgr/ws_configmgrService.cpp
kenrowland: No. The getConfigSession method will throw an exception if the ID is not valid.

esp/services/ws_configmgr/ws_configmgrService.cpp
kenrowland: No, provided 'true' was passed as the second parameter to getConfigSession. That indicates that a loaded environment is required.

kenrowland (commented):
@jpmcmu Please see my comments. If you are OK with the responses, please approve. Thanks for the review.

jpmcmu:
@kenrowland looks good

kenrowland:
@richardkchapman please merge. Thanks.

PR #12112, opened 2019-01-29
HPCC-20334 Remove Boost regex dependency

kenrowland:
Convert to use jlib regex functions

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Premature optimization

### DIFF
--- a/configuration/configmgr/configmgrlib/SchemaTypeStringLimits.cpp
+++ b/configuration/configmgr/configmgrlib/SchemaTypeStringLimits.cpp
@@ -17,11 +17,10 @@
 
 #include "SchemaTypeStringLimits.hpp"
 #include "EnvironmentValue.hpp"
-#include <boost/regex.hpp>
+#include "jregexp.hpp"
 
 std::string SchemaTypeStringLimits::getLimitString() const
 {
-
     return "String limit info";
 }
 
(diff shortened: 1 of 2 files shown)


kenrowland:
@timothyklemm Please review

configuration/configmgr/configmgrlib/SchemaTypeStringLimits.cpp
```diff
@@ -17,26 +17,25 @@
...
 {
     bool isValid;
-    int len = testValue.length();
+    size_t len = testValue.length();
     isValid = len >= m_minLength && len <= m_maxLength;
 
-    // test patterns
-    for (auto it = m_patterns.begin(); it != m_patterns.end() && isValid; ++it)
+    // test patterns (if any, there will likely be only one)
+    for (auto &pattern: m_patterns)
```
timothyklemm: As much as I like the new loop syntax, I would prefer to keep the explicit check for isValid. The loop can be avoided entirely if the length test fails, and the overhead of creating and destroying unneeded RegExpr instances after a failure is unnecessary. I see the comment stating that the most likely scenario will be at most one pattern, but if it's necessary to loop then we should get out of the loop as soon as the outcome is known.

kenrowland: Returned to old format. @timothyklemm

timothyklemm (approved):

PR #12115, opened 2019-01-30
HPCC-21390 Remove default implementation of virtual functions from eclhelper.hpp

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/common/thorhelper/thorcommon.ipp
+++ b/common/thorhelper/thorcommon.ipp
@@ -129,6 +129,7 @@ public:
     virtual void toXML(const byte * self, IXmlWriter & out) { }
     virtual unsigned getVersion() const                     { return OUTPUTMETADATA_VERSION; }
     virtual unsigned getMetaFlags()                         { return std::is_pod<T>() ? 0 : MDFneeddestruct; }
+    virtual const RtlTypeInfo * queryTypeInfo() const       { return nullptr; }
     virtual void destruct(byte * self)                      { reinterpret_cast<T *>(self)->~T(); }
     virtual IOutputRowSerializer * createDiskSerializer(ICodeContext * ctx, unsigned activityId) { return NULL; }
     virtual IOutputRowDeserializer * createDiskDeserializer(ICodeContext * ctx, unsigned activityId) { return NULL; }
@@ -136,6 +137,7 @@ public:
     virtual IOutputMetaData * querySerializedDiskMeta() { return this; }
     virtual IOutputRowSerializer * createInternalSerializer(ICodeContext * ctx, unsigned activityId) { return NULL; }
     virtual IOutputRowDeserializer * createInternalDeserializer(ICodeContext * ctx, unsigned activityId) { return NULL; }
+    virtual void process(const byte * self, IFieldProcessor & target, unsigned from, unsigned to) {}
     virtual void walkIndirectMembers(const byte * self, IIndirectMemberVisitor & visitor) {}
     virtual IOutputMetaData * queryChildMeta(unsigned i) { return NULL; }
     virtual const RtlRecord &queryRecordAccessor(bool expand) const { throwUnexpected(); } // could provide a static implementation if needed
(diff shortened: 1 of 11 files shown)


ghalliday:
Tested all examples in regression suite and ecl/regress to ensure they still compile.  (And found 1 problem in the process).
@shamser please review

richardkchapman:
Curious how you chose which to move (doesn't look like you moved all).

Do any of these changes cause existing compiled queries to need to be recompiled?

ghalliday:
I don't think it will require any queries to be recompiled, but not 100% sure.  I think they should previously be implemented in the derived classes.  (With the hidden symbols enabled we didn't get link errors from queries.)

I tried to move them all.  I think the only ones remaining were virtual destructors and a couple in a class that wasn't an activity.  My main concern was to remove things like canMatch() which are potentially called for each row.  Any that are only called once per query will have little effect.

ghalliday:
note, I didn't move inline functions - so that may be what you were seeing.

richardkchapman:
It was ones in other include files (not eclhelper.hpp) where some had moved and some had not

common/thorhelper/thorcommon.ipp
```diff
@@ -381,6 +384,10 @@ public:
     {
         return new CPrefixedRowDeserializer(offset, original->createInternalDeserializer(ctx, activityId));
     }
+    virtual void process(const byte * self, IFieldProcessor & target, unsigned from, unsigned to)
+    {
```
shamser: @ghalliday This definition for process()  seems to be different from the previous base class definition for process().  If this ticket simply removes default implementation of virtual functions, then I would have thought that this definition should reflect base class definition?

rtl/eclrtl/eclhelper_base.cpp
```diff
@@ -526,9 +536,39 @@ const char * CThorCsvWriteArg::getCluster(unsigned idx) { return NULL; }
...
+
+unsigned CThorFetchArg::getFetchFlags() { return 0; }
+unsigned CThorFetchArg::getDiskFormatCrc() { return 0; }
+unsigned CThorFetchArg::getProjectedFormatCrc() { return getDiskFormatCrc(); }  // Should really be crc of queryProjectedDiskRecordSize layout
+void CThorFetchArg::getFileEncryptKey(size32_t & keyLen, void * & key) { keyLen = 0; key = 0; }
+unsigned __int64 CThorFetchArg::getRowLimit() { return (unsigned __int64) -1; }
+void CThorFetchArg::onLimitExceeded()         { }
+size32_t CThorFetchArg::extractJoinFields(ARowBuilder & rowBuilder, const void * _right) { return 0; }
+bool CThorFetchArg::extractAllJoinFields()    { return false; }
+IOutputMetaData * CThorFetchArg::queryExtractedSize() { return NULL; }
```
shamser: @ghalliday should nullptr be returned (instead of NULL) here and below

shamser (commented):
@ghalliday

common/thorhelper/thorcommon.ipp
ghalliday: You're right, this is a bug fix added at the same time.  I will move them to a different PR.

rtl/eclrtl/eclhelper_base.cpp
ghalliday: Yes, but this was copied code, so fine to stay as NULL.

ghalliday:
@shamser see replies

shamser (approved):
@richardkchapman Code reviewed.

richardkchapman:
@ghalliday Did you determine whether this would cause any queries to need to be recompiled?

ghalliday:
@richardkchapman I have verfied that if I build workunits with the previous code they still load in a system with the changes, and also none of the functions are exported from the dlls.  So I am fairly confident it will not cause queries to need to be recompiled.

PR #12125, opened 2019-01-31
HPCC-21403 Wrap Std.Date.FromStringToSeconds() in a compile-time version check

dcamper:
Trailing whitespace in a few other places has also been removed.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Testing:

Manual tests with code supplied in the associated Jira and other code.

### DIFF
--- a/ecllibrary/std/Date.ecl
+++ b/ecllibrary/std/Date.ecl
@@ -6,9 +6,9 @@
 
 IMPORT lib_stringlib.StringLib;
 IMPORT lib_timelib.TimeLib;
+IMPORT Std.System.Util;
 
 EXPORT Date := MODULE
-
 
 // A record structure with the different date elements separated out.
 EXPORT Date_rec := RECORD
@@ -687,8 +687,11 @@ EXPORT Time_t FromStringToTime(STRING time_text, VARSTRING format) :=
     StringLib.StringToTimeOfDay(time_text, format);
 
 
+#IF(Util.PlatformVersionCheck('7.2.0'))
 /**
  * Converts a string to a Seconds_t using the relevant string format.
+ *
+ * Note: This function first appeared in HPCC Systems v7.2.0.
  *
  * @param datetime_text The string to be converted.
  * @param format        The format of the input string.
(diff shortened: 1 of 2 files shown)


richardkchapman:
Couple of general comments:

1. The PlatformVersionCheck function should probably live in a different Std module (not sure which - maybe Std.System.Util)
2. Feels like this may just swap one obscure error message for another. The #if will be compiled differently at syntax-check/archive creation time from workunit creation time, and is still going to come out as 'Unknown identifier' rather than 'Server version too old'. I don't know if it's possible to put in a #else that will give a more helpful message? 

@ghalliday Any thoughts?

dcamper:
I like the idea of moving PlatformVersionCheck to Std.System.Util, and marking it as EXPORT.  There have been some similar recent problems that are causing our internal dev teams to explicitly reference a particular client tools version to work around the issue.  I suspect that using PlatformVersionCheck would solve that as well.  I will move the function.

I think the problem with "obscure message" is really just a documentation issue.  If Std.Date. FromStringToSeconds() is documented as first appearing in 7.2.0 -- which it is not as of this writing -- then the developer should be aware that both client tools and the cluster must both be at 7.2.0 or later.  With the documentation in place, the error that would appear ("FromStringToSeconds cannot be found in Std.Date" or whatever it is) will make sense at both syntax check and WU creation.

Without wrapping this function in a version check, the developer can get tripped up by this error when trying to use any Std.Date function in a mismatched client tools/cluster setup.  The error in that case is very confusing, as it references something that the developer wasn't even trying to use.  This scenario can occur easily:  The cluster is pre-7.2.0 and the developer upgrades their client tools.

Interested in hearing alternate points of view.

richardkchapman:
Worth thinking about the different cases:

Adding a new function 
Adding a new parameter
Removing a function
Using new ECL functionality

Both for plugins and for the standard ECL library. Isn't there some rule saying we use the remove version of the plugin/std library ECL even though the local version has been embedded in the archive?

ghalliday:
I have created a jira to document the new function in Std.Util

ghalliday:
Yes - that means that code with old standard libraries will still work on new systems.  This is different - it allows new functions to work on old systems.  I think to work it would mean that any changes have to be wrapped in version checks.
I merged the change, but on reflection I'm not sure we want to support backward compatibility this way in general.

PR #12139, opened 2019-02-04
Revert "HPCC-21403 Wrap Std.Date.FromStringToSeconds() in a compile-time version check"

ghalliday:
Reverts hpcc-systems/HPCC-Platform#12125

On reflection this isn't the correct fix, and it wouldn't be possible to use this approach in general.  Better is to use the standard library from the remote system as well as the plugins.


ghalliday:
@richardkchapman please merge

PR #12147, opened 2019-02-06
HPCC-21424 Add envchk CLI to validate environment file

kenrowland:
Add new CLI to run config manager core library validate function
and print results to stdout
Improve some XSDs to better detect invalid configurations
Improve config manager core library validation to remove
duplicate errors and improve error messages

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Premature optimization


kenrowland:
@jpmcmu Please review. Thanks

configuration/cli/envchk/envchk.cpp
```diff
@@ -0,0 +1,341 @@
...
+    size_t  start = 0, end = 0, delimLen = delim.length();
+    std::vector<std::string> list;
+
+    while (end != std::string::npos)
+    {
+        end = input.find(delim, start);
+        std::string item = input.substr(start, (end == std::string::npos) ? std::string::npos : end - start);
+        if (!item.empty())
+            list.push_back(item);
+        start = ((end > (std::string::npos - delimLen)) ? std::string::npos : end + delimLen);
```
jpmcmu: This seems incorrect. ``end > (std::string::npos - delimLen)`` I think this would always be true. End is always >= -1 and (-1 - delimLen) should always be less than -1

configuration/configmgr/configmgrlib/EnvironmentNode.cpp
```diff
@@ -234,23 +234,23 @@ void EnvironmentNode::validate(Status &status) const
...
+//            {
+//                auto ret = unquieValues.insert(*it);
+//                found = !ret.second;
+//            }
+//
+//            if (found)
+//            {
+//                status.addUniqueMsg(statusMsg::error, m_id, attrIt.second->getName(), "Attribute value must be unique");
+//            }
+//        }
```
jpmcmu: Do we need to keep the commented out code around?

configuration/configmgr/configmgrlib/SchemaItem.cpp
```diff
@@ -652,88 +660,115 @@ void SchemaItem::setRequiredInstanceComponents(const std::string list)
...
 
-    //
-    // Now validate each environment node (at this time go ahead and reset the member variable of weak pointers as well
-    // to get rid of any stale values)
-    for (auto &pWeakEnvNode: m_envNodes)
-    {
-        if (!isHidden() || includeHiddenNodes)
+        //
+        // Check the attributes for uniqueness.
+        for (auto &attribute: m_attributes)
```
jpmcmu: I don't know how many attributes we are searching here, but would we ever run into the case where the n^2 search here might get out of hand? Sorting  & using something like http://www.cplusplus.com/reference/algorithm/set_intersection/  might also be less code.

initfiles/componentfiles/configschema/xsd/types.xsd
```diff
@@ -47,20 +47,21 @@
     <xs:simpleType name="cronSchedule">
         <xs:restriction base="xs:string">
             <xs:pattern
-                    value="^(\*|((\*\/)?[1-5]?[0-9])) (\*|((\*\/)?[1-5]?[0-9])) (\*|((\*\/)?(1?[0-9]|2[0-3]))) (\*|((\*\/)?([1-9]|[12][0-9]|3[0-1]))) (\*|((\*\/)?([1-9]|1[0-2]))) (\*|((\*\/)?[0-6]))$"/>
+                    value="^(\*|([0-9]|1[0-9]|2[0-9]|3[0-9]|4[0-9]|5[0-9])|\*\/([0-9]|1[0-9]|2[0-9]|3[0-9]|4[0-9]|5[0-9])) (\*|([0-9]|1[0-9]|2[0-3])|\*\/([0-9]|1[0-9]|2[0-3])) (\*|([1-9]|1[0-9]|2[0-9]|3[0-1])|\*\/([1-9]|1[0-9]|2[0-9]|3[0-1])) (\*|([1-9]|1[0-2])|\*\/([1-9]|1[0-2])) (\*|([0-6])|\*\/([0-6]))$"/>
```
jpmcmu: May not be important, but it doesn't seem like the above regex would allow for lists or ranges in the cron schedule. Checked regex lib, there doesn't seem to be a simple expression that covers everything. This covers ranges and lists but is more than a little complicated: http://regexlib.com/REDetails.aspx?regexp_id=20201

jpmcmu (commented):
@kenrowland  a few comments / questions

configuration/cli/envchk/envchk.cpp
kenrowland: It would seem this is the case, however the code will successfully parse and split the string. The size_t is unsigned, so the test really looks at end being the case where the input.find did not find a match. In this case it just assigns start so that the next find will return npos. However, since end is already npos, it will exit the loop anyway. This could be cleaner. Probably as simple check using the length of the input.

configuration/configmgr/configmgrlib/SchemaItem.cpp
kenrowland: I don't think this is an issue, however that could be an improvement. I shall do some research.

initfiles/componentfiles/configschema/xsd/types.xsd
kenrowland: I did some searching for a regex to validate a cron schedule. I think some of the issues I had as I tested was that the jlib regex wasn't working, hence some expressions I tested may be been OK. I was looking for something that gave reasonable results. That one you mentioned, and by the way thanks for the reference to that site, is more complicated and probably beyond the scope of the simple schedules we want to validate.

kenrowland (commented):
@jpmcmu  Please see my responses. I will make an additional commit and let you know when it is done.

kenrowland:
@jpmcmu I made another commit. Please take a look.

jpmcmu (commented):
@kenrowland looks good

kenrowland:

PR #12158, opened 2019-02-10
HPCC-21469 Prevent splitter stop state getting out of sync

jakesmith:
If a downstream activity called stop on the same input >1, then
the stop count/state could get out of sync, causing the input
to stop prematurely, and/or the input to be restarted and then
not stopped on the next iteration of the subgraph if within a
child graph.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/thorlcr/activities/nsplitter/thnsplitterslave.cpp
+++ b/thorlcr/activities/nsplitter/thnsplitterslave.cpp
@@ -346,7 +346,6 @@ public:
             {
                 writer.stop();
                 PARENT::stop();
-                inputPrepared = false;
             }
         }
     }
@@ -462,6 +461,8 @@ void CSplitterOutput::start()
 // IEngineRowStream
 void CSplitterOutput::stop()
 { 
+    if (stopped)
+        return;
     stopped = true;
     activity.inputStopped(outIdx);
     dataLinkStop();
--- a/thorlcr/graph/thgraphslave.hpp
+++ b/thorlcr/graph/thgraphslave.hpp
@@ -65,11 +65,13 @@ public:
 #ifdef _TESTING
         owner.ActPrintLog("ITDL starting for output %d", outputId);
 #endif
-#ifdef _TESTING
-        bool started = hasStarted();
-        bool stopped = hasStopped();
-        assertex(!hasStarted() || hasStopped());      // ITDL started twice
-#endif
+        if (hasStarted())
+        {
+            if (!hasStopped())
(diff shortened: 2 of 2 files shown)


thorlcr/activities/nsplitter/thnsplitterslave.cpp
```diff
@@ -346,7 +346,6 @@ class NSplitterSlaveActivity : public CSlaveActivity, implements ISharedSmartBuf
             {
                 writer.stop();
                 PARENT::stop();
-                inputPrepared = false;
```
jakesmith: NB: this is reset in reset() - it was premature to reset here.

thorlcr/graph/thgraphslave.hpp
```diff
@@ -65,11 +65,13 @@ class CEdgeProgress
...
-        bool stopped = hasStopped();
-        assertex(!hasStarted() || hasStopped());      // ITDL started twice
-#endif
+        if (hasStarted())
+        {
+            if (!hasStopped())
+                throw MakeActivityException(&owner, 0, "Starting without being stopped 1st");
+            else
+                throw MakeActivityException(&owner, 0, "Started and stopped states both set");
+        }
```
jakesmith: Not directly related to the bug, but would have made the error hit/being debugged easier to diagnose.

jakesmith:
I've targeted 7.0.12 - because it is a small/safe fix, but given there's a workaround we can move it to 7.2 if required.

@ghalliday - please review/comment on target.

thorlcr/graph/thgraphslave.hpp
ghalliday: I agree the new code is clearer.   Does this change anything (other than improving the error message?)  I don't think started and stopped can ever be set at the same time.  (If so how can they?)

jakesmith: it doesn't change anything, purely to make error clearer.
I didn't think they could both be set at the same time, but was reluctant to remove that check in case could (e.g. corrupted)

ghalliday:
@richardkchapman looks ready (and safe) to merge.  Not sure which version you want to merge it into though.

richardkchapman:
@jakesmith Can you rebase onto 7.0.x please

jakesmith:
@richardkchapman - rebased

PR #12160, opened 2019-02-11
HPCC-21463 Refactor CThorContiguousRowBuffer

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/rtl/eclrtl/rtlcommon.hpp
+++ b/rtl/eclrtl/rtlcommon.hpp
@@ -7,15 +7,76 @@
 #include "eclrtl.hpp"
 #include "eclhelper.hpp"
 
-//The CThorContiguousRowBuffer is the source for a readAhead call to ensure the entire row
-//is in a contiguous block of memory.  The read() and skip() functions must be implemented
-class ECLRTL_API CThorContiguousRowBuffer : implements IRowPrefetcherSource
+//The CContiguousRowBuffer is a buffer used for reading ahead into a file, and unsuring there is a contiguous
+//block of data available to the reader.  Fixed size files could use this directly.
+class ECLRTL_API CContiguousRowBuffer
 {
 public:
-    CThorContiguousRowBuffer() {};
+    CContiguousRowBuffer() = default;
+    CContiguousRowBuffer(ISerialStream * _in);
+
+    void setStream(ISerialStream *_in);
+    const byte * peekBytes(size32_t maxSize);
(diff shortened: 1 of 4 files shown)


ghalliday:
NOT ready for review.  Pushed for smoketest results.

ghalliday:
@jakesmith please can you review this.  I think it is complete, but I am hesitant about the scope for breaking things.  For instance there was a problem in roxie which was caught by only one case - on a 4 way data file.

rtl/eclrtl/rtlcommon.hpp
```diff
@@ -7,15 +7,72 @@
...
-class ECLRTL_API CThorContiguousRowBuffer : implements IRowPrefetcherSource
+class ECLRTL_API CThorContiguousRowBuffer : public CContiguousRowBuffer, implements IRowPrefetcherSource
 {
 public:
-    CThorContiguousRowBuffer() {};
+    CThorContiguousRowBuffer() = default;
     CThorContiguousRowBuffer(ISerialStream * _in);
 
-    inline void setStream(ISerialStream *_in) { in = _in; maxOffset = 0; readOffset = 0; }
+    inline void setStream(ISerialStream *_in) { CContiguousRowBuffer::setStream(_in); }
```
jakesmith: why override?

jakesmith:
@ghalliday - I've looked through it a couple of times and can't spot any issues, and it looks clearer/cleaner.

What was the roxie issue caught ?

ghalliday:
The problem in roxie was that it was calling inputSteam->tell() instead of deserizlizerSource.tell().  That caused problems because it no longer skips the input stream at the start of each row.  I also had a similar problem with calling inputStream->eos() which I worked around.  I plan to restructure the roxie code further anyway.

rtl/eclrtl/rtlcommon.hpp
ghalliday: I'll check.  It should possibly be clearing readOffset as well.

ghalliday: Fixed that problem.

ghalliday:
@richardkchapman please can you review this.  I think it will be better to merge before starting the next stage.

PR #12186, opened 2019-02-15
HPCC-20996 Stream write support and services

jakesmith:
+ add streaing write support to dafilesrv
+ add a dafilesrv generic json handler, so JSON command can 
proceed data (needed for write)
+ restructure dafilesrv service so activity deals with processing
+ New DFU file interfaces and streaming interface implementations,
that connect directly to Esp and provide partition stream read
and write services, plus create and publish services.
+ Add standalone unittests for streaming interface interfaction
with dafilersv
+ Add datest test cases for read, write, copy, publish
+ Add getAccessibleServiceURLList service to environment to
discover WsDfu services (similar to fileservices and needs
additional commoning up and clarification)
+ Expose removePhysicalPartFiles to use on file before
publish
+ Add new WsDfu services to handle create without locations etc.
    
## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization


jakesmith:
There a few minor issues to clear up, but near ready, good enough for review.

@jpmcmu - please review/test.

AttilaVamos:
@AttilaVamos Those spray errors has been caused by my merged PR-https://github.com/hpcc-systems/HPCC-Platform/pull/12206 . I'm working on the fix. I am sorry about inconvenience caused.

AttilaVamos:
@jakesmith back to you.

jakesmith:
This should have been back to @jpmcmu

common/remote/CMakeLists.txt
```diff
@@ -51,16 +52,22 @@ if (USE_URIPARSER)
...
+         ${HPCC_SOURCE_DIR}/system/security/shared
+         ${HPCC_SOURCE_DIR}/system/security/securesocket
+         ${HPCC_SOURCE_DIR}/system/security/cryptohelper
+         ${HPCC_SOURCE_DIR}/testing/unittests
+         ${HPCC_SOURCE_DIR}/rtl/include
+         ${HPCC_SOURCE_DIR}/rtl/eclrtl
+         ${HPCC_SOURCE_DIR}/ecl/hql
+         ${HPCC_SOURCE_DIR}/common/deftype
+         ${HPCC_SOURCE_DIR}/dali/base
+         ${CMAKE_BINARY_DIR}
```
Michael-Gardner: @jakesmith since hql has been added to this project, to build correctly under LN we need to add ${CMAKE_BINARY_DIR}/oss here after the ${CMAKE_BINARY_DIR} entry.

Michael-Gardner: @jakesmith since hql has been added to this project, to build correctly under LN we need to add ${CMAKE_BINARY_DIR}/oss here after the ${CMAKE_BINARY_DIR} entry.

Michael-Gardner (changes requested):
Requesting a change at line 71.  Needs to add ${CMAKE_BINARY_DIR}/oss to compile properly when building against LN.

esp/services/ws_dfu/ws_dfuService.cpp
```diff
@@ -6245,117 +6180,130 @@ void CWsDfuEx::exportRecordDefinitionBinaryType(const char *recordDefinition, Me
...
+        {
+            getClusterGroupName(groupName, clusterName);
+            if (isEmptyString(groupName))
+                 throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "Group for cluster %s not found.", clusterName);
+            group.setown(queryNamedGroupStore().lookup(groupName.str()));
+            if (!group)
+                throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "Failed to get Group %s.", groupName.str());
+
+            tempFileName.append(".").append(dfuCreateUniqId++); // avoid potential clash if >1 creating file. One will succeed at publish time.
+            tempFileName.append(DFUFileCreate_FileNamePostfix);
```
jpmcmu: We need to setFileId here as well.

esp/services/ws_dfu/ws_dfuService.cpp
```diff
@@ -6370,79 +6318,129 @@ bool CWsDfuEx::onDFUFilePublish(IEspContext &context, IEspDFUFilePublishRequest
     {
```
jpmcmu: Need to clean up temp files on failure here.

esp/scm/ws_dfu.ecm
```diff
@@ -871,44 +879,72 @@ ESPStruct DFUFilePart
...
-    string MetaInfoBlob;
-    string ExpiryTime;
-    // {NumParts, FileParts} depend on ReturnFileInfo in request
-    int NumParts;        // number of parts in logical file
-    ESParray<EspStruct DFUPartLocation> FileLocations;
-    ESParray<EspStruct DFUFilePart> FileParts;
+    string MetaInfoBlob;                               // Contains serialized from of all meta data required + signature if security enabled
+    string ExpiryTime;                                 // After this time, the meta information cannot be used to communicate with dafilesrv's 
+    int NumParts;                                      // number of parts in logical file
+    ESParray<EspStruct DFUPartLocation> FileLocations; // List of hosts of parts
```
jpmcmu: Can the row service port change? If so could we add the row service port number to DFUPartLocation?

esp/services/ws_dfu/ws_dfuService.cpp
```diff
@@ -6245,117 +6180,130 @@ void CWsDfuEx::exportRecordDefinitionBinaryType(const char *recordDefinition, Me
...
-}
-
 bool CWsDfuEx::onDFUFileCreate(IEspContext &context, IEspDFUFileCreateRequest &req, IEspDFUFileCreateResponse &resp)
 {
     try
     {
-        IConstDFUFileAccessRequestBase &requestBase = req.getRequestBase();
-        SecAccessFlags accessType = translateToSecAccessFlags(requestBase.getAccessType());
-        if (SecAccess_None == accessType)
+        context.ensureFeatureAccess(FEATURE_URL, SecAccess_Write, ECLWATCH_DFU_ACCESS_DENIED, "WsDfu::DFUFileCreate: Permission denied.");
```
jpmcmu: Can we check if the requested file exists and is locked here?

common/remote/sockfile.cpp
```diff
@@ -4576,22 +5031,52 @@ void verifyMetaInfo(IPropertyTree &actNode, bool authorizedOnly, const IProperty
 #endif
         metaInfo.set(metaInfoEnvelope);
 
-    IPropertyTree *fileInfo = metaInfo->queryPropTree("FileInfo");
-    assertex(fileInfo);
-
-    // extra filename based on part/copy.
     assertex(actNode.hasProp("filePart"));
     unsigned partNum = actNode.getPropInt("filePart");
```
jpmcmu: This code seems to be executed for both read & write. However, the write JSON format template doesn't have filePart as a requirement. Do we need to specify filePart for write?

jpmcmu (changes requested):
@jakesmith a few comments / questions.

common/remote/CMakeLists.txt
jakesmith: Right, will add.

esp/scm/ws_dfu.ecm
jakesmith: It can't vary for any given single request.
So the single 'fileAccessPort' field in this response should suffice.

common/remote/sockfile.cpp
jakesmith: Yes, needs to be present whether reading or writing.
I'll update the examples + add some notes.

esp/services/ws_dfu/ws_dfuService.cpp
jakesmith: Without locking support from outside of the environment (future work), e.g. the spark client, there won't be any guarantees. i.e. the file might be created after any checking here.
I think the best we can do for now, is complain at publish time (if overwrite=false).

esp/services/ws_dfu/ws_dfuService.cpp
jakesmith: Yes, will add code to clearup in event of failure

jakesmith:
@rpastrana @jpmcmu - please review/test. Keeping a separate commits to aide review process for now.

jpmcmu:
@jakesmith the onDFUFilePublish method needs to be aware of the file compression used while writing

jakesmith:
> @jakesmith the onDFUFilePublish method needs to be aware of the file compression used while writing

it should be an option on create, but it won't be on the legacy methods, which will remain untouched.
So the new ability to indicate compressed will be on onDFUCreateFile, and will probably look like:
```
ESPrequest [nil_remove] DFUCreateFileRequest
{
    string Name;                        // the logical file name
    string Cluster;                     // needed only if on >1 Cluster and want to disambiguate
    ESPenum DFUFileType Type;           // e.g. flat, csv, xml
    string ECLRecordDefinition;                           // ecl text record definition
    string RequestId;                   // optional 
    int ExpirySeconds(60);              // how long the retrieved meta info is valid for, to access physical files (in dafilesrv) 
    bool ReturnTextResponse(false);     // Returns response in full not just in blob
    **bool Compressed(false);             // The new file will be compressed on disk**
};
```

esp/clients/wsdfuaccess/wsdfuaccess.cpp
```diff
@@ -15,42 +15,72 @@
...
-    dfuClient->addServiceUrl(serviceUrl);
-    dfuClient->setUsernameToken(user, token, "");
+    if (0 == dfuServiceUrls.size())
+        return nullptr;
+    if (currentDfuServiceUrl == dfuServiceUrls.size())
+        currentDfuServiceUrl = 0;
+    return dfuServiceUrls[currentDfuServiceUrl].c_str();
+}
 
+static IClientDFUFileAccessResponse *doLookupDFUFileDeprecated(IClientWsDfu *dfuClient, const char *logicalName, const char *requestId, unsigned expirySecs)
```
rpastrana: doLookupDFUFileLegacy or doLookupDFUFile1_39?

esp/clients/wsdfuaccess/wsdfuaccess.cpp
```diff
@@ -59,21 +89,667 @@ bool getFileAccess(StringBuffer &metaInfo, const char *serviceUrl, const char *j
...
+    try
+    {
+        dfuResp.setown(doLookupDFUFile(dfuClient, logicalName, requestId, expirySecs));
+    }
+    catch (IException *e)
+    {
+        /* NB: there should really be a different IException class and a specific error code
+         * The server knows it's an unsupported method.
+         */
+        if (SOAP_SERVER_ERROR != e->errorCode())
```
rpastrana: It might be less expensive to explicitly check the target HPCC version to know which wsdfu to target.

esp/clients/wsdfuaccess/wsdfuaccess.cpp
```diff
@@ -59,21 +89,667 @@ bool getFileAccess(StringBuffer &metaInfo, const char *serviceUrl, const char *j
...
+            throw;
+        // fall through and try deprecated method
+        e->Release();
```
rpastrana: could we potentially already know the appropriate target wsdfu interface since this requires the file access ?

rpastrana (changes requested):
@jakesmith a few comments. Will continue to review

esp/services/ws_dfu/ws_dfuService.cpp
```diff
@@ -6153,32 +6022,86 @@ SecAccessFlags translateToSecAccessFlags(CSecAccessType from)
...
+        const char *clusterGroup = df->getClusterName(c, clusterName.clear());
+        groups.push_back(clusterGroup);
+    }
+
+    StringBuffer keyPairName;
+    unsigned port;
+    bool secure;
+    getFileDafilesrvConfiguration(keyPairName, port, secure, fileName, groups);
+
+    Owned<IPropertyTree> metaInfo = createDFUFileMetaInfo(fileName, fileDesc, requestId, "READ", expirySecs, userDesc, keyPairName, port, secure, maxFileAccessExpirySeconds);
```
rpastrana: Is it worthwhile to create an enum for the accesstype (I thought one might already exist to be honest) to avoid stringwise operations on the "READ" and/or "WRITE" ?

esp/services/ws_dfu/ws_dfuService.cpp
```diff
@@ -6153,32 +6022,86 @@ SecAccessFlags translateToSecAccessFlags(CSecAccessType from)
...
+    StringBuffer metaInfoBlob;
+    encodeDFUFileMeta(metaInfoBlob, metaInfo, env);
+    accessInfo.setMetaInfoBlob(metaInfoBlob);
+
+    if (returnTextResponse)
+    {
+        getFileAccessInfo(context, *fileDesc, true, accessInfo);
+
+        accessInfo.setExpiryTime(metaInfo->queryProp("keyPairName"));
```
rpastrana: something seems off here, what's the correlation between the prop keyPairName and the expirytime?

esp/services/ws_dfu/ws_dfuService.cpp
```diff
@@ -6153,32 +6022,86 @@ SecAccessFlags translateToSecAccessFlags(CSecAccessType from)
...
+        accessInfo.setFileAccessPort(metaInfo->getPropInt("port"));
+        accessInfo.setFileAccessSSL(metaInfo->getPropBool("secure"));
+    }
+
+    StringBuffer userName;
```
rpastrana: is this different than userID from above?

esp/services/ws_dfu/ws_dfuService.cpp
```diff
@@ -6187,7 +6110,28 @@ bool CWsDfuEx::onDFUFileAccess(IEspContext &context, IEspDFUFileAccessRequest &r
     return true;
 }
 
-IGroup *CWsDfuEx::getDFUFileIGroup(const char *clusterName, ClusterType clusterType, const char *clusterTypeEx, StringArray &locations, StringBuffer &groupName)
+bool CWsDfuEx::onDFULookupFile(IEspContext &context, IEspDFULookupFileRequest &req, IEspDFUFileAccessResponse &resp)
+{
+    try
+    {
+        context.ensureFeatureAccess(FEATURE_URL, SecAccess_Read, ECLWATCH_DFU_ACCESS_DENIED, "WsDfu::DFULookupFile: Permission denied.");
```
rpastrana: FYI the annotation in the method definition ""DfuAccess:READ":
ESPmethod [auth_feature("DfuAccess:READ")] DFULookupFile(DFULookupFileRequest, DFUFileAccessResponse);
generates this same context.ensureFeatureAccess call lower in the stack. We can either change the method declaration to:
auth_feature("DEFERRED"), //This declares that the method logic handles feature level authorization
or remove this block from the CWsDfuEx::onXYZ methods

esp/services/ws_dfu/ws_dfuService.cpp
```diff
@@ -6268,119 +6212,194 @@ void CWsDfuEx::exportRecordDefinitionBinaryType(const char *recordDefinition, Me
...
     {
-        IConstDFUFileAccessRequestBase &requestBase = req.getRequestBase();
-        SecAccessFlags accessType = translateToSecAccessFlags(requestBase.getAccessType());
-        if (SecAccess_None == accessType)
-        {
-            context.setAuthStatus(AUTH_STATUS_NOACCESS);
-            throw MakeStringException(ECLWATCH_DFU_ACCESS_DENIED, "WsDfu::DFUFileCreate - Permission denied.");
-        }
-        context.ensureFeatureAccess(FEATURE_URL, accessType, ECLWATCH_DFU_ACCESS_DENIED, "WsDfu::DFUFileCreate: Permission denied.");
+        context.ensureFeatureAccess(FEATURE_URL, SecAccess_Write, ECLWATCH_DFU_ACCESS_DENIED, "WsDfu::DFUFileCreate: Permission denied.");
```
rpastrana: See above comment on ensureFeatureAccess

esp/services/ws_dfu/ws_dfuService.cpp
```diff
@@ -6268,119 +6212,194 @@ void CWsDfuEx::exportRecordDefinitionBinaryType(const char *recordDefinition, Me
...
+        if (requestId.isEmpty())
+            requestId.appendf("Create %s on %s", normalizedFileName.str(), clusterName);
 
-        Owned<IFileDescriptor> fileDesc = createFileDescriptor(tempFileName, clusterTypeEx, groupName, group);
-        fileDesc->queryProperties().setProp("@job", jobId);
+        Owned<IFileDescriptor> fileDesc = createFileDescriptor(tempFileName, clusterTypeString(clusterType, false), groupName, group);
+        fileDesc->queryProperties().setProp("@job", requestId);
         if (!userId.isEmpty())
             fileDesc->queryProperties().setProp("@owner", userId);
+        fileDesc->queryProperties().setProp("ECL", recordDefinition);
```
rpastrana: perhaps out of scope, but is it worthwhile to normalize the recordDef before setting it as the ECL prop? From the JAPI perspective, the ECL prop is important to determine the file structure, but we end up having to support multiple styles of record definitions in ECL. Just a thought

esp/services/ws_dfu/ws_dfuService.cpp
```diff
@@ -6268,119 +6212,194 @@ void CWsDfuEx::exportRecordDefinitionBinaryType(const char *recordDefinition, Me
...
-        file->queryAttributes().setPropInt64("@recordCount", 0);
-        file->queryAttributes().setPropInt64("@size", 0);
+        MemoryBuffer layoutBin;
+        exportRecordDefinitionBinaryType(recordDefinition, layoutBin);
+        if (0 == layoutBin.length())
+            throw makeStringExceptionV(ECLWATCH_INVALID_ECLRECDEF, "DFUCreateFile: Failed to parse ECL record definition: %s.", recordDefinition);
+        fileDesc->queryProperties().setPropBin("_rtlType", layoutBin.length(), layoutBin.toByteArray());
 
-        getFileAccessBeforePublish(context, tempFileName, clusterName, jobId, layoutBin, fileDesc, userDesc, requestBase, resp);
+        // NB: if file has copies on >1 cluster, they must share the same key
```
rpastrana: Do we detect/report mismatched key in the situation described above? If not, will the user simply receive a failed file access?

esp/services/ws_dfu/ws_dfuService.cpp
```diff
@@ -6268,119 +6212,194 @@ void CWsDfuEx::exportRecordDefinitionBinaryType(const char *recordDefinition, Me
...
+
+        StringBuffer metaInfoBlob;
+        encodeDFUFileMeta(metaInfoBlob, metaInfo, env);
+
+        IEspDFUFileAccessInfo &accessInfo = resp.updateAccessInfo();
+        accessInfo.setMetaInfoBlob(metaInfoBlob);
+        if (returnTextResponse)
+        {
+            getFileAccessInfo(context, *fileDesc, true, accessInfo);
+            accessInfo.setExpiryTime(metaInfo->queryProp("keyPairName"));
```
rpastrana: noticed the same oddity above, perhaps this is intentional, but the "keyPairName" prop doesn't seem to be directly correlated to the expirytime.

esp/services/ws_dfu/ws_dfuService.cpp
```diff
@@ -6389,57 +6408,71 @@ bool CWsDfuEx::onDFUFileCreate(IEspContext &context, IEspDFUFileCreateRequest &r
     return true;
 }
 
+// NB: deprecated from ver >= 1.50
 bool CWsDfuEx::onDFUFilePublish(IEspContext &context, IEspDFUFilePublishRequest &req, IEspDFUFilePublishResponse &resp)
 {
+    Owned<IException> exception;
```
rpastrana: Hopefully we don't deprecate the entire method, but we deprecate certain fields within the req/resp structure. 
In any case, we need to ensure we're processing correctly for the current client version:

double version = context.getClientVersion();
 if (version > 1.50)
{ ... }
else
{ ... }

esp/services/ws_dfu/ws_dfuService.cpp
```diff
@@ -6448,36 +6481,205 @@ bool CWsDfuEx::onDFUFilePublish(IEspContext &context, IEspDFUFilePublishRequest
...
+    Owned<IFileDescriptor> fileDesc;
+    StringBuffer normalizeTempFileName;
+    try
+    {
+        context.ensureFeatureAccess(FEATURE_URL, SecAccess_Write, ECLWATCH_DFU_ACCESS_DENIED, "WsDfu::DFUPublishFile: Permission denied.");
+
+        const char *fileId = req.getFileId();
+        if (isEmptyString(fileId))
+             throw makeStringException(ECLWATCH_INVALID_INPUT, "No FileId defined.");
```
rpastrana: "DFUPublishFile: No FileId defined."

esp/services/ws_dfu/ws_dfuService.cpp
```diff
@@ -6448,36 +6481,205 @@ bool CWsDfuEx::onDFUFilePublish(IEspContext &context, IEspDFUFilePublishRequest
...
+
+        const char *recordDefinition = req.getECLRecordDefinition();
+
```
rpastrana: can we bail out of recordDefinition  is empty?

esp/services/ws_dfu/ws_dfuService.cpp
```diff
@@ -6448,36 +6481,205 @@ bool CWsDfuEx::onDFUFilePublish(IEspContext &context, IEspDFUFilePublishRequest
...
+            normalizeTempFileName.set(lfn.get());
+            fileDesc->getClusterGroupName(0, groupName);
+        }
+        else
+        {
+            groupName.set(fileIdItems.item(0));
+            const char *clusterName = fileIdItems.item(1);
+            tempFileName = fileIdItems.item(2);
+            if (isEmptyString(groupName))
+                 throw makeStringException(ECLWATCH_INVALID_INPUT, "Invalid FileId: empty groupName.");
```
rpastrana: let's provide method name in all err messages for context

rpastrana (changes requested):
@jakesmith a few more comments

esp/services/ws_dfu/ws_dfuService.cpp
```diff
@@ -6037,104 +5980,30 @@ void CWsDfuEx::getFileDafilesrvConfiguration(StringBuffer &keyPairName, unsigned
...
-    CDfsLogicalFileName lfn;
-    lfn.set(req.getName());
-    StringBuffer fileName = lfn.get();
-    if (!isEmptyString(req.getCluster()))
-        fileName.append("@").append(req.getCluster());
-
-    checkLogicalName(fileName, udesc, readPermissions, writePermissions, false, nullptr);
-
-    switch (accessType)
+    getFilePartsInfo(context, fileDesc, false, accessInfo);
```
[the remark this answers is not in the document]
jakesmith: well spotted, that's a bug (at least for legacy clients). Will fix.

esp/services/ws_dfu/ws_dfuService.cpp
```diff
@@ -6153,32 +6022,86 @@ SecAccessFlags translateToSecAccessFlags(CSecAccessType from)
     }
 }
 
+void CWsDfuEx::dFULookupFileCommon(IEspContext &context, const CDfsLogicalFileName &lfn, const char *requestId, unsigned expirySecs, bool returnTextResponse, IEspDFUFileAccessInfo &accessInfo)
+{
+    StringBuffer fileName;
+    lfn.get(fileName, false, true);
+    if (0 == fileName.length())
+         throw MakeStringException(ECLWATCH_INVALID_INPUT, "No Name defined.");
```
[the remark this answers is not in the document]
jakesmith: I can add requestId and the fact it's a lookup, probably not much useful context apart from that though.

esp/services/ws_dfu/ws_dfuService.cpp
```diff
@@ -6153,32 +6022,86 @@ SecAccessFlags translateToSecAccessFlags(CSecAccessType from)
...
+    {
+        StringBuffer clusterName;
+        const char *clusterGroup = df->getClusterName(c, clusterName.clear());
+        groups.push_back(clusterGroup);
+    }
+
+    StringBuffer keyPairName;
+    unsigned port;
+    bool secure;
+    getFileDafilesrvConfiguration(keyPairName, port, secure, fileName, groups);
```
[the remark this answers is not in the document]
jakesmith: It's not an error not to be found here, it would simply mean that this cluster/group has not been configured for security.
It could be that the firewalls/configuration elsewhere means that's okay.
It could also be the case that the meta info was being used, but the physical streams were never used, i.e. the keys were never needed.

It's not until you come to use the meta info to try to connect to a [secure] dafilesrv where you have no keys that it's an issue.

I suspect you're right that could do with clarification somewhere, but I don't think it can go here.
Please open another JIRA re. this, with an example of the error seen when the configuration is wrong and access is denied.

esp/services/ws_dfu/ws_dfuService.cpp
jakesmith: not sure, there's no string comparison until it hits the server (dafilesrv).
i.e. "READ" is in human readable form in the JSON request. 
Where possible, and cost is inconsequential, I think it's best that the JSON request fields remains readable, vs numbers matching enums in a header file somewhere.

esp/services/ws_dfu/ws_dfuService.cpp
jakesmith: hmm agreed!
The keyPairName is already in the metaInfo, this should expirySecs.

esp/services/ws_dfu/ws_dfuService.cpp
jakesmith: nope, I'll delete this code and use userID

esp/services/ws_dfu/ws_dfuService.cpp
jakesmith: well if it is malformed the code that follows will fail and generate a ' Failed to parse ECL ..' error.
I'll move the setProp("ECL ) line below, to make it more clear.

esp/services/ws_dfu/ws_dfuService.cpp
jakesmith: Yes, getFileDafilesrvConfiguration will generate an error if mismatches.

esp/services/ws_dfu/ws_dfuService.cpp
jakesmith: yep, will fix.

esp/services/ws_dfu/ws_dfuService.cpp
jakesmith: Lookup is very similar, but Create/Publish were never quite right and are substantially different
Deprecating is desirable and means more readable code. I'll comment on this at the end as well.

esp/services/ws_dfu/ws_dfuService.cpp
jakesmith: will change.

esp/services/ws_dfu/ws_dfuService.cpp
jakesmith: It is quite normal to publish a file without a record definition. Sprayed files tend to be like that, so I don't want to make it illegal.

esp/services/ws_dfu/ws_dfuService.cpp
jakesmith: agreed, will change.

esp/clients/wsdfuaccess/wsdfuaccess.cpp
jakesmith: It would be better yes, but further to our offline discussion last week:

Esp should offer a standard way to discover a service version, but there is none at the moment.
I'm going to open a JIRA for this.

You mentioned a kludgy route that's used sometimes involving a soap call and parsing soap:address.
As a stop-gap that may be worth pursuing.

@rpastrana - can you point to a code example of this? Is it in a utility class/method I can reuse?

jakesmith:
Re. comments re. keeping existing *Request structures and augmenting them, by adding new fields and deprecating others.

It used to keep common *Request structures, and fill in everything so that it would work with all versions.
New version would ignore the unwanted transmitted requests data, but further to our meeting and discussion with Richard,
it was decided to implement the new *Request format and failover to legacy if dealing with a esp that didn't support it.

In fact if you look at the PR commits, it was like this until last week, see commit: "Split Legacy/Depr. methods, vs new" in the PR.

I __could__ revert to common *Request structures that were an amalgamation of old + new fields with min_ver + depr_ver etc..
But that would absolutely require to know the server version before trying the request with new fields only,
since a legacy server with some old fields populate would likely return with a variety of different obscure errors.
Instead of a 'clean' method not supported if using a new request.

I do not think there is much to gain by keeping a single combination of old+new field request meta in a single combined Requests.
If it were a minor change, e.g. 1 or 2 fields being added or deprecated + didn't involve a big change on the server side interpreting
the new changes, then I'd agree. 

But this is a significant change and the client must differentiate between 2 flavours of request whichever approach is taken.
Consoldating the new *Request's into there own definitions and likewise having the new request handlers separate from the 
legacy request handlers makes for better code readability and it will be easier to maintain going forward.
Hopefully we will be able to delete the deprecated requests and handlers in time.

jakesmith:
@rpastrana - rebased to latest 7.2.x and push a couple of commits based on review.

Please check.

jpmcmu (changes requested):
@jakesmith a few comments / issues. Still reading through though

common/environment/environment.cpp
```diff
@@ -2362,3 +2362,54 @@ extern ENVIRONMENT_API unsigned long readSizeSetting(const char * sizeStr, const
...
+                    {
+                        Owned<IPropertyTreeIterator> espInstanceIter = espProcessIter->query().getElements("Instance");
+                        ForEach(*espInstanceIter)
+                        {
+                            if (espInstanceIter->query().getProp("@computer", espInstanceComputerName.clear()))
+                            {
+                                xpath.setf("Hardware/Computer[@name=\"%s\"]/@netAddress", espInstanceComputerName.str());
+                                if (env->getProp(xpath.str(), instanceAddress.clear()))
+                                {
+                                    fileMetaServiceUrl.setf("%s://%s:%d", bindingProtocol.str(), instanceAddress.str(), espBindingIter->query().getPropInt("@port",8010));
```
rpastrana: If we're trying to find all availble eclwatch instances, looking for 8010 might be ok, but there's nothing keeping an admin from assigning to another port and btw 18010 is the default for https. We might be better off looking for bound EspServices of type=WsSCM or plugin= ws_smc.
If we're looking for wsdfu, we might want to search for bound EspServices of type="WsDfu" and/or plugin="ws_dfu"

jakesmith: Agreed, this code was essentially a tidied up copy of code that does the same thing in fileservices.cpp.
Which also is hard-wired to look for port 8010 - that's ancient code afaik - but also wrong for the same reasons.

The code in fileservices.cpp should use a/this common method as well, then at least there's 1 place to fix this issue - but it's not really a new issue and given fileservices has been living with it for years, I think we can think about how best to do it for all such client discovering services in one place (like environment.*) and fix it once.
Let's open a new JIRA for this issue.

common/remote/dafsstream.cpp
```diff
@@ -0,0 +1,1329 @@
...
+#endif
+            MemoryBuffer metaInfoBlob;
+            metaInfoEnvelope->getPropBin("metaInfoBlob", metaInfoBlob);
+
+            metaInfo.setown(createPTree(metaInfoBlob));
+        }
+        else
+            metaInfo.set(metaInfoEnvelope);
+
+        IPropertyTree *fileInfo = metaInfo->queryPropTree("FileInfo");
```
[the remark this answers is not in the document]
jakesmith: It should never be missing, but worth a check/error JIC.

common/remote/dafsstream.cpp
```diff
@@ -0,0 +1,1329 @@
...
+
+        hosts.resize(numParts*2); // typical max size
+        for (unsigned p=0; p<numParts; p++)
+        {
+            unsigned numCopies = fileDesc->numCopies(p);
+            for (unsigned c=0; c<numCopies; c++)
+            {
+                StringBuffer host;
+                fileDesc->queryNode(p, c)->endpoint().getUrlStr(host);
+                unsigned pos = p*numParts+c;
```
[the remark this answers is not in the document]
jakesmith: yes, that's wrong, and yes should be calculated based on max copies per part.
All parts should have same # copies.
Will fix.

common/remote/dafsstream.cpp
```diff
@@ -0,0 +1,1329 @@
...
+    {
+        return fileDesc->isCompressed();
+    }
+    virtual const char *queryClusterGroupName() const override
+    {
+        return groupName;
+    }
+    virtual const char *queryPartHost(unsigned part, unsigned copy=0) const override
+    {
+        return hosts[part*numParts+copy].c_str();
```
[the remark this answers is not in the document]
jakesmith: Yes same issue, will fix.

common/remote/dafsstream.cpp
```diff
@@ -0,0 +1,1329 @@
...
+    {
+        commCompType.set(compType);
+    }
+    virtual DFUFileOption queryFileOptions() const override
+    {
+        return fileOptions;
+    }
+    virtual bool isFileOptionSet(DFUFileOption opt) const override
+    {
+        return ((unsigned)fileOptions | (unsigned)opt);
```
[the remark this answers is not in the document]
jakesmith: Yes it could be, will change.

common/environment/environment.cpp
rpastrana: ok opened this: https://track.hpccsystems.com/browse/HPCC-21636

common/remote/dafsstream.cpp
```diff
@@ -0,0 +1,1329 @@
...
+            compressor->close();
+            len = compressor->buflen();
+            sendMb.setLength(startPos + sizeof(len)+len);
+        }
+        else
+            len = sendMb.length()-startPos;
+        sendMb.writeEndianDirect(startPos, sizeof(len), &len);
+        return PARENT::send(replyMb);
+    }
+    void _sendWriteFirst()
```
rpastrana: let's rename these 2 methods ( _sendWriteFirst and/or sendWriteFirst) so their purpose is clear from the name...

esp/scm/ws_dfu.ecm
```diff
@@ -950,9 +1002,14 @@ ESPservice [
     ESPmethod DFURecordTypeInfo(DFURecordTypeInfoRequest, DFURecordTypeInfoResponse);
     ESPmethod EclRecordTypeInfo(EclRecordTypeInfoRequest, EclRecordTypeInfoResponse);
 
+// NB: these 3 methods are deprecated from ver >= 1.50
     ESPmethod [auth_feature("DfuAccess:READ"), min_var("1.39")] DFUFileAccess(DFUFileAccessRequest, DFUFileAccessResponse);
     ESPmethod [auth_feature("DfuAccess:FULL"), min_var("1.39")] DFUFileCreate(DFUFileCreateRequest, DFUFileCreateResponse);
     ESPmethod [auth_feature("DfuAccess:FULL"), min_var("1.39")] DFUFilePublish(DFUFilePublishRequest, DFUFilePublishResponse);
+
+    ESPmethod [auth_feature("DfuAccess:READ"), min_var("1.50")] DFULookupFile(DFULookupFileRequest, DFUFileAccessResponse);
```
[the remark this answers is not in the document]
rpastrana: min_var or min_ver?

esp/services/ws_dfu/ws_dfuService.cpp
rpastrana: what I was after was normalization of the ECL used to declare the rec def.
I don't have examples on hand, but from memory, it appears we support multiple ways of declaring a rec def.
As mentioned above, this is perhaps out of scope, but worth having the conversation.

esp/clients/wsdfuaccess/wsdfuaccess.cpp
rpastrana: @jakesmith no example on hand, I don't think this situation has come up before to be honest (deprecated methods, in favor of new methods). 

Until we get a sensible mechanism from the ESP to provide the interface version we might be stuck with your approach, the stop-gap measures I mentioned are indeed a kludge and not necessarily endorsed:

1) fetch HPCC build ver from wssmc and infer wsdfu interface ver.
http://n.n.n.n:8010/WsSMC/Activity returns a "Build" field, which would need to be parsed.

2) Requesting the WSDL http://n.n.n.n:8010/WsDfu/?wsdl
Will provide the default/latest interface version, but also will list the supported methods (but forces the client to interpret WSDL)

3) Requesting esp form: http://n.n.n.n:8010/WsDfu/ <- esp form returns the following HTML:
<html><head><meta http-equiv="Content-Type" content="text/html; charset=utf-8"><title>Enterprise Services Platform</title></head><body><h1>WsDfu</h1><h2>**Version: 1.41**, Build: internal_7.3.0-trunk0[remotes/origin/master-0-g5427b5]</h2>

jakesmith:
Opened JIRA for generic esp service version access : https://track.hpccsystems.com/browse/HPCC-21653

common/remote/dafsstream.cpp
```diff
@@ -0,0 +1,1329 @@
...
+            return;
+        }
+        MemoryBuffer newReplyMb;
+        size32_t oldRemaining = bufRemaining;
+        size32_t leadingSpace = bufRemaining;
+        size32_t hdrSize = sizeof(handle) + sizeof(bufRemaining);
+        if (leadingSpace > hdrSize)
+            leadingSpace -= hdrSize;
+        // reserve space to insert what remains in current replyMb into new, to ensure contiguous
+        newReplyMb.ensureCapacity(leadingSpace);
```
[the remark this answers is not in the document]
jakesmith: No, the logic here is that the new reply (all replies) contain a leading header before the row data.
We want the old remaining data to be copied in front of the row data, and to overwrite this header data.
So as long as the remaining row data is larger than the header data, we reduce 'leadingSpace' by that amount and allocates that much room, such that later when it writes to the front of the reply buffer, it overwrites the reserved space and the head data.

However, this code hadn't been tested! And I found a number of problems, which I've now fixed (in next push).

common/remote/dafsstream.cpp
```diff
@@ -0,0 +1,1329 @@
...
+            // resend original request with cursor
+            size32_t cursorLength;
+            replyMb.skip(bufRemaining);
+            replyMb.read(cursorLength);
+            requestTree->setPropBin("cursorBin", cursorLength, replyMb.readDirect(cursorLength));
+            handle = sendReadStart(); // new handle
+        }
+        ensureAvailable(); // reads from replyMb
+        // now have read new handle and new bufRemaining, copy it back to reserved space
+        replyMb.writeDirect(0, oldRemaining, newReplyMb.readDirect(oldRemaining)); // NB: newReplyMb was swapped and refers to old here
```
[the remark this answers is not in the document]
jakesmith: it should be reading from the old pos, before skipping the cursor, yes.
Part of the fix mentioned above.

esp/services/ws_dfu/ws_dfuService.cpp
jakesmith: ok, I don't think there's currently a way to convert the binary info to ECL (which I want for other reasons).
If we had that, then parsing to the rtlType (which will also validate it) and then converting to ECL - will be a way to normalize it.
I'll open a separate JIRA for this.

common/remote/dafsstream.cpp
jakesmith: done

esp/clients/wsdfuaccess/wsdfuaccess.cpp
jakesmith: it still might be worth kludge, but it's not critical just an improvement for connecting to 7.0.x esp's.
Let's concentrate on getting this merged and I'll open a new JIRA re. adding a kludged getEspVersion() to use in the interim until HPCC-21653 is available

common/remote/dafsstream.cpp
```diff
@@ -0,0 +1,1383 @@
...
+                        replicated = true;
+                    const char *host = part.queryProp("Copy[1]/@host");
+                    SocketEndpoint ep(host);
+                    eps.append(ep);
+                }
+                StringBuffer groupText;
+                eps.getText(groupText);
+                Owned<IGroup> group = createIGroup(eps);
+                ClusterPartDiskMapSpec mspec;
+                mspec.defaultCopies = replicated?DFD_DefaultCopies:DFD_NoCopies;
```
rpastrana: in the replicated case, shouldn't we get the actual number of copies from the file structure rather than defaulting?

esp/scm/ws_dfu.ecm
```diff
@@ -950,9 +1003,14 @@ ESPservice [
...
-    ESPmethod [auth_feature("DfuAccess:FULL"), min_var("1.39")] DFUFileCreate(DFUFileCreateRequest, DFUFileCreateResponse);
-    ESPmethod [auth_feature("DfuAccess:FULL"), min_var("1.39")] DFUFilePublish(DFUFilePublishRequest, DFUFilePublishResponse);
+// NB: these 3 methods are deprecated from ver >= 1.50
+    ESPmethod [auth_feature("DfuAccess:READ"), min_ver("1.39")] DFUFileAccess(DFUFileAccessRequest, DFUFileAccessResponse);
+    ESPmethod [auth_feature("DfuAccess:FULL"), min_ver("1.39")] DFUFileCreate(DFUFileCreateRequest, DFUFileCreateResponse);
+    ESPmethod [auth_feature("DfuAccess:FULL"), min_ver("1.39")] DFUFilePublish(DFUFilePublishRequest, DFUFilePublishResponse);
+
+    ESPmethod [auth_feature("DfuAccess:READ"), min_ver("1.50")] DFUFileAccessV2(DFUFileAccessV2Request, DFUFileAccessResponse);
+    ESPmethod [auth_feature("DfuAccess:FULL"), min_ver("1.50")] DFUFileCreateV2(DFUFileCreateV2Request, DFUFileCreateResponse);
+    ESPmethod [auth_feature("DfuAccess:FULL"), min_ver("1.50")] DFUFilePublishV2(DFUFilePublishV2Request, DFUFilePublishResponse);
```
rpastrana: per offline talk, dfufilepublishv2 can go away in favor of legacy method + overwrite flag

rpastrana (commented):
@jakesmith just a couple more comments

common/remote/dafsstream.cpp
```diff
@@ -0,0 +1,1383 @@
...
+         * But find max just in case.
+         */
+        for (unsigned p=0; p<numParts; p++)
+        {
+            unsigned numCopies = fileDesc->numCopies(p);
+            if (numCopies > maxCopiesPerPart)
+                maxCopiesPerPart = numCopies;
+        }
+        hosts.resize(numParts*maxCopiesPerPart);
+        for (unsigned p=0; p<numParts; p++)
```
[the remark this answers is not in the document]
jakesmith: that's not possible (or numParts==0)

common/remote/dafsstream.cpp
```diff
@@ -0,0 +1,1383 @@
...
+        MemoryBuffer compressedMetaInfoMb;
+        JBASE64_Decode(metaInfoBlobB64, compressedMetaInfoMb);
+        MemoryBuffer decompressedMetaInfoMb;
+        fastLZDecompressToBuffer(decompressedMetaInfoMb, compressedMetaInfoMb);
+        Owned<IPropertyTree> metaInfoEnvelope = createPTree(decompressedMetaInfoMb);
+        StringBuffer metaInfoSignature;
+        if (metaInfoEnvelope->getProp("signature", metaInfoSignature))
+        {
+#if !defined(_USE_OPENSSL) || !defined(_WIN32)
+            // implies will not be able to use this to communicate with secure dafilesrv's
```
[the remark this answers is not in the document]
jakesmith: yes, even if no openssl, it could still be valid, it's the dafilesrv that will check the signature.
So I'll remove these nop lines.

common/remote/dafsstream.cpp
jakesmith: yes, but this is legacy (case 1:) where Esp didn't provide enough info in the response.
In the case of 7.2 esp (case 2:), the whole file descriptor is serialized, which contains much the reasl ClusterPartDiskMapSpec + other info.

esp/scm/ws_dfu.ecm
jakesmith: Merge in latest commit.

jakesmith:
>We might at least want to add logic to account for eclwatch instances bound to https protocol on 18010

Was there a JIRA opened re. this ? - fileservices has been living with talking to WsDfu like this for years...

jakesmith:
There's an intermittent failure in remoteread.ecl (optForceRemote=true), I've now looked at some sample logs, and the reason (at least in part) is that dafilesrv is crashing here:
```
00000055 2019-03-13 19:05:58.640  7247 17411 "================================================"
00000056 2019-03-13 19:05:58.640  7247 17411 "Program:   10.241.40.11:/mnt/disk1/home/vamosax/smoketest/PR-12186/hpcc/opt/HPCCSystems/bin/dafilesrv"
00000057 2019-03-13 19:05:58.640  7247 17411 "Signal:    11 Segmentation fault"
00000058 2019-03-13 19:05:58.640  7247 17411 "Fault IP:  00007FDA91E25341"
00000059 2019-03-13 19:05:58.640  7247 17411 "Accessing: 0000000000000000"
0000005A 2019-03-13 19:05:58.640  7247 17411 "Backtrace:"
0000005B 2019-03-13 19:05:58.640  7247 17411 "  /mnt/disk1/home/vamosax/smoketest/PR-12186/hpcc/opt/HPCCSystems/lib/libjhtree.so(_ZN10CKeyCursor4nextEPcR17KeyStatsCollector+0x31) [0x7fda91e25341]"
0000005C 2019-03-13 19:05:58.640  7247 17411 "  /mnt/disk1/home/vamosax/smoketest/PR-12186/hpcc/opt/HPCCSystems/lib/libjhtree.so(_ZN10CKeyCursor7_lookupEbjR17KeyStatsCollector+0x5f) [0x7fda91e2568f]"
0000005D 2019-03-13 19:05:58.640  7247 17411 "  /mnt/disk1/home/vamosax/smoketest/PR-12186/hpcc/opt/HPCCSystems/lib/libremote.so(_ZN24CRemoteIndexReadActivity7nextRowER19MemoryBufferBuilderRj+0x7b) [0x7fda920bfd8b]"
0000005E 2019-03-13 19:05:58.640  7247 17411 "  /mnt/disk1/home/vamosax/smoketest/PR-12186/hpcc/opt/HPCCSystems/lib/libremote.so(_ZN14CRemoteRequest11processReadEP13IPropertyTreeR12MemoryBuffer+0x424) [0x7fda920cbd54]"
0000005F 2019-03-13 19:05:58.640  7247 17411 "  /mnt/disk1/home/vamosax/smoketest/PR-12186/hpcc/opt/HPCCSystems/lib/libremote.so(_ZN17CRemoteFileServer15cmdStreamCommonEP13IPropertyTreeR12MemoryBufferS3_RNS_20CRemoteClientHandlerE+0x340) [0x7fda920d16d0]"
00000060 2019-03-13 19:05:58.640  7247 17411 "  /mnt/disk1/home/vamosax/smoketest/PR-12186/hpcc/opt/HPCCSystems/lib/libremote.so(_ZN17CRemoteFileServer19cmdStreamReadCommonER12MemoryBufferS1_RNS_20CRemoteClientHandlerE+0x4e) [0x7fda920d22fe]"
00000061 2019-03-13 19:05:58.640  7247 17411 "  /mnt/disk1/home/vamosax/smoketest/PR-12186/hpcc/opt/HPCCSystems/lib/libremote.so(_ZN17CRemoteFileServer14processCommandEhR12MemoryBufferS1_PNS_20CRemoteClientHandlerEPNS_10CThrottlerE+0x27c) [0x7fda920d328c]"
00000062 2019-03-13 19:05:58.640  7247 17411 "  /mnt/disk1/home/vamosax/smoketest/PR-12186/hpcc/opt/HPCCSystems/lib/libremote.so(_ZN17CRemoteFileServer10CThrottler10addCommandEhR12MemoryBufferPNS_20CRemoteClientHandlerE+0x1dc) [0x7fda920d53fc]"
00000063 2019-03-13 19:05:58.641  7247 17411 "  /mnt/disk1/home/vamosax/smoketest/PR-12186/hpcc/opt/HPCCSystems/lib/libremote.so(_ZN17CRemoteFileServer17cCommandProcessor10threadmainEv+0x1e6) [0x7fda920d7336]"
```

@ghalliday - I think I've seen something go by re. an issue in this area?

jakesmith:
Have opened a separate JIRA: https://track.hpccsystems.com/browse/HPCC-21670

rpastrana (approved):
@jakesmith looks reasonable.

jakesmith:
I may as Richard suggested close this PR and open a fresh one.
But also, holding out on investigation James is doing re. apparent stall over secure sockets.

jakesmith:
The issue that @jpmcmu is hitting, seems specific to SSL setups.
It doesn't look related to new code. Still checking though.

Also HPCC-21670, @ghalliday has had a look and can't see how it can happen.
I can't reproduce the issue.
When this PR was pushed, it seemed to hit it intermittently a few times.
I'm going to try and force push this commit a few times (with 0 changes) to provoke the smoketest - to see if we can get more info.

jakesmith:
Closing in favour of : https://github.com/hpcc-systems/HPCC-Platform/pull/12305

PR #12187, opened 2019-02-15
HPCC-21491 LDAP security lib does not send back correct scope permission

RussWhitehead:
When the file scope permission cache has not yet been populated, calls to
get the permission of a child scope erroneously return the permission of the
topmost parent scope.  This PR updates the cache lookup logic to detect that
condition and provide a return code the informs the caller to call to LDAP

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [x] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/system/security/shared/caching.cpp
+++ b/system/security/shared/caching.cpp
@@ -627,7 +627,13 @@ bool CPermissionsCache::queryPermsManagedFileScope(ISecUser& sec_user, const cha
                     return true;
                 }
                 else
+                {
                     matchedRes = res;//allowed at this scope, but must also look at child scopes
+                }
+            }
+            else
+            {
+                matchedRes = nullptr;//don't use parent scope permissions for managed child scope, need to get from LDAP
             }
         }
     }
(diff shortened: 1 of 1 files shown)


RussWhitehead:
@wangkx Please review

system/security/shared/caching.cpp
```diff
@@ -627,7 +627,16 @@ bool CPermissionsCache::queryPermsManagedFileScope(ISecUser& sec_user, const cha
...
+                {
                     matchedRes = res;//allowed at this scope, but must also look at child scopes
+                }
+            }
+            else
+            {
+#ifdef _DEBUG
+                DBGLOG("FileScope %s for user %s managed but not cached", fullScope, sec_user.getName());
+#endif
+                return false;//force callout to LDAP
```
wangkx: Would it be better if you set the matchedRes = false and break so that you may let line 656-663 to handle this case.

RussWhitehead:
@wangkx Please rereview

system/security/shared/caching.cpp
```diff
@@ -627,7 +627,13 @@ bool CPermissionsCache::queryPermsManagedFileScope(ISecUser& sec_user, const cha
                     return true;
                 }
                 else
+                {
                     matchedRes = res;//allowed at this scope, but must also look at child scopes
+                }
+            }
+            else
+            {
+                matchedRes = nullptr;//don't use parent scope permissions for managed child scope, need to get from LDAP
```
wangkx: @RussWhitehead why not break?

RussWhitehead: Because it may find the exact match in a subsequent loop iteration

wangkx: Make sense.

wangkx (approved):
@RussWhitehead looks good now.

RussWhitehead:
@richardkchapman Please merge

PR #12231, opened 2019-02-27
HPCC-21563 Correct xpath evaluation errors

timothyklemm:
Refactor the xpath evaluation functions. This eliminates memory leaks,
produces consistent and accurate error handling, and produces consistent
evaluated results.

Catch exceptions thrown by the evaluation functions. This eliminates memory
leaks and enables meaningful debug output.

Reduce the number of evaluation exceptions by permitting empty variable values.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Premature optimization

### DIFF
--- a/esp/services/esdl_svc_engine/esdl_svc_custom.cpp
+++ b/esp/services/esdl_svc_engine/esdl_svc_custom.cpp
@@ -78,6 +78,18 @@ void CEsdlCustomTransformChoose::processClauses(IPropertyTree *request, IXpathCo
                             request->appendProp(targetField, evaluatedValue.str());
                         }
                     }
+                    catch (IException* e)
+                    {
+                        StringBuffer eMsg;
+                        e->errorMessage(eMsg);
+                        e->Release();
+
+                        VStringBuffer msg("Could not process Custom Transform: '%s' [%s]", cur.queryName(), eMsg.str());
+                        if (!optional)
+                            throw MakeStringException(-1, "%s", msg.str());
+                        else
+                            ERRLOG("%s", msg.str());
+                    }
                     catch (...)
                     {
                         VStringBuffer msg("Could not process Custom Transform: '%s' ", cur.queryName());
@@ -207,9 +219,15 @@ bool CEsdlCustomTransformChoose::evaluate(IXpathContext * xpathContext)
     {
         evalresp = xpathContext->evaluateAsBoolean(m_compiledConditionalXpath.getLink());
     }
+    catch (IException* e)
+    {
+        StringBuffer msg;
+        DBGLOG("%s", e->errorMessage(msg).str());
+        e->Release();
+    }
     catch (...)
     {
-        DBGLOG("CEsdlCustomTransformChoose:evaluate: Could not evaluate xpath '%s'", xpathContext->getXpath());
+        DBGLOG("CEsdlCustomTransformChoose:evaluate: Could not evaluate xpath '%s'", m_compiledConditionalXpath->getXpath());
     }
     return evalresp;
 }
--- a/system/xmllib/libxml_xpathprocessor.cpp
+++ b/system/xmllib/libxml_xpathprocessor.cpp
@@ -34,6 +34,7 @@
 #include <libxml/hash.h>
 
 #include "xpathprocessor.hpp"
+#include "xmlerror.hpp"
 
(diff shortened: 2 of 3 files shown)


timothyklemm:
@afishbeck Please review.

esp/services/esdl_svc_engine/esdl_svc_custom.cpp
```diff
@@ -78,6 +78,18 @@ void CEsdlCustomTransformChoose::processClauses(IPropertyTree *request, IXpathCo
...
                     }
+                    catch (IException* e)
+                    {
+                        StringBuffer eMsg;
+                        e->errorMessage(eMsg);
+                        e->Release();
+
+                        VStringBuffer msg("Could not process Custom Transform: '%s' [%s]", cur.queryName(), eMsg.str());
+                        if (!optional)
+                            throw MakeStringException(-1, "%s", msg.str());
```
afishbeck: Really prefer to declare new codes rather than use -1.

system/xmllib/libxml_xpathprocessor.cpp
```diff
@@ -121,116 +121,31 @@ class CLibXpathContext : public CInterface, public IXpathContext
...
-        {
-            bresult = evaluatedXpathObj->boolval;
-        }
-        else
-            throw MakeStringException(-1,"XpathProcessor:evaluateAsBoolean: Error: Could not evaluate XPATH '%s' as boolean", m_xpath.str());
-
-        xmlXPathFreeObject(evaluatedXpathObj);
-        return bresult;
+        if (!clibCompiledXpath)
+            throw MakeStringException(-1,"XpathProcessor:evaluateAsBoolean: Error: Missing compiled XPATH");
```
afishbeck: Again not a fan of -1 codes.

system/xmllib/libxml_xpathprocessor.cpp
```diff
@@ -257,6 +172,70 @@ class CLibXpathContext : public CInterface, public IXpathContext
         return false;
     }
 
+    bool evaluateAsBoolean(xmlXPathObjectPtr evaluatedXpathObj, const char* xpath)
+    {
+        if (!evaluatedXpathObj)
+        {
+            throw MakeStringException(-1, "XpathProcessor:evaluateAsBoolean: Error: Could not evaluate XPATH '%s'", xpath);
+        }
+        else if (XPATH_BOOLEAN != evaluatedXpathObj->type)
```
afishbeck: Don't really need an else because of the above throw, but ok.

afishbeck (commented):
@timothyklemm minor comments, but looks good.

esp/services/esdl_svc_engine/esdl_svc_custom.cpp
timothyklemm: I generally agree, although -1 is a frequently used code. In this case, the expectation is that the exception thrown here will be consumed by the caller. The callers currently ignore the code, using only the message text for debug logging. Even when the caller propagates an exception, it is not this exception with this code.

system/xmllib/libxml_xpathprocessor.cpp
timothyklemm: Agreed. There were other occurrences I cleaned up, but missed this one.

timothyklemm:
@afishbeck I've updated based on your comments. I will squash following review.

afishbeck (commented):
@timothyklemm looks good.

afishbeck:

richardkchapman:
@timothyklemm Please squash

timothyklemm:
@richardkchapman It's been squashed.

richardkchapman:
Slightly wondering if this should be targeting 7.2.x rather than 7.0.x ?

richardkchapman:
@afishbeck Thoughts?

afishbeck:
@richardkchapman it mixes a bit of refactoring with some memory leak fixes.  We could make this 7.2.x and do memory leak only fixes for 7.0.x.

timothyklemm:
The refactoring was prompted by the observation that each leak occurred twice. By refactoring, I only had to fix each leak once.

PR #12238, opened 2019-02-28
HPCC-21250 Uber H3 geospatial plugin

GordonSmith:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [x] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization


plugins/h3/README.md
```diff
@@ -0,0 +1,231 @@
+H3 Plugin
+================
+
+This plugin exposes a hexagonal hierarchical geospatial indexing system to ECL.  It is a wrapper around Ubers H3 library:
```
richardkchapman: Picky: Missing apostrophe in Uber's

plugins/h3/h3.cpp
```diff
@@ -0,0 +1,247 @@
...
+}
+
+ECL_H3_API void ECL_H3_CALL center(ICodeContext *_ctx, size32_t &__lenResult, void *&__result, unsigned __int64 index)
+{
+    GeoCoord location;
+    ::h3ToGeo(index, &location);
+    __lenResult = 2 * sizeof(double);
+    __result = rtlMalloc(__lenResult);
+    double *cur = static_cast<double *>(__result);
+    *cur = ::radsToDegs(location.lat);
```
richardkchapman: Slightly weird style -I would have said *cur++ on the first line and *cur on the second

plugins/h3/h3.cpp
```diff
@@ -0,0 +1,247 @@
...
+{
+    GeoBoundary boundary;
+    ::h3ToGeoBoundary(index, &boundary);
+    __lenResult = boundary.numVerts * 2 * sizeof(double);
+    __result = rtlMalloc(__lenResult);
+
+    double *cur = static_cast<double *>(__result);
+    for (int v = 0; v < boundary.numVerts; v++)
+    {
+        *cur = ::radsToDegs(boundary.verts[v].lat);
```
richardkchapman: weird style again - *cur++ on both lines would be clearer maybe?

plugins/h3/h3.cpp
```diff
@@ -0,0 +1,247 @@
...
+}
+
+ECL_H3_API uint32_t ECL_H3_CALL baseCell(ICodeContext *_ctx, unsigned __int64 index)
+{
+    return ::h3GetBaseCell(index);
+}
+
+ECL_H3_API void ECL_H3_CALL toString(ICodeContext *_ctx, size32_t &lenVarStr, char *&varStr, unsigned __int64 index)
+{
+    varStr = static_cast<char *>(rtlMalloc(17));
```
richardkchapman: What's special about 17 ? Should it be defined as a constant somewhere?

plugins/h3/h3.cpp
```diff
@@ -0,0 +1,247 @@
...
+    H3Index *buff = static_cast<H3Index *>(calloc(maxBuff, sizeof(H3Index)));
+    ::kRing(index, k, buff);
+    toSetOf(__isAllResult, __lenResult, __result, buff, maxBuff);
+    free(buff);
+}
+
+ECL_H3_API void ECL_H3_CALL hexRing(ICodeContext *_ctx, bool &__isAllResult, size32_t &__lenResult, void *&__result, unsigned __int64 index, uint32_t k)
+{
+    int maxBuff = ::maxKringSize(k);
+    H3Index *buff = static_cast<H3Index *>(calloc(maxBuff, sizeof(H3Index)));
```
richardkchapman: I don't know if exceptions are a possibility but if they are buff is likely to leak. Use of a rtlDataAttr may be preferable.

Also seems a shame to have to copy the data twice - is this to eliminate zeros?

plugins/h3/h3.hpp
```diff
@@ -0,0 +1,48 @@
...
+{
+    ECL_H3_API bool getECLPluginDefinition(ECLPluginDefinitionBlock *pb);
+    ECL_H3_API void setPluginContext(IPluginContext *_ctx);
+}
+
+extern "C++"
+{
+    namespace h3
+    {
+    } // namespace h3
```
richardkchapman: Is there supposed to be something in here?

testing/regress/ecl/h3.ecl
```diff
@@ -0,0 +1,51 @@
+IMPORT lib_h3;
```
richardkchapman: Should maybe have a copyright message, and a //3rdparty tag?

richardkchapman (commented):
Generally looks fine, some minor quibbles

plugins/h3/h3.cpp
GordonSmith: Its from their docs and code:  https://github.com/hpcc-systems/h3/blob/master/src/h3lib/lib/h3Index.c#L66

plugins/h3/h3.cpp
GordonSmith: @richardkchapman - Given H3 is a "c" library they should not be throwing any explicit exceptions - is it still wise to use rtlDataArr?
Yes unfortunately the data can come back "gappy" and either we remove the null records here, or force all the consumers to be aware of that.

testing/regress/ecl/h3.ecl
GordonSmith: Inside the regress suite or inside lib_h3.eclib?
FWIW the h3 sources are also under Apache 2.0 license which is included in the h3 submodule AND non of our other plugins appear to reference there dependent licenses?

plugins/h3/h3.cpp
richardkchapman: You actually copy the data 3 times I think - once into buff, then from there into a vector, then from there into the set that you return.

How common are the gaps, and how often are there less than maxBuff elements added to the ring? It may be preferable to get the data directly into the target set and then close up gaps.

Or this whole discussion may come under the "premature optimization" banner.

GordonSmith: Agree - testing simpler version now.

testing/regress/ecl/h3.ecl
richardkchapman: I was meaning that maybe the regression test file should have a copyright header (some but not all others do), and that a comment in the regression test file of the form

//class=3rdparty

might be appropriate (they are used when deciding what subsets of the regression suite to run).

plugins/h3/h3.cpp
GordonSmith: @richardkchapman ready for re-review

plugins/h3/h3.cpp
```diff
@@ -0,0 +1,247 @@
...
+    int finger = 0;
+    for (int i = 0; i < maxBuffLen; ++i)
+    {
+        if (buff[i] == 0)
+            continue;
+
+        if (i > finger)
+        {
+            buff[finger] = buff[i];
+            buff[i] = 0;
```
GordonSmith: Probably not needed.

richardkchapman: I think it is, otherwise you would duplicate n+1 into n rather than moving it (and subsequent)

richardkchapman:
Looks ok to me - @ghalliday Can you check

plugins/h3/h3.cpp
GordonSmith: I don't think so `buff[that Specific "I"]` never gets read again and will be overridden if/when finger gets there.

plugins/h3/h3.cpp
```diff
@@ -0,0 +1,247 @@
...
+        }
+        ++finger;
+    }
+    __isAllResult = false;
```
GordonSmith: @ghalliday / @richardkchapman I meant to ask - what does ALL indicate for a `SET OF STRING`?

ghalliday: It means all potential strings.  So myString IN ALL is always true.

plugins/h3/README.md
```diff
@@ -0,0 +1,231 @@
...
+
+Returns the parent (coarser) index containing `idx` as resolution `resolution`.
+
+#### children
+
+```c
+SET OF h3_index_t children(CONST h3_index_t idx, CONST h3_resolution_t resolution)
+```
+
+Retruns `children` with the indexes contained by `idx` at resolution `resolution`.
```
ghalliday: typo: Returns

plugins/h3/h3.cpp
ghalliday: I agree the line probably isn't needed

plugins/h3/h3.cpp
```diff
@@ -0,0 +1,247 @@
...
+    pb->ECL = NULL;
+    pb->flags = PLUGIN_IMPLICIT_MODULE;
+    pb->description = "ECL plugin library for uber h3\n";
+    return true;
+}
+
+#include <h3api.h>
+#define H3_INDEX_MINSTRLEN 17
+
+void toSetOf(bool &__isAllResult, size32_t &__lenResult, void *&__result, H3Index *buff, int maxBuffLen)
```
ghalliday: best to add static qualifier so it cannot clash with anything else.

plugins/h3/h3.cpp
```diff
@@ -0,0 +1,247 @@
...
+ECL_H3_API bool ECL_H3_CALL isValid(ICodeContext *_ctx, unsigned __int64 index)
+{
+    return ::h3IsValid(index) != 0;
+}
+
+//  Traversal  ---
+ECL_H3_API void ECL_H3_CALL kRing(ICodeContext *_ctx, bool &__isAllResult, size32_t &__lenResult, void *&__result, unsigned __int64 index, int32_t k)
+{
+    int maxBuff = ::maxKringSize(k);
+    H3Index *buff = static_cast<H3Index *>(calloc(maxBuff, sizeof(H3Index)));
```
ghalliday: theoretically, if this memory is being returned to the query it should be using rtlMalloc instead of calloc.  In practice I don't think it will cause a problem, other plugins do the same (e.g., returning buffer.detach)

testing/regress/ecl/h3.ecl
```diff
@@ -0,0 +1,71 @@
+/*##############################################################################
+
+    HPCC SYSTEMS software Copyright (C) 2014 HPCC Systems®.
```
ghalliday: date needs updating

ghalliday (commented):
@GordonSmith a few minor comments

plugins/h3/h3.cpp
```diff
@@ -0,0 +1,247 @@
...
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+############################################################################## */
+
+#include "platform.h"
+#include "eclrtl.hpp"
+#include "h3.hpp"
+
+#include <vector>
```
richardkchapman: are you still using this?

plugins/h3/h3.cpp
richardkchapman: So do I , I think

plugins/h3/h3.cpp
GordonSmith: Nope

plugins/h3/h3.cpp
GordonSmith: do we have a rtlCalloc?  otherwise I will need to pre clear it.

GordonSmith:
@ghalliday updated with suggested changes (I added a rtlCalloc you will want to review).

ghalliday:
@richardkchapman looks good to merge.

PR #12254, opened 2019-03-04
HPCC-21456 verifyWorkunitDAToken has security hole

RussWhitehead:
verifyWorkunitToken in workunit.cpp should accept a username, and the function
should verify that username matches the username embedded in the signed token

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [x] Security
  - [ ] Thread-safety
  - [ ] Premature optimization


RussWhitehead:
@afishbeck  Please review

richardkchapman:
@ghalliday Can you look too. Should this be in 7.0.x?

common/workunit/workunit.cpp
```diff
@@ -6894,11 +6894,17 @@ bool extractFromWorkunitDAToken(const char * distributedAccessToken, StringBuffe
 //   1 : Signature does not verify (wuid/username don't match, or signature does not verify)
 //   2 : Workunit not active
 //   Throws if unable to open workunit
-wuTokenStates verifyWorkunitDAToken(const char * distributedAccessToken)
+wuTokenStates verifyWorkunitDAToken(const char * _user, const char * _distributedAccessToken)
```
ghalliday: style: I wouldn't prefix the parameter names with _.
Normally parameters prefixed with _ are used for constructors or set functions where the names of the parameters clash with the name of the member.  Occasionally they are also used for factory methods that call those constructors (but generally not).

common/workunit/workunit.cpp
```diff
@@ -6894,11 +6894,17 @@ bool extractFromWorkunitDAToken(const char * distributedAccessToken, StringBuffe
...
 {
+    if (isEmptyString(_user) || isEmptyString(_distributedAccessToken))
```
ghalliday: What happens on a system that doesn't have security enabled?  This will probably cause additional failures.
If it is blank in the access token, and the user is blank then that should match.

E.g.
                 distributedAccessToken="HPCC[u=,w=W20190221-190418];"

 from my local machine.

common/workunit/workunit.cpp
```diff
@@ -6930,14 +6936,19 @@ wuTokenStates verifyWorkunitDAToken(const char * distributedAccessToken)
...
 
     //Verify user matches
-    if (!isEmptyString(cw->queryUser()) || !isEmptyString(tokUser.str()))
+    if (!isEmptyString(cw->queryUser()) && !isEmptyString(tokUser.str()))
     {
-        if (!streq(cw->queryUser(), tokUser.str()))
+        if (!streq(cw->queryUser(), tokUser.str()) || !streq(cw->queryUser(), _user))
         {
-            ERRLOG("verifyWorkunitDAToken : token user does not match workunit");
+            ERRLOG("verifyWorkunitDAToken : token user (%s) does not match WU user (%s) or context user (%s)", tokUser.str(), cw->queryUser(), _user);
```
ghalliday: Worth reporting the different errors separately so they can be distinguished from the messages.

ghalliday: At some point we will probably want to add both of these errors to the audit log since they suggest someone is trying to circumvent the security.  (but not until logging changes have been merged)

ghalliday (commented):
@RussWhitehead a few comments

common/workunit/workunit.cpp
RussWhitehead: Good catch, you are correct

common/workunit/workunit.cpp
RussWhitehead: They can be distinguished from the message, since all 3 usernames are displayed as part of the error message

afishbeck (commented):
@RussWhitehead I don't see any issues beyond Gavin's comments.

RussWhitehead:
@ghalliday @afishbeck Please rereview.

afishbeck (approved):
@RussWhitehead looks good.

richardkchapman:
@ghalliday Please rview and merge if happy

common/workunit/workunit.cpp
ghalliday: true, but they cannot be distinguished by grepping the log messages.

If the token does not match the workunit user then that suggests an internal error.
If the token user does not match the requested user that suggests something that should be audited.
So the messages should really have different target audiences.

common/workunit/workunit.cpp
```diff
@@ -7020,11 +7020,17 @@ bool extractFromWorkunitDAToken(const char * distributedAccessToken, StringBuffe
 //   1 : Signature does not verify (wuid/username don't match, or signature does not verify)
 //   2 : Workunit not active
 //   Throws if unable to open workunit
-wuTokenStates verifyWorkunitDAToken(const char * distributedAccessToken)
+wuTokenStates verifyWorkunitDAToken(const char * ctxUser, const char * daToken)
 {
+    if (isEmptyString(ctxUser) || isEmptyString(daToken))
```
ghalliday: What happens if there is no security - ctxUser will be empty.

common/workunit/workunit.cpp
```diff
@@ -7056,14 +7062,27 @@ wuTokenStates verifyWorkunitDAToken(const char * distributedAccessToken)
...
+        if (!streq(cw->queryUser(), tokUser.str()) || !streq(cw->queryUser(), ctxUser))
         {
-            ERRLOG("verifyWorkunitDAToken : token user does not match workunit");
+            ERRLOG("verifyWorkunitDAToken : token user (%s) does not match WU user (%s) or context user (%s)", tokUser.str(), cw->queryUser(), ctxUser);
             return wuTokenInvalid;
         }
     }
+    else if (!wuUserExist && !tokUserExist)//both users will be empty if no security enabled
+    {
+        //both users will be empty if no security enabled
```
ghalliday: this should probably return false if ctxUser is not empty - only possible to occur on a system started without security, but security then enabled.

ghalliday (commented):
@RussWhitehead a couple of comments (and one reply to the comment on separating the logging messages).
Also needs squashing and removing the merge from the commit list.

common/workunit/workunit.cpp
RussWhitehead: This method is not called if there is no security manager

common/workunit/workunit.cpp
RussWhitehead: This method is not called if there is no security manager

RussWhitehead:
Cancelled, replaced with https://github.com/hpcc-systems/HPCC-Platform/pull/12301

PR #12284, opened 2019-03-11
HPCC-21539 Add All option to WsWorkunits.WUQueryGetSummaryStats

wangkx:
All option: TRUE/FALSE and defaults to FALSE. When set to TRUE,
it returns whatever the roxie's <control:queryAggregates> returns.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/esp/scm/ws_workunits.ecm
+++ b/esp/scm/ws_workunits.ecm
@@ -1187,11 +1187,13 @@ ESPrequest WUQueryGetSummaryStatsRequest
     string QueryId;
     string FromTime; //YYYY-MM-DDTHH:MM:SS
     string ToTime;
+    [min_ver("1.75")] bool All(false); //Return an array of all the raw stats that are available in the nominated time frame
 };
 
 ESPresponse [exceptions_inline] WUQueryGetSummaryStatsResponse
 {
     ESParray<ESPstruct QuerySummaryStats> StatsList;
+    [min_ver("1.75")] string RawStats;
 };
 
 ESPrequest WUExportRequest
@@ -2265,7 +2267,7 @@ ESPresponse [exceptions_inline] WUEclDefinitionActionResponse
 // ----------------------------------------------------------------------------------
 ESPservice [
     auth_feature("DEFERRED"), //This declares that the method logic handles feature level authorization
-    version("1.74"), default_client_version("1.74"), cache_group("ESPWsWUs"),
+    version("1.75"), default_client_version("1.75"), cache_group("ESPWsWUs"),
     noforms,exceptions_inline("./smc_xslt/exceptions.xslt"),use_method_name] WsWorkunits
 {
     ESPmethod [cache_seconds(60), resp_xsl_default("/esp/xslt/workunits.xslt")]     WUQuery(WUQueryRequest, WUQueryResponse);
--- a/esp/services/ws_workunits/ws_workunitsQuerySets.cpp
+++ b/esp/services/ws_workunits/ws_workunitsQuerySets.cpp
@@ -3304,7 +3304,8 @@ bool CWsWorkunitsEx::onWUGetNumFileToCopy(IEspContext& context, IEspWUGetNumFile
     return true;
 }
 
-void getSummaryStatsByQueryId(const char *target, const char *queryId, const char *fromTime, const char *toTime, IArrayOf<IEspQuerySummaryStats>& querySummaryStatsList)
+void getSummaryStatsByQueryId(const char *target, const char *queryId, const char *fromTime,
+    const char *toTime, bool returnAllRawStats, StringBuffer &rawStats, IArrayOf<IEspQuerySummaryStats>& querySummaryStatsList)
 {
     if (!target || !*target)
         throw MakeStringException(ECLWATCH_MISSING_PARAMS, "Target name required");
@@ -3336,6 +3337,12 @@ void getSummaryStatsByQueryId(const char *target, const char *queryId, const cha
         StringBuffer sb;
         toXML(queryAggregates, sb);
         DBGLOG("getSummaryStatsByQueryId(): '%s' => '%s'", control.str(), sb.str());
+    }
+
+    if (returnAllRawStats)
+    {
+        toXML(queryAggregates, rawStats);
+        return;
     }
 
     //Parse queryAggregates and build querySummaryStatsList.
@@ -3383,6 +3390,7 @@ void getSummaryStatsByQueryId(const char *target, const char *queryId, const cha
             querySummaryStats->setStatus(status);
         querySummaryStatsList.append(*querySummaryStats.getLink());
     }
+
     return;
 }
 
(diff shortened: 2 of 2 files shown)


wangkx:
@afishbeck please review.

wangkx:
@AttilaVamos this PR is not couchbase related. Could you please check the errors in the Automated Smoketest?

AttilaVamos:
@wangkx Yes, you are right. This PR is not Couchbase related, but the platform build unfortunately is and business as usual, the problem again with a failed on-fly download.:
```
CMake Error at cmake/Modules/DownloadLcbDep.cmake:18 (MESSAGE):
  Failed to download
  https://packages.couchbase.com/clients/c/mock/CouchbaseMock-1.5.19.jar
````
Rescheduled.

File: esp/services/ws_workunits/ws_workunitsQuerySets.cpp [hunk 1]
```diff
@@ -3338,6 +3339,12 @@ void getSummaryStatsByQueryId(const char *target, const char *queryId, const cha
         DBGLOG("getSummaryStatsByQueryId(): '%s' => '%s'", control.str(), sb.str());
     }
 
+    if (returnAllRawStats)
+    {
+        toXML(queryAggregates, rawStats);
```
afishbeck: Should we also support JSON? 
toJSON(queryAggregates, rawStats);

afishbeck:
@GordonSmith can you make sure this is what you need?

afishbeck (commented):
@wangkx one comment and also asked Gordon to comment.

esp/services/ws_workunits/ws_workunitsQuerySets.cpp
GordonSmith: JSON is my preferred defaults - but I was expecting the same result "shape", just with more items of different time periods?

esp/scm/ws_workunits.ecm
```diff
@@ -1187,11 +1187,13 @@ ESPrequest WUQueryGetSummaryStatsRequest
     string QueryId;
     string FromTime; //YYYY-MM-DDTHH:MM:SS
     string ToTime;
+    [min_ver("1.75")] bool All(false); //Return an array of all the raw stats that are available in the nominated time frame
 };
 
 ESPresponse [exceptions_inline] WUQueryGetSummaryStatsResponse
 {
     ESParray<ESPstruct QuerySummaryStats> StatsList;
+    [min_ver("1.75")] string RawStats;
```
GordonSmith: I was expecting the raw stats to be of the same shape as a `QuerySummaryStats`?

afishbeck: I thought you might say that.  So All would just be a way of removing all the filters, but the info would still be returned the same way.

GordonSmith: Yes

esp/services/ws_workunits/ws_workunitsQuerySets.cpp
wangkx: @GordonSmith  I can check and return a JSON format. A time period is specified in the WUQueryGetSummaryStats request. Are you saying that I should return the Stats without using the time period?

esp/scm/ws_workunits.ecm
GordonSmith: @wangkx - at the moment if you pass in no From/To you get 1 stat back with the entire date range.
If "All" is true I would expect an array of all the stats back (in the same format).

Maybe it might be better if we called it "Aggregate" and defaulted it on?  That way I can ask for a time range and get the aggregated result (singular) or the non - aggregated results (plural)?

wangkx: Correct me if I do not understand correctly:
1. Change the "All" to "Aggregate".
2. default "Aggregate" to on. 
3. When it is on, the WUQueryGetSummaryStats returns as what we have before this PR.
4. When it is not on,  the WUQueryGetSummaryStats returns the non aggregated results  for multiple time ranges.
Questions/concerns: 1. who defines the multiple time ranges? 2. Current QuerySummaryStats (one for each roxie node) already contains all of stats data items returned from roxie. What if some stats data items are added in future (that is why I am using the string RawStats here)? 
@GordonSmith

GordonSmith: 1 - Roxie does
2 - Its not about new "types" of stats - its about new stat records for new time frames.

My understanding is that Roxie maintains stats for:
* every minute (for the last hour)
* every hour (for the last day)
* every day since the server came up

I would like to get that entire list - rather than 1 stat for the specified time frame?

wangkx: Ok, this PR is about new stat records for new time frames.

The response structure may be changed:
ESPStruct QuerySummaryStatsInATimeFrame
{
    string From;
   string To;
   ESParray<ESPstruct QuerySummaryStats> Stats;
};

ESPresponse [exceptions_inline] WUQueryGetSummaryStatsResponse
{
    ESParray<ESPstruct QuerySummaryStatsInATimeFrame> Stats;
};

I need to find out how to get that entire list (hopefully in one roxie control call).

wangkx: @richardkchapman which roxie <control> call can be used to get all (not aggregated) stat records in all time frames for a roxie query (please see Gordon's comment above)?

richardkchapman: They are aggregated as they get less current, for reasons of space. Gordon wants all available information, not all information that has ever existed...

richardkchapman:
@wangkx

esp/scm/ws_workunits.ecm
wangkx: @richardkchapman could you please tell me how to get the 'all available information'?

richardkchapman: else if (stricmp(queryName, "control:queryAggregates")==0)
            {
                time_t from;
                const char *fromTime = control->queryProp("@from");
                if (fromTime)
                {
                    CDateTime f;
                    f.setString(fromTime, NULL, true);
                    from = f.getSimple();
                }
                else
                    from = startupTime;
                time_t to;
                const char *toTime = control->queryProp("@to");
                if (toTime)
                {
                    CDateTime t;
                    t.setString(toTime, NULL, true);
                    to = t.getSimple();
                }
                else
                    time(&to);
                const char *id = control->queryProp("Query/@id");
                if (id)
                {
                    Owned<IQueryFactory> f = getQuery(id, NULL, NULL, logctx);
                    if (f)
                    {
                        Owned<const IPropertyTree> stats = f->getQueryStats(from, to);
                        toXML(stats, reply);
                    }
                    else
                        throw MakeStringException(ROXIE_CONTROL_MSG_ERROR, "Unknown query %s", id);
                }
                else
                {
                    bool includeAllQueries = control->getPropBool("@all", true);
                    Owned<const IPropertyTree> stats = getAllQueryStats(includeAllQueries, from, to);
                    toXML(stats, reply);
                }
            }

richardkchapman:
@wangkx

wangkx:
I did some research on 'control:queryAggregates' related code. Looks like that roxie does store the stats in multiple buffers (one per hour?  std::deque<QueryStatsAggregateRecord> aggregated;). When the 'control:queryAggregates' is called with all=1, those stats inside the buffers are merged (aggregator.mergeStats(thisSlot)) into one stat. I may call the 'control:queryAggregates' for every hour and collect every stats. But, that is not efficient. It could be better if one roxie control call can retrieve all of those stats.

PR #12303, opened 2019-03-14
HPCC-21671 Embedded Java persist semantics tweaks

richardkchapman:
Also allow signature to be optional on all java imports, not just when object
supplied.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] I will update the blog for documentation to plagiarize
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/plugins/javaembed/javaembed.cpp
+++ b/plugins/javaembed/javaembed.cpp
@@ -49,6 +49,7 @@ static const char *version = "Java Embed Helper 1.0.0";
 //#define TRACE_GLOBALREF
 //#define TRACE_CLASSFILE
 //#define CHECK_JNI
+//#define FORCE_GC
 /* Note - if you enable CHECK_JNI and see output like:
  *   WARNING in native method: JNI call made without checking exceptions when required to from CallObjectMethodV
  * where for 'from' may be any of several functions, then the cause is likely to be a missing call to checkException()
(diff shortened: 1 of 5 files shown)


richardkchapman:
@ghalliday Please review

plugins/javaembed/javaembed.cpp
```diff
@@ -3957,34 +3934,51 @@ class JavaEmbedImportContext : public CInterfaceOf<IEmbedFunctionContext>
                 {
                     if (streq(methodName, "<init>"))
                     {
-                        result.l = JNIenv->NewObjectA(javaClass, javaMethodID, args);
+                        if (persistMode == persistNone)
```
richardkchapman: If instance is set, should return it immediately

richardkchapman:
There is a potential issue - bug or feature - concerning number of objects created when using multiple channels per slave together with PERSIST('workunit'). As coded you will get one per slave process which makes it harder for an ECL user to handle if for example they are using the object to aggregate a value that they then want to rollup further.

plugins/javaembed/javaembed.cpp
richardkchapman: Actually looks like it already does.

richardkchapman:
Pushed a second commit adding PERSIST('Channel') support

@rpastrana Can you check whether I broke anything on the 'Java services in ESP' side - I think I spotted a couple of potential issues

PR #12345, opened 2019-03-22
HPCC-21548 Support multiple active fail-safe log files

wangkx:
1. When a tank file is rollovered, it should not be renamed to
.old. This is because, when restarting logging agent thread,
only .log files are checked for not-acked LogReqs;
2. When restarting logging agent thread, we need to remember the
old .log files and rename those files to .old file. Current code
renames all of .log files (except for 'current' .log files),
including possible newly created, but, rollovered .log files;
3. When restarting logging agent thread, read acked LogReqs from
all of ack files. Filter out those acked LogReqs from all of
sending files. Non-acked LogReqs are added into m_PendingLogs.

Also fix 2 bugs.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/esp/logging/logginglib/LogFailSafe.cpp
+++ b/esp/logging/logginglib/LogFailSafe.cpp
@@ -34,7 +34,6 @@
 #define RECEIVING "_acked_"
 #define SENDING   "_sending_"
 
-const char* const RolloverExt=".old";
 const unsigned int TRACE_PENDING_LOGS_MIN = 10;
 const unsigned int TRACE_PENDING_LOGS_MAX = 50;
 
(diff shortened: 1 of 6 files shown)


wangkx:
@timothyklemm please review.

esp/logging/logginglib/LogFailSafe.cpp
```diff
@@ -153,30 +154,35 @@ void CLogFailSafe::LoadOldLogs(StringArray& oldLogData)
...
-        GuidSet ackedSet;
-        getReceiveFileName(file.queryFilename(),ackedName);
-        CLogSerializer ackedLog(ackedName.str());
-        ackedLog.loadAckedLogs(ackedSet);
-
+        oldLogs.append(file.queryFilename()); //add to the files for rollover
         CLogSerializer sendLog(file.queryFilename());
         unsigned long total_missed = 0;
         {//scope needed for critical block below
             CriticalBlock b(m_critSec); //since we plan to use m_PendingLogs
```
timothyklemm: This block is not currently needed as the method is private and only called from the constructor. No contention is possible until the object is constructed.

If the concern is for possible future contention, the block is insufficient. loadSendLogs looks in m_PendingLogs to see if a GUID has already been scheduled for re-processing. If m_PendingLogs is being changed outside this function, it is possible that a GUID is removed from the map before a second reference is encountered. It would be necessary to cache the GUIDs popped from the map so loadSendLogs can see that it is not pending and has not been processed.

esp/logging/logginglib/LogFailSafe.cpp
```diff
@@ -332,30 +338,24 @@ void CLogFailSafe::SafeRollover()
...
+//Only rollover the files inside the oldLogs.
+//This is called only when a log agent is starting.
 void CLogFailSafe::RollOldLogs()
 {
-    StringBuffer filesToFind;
-    filesToFind.appendf("%s*.log",m_LogType.str());
-
-    Owned<IDirectoryIterator> di = createDirectoryIterator(m_logsdir.str(), filesToFind.str());
-    ForEach (*di)
+    ForEachItemIn(i, oldLogs)
```
timothyklemm: Is it guaranteed that if we get here then every previously unacknowledged transaction will exist in a new file? If we process this loop before every transaction is re-recorded, we risk losing transactions in renamed files.

esp/logging/logginglib/logthread.cpp
```diff
@@ -429,7 +429,7 @@ void CLogThread::checkPendingLogs(bool bOneRecOnly)
         }
         //if everything went ok then we should be able to rollover the old logs.
         if (!queueLogError && !bOneRecOnly)
```
timothyklemm: queueLogError is set if enqueue fails for any one transaction. If enqueue successfully submits pending transaction for processing and fails on another, the files will not be rolled over. The problem is that the successfully submitted transaction, if it does not acknowledged, will be recorded in two tank files with two different GUIDs. When the process restarts, it will see the same transaction multiple times without recognizing that they are the same, and duplicate entries could be submitted.

Should enqueue be changed (or a new version defined) that accepts the original GUID, which is available?

esp/logging/logginglib/LogFailSafe.cpp
wangkx: Yes, I will remove the block since this part of the code is only used when the object is constructed.

esp/logging/logginglib/LogFailSafe.cpp
wangkx: All of previously unacknowledged transactions are added into a new tank file when CLogFailSafe is constructed. The CLogFailSafe is constructed when CLogThread is constructed. This part of the code will only be called in CLogThread::run(). So, it is guaranteed that we get here after every previously unacknowledged transaction exists in the new file.

esp/logging/logginglib/logthread.cpp
wangkx: This is a good point. Not sure why two different GUIDs are used. We may have to figure it out.

wangkx:
@timothyklemm please review.

timothyklemm (approved):

wangkx:

PR #12367, opened 2019-03-28
HPCC-21784 Add option to ForeignLogicalFileName to suppress cluster prefix

jakesmith:
…refix

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/docs/EN_US/ECLStandardLibraryReference/SLR-Mods/ForeignLogicalFileName.xml
+++ b/docs/EN_US/ECLStandardLibraryReference/SLR-Mods/ForeignLogicalFileName.xml
@@ -13,6 +13,7 @@
     </indexterm>(</emphasis> <emphasis>filename </emphasis> <emphasis
   role="bold">[,</emphasis> <emphasis>foreigndali </emphasis> <emphasis
   role="bold">] [,</emphasis> <emphasis>absolutepath</emphasis> <emphasis
+  role="bold">] [,</emphasis> <emphasis>omitClusterPrefix</emphasis> <emphasis
   role="bold">] )</emphasis></para>
 
   <informaltable colsep="1" frame="all" rowsep="1">
@@ -44,6 +45,14 @@
           <entry>Optional. A boolean TRUE/FALSE to indicate whether to prepend
           a tilde (~) to the resulting foreign logical file name. If omitted,
           the default is FALSE.</entry>
+        </row>
+
+        <row>
+          <entry><emphasis>omitClusterPreifx</emphasis></entry>
+
+          <entry>Optional. A boolean TRUE/FALSE to indicate whether the target
+          cluster's prefix should automatically be added if 'filename' is a
+          relative logical file name. If omitted, the default is FALSE.</entry>
         </row>
 
         <row>
--- a/plugins/fileservices/fileservices.cpp
+++ b/plugins/fileservices/fileservices.cpp
@@ -92,7 +92,7 @@ static const char * EclDefinition =
 "  SwapSuperFile(const varstring lsuperfn1,const varstring lsuperfn2) : c,action,globalcontext,entrypoint='fsSwapSuperFile'; \n"
 "  ReplaceSuperFile(const varstring lsuperfn,const varstring lfn,const varstring bylfn) : c,action,globalcontext,entrypoint='fsReplaceSuperFile'; \n"
 "  FinishSuperFileTransaction(boolean rollback=false) : c,action,globalcontext,entrypoint='fsFinishSuperFileTransaction'; \n"
-"  varstring ForeignLogicalFileName(const varstring name, const varstring foreigndali='', boolean abspath=false) : c,context,entrypoint='fsForeignLogicalFileName'; \n"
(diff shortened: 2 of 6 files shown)


jakesmith:
@dcamper - FYI

@AttilaVamos - please review

AttilaVamos (approved):
@jakesmith It seems good.

jakesmith:

docs/EN_US/ECLStandardLibraryReference/SLR-Mods/ForeignLogicalFileName.xml
```diff
@@ -46,6 +47,14 @@
           the default is FALSE.</entry>
         </row>
 
+        <row>
+          <entry><emphasis>omitClusterPreifx</emphasis></entry>
+
+          <entry>Optional. A boolean TRUE/FALSE to indicate whether the target
+          clusters prefix should automatically be added if 'filename' is a
```
richardkchapman:
```suggestion
          cluster's prefix should automatically be added if 'filename' is a
```

plugins/fileservices/fileservices.cpp
```diff
@@ -1611,10 +1611,14 @@ FILESERVICES_API void FILESERVICES_CALL fslFinishSuperFileTransaction(ICodeConte
...
-FILESERVICES_API char *  FILESERVICES_CALL fsForeignLogicalFileName(ICodeContext *ctx, const char *_lfn,const char *foreigndali,bool abspath)
+static char * implementForeignLogicalFileName(ICodeContext *ctx, const char *_lfn, const char *foreigndali, bool abspath, bool omitClusterPrefix)
 {
     StringBuffer lfns;
-    constructLogicalName(ctx, _lfn, lfns);
+
+    Owned<IConstWorkUnit> wu;
+    if (!omitClusterPrefix && _lfn&&(*_lfn != '~'))
+        wu.setown(getWorkunit(ctx));
+    constructLogicalName(wu, _lfn, lfns);
```
richardkchapman: Can wu end up NULL (in Roxie, for example)?

richardkchapman:
@jakesmith Couple of questions

plugins/fileservices/fileservices.cpp
jakesmith: it's okay for wu to be null, code in constructLogicalName(IConstWorkUnit *,..
will check for null.

NB: in this regard, the logic is the same as before this PR, i.e. it used to unconditionally get a 'wu' (if not absolute prefix), and then call on as it does now. 'wu' could still be null.

jakesmith:
@richardkchapman - fixed doc. issue + answered other question.

richardkchapman:
@jakesmith A test in regression suite would be good

richardkchapman:
I think this is going to cause compatibility issues if a query compiled with older version tries to link to new libfileservices or vice versa (e.g. deployed roxie queries). May need to leave the old function present and reference a new one

jakesmith:
@richardkchapman - the old exported symbol 'fsForeignLogicalFileName' is still there/the same.
With a new symbol introduced 'fsForeignLogicalFileName_v2' for the variance and the EclDefinition pointing to it. This is the technique used before afaics, to avoid this problem.
So I think older compiles linking to this newer version of fileservices should be ok?

richardkchapman:
Sorry, missed that.

@ghalliday Canyou sanity check

ghalliday:
@richardkchapman the changes looks good.  I agree a regression test would be useful.

richardkchapman:
@jakesmith Did you want to add a test?

jakesmith:
I'll have a look, not sure how to off hand, because it would generate a different prefix on each engine it ran on, and I'm not sure if there's a way to ask for the cluster prefix (but I could add a new getClusterPrefix() method too).

jakesmith:
@ghalliday - have extended the fileservice.ecl regression test (squashed in).

testing/regress/ecl/fileservice.ecl
```diff
@@ -47,4 +53,8 @@ SEQUENTIAL(
   File.RenameLogicalFile(prefix + 'renametest.d00', prefix + 'afterrename1.d00', true),
   output(compareDatasets(ds, ds2)),
   File.DeleteLogicalFile(prefix + 'afterrename1.d00'),
+  RemovePrefix(File.ForeignLogicalFileName('somescope::somefilename', '192.168.168.168', true), 3);
```
richardkchapman: That ip looks like it won't work on most environments?

richardkchapman:
@jakesmith One question re the test case

testing/regress/ecl/fileservice.ecl
jakesmith: It is a dummy IP. The function only uses it to construct a logical filename, e.g. ~foreign::192.168.168.168::blah

jakesmith:
@richardkchapman - see reply above.

richardkchapman:
@ghalliday

PR #12392, opened 2019-04-04
HPCC-20493 Allow connect and read timeout to be configured in ESP

shamser:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/esp/platform/espcfg.cpp
+++ b/esp/platform/espcfg.cpp
@@ -35,6 +35,7 @@
 #include "xslprocessor.hpp" 
 #include "espcontext.hpp"
 #include "mplog.hpp"
+#include "rmtfile.hpp"
 
 #include <dalienv.hpp>
 
@@ -324,6 +325,10 @@ CEspConfig::CEspConfig(IProperties* inputs, IPropertyTree* envpt, IPropertyTree*
         StringBuffer daliservers;
         if (m_cfg->getProp("@daliServers", daliservers))
             initDali(daliservers.str()); //won't init if detached
+
+        const unsigned dafilesrvConnectTimeout = m_cfg->getPropInt("@dafilesrvConnectTimeout", 10)*1000;
+        const unsigned dafilesrvReadTimeout = m_cfg->getPropInt("@dafilesrvReadTimeout", 10)*1000;
+        setRemoteFileTimeouts(dafilesrvConnectTimeout, dafilesrvReadTimeout);
 
 #ifndef _DEBUG
         startPerformanceMonitor(m_cfg->getPropInt("@perfReportDelay", 60)*1000);
--- a/initfiles/componentfiles/configschema/xsd/esp.xsd
+++ b/initfiles/componentfiles/configschema/xsd/esp.xsd
@@ -380,7 +380,14 @@
                 <xs:attribute name="maxPersistentRequests" type="xs:integer"
                               hpcc:displayName="Max Persistent Requests" hpcc:presetValue="100"
                               hpcc:tooltip="Maximum number of query requests per persistent http connection. (-1 for unlimited, 0 to disable)"/>
-
+                <xs:attribute name="dafilesrvConnectTimeout" type="xs:nonNegativeInteger"
+                              hpcc:displayName="Remote file connection timeout in seconds"
+                              hpcc:presentValue="10"
+                              hpcc:tooltip="Remote file access connection timeout in seconds"/>
+                <xs:attribute name="dafilesrvReadTimeout" type="xs:nonNegativeInteger"
+                              hpcc:displayName="Remote file read timeout in seconds"
+                              hpcc:presentValue="10"
+                              hpcc:tooltip="Remote file access read timeout in seconds"/>
             </xs:complexType>
         </xs:element>
     </hpcc:insert>
--- a/esp/platform/CMakeLists.txt
+++ b/esp/platform/CMakeLists.txt
@@ -42,6 +42,7 @@ include_directories (
          ./../bindings/http/platform 
          ./../../common/environment 
          ./../../system/mp 
+         ./../../fs/dafsclient
          ${CMAKE_BINARY_DIR}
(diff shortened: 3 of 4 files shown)


shamser:
TODO: still working on providing the test cases for this.

shamser:
@afishbeck Please can you review these changes relating being able to set the timeouts for ESP services.   By the way, I've tested this manually but (1) created a multi-node thor (2) spraying to this thor (3) shutting down the remote thor node's dafilesrv (4) executing WUInfo, DFUQuery and DFUInfo ESP queries.  Compare how long it takes to timeout with and without the new settings.  (I'm inclined not produce test cases because it seems to be quite a lot of work for not too much benefit)

esp/platform/espcfg.cpp
```diff
@@ -325,6 +326,10 @@ CEspConfig::CEspConfig(IProperties* inputs, IPropertyTree* envpt, IPropertyTree*
         if (m_cfg->getProp("@daliServers", daliservers))
             initDali(daliservers.str()); //won't init if detached
 
+        const unsigned dafilesrvConnectTimeout = m_cfg->getPropInt("@dafilesrvConnectTimeout", 2000);
```
afishbeck: How were the defaults decided on?  Will a change in default behavior catch the users off guard?  Looks like old behavior was the equivalent of 0's here?

afishbeck (commented):
@shamser one question.

esp/platform/espcfg.cpp
shamser: We discussed the default timeout period last Monday.  @jakesmith suggested that having shorter timeout period would mean be sensible (resolves issues such as https://track.hpccsystems.com/browse/HPCC-19002) and if it were to cause problems, the timeout period could be extended with the configuration option. 

I agree that it would cause problems if say a 30 second timeout is necessary and eventually returns a results.  But does any remote file access require this length of time?  May be 2 seconds timeout is too short?

@jakesmith ?

jakesmith: My suggestion was that global default (not just in esp) should be much shorter than it is now.
( Currently the defaults are 100 seconds and 3 retries. )
And it should be globally configurable (e.g. via environment.conf), so that if need / if problematic, it can be raised.

I think I had suggested changing it to 10 seconds (3 retries may be still ok).

Once that is done, whilst having a separate configuration in Esp is okay, it becomes less relevant.

There is also the further improvement of implementing a better mechanism via a timed blacklist, so clients don't keep retrying the same endponit and repeatedly timing out, which cumulatively can be significant even if short. (There is an implementation like this in thorsoapcall, which should be looked at to see whether it can be reused).
@shamser - was a JIRA opened for the blacklist implementation ?

shamser: @jakesmith I have just created the blackout jira: https://track.hpccsystems.com/browse/HPCC-21927. @afishbeck Do you think it should be set to '0' or is 10 seconds ok?

jakesmith: >I have just created the blackout jira
@shamser - blacklist?

jakesmith: > Do you think it should be set to '0' or is 10 seconds ok?

@shamser - are you asking re. this esp custom timeout, or the global timeout I was talking about above?
Has a JIRA been opened to change the global timeout?

If the global timeout is set as suggested (10 secs) - then I'd  default the esp timeout to unset , will default to global timeout - but they (ops) can override it if they want to.

shamser: @jakesmith I'm talking about the ESP custom and default timeouts.  This one is designed to address a number of issues reported relating to ESP services (making remote file requests) taking too long to timeout.  As far as other aspects related to timeouts, I feel that i may not be immediately pressing as no has complained (yet) and we can merge this one until we have agreed the more comprehensive approach for timings.   By the way, there was a bug in the previous commit, which I have now fixed and pushed. (I've also set the default timeout period to 10 seconds).

initfiles/componentfiles/configschema/xsd/esp.xsd
```diff
@@ -380,7 +380,14 @@
                 <xs:attribute name="maxPersistentRequests" type="xs:integer"
                               hpcc:displayName="Max Persistent Requests" hpcc:presetValue="100"
                               hpcc:tooltip="Maximum number of query requests per persistent http connection. (-1 for unlimited, 0 to disable)"/>
-
+                <xs:attribute name="dafilesrvConnectTimeout" type="xs:nonNegativeInteger"
+                              hpcc:displayName="Remote file connection timeout in seconds"
+                              hpcc:presentValue="2"
```
jakesmith: this should be 10 now ?

initfiles/componentfiles/configschema/xsd/esp.xsd
```diff
@@ -380,7 +380,14 @@
...
+                              hpcc:tooltip="Remote file access connection timeout in seconds"/>
+                <xs:attribute name="dafilesrvReadTimeout" type="xs:nonNegativeInteger"
+                              hpcc:displayName="Remote file read timeout in seconds"
+                              hpcc:presentValue="2"
```
jakesmith: this should be 10 now ?

jakesmith:
@shamser - 1 comment. Please go ahead and squash after changing.

shamser:
@jakesmith Squashed.  Please can you do final review.

jakesmith:
Looks good.

richardkchapman:
@JamesDeFabia @shamser Is there a documentation impact?

shamser:
> @JamesDeFabia @shamser Is there a documentation impact?

@jamesdefabia @richardkchapman I have created a documentation Jira: https://track.hpccsystems.com/browse/HPCC-21974. This change affects current behaviour.  How do we get this documented in the Redbook?

richardkchapman:
I have added a tag

PR #12454, opened 2019-04-15
HPCC-21944 Folding function calls that return real values may fail

richardkchapman:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

## Testing:
Using Ubuntu 1804 and the changes in HPCC-21879, zerodivide1.ecl test was failing prior to this fix.

Note that the changes in HPCC21879 are not directly related to this issue, they just happened to give the gcc compiler just that little bit more ability to optimize generated code and mis-optimize this bit.

### DIFF
--- a/ecl/hql/hqlfold.cpp
+++ b/ecl/hql/hqlfold.cpp
@@ -1053,7 +1053,7 @@ IValue * doFoldExternalCall(IHqlExpression* expr, unsigned foldOptions, ITemplat
             {
                 __asm__  __volatile__(
                     "movss  %%xmm0,(%%rdi) \n\t"
-                    :
+                    :"=m"(floatresult)
                     : "D"(&(floatresult))
                 );
             }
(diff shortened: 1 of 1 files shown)


richardkchapman:
@ghalliday 

Note that I have only tested the x64 version - I don't know if the arm / x86 versions were affected or whether the fix here for them isright (though so long as it compiles, it should be at least no worse).

ecl/hql/hqlfold.cpp
```diff
@@ -1230,15 +1230,15 @@ IValue * doFoldExternalCall(IHqlExpression* expr, unsigned foldOptions, ITemplat
             {
                 __asm__  __volatile__(
                     "fsts  s0,[%[fresult]] \n\t"
-                    :
+                    :"=m"floatresult)
```
ghalliday: missing (

ghalliday (commented):
@richardkchapman one problem.  Who can test the different variants?

richardkchapman:
@ghalliday Fixed the missing (, and tested the ARM variants as best I can using https://godbolt.org/

ghalliday:
change looks good.  Should it target an earlier version though.

PR #12455, opened 2019-04-15
HPCC-21935 Fix lazy file io crash

jakesmith:
It was possible for requests from the cache to overlap with
previous requests expiring and being removed.
The overlaping request for a previously cached result, that
was about to be destroyed linked the result before the
PSEUDO_COUNT was set, meaning it could be destroyed twice.

Protect by checking isAlive() when looking up and locking
mutex within beforeDispose()

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/thorlcr/graph/thgraphslave.cpp
+++ b/thorlcr/graph/thgraphslave.cpp
@@ -1945,6 +1945,8 @@ IActivityReplicatedFile *createEnsurePrimaryPartFile(const char *logicalFilename
 class CFileCache;
 class CLazyFileIO : public CInterfaceOf<IFileIO>
 {
+    typedef CInterfaceOf<IFileIO> PARENT;
+
     CFileCache &cache;
     Owned<IActivityReplicatedFile> repFile;
     Linked<IExpander> expander;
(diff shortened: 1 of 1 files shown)


jakesmith:
@ghalliday - please review

ghalliday:
I'm not sure this is the correct change.  As far as I can see the original code didn't have a call to isAlive() which is required if you have a cache which doesn't link the items.  
The code should be:

        Linked<CLazyFileIO> file = files.find(id);
        if (!file || !file->isAlive())

I suspect your change would actually introduce a race condition.
@jakesmith

ghalliday:
See the comments at the bottom of jiface.hpp on the pattern needed.

ghalliday:
ok, looking at the code again I don't think it introduces a new race condition, but it does mean that each call to Release() enters a critical section.

jakesmith:
Yes, I can't see the potential race with mutex lock in Release(), but beforeDispose() pattern is better , as avoids crit per Release(), so I'll change to that.

jakesmith:
@ghalliday - please see revised commit.

PR #12525, opened 2019-05-08
HPCC-21198 Update workunit information using WorkUnitAborts

wangkx:
Workunit aborting state may not be stored in /WorkUnits when the
workunit is locked. In this fix, /WorkUnitAborts is used to
update workunit state.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/common/workunit/workunit.cpp
+++ b/common/workunit/workunit.cpp
@@ -5173,6 +5173,31 @@ static void asyncRemoveFile(const char * ip, const char * name)
     deleteDllWorkQ->post(new asyncRemoveRemoteFileWorkItem(ip, name));
 }
 
+static WUState getWUState(const IPropertyTree * p, std::function<bool()> isAbortingFunc)
+{
+    WUState state = (WUState) getEnum(p, "@state", states);
+    switch (state)
+    {
+    case WUStateRunning:
+    case WUStateDebugPaused:
+    case WUStateDebugRunning:
+    case WUStateBlocked:
+    case WUStateCompiling:
+        if (isAbortingFunc())
+        {
+            state = WUStateAborting;
+        }
+        break;
+    case WUStateSubmitted:
+        if (isAbortingFunc())
+        {
+            state = WUStateAborted;
+        }
+        break;
+    }
+    return state;
+}
+
 class CDaliWorkUnitFactory : public CWorkUnitFactory, implements IDaliClientShutdown
 {
 public:
@@ -5462,8 +5487,64 @@ public:
                 }
                 if (!iter)
                     return NULL;
-                sortElements(iter, sortOrder.get(), nameFilterLo.get(), nameFilterHi.get(), unknownAttributes, elements);
+
+                Owned<IPropertyTreeIterator> lWWUPTreeIter = createWorkunitPTreeIterator(iter);
+                sortElements(lWWUPTreeIter, sortOrder.get(), nameFilterLo.get(), nameFilterHi.get(), unknownAttributes, elements);
                 return conn.getClear();
+            }
+            IPropertyTreeIterator* createWorkunitPTreeIterator(IPropertyTreeIterator* originalIter)
+            {
+                Owned<IRemoteConnection> conWorkUnitAborts = querySDS().connect("/WorkUnitAborts/", myProcessSession(), 0, SDS_LOCK_TIMEOUT);
+                if (!conWorkUnitAborts)
+                    return LINK(originalIter);
+
+                bool hasDynamicAbortingState = false;
+                Owned<IPropertyTree> lWWUPTreeRoot = createPTree("lWWUs");
+                IPropertyTree* wuAbortsPTree = conWorkUnitAborts->queryRoot();
+                ForEach(*originalIter)
+                {
+                    IPropertyTree& cur = originalIter->query();
+
+                    const char* wuid = cur.queryName();
+                    auto isAbortingFunc = [wuid, wuAbortsPTree] () { return wuAbortsPTree->hasProp(wuid); };
+
+                    //NB: In CLocalWorkUnit::getState() it also uses checkAgentRunning() to verify
+                    //the sessionId that the workunit is associated with is still alive. In this
+                    //implementation the cost of doing that per workunit would be prohibitively
+                    //expensive, so it's possible in the rare cases when the state is stale because
+                    //the sessionId is dead, the state from this workunit paging code will be out
+                    //of date.
+                    WUState state = (WUState) getEnum(&cur, "@state", states);
+                    WUState dynamicAbortingState = getWUState(&cur, isAbortingFunc);
+                    if (state == dynamicAbortingState)
+                        lWWUPTreeRoot->addPropTree(wuid, LINK(&cur));
+                    else
+                    {
+                        Owned<IPropertyTree> lWWUPTree = createLightweightWorkunitPTree(&cur, dynamicAbortingState);
+                        lWWUPTreeRoot->addPropTree(wuid, lWWUPTree.getClear());
+                        hasDynamicAbortingState = true;
+                    }
+                }
+                return hasDynamicAbortingState ? lWWUPTreeRoot->getElements("*") : LINK(originalIter);
+            }
+            IPropertyTree* createLightweightWorkunitPTree(IPropertyTree* originalWUTree, WUState& state)
+            {
+                Owned<IPropertyTree> lWWUPTree = createPTree();
+                lWWUPTree->setProp("@jobName", originalWUTree->queryProp("@jobName"));
+                lWWUPTree->setProp("@submitID", originalWUTree->queryProp("@submitID"));
+                lWWUPTree->setProp("@clusterName", originalWUTree->queryProp("@clusterName"));
+                lWWUPTree->setProp("@timeScheduled", originalWUTree->queryProp("@timeScheduled"));
+                lWWUPTree->setProp("@scope", originalWUTree->queryProp("@scope"));
+                lWWUPTree->setProp("Action", originalWUTree->queryProp("Action"));
+                lWWUPTree->setProp("@totalThorTime", originalWUTree->queryProp("@totalThorTime"));
+                lWWUPTree->setProp("@priorityClass", originalWUTree->queryProp("@priorityClass"));
+                lWWUPTree->setPropBool("@protected", originalWUTree->getPropBool("@protected"));
+                setEnum(lWWUPTree, "@state", state, states);
+
+                Owned<IPropertyTree> application = originalWUTree->getBranch("Application");
+                if (application)
+                    lWWUPTree->addPropTree("Application", application.getClear());
+                return lWWUPTree.getClear();
             }
             virtual bool allMatchingElementsReceived() { return true; }//For now, dali always returns all of matched WUs.
         };
@@ -6778,22 +6859,7 @@ void CLocalWorkUnit::checkAgentRunning(WUState & state)
 WUState CLocalWorkUnit::getState() const 
 {
     CriticalBlock block(crit);
-    WUState state = (WUState) getEnum(p, "@state", states);
-    switch (state)
-    {
-    case WUStateRunning:
-    case WUStateDebugPaused:
-    case WUStateDebugRunning:
-    case WUStateBlocked:
-    case WUStateCompiling:
-        if (aborting())
-            state = WUStateAborting;
-        break;
-    case WUStateSubmitted:
-        if (aborting())
-            state = WUStateAborted;
-        break;
-    }
+    WUState state = getWUState(p, std::bind(&CLocalWorkUnit::aborting, this));
     const_cast<CLocalWorkUnit *>(this)->checkAgentRunning(state); //need const_cast as will change state if agent has died
     return state;
 }


wangkx:
@jakesmith please review.

AttilaVamos:
@wangkx The ESP cored 
```
0000007D 2019-05-08 11:54:43.165  1675  1675 "Initializing WsWorkunits_EclWatch_myesp service [process = myesp]"
0000007E 2019-05-08 11:54:43.168  1675  1736 "ECLWorkunit WUSchedule Thread started."
0000007F 2019-05-08 11:54:43.169  1675  1675 "QueryFilesInUse.subscribe() called: QuerySets PackageMaps PackageSets"
00000080 2019-05-08 11:54:43.172  1675  1736 "================================================"
00000081 2019-05-08 11:54:43.172  1675  1736 "Program:   10.241.40.11:/mnt/disk1/home/vamosax/smoketest/PR-12525/hpcc/opt/HPCCSystems/bin/esp"
00000082 2019-05-08 11:54:43.172  1675  1736 "Signal:    11 Segmentation fault"
00000083 2019-05-08 11:54:43.172  1675  1736 "Fault IP:  00007F1E0B31728C"
00000084 2019-05-08 11:54:43.172  1675  1736 "Accessing: 0000000000000000"
00000085 2019-05-08 11:54:43.172  1675  1736 "Backtrace:"
00000086 2019-05-08 11:54:43.195  1675  1675 "CSmartSocketFactory::CSmartSocketFactory(10.241.40.11:9876)"
00000087 2019-05-08 11:54:43.196  1675  1736 "  /mnt/disk1/home/vamosax/smoketest/PR-12525/hpcc/opt/HPCCSystems/lib/libws_workunits.so(_ZN12ws_workunits10WUSchedule3runEv+0x1fc) [0x7f1e0b31728c]"
00000088 2019-05-08 11:54:43.196  1675  1736 "  /mnt/disk1/home/vamosax/smoketest/PR-12525/hpcc/opt/HPCCSystems/lib/libjlib.so(_ZN6Thread5beginEv+0x2c) [0x7f1ea4d7b42c]"
00000089 2019-05-08 11:54:43.196  1675  1736 "  /mnt/disk1/home/vamosax/smoketest/PR-12525/hpcc/opt/HPCCSystems/lib/libjlib.so(_ZN6Thread11_threadmainEPv+0x1e) [0x7f1ea4d7c99e]"
0000008A 2019-05-08 11:54:43.196  1675  1736 "  /lib64/libpthread.so.0(+0x7e25) [0x7f1ea2e1fe25]"
0000008B 2019-05-08 11:54:43.196  1675  1736 "  /lib64/libc.so.6(clone+0x6d) [0x7f1ea2b4d34d]"
0000008C 2019-05-08 11:54:43.196  1675  1736 "Registers:" 
```
From a core trace:
```
Using host libthread_db library "/lib64/libthread_db.so.1".
Core was generated by `esp snmpid=25336'.
Program terminated with signal 11, Segmentation fault.
#0  0x00007f1e0b31728c in ws_workunits::WUSchedule::run (this=0x3e0d370) at /mnt/disk1/home/vamosax/smoketest/PR-12525/HPCC-Platform/esp/services/ws_workunits/ws_workunitsHelpers.cpp:3048
3048	                                if (m_container->hasCacheClient())
 ```

**Comment by wangkx** [comment]
@AttilaVamos the core is not related to this PR. Looks like that a scheduled ECL job is running as soon as the ws_workunits service is loaded (before the service is bound and the m_container is set). I think that  no scheduled job should be picked up before the service is bound. This is a bug in the ESP code. Please create a JIRA and assign to me.

**Comment by AttilaVamos** [comment]
@wangkx done: https://track.hpccsystems.com/browse/HPCC-22091

**Comment by wangkx** [comment]
Thanks.

**Comment by richardkchapman** [comment]
@jakesmith Please review

**Comment by jakesmith** [comment]
@wangkx - can you push force an update, so the smoketest reruns for a clean smoketest?

File: common/workunit/workunit.cpp [hunk 1]
```diff
@@ -5462,6 +5490,8 @@ class CDaliWorkUnitFactory : public CWorkUnitFactory, implements IDaliClientShut
                 }
                 if (!iter)
                     return NULL;
+
+                updateWUState(iter);
```
jakesmith: You should do this after sortElements which filters based on >filter-lo <filter-hi,
otherwise it's potentially checking a lot of workunits that it doesn't ever return.

common/workunit/workunit.cpp
```diff
@@ -5416,6 +5416,34 @@ class CDaliWorkUnitFactory : public CWorkUnitFactory, implements IDaliClientShut
...
+                    WUState state = (WUState) getEnum(&wu, "@state", states);
+                    switch (state)
+                    {
+                    case WUStateRunning:
+                    case WUStateDebugPaused:
+                    case WUStateDebugRunning:
+                    case WUStateBlocked:
+                    case WUStateCompiling:
+                        if (WUAbortPTree->queryPropTree(wu.queryName()))
+                            setEnum(&wu, "@state", WUStateAborting, states);
```
jakesmith: This is a bit dodgy, you're connecting on an unlocked connecting, then modifying state behind potentially locked workunits.
You should use the interface to query the state: workunit->getState() - that returns state and checks if aborting() without modifying the workunit.

common/workunit/workunit.cpp
```diff
@@ -5416,6 +5416,34 @@ class CDaliWorkUnitFactory : public CWorkUnitFactory, implements IDaliClientShut
...
+                        break;
+                    case WUStateSubmitted:
+                        if (WUAbortPTree->queryPropTree(wu.queryName()))
+                            setEnum(&wu, "@state", WUStateAborted, states);
```
jakesmith: as per above.

But also, in stead of re-coding the state logic here (the cases statement), calling IConstWorkunit::getState() on each workunit should in itself be enough (it checks state and aborting())

jakesmith (changes requested):
@wangkx - a few comments.

jakesmith:
NB: getWorkUnitsSorted() already returns IConstWorkUnitIterator , so the caller is dealing with IConstWorkUnit's.

So *it* should be calling IConstWorkUnit::getState() and then it should work and be consistent (regardless of impl. dali/cassandra).

However, I think the problem is that the pager creates an array of CLightweightWorkunitInfo, which only look at @state.

So I think the proper fix is to ensure that the pager's IConstWorkUnits also have pre-checked aborting state (to avoid it going to Dali to check per workunit dynamically).

Given the rest is read-only (and you shouldn't alter the info in Dali), you may need to pass in the calculated state into the CLightweightWorkunitInfo's as you create them.

wangkx:
@jakesmith If I check/update WU state when I create the CLightweightWorkunitInfo (after the sortElements()) and a user wants to sort the WUs based on the state, will the WUs be return in correct order?

jakesmith:
@wangkx then in that case you will have to establish the state before sort.

wangkx:
@jakesmith 3 questions:
1. Is it worth using 2 groups of code to update WU states? If WUs are sorted by state, update WU states before the sortElements(); If not, update WU states when creating the CLightweightWorkunitInfos.
2. If updating WU states before sortElements(), to avoid alter the WU info in Dali, do I have to make a copy of the WU IPropertyTreeIterator from dali and update the WU states inside the copy?
3. How to check WU state with aborting? I prefer to use the /WorkUnitAborts since there is only one call to dali. If I use the IConstWorkUnit::getState(), I have to go to dali for every WUs.

jakesmith:
@wangkx - what I would do is:

1) create a function that creates minimal copy of the wuid tree (only properties that CLightweightWorkunitInfo needs), e.g. function could be called createLightweitghtWorkunitPTree(originalTree)
2) Refactor CLocalWorkUnit::getState() so it calls a utility function which does the same thing, that you can call from your code, to avoid cloning the code. Utility function will have to be passed a function to determine if aborting. Could use a std::function and lambda function to call existing implementation.
3) create a method that creates a IPropertyTreeIterator (that can be passed to sortElements instead of the existing bare iterator). Pre-populate the object that implements the iterator with a list of workunit trees, by iterating the original, whilst iterating, check if in aborting list and check state, if it is aborting, then create a copy with function in (1) and set state.
4) If there are any workunits in new iterator that were changed in (3) use that iterator, otherwise throw it away and use original iterator.

This way you are only copying, changing and using the bare minimum required by the keyed sorting process (sortElements).

common/workunit/workunit.cpp
```diff
@@ -5416,6 +5416,34 @@ class CDaliWorkUnitFactory : public CWorkUnitFactory, implements IDaliClientShut
             StringArray unknownAttributes;
             Owned<CQueryOrFilter> orFilter;
 
+            void updateWUState(IPropertyTreeIterator* wuIter)
+            {
+                Owned<IRemoteConnection> conWorkUnitAborts = querySDS().connect("/WorkUnitAborts/", myProcessSession(), 0, SDS_LOCK_TIMEOUT);
+                IPropertyTree* WUAbortPTree = conWorkUnitAborts->queryRoot();
```
jakesmith: this could crash - conWorkUnitAborts could be null.

wangkx:
@jakesmith to refactor CLocalWorkUnit::getState(), should I also pass the checkAgentRunning to the utility function?

richardkchapman:
What are we trying to fix here? Sounds like some very low-level changes being discussed (to stuff that was coded the way it was because of efficiency concerns)...

jakesmith:
I think the basic issue (that Gordon reported) is that the paging/workunit listing code, doesn't necessarily report the same state for workunits, that you would see if you looked at the workunit individually.

Because the listing/paging code (for efficiency reasons), uses a lightweight implementation of IConstWorkUnit that uses the properties retrieved (en-masse) during the listing search, whereas the standard IConstWorkUnit implementation, connects to dali dynamically during getState() (it also connects to the session manager to see if agent session is still alive), which would be too costly for the paging code to do for per workunit.

So the consistency issue would be fixed quite trivially I think, by changing the listing/paging code to use the standard (non-lightwegitht) version of IConstWorkUnit, thus using the standard getState and e.g. dynamically picking up the same state, but would have a significant performance impact.

This JIRA/PR is trying to fix the state consistency issue efficiently, by ensuring the existing lightweight implementation reflects the dynamic aborting state.

jakesmith:
>to refactor CLocalWorkUnit::getState(), should I also pass the checkAgentRunning to the utility function?

I don't think so.
CLocalWorkUnit::getState() would still look the same, except for the bit that checks if aborting, which would instead be contained in a helper function that was passed a isAborting() function.

richardkchapman:
Is it still lightweight though, if it's doing the extra check for the dynamic aborting state?

jakesmith:
Yes because it only needs to perform 1 connection (for /WorkUnitAborts) when resolving the workunit search list (which is cached and paged through), vs a connection for every workunit in the list.
As it creates the lightweight IConstWorkUnits, it can check (and only if necessary depending on state - as per getState()), against the aborts branch. Most of the time it will not have, but even if it does, it will be quick as 1 connection, 1 fetch of the abort states.

It looks like Cassandra impl. uses the lightweigh workunit impl. also (CLightweightWorkunitInfo), and so IConstWorkUnit::aborting() would not 'see' workunits that are currently aborting. So that would need a similar change for this to be fixed.

The other alternative is to accept that the workunit lists/search results are not necessarily accurate/up-to-date if workunits are currently aborting, i.e. keep it as it is.

richardkchapman:
If we can give the right answer without significant overhead, fine. If we can't, then accept that the answer is not 100% right.

richardkchapman:
@wangkx I think this is back to you

wangkx:
@jakesmith if I do not pass the checkAgentRunning to the utility (only call it from the CLocalWorkUnit::getState()), is it possible that the utility function returns one state, but, the CLocalWorkUnit::getState() returns a different state?

jakesmith:
@wangkx - if it were possible to do cheaply then you we want to, but there is no way to do that without a prohibitively expensive call per workunit to the Dali session manager... 
So yes, it's possible an eclagent is down and therefore not checking it would give an out-of-date result in those unusual circumstances, but there's no effective way to avoid that.

Worth a comment in the workunit get list code to say something about this caveat.

wangkx:
@jakesmith please review the new commit.

wangkx:
Looks like the Smoketest error is unrelated.

common/workunit/workunit.cpp
```diff
@@ -5462,9 +5490,70 @@ class CDaliWorkUnitFactory : public CWorkUnitFactory, implements IDaliClientShut
...
+                bool hasDynamicAbortingState = false;
+                Owned<IPropertyTree> lWWUPTreeRoot = createPTree("lWWUs");
+                IPropertyTree* wuAbortsPTree = conWorkUnitAborts->queryRoot();
+                ForEach(*originalIter)
+                {
+                    IPropertyTree& cur = originalIter->query();
+
+                    bool dynamicAbortingState = false;
+                    const char* wuid = cur.queryName();
+                    WUState state = getState(&cur, wuAbortsPTree, dynamicAbortingState);
```
jakesmith: perhaps slightly cleaner if you removed the boolean here and in definition and in getWUState etc., and simply returned state, then code as:
```
    WUState state = (WUState) getEnum(&cur, "@state", states);
    WUState dynamicAbortingState = getState(&cur, wuAbortsPTree);
    if (state == dynamicAbortingState)
        lWWUPTreeRoot->addPropTree(wuid, LINK(&cur));
...
```

jakesmith: minor: this getState() function which mainly defines a lambda function, could be inlined here as:
```
    auto isAbortingFunc = [wuid, wuAbortsPTree] () { return wuAbortsPTree->hasProp(wuid); };
```
so above code becomes:
```
                    auto isAbortingFunc = [wuid, wuAbortsPTree] () { return wuAbortsPTree->hasProp(wuid); };

                    WUState state = (WUState) getEnum(&cur, "@state", states);
                    WUState dynamicAbortingState = getWUState(&cur, isAbortingFunc);
                    if (state == dynamicAbortingState)
                        lWWUPTreeRoot->addPropTree(wuid, LINK(&cur));
```

common/workunit/workunit.cpp
```diff
@@ -5462,9 +5490,70 @@ class CDaliWorkUnitFactory : public CWorkUnitFactory, implements IDaliClientShut
...
+                //In the CLocalWorkUnit::getState(), the checkAgentRunning() is used to get the latest
+                //state for one WU using Dali session manager. But, this getState() may give an
+                //out-of-date state in some unusual circumstances because no effective way can be used
+                //to get the latest state using Dali session manager.
+                const char* wuid = originalWUTree->queryName();
+                auto aborting = [wuid, wuAbortsPTree] ()
+                {
+                    return wuAbortsPTree->queryPropTree(wuid) ? true : false;
+                };
+                return getWUState(originalWUTree, aborting, dynamicAbortingState);
```
jakesmith: minor: as per above comments, as this method is really only defining a 1 line lambda function, I'd move to inline above.

common/workunit/workunit.cpp
```diff
@@ -5462,9 +5490,70 @@ class CDaliWorkUnitFactory : public CWorkUnitFactory, implements IDaliClientShut
...
+                if (application)
+                    lWWUPTree->addPropTree("Application", application.getClear());
+                return lWWUPTree.getClear();
+            }
+            WUState getState(IPropertyTree* originalWUTree, IPropertyTree* wuAbortsPTree, bool& dynamicAbortingState)
+            {
+                //In the CLocalWorkUnit::getState(), the checkAgentRunning() is used to get the latest
+                //state for one WU using Dali session manager. But, this getState() may give an
+                //out-of-date state in some unusual circumstances because no effective way can be used
+                //to get the latest state using Dali session manager.
```
jakesmith: I'd slightly reword as:

> NB: In CLocalWorkUnit::getState() it also uses checkAgentRunning() to verify the sessionId that the workunit is associated with is still alive. In this implementation the cost of doing that per workunit would be prohibitively expensive, so it's possible in the rare cases when the state is stale because the sessionId is dead, the state from this workunit paging code will be out of date.

common/workunit/workunit.cpp
```diff
@@ -5173,6 +5173,34 @@ static void asyncRemoveFile(const char * ip, const char * name)
     deleteDllWorkQ->post(new asyncRemoveRemoteFileWorkItem(ip, name));
 }
 
+extern WUState getWUState(const IPropertyTree * p, std::function<bool()> aborting, bool & dynamicAbortingState)
```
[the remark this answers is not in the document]
jakesmith: minor: "aborting", "isAbortingFunc" would clarify that it's a function.

jakesmith:
@wangkx - looks good, all comments above are minor. It could remain as it is.

The Cassandra implementation will have the same issue which should be looked at too, but a separate JIRA can be opened and referenced / linked to this one.

common/workunit/workunit.cpp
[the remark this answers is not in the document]
wangkx: Will use static and "isAbortingFunc".

common/workunit/workunit.cpp
```diff
@@ -5462,9 +5490,70 @@ class CDaliWorkUnitFactory : public CWorkUnitFactory, implements IDaliClientShut
...
+
+                Owned<IPropertyTreeIterator> lWWUPTreeIter = createWorkunitPTreeIterator(iter);
+                sortElements(lWWUPTreeIter ? lWWUPTreeIter : iter, sortOrder.get(), nameFilterLo.get(), nameFilterHi.get(), unknownAttributes, elements);
                 return conn.getClear();
             }
+            IPropertyTreeIterator* createWorkunitPTreeIterator(IPropertyTreeIterator* originalIter)
+            {
+                Owned<IRemoteConnection> conWorkUnitAborts = querySDS().connect("/WorkUnitAborts/", myProcessSession(), 0, SDS_LOCK_TIMEOUT);
+                if (!conWorkUnitAborts)
+                    return nullptr;
```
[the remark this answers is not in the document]
wangkx: Will do

common/workunit/workunit.cpp
```diff
@@ -5462,9 +5490,70 @@ class CDaliWorkUnitFactory : public CWorkUnitFactory, implements IDaliClientShut
...
+                    if (!dynamicAbortingState)
+                        lWWUPTreeRoot->addPropTree(wuid, LINK(&cur));
+                    else
+                    {
+                        Owned<IPropertyTree> lWWUPTree = createLightWeitghtWorkunitPTree(&cur, state);
+                        lWWUPTreeRoot->addPropTree(wuid, lWWUPTree.getClear());
+                        hasDynamicAbortingState = true;
+                    }
+                }
+                return hasDynamicAbortingState ? lWWUPTreeRoot->getElements("*") : nullptr;
```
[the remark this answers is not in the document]
wangkx: Ok

common/workunit/workunit.cpp
wangkx: Will change

common/workunit/workunit.cpp
wangkx: ok

common/workunit/workunit.cpp
wangkx: Will move this comment inside the createWorkunitPTreeIterator() since the getState() here is not needed anymore.

wangkx:
@jakesmith I made the changes based on your review. Please review the changes.

jakesmith:
@wangkx - looks good, please squash. Also need to open and reference a JIRA for the cassandra side.

wangkx:
@jakesmith Squashed. New JIRA is opened: https://track.hpccsystems.com/browse/HPCC-22141

jakesmith:
Am not sure this is a fix that should target 7.2.x ..
@richardkchapman do you think should retarget master? (for 7.4)

common/workunit/workunit.cpp
```diff
@@ -5462,9 +5487,65 @@ class CDaliWorkUnitFactory : public CWorkUnitFactory, implements IDaliClientShut
...
+                    //expensive, so it's possible in the rare cases when the state is stale because
+                    //the sessionId is dead, the state from this workunit paging code will be out
+                    //of date.
+                    WUState state = (WUState) getEnum(&cur, "@state", states);
+                    WUState dynamicAbortingState = getWUState(&cur, isAbortingFunc);
+                    if (state == dynamicAbortingState)
+                        lWWUPTreeRoot->addPropTree(wuid, LINK(&cur));
+                    else
+                    {
+                        Owned<IPropertyTree> lWWUPTree = createLightWeitghtWorkunitPTree(&cur, state);
```
richardkchapman: that's not how you spell lightweight

richardkchapman:
7.4 seems more appropriate than 7.2. I don't see any changes in the Cassandra code here?

Still not 100% convinced that the gain is worth the pain here...

jakesmith:
I think the performance pain is minimal (could do with some proof with tests I suppose).
But the code changes and future maintenance of more complex code is bigger consideration (+ still need to Cassandra impl. to be fully consistent - a separate JIRA was opened, referenced above).

wangkx:
I am fixing the typo and changing the base branch to master.

wangkx:
Another way to fix the conflicting state information is to add new field (Aborting) to the WUQuery/WUInfo responses. WUInfo always returns the state from the "@state" and the Aborting from the "/WorkUnitAborts/". WUQuery always returns the state from the "@state". If an IncludeAborting flag is set in WUQuery request, WUQuery returns the Aborting.

richardkchapman:
@wangkx This seems to have stalled?

wangkx:
This PR may be reopened when the code is ready for https://track.hpccsystems.com/browse/HPCC-22141.

PR #12531, opened 2019-05-09
HPCC-22093 Add DataPatterns to STD

GordonSmith:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/ecllibrary/std/DataPatterns/BestRecordStructure.ecl
+++ b/ecllibrary/std/DataPatterns/BestRecordStructure.ecl
@@ -1,0 +1,366 @@
+/***
+ * Function macro that leverages DataPatterns to return a string defining the
+ * best ECL record structure for the input data.
+ *
+ * @param   inFile          The dataset to process; REQUIRED
+ * @param   sampling        A positive integer representing a percentage of
+ *                          inFile to examine, which is useful when analyzing a
+ *                          very large dataset and only an estimatation is
+ *                          sufficient; valid range for this argument is
+ *                          1-100; values outside of this range will be
+ *                          clamped; OPTIONAL, defaults to 100 (which indicates
+ *                          that the entire dataset will be analyzed)
+ * @param   emitTransform   Boolean governing whether the function emits a
+ *                          TRANSFORM function that could be used to rewrite
+ *                          the dataset into the 'best' record definition;
+ *                          OPTIONAL, defaults to FALSE.
+ * @param   textOutput      Boolean governing the type of result that is
+ *                          delivered by this function; if FALSE then a
+ *                          recordset of STRINGs will be returned; if TRUE
+ *                          then a dataset with a single STRING field, with
+ *                          the contents formatted for HTML, will be
+ *                          returned (this is the ideal output if the
+ *                          intention is to copy the output from ECL Watch);
+ *                          OPTIONAL, defaults to FALSE
+ *
+ * @return  A recordset defining the best ECL record structure for the data.
+ *          If textOutput is FALSE (the default) then each record will contain
+ *          one field declaration, and the list of declarations will be wrapped
+ *          with RECORD and END strings; if the emitTransform argument was
+ *          TRUE, there will also be a set of records that that comprise a
+ *          stand-alone TRANSFORM function.  If textOutput is TRUE then only
+ *          one record will be returned, containing an HTML-formatted string
+ *          containing the new field declarations (and optionally the
+ *          TRANSFORM); this is the ideal format if the intention is to copy
+ *          the result from ECL Watch.
+ */
+EXPORT BestRecordStructure(inFile, sampling = 100, emitTransform = FALSE, textOutput = FALSE) := FUNCTIONMACRO
+    LOADXML('<xml/>');
+    #EXPORTXML(bestInFileFields, RECORDOF(inFile));
+    #UNIQUENAME(bestFieldStack);
+    #UNIQUENAME(bestStructType);
+    #UNIQUENAME(bestLayoutType);
+    #UNIQUENAME(bestCapturedPos);
+    #UNIQUENAME(bestPrevCapturedPos);
+    #UNIQUENAME(bestLayoutName);
+    #UNIQUENAME(bestNeedsDelim);
+    #UNIQUENAME(bestNamePrefix);
+
+    IMPORT Std;
+
+    LOCAL OutRec := {STRING s};
+
+    // Build a dataset containing information about embedded records and
+    // child datasets; we need to track the beginning and ending positions
+    // of the fields defined within those structures
+    LOCAL ChildRecInfoLayout := RECORD
+        STRING      layoutType;
+        STRING      layoutName;
+        STRING      fieldName;
+        UNSIGNED2   startPos;
+        UNSIGNED2   endPos;
+    END;
+
+    LOCAL childRecInfo := DATASET
+        (
+            [
+                #SET(bestFieldStack, '')
+                #SET(bestNeedsDelim, 0)
+                #FOR(bestInFileFields)
+                    #FOR(Field)
+                        #IF(%{@isRecord}% = 1 OR %{@isDataset}% = 1)
+                            #IF(%{@isRecord}% = 1)
+                                #SET(bestStructType, 'r')
+                            #ELSE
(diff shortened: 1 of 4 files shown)


GordonSmith:
@richardkchapman @ghalliday First shot at including DataPatterns into STD - are you happy with the submodule location and naming etc.?

richardkchapman:
You will need some code in ecllibrary/teststd too - not sure how to manage that with submodule (I suppose it's ok so long as you don't mind it only being there after the install / make install, but not in the dev directory).

Will also need some documentation.

I assume the purpose of using a submodule is so that it can stay as a bundle while maintaining common code. That's a worthy goal I guess but may end up causing more trouble than it's worth in the long term - perhaps freezing the existing bundle and doing all future development on a copy in the HPCC-Platform repo might make more sense?

GordonSmith:
@dcamper thoughts on dual citizenship of DataPatterns as a Bundle and a STD lib? 
 
I suspect it will be desirable as long as we are supporting versions < 7.4 (and while DataPatterns is undergoing active development)?

GordonSmith:
@richardkchapman updated with extra files and tests 
@JamesDeFabia (heads up on some future documentation needs)

richardkchapman:
Now we have a situation where the code is in one repo and the tests in another... Not conducive to remembering to add tests when adding new functions.

GordonSmith:
That could be improved on, but will require some changes in the DataPatterns submodule (restructure the tests so that they can be run against std + non std DataPatterns module).

(I don't have any personal preference as to the final approach and master code location...)

dcamper:
It is possible to maintain both a bundle version and a Std version, but I think that if we leave DataPatterns as a bundle then we'll run into problems maintaining dual code bases, particularly around the test code.  My initial take on this is that we just just plan on migrating the code to Std.  Until the platform release that contains Std.DataPatterns is published we'll have to maintain both repos, but that is only a temporary problem that generates work only when the DataPatterns code changes.

Does anyone see advantages to retaining both versions that I may be missing?

GordonSmith:
In which case I will simply copy/paste the code into the HPCC-Platform and periodically update it (targeting 7.4)?

dcamper:
Yes, that would be the "work" I was talking about.

There are outstanding issues on the bundle repo that may be worth resolving before 7.4 (feedback appreciated).  Other than outright bugs, the bundle issues are the only things that may force a propagation of code to this PR.

GordonSmith:
Question:  In this new world it may end up that the primary interface will be ECLWatch and that the user may never (or very rarely) invoke the profile call from within their ECL.  
If this turns out to be true, is STD the correct place for it?

dcamper:
I think that depends on the user and use case.  DataPatterns is used today by at least one outside customer as part of their data ingestion process, and I'm currently working on a project that leverages profile output for different reasons.  I think it makes sense to have it located in Std in order to 1) anchor it in a known location and 2) ensure that it is available.

GordonSmith:
Ok, STD it is - with a manual copy pasta until the bundle is retired.

dcamper:
@GordonSmith @richardkchapman Please spend a couple of minutes and look through the issues listed in the bundle (https://github.com/hpcc-systems/DataPatterns/issues).  Are any of them important enough to resolve before 7.4?  Just looking to prioritize work....

JamesDeFabia:
I created a Jira for documentation 
https://track.hpccsystems.com/browse/HPCC-22095

GordonSmith:
I would like to see:
* Histogram
* "Data Discovery" - this would not need to be anything comprehensive for 7.4, but should really outline (by example) a convention for extending with new rules - I would expect a regex based one would be a quick win (phone numbers?) - I know you have given this some thought already, but I vote for including the field ID(s) as part of the input (if a file has "lat" or "latitude" and "lng" or "lon" or "Longitude" as field names, its a pretty big hint whats in those fields!).

richardkchapman:
@dcamper The issues look like they are missing functionality rather than bugs, that could be added at any point in the future and would not block us going into 7.4. Having said that, we do need to make sure that we handle the situation gracefully when EclWatch is presented with a dataset that DataPatterns cannot handle - for example a child dataset or a top-level set. I would think adding support for those two in particular should be considered reasonably high priority or the usefulness of this in 7.4 will be restricted.

GordonSmith:
Question:
Do we continue to include the "HTML Report" in the "std" version - or only include it when the user invokes it from within ECL Watch?
(it does "just work" - but they way its developed and included is slightly out of kilter with the rest of the repository) - I will include in my initial release and we can discuss then...

dcamper:
Thanks for reviewing, guys.  I fully agree with adding SET and child dataset support.  The former is relatively easy, but the latter can get so crazy (e.g. multi-level) I have trouble envisioning the solution, but I'm probably overthinking it.  I'll prioritize those two issues so as to provide some kind of support/output for any dataset presented to DataPatterns.

Histogram should be relatively easy to add as well.

"Data discovery" is a discussion that should probably be taken offline.  I'm fully involved with that now, as a separate project, and one of its goals is the ability to easily extend content detection.  Providing some Jiras relating to module inheritance are resolved....

dcamper:
@GordonSmith Good question re: the HTML report.  The use case for that probably is very closely related to ECL Watch.  I don't see many ECL programmers manually generating a data patterns output and then viewing the graph, while I do see people kicking off the process from within ECL Watch and then wanting to see the results right there.  It may make sense to decouple the two, but is there a big payoff for doing so?  If not then perhaps we should leave it as-is for those (probably few) ECL people who want to see the graphs.

GordonSmith:
Main payoff is for a consistent build paradigm (not a huge selling point) - plus it keeps the std folder clean and to the point.
My instinct says to decouple.

dcamper:
Let's decouple, then.

GordonSmith:
@dcamper is this what you would expect to include?  Or would you also expect the "FromPath" variants as well?

dcamper:
@GordonSmith At the moment, let's leave the FromPath variants out.  They don't work right when a sprayed CSV file is referenced, due to a missing headerLength metadata value in Dali (HPCC-21736).  HPCC-21459 should also probably be resolved.  Once those two issues are resolved we can add the FromPath variants.

GordonSmith:
@AttilaVamos - I believe these failures are unrelated to this PR specifically, is there anyway you could elevate the failing ECL filename into this summary (or even into the failing log file)?

dcamper:
@GordonSmith Update to DataPatterns repo should take care of the smoketest failures.  Diff v1.4.1.

GordonSmith:
@dcamper this PR already had the same fixes in it...
@AttilaVamos I may be missing something obvious but I am unable to track down which test actually failed, the closest I get to is:
```
Workunit:     W20190510-023245
User:         regress
Build Version:community_7.3.0-trunk0[heads/master-0-gc3d4a2]
Cluster:      thor
ESP:          10.241.40.11
Exceptions:   
	Assert failed: TRANSFER(trim('   '), data) != TRANSFER('    ', data)

Problem:      "Failed

What Changed: 

Timing:       
```  
Which isn't too helpful...

richardkchapman:
@ghalliday That looks a little reminiscent of a recent codegen issue?

ghalliday:
Which codegen issue was that?

richardkchapman:
The one where data was not zero-padding before comparing a value of different size.

I think the conclusion was that this was correct behaviour, but if so, perhaps the assert is correct to fail?

ghalliday:
@AttilaVamos the error does have an associated filename and line which would have made it easy for you to locate the test.  E.g. I get:
    /home/gavin/dev/hpcc/ecllibrary/teststd/str/TestTrim.ecl(12,14): error C100000: Assert failed: TRANSFER(trim('   '), data) != TRANSFER('    ', data)

Fix released in https://github.com/hpcc-systems/HPCC-Platform/pull/12566

GordonSmith:
@miguelvazq @jbrundage Please Review

GordonSmith:
Closing in favour of:  https://github.com/hpcc-systems/HPCC-Platform/pull/12568

PR #12550, opened 2019-05-14
HPCC-22013 Replace website-antlr3 submodule with libantlr3c

Michael-Gardner:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/services/ws_sql/antlr3c.cmake
+++ b/esp/services/ws_sql/antlr3c.cmake
@@ -41,7 +41,7 @@ endif()
 include(ExternalProject)
 ExternalProject_Add(
     antlr3c
-    DOWNLOAD_COMMAND tar xzf ${CMAKE_CURRENT_SOURCE_DIR}/website-antlr3/download/C/libantlr3c-3.4.tar.gz
+    DOWNLOAD_COMMAND tar xzf ${CMAKE_CURRENT_SOURCE_DIR}/libantlr3c/libantlr3c-3.4.tar.gz
     SOURCE_DIR ${CMAKE_CURRENT_BINARY_DIR}/antlr3c/src/libantlr3c-3.4
     CONFIGURE_COMMAND ${CMAKE_CURRENT_BINARY_DIR}/antlr3c/src/libantlr3c-3.4/configure ${ANTLRcCONFIGURE_COMMAND_PARAMS} --prefix=${CMAKE_CURRENT_BINARY_DIR}/antlr3c 2> antlr.cfg.err
     PREFIX ${CMAKE_CURRENT_BINARY_DIR}/antlr3c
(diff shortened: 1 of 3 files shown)


Michael-Gardner:
@mckellyln Please review

esp/services/ws_sql/antlr3c.cmake
```diff
@@ -41,7 +41,7 @@ endif()
 include(ExternalProject)
 ExternalProject_Add(
     antlr3c
-    DOWNLOAD_COMMAND tar xzf ${CMAKE_CURRENT_SOURCE_DIR}/website-antlr3/download/C/libantlr3c-3.4.tar.gz
+    DOWNLOAD_COMMAND tar xzf ${CMAKE_CURRENT_SOURCE_DIR}/libantlr3c/libantlr3c-3.4.tar.gz
```
mckellyln: It is a little convoluted that the build untars a tarfile with the same files as what is already in the source.  But this is not really any different than it was.
The:
build/esp/services/ws_sql/antlr3c/src/libantlr3c-3.4/src
dir created from the tar xf cmd, and the:
HPCC-Platform/esp/services/ws_sql/libantlr3c/src
are identical.  Same for include dir and m4 dir.

mckellyln:
@Michael-Gardner I am ok with this change.  It is better than before by having the submodule only have the files we use, and it works correctly.  But I think it may be worth another look, as it seems the submodule source from the tarfile appears to match the files already in the source dir.

richardkchapman:
I think there's a reason we do it this way (but I can't remember what it is!)

mckellyln:
Perhaps then we don't need the source in the source dir ?
Something to look at, but as a separate issue ?
But I am good with this PR as is, it is good and an improvement over the previous.

Michael-Gardner:
@mckellyln Originally I had thought I needed to include the source as is, with the license file, but I also wanted to include the archive to retain the same logic that we were using in the ExternalProject_Add's DOWNLOAD_COMMAND.  I could probably just change the DOWNLOAD_COMMAND to cp that entire directory over into the build directory, and then remove the archive.  I'll do that tomorrow and get back to you.

mckellyln:
@Michael-Gardner back to you one more time.

Michael-Gardner:
@mckellyln I modified the libantlr3c repository and removed that redundant jar.  I altered the DOWNLOAD step on add external project to do a copy instead of an extraction.  Had to make other small changes to insure that it worked correctly, and could find all the necessary files, but nothing major.

richardkchapman:
7.4 would be a more appropriate target

Michael-Gardner:
Rebasing it now

mckellyln:
I see it has already been merged.

PR #12555, opened 2019-05-15
HPCC-dafilesrv-csvsupport

jakesmith:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/fs/dafsserver/dafsserver.cpp
+++ b/fs/dafsserver/dafsserver.cpp
@@ -57,6 +57,10 @@
 
 #include "dafdesc.hpp"
 
+#include "thorcommon.hpp"
+#include "csvsplitter.hpp"
+#include "thorxmlread.hpp"
+
 #include "dafscommon.hpp"
 #include "rmtfile.hpp"
 #include "rmtclient_impl.hpp"
@@ -1124,8 +1128,16 @@ static IOutputMetaData *getTypeInfoOutputMetaData(IPropertyTree &actNode, const 
 
 class CRemoteActivityBase : public CSimpleInterface
 {
-protected:
+public:
+    virtual void init(IPropertyTree &config) { }
 };
+
+template<class ActivityClass> ActivityClass *constructAndInitialize(IPropertyTree &actNode, IFileDescriptor *fileDesc)
+{
+    Owned<ActivityClass> ret = new ActivityClass(actNode, fileDesc);
+    ret->init(actNode); // post constructor initialization
+    return ret.getClear();
+}
 
 class CRemoteDiskBaseActivity : public CRemoteActivityBase, implements IRemoteReadActivity, implements IVirtualFieldCallback
 {
@@ -1133,7 +1145,6 @@ protected:
     StringAttr fileName; // physical filename
     Linked<IOutputMetaData> inMeta, outMeta;
     unsigned __int64 processed = 0;
-    Owned<const IDynamicTransform> translator;
     bool outputGrouped = false;
     bool opened = false;
     bool eofSeen = false;
@@ -1141,25 +1152,22 @@ protected:
     RowFilter filters;
     // virtual field values
     StringAttr logicalFilename;
+    unsigned numInputFields = 0;
 
-    void initCommon(IPropertyTree &config)
+public:
+    IMPLEMENT_IINTERFACE_USING(CRemoteActivityBase)
+
+    CRemoteDiskBaseActivity(IPropertyTree &config, IFileDescriptor *fileDesc)
     {
         fileName.set(config.queryProp("fileName"));
         if (isEmptyString(fileName))
             throw createDafsException(DAFSERR_cmdstream_protocol_failure, "CRemoteDiskBaseActivity: fileName missing");
-
-        record = &inMeta->queryRecordAccessor(true);
-        translator.setown(createRecordTranslator(outMeta->queryRecordAccessor(true), *record));
-        Owned<IPropertyTreeIterator> filterIter = config.getElements("keyFilter");
-        ForEach(*filterIter)
-            filters.addFilter(*record, filterIter->query().queryProp(nullptr));
         logicalFilename.set(config.queryProp("virtualFields/logicalFilename"));
     }
-public:
-    IMPLEMENT_IINTERFACE_USING(CRemoteActivityBase)
-
-    CRemoteDiskBaseActivity(IPropertyTree &config)
+    virtual void init(IPropertyTree &config) override
     {
+        record = &inMeta->queryRecordAccessor(true);
+        numInputFields = record->getNumFields();
     }
 // IRemoteReadActivity impl.
     virtual unsigned __int64 queryProcessed() const override
@@ -1205,41 +1213,41 @@ public:
     }
 };
 
-class CRemoteDiskReadActivity : public CRemoteDiskBaseActivity
+class CRemoteStreamReadBaseActivity : public CRemoteDiskBaseActivity
 {
     typedef CRemoteDiskBaseActivity PARENT;
 
-    CThorContiguousRowBuffer prefetchBuffer;
-    Owned<ISourceRowPrefetcher> prefetcher;
+protected:
     Owned<ISerialStream> inputStream;
     Owned<IFileIO> iFileIO;
     unsigned __int64 chooseN = 0;
     unsigned __int64 startPos = 0;
     bool compressed = false;
-    bool inputGrouped = false;
     bool cursorDirty = false;
-    mutable bool eogPending = false;
-    mutable bool someInGroup = false;
     RtlDynRow *filterRow = nullptr;
     // virtual field values
     unsigned partNum = 0;
     offset_t baseFpos = 0;
 
-    void checkOpen()
+    virtual bool refreshCursor()
+    {
+        if (inputStream->tell() != startPos)
+        {
+            inputStream->reset(startPos);
+            return true;
+        }
+        return false;
+    }
+    bool checkOpen() // NB: returns true if this call opened file
     {
         if (opened)
         {
             if (!cursorDirty)
-                return;
-            if (prefetchBuffer.tell() != startPos)
-            {
-                inputStream->reset(startPos);
-                prefetchBuffer.clearStream();
-                prefetchBuffer.setStream(inputStream);
-            }
+                return false;
+            refreshCursor();
             eofSeen = false;
             cursorDirty = false;
-            return;
+            return false;
         }
         cursorDirty = false;
         OwnedIFile iFile = createIFile(fileName);
@@ -1265,11 +1273,10 @@ class CRemoteDiskReadActivity : public CRemoteDiskBaseActivity
             }
         }
         inputStream.setown(createFileSerialStream(iFileIO, startPos));
-        prefetchBuffer.setStream(inputStream);
-        prefetcher.setown(inMeta->createDiskPrefetcher());
 
         opened = true;
         eofSeen = false;
+        return true;
     }
     void close()
     {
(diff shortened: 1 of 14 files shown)


rtl/eclrtl/rtldynfield.cpp
```diff
@@ -1405,202 +1381,516 @@ class GeneralRecordTranslator : public CInterfaceOf<IDynamicTransform>
...
+            double res = rtlStrToReal(srcLen, source);
+            offset = destType.buildReal(builder, offset, field, res);
+            break;
+        }
+        case type_data:
+        case type_string:
+            //Special case if source and destination types are identical to avoid cloning strings
+            if ((destType.fieldType & (RFTMkind|RFTMebcdic)) == sourceType.getType())
+            {
+                offset = destType.buildString(builder, offset, field, srcLen, source);
```
ghalliday: All calls will need to be buildUtf8 (should possibly be called buildFromUtf8).  Also length is length in codepoints, rather than bytes, so need to call rtlUtf8Length(srcLen, source).  Parameter is clearer if it is called srcSize.

rtl/eclrtl/rtldynfield.cpp
```diff
@@ -1405,202 +1381,516 @@ class GeneralRecordTranslator : public CInterfaceOf<IDynamicTransform>
...
-        if (matchFlags && !destRecInfo.getFixedSize())
+        case type_utf8:
+            //MORE: Could special case casting from utf8 to utf8 similar to strings above
+        case type_unicode:
+        case type_varunicode:
         {
-            for (unsigned idx = 0; idx < sourceRecInfo.getNumFields(); idx++)
+            size32_t utf8chars;
+            rtlDataAttr utf8Text;
+            rtlUtf8ToUtf8X(utf8chars, utf8Text.refstr(), srcLen, source);
```
ghalliday: not needed - assignment is direct, same as all the other calls above.

rtl/eclrtl/rtldynfield.cpp
jakesmith: right, will change.

rtl/eclrtl/rtldynfield.cpp
jakesmith: right, will change.

rtl/eclrtl/rtldynfield.hpp
```diff
@@ -108,11 +108,25 @@ enum class RecordTranslationMode:byte { None = 0, All = 1, Payload = 2, AlwaysDi
 extern ECLRTL_API RecordTranslationMode getTranslationMode(const char *modeStr);
 extern ECLRTL_API const char *getTranslationModeText(RecordTranslationMode val);
 
+interface IDynamicRowIterator;
+interface IDynamicFieldValueFetcher : extends IInterface
+{
+    virtual const byte *queryValue(unsigned fieldNum, size_t &sz) const = 0;
+    virtual IDynamicRowIterator *getNestedIterator(unsigned fieldNum) const = 0;
+    virtual size_t getSize(unsigned fieldNum) const = 0;
```
ghalliday: not used, but reasonable to keep

ghalliday:
@jakesmith looks fairly good.  I think quite a lot of code will move around at a later date, but should be reasonable to merge after a bit more work.

jakesmith:
@ghalliday - have combined into single translator, can you take a look?

ghalliday (commented):
Looks like there were some pending comments - not sure what they were

ghalliday:
I'm not sure about the csv/xml filtering for several reasons
- the fields being filtered on are likely to not be in the output.
- the logic should probably be in the disk reading classes rather than the the dafilesrver activity.  For csv/xml it may require a compound reading class (not quite sure).

Separately I'm not sure if it should be a filter on the expected or the actual dataset.

rtl/eclrtl/rtldynfield.cpp
```diff
@@ -1099,9 +1109,9 @@ class GeneralRecordTranslator : public CInterfaceOf<IDynamicTransform>
         unsigned numOffsets = sourceRecInfo.getNumVarFields() + 1;
         size_t * variableOffsets = (size_t *)alloca(numOffsets * sizeof(size_t));
         RtlRow sourceRow(sourceRecInfo, sourceRec, numOffsets, variableOffsets);  // MORE - could save the max source offset we actually need, and only set up that many...
-        return doTranslate(builder, callback, offset, sourceRow);
+        return doTranslateOpaqueType(builder, callback, offset, &sourceRow);
     }
-    size32_t doTranslate(ARowBuilder &builder, IVirtualFieldCallback & callback, size32_t offset, const RtlRow &sourceRow) const
+    size32_t doTranslateOpaqueType(ARowBuilder &builder, IVirtualFieldCallback & callback, size32_t offset, const void *sourceRow) const
```
ghalliday: can be renamed back to doTranslate

rtl/eclrtl/rtldynfield.cpp
```diff
@@ -1149,9 +1162,23 @@ class GeneralRecordTranslator : public CInterfaceOf<IDynamicTransform>
...
+                if (binarySource)
+                {
+                    const RtlRow &rtlRow = *(const RtlRow *)sourceRow;
+                    sourceOffset = rtlRow.getOffset(matchField);
+                    source = rtlRow.queryRow() + sourceOffset;
+                    copySize = rtlRow.getSize(matchField);
+                }
+                else
+                {
+                    const IDynamicFieldValueFetcher &callbackRowHandler = *(const IDynamicFieldValueFetcher *)sourceRow;
```
ghalliday: Could possibly move this to the branch match_typecast|match_dynamic, since I think it is the only place it is used.

rtl/eclrtl/rtldynfield.cpp
```diff
@@ -1363,47 +1449,50 @@ class GeneralRecordTranslator : public CInterfaceOf<IDynamicTransform>
...
-            expectedSize -= sourceRow.getSize(fieldNo);
-            //DBGLOG("Reducing estimated size by %d to %d for omitted field %d (%s)", (int) sourceRow.getSize(fieldNo), expectedSize, fieldNo, sourceRecInfo.queryName(fieldNo));
+        case type_boolean:
+        case type_int:
+        case type_swapint:
+        case type_packedint:
+        case type_filepos:
+        case type_keyedint:
+        {
+            __int64 res = rtlUtf8ToInt(srcSize, source);
```
ghalliday: Should probably call rtlStrToInt8 - otherwise the length should be in codepoints rather than bytes.
(Not there seems to be a bug in rtlUtf8ToInt - it should be converting the length to bytes)

ghalliday (commented):
A few minor minor comments

rtl/eclrtl/rtldynfield.cpp
jakesmith: would clash with line 1107, that deals with binary rows only.

This method deals with a 'sourceRow' that can either be a binary source row, or a dynamic callback.

rtl/eclrtl/rtldynfield.cpp
jakesmith: you're right, I'll move.

rtl/eclrtl/rtldynfield.cpp
jakesmith: will change.

fs/dafsserver/dafsserver.cpp
```diff
@@ -1124,42 +1128,46 @@ static IOutputMetaData *getTypeInfoOutputMetaData(IPropertyTree &actNode, const
 
 class CRemoteActivityBase : public CSimpleInterface
 {
-protected:
+public:
+    virtual void init(IPropertyTree &config) { }
```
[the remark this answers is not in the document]
jakesmith: ok, will add a comment

fs/dafsserver/dafsserver.cpp
```diff
@@ -1288,29 +1295,86 @@ class CRemoteDiskReadActivity : public CRemoteDiskBaseActivity
...
+    CThorContiguousRowBuffer prefetchBuffer;
+    Owned<ISourceRowPrefetcher> prefetcher;
+    bool inputGrouped = false;
+    mutable bool eogPending = false;
+    mutable bool someInGroup = false;
+    Owned<const IDynamicTransform> translator;
+
+    virtual bool refreshCursor() override
+    {
+        if (prefetchBuffer.tell() != startPos)
```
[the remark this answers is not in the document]
jakesmith: it can, but needs to check CThorContiguousRowBuffer::tell() ..
So could make it:
```
    virtual bool refreshCursor() override
    {
        if (prefetchBuffer.tell() != startPos)
        {
            PARENT::refreshCursor();
            prefetchBuffer.clearStream();
            prefetchBuffer.setStream(inputStream);
            return true;
        }
        return false;
    }

```

fs/dafsserver/dafsserver.cpp
```diff
@@ -1396,21 +1461,370 @@ class CRemoteDiskReadActivity : public CRemoteDiskBaseActivity
...
+    unsigned __int64 headerLines = 0;
+    unsigned __int64 maxRowSize = 0;
+    bool preserveWhitespace = false;
+    CSVSplitter csvSplitter;
+    Owned<const IDynamicFieldValueFetcher> fieldFetcher;
+    Owned<const IDynamicTransform> translator;
+
+    class CFieldFetcher : public CSimpleInterfaceOf<IDynamicFieldValueFetcher>
+    {
+        CSVSplitter &csvSplitter;
```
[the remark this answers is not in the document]
jakesmith: because of the way the splitter [potentially] 1st removes header lines, and how it's effectively used as a stream (eos) as well, I thought it made more sense to be used purely as a utility when dealing with the current line/fields.

jakesmith:
```
I'm not sure about the csv/xml filtering for several reasons

the fields being filtered on are likely to not be in the output.
the logic should probably be in the disk reading classes rather than the the dafilesrver activity. For csv/xml it may require a compound reading class (not quite sure).
Separately I'm not sure if it should be a filter on the expected or the actual dataset.
```

Nor was I, I had left comments in the filter setup for both saying:
```
        /* JCSMORE - this needs resolving properly, currently
         * diskread filters on actual format, and this is filtering on output(projected) format.
         * We probably need expected format sent too, and filter on that..
         */
```

I think needs to be on the expected input (and the same for diskread)

jakesmith:
@ghalliday - have pushed a few minor changes as mentioned in replies.

I agree filtering for csv/xml needs changing, but could we postpone until after this is merged since I think diskread should also change for similar reasons (actual used at moment, but should I think be expected).
Also agree needs further refactoring to e.g. to deal with filtering etc. consistently in a diskread activity.

I think this is useful/working sufficiently to be useful, e.g. for Spark to read csv/xml files which something being worked around at the moment.
Do you agree, shall I move the branch to regular PR?

fs/dafsserver/dafsserver.cpp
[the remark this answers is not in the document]
ghalliday: Leave for now.  When csv and disk read get refactored it will be obvious where the class needs to go.

rtl/eclrtl/rtldynfield.hpp
```diff
@@ -108,11 +108,25 @@ enum class RecordTranslationMode:byte { None = 0, All = 1, Payload = 2, AlwaysDi
...
+    virtual size32_t getRecordSize() const = 0;
```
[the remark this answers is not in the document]
ghalliday: I don't think the implementation for csv makes much sense.  I suspect the estimate size function shouldn't be shared.

fs/dafsserver/dafsserver.cpp
```diff
@@ -1396,21 +1460,413 @@ class CRemoteDiskReadActivity : public CRemoteDiskBaseActivity
...
+class CRemoteCsvReadActivity : public CRemoteStreamReadBaseActivity
+{
+    typedef CRemoteStreamReadBaseActivity PARENT;
+
+    StringBuffer csvQuote, csvSeparate, csvTerminate, csvEscape;
+    unsigned __int64 headerLines = 0;
+    unsigned __int64 maxRowSize = 0;
+    bool preserveWhitespace = false;
+    CSVSplitter csvSplitter;
+    Owned<const IDynamicFieldValueFetcher> fieldFetcher;
```
ghalliday: Cleaner as Owned<CFieldFetcher> - would avoid a cast elsewhere.

rtl/eclrtl/rtlds.cpp
```diff
@@ -1892,3 +1892,14 @@ void MemoryBufferBuilder::finishRow(size32_t length)
     self = NULL;
     reserved = 0;
 }
+
+void MemoryBufferBuilder::removeBytes(size32_t len)
+{
+    dbgassertex(buffer);
+    if (reserved>len)
```
ghalliday: What are these two lines here for?   I would have expected < rather than >

fs/dafsserver/dafsserver.cpp
```diff
@@ -1396,21 +1460,413 @@ class CRemoteDiskReadActivity : public CRemoteDiskBaseActivity
...
+                const void *ret = outBuilder.getSelf();
+                if (fieldFilterMatch(ret))
+                {
+                    outBuilder.finishRow(retSz);
+                    ++processed;
+                    inputStream->skip(lineLength);
+                    return ret;
+                }
+                else
+                    outBuilder.removeBytes(retSz);
```
ghalliday: minor: cleaner might be outBuilder.cancelRow() - which would use the reserved length

ghalliday (commented):
@jakesmith a few more comments.

It would be worth squashing the commits so far.
Also, I don't think this currently supports reading from ascii csv (rather than utf8).  It might be worth trying to get this merged, and then go back to supporting that.

fs/dafsserver/dafsserver.cpp
```diff
@@ -1396,21 +1460,413 @@ class CRemoteDiskReadActivity : public CRemoteDiskBaseActivity
...
+            currentMatch.set(_currentMatch);
+        }
+    // IDynamicFieldValueFetcher impl.
+        virtual const byte *queryValue(unsigned fieldNum, size_t &sz) const override
+        {
+            dbgassertex(fieldNum < numInputFields);
+            dbgassertex(currentMatch);
+
+            size32_t rawSz;
+            const char *ret = currentMatch->readRaw(recInfo.queryName(fieldNum), rawSz);
```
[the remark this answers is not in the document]
ghalliday: i.e. why it would be modified from a const function.

rtl/eclrtl/rtlds.cpp
jakesmith: I think it should be:
```
   assertex(len >= reserved);
```

fs/dafsserver/dafsserver.cpp
jakesmith: well the cast (there's 1) is casting to non-const, but it's also/mainly casting to the impl. the interface doesn't have a setCurrentMatch  (which is specific to the xml impl).

fs/dafsserver/dafsserver.cpp
jakesmith: ok, yes agree that would be cleaner, I'll change.

rtl/eclrtl/rtldynfield.cpp
jakesmith: I think I'll leave as is.

rtl/eclrtl/rtldynfield.hpp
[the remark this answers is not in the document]
jakesmith: isn't the estimate based on csv columns better than nothing?

fs/dafsserver/dafsserver.cpp
```diff
@@ -1288,29 +1298,82 @@ class CRemoteDiskReadActivity : public CRemoteDiskBaseActivity
...
+};
+
+class CRemoteDiskReadActivity : public CRemoteStreamReadBaseActivity
+{
+    typedef CRemoteStreamReadBaseActivity PARENT;
+
+    CThorContiguousRowBuffer prefetchBuffer;
+    Owned<ISourceRowPrefetcher> prefetcher;
+    bool inputGrouped = false;
+    mutable bool eogPending = false;
```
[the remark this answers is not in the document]
jakesmith: I don't know, I'll change.

fs/dafsserver/dafsserver.cpp
jakesmith: hm, actually.. don't see how that can work.
reserved is already extended (during translate, when it ensures..).
removeBytes is reducing reserved.

fs/dafsserver/dafsserver.cpp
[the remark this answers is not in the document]
jakesmith: >Should be xpath rather than name.

right, yes, will fix.

PR #12556, opened 2019-05-15
HPCC-22124 File services permits and defaults to copying a super index into a invalid single index with multiple TLK's

AttilaVamos:
Add Superfile check for a simple logical file copy.

Add Regression Suite test case.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

## Testing:

Tested manually with ECLWatch, ECL code and DFUPlus


AttilaVamos:
@jakesmith please take a look.

AttilaVamos:
@jakesmith should I prevent both Superfile and Superindex to copy as a simple logical file or it should check for Superindex only?

jakesmith:
@AttilaVamos - this issue only effects super index's.

A flat file copied in this way is still valid.
So this PR should only change behaviour for super index files.

AttilaVamos:
@jakesmith  please review.

jakesmith:
@AttilaVamos PR check boxes need filling

jakesmith:
@AttilaVamos - looks good, apart from checkboxes.

I did wonder whether perhaps new regression test is a bit verbose, it could be much shorter, e.g. inline datasets could be 1 row for the purposes of what it's testing.
Or perhaps should be combined with another test.

AttilaVamos:
@jakesmith checkboxes are filled. Honestly I borrowed this dataset from another super index related test and I can agree it would be better to put it into somewhere in setup or cut it shorter.

jakesmith:
@AttilaVamos - the easiest thing to do, is delete everything but the first line of ds.

AttilaVamos:
@jakesmith DsS size reduced and fixed remote/foreign superkey copy. Please re-review.

dali/dfu/dfurun.cpp
```diff
@@ -1220,8 +1220,13 @@ class CDFUengine: public CInterface, implements IDFUengine
                             throw MakeStringException(-1,"Source file %s could not be found in Dali %s",tmp.str(),foreigndalinode->endpoint().getUrlStr(s).str());
                         }
                         kind.set(foreignfdesc->queryProperties().queryProp("@kind"));
-                        oldRoxiePrefix.set(foreignfdesc->queryProperties().queryProp("@roxiePrefix"));
                         iskey = strsame("key", kind);
+                        bool isForeignSuperfile = queryDistributedFileDirectory().isSuperFile(tmp.str(),foreignuserdesc,foreigndalinode);
```
jakesmith: this approach works, but it's a bit costly..
It makes another call to the foreign Dali to get the file tree to check if it's a super, when it has already pulled all the info above (line 1217) when it called getFileDescriptor().
Instead of calling getFileDescriptor() why doesn't it call lookup() (in exactly same way that it would for non-local) to get a IDistributedFile ?
Then both could use the standard way to check for a super which is IDistributedFile::querySuperFile().

The foreign code to that calls getFileDescriptor() vs lookup() isn't new, but I can't see why it does it that way?

jakesmith:
@AttilaVamos - back to you for now, to see if you can't simplify code (why is foreigncopy code is using getFileTree() etc.) and thereby avoid inefficiency of making separate isSuperFile Dali call.

AttilaVamos:
@jakesmith the code block is simplified. Please review my changes.

dali/dfu/dfurun.cpp
```diff
@@ -1199,56 +1199,65 @@ class CDFUengine: public CInterface, implements IDFUengine
...
+                        foreignfdesc.setown(srcFile->getFileDescriptor());
+
+                        if (options->getPush())
+                        {
+                            // need to set ftslave location
                             StringBuffer progpath;
                             StringBuffer workdir;
-                            INode *n = foreignfdesc->queryNode(0);
-                            if (n&&getRemoteRunInfo("FTSlaveProcess", "ftslave", NULL, n->endpoint(), progpath, workdir, foreigndalinode, 1000*60*5)) {
+                            if (foreigndalinode && getRemoteRunInfo("FTSlaveProcess", "ftslave", NULL, foreigndalinode->endpoint(), progpath, workdir, foreigndalinode, 1000*60*5))
```
jakesmith: this used to be:
```
                            INode *n = foreignfdesc->queryNode(0);
                            if (n&&getRemoteRunInfo("FTSlaveProcess", "ftslave", NULL, n->endpoint(), progpath, workdir, foreigndalinode, 1000*60*5)) {
 ```

i.e. it's using the endpoint() of the 1st node of the foreign file.
But now it's using the endpoint of the dalinode - which is not the same and incorrect.

**Review by jakesmith — CHANGES_REQUESTED** [review]

**Comment by jakesmith** [comment]
@AttilaVamos - a couple issues.

Also, I think now that srcFile is always valid, a lot (maybe all) of the places that reference foreignfdesc could be simplified also, i.e. avoid the special casing. You may be able to get rid of foreignfdesc altogether.
That's true for some but not all of the conditional coding on foreigncopy too.

It's not a new issue, but it should make the code simpler and more maintainable.

You could tackle in a separate JIRA if you prefer.
But if this JIRA (as I think it should) is going to target master (for 7.4), I'd be tempted to consolidate these other simplifications, with the related one you've already made here.

File: dali/dfu/dfurun.cpp [hunk 3]
```diff
@@ -1199,56 +1199,65 @@ class CDFUengine: public CInterface, implements IDFUengine
...
+                        throw MakeStringException(-1,"Source file %s could not be found",tmp.str());
+
+                    iskey = isFileKey(srcFile);
+                    if ((cmd==DFUcmd_copy) && (srcFile->querySuperFile() != nullptr) && iskey)
+                        throwError1(DFTERR_InvalidSuperindexCopy, srcFile->queryLogicalName());
+
+                    oldRoxiePrefix.set(srcFile->queryAttributes().queryProp("@roxiePrefix"));
+                    kind.set(srcFile->queryAttributes().queryProp("@kind"));
+
+                    if (destination->getWrap()||iskey)
```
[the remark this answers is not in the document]
AttilaVamos: By mistake and misunderstanding.  (In the original foreigncopy block didn't have it.)
Added

dali/dfu/dfurun.cpp
AttilaVamos: Changed

AttilaVamos:
Replaced by https://github.com/hpcc-systems/HPCC-Platform/pull/12606

PR #12559, opened 2019-05-16
HPCC-22278 Implement a new hthor disk read activity

ghalliday:
## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/common/thorhelper/thorcommon.cpp
+++ b/common/thorhelper/thorcommon.cpp
@@ -819,6 +819,7 @@ extern const char * getActivityText(ThorActivityKind kind)
     case TAKspillread:              return "Spill Read";
     case TAKspillwrite:             return "Spill Write";
     case TAKnwaydistribute:         return "Nway Distribute";
+    case TAKnewdiskread:            return "Disk Read";
     }
     throwUnexpected();
 }
@@ -2097,6 +2098,7 @@ static bool getTranslators(Owned<const IDynamicTransform> &translator, Owned<con
                 if (keyedTranslator && (sourceFormat != expectedFormat))
                 {
                     Owned<const IKeyTranslator> _keyedTranslator = createKeyTranslator(sourceFormat->queryRecordAccessor(true), expectedFormat->queryRecordAccessor(true));
+                    //MORE: What happens if the key filters cannot be translated?
                     if (_keyedTranslator->needsTranslate())
                         keyedTranslator->swap(_keyedTranslator);
                 }
--- a/common/thorhelper/thorread.cpp
+++ b/common/thorhelper/thorread.cpp
@@ -1,0 +1,805 @@
+/*##############################################################################
+
+    HPCC SYSTEMS software Copyright (C) 2019 HPCC Systems®.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+############################################################################## */
+
+#include "jliball.hpp"
+
+#include "thorfile.hpp"
+
+#include "eclhelper.hpp"
+#include "eclrtl.hpp"
+#include "eclrtl_imp.hpp"
+#include "rtlfield.hpp"
+#include "rtlds_imp.hpp"
+#include "rtldynfield.hpp"
+
+#include "rmtclient.hpp"
+#include "rmtfile.hpp"
+
+#include "thorread.hpp"
+#include "rtlcommon.hpp"
+#include "thorcommon.hpp"
+
+//---------------------------------------------------------------------------------------------------------------------
+
+constexpr size32_t defaultReadBufferSize = 0x10000;
+
+class DiskRowReader : extends CInterfaceOf<IAllocRowStream>, implements IRawRowStream, implements IDiskRowReader, implements IThorDiskCallback
+{
+public:
+    DiskRowReader(unsigned _expectedCrc, IOutputMetaData & _expected, unsigned _projectedCrc, IOutputMetaData & _projected, unsigned _actualCrc, IOutputMetaData & _actual, const IPropertyTree * options);
+    IMPLEMENT_IINTERFACE_USING(CInterfaceOf<IAllocRowStream>)
+
+    virtual IRawRowStream * queryRawRowStream() override;
+    virtual IAllocRowStream * queryAllocatedRowStream(IEngineRowAllocator * _outputAllocator) override;
+
+    virtual bool getCursor(MemoryBuffer & cursor) override;
+    virtual void setCursor(MemoryBuffer & cursor) override;
+    virtual void stop() override;
+
+    virtual void clearInput() override;
+    virtual bool matches(const char * format, bool streamRemote, unsigned _expectedCrc, IOutputMetaData & _expected, unsigned _projectedCrc, IOutputMetaData & _projected, unsigned _actualCrc, IOutputMetaData & _actual, const IPropertyTree * options) override;
+
+// IThorDiskCallback
+    virtual offset_t getFilePosition(const void * row) override;
+    virtual offset_t getLocalFilePosition(const void * row) override;
+    virtual const char * queryLogicalFilename(const void * row) override;
+    virtual const byte * lookupBlob(unsigned __int64 id) override { UNIMPLEMENTED; }
+
+
+protected:
+    offset_t getLocalOffset();
+
+protected:
+    Owned<ISerialStream> input;
+    Owned<IFileIO> inputfileio;
+    CThorContiguousRowBuffer inputBuffer;
+    Owned<IEngineRowAllocator> outputAllocator;
+    Owned<const IDynamicTransform> translator;
+    Owned<const IKeyTranslator> keyedTranslator;
+    Linked<IOutputMetaData> expectedDiskMeta = nullptr;
+    Linked<IOutputMetaData> projectedDiskMeta = nullptr;
+    Linked<IOutputMetaData> actualDiskMeta = nullptr;
+    unsigned expectedCrc = 0;
+    unsigned projectedCrc = 0;
+    unsigned actualCrc = 0;
+    MemoryBuffer encryptionKey;
+    size32_t readBufferSize = defaultReadBufferSize;
+    bool grouped = false;
+    bool stranded = false;
+    bool compressed = false;
+    bool blockcompressed = false;
+    bool rowcompressed = false;
+
+//The following refer to the current input file:
+    offset_t fileBaseOffset = 0;
+    StringAttr logicalFilename;
+    unsigned filePart = 0;
+};
+
+
+DiskRowReader::DiskRowReader(unsigned _expectedCrc, IOutputMetaData & _expected, unsigned _projectedCrc, IOutputMetaData & _projected, unsigned _actualCrc, IOutputMetaData & _actual, const IPropertyTree * options)
+: expectedCrc(_expectedCrc), expectedDiskMeta(&_expected), projectedCrc(_projectedCrc), projectedDiskMeta(&_projected), actualCrc(_actualCrc), actualDiskMeta(&_actual)
+{
+    assertex(options);
+
+    //Not sure this should really be being passed in here...
+    RecordTranslationMode translationMode = (RecordTranslationMode)options->getPropInt("translationMode", (int)RecordTranslationMode::All);
+    //MORE: HPCC-22287 This is too late to be able to do something different if the keyed filters cannot be translated.
+    getTranslators(translator, keyedTranslator, "BinaryDiskRowReader", expectedCrc, expectedDiskMeta, actualCrc, actualDiskMeta, projectedCrc, projectedDiskMeta, translationMode);
+
+    //Options contain information that is the same for each file that is being read, and potentially expensive to reconfigure.
+    if (options->hasProp("encryptionKey"))
+    {
+        encryptionKey.resetBuffer();
+        options->getPropBin("encryptionKey", encryptionKey);
+    }
+    readBufferSize = options->getPropInt("readBufferSize", defaultReadBufferSize);
+}
+
+IRawRowStream * DiskRowReader::queryRawRowStream()
+{
+    return this;
+}
+
+IAllocRowStream * DiskRowReader::queryAllocatedRowStream(IEngineRowAllocator * _outputAllocator)
+{
+    outputAllocator.set(_outputAllocator);
+    return this;
+}
+
+void DiskRowReader::clearInput()
+{
+    inputBuffer.setStream(nullptr);
+    input.clear();
+}
+
+bool DiskRowReader::matches(const char * format, bool streamRemote, unsigned _expectedCrc, IOutputMetaData & _expected, unsigned _projectedCrc, IOutputMetaData & _projected, unsigned _actualCrc, IOutputMetaData & _actual, const IPropertyTree * options)
+{
+    if ((expectedCrc != _expectedCrc) || (projectedCrc != _projectedCrc) || (actualCrc != _actualCrc))
+        return false;
+
+    //MORE: Check translation mode
+
+    //MORE: Is the previous check sufficient?  If not, once getDaliLayoutInfo is cached the following line could be enabled.
+    //if ((expectedDiskMeta != &_expected) || (projectedDiskMeta != &_projected) || (actualDiskMeta != &_actual))
+    //    return false;
+
+    if (options->hasProp("encryptionKey"))
+    {
+        MemoryBuffer tempEncryptionKey;
+        options->getPropBin("encryptionKey", tempEncryptionKey);
+        if (!encryptionKey.matches(tempEncryptionKey))
+            return false;
+    }
+    if (readBufferSize != options->getPropInt("readBufferSize", defaultReadBufferSize))
+        return false;
+    return true;
+}
+
+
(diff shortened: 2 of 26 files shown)


ghalliday:
@jakesmith for discussion - particularly the end of group rows, splitting code from hthor to a reader class and various problems that introduces/highlights.

common/thorhelper/thorread.cpp
```diff
@@ -0,0 +1,510 @@
...
+    offset_t fileBaseOffset = 0;
+    StringAttr logicalFilename;
+    unsigned filePart = 0;
+};
+
+
+DiskRowReader::DiskRowReader(IOutputMetaData & expected, IOutputMetaData & projected, IOutputMetaData & actual, IPropertyTree * options)
+{
+    compressed = options->getPropBool("compressed", compressed);
+    blockcompressed = options->getPropBool("blockCompressed", blockcompressed);
```
jakesmith: do we need both?

common/thorhelper/thorread.cpp
```diff
@@ -0,0 +1,510 @@
...
+
+    return (filesize != 0);
+}
+
+//Implementation of IRowStream
+const void *BinaryDiskRowReader::nextRow()
+{
+    for (;;)
+    {
+        //This may return multiple eog in a row with no intervening records.  It is up to the caller to filter.
```
jakesmith: how?

ecl/hthor/hthor.cpp
```diff
@@ -10358,6 +10359,597 @@ void CHThorExternalActivity::stop()
...
+    {
+        StringBuffer mangledFilename;
+        mangleLocalTempFilename(mangledFilename, mangledHelperFileName.str());
+        tempFileName.set(agent.queryTemporaryFile(mangledFilename.str()));
+        logicalFileName = tempFileName.str();
+        gatherInfo(NULL);
+    }
+    else
+    {
+        ldFile.setown(resolveLFNFlat(agent, mangledHelperFileName.str(), "Read", 0 != (helper.getFlags() & TDRoptional)));
```
jakesmith: perhaps not new, but wondering what resolveLFNFlat does vs other types of files?

ecl/hthor/hthor.cpp
```diff
@@ -10358,6 +10359,597 @@ void CHThorExternalActivity::stop()
...
+        if ( mangledHelperFileName.charAt(0) == '~')
+            logicalFileName = mangledHelperFileName.str()+1;
+        else
+            logicalFileName = mangledHelperFileName.str();
```
jakesmith: suspect mangleHelperFileName should do this.

ecl/hthor/hthor.cpp
```diff
@@ -10358,6 +10359,597 @@ void CHThorExternalActivity::stop()
...
+{
+    if (useRawStream)
+        rawRowStream = queryNullRawRowStream();
+    else
+        roxieRowStream = queryNullRoxieRowStream();
+}
+
+bool CHThorNewDiskReadBaseActivity::openFilePart(ILocalOrDistributedFile * localFile, IDistributedFilePart * filePart, unsigned whichPart)
+{
+    unsigned numCopies = filePart?filePart->numCopies():ldFile->numPartCopies(partNum);
```
jakesmith: if no filePart, does ldFile->numPartCopies() ever equal anything other than 1?

common/thorhelper/thorread.cpp
```diff
@@ -0,0 +1,510 @@
...
+                return outputAllocator->finalizeRow(sizeRead, result, allocatedSize);
+            }
+        }
+    }
+}
+
+
+//Implementation of IRawRowStream
+const void *BinaryDiskRowReader::nextRow(size32_t & size)
+{
```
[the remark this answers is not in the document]
ghalliday: correct it would also return eog.  I'll remove the comment from the header.

common/thorhelper/thorread.cpp
```diff
@@ -0,0 +1,510 @@
...
+#include "eclrtl_imp.hpp"
+#include "rtlfield.hpp"
+#include "rtlds_imp.hpp"
+#include "rtldynfield.hpp"
+
+#include "thorread.hpp"
+#include "rtlcommon.hpp"
+#include "thorcommon.hpp"
+
+class DiskRowReader : implements IRowStream, implements IRawRowStream, implements IDiskRowReader, implements IThorDiskCallback, public CInterface
```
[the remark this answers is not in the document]
ghalliday: cannot be because there are multiple base interfaces that need to be implemented.

ghalliday:
@richardkchapman @jakesmith ready for a first review.  A few items that stand out for discussion:

* Alien data types cause some problems (imgkey.ecl).  If they cannot be processed by the tanslation layer then the read activity still needs to support fileposition callbacks, which might be a pain.
* Should the remote filter be a filter on the expected filter, or the actual filter?
* The code may be cleaner with a struct to encapsulate the parameters to the reader, especially if more are needed (e.g., "can filter" may be needed for the csv/xml implementations?).
* Do we need to pass a few more flags - e.g., translation mode? to the remote dafilesrv, or should that all be done client side.

common/thorhelper/thorread.cpp
```diff
@@ -0,0 +1,825 @@
...
+class DiskRowReader : implements IAllocRowStream, implements IRawRowStream, implements IDiskRowReader, implements IThorDiskCallback, public CInterface
+{
+public:
+    DiskRowReader(unsigned _expectedCrc, IOutputMetaData & _expected, unsigned _projectedCrc, IOutputMetaData & _projected, unsigned _actualCrc, IOutputMetaData & _actual, IPropertyTree * options);
+    IMPLEMENT_IINTERFACE
+
+    virtual IRawRowStream * queryRawRowStream() override;
+    virtual IAllocRowStream * queryAllocatedRowStream(IEngineRowAllocator * _outputAllocator) override;
+
+    virtual bool getCursor(MemoryBuffer & cursor) override;
```
jakesmith: can be const method?

common/thorhelper/thorread.cpp
```diff
@@ -0,0 +1,825 @@
...
+    Owned<ISerialStream> input;
+    Owned<IFileIO> inputfileio;
+    CThorContiguousRowBuffer inputBuffer;
+    Owned<IEngineRowAllocator> outputAllocator;
+    IOutputMetaData * expectedDiskMeta = nullptr;
+    IOutputMetaData * projectedDiskMeta = nullptr;
+    IOutputMetaData * actualDiskMeta = nullptr;
+    unsigned expectedCrc;
+    unsigned projectedCrc;
+    unsigned actualCrc;
```
jakesmith: coding style question: should be initialize all members as a matter of good practice, even if they are also initialize in ctor.
I think for members that have ctor initializers - compiler guarantees won't do twice.

common/thorhelper/thorread.cpp
```diff
@@ -0,0 +1,825 @@
...
+}
+
+IRawRowStream * DiskRowReader::queryRawRowStream()
+{
+    return this;
+}
+
+IAllocRowStream * DiskRowReader::queryAllocatedRowStream(IEngineRowAllocator * _outputAllocator)
+{
+    outputAllocator.set(_outputAllocator);
```
jakesmith: feels like it's doing more than method name suggests (queryAllocatedRowStream()).
Should row allocate be set via a different method instead?

common/thorhelper/thorread.cpp
```diff
@@ -0,0 +1,825 @@
...
+
+void DiskRowReader::clearInput()
+{
+    inputBuffer.setStream(nullptr);
+    input.clear();
+}
+
+bool metaMatches(IOutputMetaData * left, IOutputMetaData * right)
+{
+    return left == right;
```
jakesmith: should this do further checking ?

common/thorhelper/thorread.cpp
```diff
@@ -0,0 +1,825 @@
...
+
+    //Is the previous check sufficient?
+    //if (!metaMatches(expectedDiskMeta, &_expected) || !metaMatches(projectedDiskMeta, &_projected) || !metaMatches(actualDiskMeta, &_actual))
+    //    return false;
+
+    if (options->hasProp("encryptionKey"))
+    {
+        MemoryBuffer tempEncryptionKey;
+        options->getPropBin("encryptionKey", tempEncryptionKey);
+        if (!encryptionKey.matches(tempEncryptionKey))
```
jakesmith: I'm not clear under what circumstances this would happen.. should it throw an exception if mismatches?

common/thorhelper/thorread.cpp
```diff
@@ -0,0 +1,825 @@
...
+    return LocalDiskRowReader::matches(format, streamRemote, _expectedCrc, _expected, _projectedCrc, _projected, _actualCrc, _actual, options);
+}
+
+bool BinaryDiskRowReader::setInputFile(IFile * inputFile, const char * _logicalFilename, unsigned _partNumber, offset_t _baseOffset, IPropertyTree * meta, const FieldFilterArray & expectedFilter)
+{
+    if (!LocalDiskRowReader::setInputFile(inputFile, _logicalFilename, _partNumber, _baseOffset, meta, expectedFilter))
+        return false;
+
+    actualFilter.clear();
+    actualFilter.appendFilters(expectedFilter);
```
jakesmith: just to check, actual=expect filter here, because rows will already have been translated by time actualFilter used in this context, right?

ghalliday (commented):
Forgot to submit comments.  Removed draft status.

common/thorhelper/thorcommon.cpp
```diff
@@ -2097,6 +2098,7 @@ static bool getTranslators(Owned<const IDynamicTransform> &translator, Owned<con
                 if (keyedTranslator && (sourceFormat != expectedFormat))
                 {
                     Owned<const IKeyTranslator> _keyedTranslator = createKeyTranslator(sourceFormat->queryRecordAccessor(true), expectedFormat->queryRecordAccessor(true));
+                    //MORE: What happens if the key filters cannot be translated?
```
richardkchapman: Has this question been answered? Is there a Jira to revisit it?

ecl/hqlcpp/hqlsource.cpp
```diff
@@ -1135,14 +1137,17 @@ void SourceBuilder::rebindFilepositons(BuildCtx & ctx, IHqlExpression * dataset,
...
-        ctx.associateExpr(selectorFpos, match->queryExpr());
+        ctx.associateOwn(*new MonitoredDefinedValue(transformUsesVirtualFilePosition, selectorFpos, match->queryExpr()));
     }
 }
 
 
 void SourceBuilder::rebindFilepositons(BuildCtx & ctx, IHqlExpression * dataset, node_operator side, IHqlExpression * selSeq)
 {
-    bool savedIsVirtualLogicalFilenameUsed = isVirtualLogicalFilenameUsed;  // don't allow the rebinding to set the flag.
+    // don't allow the rebinding to modify these flags.
```
richardkchapman: This code looks a bit iffy. Wouldn't just passing a dummy local bool on line 1159 have the same effect?

ecl/hthor/hthor.cpp
```diff
@@ -10363,6 +10364,681 @@ void CHThorExternalActivity::stop()
...
+{
+    Owned<IPropertyTree> meta = createPTree();
+    unsigned actualCrc = helper.getDiskFormatCrc();
+    Linked<IOutputMetaData> actualDiskMeta = expectedDiskMeta;
+    bool compressed = false;
+    bool blockcompressed = false;
+
+    if (distributedFile)
+    {
+        //MORE:
```
richardkchapman: Not clear what this comment is telling me... is it still valid (and if so is there a JIRA to cover the additional work that is needed?)

ecl/hthor/hthor.cpp
```diff
@@ -10363,6 +10364,681 @@ void CHThorExternalActivity::stop()
...
+        size32_t kl;
+        helper.getEncryptKey(kl,k);
+        if (kl)
+        {
+            meta->setPropBin("encryptionKey", kl, k);
+            blockcompressed = true;
+            compressed = true;
+        }
+    }
+    else
```
richardkchapman: Do we need an empty else here? Should there be a comment in the else saying "No need to do X because Y"

ecl/hthor/hthor.cpp
```diff
@@ -10363,6 +10364,681 @@ void CHThorExternalActivity::stop()
...
+    if (curPart)
+        curPart->getFilename(rfilename,copy);
+    else
+        localFile->getPartFilename(rfilename,partNum,copy);
+}
+
+bool CHThorNewDiskReadBaseActivity::openFirstPart()
+{
+    partNum = 0;
+    if (dfsParts)       // more should really be fileDesc or something
```
richardkchapman: So why isn't it?

ecl/hthor/hthor.cpp
```diff
@@ -10363,6 +10364,681 @@ void CHThorExternalActivity::stop()
...
+            return &cur;
+    }
+    IDiskRowReader * reader = createDiskReader(format, streamRemote, expectedCrc, expected, projectedCrc, projected, actualCrc, actual, options);
+    readers.append(*reader);
+    return reader;
+}
+
+bool CHThorNewDiskReadBaseActivity::openFilePart(const char * filename)
+{
+    const char * format = "thor";   // more - should extract from the current file (could even mix flat and csv...)
```
richardkchapman: more?

ecl/hthor/hthor.cpp
```diff
@@ -10363,6 +10364,681 @@ void CHThorExternalActivity::stop()
...
+
+    bool tryRemoteStream = actualDiskMeta->queryTypeInfo()->canInterpret() && actualDiskMeta->queryTypeInfo()->canSerialize() &&
+                           projectedDiskMeta->queryTypeInfo()->canInterpret() && projectedDiskMeta->queryTypeInfo()->canSerialize();
+
+    /*
+     * If a file part can be accessed local, then read it locally
+     * If a file part supports a remote stream, then use that
+     * Otherwise failover to the legacy remote access.
+     */
+    const char * format = "thor";   // more - should extract from the current file (could even mix flat and csv...)
```
richardkchapman: ...

system/jlib/jrowstream.cpp
```diff
@@ -0,0 +1,68 @@
+/*##############################################################################
+
+    HPCC SYSTEMS software Copyright (C) 2012 HPCC Systems®.
```
richardkchapman: Is it still 2012? I wish it was...

system/jlib/jrowstream.hpp
```diff
@@ -0,0 +1,89 @@
+/*##############################################################################
+
+    HPCC SYSTEMS software Copyright (C) 2018 HPCC Systems®.
```
richardkchapman: Closer...

testing/regress/ecl/translatedisk.ecl
```diff
@@ -17,15 +17,18 @@
 
 //version multiPart=false
 //version multiPart=true
+//version multiPart=true,optRemoteRead=true
```
richardkchapman: Should there be some variants that force the old activity ?

common/thorhelper/thorread.cpp
```diff
@@ -0,0 +1,806 @@
+/*##############################################################################
+
+    HPCC SYSTEMS software Copyright (C) 2012 HPCC Systems®.
```
richardkchapman: 2019

common/thorhelper/thorread.cpp
```diff
@@ -0,0 +1,825 @@
...
+
+//---------------------------------------------------------------------------------------------------------------------
+
+static Owned<IPropertyTree> emptyTree = createPTree();
+
+//---------------------------------------------------------------------------------------------------------------------
+
+constexpr size32_t defaultReadBufferSize = 0x10000;
+
+//Cannot use InterfaceOf<> because multiple base interfaces are derived from IInterface
```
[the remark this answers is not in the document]
richardkchapman: Should you delete the comment then?

common/thorhelper/thorcommon.cpp
ghalliday: There is a jira HPCC-22287.  Currently we don't spot potential problems.

common/thorhelper/thorread.cpp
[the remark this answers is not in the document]
ghalliday: done

common/thorhelper/thorread.cpp
ghalliday: the current implementation calls a non const function.  Something to return to, but not change at the moment.

ecl/hqlcpp/hqlsource.cpp
ghalliday: No, that does need be the correct value.  The problem is that the act of looking up what the fileposition maps to sets the flag - but we only want it to be set if the generated code looks up the fileposition.

ecl/hthor/hthor.cpp
ghalliday: This probably does need addressing.  I think it is where it is reading a binary file that was sprayed as csv - don't use the structure it was sprayed as (see lines 8080 in original activity).

ecl/hthor/hthor.cpp
ghalliday: I'll look at it.

ecl/hthor/hthor.cpp
ghalliday: At the moment this activity is only used for binary reads - in the future the same activity will be used for binary/csv/xml reads

testing/regress/ecl/translatedisk.ecl
ghalliday: imgkey.ecl will force the old activity.

ghalliday (commented):
Some replies.  Need to revisit the comments about reading csv files as binary

ghalliday:
@richardkchapman squashed, rebased.  I have also verified that setting
  #option ('useNewDiskReadActivity', false);
Avoids using the new code.

PR #12563, opened 2019-05-16
HPCC-21919 Right-justify numeric values in ECLWatch

kunalaswani:
All numeric values are now right justified.

Signed-off-by: Aswani, Kunal <kunal.aswani@lexisnexis.com>

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization
  - [x] I have tested my changes in multiple modern browsers

## Smoketest:

### DIFF
--- a/esp/src/eclwatch/css/hpcc.css
+++ b/esp/src/eclwatch/css/hpcc.css
@@ -260,6 +260,10 @@ hr.dashedLine {
     white-space: normal; 
 }
 
+.justify-right {
+    text-align: right !important;
+}
+
 .flat .hpccCentered { 
     text-align: center; 
 }
--- a/esp/src/src/Utility.ts
+++ b/esp/src/src/Utility.ts
@@ -107,6 +107,14 @@ export function convertedSize(intsize: number): string {
     }
 }
 
+export function valueCleanUp(intsize): string {
+    if (intsize === null || intsize === undefined) {
+        return "";
+    } else {
+        return intsize;
+    }
+}
+
 export function unitTest(size, unit) {
     var nsIndex = size.indexOf(unit);
     if (nsIndex !== -1) {
--- a/esp/src/eclwatch/DFUQueryWidget.js
+++ b/esp/src/eclwatch/DFUQueryWidget.js
@@ -658,14 +658,27 @@ define([
                         SuperOwners: { label: this.i18n.SuperOwner, width: 150 },
                         Description: { label: this.i18n.Description, width: 150 },
                         NodeGroup: { label: this.i18n.Cluster, width: 108 },
-                        RecordCount: { label: this.i18n.Records, width: 85 },
+                        RecordCount: {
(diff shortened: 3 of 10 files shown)


kunalaswani:
@GordonSmith  please review,
Utility.ts - new function added, to help ensure values don't have stray undefined or invalid values.
Addressed all files with numeric values in grids.

esp/src/eclwatch/css/hpcc.css
```diff
@@ -260,6 +260,10 @@ hr.dashedLine {
     white-space: normal; 
 }
 
+.justify-right {
+    text-align: right !important;
```
GordonSmith: Is the `!important` needed?  In general it should only be used as a last measure (adding a more specific CSS selector may be a better approach).

esp/src/src/Utility.ts
```diff
@@ -99,14 +99,22 @@ export function espTime2SecondsTests() {
 
 export function convertedSize(intsize: number): string {
     const unitConversion = ["Bytes", "KB", "MB", "GB", "TB", "PB", "EB", "ZB", "YB"];
-    if (intsize === null || intsize === undefined) {
+    if (intsize === null || intsize === undefined || intsize === -1) {
```
GordonSmith: Is the "-1" safe for all calls to this function?  It might be better to have an additional param option `treatMinusOneAsUndefined`?
Or have the -1 === undefined logic closer to where the data comes back from the server or in the grid formatter?

GordonSmith (changes requested):

esp/src/eclwatch/css/hpcc.css
kunalaswani: Yes, we are using the "!important" because the parent class overwrites this every time otherwise.

GordonSmith: You might be able to "win" with better CSS specificity?

kunalaswani:
@GordonSmith  changes made, added two more files where right justify was needed.

kunalaswani:
@GordonSmith after explanation of the use of the "!important" Flag, does this still need to be changed or is it ok?

esp/src/src/Utility.ts
```diff
@@ -107,6 +107,14 @@ export function convertedSize(intsize: number): string {
     }
 }
 
+export function valueCleanUp(intsize): string {
+    if (intsize === null || intsize === undefined || intsize === -1) {
```
GordonSmith: Should the test for "-1" be here now?

kunalaswani: I have removed the "-1" from here and have checked pages to make sure it is not needed, where it is needed I have added a check for it.

GordonSmith (approved):
Looks good.

PR #12586, opened 2019-05-23
HPCC-22202 Set "local" attribute on hthor and roxie built indexes

jakesmith:
.. which are always 1 part and by definition local (not distributed)

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/ecl/hthor/hthor.cpp
+++ b/ecl/hthor/hthor.cpp
@@ -1223,6 +1223,7 @@ void CHThorIndexWriteActivity::execute()
     // properties of the logical file
     IPropertyTree & properties = desc->queryProperties();
     properties.setProp("@kind", "key");
+    properties.setPropBool("@local", true);
     properties.setPropInt64("@size", indexFileSize);
     properties.setPropInt64("@recordCount", reccount);
     properties.setProp("@owner", agent.queryWorkUnit()->queryUser());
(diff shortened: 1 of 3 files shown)


jakesmith:
@ghalliday - please check.

richardkchapman:
You can argue that a 1-part index is distributed (1-way).

What is the benefit of marking them local?

jakesmith:
It's more that that they're inconsistent between engines.
Could mark Thor's 1-way indexes as 'distributed' too.

Not sure (should check) what hthor does if asked to create , LOCAL index - which would be exactly the same as it's current default "distributed" 1-way indexes.

richardkchapman:
Also what roxie/hthor do when they encounter an index that is marked both local and 1-way for reading.

ghalliday:
@jakesmith I think waiting for some answers from you

jakesmith:
>Also what roxie/hthor do when they encounter an index that is marked both local and 1-way for reading.

Afaics, after this change they'd be treated in exactly same way as if it was a 1-way without @local.

jakesmith:
>Not sure (should check) what hthor does if asked to create , LOCAL index - which would be exactly the same as it's current default "distributed" 1-way indexes.

Neither hthor or Roxie (prior to this change) used to set the @local attribute under any circumstances, i.e. the only thing that does is Thor when ECL tell it to create a local key.

>You can argue that a 1-part index is distributed (1-way).

So that's the alternative, since hthor and roxie do not mark 1-way (i.e. all) index as local, we could instead change it so that 1-part (,FEW) indexes created by Thor don't either.

Still feels like 1-way's (without TLK) should be "local".. I think that is most logical/consistent.

jakesmith:
@richardkchapman - apart from fixing the inconsistency issue (or because it does), it means you can mix hthor and thor 1 part index in a super.
Without this change you get an error when adding mixed subfiles (some "local", some not).

jakesmith:
Closing to replace with: https://track.hpccsystems.com/browse/HPCC-22653

PR #12592, opened 2019-05-28
HPCC-22030 Add white list mechanism to Dali

jakesmith:
NB: also adjust a few regression suite queries, that tried
to access Dali from thor slaves, by from within the query,
enabling 'slaveDaliClient'.
Which now, even when enabled would give 'Access denied' by
default via white list mechanism.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dali/base/dasds.cpp
+++ b/dali/base/dasds.cpp
@@ -6477,14 +6477,11 @@ void CCovenSDSManager::saveDelta(const char *path, IPropertyTree &changeTree)
     if (externalEnvironment)
     {
         // don't save any changed to /Environment if external
-        if (0 == strncmp("/Environment", path, strlen("/Environment")))
+
+        if (startsWith(path, "/Environment") || (streq(path, "/") && changeTree.hasProp("*[@name=\"Environment\"]")))
         {
-            OWARNLOG("Attempt to change read-only Dali environment, path = %s", path);
-            return;
-        }
-        if (0 == strcmp("/", path) && changeTree.hasProp("*[@name=\"Environment\"]"))
-        {
-            OWARNLOG("Attempt to change read-only Dali environment, path = %s", path);
+            querySessionManager().refreshWhiteList();
+            PROGLOG("Dali Environment updated, path = %s", path);
             return;
         }
     }
--- a/dali/base/dasess.cpp
+++ b/dali/base/dasess.cpp
@@ -16,6 +16,10 @@
 ############################################################################## */
 
 #define da_decl DECL_EXPORT
+
+#include <unordered_map>
+#include <unordered_set>
+
 #include "platform.h"
 #include "jlib.hpp"
 #include "jfile.hpp"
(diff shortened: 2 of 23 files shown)


jakesmith:

dali/base/dasess.cpp
```diff
@@ -41,32 +45,62 @@ using namespace cryptohelper;
 #pragma warning (disable : 4355)
 #endif
 
+static std::unordered_map<std::string, DaliClientRole> daliClientRoleMap = {
```
richardkchapman: Is there not such a thing as a bidirectional map?

richardkchapman:
Couple of general questions:

1. Will all of this break if we use hostnames or bring up machines dynamically? Will we need to update the dali whitelist whenever bringing up a machine dynamically (e.g. in a kubernetes scenario)
2. Do we need support for whitelisting ranges of IPs?
3. Is it possible to disable the functionality (i.e allow connections from anywhere) as an alternative to rollback if unforeseen issues arise?
4. Is it worth having a mode where we log but do not reject any connection that is not accepted by the whitelist, as a way of checking that enabling the whitelist is not going to cause issues when first deploying.

richardkchapman:
I didn;t spot anything glaring but @mckellyln and @ghalliday should look too

jakesmith:
>Will all of this break if we use hostnames or bring up machines dynamically? Will we need to update the dali whitelist whenever bringing up a machine dynamically (e.g. in a kubernetes scenario)

The list is build from the environment, as it's build it binds to IP's.
If the IP for a hostname changes, then the whitelist will need updating.
If the environment changes (or is pushed), the whitelist is rebuild, so if hostame IP's change we'd currently have to have something request an update.

> Will we need to update the dali whitelist whenever bringing up a machine dynamically (e.g. in a kubernetes scenario)

So, yes, a authorized node with updtdalienv would need to update environment Whitelist for a new host and push update to Dali.

> Do we need support for whitelisting ranges of IPs?

Could do, but would make the lookups a bit more painful - would need some kind of search, currently it's a hash lookup.
At the moment (apart from auto-populated component/roles), you can add a csv list of hosts, with a csv list of roles, e.g.:
```
<WhiteList>
 <Entry hosts="host1,host2,host3" roles="daliadmin,dalidiag,swapnode"/>
 <Entry hosts="hostA,hostB,hostC" roles="thormaster,eclccserver,xref"/>
</WhiteList>
```
> Is it possible to disable the functionality (i.e allow connections from anywhere) as an alternative to rollback if unforeseen issues arise?

Yes, @enabled="false" in \<WhiteList\> disables.

> Is it worth having a mode where we log but do not reject any connection that is not accepted by the whitelist, as a way of checking that enabling the whitelist is not going to cause issues when first deploying.

With \<WhiteList enabled="false"\> - it will still log 'unauthorized' connections.

dali/base/dasess.cpp
jakesmith: Not a STL one, I think there might be a boost impl.

dali/base/dasds.cpp
```diff
@@ -6477,14 +6477,11 @@ void CCovenSDSManager::saveDelta(const char *path, IPropertyTree &changeTree)
...
-            OWARNLOG("Attempt to change read-only Dali environment, path = %s", path);
-            return;
-        }
-        if (0 == strcmp("/", path) && changeTree.hasProp("*[@name=\"Environment\"]"))
+
+        if (startsWith(path, "/Environment") || (streq(path, "/") && changeTree.hasProp("*[@name=\"Environment\"]")))
         {
-            OWARNLOG("Attempt to change read-only Dali environment, path = %s", path);
+            querySessionManager().refreshWhiteList();
+            PROGLOG("Dali Environment updated, path = %s", path);
```
mckellyln: Why drop the read-only message ?  Because the whitelist is updated now ?

jakesmith: It wasn't correct before. It is valid to update the /Environment via some routes, e.g. updtdalienv - so the warning was misleading.

mckellyln:
@jakesmith this looks good.  I do not completely follow the xor hash combine code yet but otherwise looks very good.

jakesmith:
@ghalliday - reviewed by Richard and Mark, can you give it a once over before merging?

dali/base/dasess.cpp
```diff
@@ -1118,6 +1152,16 @@ class CClientSessionManager: public CSessionManagerBase, implements IConnectionM
         queryCoven().sendRecv(mb,RANK_RANDOM,MPTAG_DALI_SESSION_REQUEST,SESSIONREPLYTIMEOUT);
     }
 
+    virtual void refreshWhiteList() override
+    {
+        assertex(!"refreshWhiteList called on client");
```
ghalliday: minor: more recently throwUnexpected("refreshWhiteList called on client"); tends to be how this would be coded

dali/base/dasess.cpp
```diff
@@ -1267,9 +1311,246 @@ class CLdapWorkItem : public Thread
...
+        for (auto &it: whiteList)
+            out.append(it.first.c_str()).append(", ").append(queryRoleName(it.second)).append("\n");
+        return out;
+    }
+    void refresh()
+    {
+        /* NB: clear only, so that next usage will re-populated
+         * Do not want to repopulate now, because refresh() is likely called within a update write transaction
+         */
+        CriticalBlock block(populatedCrit);
```
ghalliday: How common is it likely to be that refresh will block the whitelisting checks?  If common you could use a different approach to reduce the scope of the critical section.

ghalliday (commented):
@jakesmith a couple of minor questions/comments.

dali/base/dasess.cpp
jakesmith: yeah agree, but would then have wanted to change all the other cases in this file and decided against it at this stage.

Will change this and others in this file now though.

dali/base/dasess.cpp
jakesmith: I do not think it will be refreshing often, or that client registration time will be so rapid and frequent that it will be an issue.

jakesmith:
@ghalliday - have changed the assertex(!".. to throwUnexpectedX, and squashed in.

wangkx:
Would the change of the ServerVersion here (from 3.15 to 3.2) break backward compatibility (please see the code in CDaliVersion::compare())?

jakesmith:
@wangkx - yes it did cause problem, fixed by: https://track.hpccsystems.com/browse/HPCC-22352

PR #12594, opened 2019-05-28
HPCC-22231 Analyse workunit stats for distribute output row skews

shamser:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:


shamser:
@ghalliday

common/wuanalysis/anarule.cpp
```diff
@@ -0,0 +1,81 @@
...
+        if (!outputEdge)
+            return false;
+
+        stat_t rowsAvg = outputEdge->getStatRaw(StNumRowsProcessed, StAvgX);
+        if (rowsAvg < rowsThreshold)
+            return false;
+        stat_t rowsMaxSkew = outputEdge->getStatRaw(StNumRowsProcessed, StSkewMax);
+        if (rowsMaxSkew > options.skewThreshold)
+        {
+            // Use upstream activity time to calculate approximate cost
```
ghalliday: I think the output should be downstream rather than upstream

common/wuanalysis/anarule.cpp
```diff
@@ -0,0 +1,81 @@
...
+            IWuActivity * targetActivity = outputEdge->queryTarget();
+            assertex(targetActivity);
+            stat_t timeMaxLocalExecute = targetActivity->getStatRaw(StTimeLocalExecute, StMaxX);
+            stat_t timeMinLocalExecute = targetActivity->getStatRaw(StTimeLocalExecute, StMinX);
+            __uint64 cost = (timeMaxLocalExecute - timeMinLocalExecute) / (rowsMaxSkew / 10000);
```
ghalliday: I'm not sure about this calculation.  I think this should be subtracting the average time from the max time, and not dividing it by anything.  My reasoning: assuming the timing is linear with the number of row then if there was no skew it would execute in the average time.  So the extra cost is the max time minus the average time.

If you're using the max and min time, then you should be using the min and max skews somehow in your calculation.

Unfortunately I think this is one of the hardest rules to come up with a cost for, since it is hard to know what effect it has on the downstream activities.

ghalliday: You're going to have rounding issues.  rowMaxSkew/10000 will likely evaluate to 1.  Better to multiply by 10000 and then divide by rowsMaxSkew.

common/wuanalysis/anawu.cpp
```diff
@@ -0,0 +1,429 @@
...
+    for (auto & cur : scopes)
+        toWalk.append(cur);
+
+    ForEachItemIn(i, toWalk)
+    {
+        WuScope & cur = toWalk.item(i);
+
+        if (cur.queryScopeType() == SSTedge)
+        {
+            const char * tmp = attrs->queryProp("@IdSource");
```
ghalliday: Looks like old debugging code

common/wuanalysis/anawu.cpp
```diff
@@ -0,0 +1,429 @@
...
+
+//-----------------------------------------------------------------------------------------------------------
+WorkunitAnalyser::WorkunitAnalyser(WuAnalyseOptions & _options) : root("", nullptr), options(_options)
+{
+    gatherRules(rules);
+}
+
+void WorkunitAnalyser::check(const char * scope, IWuActivity & activity)
+{
+    if (activity.queryThorActivityKind()!=TAKhashdistribute)
```
ghalliday: This test is a bit arbitrary - it suggests the minimum time test should be done somewhere else, or in a different way.

common/wuanalysis/anawu.cpp
```diff
@@ -0,0 +1,429 @@
...
+        }
+        catch (IException * e)
+        {
+            e->Release();
+        }
+    }
+}
+
+// Only keep highest cost result per scope
+void WorkunitAnalyser::limitResultsPerScope()
```
ghalliday: This logic could probably be moved to WorkunitAnalyser::check - comparing the cost with previous entries, and then only adding one at the end.

common/wuanalysis/anawu.cpp
```diff
@@ -0,0 +1,429 @@
...
+{
+    issues.sort(compareIssuesScopeOrder);
+
+    int prevIssue = 0;
```
ghalliday: minor: unsigned probably best

common/wuanalysis/anawu.cpp
```diff
@@ -0,0 +1,429 @@
...
+    int ordinality = issues.ordinality();
+    int indx = 1;
+    while (indx < ordinality)
+    {
+        if (issues.item(indx).compareScope(issues.item(prevIssue))==0)   // Same scope so remove one
+        {
+            if (issues.item(indx).compareCost(issues.item(prevIssue))<0) // remove the lower cost of the 2 issues
+            {
+                prevIssue = indx;
```
ghalliday: I think this is wrong - since indx-1 is removed - so the current indx item will now be at indx-1.  prevIssue is always indx-1

common/wuanalysis/anawu.cpp
```diff
@@ -0,0 +1,429 @@
...
+        {
+            e->Release();
+        }
+    }
+}
+
+// Only keep highest cost result per scope
+void WorkunitAnalyser::limitResultsPerScope()
+{
+    issues.sort(compareIssuesScopeOrder);
```
ghalliday: Following code could be simpler if the sort was by scope and cost, but simpler still if done above.

system/jlib/jstatcodes.h
```diff
@@ -251,4 +251,5 @@ enum StatisticKind
 };
 constexpr StatisticKind operator |(StatisticKind l, StatisticKind r) { return (StatisticKind)((unsigned)l | (unsigned)r); }
 
+typedef __int64 stat_t;
```
ghalliday: This should be unsigned.

ghalliday:
@shamser a few comments.  Also please can you link the jira to the main issue.

ghalliday:
This also needs an example that provokes the warning.

system/jlib/jstatcodes.h
shamser: Some statistics (such as SkewMinRowsProcessed) are negative.

common/wuanalysis/anarule.cpp
shamser: Your cost calculation formation (max execute - average time) seems good enough for now.  I've noted in the comments that we should look at ways of improving the cost calculation.

shamser: No longer relevant with the change in formula.

common/wuanalysis/anawu.cpp
shamser: Yes, this slipped in by mistake...removed.

common/wuanalysis/anawu.cpp
shamser: The time taken to distribute does not seem relevant to skewed distribution rows.  We could check the minimum time anyway for hash distributes.  What do you think?

common/wuanalysis/anawu.cpp
shamser: i was thinking that at some point the number of issues per activity reported would be configurable.  However, I have simplified the code as suggested and moved the logic to ::check()

common/wuanalysis/anawu.cpp
shamser: Function removed. No longer relevant.

common/wuanalysis/anawu.cpp
shamser: Function removed. No longer relevant.

common/wuanalysis/anawu.cpp
shamser: Function removed. No longer relevant.

shamser:
@ghalliday None of the performance suite ecl queries is generating enough skew & cost to generate a issue report.  Where should I provide the example?  Please see the changes & comments on the review.

common/wuanalysis/anacommon.hpp
```diff
@@ -0,0 +1,47 @@
...
+    #define WUANALYSIS_API DECL_IMPORT
+#endif
+
+class PerformanceIssue : public CInterface
+{
+public:
+    int compareCost(const PerformanceIssue & other) const;
+
+    void print() const;
+    void set(__uint64 _cost, const char * msg, ...);
```
ghalliday: Should really have:

`__attribute__((format(printf, 3, 4)));
`
on the definition.

common/wuanalysis/anarule.hpp
```diff
@@ -0,0 +1,42 @@
...
+#define ANARULE_HPP
+
+#include "anacommon.hpp"
+#include "anawu.hpp"
+#include "jstatcodes.h"
+#include "wuattr.hpp"
+
+struct WuAnalyseOptions
+{
+    __uint64 minInterestingTime = msecs2StatUnits(10);// ignore anything under 10 millisecond
```
ghalliday: Should probably use stat_t

common/wuanalysis/anawu.cpp
```diff
@@ -0,0 +1,391 @@
...
+
+class WuScope;
+class WorkunitAnalyser;
+
+class WuScopeHashTable : public SuperHashTableOf<WuScope, const char>
+{
+public:
+    ~WuScopeHashTable() { _releaseAll(); }
+
+    void addNew(WuScope * newWuScope) { SuperHashTableOf::addNew(newWuScope);}
```
ghalliday: does this do anything?  Left over from debugging?

ghalliday (commented):
@shamser a few minor comments.

ghalliday:
@shamser I would suggest creating a new directory testing/analyse or something, and create a regression suite of analysis queries there.

common/wuanalysis/anawu.cpp
shamser: This is used in WuScope::select to add a new scope (if it's not already in the hash table)

shamser:
@ghalliday I provided some test queries.  They'll need to be executed manually for now.  I intend to provide a script to automate the testing after integrating the automate the analysis post-execution.

shamser:
@ghalliday

PR #12611, opened 2019-06-03
HPCC-20914 Log why extra log line is added to ESP log

wangkx:
When the logLevel is set to 1 and the logRequests is set to
false, ESP will log requests if there is some exception or
the processing time is longer than a pre-defined value. In
this fix, ESP will also log why requests are logged.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/bindings/http/platform/httpservice.cpp
+++ b/esp/bindings/http/platform/httpservice.cpp
@@ -73,7 +73,12 @@ CEspHttpServer::~CEspHttpServer()
                 if (paramStr && *paramStr)
                     logStr.appendf("?%s", paramStr);
 
-                DBGLOG("Request[%s]", logStr.str());
+                if (ctx->queryHasException())
+                    DBGLOG("Log request due to exception. Request[%s]", logStr.str());
+                else
+                    DBGLOG("Log request due to long processing time: %u seconds. Request[%s]",
+                        (unsigned) (ctx->queryProcessingTime() * 0.001), logStr.str());
+
                 if (m_request->isSoapMessage())
                 {
                     StringBuffer requestStr;
--- a/esp/bindings/http/platform/httpservice.cpp
+++ b/esp/bindings/http/platform/httpservice.cpp
@@ -64,7 +64,7 @@ CEspHttpServer::~CEspHttpServer()
             //has been logged and it should not be logged here.
             ctx->setProcessingTime();
             if ((ctx->queryHasException() || (ctx->queryProcessingTime() > getSlowProcessingTime())) &&
-                !getEspLogRequests() && (getEspLogLevel() <= LogNormal))
+                (getEspLogRequests() == LogRequestsWithIssuesOnly))
             {
                 StringBuffer logStr;
                 logStr.appendf("%s %s", m_request->queryMethod(), m_request->queryPath());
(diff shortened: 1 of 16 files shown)


wangkx:
@afishbeck please review.

esp/bindings/http/platform/httpservice.cpp
```diff
@@ -73,7 +73,12 @@ CEspHttpServer::~CEspHttpServer()
                 if (paramStr && *paramStr)
                     logStr.appendf("?%s", paramStr);
 
-                DBGLOG("Request[%s]", logStr.str());
+                if (ctx->queryHasException())
+                    DBGLOG("Log request due to exception. Request[%s]", logStr.str());
```
afishbeck: Perhaps simplify slightly to "Request with exception: Request[%s]" and "Request with long processing time: %u seconds. Request[%s]"?

Unless you think its important to be able to search or parse for "Log Request"?

afishbeck (commented):
@wangkx very minor comment.  Otherwise change is fine.

esp/bindings/http/platform/httpservice.cpp
wangkx: @afishbeck do you mean "Request with exception. Request[%s]"?

richardkchapman:
I'm a bit concerned that we may end up logging requests (which could contain sensitive information that should not be logged) when logging of requests has been explicitly turned off (perhaps because they contain sensitive information...). 

I would have thought we need a way to say loggingRequests=never|only-ones-with-issues|all

wangkx:

esp/platform/espcontext.cpp
```diff
@@ -820,6 +820,22 @@ static IEspContainer*& getContainer()
...
+    if (isEmptyString(req))
+        return LogRequestsNever;
+
+    if (strieq(req, "all"))
+        return LogRequestsAlways;
+    if (strieq(req, "never"))
+        return LogRequestsNever;
+    if (strieq(req, "only-ones-with-issues"))
+        return LogRequestsWithIssuesOnly;
+    if (strieq(req, "true"))
```
afishbeck: should probably support "1" as a boolean true value.

esp/platform/espcontext.cpp
```diff
@@ -820,6 +820,22 @@ static IEspContainer*& getContainer()
     return gContainer;
 }
 
+LogRequest readLogRequest(char const* req)
+{
+    if (isEmptyString(req))
```
afishbeck: should probably treat old values of logRequests=false and logRequests=0 as LogRequestsNever.

afishbeck (changes requested):
@wangkx a couple comments.

wangkx:
@afishbeck please review.

esp/platform/espcontext.cpp
richardkchapman: There's a standard function we use for such situations (I can't remember its name)

wangkx: I guess I can use getEnum().

wangkx:
@afishbeck please review the last 2 commits.

wangkx:
@afishbeck please review.

afishbeck (approved):
@wangkx looks good.

PR #12629, opened 2019-06-06
HPCC-21923 Save all custom binding data for a published ESDL binding

timothyklemm:
The ESDL store accepts a property tree describing either a Binding or a
binding's Definition. In either case, custom root node attributes and all
child elements are added to the new binding, not just the Methods element
previously supported.

The ESDL configuration service passes the entire request configuration to
the ESDL store to be saved, instead of only the contained Methods element.
If no configuration is given, the existing behavior is unchanged.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Premature optimization

## Smoketest:

### DIFF
--- a/esp/services/esdl_svc_engine/esdl_store.cpp
+++ b/esp/services/esdl_svc_engine/esdl_store.cpp
@@ -561,28 +561,49 @@ public:
         return configureLogTransform(bindingId, logTransformName, configTree, overwrite, message);
     }
 
-    void preserveBindingContent(IPTree* src, IPTree* dst, IPTree* exclusion) const
+    void mergePTreeWithExclusion(IPTree* src, IPTree* dst, IPTree* exclusion, bool overwriteAttributes = false) const
     {
         if (src)
         {
             Owned<IPTreeIterator> children(src->getElements("*"));
-
             ForEach(*children)
             {
                 auto child = &children->query();
-
                 if (child != exclusion)
                     dst->addPropTree(child->queryName(), LINK(child));
             }
 
             Owned<IAttributeIterator> attributes(src->getAttributes());
-
             ForEach(*attributes)
             {
-                if (!dst->hasProp(attributes->queryName()))
+                if (overwriteAttributes || !dst->hasProp(attributes->queryName()))
                     dst->addProp(attributes->queryName(), attributes->queryValue());
             }
         }
+    }
+
+    bool checkEsdlConfigLevel(IPTree*& methodsConfig, const char* root, const char* child, IPTree*& node)
+    {
+        node = nullptr;
+        if (methodsConfig && (isEmptyString(root) || streq(methodsConfig->queryName(), root)))
+        {
+            node = methodsConfig;
+            if (!isEmptyString(child))
+            {
+                switch (node->getCount(child))
+                {
+                case 0:
+                    methodsConfig = nullptr;
+                    break;
+                case 1:
+                    methodsConfig = node->queryBranch(child);
+                    break;
+                default:
+                    return false;
+                }
+            }
+        }
+        return true;
     }
 
     virtual int bindService(const char* bindingName,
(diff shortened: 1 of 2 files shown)


timothyklemm:
@afishbeck Please review.

esp/services/esdl_svc_engine/esdl_store.cpp
```diff
@@ -770,13 +795,43 @@ class CEsdlSDSStore : implements IEsdlStore, public CInterface
         esdldeftree->setProp("@id", lcId.str());
         esdldeftree->setProp("@esdlservice", esdlServiceName);
 
-        if (methodsConfig != nullptr)
-            esdldeftree->addPropTree("Methods", LINK(methodsConfig));
+#define IDENTIFY_ESDL_ENTRY(node, root, child) \
```
afishbeck: Wouldn't this be more clear as a function.   And a name that was more self documenting would be nice. For example, passing methodsConfig as a reference?:

IPTree *srcBinding = walkEsdlConfigLevel(methodsConfig, ESDL_BINDING_ENTRY, ESDL_DEF_ENTRY);

I would imediately realize you were walking the levels and potentially updating methodsConfig as you went.

Name could be something else, but meaning that it is walking or checking each potential level.
checkEsdlConfigLevel? getEsdlConfigLevel?

afishbeck: And the last one would be clearest as just srcMethods = methodsConfig;

esp/services/esdl_svc_engine/esdl_store.cpp
```diff
@@ -560,6 +561,30 @@ class CEsdlSDSStore : implements IEsdlStore, public CInterface
         return configureLogTransform(bindingId, logTransformName, configTree, overwrite, message);
     }
 
+    void preserveBindingContent(IPTree* src, IPTree* dst, IPTree* exclusion) const
+    {
+        if (src)
+        {
+            Owned<IPTreeIterator> children(src->getElements("*"));
+
```
afishbeck: Extra blank lines?

esp/services/esdl_svc_engine/esdl_store.cpp
```diff
@@ -560,6 +561,30 @@ class CEsdlSDSStore : implements IEsdlStore, public CInterface
         return configureLogTransform(bindingId, logTransformName, configTree, overwrite, message);
     }
 
+    void preserveBindingContent(IPTree* src, IPTree* dst, IPTree* exclusion) const
```
afishbeck: This behavior is more generic than the name implies.  It's really a sort of mergePTreeExcludingNode() and we could find use for it or evolve it over time if it was named more generically.

esp/services/esdl_svc_engine/esdl_store.cpp
```diff
@@ -560,6 +561,30 @@ class CEsdlSDSStore : implements IEsdlStore, public CInterface
...
+
+                if (child != exclusion)
+                    dst->addPropTree(child->queryName(), LINK(child));
+            }
+
+            Owned<IAttributeIterator> attributes(src->getAttributes());
+
+            ForEach(*attributes)
+            {
+                if (!dst->hasProp(attributes->queryName()))
```
afishbeck: Not sure of the exact scenarios, but just checking that the changed values shouldn't win?

esp/services/esdl_svc_engine/esdl_store.cpp
```diff
@@ -770,13 +795,43 @@ class CEsdlSDSStore : implements IEsdlStore, public CInterface
...
+                methodsConfig = node->queryBranch(child); \
+                break; \
+            default: \
+                message.setf("Could not configure Service '%s' with multiple %s elements", esdlServiceName, child); \
+                return -1; \
+            } \
+        } \
+    }
+
+        IDENTIFY_ESDL_ENTRY(srcBinding, ESDL_BINDING_ENTRY, ESDL_DEF_ENTRY);
```
afishbeck: IF we haven't ever allowed the caller to send in a definition section, and IF definition section should always a single service (unlike the ESDL interface definition part in DALI which eventually could have multiple services defined), do you think we could take this opportunity to have the user exposed name be Service instead of Definition.  Even if switched internally it might be more user friendly.

afishbeck (changes requested):
@timothyklemm some comments.

esp/services/esdl_svc_engine/esdl_store.cpp
timothyklemm: I would eventually like to be able to get a binding from one service and republish it to another without having to strip Binding and Definition attributes from the configuration first. This was a way to let the new values win. The destination node is assumed to have been assigned attributes that identify the new binding and, in the event that the source node contains potentially stale values of those attributes, the destination node's values need to win.

An alternative could be to rearrange the code that calls this to only add the new attributes after, overwriting anything preserved here.

esp/services/esdl_svc_engine/esdl_store.cpp
timothyklemm: It is and it isn't. The implementation really assumes that the destination node contains no children and that its attributes take precedence over conflicting values in the source. More than a mergePTreeExcludingNode, its appendChildrenExcludingNode.

I think a generic merge would need to apply a predicate on each source node to decide if it should be included in the merge, and possibly a second predicate to decide how each attribute should be handled.

esp/services/ws_esdlconfig/ws_esdlconfigservice.cpp
```diff
@@ -32,15 +32,15 @@ enum ESDLConfigInfoType
     ESDLLogTransform = 1,
 };
 
-IPropertyTree * fetchConfigInfo(const char * config,
+IPropertyTree * fetchConfigInfo(Owned<IPropertyTree> configTree,
```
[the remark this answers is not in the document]
timothyklemm: It was supposed to be a reference, which would have eliminated questions about reference counting, but point taken.

esp/services/esdl_svc_engine/esdl_store.cpp
timothyklemm: The store has always assumed it was receiving Methods or nothing. But the calling service does accept Definition, and obviously returns Definition when retrieving a binding. I like the idea, but I think it should be its own JIRA that describes the end-to-end solution (acceptable input, what's saved, and what's returned).

esp/services/esdl_svc_engine/esdl_store.cpp
timothyklemm: I chose a macro because it simplified the code that used it, compared to making function calls.
* It handles the declaration of the variable in the calling environment, which may or may not be a good thing.
* It returns errors from the calling environment without needing two return values (or an additional reference parameter) from a function.
* It doesn't require the function to take extra parameters just to prepare the error message, or require me to replicate error message preparation in the caller. Replication can probably be handled by a macro.

I also reused the macro for Methods because I wasn't sure if we would ever want to enforce that we actually have a Methods element, or will always assume that the user of the store always sends either Binding, Definition, or Methods.

I'll refactor.

esp/services/esdl_svc_engine/esdl_store.cpp
afishbeck: If you define "merge" based on behavior of the existing common function "mergePTree" then this would be mergePTreeExcludingNode().  mergePTree keeps things simple as does this.

The other complex method you described would be something else and probably too complex to include in jptree.hpp  This function shouldn't go there as is either but is close to the keep it simple philosophy.

esp/services/esdl_svc_engine/esdl_store.cpp
afishbeck: The problem with its own JIRA is just that we have to make sure this new feature is never released using one tag and then switched in a future release... or we have to forever maintain backward compatibility.

afishbeck: if you think we already have to maintain bw compatibility then of course a new JIRA is fine.

timothyklemm: To summarize an offline discussion, the ESDL config service already processes binding input configurations that contain Binding and/or Definition elements. This story simply prevents those elements from being dropped by the service. Changing Definition to Service in the end user's command line interface does not eliminate the need to maintain support for Definition. Such changes are beyond the scope of this story and should be tracked by a new story.

timothyklemm:
@afishbeck Please review my updates.

afishbeck (approved):
@timothyklemm looks fine.

PR #12633, opened 2019-06-07
HPCC-22205 Note which node to run config scripts on

g-pan:
Signed-off-by: G-Pan <greg.panagiotatos@lexisnexis.com>

@JamesDeFabia please review.  
  Note: Review of Sys. Admin doc did not reveal any others

http://10.240.32.243/view/Docs%207.0/job/AnotherDocBuild7/62/CE=ubuntu-18.04-amd64-2/artifact/build-docs/Release/docs/EN_US/HPCCSystemAdministratorsGuide_EN_US-7.2.17-closedown0.pdf#page=68

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

## Testing:

Successful Build Test: http://10.240.32.243/view/Docs%207.0/job/AnotherDocBuild7/

### DIFF
--- a/docs/EN_US/Installing_and_RunningTheHPCCPlatform/Inst-Mods/hpcc_ldap.xml
+++ b/docs/EN_US/Installing_and_RunningTheHPCCPlatform/Inst-Mods/hpcc_ldap.xml
@@ -812,6 +812,8 @@ Proceed?  y/n
       for example:</para>
 
       <para><programlisting>/opt/HPCCSystems/bin/addScopes /var/lib/HPCCSystems/mydali/daliconf.xml</programlisting></para>
+
+      <para>Run addScopes on the Dali node.</para>
     </sect3>
   </sect2>
 </sect1>


JamesDeFabia (commented):
@g-pan 
What about 
 /opt/HPCCSystems/bin/initldap and addScopes? Does it matter from which node these are run?

g-pan:
> @g-pan
> What about
> /opt/HPCCSystems/bin/initldap and addScopes? Does it matter from which node these are run?

On a multi-node system you could run addScopes an any node where
 you can access the daliconf.xml file.

g-pan:
re: initldap is in the config section where we state: &lt;configmgr_IP_Address&gt; is the IP address of the node running Configuration Manager

JamesDeFabia (approved):
@richardkchapman This is good to merge

richardkchapman:
@jakesmith is this still accurate after whitelist changes?

jakesmith:
hm, not familiar with addScopes.. 
It seems to interact directly with LDAP in one mode (to add)  (security?)

And in other mode (clear) to try to talk to Dali's session manager.. 
Afaics it would have always failed to do so though, as there is no initClientProcess call, and it's attempt to use the session manager would therefore fail.

It doesn't seem right that it gets it's config info from daliconf.xml.

Is this program actually used? I suspect it needs revisiting.
@RussWhitehead - can you comment?

RussWhitehead:
@jakesmith @g-pan addScopes is meant to enumerate all the LDAP users and ensure their hpccinternal::<username> file scope exists with the default permissions. To find the correct LDAP Server and binding information, it needs to extract it from the ldapSecurity block of the daliconf.xml.

jakesmith:
> To find the correct LDAP Server and binding information, it needs to extract it from the ldapSecurity block of the daliconf.xml.

which necessitates it having to run on the same node as Dali?

If it has to connect to Dali.. (?).. that needs testing (since initClientProcess appears to be missing), and if it connects to Dali, why does it need to read daliconf.xml? It could read the info. it needs from the environment , and therefore could be run on any node.

RussWhitehead:
It needs to run on a node that has a copy of the current daliconf.xml.  The only reason it connects to Dali (that I am aware of) is to clear the permissions cache. This code could probably be removed

g-pan:
@jakesmith  would the white list invalidate any of these changes to the text? 
It does not seem to.  Would this then be OK to merge?

jakesmith:
Well, no because addscopes didn't work if it tried to connect to Dali anyway (as I say above, there's no initClientProcess or mp connect).
Afaics it will fail (crash or assert) if it tries to access Dali.
@RussWhitehead  - can you take a look, check that and open a JIRA if confirmed? 

>On a multi-node system you could run addScopes an any node where
>you can access the daliconf.xml file. By default the daliconf.xml file
>can be found on the node where your Dali is.

@g-pan That's a bit misleading - daliconf.xml cannot be anywhere else, but the Dali node.
(it's odd that addRoles uses daliconf.xml at all and not the environment, but that's another matter)

g-pan:
@jakesmith would this change suffice?

JamesDeFabia (changes requested):

g-pan:
@JamesDeFabia @jakesmith please review latest changes. Thanks.

JamesDeFabia (approved):
Looks good to me

richardkchapman:
@RussWhitehead @jakesmith Can you confirm?

RussWhitehead:
@richardkchapman  THis is correct as written

PR #12636, opened 2019-06-10
HPCC-22285 Execute WU Analysis post execution and write results to WU

shamser:
## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/common/wuanalysis/anacommon.hpp
+++ b/common/wuanalysis/anacommon.hpp
@@ -19,6 +19,8 @@
 #define ANACOMMON_HPP
 
 #include "jliball.hpp"
+#include "wuattr.hpp"
+#include "eclhelper.hpp"
 
 #ifdef WUANALYSIS_EXPORTS
     #define WUANALYSIS_API DECL_EXPORT
@@ -26,15 +28,46 @@
     #define WUANALYSIS_API DECL_IMPORT
 #endif
 
+interface IWuScope
+{
+    virtual stat_type getStatRaw(StatisticKind kind, StatisticKind variant = StKindNone) const = 0;
+    virtual unsigned getAttr(WuAttr kind) const = 0;
+    virtual void getAttr(StringBuffer & result, WuAttr kind) const = 0;
+};
+
+interface IWuActivity;
+interface IWuEdge : public IWuScope
+{
+    virtual IWuActivity * querySource() = 0;
+    virtual IWuActivity * queryTarget() = 0;
+};
+
+interface IWuActivity : public IWuScope
+{
+    virtual const char * queryName() const = 0;
+    virtual IWuEdge * queryInput(unsigned idx) = 0;
+    virtual IWuEdge * queryOutput(unsigned idx) = 0;
+    inline IWuActivity * queryInputActivity(unsigned idx)
+    {
+        IWuEdge * edge = queryInput(idx);
+        return edge ? edge->querySource() : nullptr;
+    }
+    inline ThorActivityKind queryThorActivityKind()
+    {
+        return (ThorActivityKind) getAttr(WaKind);
+    }
+};
+
 class PerformanceIssue : public CInterface
 {
 public:
     int compareCost(const PerformanceIssue & other) const;
-
     void print() const;
+    void createException(IWorkUnit * we);
+
     void set(stat_type _cost, const char * msg, ...) __attribute__((format(printf, 3, 4)));
-    void setScope(const char *_scope) { scope.set(scope); };
-    stat_type getCost() const          { return cost; }
+    void setScope(const char *_scope) { scope.set(scope); }
+    stat_type getCost() const         { return cost; }
 
 private:
     StringAttr scope;
--- a/common/wuanalysis/anawu.cpp
+++ b/common/wuanalysis/anawu.cpp
@@ -88,17 +88,17 @@ public:
 
     void check(const char * scope, IWuActivity & activity);
     void analyse(IConstWorkUnit * wu);
+    void print();
+    void update(IWorkUnit *wu);
 
 protected:
     void collateWorkunitStats(IConstWorkUnit * workunit, const WuScopeFilter & filter);
-    void printWarnings();
     WuScope * selectFullScope(const char * scope);
 
-protected:
     CIArrayOf<AActivityRule> rules;
     CIArrayOf<PerformanceIssue> issues;
     WuScope root;
-    WuAnalyseOptions & options;
+    WuAnalyseOptions options;
 };
 
 //-----------------------------------------------------------------------------------------------------------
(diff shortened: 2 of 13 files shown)


shamser:
@ghalliday

ecl/eclagent/eclagent.cpp
```diff
@@ -1950,6 +1951,12 @@ void EclAgent::doProcess()
             break;
         }
 
+        if (w->getState() == WUStateCompleted)
+        {
+            const char * clustername = w->queryClusterName();
+            if (getClusterTypeByClusterName(clustername)==ThorLCRCluster)
+                analyseWorkunit(static_cast<IConstWorkUnit*>(w.get()));
```
ghalliday: Minor: would be better to pass in w as an IWorkUnit * - then there would be no need for the code to lock the workunit.  (There also shouldn't need to be a cast to convert a IWorkUnit * to a IConstWorkUnit *.

ecl/eclagent/eclagent.cpp
```diff
@@ -1950,6 +1951,12 @@ void EclAgent::doProcess()
             break;
         }
 
+        if (w->getState() == WUStateCompleted)
+        {
+            const char * clustername = w->queryClusterName();
+            if (getClusterTypeByClusterName(clustername)==ThorLCRCluster)
```
ghalliday: trivial: possibly better to check clusterType - may have less overhead.

ecl/eclagent/eclagent.cpp
```diff
@@ -1950,6 +1951,12 @@ void EclAgent::doProcess()
             break;
         }
 
+        if (w->getState() == WUStateCompleted)
```
ghalliday: We should have a #option that can be used to disable this code - just in case there is a problem with it that causes eclagent to crash.

ghalliday (commented):
@shamser a few minor comments.  Also please rebase once the other PR is merged.

shamser:
@ghalliday A new ticket has been created to document the configuration and #option (HPCC-22291)  Please can you review.

ecl/eclagent/eclagent.cpp
```diff
@@ -1950,6 +1951,25 @@ void EclAgent::doProcess()
...
         }
 
+        if (w->getState() == WUStateCompleted && getClusterType(clusterType)==ThorLCRCluster)
+        {
+            if (agentTopology->getPropBool("@analyzeWorkunit", true) && w->getDebugValueBool("analyzeWorkunit", true))
+            {
+                WuAnalyseOptions options;
+                CIArrayOf<PerformanceIssue> issues;
+                analyseWorkunit(w.get(), options, issues);
+                ForEachItemIn(i, issues)
```
ghalliday: I preferred the structure of the previous commit.  You are breaking encapsulation of the analysis logic - previously these classes were not part of the public interface (and they are not exported from the dll).  It also means you have lost the option to run the analyser from the wutool and write the results to the workunit.  I can imagine that being useful - e.g., if you want to run new rules on an old workunit.

common/wuanalysis/anacommon.hpp
```diff
@@ -19,6 +19,7 @@
 #define ANACOMMON_HPP
 
 #include "jliball.hpp"
+#include "workunit.hpp"
```
ghalliday: Why is this needed?

common/wuanalysis/anacommon.hpp
```diff
@@ -44,4 +46,11 @@ class PerformanceIssue : public CInterface
 
 extern int compareIssuesCostOrder(CInterface * const * _l, CInterface * const * _r);
 
+struct WuAnalyseOptions
```
ghalliday: This was original intended to be a private include file - this should be in anawu.hpp which is the public interface to the analyser

ecl/eclagent/eclagent.cpp
```diff
@@ -1950,6 +1951,25 @@ void EclAgent::doProcess()
             break;
         }
 
+        if (w->getState() == WUStateCompleted && getClusterType(clusterType)==ThorLCRCluster)
+        {
+            if (agentTopology->getPropBool("@analyzeWorkunit", true) && w->getDebugValueBool("analyzeWorkunit", true))
```
ghalliday: The following is better - since it allows explicit overriding both ways:

`(w->getDebugValueBool("analyzeWorkunit", agentTopology->getPropBool("@analyzeWorkunit", true))
`

ghalliday (commented):
@shamser see some comments

ecl/eclagent/eclagent.cpp
shamser: My thinking was that analyseWorkunit would do only that - analyse.  Updating the workunit or printing out the results doesn't seem part of the analysis. We could provide a separate function to write the issues to the WU and another function to print out the issues.

The other question is that if we were to do it the way it was done before, would we pass a locked or an unlocked workunit?  A locked workunit is not necessary for Analyze & Print but an locked workunit is required for Analyse & Update WU.  It seems that we'd need to either lock WU unnessary or need to relock the WU (that may already be locked).

common/wuanalysis/anacommon.hpp
shamser: It doesn't need to be here - I'm removing.

common/wuanalysis/anacommon.hpp
shamser: I understand. Will move public interface to anawu.hpp.

ecl/eclagent/eclagent.cpp
shamser: Neat. That works better.

ecl/eclagent/eclagent.cpp
ghalliday: There is always a balancing act for where code goes - but generally it is best to try and encapsulate code to minimize information leaking.  For instance if the results of the analysis were then going to be processed by multiple other stages then it would make sense to export the information about the results.  If (as in this case) the results have little further processing it makes sense to keep the details hidden.

As for the details, in this case I would have had two functions, one that took a IWorkUnit and added the results to the workunit exceptions, the other than took an IConstWorkUnit (which is a base interface of IWorkUnit) and printed them.  They would use common code within the anawu module.

Slightly related one rule I have found useful is that if most of the code within a function is accessing members of an object that as a good indication that should be a member function.  Which would argue against removing PerformanceIssue::print() - and similarly the code to add an issue as a workunit exception would be cleaner as a member function.

common/wuanalysis/anawu.cpp
```diff
@@ -354,14 +352,6 @@ void WorkunitAnalyser::collateWorkunitStats(IConstWorkUnit * workunit, const WuS
     }
 }
 
-void WorkunitAnalyser::printWarnings()
```
ghalliday: Better encapsulation to  keep this function and also add a createExceptions(wu) member, and avoids the refeference parameter on the constructor.

ghalliday:
@shamser looks better.  One comment - I don't mind if you make the change or not.  Please squash into a single commit, and I'll merge.

shamser:
@ghalliday

ghalliday:
Great, looks clean.

PR #12642, opened 2019-06-11
HPCC-22196 Ensure text of failure is available to FAILMESSAGE

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/testing/regress/ecl/thor/issue22196.xml
+++ b/testing/regress/ecl/thor/issue22196.xml
@@ -1,0 +1,10 @@
+<Exception><Code>100000</Code><Filename>issue22196.ecl</Filename><Line>4</Line><Source>user</Source><Message>Letter 'A' not allowed</Message></Exception>
+<Dataset name='Result 1'>
+ <Row><Result_1>B</Result_1></Row>
+</Dataset>
+<Dataset name='Result 2'>
+ <Row><Result_2>Graph graph1[1], simpleaction[2]: Letter &apos;A&apos; not allowed, Master exception - caused by (100000, Letter &apos;A&apos; not allowed) (in item 1)</Result_2></Row>
+</Dataset>
+<Dataset name='Result 3'>
+ <Row><Result_3>Recover</Result_3></Row>
+</Dataset>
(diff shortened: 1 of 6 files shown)


ghalliday:
@jakesmith please review

testing/regress/ecl/thor/issue22196.xml
```diff
@@ -0,0 +1,10 @@
+<Exception><Code>100000</Code><Filename>/home/gavin/dev/hpcc/testing/regress/ecl/issue22196.ecl</Filename><Line>4</Line><Source>user</Source><Message>Letter &apos;A&apos; not allowed</Message></Exception>
```
jakesmith: test mismatching, because has personal make install path.. 
Can you transform to regex out the leading path, so always matches?

jakesmith:
@ghalliday - looks fine, but regression test failing because different path in message.
Checkbox's need ticking also.

ghalliday:
@AttilaVamos any ideas what to do about the difference in filename?  It is not a result, so there is no way of modifying it from within the ecl.

AttilaVamos:
@ghalliday I see. So, it seems in the Regression Test Engine I should remove the path from <Exception> ...<Filename> ... </Filename> result . (As I did it before for <Warning>.)  
After this the key file should have the filename only (<Exception><Code>100000</Code><Filename>issue22196.ecl</Filename> ...).
Is this solution OK for you?

ghalliday:
@AttilaVamos yes that would solve it

testing/regress/ecl/thor/issue22196.xml
richardkchapman: We should consider stripping out path information from the filename at compile time - you could argue relative info is valuable but if it's simpler to strip all path info I would not be too upset

ghalliday:
@AttilaVamos this doesn't look like the path has been removed from the exception.

AttilaVamos:
@ghalliday my fix has been merged into 7.2.x branch. If you up-merge it into master, this PR will be clean.

richardkchapman:
I have upmerged - if the smoketest can be restarted

AttilaVamos:
@ghalliday the keyfile has a full test file path and the \&apos;A\&apos; should change to 'A'.

PR #12660, opened 2019-06-13
HPCC-22313 Strip path from <Filename>filename</Filename> in <Exception> line of test result.

AttilaVamos:
Add check to enable strip the path from <Exception> line.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

## Testing:

Tested locally and manually with "HPCC-22196 Ensure text of failure is available to FAILMESSAGE #12642" code.

### DIFF
--- a/testing/regress/ecl/key/assert.xml
+++ b/testing/regress/ecl/key/assert.xml
@@ -15,7 +15,7 @@
 <Exception><Code>100000</Code><Filename>assert.ecl</Filename><Line>46</Line><Source>user</Source><Message>Assert (2 = 1) failed [val1 = 1]</Message></Exception>
 <Exception><Code>100000</Code><Filename>assert.ecl</Filename><Line>47</Line><Source>user</Source><Message>Abc7</Message></Exception>
 <Exception><Code>100000</Code><Filename>assert.ecl</Filename><Line>67</Line><Source>user</Source><Message>Assert (2 = 9) failed [COUNT(ds) = 9]</Message></Exception>
-<Exception><Code>100000</Code><Filename>assert.ecl</Filename><Line>71</Line><Source>user</Source><Message>Assert (&#226;&#130;&#172; = &#226;&#130;&#172;&#226;&#130;&#172;) failed [euro = U&apos;&#226;&#130;&#172;&#226;&#130;&#172;&apos;]</Message></Exception>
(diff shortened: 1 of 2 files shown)


richardkchapman:
I'm not convinced this is the right approach. We should not be generating fully-expanded paths in our exceptions (or in our graphs).

@ghalliday What do you think?

AttilaVamos:
@ghalliday @richardkchapman this PR is ready to decision.

(BTW: If I see well in assert.ecl, there is already implemented something similar : "#option ('reportAssertFilenameTail', true);" )

ghalliday:
@richardkchapman I had a look at fixing this in the code generator.  It was relatively simple to fix the paths found via -I to only include the relevant trailing path.  However it isn't clear what to do with the filename that was supplied to compile the query.  If it is within the include paths then it could be mapped, but what if it isn't?

ghalliday:
The other problem would be that filenames reported from the code generator would no longer be absolute pathnames - which would make it hard for ecl ide to resolve the correct file.

testing/regress/ecl/key/assert.xml
```diff
@@ -15,7 +15,7 @@
 <Exception><Code>100000</Code><Filename>assert.ecl</Filename><Line>46</Line><Source>user</Source><Message>Assert (2 = 1) failed [val1 = 1]</Message></Exception>
 <Exception><Code>100000</Code><Filename>assert.ecl</Filename><Line>47</Line><Source>user</Source><Message>Abc7</Message></Exception>
 <Exception><Code>100000</Code><Filename>assert.ecl</Filename><Line>67</Line><Source>user</Source><Message>Assert (2 = 9) failed [COUNT(ds) = 9]</Message></Exception>
-<Exception><Code>100000</Code><Filename>assert.ecl</Filename><Line>71</Line><Source>user</Source><Message>Assert (&#226;&#130;&#172; = &#226;&#130;&#172;&#226;&#130;&#172;) failed [euro = U&apos;&#226;&#130;&#172;&#226;&#130;&#172;&apos;]</Message></Exception>
+<Exception><Code>100000</Code><Filename>assert.ecl</Filename><Line>71</Line><Source>user</Source><Message>Assert (&#226;&#130;&#172; = &#226;&#130;&#172;&#226;&#130;&#172;) failed [euro = U'&#226;&#130;&#172;&#226;&#130;&#172;']</Message></Exception>
```
richardkchapman: This change seems unrelated

PR #12665, opened 2019-06-14
HPCC-22328 WU Analysis store full scope in Exception->scope

shamser:
Signed-off-by: Shamser Ahmed <shamser.ahmed@lexisnexis.co.uk>

## Type of change:
- [ ] This change is a bug fix (non-breaking change which fixes an issue).
- [ ] This change is a new feature (non-breaking change which adds functionality).
- [x] This change improves the code (refactor or other change that does not change the functionality)
- [ ] This change fixes warnings (the fix does not alter the functionality or the generated code)
- [ ] This change is a breaking change (fix or feature that will cause existing behavior to change).
- [ ] This change alters the query API (existing queries will have to be recompiled)

## Checklist:
- [x] My code follows the code style of this project.
  - [x] My code does not create any new warnings from compiler, build system, or lint.
- [x] The commit message is properly formatted and free of typos.
  - [ ] The commit message title makes sense in a changelog, by itself.
  - [ ] The commit is signed.
- [ ] My change requires a change to the documentation.
  - [ ] I have updated the documentation accordingly, or...
  - [ ] I have created a JIRA ticket to update the documentation.
  - [ ] Any new interfaces or exported functions are appropriately commented.
- [x] I have read the CONTRIBUTORS document.
- [x] The change has been fully tested:
  - [ ] I have added tests to cover my changes.
  - [ ] All new and existing tests passed.
  - [ ] I have checked that this change does not introduce memory leaks.
  - [ ] I have used Valgrind or similar tools to check for potential issues.
- [ ] I have given due consideration to all of the following potential concerns:
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization
  - [ ] Existing deployed queries will not be broken
  - [ ] This change fixes the problem, not just the symptom
  - [ ] The target branch of this pull request is appropriate for such a change.
- [ ] There are no similar instances of the same problem that should be addressed
  - [ ] I have addressed them here
  - [ ] I have raised JIRA issues to address them separately
- [ ] This is a user interface / front-end modification
  - [ ] I have tested my changes in multiple modern browsers
  - [ ] The component(s) render as expected

## Smoketest:
- [ ] Send notifications about my Pull Request position in Smoketest queue.
- [ ] Test my draft Pull Request.

PR #12669, opened 2019-06-14
HPCC-22335 Regression fix in remote index read if no outMeta

jakesmith:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/fs/dafsserver/dafsserver.cpp
+++ b/fs/dafsserver/dafsserver.cpp
@@ -1984,7 +1984,6 @@ protected:
     bool isTlk = false;
     bool allowPreload = false;
     unsigned fileCrc = 0;
-    Owned<const IDynamicTransform> translator;
     Owned<IKeyIndex> keyIndex;
     Owned<IKeyManager> keyManager;
 
@@ -2023,7 +2022,6 @@ public:
         isTlk = config.getPropBool("isTlk");
         allowPreload = config.getPropBool("allowPreload");
         fileCrc = config.getPropInt("crc");
-        translator.setown(createRecordTranslator(outMeta->queryRecordAccessor(true), *record));
     }
 };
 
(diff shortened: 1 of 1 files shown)


fs/dafsserver/dafsserver.cpp
```diff
@@ -2056,8 +2057,16 @@ class CRemoteIndexReadActivity : public CRemoteIndexBaseActivity
...
                     const byte *keyRow = keyManager->queryKeyBuffer();
-                    retSz = translator->translate(outBuilder, *this, keyRow);
-                    if (retSz)
+                    if (translator)
+                        retSz = translator->translate(outBuilder, *this, keyRow);
+                    else
+                    {
+                        outBuilder.ensureRow();
+                        retSz = keyManager->queryRowSize();
+                        memcpy(outBuilder.getSelf(), keyRow, retSz);
```
jakesmith: @ghalliday - is this right way to copy index fields if no translation?

NB: before this fix and before HPCC-18484, it always used a translator - so this is secondary to the fix for this JIRA.

fs/dafsserver/dafsserver.cpp
```diff
@@ -2056,8 +2057,16 @@ class CRemoteIndexReadActivity : public CRemoteIndexBaseActivity
...
+                    }
+
+                    if (retSz) // JCSMORE is this poss.?
```
jakesmith: @ghalliday - are there any circumstacnes where this could be 0?

jakesmith:
@ghalliday - the change to build row without translator isn't directly related to the crash, but seems translator should be avoided if there is no output meta.
Can you check if approach is correct (couple of questions above inline) ? Thanks.

richardkchapman:
Not sure what the details of when this happens are - why is there no output meta?

jakesmith:
Btw this regression is the cause of the cores that Attila saw in https://github.com/hpcc-systems/HPCC-Platform/pull/12665.

If actual == project format then outMeta is not sent. Prior to the introduction of this regression that meant dafilesrv would spot no outMeta and use inMeta as outMeta

fs/dafsserver/dafsserver.cpp
ghalliday: This will have problems for variable length keys. I think it needs to be

    memcpy(outBuilder.ensureCapacity(retSz), keyRow, retSz);

You should be able to remove the ensureRow call as well.

fs/dafsserver/dafsserver.cpp
ghalliday: I don't think so - same as the situation with the disk read, transform() cannot filter.

ghalliday:
@jakesmith see comment/reply

fs/dafsserver/dafsserver.cpp
jakesmith: yes and actually I'd already changed/pushed it prior to your comment to:
```
                        retSz = keyManager->queryRowSize();
                        outBuilder.ensureCapacity(retSz, nullptr);
                        memcpy(outBuilder.getSelf(), keyRow, retSz);
```
but github has preserved old code with my/your comment it seems.
Think that's equivalent to what you have suggested.

fs/dafsserver/dafsserver.cpp
jakesmith: ok I'll change it to a dbgassertex as in other cases.

PR #12606, opened 2019-05-31
HPCC-22124 File services permits and defaults to copying a super index into a invalid single index with multiple TLK's

AttilaVamos:
Add Superfile check for a simple logical file copy.

Add supercopy.ecl to Regression Suite

Update 3 after review,

Fix problems caused by the merge local and remote source file handling

Remove 'foreignfdesc'

Signed-off-by: Attila Vamos <attila.vamos@gmail.com>

## Type of change:
- [x] This change is a bug fix (non-breaking change which fixes an issue).
- [ ] This change is a new feature (non-breaking change which adds functionality).
- [ ] This change improves the code (refactor or other change that does not change the functionality)
- [ ] This change fixes warnings (the fix does not alter the functionality or the generated code)
- [ ] This change is a breaking change (fix or feature that will cause existing behavior to change).
- [ ] This change alters the query API (existing queries will have to be recompiled)

## Checklist:
- [x] My code follows the code style of this project.
  - [x] My code does not create any new warnings from compiler, build system, or lint.
- [x] The commit message is properly formatted and free of typos.
  - [x] The commit message title makes sense in a changelog, by itself.
  - [x] The commit is signed.
- [ ] My change requires a change to the documentation.
  - [ ] I have updated the documentation accordingly, or...
  - [ ] I have created a JIRA ticket to update the documentation.
  - [ ] Any new interfaces or exported functions are appropriately commented.
- [x] I have read the CONTRIBUTORS document.
- [x] The change has been fully tested:
  - [x] I have added tests to cover my changes.
  - [x] All new and existing tests passed.
  - [ ] I have checked that this change does not introduce memory leaks.
  - [ ] I have used Valgrind or similar tools to check for potential issues.
- [x] I have given due consideration to all of the following potential concerns:
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization
  - [ ] Existing deployed queries will not be broken
  - [ ] This change fixes the problem, not just the symptom
  - [x] The target branch of this pull request is appropriate for such a change.
- [x] There are no similar instances of the same problem that should be addressed
  - [x] I have addressed them here
  - [ ] I have raised JIRA issues to address them separately
- [ ] This is a user interface / front-end modification
  - [ ] I have tested my changes in multiple modern browsers
  - [ ] The component(s) render as expected

## Smoketest:
- [x] Send notifications about my Pull Request position in Smoketest queue.
- [x] Test my draft Pull Request.

[...]

PR #12674, opened 2019-06-17
HPCC-22351 Force super file option when copying a super index

wangkx:
In this fix, ESP FileSpray.Copy is changed to force super file
option when copying a super index.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/services/ws_fs/ws_fsService.cpp
+++ b/esp/services/ws_fs/ws_fsService.cpp
@@ -2490,32 +2490,41 @@ bool CFileSprayEx::onCopy(IEspContext &context, IEspCopy &req, IEspCopyResponse 
         StringBuffer fileMask;
         constructFileMask(destTitle.str(), fileMask);
 
+        Owned<IUserDescriptor> udesc=createUserDescriptor();
         const char* srcDali = req.getSourceDali();
-        bool supercopy = req.getSuperCopy();
-        if (supercopy)
+        const char* srcu = req.getSrcusername();
+        if (!isEmptyString(srcDali) && !isEmptyString(srcu))
+        {
+            udesc->set(srcu, req.getSrcpassword());
+        }
+        else
         {
             StringBuffer user, passwd;
             context.getUserID(user);
             context.getPassword(passwd);
-            StringBuffer u(user);
-            StringBuffer p(passwd);
-            Owned<INode> foreigndali;
-            if (srcDali)
-            {
-                SocketEndpoint ep(srcDali);
-                foreigndali.setown(createINode(ep));
-                const char* srcu = req.getSrcusername();
-                if(srcu && *srcu)
-                {
-                    u.clear().append(srcu);
-                    p.clear().append(req.getSrcpassword());
-                }
-            }
-            Owned<IUserDescriptor> udesc=createUserDescriptor();
-            udesc->set(u.str(),p.str());
-            if (!queryDistributedFileDirectory().isSuperFile(srcname,udesc,foreigndali))
+            udesc->set(user, passwd);
+        }
+
+        CDfsLogicalFileName logicalName;
+        logicalName.set(srcname);
+        if (!isEmptyString(srcDali))
+        {
+            SocketEndpoint ep(srcDali);
+            logicalName.setForeign(ep,false);
+        }
+
+        Owned<IDistributedFile> file = queryDistributedFileDirectory().lookup(logicalName, udesc);
+        if (!file)
+            throw MakeStringException(ECLWATCH_FILE_NOT_EXIST, "Failed to find file: %s", logicalName.get());
+
+        bool supercopy = req.getSuperCopy();
+        if (supercopy)
+        {
+            if (!file->querySuperFile())
                 supercopy = false;
         }
+        else if (file->querySuperFile() && isFileKey(file))
+            supercopy = true;
 
         Owned<IDFUWorkUnitFactory> factory = getDFUWorkUnitFactory();
         Owned<IDFUWorkUnit> wu = factory->createWorkUnit();
--- a/testing/regress/ecl/key/supercopy.xml
+++ b/testing/regress/ecl/key/supercopy.xml
@@ -13,7 +13,7 @@
  <Row><result>Copy superfile as superfle: Pass</result></Row>
 </Dataset>
 <Dataset name='Result 7'>
- <Row><result>Copy superindex as logical file: Fail</result></Row>
+ <Row><result>Copy superindex as logical file: Pass</result></Row>
 </Dataset>
 <Dataset name='Result 8'>
  <Row><result>Copy superindex as superfle: Pass</result></Row>


wangkx:
@jakesmith please review

wangkx:
I am not sure why Smoketests failed when the supercopy is on hthor and roxie. Should I disable the super file option if the target group is hthor or roxie?

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -2490,32 +2490,37 @@ bool CFileSprayEx::onCopy(IEspContext &context, IEspCopy &req, IEspCopyResponse
...
+        udesc->set(user, passwd);
+
+        bool supercopy = req.getSuperCopy();
+        if (supercopy)
+        {
+            if (!queryDistributedFileDirectory().isSuperFile(srcname, udesc, foreigndali))
                 supercopy = false;
         }
+        else if (queryDistributedFileDirectory().isSuperIndexFile(srcname, udesc, foreigndali))
+            supercopy = true;
```
AttilaVamos: @wangkx practically those two lines are ok, but theoretically I think  they are wrong because:
- it changes the meaning/usage of 'asSuper' parameter behind the scenes
- it is entirely disable my changes in dfu server because it always force the superkey/superindex copy as supercopy 
(- and finally in educational point of view, it prevents the user/ECL programmer faces to their mistake)

We can start a debate about this changes and about it is good if our platform are cleverer than its users. There has a sense to copy a superfile as a superfile or a simple logical file, but there isn't with a superkey. So, I think in this case we should separate to copy a superfile from copy a superkey. 

If I remove the two lines above from your PR code then the  supercopy.ecl finish well. (With this I remove the main part of your PR as well, sorry)

I think this PR is  a duplicate of the already merged  PR #12606: HPCC-22124 File services permits and defaults to copying a super index into a invalid single index with multiple TLK's by AttilaVamos in hpcc-systems/HPCC-Platform on GitHub. 
Perhaps we should ask an architect.

AttilaVamos (changes requested):

esp/services/ws_fs/ws_fsService.cpp
wangkx: I wish this PR is a duplicate of https://github.com/hpcc-systems/HPCC-Platform/pull/12606. But, I found that https://github.com/hpcc-systems/HPCC-Platform/pull/12606 may not be good enough for ESP. When a user copies a super index without the super file option using ESP, the existing ESP code just creates a job for DFU server. With https://github.com/hpcc-systems/HPCC-Platform/pull/12606, dfuserver may log an error. But, if the user does not read the log, the user does not know what happens. 
I was considering two options for this PR. 1. Throw an exception when a user copies a super index without the super file option. 2. Force the super file option when a user copies a super index. I am choosing the second option because ESP does the similar thing in line 2516-2517 in the existing ws_fsService.cpp. @jakesmith please comment.

AttilaVamos: @wangkx I think the parent JIRA (https://track.hpccsystems.com/browse/HPCC-22139) of this means it would be enough if the ECLWatch  prevents (on or immediately after the copy dialog) the user to trying copy superkey without using  'asSuper' parameter. 

Any other case, if a copy (workunit) fails (the https://github.com/hpcc-systems/HPCC-Platform/pull/12606 throws an exception) and the user/ECL programmer doesn't aware of it /check the logs/results, then I don't know hat we can/should do.

wangkx: @AttilaVamos an ESP service may be called without ECLWatch UI. For example, a user may send a SOAP call to ESP.

AttilaVamos: @wangkx  Yes, but every time the copy goes through DFU server (if I know well), where all necessary checks are already implemented.

wangkx: @AttilaVamos The copy does go through DFU server and DFU server does throw an exception. But, I do not think that most of ECL programmers will check DFU server log even if the copy fails.

AttilaVamos: @wangkx as I mentioned earlier, we can start a debate about...
@jakesmith what is your opinion.

dali/base/dadfs.cpp
```diff
@@ -7549,6 +7550,17 @@ bool CDistributedFileDirectory::isSuperFile(    const char *logicalname,
     return tree.get()&&(strcmp(tree->queryName(),queryDfsXmlBranchName(DXB_SuperFile))==0);
 }
 
+bool CDistributedFileDirectory::isSuperIndexFile(const char *logicalname,
+    IUserDescriptor *user, INode *foreigndali, unsigned timeout)
+{
+    Owned<IPropertyTree> tree = getFileTree(logicalname, user, foreigndali, timeout, false);
+    if (!tree.get())
```
jakesmith: I am concerned with [lack of] efficiency of this call and the previous isSuperFile(),
especially as the calling code may call both..

If it's a foreign file - you may have to do a getFileTree(), but should avoid doing it twice.

I would not introduce this new method and not use isSuperFile(), but instead
call something that returns you a IFileDescriptor or a IDistributedFile (e.g. IDistributedFileDirectory::getFileDescriptor()) - and test that for super and index.

AttilaVamos: @jakesmith I think the current implementation is only the first half of the question. The second part is (I think): Is this a duplicated solution of an already merged PR?
And the bonus (design/behaviour/educational) question is  Should we hide a user/ECL programmer error when they wants to copy a superkey file as a simple logical file.

jakesmith: From a IDistributedFile you should be able to tell:
1) if it's a super (cheaply) by querying querySuperFile().
2) if it's a index/key file (cheaply) via isFileKey()

So afaik, you should be able to do a normal lookup (that returns you a IDistributedFile) , then test.
And as per previous comment - throw an exception if the're attempting to copy a super index without preserving the superfile.

esp/services/ws_fs/ws_fsService.cpp
jakesmith: I've changed my mind..
Eclwatch users might be expected to look at an index and choose appropriate options (by Eclwatch now suppresses the invalid options).
Other users (SOAP and ECL) who are asking to copy an index, generally have generic code and cannot know (and do not are). This option should be irrelevant (not even an option) for them when copying an index. It is has 1 answer and it can never validly be anything else.
You could argue there should be a separate service call or a separate fileservices call (e.g. CopyIndex), but using std. Copy it just means some options do not apply (make any sense), supercopy is one of them.

If we make it fire an error when they pick badly, then they'll be nasty conditional code reading and testing to set appropriate setting where only 1 answer is ever correct for indexes.
We already ignore supercopy=true if the file isn't a superfile, which also seems reasonable.

So I think it should just ignore supercopy=false if a superindex, i.e. perform as if supercopy=true.
i.e. as @wangkx has done.

dali/base/dadfs.cpp
AttilaVamos: The DFU server (in the #12606) already throws an exception. 
Hidden in the ESP this code to fix the programmer/user error and my question is: do we really want to do this?

jakesmith: See my other comments.
https://github.com/hpcc-systems/HPCC-Platform/pull/12674#discussion_r295375113

wangkx:
@jakesmith please review.

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -2490,32 +2490,30 @@ bool CFileSprayEx::onCopy(IEspContext &context, IEspCopy &req, IEspCopyResponse
         StringBuffer fileMask;
         constructFileMask(destTitle.str(), fileMask);
 
+        Owned<IUserDescriptor> udesc=createUserDescriptor();
         const char* srcDali = req.getSourceDali();
-        bool supercopy = req.getSuperCopy();
-        if (supercopy)
+        const char* srcu = req.getSrcusername();
+        if (srcDali && !isEmptyString(srcu))
```
jakesmith: If 'srcu' could be "" (and hence isEmptyString()), then can't 'srcDali' also be?

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -2490,32 +2490,30 @@ bool CFileSprayEx::onCopy(IEspContext &context, IEspCopy &req, IEspCopyResponse
...
-                    p.clear().append(req.getSrcpassword());
-                }
-            }
-            Owned<IUserDescriptor> udesc=createUserDescriptor();
-            udesc->set(u.str(),p.str());
-            if (!queryDistributedFileDirectory().isSuperFile(srcname,udesc,foreigndali))
+            udesc->set(user, passwd);
+        }
+
+        Owned<IDistributedFile> file = queryDistributedFileDirectory().lookup(srcname, udesc.get());
```
jakesmith: it doesn't look like you're taking into account srcDali here? Does srcname already include ~foreign:: ..?
If not you'll have to build it.

And needs testing to confirm doesn't break foreign copies.

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -2490,32 +2490,30 @@ bool CFileSprayEx::onCopy(IEspContext &context, IEspCopy &req, IEspCopyResponse
...
+        bool supercopy = req.getSuperCopy();
+        if (supercopy)
+        {
+            if (file && !file->querySuperFile())
```
jakesmith: should you test 'file' after lookup, and stop/error ?

jakesmith:
@wangkx - a few comments.

esp/services/ws_fs/ws_fsService.cpp
wangkx: I should.

esp/services/ws_fs/ws_fsService.cpp
wangkx: I will check.

esp/services/ws_fs/ws_fsService.cpp
wangkx: I will test after 160 comes back.

AttilaVamos:
Perhaps the indexlimits (optRemoteRead=true) failure caused by the https://github.com/hpcc-systems/HPCC-Platform/pull/12606 merged into master yesterday.

The  supercopy failures comes from this PR because it changed the underlying behaviour. (That's what I tried to explaining earlier)

wangkx:
I do not know why the Smoketests fail. Let me rebase to the latest 7.4.x and see if the Smoketests still fail.

jakesmith:
>Perhaps the indexlimits (optRemoteRead=true) failure caused by the #12606 merged into master yesterday.

the indexlimits failure can be ignored, there's a fix not yet merged in another PR.

> The supercopy failures comes from this PR because it changed the underlying behaviour. (That's what I tried to explaining earlier)

Does supercopy.ecl test test failure of a super index without preserving the super file structure?
If so, yes it will need changing, as it will no longer fail.

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -2490,32 +2490,48 @@ bool CFileSprayEx::onCopy(IEspContext &context, IEspCopy &req, IEspCopyResponse
...
+            ((strlen(srcname) > 10) && (strnicmp(srcname, "~foreign::", 10) == 0)))
+            file.setown(queryDistributedFileDirectory().lookup(srcname, udesc.get()));
+        else
+        {
+            StringBuffer fileToLookup;
+            fileToLookup.set("foreign::").append(srcDali).append("::");
+            if ((strlen(srcname) > 3) && (strnicmp(srcname, ".::", 3) == 0))
+                fileToLookup.append(srcname + 3);
+            else
+                fileToLookup.append(srcname);
```
jakesmith: You should use CDfsLogicalFileName to manipulate (add foreign) and get the correct format.

jakesmith: See example code that uses CDfsLogicalFileName::setForeign()

wangkx: I thought that there might be some existing code to do that. I just do not know where. I will try the CDfsLogicalFileName ...

jakesmith:
>Does supercopy.ecl test test failure of a super index without preserving the super file structure?
>If so, yes it will need changing, as it will no longer fail.

Ok I see it does, i.e. it tests for the failure when spraying a super index.
@wangkx The test and the key file needs changing. I'd probably leave the test in and reflect the new (successful) result in the key

AttilaVamos:
@wangkx to fix supercopy error:
1. add testing/regress/ecl/key/supercopy.xml to this PR
2. change 
```
<Row><result>Copy superindex as logical file: Fail</result></Row>
```
to
```
<Row><result>Copy superindex as logical file: Pass</result></Row>
```

wangkx:
@jakesmith please review my last 2 commits.

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -2490,32 +2490,41 @@ bool CFileSprayEx::onCopy(IEspContext &context, IEspCopy &req, IEspCopyResponse
...
-            }
-            Owned<IUserDescriptor> udesc=createUserDescriptor();
-            udesc->set(u.str(),p.str());
-            if (!queryDistributedFileDirectory().isSuperFile(srcname,udesc,foreigndali))
+            udesc->set(user, passwd);
+        }
+
+        CDfsLogicalFileName logicalName;
+        logicalName.set(srcname);
+        if (srcDali)
```
jakesmith: minor: should really also be !isEmptyString() - and you could combine this block with the other conditional block

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -2490,32 +2490,41 @@ bool CFileSprayEx::onCopy(IEspContext &context, IEspCopy &req, IEspCopyResponse
...
+        {
+            SocketEndpoint ep(srcDali);
+            logicalName.setForeign(ep,false);
+        }
+
+        Owned<IDistributedFile> file = queryDistributedFileDirectory().lookup(logicalName.get(), udesc.get());
```
jakesmith: tivial: you don't need the .get()'s

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -2490,32 +2490,41 @@ bool CFileSprayEx::onCopy(IEspContext &context, IEspCopy &req, IEspCopyResponse
...
+        if (!file)
+            throw MakeStringException(ECLWATCH_FILE_NOT_EXIST, "Failed to find file: %s", srcname);
```
jakesmith: would be clear if include foreign scope if foreign logical file.
So logicalName.get()

jakesmith:
@wangkx - a couple of minor comments. Please go ahead and squash

esp/services/ws_fs/ws_fsService.cpp
wangkx: Sorry, I should.

esp/services/ws_fs/ws_fsService.cpp
wangkx: Ok

esp/services/ws_fs/ws_fsService.cpp
wangkx: Will switch to logicalName.get().

wangkx:
@jakesmith I made the changes based on your last review. Please review.

jakesmith:
@wangkx - looks good

wangkx:

richardkchapman:
@AttilaVamos Are you happy with this PR now?

AttilaVamos:
@richardkchapman I do not agree with 100%, but I can accept it.

PR #12679, opened 2019-06-18
HPCC-22221 hosts in groups

jakesmith:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/common/environment/environment.cpp
+++ b/common/environment/environment.cpp
@@ -168,7 +168,7 @@ private:
     // NOTE - order is important - we need to construct before p and (especially) destruct after p
     Owned<IRemoteConnection> conn;
     Owned<IPropertyTree> p;
-    mutable MapStringToMyClass<IConstEnvBase> cache;
+    mutable MapStringToMyClass<IConstEnvBase> cache; // NB: map of 'MappingStringToIInterface' that Link's the added IConstEnvBase, and Release's on element removal.
     mutable Mutex safeCache;
     mutable bool dropZoneCacheBuilt;
     mutable bool machineCacheBuilt;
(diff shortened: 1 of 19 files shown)


jakesmith:
@mckellyln - I've done quite a bit of testing now on my mini cluster. I think it's ready for review.
I will leave as separate commits to help the review process for now.

jakesmith:
@mckellyln - rebased to latest master to fix clashes.

Please review.

common/environment/environment.cpp
```diff
@@ -1589,53 +1587,45 @@ IConstMachineInfo * CLocalEnvironment::getMachine(const char * name) const
...
-        {
-            cached = new CConstMachineInfo((CLocalEnvironment *) this, d);
-            setCache(xpath1.str(), cached);
-            setCache(xpath.str(), cached);
-        }
     }
+    if (!d)
+        return nullptr;
+    cached = new CConstMachineInfo((CLocalEnvironment *) this, d);
+    setCache(xpath.str(), cached);
```
mckellyln: Why not add @name to cache as before ?

tools/swapnode/swapnode.cpp
```diff
@@ -268,24 +268,34 @@ int main(int argc, const char *argv[])
...
-                            epa.fromText(ipOrRange, 0);
-                            ForEachItemIn(e, epa)
-                                allEps.append(epa.item(e));
+                            epa.fromText(hostOrIpRange, 0);
+                            if (epa.ordinality()>1)
+                            {
+                                ForEachItemIn(e, epa)
+                                {
+                                    StringBuffer ipStr;
+                                    epa.item(e).getIpText(ipStr);
```
mckellyln: Won't this always return an IP ?  Did you want to preserve the name if it was that ??

jakesmith: yes, but this is if it's detected an IP range (e.g. 192.168.11-15) in the parameters - so it's explicitly dealing with IP's at this point.
If not a range (the else block a few lines below), it doesn't touch it, i.e. it leaves it as a host if that's what it was.

common/environment/environment.cpp
jakesmith: I should have made a comment in the PR to remember why.
I think it's because it's already in the cache.. built by buildMachineCache initially (and whenever dirty)
So adding by name here was pointless before since already there.

Whereas adding by @netAddress still useful as could be different from what it finds, i.e. could be myhost or myhost.domain, or IP..

mckellyln (approved):

mckellyln:
@jakesmith very good PR.
Looks good to merge.
I will create a new JIRA issue for how the local hostname gets resolved to an IP when interface=*, but for now this is handled fine with the use of "."

jakesmith:

PR #12683, opened 2019-06-18
HPCC-22235 DESDL Namespace compatibility

asselitx:
## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Premature optimization

## Smoketest:

### DIFF
--- a/esp/esdllib/esdl_transformer2.cpp
+++ b/esp/esdllib/esdl_transformer2.cpp
@@ -1381,7 +1381,8 @@ void Esdl2Response::processChildNamedResponse(Esdl2TransformerContext &ctx, cons
                 if (ctx.do_output_ns)
                 {
                     ctx.writer->outputBeginNested(out_name, true);
-                    ctx.writer->outputXmlns(out_name, ctx.ns.str());
+                    // This ensures the namespace prefix is just 'xmlns'
+                    ctx.writer->outputXmlns("xmlns", ctx.ns.str());
                     ctx.do_output_ns=false;
                 }
                 else
--- a/esp/services/esdl_svc_engine/esdl_binding.cpp
+++ b/esp/services/esdl_svc_engine/esdl_binding.cpp
@@ -2106,8 +2106,7 @@ int EsdlBindingImpl::HandleSoapRequest(CHttpRequest* request,
             Owned<IPropertyTree> tgtcfg;
             Owned<IPropertyTree> tgtctx;
 
-            ns.clear();
-            generateNamespace(*ctx, request, srvdef->queryName(), mthdef->queryName(), ns);
+            // Echo back the reqeust namespace, don't generate it here
             getSchemaLocation(*ctx, request, schemaLocation);
 
             ctx->setESDLBindingID(m_bindingId.get());


asselitx:
@timothyklemm Could you please review this change?

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -2184,8 +2183,6 @@ StringBuffer & EsdlBindingImpl::generateNamespace(IEspContext &context, CHttpReq
 
     ns.appendf("%s%c%s", m_pESDLService->m_serviceNameSpaceBase.str(), m_pESDLService->m_usesURLNameSpace ? '/' : ':', serv);
 
-    if (method && *method)
```
timothyklemm: This method already includes a condition on the configured namespaceScheme being "scapps". This condition should probably be conditional on it not being "scapps".

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -2106,8 +2106,7 @@ int EsdlBindingImpl::HandleSoapRequest(CHttpRequest* request,
             Owned<IPropertyTree> tgtcfg;
             Owned<IPropertyTree> tgtctx;
 
-            ns.clear();
```
timothyklemm: This should probably always happen. Let the namespace generator, which already has knowledge of scapps ensure a scapps compatible namespace if and only if the service configuration identifies itself as using the scapps namespace scheme.

esp/esdllib/esdl_transformer2.cpp
```diff
@@ -1381,7 +1381,8 @@ void Esdl2Response::processChildNamedResponse(Esdl2TransformerContext &ctx, cons
                 if (ctx.do_output_ns)
                 {
                     ctx.writer->outputBeginNested(out_name, true);
-                    ctx.writer->outputXmlns(out_name, ctx.ns.str());
```
timothyklemm: Not sure what the best answer is here, but this change breaks non-scapps services that could depend on this implementation.

timothyklemm (changes requested):
The JIRA description says "Provide a mechanism to enable compatibility..." It looks like this change forces all ESDL services, including those unrelated to scapps, to conform to the scapps implementation of namespaces. We can't break everybody else for the sake of helping scapps.

asselitx:
> The JIRA description says "Provide a mechanism to enable compatibility..." It looks like this change forces all ESDL services, including those unrelated to scapps, to conform to the scapps implementation of namespaces. We can't break everybody else for the sake of helping scapps.

These changes only affect how the namespace is presented in a method response, while preserving how namespaces are presented in the WSDL/XSD. This way we ensure that the namespace present in the response matches what was submitted in the request. It is not forcing all clients to conform to the scapps style namespace convention.

- If the client was built with a scapps WSDL, and submits requests with that namespace, then that is the namespace used in the response. 

- Otherwise, if a client was build with a full-featured hpcc WSDL,  and submits requests with that namespace, then that full-featured namespace is returned in the response.

esp/esdllib/esdl_transformer2.cpp
asselitx: This is a change but Tony called it out as a bug, and we shouldn't return it like that.

esp/services/esdl_svc_engine/esdl_binding.cpp
asselitx: The _scapps_ style namespace is different even from this behavior. We could implement some configurable behavior (probably extending the range of values of namespaceScheme), but Tony recommended an echo-back categorically because of the complexity of the long-term vision for namespace handling.

esp/services/esdl_svc_engine/esdl_binding.cpp
asselitx: My reply on this is in the pull-level comment below.

esp/services/esdl_svc_engine/esdl_binding.cpp
timothyklemm: Based on offline conversation, this change does not appear to be necessary to achieve the intent of the story. Either the change should not be made (because it can affect the namespace returned in the WSDL) or is acceptable (because it is the removal of dead code).

esp/esdllib/esdl_transformer2.cpp
timothyklemm: Based on offline conversation, this is OK as-is.

esp/services/esdl_svc_engine/esdl_binding.cpp
timothyklemm: Based on offline conversation, this is OK as-is.

esp/services/esdl_svc_engine/esdl_binding.cpp
asselitx: You are correct- removing this code was an error. I'm patching it back in.

asselitx:
@afishbeck could you please review- Tim has finished.

asselitx:
@afishbeck could you please review- Tim has finished.

afishbeck (approved):
@asselitx looks good.

afishbeck:

PR #12684, opened 2019-06-18
HPCC-22240 Support ESDL binding feature authorization

timothyklemm:
* Add "auth_feature" attribute to Binding/Definition element.
* Add "auth_feature" and "override_service_auth_feature" attributes to
  Method element.
* If the binding declares security, even to say no security, the binding takes
  precedence. If the binding does not declare security, the ESDL definition
  security, if any, is applied.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/esdllib/esdl_def.cpp
+++ b/esp/esdllib/esdl_def.cpp
@@ -1009,6 +1009,8 @@ static void parseAccessList(const char * rawServiceAccessList, MapStringTo<SecAc
                 if (currAccessName.isEmpty())
                     currAccessName.setf("%sAccess",  defaultaccessname);
             }
+            else if (isEmptyString(currAccessName)) // an empty string will have length 1 for the NULL terminator
+                continue;
 
             for (;currIndex <= entrylen; currIndex++ )
             {
--- a/esp/services/esdl_svc_engine/esdl_binding.cpp
+++ b/esp/services/esdl_svc_engine/esdl_binding.cpp
@@ -411,10 +411,99 @@ static void ensureMergeOrderedEsdlTransform(Owned<IPropertyTree> &dest, IPropert
     mergePTree(dest, copy);
 }
 
+void EsdlServiceImpl::configureBindingAccessMap(const IPTree* node, const char* service, const char* method, const char* serviceSecurity)
+{
+    StringBuffer methodSecurity;
+    StringArray  accessList;
+    StringBuffer accessName;
+    StringBuffer accessLevel;
+    Owned<MethodAccessMap> methodAccess;
+
+    if (!isEmptyString(serviceSecurity))
+    {
+        // Service and method access requirements are combined by default, but the method may opt
+        // to ignore the service requirements.
+        //
+        // Consider a service level requirement of "{$method}Access" that applies a method-specific
+        // requirement to every method. Some methods, such as Ping or EchoTest, may require no
+        // security. The 'inherited' requirement may be incorrect for other methods, perhaps because
+        // the name is too long. Methods must be able to override the service's settings.
+        if (!node->getPropBool("@override_service_auth_feature", false))
+            methodSecurity.set(serviceSecurity).append(',');
+    }
+    methodSecurity.append(node->queryProp("@auth_feature"));
+    methodSecurity.replaceString("{$service}", service);
+    methodSecurity.replaceString("{$method}", method);
+
+    accessList.appendList(methodSecurity, ",");
+    if (accessList.ordinality() == 0)
+    {
+        // Either the binding's service and method both omit the auth_feature attribute or service
+        // level security is given and the method explicitly overrides it without specifying its
+        // own entries. The binding defers.
+        return;
+    }
+
+    // At least one security entry exists in the binding. The binding will define a method access
+    // map (even if empty) so it can take precedence.
+
+    methodAccess.setown(new MethodAccessMap());
+    m_bindingAccessMaps.setValue(method, methodAccess.getLink());
+    ForEachItemIn(idx, accessList)
+    {
+        auto accessEntry = accessList.item(idx);
+        auto entryLen = strlen(accessEntry);
+        decltype(entryLen) entryIdx = 0;
+        SecAccessFlags flag = SecAccess_Unknown;
+
+        accessName.clear();
+        for (; entryIdx <= entryLen && accessEntry[entryIdx] != ':'; entryIdx++)
+        {
+            if (!isspace(accessEntry[entryIdx]))
+                accessName.append(accessEntry[entryIdx]);
+        }
+        if (strieq(accessName, "None") || strieq(accessName, "Deferred"))
+        {
+            // No map entry is created for this list item, but other list items may still yield
+            // map entries. It is the binding author's responsibility to use "none" logically.
+            continue;
+        }
+        if (':' == accessEntry[entryIdx])
+        {
+            entryIdx++;
+            if (accessName.isEmpty())
+                accessName.setf("%sAccess", service);
+        }
+        else if (isEmptyString(accessName)) // an empty string will have length 1 for the NULL terminator
+            continue;
+
+        accessLevel.clear();
+        for (; entryIdx <= entryLen; entryIdx++)
+        {
+            if (!isspace(accessEntry[entryIdx]))
+                accessLevel.append(accessEntry[entryIdx]);
+        }
+
+        switch (flag = getSecAccessFlag(accessLevel))
+        {
+        case SecAccess_Unknown:
+            DBGLOG("Unknown access level: %s", accessLevel.str());
+            throw MakeStringException(0, "Unknown access level: %s", accessLevel.str());
+        case SecAccess_Unavailable:
+            methodAccess->setValue(accessName, SecAccess_Full);
+            break;
+        default:
+            methodAccess->setValue(accessName, flag);
+            break;
+        }
+    }
+}
+
 void EsdlServiceImpl::configureTargets(IPropertyTree *cfg, const char *service)
 {
-    VStringBuffer xpath("Definition[@esdlservice='%s']/Methods", service);
-    IPropertyTree *target_cfg = cfg->queryPropTree(xpath.str());
+    VStringBuffer xpath("Definition[@esdlservice='%s']", service);
+    IPropertyTree* definition_cfg = cfg->queryPropTree(xpath);
+    IPropertyTree *target_cfg = definition_cfg->queryPropTree("Methods");
     DBGLOG("ESDL Binding: configuring method targets for esdl service %s", service);
 
     if (target_cfg)
@@ -456,6 +545,9 @@ void EsdlServiceImpl::configureTargets(IPropertyTree *cfg, const char *service)
 
         MapStringToMyClass<IEspPlugin> localCppPluginMap;
 
+        StringBuffer serviceSecurity(definition_cfg->queryProp("@auth_feature"));
+        m_bindingAccessMaps.kill();
+
         Owned<IPropertyTreeIterator> iter = m_pServiceMethodTargets->getElements("Target");
         ForEach(*iter)
         {
@@ -463,6 +555,8 @@ void EsdlServiceImpl::configureTargets(IPropertyTree *cfg, const char *service)
             const char *method = methodCfg.queryProp("@name");
             if (!method || !*method)
                 throw MakeStringException(-1, "ESDL binding - found target method entry without name!");
+
+            configureBindingAccessMap(&methodCfg, service, method, serviceSecurity);
 
             m_methodCRTransformErrors.remove(method);
             m_customRequestTransformMap.remove(method);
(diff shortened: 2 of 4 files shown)


timothyklemm:
@afishbeck Please review. I don't think my changes explain the regression test failure.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -411,10 +411,99 @@ static void ensureMergeOrderedEsdlTransform(Owned<IPropertyTree> &dest, IPropert
...
+    // At least one security entry exists in the binding. The binding will define a method access
+    // map (even if empty) so it can take precedence.
+
+    methodAccess.setown(new MethodAccessMap());
+    m_bindingAccessMaps.setValue(method, methodAccess.getLink());
+    ForEachItemIn(idx, accessList)
+    {
+        auto accessEntry = accessList.item(idx);
+        auto entryLen = strlen(accessEntry);
+        decltype(entryLen) entryIdx = 0;
```
afishbeck: Clever but at the cost of readability.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -411,10 +411,99 @@ static void ensureMergeOrderedEsdlTransform(Owned<IPropertyTree> &dest, IPropert
...
+            // map entries. It is the binding author's responsibility to use "none" logically.
+            continue;
+        }
+        if (':' == accessEntry[entryIdx])
+        {
+            entryIdx++;
+            if (accessName.isEmpty())
+                accessName.setf("%sAccess", service);
+        }
+        else if (isEmptyString(accessName)) // an empty string will have length 1 for the NULL terminator
```
afishbeck: Comment is still confusing.  Maybe because byte length and string length are different concepts.  I don't think the comment adds value.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -411,10 +411,99 @@ static void ensureMergeOrderedEsdlTransform(Owned<IPropertyTree> &dest, IPropert
...
+            continue;
+
+        accessLevel.clear();
+        for (; entryIdx <= entryLen; entryIdx++)
```
afishbeck: You could write this loop as a common inline function getAccessEntryText(len, entryStr, terminator) and share for both this and the above loop.  Or even better even if you made the parsing bits one function, and then do the checks like:

 if (strieq(accessName, "None") || strieq(accessName, "Deferred"))
 if (hasValue) //(':' == accessEntry[entryIdx])
 else if (isEmptyString(accessName))

After parsing.  It might be a tiny bit wasteful in some scenarios, but could be much more readable.

system/security/shared/seclib.hpp
```diff
@@ -78,6 +78,23 @@ inline const char * getSecAccessFlagName(SecAccessFlags flag)
     }
 }
 
+inline SecAccessFlags getSecAccessFlag(const char* name)
+{
+    if (isEmptyString(name))
+        return SecAccess_Unavailable;
```
afishbeck: Term unavailable seems slightly confusing, but you could convince me.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -456,6 +545,9 @@ void EsdlServiceImpl::configureTargets(IPropertyTree *cfg, const char *service)
 
         MapStringToMyClass<IEspPlugin> localCppPluginMap;
 
+        StringBuffer serviceSecurity(definition_cfg->queryProp("@auth_feature"));
+        m_bindingAccessMaps.kill();
```
afishbeck: Just checking, but you made sure this doesn't leak the contents?

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -551,6 +645,75 @@ static inline bool isPublishedQuery(EsdlMethodImplType implType)
     return (implType==EsdlMethodImplRoxie || implType==EsdlMethodImplWsEcl);
 }
 
+void EsdlServiceImpl::handleFeatureAuthorization(IEspContext& context, IEsdlDefMethod& esdlDef)
+{
+    auto method = esdlDef.queryName();
```
afishbeck: Convince me that in these places auto is helpful and not just a shortcut that isn't self documenting.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -551,6 +645,75 @@ static inline bool isPublishedQuery(EsdlMethodImplType implType)
...
+    auto bindingMethodAccessMap = m_bindingAccessMaps.getValue(method);
```
afishbeck: Below you make some pretty big assumptions about the type for not knowing what it is ahead of time. (auto).  And I found it hard to determine exactly what that type was.  Mostly because the type itself is owned rather than having the map just own its entries without them using an owned wrapper.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -551,6 +645,75 @@ static inline bool isPublishedQuery(EsdlMethodImplType implType)
...
+        accessMap = bindingMethodAccessMap->get();
+#ifdef DEBUG
+        accessMapSource = "binding";
+#endif
+    }
+    else
+    {
+        // The absence of an access map from the binding defers the security specification
+        // to the ESDL definition.
+        accessMap = &const_cast<MethodAccessMap&>(esdlDef.queryAccessMap());
```
afishbeck: Often demonstrates the the use of const was wrong, either in the source or destination, if you have to cast it away.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -551,6 +645,75 @@ static inline bool isPublishedQuery(EsdlMethodImplType implType)
...
+                const char * key = (const char *)cur.getKey();
+                features.appendf("%s%s:%s", (index++ == 0 ? "" : ", "), key, getSecAccessFlagName(*accessMap->getValue(key)));
+            }
+            const char * user = context.queryUserId();
+            throw MakeStringException(-1, "%s::%s access denied for user '%s' - Method requires: '%s'.", context.queryServiceName(nullptr), method, !isEmptyString(user) ? user : "Anonymous", features.str());
+        }
+#ifdef DEBUG
+        else
+        {
+            StringBuffer features;
```
afishbeck: Might want to check the log level before calculating all of this if ESPLOG(LogMax will just discard the entry and make the work useless.

afishbeck (commented):
@timothyklemm some comments.

esp/esdllib/esdl_def.cpp
```diff
@@ -1009,6 +1009,8 @@ static void parseAccessList(const char * rawServiceAccessList, MapStringTo<SecAc
                 if (currAccessName.isEmpty())
                     currAccessName.setf("%sAccess",  defaultaccessname);
             }
+            else if (isEmptyString(currAccessName)) // an empty string will have length 1 for the NULL terminator
```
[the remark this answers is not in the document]
timothyklemm: The loop terminator on line 1000, which I didn't write, enables the accessEntry NULL terminator to be added to currAccessName as if it was any other character. currAccessName.length() returns 1, and currAccessName.isEmpty() returns false. The comment was an (apparently) unsuccessful attempt to explain why I didn't use the the isEmpty method that was used three lines above.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -411,10 +411,99 @@ static void ensureMergeOrderedEsdlTransform(Owned<IPropertyTree> &dest, IPropert
...
+    if (!isEmptyString(serviceSecurity))
+    {
+        // Service and method access requirements are combined by default, but the method may opt
+        // to ignore the service requirements.
+        //
+        // Consider a service level requirement of "{$method}Access" that applies a method-specific
+        // requirement to every method. Some methods, such as Ping or EchoTest, may require no
+        // security. The 'inherited' requirement may be incorrect for other methods, perhaps because
+        // the name is too long. Methods must be able to override the service's settings.
+        if (!node->getPropBool("@override_service_auth_feature", false))
```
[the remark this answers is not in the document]
timothyklemm: The name is for the binding author that is defining security. Being a completely new feature, I'm not sure how often will will ultimately be used.

The ESDL definition used auth_feature, so I kept the name in the binding so people would understand the same concept was being expressed in both places. When I realized I wanted the method to be able to replace a service level specification I needed a name that conveyed its purpose without being ambiguous if we every defined something with similar usage. I'm open to suggestions if it is too verbose.

esp/services/esdl_svc_engine/esdl_binding.cpp
timothyklemm: To reiterate my previous comment, it is an attempt to explain why I passed a StringBuffer, which has isEmpty, to isEmptyString when isEmpty was used three lines earlier. The value I see in the comment is that if I was reviewing the code I would object to the use of isEmptyString on a StringBuffer, and the comment suggests there was a reason for doing it.

esp/services/esdl_svc_engine/esdl_binding.cpp
timothyklemm: I basically copied the esdl_def code, with minor modifications. I'll look at reworking this, which will also eliminate the need for isEmptyString instead of isEmpty.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -411,10 +411,99 @@ static void ensureMergeOrderedEsdlTransform(Owned<IPropertyTree> &dest, IPropert
...
+    }
+}
+
 void EsdlServiceImpl::configureTargets(IPropertyTree *cfg, const char *service)
 {
-    VStringBuffer xpath("Definition[@esdlservice='%s']/Methods", service);
-    IPropertyTree *target_cfg = cfg->queryPropTree(xpath.str());
+    VStringBuffer xpath("Definition[@esdlservice='%s']", service);
+    IPropertyTree* definition_cfg = cfg->queryPropTree(xpath);
+    IPropertyTree *target_cfg = definition_cfg->queryPropTree("Methods");
```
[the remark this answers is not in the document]
timothyklemm: Probably should.

esp/services/esdl_svc_engine/esdl_binding.cpp
timothyklemm: I agree completely. IEsdlDefMethod returns const and IEspContext requires non-const. If you check original line 575 you'll see that this cast was necessary before my changes.

system/security/shared/seclib.hpp
timothyklemm: translateAuthLevel, in esdl_def.cpp, maps an empty string to SecAccess_Full, and EsdlServiceImpl::configureBindingAccessMap maps SecAccess_Unavailable to SecAccess_Full. My security manager plugin, however, considers SecAccess_Unavailable to be an error.

Maybe the plugin is wrong and should also treat SecAccess_Unavailable like SecAccess_Full (hint: this is a request for an opinion). But I'm not convinced that a generic utility should assume that every caller wants this equivalence. The generic behavior should indicate that nothing was given, allowing the caller to treat the condition in whatever way it chooses.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -411,10 +411,99 @@ static void ensureMergeOrderedEsdlTransform(Owned<IPropertyTree> &dest, IPropert
...
+
+    // At least one security entry exists in the binding. The binding will define a method access
+    // map (even if empty) so it can take precedence.
+
+    methodAccess.setown(new MethodAccessMap());
+    m_bindingAccessMaps.setValue(method, methodAccess.getLink());
+    ForEachItemIn(idx, accessList)
+    {
+        auto accessEntry = accessList.item(idx);
+        auto entryLen = strlen(accessEntry);
```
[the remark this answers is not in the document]
timothyklemm: In this case, strlen is unlikely to be replaced with something that returns a type other than size_t, so this may not be the best example of another good use case for auto (and the following use of decltype). This other use case is when the code is not dependent on a specific type. Is it const or not const, signed or unsigned, 32-bit or 64-bit, Auto provides an appropriate type, and decltype ensures a compatible value that should never generate a compiler warning.

I'll change (both).

esp/esdllib/esdl_def.cpp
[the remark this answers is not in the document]
afishbeck: Ah, I see, but what this really does is hide some strange behavior by using a char * function where we would normally use the member function of the current object.  I'd rather see the loop at line 1000 fixed even though it's an existing weird behavior.

esp/services/esdl_svc_engine/esdl_binding.cpp
afishbeck: Again, I'd rather the weirdness be fixed, but if not the comment should clarify why the extra null is there, so the real reason can be looked at and understood in the future.

esp/services/esdl_svc_engine/esdl_binding.cpp
afishbeck: Understood.  Just making sure it was thought about.  When const_cast is appropriate is when something is meant to be const for just about everyone, but you are well aware you are the one case that wants to change the contents.

system/security/shared/seclib.hpp
afishbeck: Let's discuss this afternoon.  At the level of this function my main concern was that if you squint your eyes hard enough, the term "Unavailable" starts to sound like it has a real security meaning.  Rather than just meaning that the string was empty.  But I'd like to better understand each context that this might be called now and in the future.

esp/services/esdl_svc_engine/esdl_binding.cpp
jakesmith: Saw this going by, const_cast is nasty.. I suspect the original code on line 575 was abusing it too.

>..the one case that wants to change the contents.

I don't think anything is changing the content.. 
Looking at how 'accessMap' is being used here, it looks like 'validateFeaturesAccess' is the only method that is requiring it to be non-const, and having had a look at it, it looks like the map it accepts as an argument should be const. i.e. validateFeaturesAccess is not const correct (it does not alter the object that's passed into it).

So I think the correct fix is to make validateFeaturesAccess accept a const and then you wouldn't need the const_cast kludge.

esp/services/esdl_svc_engine/esdl_binding.cpp
timothyklemm: Auto doesn't mean I don't have to know how to use the value. It simply means I don't have to know (or type) its name. Not an argument for keeping it; just a response to the "big assumptions" comment.

esp/services/esdl_svc_engine/esdl_binding.cpp
timothyklemm: I've run valgrind. Unless I'm completely misreading the output, no leaks are reported with configureTargets in the allocation stack trace. Is there something else I should be looking for?

esp/services/esdl_svc_engine/esdl_binding.cpp
afishbeck: Understood.  But someone reading the code should also be able to easily figure it out.  There are places where auto really makes sense or is needed.  Anywhere else, I would say error on the side of self documenting code.

timothyklemm:
HPCC-22458 effectively invalidates these changes without invalidating the intent. Closing this request.

PR #12695, opened 2019-06-19
HPCC-22356 Child query workunit result writer used in wrong context

jakesmith:
Also if it is used and has 0 rows, its result was never published.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/thorlcr/activities/wuidwrite/thwuidwrite.cpp
+++ b/thorlcr/activities/wuidwrite/thwuidwrite.cpp
@@ -38,32 +38,6 @@ protected:
     StringAttr resultName;
     unsigned resultSeq;
 
-public:
-    CWorkUnitWriteMasterBase(CMasterGraphElement * info) : CMasterActivity(info)
-    {
-        numResults = 0;
-        totalSize = 0;
-        resultSeq = 0;
-        appendOutput = false;
-        flushThreshold = -1;
-        workunitWriteLimit = 0;
-        mpTag = container.queryJob().allocateMPTag(); // used by local too
-        activityMaxSize = 0;
-    }
-    virtual void init()
-    {
-        CMasterActivity::init();
-        // In absense of OPT_OUTPUTLIMIT check pre 5.2 legacy name OPT_OUTPUTLIMIT_LEGACY
-        workunitWriteLimit = activityMaxSize ? activityMaxSize : getOptInt(OPT_OUTPUTLIMIT, getOptInt(OPT_OUTPUTLIMIT_LEGACY, defaultDaliResultLimit));
-        if (workunitWriteLimit>defaultDaliResultOutputMax)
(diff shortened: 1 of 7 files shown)


thorlcr/activities/wuidwrite/thwuidwrite.cpp
```diff
@@ -280,7 +280,7 @@ class CWorkUnitWriteLocalActivityMaster : public CWorkUnitWriteMasterBase
 
 CActivityBase *createWorkUnitWriteActivityMaster(CMasterGraphElement *container)
 {
-    if (container->queryLocalOrGrouped())
+    if (container->queryOwner().isLocalChild())
```
jakesmith: I think this change is correct, CWorkUnitWriteLocalActivityMaster and it's counterpart in the slave (CWorkUnitWriteLocalSlaveActivity), is used in the context of a CQ, where the partial results are streamed differently.

But prior to this change, it was also being used if the activity was marked 'grouped' when it should not have been.

thorlcr/activities/wuidwrite/thwuidwriteslave.cpp
```diff
@@ -174,27 +174,24 @@ class CWorkUnitWriteLocalSlaveActivity : public CWorkUnitWriteSlaveBase
                 ActPrintLog("WORKUNITWRITE: exception");
                 throw;
             }
-            if (numGot || totalNum)
```
jakesmith: This is the main fix, essentially removing this line, the other changes are just a indentation change
With this condition, it was not communicating with the master if there were 0 total rows, meaning the result was never published.

jakesmith:
@ghalliday - ready for review, and have tested manually with the reports archive, which does generate a workunit result in a child query. However I don't seem to be able to get my test query to generate one in a child query to test it further.

jakesmith:
@ghalliday - have now added a regression test. Please review.

thorlcr/activities/wuidwrite/thwuidwriteslave.cpp
jakesmith: this is not really the main fix actually - it should never have been using the CQ form of workunit write in the context this problem was hit, it was spuriously using this version because of  'grouped=1' in the write activity, which was wrongly interpreted as meaning in CQ context. That is fixed by the other changes.

jakesmith:
Reworked to avoid to only communicate to master if there are any results, and ensure result published when [parent] query is complete.
There was no mechanism for the master activities of CQ activities (which are created on demand) to know when done before. This change includes calling done() on all CQ activities.
The only kind that does anything is this reworked CQ workunit write, which now published the result if necessary.

@ghalliday - please review

jakesmith:
@ghalliday - this is an ordering issue, and in offline chat, we agree that order is okay, given it's >1 CQ updating(extending) the same result. Now it is the whole of one CQ's result followed by the other, before it was per iteration. This change makes the result update much more efficient avoiding an update until all is done.

I've modified apply3.ecl to have each extend in it's own CQ, but ordered by an ORDERED.

@ghalliday - please review

PR #12709, opened 2019-06-21
HPCC-22312 Failing Std.Date unit tests on macOS

dcamper:
This change specifically involves the %Y formatting directive for year output (not parsing).

Complete list of Date functions that were affected:

DateToString
SecondsToString
TimestampToString
ConvertDateFormat
ConvertDateFormatMultiple

Also included with this change is a new __OS__ ECL constant value:  'linux-macos'.

## Type of change:
- [x] This change is a breaking change (fix or feature that will cause existing behavior to change).

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

## Testing:

Manual testing plus exercising (new and modified) tests in TestStd.Date.TestFormat.

### DIFF
--- a/ecl/hql/hqllex.l
+++ b/ecl/hql/hqllex.l
@@ -1044,6 +1044,8 @@ __OS__              {
                             return lookupIdentifierToken(returnToken, lexer, lookup, activeState, CUR_TOKEN_TEXT);
 #ifdef _WIN32
                         returnToken.setExpr(createConstant("windows"));
+#elif defined(__APPLE__)
+                        returnToken.setExpr(createConstant("macos"));
 #else
                         returnToken.setExpr(createConstant("linux"));
 #endif
--- a/ecllibrary/teststd/Date/TestFormat.ecl
+++ b/ecllibrary/teststd/Date/TestFormat.ecl
@@ -77,6 +77,11 @@ EXPORT TestFormat := MODULE
     ASSERT(Date.DateToString(19700101, '%d/%m/%y') = '01/01/70');
     ASSERT(Date.DateToString(20110302, '%d %b %Y') = '02 Mar 2011');
     ASSERT(Date.DateToString(20111202, '%d %B %Y') = '02 December 2011');
+#IF(__OS__ = 'macos')
+    ASSERT(Date.DateToString(110131, '%Y-%m-%d') = '0011-01-31');
+#ELSE
+    ASSERT(Date.DateToString(110131, '%Y-%m-%d') = '11-01-31');
+#END
 
     ASSERT(Date.MatchTimeString('123456',TimeFormats) = 123456);
     ASSERT(Date.MatchTimeString('12:34:56',TimeFormats) = 123456);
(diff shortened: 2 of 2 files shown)


dcamper:
@richardkchapman @ghalliday Please review.

ecl/hql/hqllex.l
```diff
@@ -1044,6 +1044,8 @@ __OS__              {
                             return lookupIdentifierToken(returnToken, lexer, lookup, activeState, CUR_TOKEN_TEXT);
 #ifdef _WIN32
                         returnToken.setExpr(createConstant("windows"));
+#elif defined(__APPLE__)
+                        returnToken.setExpr(createConstant("linux-macos"));
```
ghalliday: @richardkchapman is this a sensible value?

ecllibrary/std/Date.ecl
```diff
@@ -56,6 +56,47 @@ END;
...
+ * precision format where needed.
+ *
+ * As of this writing, here is how our supported operating systems natively
+ * deal with small years (e.g. 11) and the %Y directive by default:
+ *
+ * macOS:	0011
+ * Windows:	0011
+ * Linux:	11
+ *
+ * All support Linux distributions have glibc, which provides support for
```
ghalliday: What happens if the system is compiled using clang?

ghalliday (commented):
@richardkchapman added a couple of questions

ecllibrary/std/Date.ecl
dcamper: Good question.  I installed Clang 7.0 on my Ubuntu 18.04 system and used (basically) the same code I used on Windows for testing (which is just spitting out the formatted year value from a valid struct tm).  The result was identical to gcc's result:  Using %Y, small years were returned without leading zeroes, and the precision specifier was supported (so %04Y did pad the result).

TL;DR version:  Clang appears to work the same as gcc, which makes sense since it tries to emulate the Gnu toolchain.  Apple may have modified their version of LLVM/Clang so that the leading zeroes are included.

ecl/hql/hqllex.l
richardkchapman: Not really. MacOS is not linux or linux-based, though it shares some similarity. I don't see how returning this value helps any with back-compatibility - existing code is still going to have to change - so we may as well get it right.

richardkchapman:
Has this been tried on Windows?

I'm not at all convinced that this is moving in the right direction, nor that it won't cause major compatibility headaches for existing code (99.99% of it running on Linux) that was relying on the previous behaviour. I suspect we should be changing MacOS implementation to return 2 digit dates (or marking as wontfix, if that's too hard)

dcamper:
It has been tested on Windows.

Richard, I agree that this feels like a heavy-handed fix.  It does normalize the returned year value, but what changes affects the majority of our actual installations and it is bit hackish on top of that.  It does not feel worth it.

However, rather than just marking the issue as Won't Fix leaves the Std.Date ECL unit tests failing on macOS.

What do you think about changing this PR so that:

1. We add a (more reasonable) __OS__ value for macOS
2. Use __OS__ to make OS-level unit tests for those that will fail due to this year inconsistency, so they will always pass on all supported platforms
3. Perhaps document the inconsistency, if it is deemed important enough

Thoughts?

dcamper:
Note:  If this PR is accepted then documentation for __OS__ in the language reference manual will need to be updated.

richardkchapman:
There is also thorlib.os() that should perhaps be changed? Can be a separate Jira though

PR #12713, opened 2019-06-21
HPCC-22403 checkTag no longer matches if there are trailing chars

shamser:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:


testing/regress/hpcc/util/ecl/file.py
```diff
@@ -337,7 +337,7 @@ def __checkTag(self,  tag):
         retVal = False
         eclText = open(self.getEcl(), 'rb')
         for line in eclText:
-            if tag in line.lower():
+            if re.search(tag+r'[^_\w]', line, re.IGNORECASE):
```
AttilaVamos: I am not convinced about this is correct. 
if tag = 'nosetup' and the line is '//nosetup' then the search(tag+r'[^_\w]', line, re.IGNORECASE) has no result.
Same as if the line is '//nosetup_roxie', there is no result. In both case the condition is false.

I would use this if I want to exact tag match:
```
re.search('^\/\/'+tag+'$', line, re.IGNORECASE)
```
BTW is this '//nosetup' and/or '//nosetup_roxie' is a new tag? I can't remember I did saw them before. Regression test engine definitely doesn't use them. What they are for?

AttilaVamos (changes requested):
@shamser back to you.

shamser:
@AttilaVamos Please can you review

testing/regress/hpcc/util/ecl/file.py
shamser: I have just tried it and it does work - please try it out. The tag parameter already contains the '//' prefixed, so it is already searching for line starting with '//'.  The regular expression '[^_\w]'  means that match a tag as long it does not end with _ or any word character (a-z, A-Z, 0-9, _).   I have just realized that I can remove the '_' character as it is already included as part of '\w'.  

The scripts are already searching for these tags.  I found out about the tags by running the script in debug.  Here's the extract from the debug log: 
[Debug]  -1. testExclusion (ecl:'setup.ecl', target: 'setup', tag:'//nosetup')
**[Debug]  -1.__checkTag (ecl:'setup.ecl', tag:'//nosetup')**
[Debug]  -1.__checkTag() returns with False
[Debug]  -1. testExclude() returns with: False
**[Debug]  -1. testExclusion (ecl:'setup.ecl', target: 'setup_roxie', tag:'//nosetup_roxie')**
[Debug]  -1.__checkTag (ecl:'setup.ecl', tag:'//nosetup_roxie')
[Debug]  -1.__checkTag() returns with False

As you can see, it checks first for //nosetup and then for //nosetup_roxie (in this case it was handling target Roxie).

I need these change for a new setup script (relating to a new regression) that will only run on thor.

shamser: I fixed some regression queries that were failing because it had tag //nothorlcr rather than //nothor, as required for exact matches of the tag.

AttilaVamos: Yes, you are right, the tag already contains '//'. However there are two more things:
1. Who handles the //nosetup_roxie tag?
2. We already have '//no<engine>'' tags to exclude execution on a given engine and it work with setup cases as well (see: roxie_echo.ecl). Why we need one more?

AttilaVamos:
@shamser I have two more questions (see above).

testing/regress/hpcc/util/ecl/file.py
shamser: 1) The new WU Analysis regression
2) Tags such as '//noroxie' and '//nothor'  is ignored when executed with ./ecl-test setup.

shamser: ./ecl-test  setup
[Action] Suite: thor (setup)
[Action] Queries: 11
[Action]   1. Test: roxie_echo.ecl
[Pass]   1. Pass roxie_echo.ecl - W20190624-120827 (0 sec)
[Pass]   1. URL http://127.0.0.1:8010/?Widget=WUDetailsWidget&Wuid=W20190624-120827

shamser: @AttilaVamos

AttilaVamos:
@shamser ok, fine.

shamser:
@ghalliday Please can you merge.

testing/regress/ecl/sqnormds1_an.ecl
```diff
@@ -15,6 +15,7 @@
     limitations under the License.
 ############################################################################## */
 
+//nothor
 //nothorlcr
```
ghalliday: Should this be deleted as no longer needed?

shamser: I'm not sure.  What's target nothorlcr?

ghalliday: It was used when we had two different thor implementations - one that used link counted rows (lcr) and one that didn't.
@AttilaVamos is nothorlcr ever changed anymore?

AttilaVamos: @ghalliday no, I don't think.

shamser:
@ghalliday I've removed //nothorlcr

ghalliday:
@shamser please squash and I will merge

PR #12725, opened 2019-06-25
HPCC-22415 Remove redundant security checking from dali

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dali/base/dasess.cpp
+++ b/dali/base/dasess.cpp
@@ -488,24 +488,6 @@ public:
     }
 };
 
-bool createDaliSignature(const char * scope, IUserDescriptor *udesc, CDateTime &now, StringBuffer &b64sig)
-{
-    IDigitalSignatureManager * pDSM = queryDigitalSignatureManagerInstanceFromEnv();
-    if (pDSM && pDSM->isDigiSignerConfigured())
-    {
-        StringBuffer username;
-        udesc->getUserName(username);
-        StringBuffer timeStr;
-        now.setNow();
-        now.getString(timeStr, false);//get UTC timestamp
-        VStringBuffer toSign("%s;%s;%s", scope, username.str(), timeStr.str());
-
-        pDSM->digiSign(b64sig, toSign);//Sign "scope;username;timeStamp"
-        return true;
-    }
-    return false;
-}
-
 class CSessionRequestServer: public Thread
 {
     bool stopped;
--- a/dali/base/dasess.cpp
+++ b/dali/base/dasess.cpp
@@ -120,7 +120,7 @@ interface ISessionManagerServer: implements IConnectionMonitor
     virtual void addSession(SessionId id) = 0;
     virtual SessionId lookupProcessSession(INode *node) = 0;
     virtual INode *getProcessSessionNode(SessionId id) =0;
-    virtual SecAccessFlags getPermissionsLDAP(const char *key,const char *obj,IUserDescriptor *udesc,unsigned flags, const char * reqSignature, CDateTime & reqUTCTimestamp, int *err)=0;
+    virtual SecAccessFlags getPermissionsLDAP(const char *key,const char *obj,IUserDescriptor *udesc,unsigned flags, int *err)=0;
     virtual bool clearPermissionsCache(IUserDescriptor *udesc) = 0;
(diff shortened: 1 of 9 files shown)


ghalliday:
@RussWhitehead and @jakesmith please review.

dali/base/dasess.cpp
```diff
@@ -986,21 +986,10 @@ class CClientSessionManager: public CSessionManagerBase, implements IConnectionM
...
-                }
-            }
-            else
-            {
-                mb.append(reqSignature);
-                reqUTCTimestamp.serialize(mb);
-            }
+            //Backwards compatibility in case another parameter is added later, otherwise could be removed
+            CDateTime reqUTCTimestamp;
+            mb.append("");
```
jakesmith: trivial: worth a comment saying what legacy serialized field this was // base64 signature.

jakesmith:
@ghalliday - looks ok afaics. I think it would be worth adding some notes to the JIRA descriptor and PR commit message to elaborate on why removing, e.g. whitelist..

@RussWhitehead - tagging you for review.

RussWhitehead:
@ghalliday The changes look reasonable, but I am not clear why this code is no longer necessary. Can you update the Jira or PR?  Also @afishbeck should be aware of this change

ghalliday:
@RussWhitehead the question is what do we gain from this code being present?  I'll update the jira, but if you can think of a good reason for keeping it then it should stay.

ghalliday:
@jakesmith I had some local changes I had forgot to delete - included in the second commit.

jakesmith:
@ghalliday - createDaliSignature not called now, so looks fine.

Tagging @RussWhitehead - for Gavin's question above.

dali/server/daldap.cpp
```diff
@@ -147,80 +147,10 @@ class CDaliLdapConnection: implements IDaliLdapConnection, public CInterface
...
-                pDSM = queryDigitalSignatureManagerInstanceFromEnv();
-            if (pDSM && pDSM->isDigiVerifierConfigured())
-            {
-                StringBuffer requestTimestamp;
-                reqUTCTimestamp.getString(requestTimestamp, false);//extract timestamp string from Dali request
-
-                CDateTime now;
-                now.setNow();
-                CDateTime daliTime(now);
-                if (requestSignatureAllowedClockVarianceSeconds)//allow for clock variance between machines
```
RussWhitehead: Need to remove this feature from the configuration

ghalliday:
@RussWhitehead thanks for the reminder.  Please check the latest commit.

RussWhitehead (commented):
@ghalliday Looks good, but what happens when a client engine with your changes calls a Dali running an older 7.x, which expects a signature in the request?

ghalliday:
@RussWhitehead 
Good question.  I think if the signature and the password are blank then the old code would treat the user as authenticated.  (Please can you double check my logic.) 

There might be a problem if a non-workunit token password was passed to an old dali.   That shouldn't ever happen when connecting from a 7.x system since the password isn't stored.
I will double check though - please can you also check the logic.

RussWhitehead:
@ghalliday It looks like you are correct, a missing signed token will be ignored by 7.2.x.   I do recommend testing it since a failure will cause production failures

ghalliday:
@jakesmith do you have a simple way of testing this?

jakesmith:
@ghalliday - doing : 
```
daliadmin <dali-ip> dfsverify <logicalfilename>
```

with theses changes to a 7.2.x dali should test it.

ghalliday:
@richardkchapman I have checked the command that Jake suggested worked.  Please review carefully.

richardkchapman:
@afishbeck Are you happy with this?

afishbeck:
@richardkchapman yes, Russ and I had discussed this offline.  We concluded the premise of this JIRA was correct.  This checking wasn't providing benefit.  Whitelists and in the future PKI based component trust will provide better intra-component security.  End user authentication and permissions will happen at ESP.

PR #12728, opened 2019-06-25
HPCC-22396 Arrange WU Analysis for automated regression

shamser:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/testing/wuanalysis/ecl/anahashdistrib1.ecl
+++ b/testing/wuanalysis/ecl/anahashdistrib1.ecl
@@ -1,0 +1,39 @@
+/*##############################################################################
+
+    HPCC SYSTEMS software Copyright (C) 2019 HPCC Systems®.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and 
+    limitations under the License.
+############################################################################## */
+
+// Analysis should show: DISTRIBUTE output skew is worse than input skew
+//
+// NOTE: For faster nodes, it may be necessary to increase the size of testfile
+
+//nokey
+//noroxie
+//nohthor
+
+IMPORT $.common.Files as Files;
+IMPORT $.common.Helper as Helper;
+
+visits := DATASET(Files.testfile1, Files.layout_visits, THOR);
+
+layout_visitCounts := RECORD
+  visits.url;
+  visits_cnt := COUNT(GROUP);
+END;
+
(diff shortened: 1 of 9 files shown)


AttilaVamos:
@shamser  These test cases  won't be run automatically with Regression Suite in Smoketest and OBT, based on they are not in the default (testing/regress/ecl/) directory.

shamser:
@AttilaVamos I know. This is the regression that I am trying to make work with the regression suite.  There will be a few dozen WU Analysis regressions in the testing/wuanalysis subdirectory.   I can execute with 

./ecl-test --suiteDir ../wuanalysis run 

For each query, there are 2 versions.  The first version of each query is a query that demonstrates an issue with the graph.  After the execution of the first version of the query, the WU Analysis writes the Analysis in the exceptions/messages section of the WU.  The second version reads back the exceptions/messages to check that the issues have been identified.  The output to the 2 versions are different and so require 2 different keys.

Unfortunately, I cannot do this with just one version as the analysis will not be written out until the first query completes.  This is why we need the 2 versions to compare with 2 different keys.

AttilaVamos:
@shamser 

> For each query, there are 2 versions. The first version of each query is a query that demonstrates an issue with the graph. After the execution of the first version of the query, the WU Analysis writes the Analysis in the exceptions/messages section of the WU. The second version reads back the exceptions/messages to check that the issues have been identified. The output to the 2 versions are different and so require 2 different keys.

I think this is not the versioning what we have implemented and expects same output for all versions. In your case I can see two separated query one for doing nasty thing and the second which one is checking the nastiness of previous.

I mean instead of  versioned anahashdistrib1.ecl you have two separated queries:

- testing/wuanalysis/ecl/anahashdistrib1_query.ecl  -  for doing nasty thing
- testing/wuanalysis/ecl/anahashdistrib1_result.ecl - to check anahashdistrib1_query.ecl did its job

With these separated ECL file you have two separated functionality, two separated keyfile and  everything will be fine,
(The Regression Test Engine, without --PQ parameter, always executes the test cases in alphabetical order.)

ghalliday:
Shamser, it would be good to target this to 7.4 - it will not break anything, and would mean we could start running the analysis regression on 7.4 code.

shamser:
@AttilaVamos @ghalliday Please can you review.

shamser:
Ticket to execute WU Analysis regression as part of OBT: https://track.hpccsystems.com/browse/HPCC-22430

AttilaVamos:
@shamser This logic, file placements and hierarchy seems good and will work if we don't use --PQ in OBT to execute these queries.

testing/wuanalysis/ecl/anahashdistrib1.ecl
```diff
@@ -19,9 +19,14 @@
 //
 // NOTE: For faster nodes, it may be necessary to increase the size of testfile
 
-IMPORT * From common;
+//nokey
+//noroxie
+//nohthor
 
-visits := DATASET(testfile1, layout_visits, THOR);
+IMPORT $.common.Files as Files;
```
ghalliday: picky: possibly simpler to say

import Files from $.common;

but fine as it stands

PR #12747, opened 2019-06-28
HPCC-22445 Implement whitelist via interface and plug into MP

jakesmith:
Validate MP connections (in Dali) so that all external clients
are screened on connection.
If they pass MP connection validation, session will still be
validated with their role.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dali/base/dasess.cpp
+++ b/dali/base/dasess.cpp
@@ -18,8 +18,6 @@
 #define da_decl DECL_EXPORT
 
 #include <unordered_map>
-#include <unordered_set>
-#include <string>
 
 #include "platform.h"
 #include "jlib.hpp"
@@ -74,7 +72,17 @@ static std::unordered_map<std::string, DaliClientRole> daliClientRoleMap = {
     { "XRef", DCR_XRef },
     { "EclMinus", DCR_EclMinus },
     { "Monitoring", DCR_Monitoring },
+    { "DaliStop", DCR_DaliStop },
 };
+
+DaliClientRole queryRole(const char *roleStr)
+{
+    const auto &it = daliClientRoleMap.find(roleStr);
+    if (it == daliClientRoleMap.end())
+        return DCR_Unknown;
+    return it->second;
+}
+
 
 const char *queryRoleName(DaliClientRole role)
 {
@@ -106,10 +114,10 @@ const char *queryRoleName(DaliClientRole role)
     case DCR_XRef: return "XRef";
     case DCR_EclMinus: return "EclMinus";
     case DCR_Monitoring: return "Monitoring";
+    case DCR_DaliStop: return "DaliStop";
     }
     return "Unknown";
 }
-
 
 
 interface ISessionManagerServer: implements IConnectionMonitor
@@ -125,7 +133,6 @@ interface ISessionManagerServer: implements IConnectionMonitor
     virtual void stopSession(SessionId sessid,bool failed) = 0;
     virtual void setClientAuth(IDaliClientAuthConnection *authconn) = 0;
     virtual void setLDAPconnection(IDaliLdapConnection *_ldapconn) = 0;
-    virtual bool authorizeConnection(const IpAddress &clientIP, DaliClientRole role) = 0;
     virtual void start() = 0;
     virtual void ready() = 0;
     virtual void stop() = 0;
@@ -557,35 +564,8 @@ public:
                 const SocketEndpoint &peerIP = coven.queryComm().queryChannelPeerEndpoint(mb.getSender());
                 Owned<INode> servernode(deserializeINode(mb));  // hopefully me, but not if forwarded
                 int role=0;
-                if (mb.length()-mb.getPos()>=sizeof(role)) { // a capability block present
+                if (mb.length()-mb.getPos()>=sizeof(role)) // a capability block present
                     mb.read(role);
-                    if (!manager.authorizeConnection(peerIP, (DaliClientRole) role))
-                    {
-                        MilliSleep(2000); // Delay makes rapid probing of all possible roles slightly more painful.
-                        SocketEndpoint sender = mb.getSender();
-                        mb.clear();
-                        mb.append((SessionId) 0);
-                        INode *na = queryNullNode();
-                        Owned<IGroup> dummyCoven = createIGroup(1, &na);
-                        dummyCoven->serialize(mb);
-                        const char *roleName = queryRoleName((DaliClientRole)role);
-                        StringBuffer ipStr;
-                        peerIP.getIpText(ipStr);
-                        Owned<IException> e = makeStringExceptionV(-1, "Access denied! [client ip=%s, role=%s]", ipStr.str(), roleName);
-                        EXCLOG(e, nullptr);
-                        serializeException(e, mb);
-                        coven.reply(mb);
-                        MilliSleep(100+getRandom()%1000); // Causes client to 'work' for a short time.
-                        Owned<INode> node = createINode(sender);
-                        coven.disconnect(node);
-                        break;
-                    }
-#ifdef _DEBUG
-                    StringBuffer eps;
-                    PROGLOG("Connection to %s at %s authorized",queryRoleName((DaliClientRole)role),mb.getSender().getUrlStr(eps).str());
-#endif
-                }
-                
                 IGroup *covengrp;
                 id = manager.registerClientProcess(node.get(),covengrp,(DaliClientRole)role);
                 mb.clear().append(id);
@@ -905,8 +885,6 @@ public:
             servernotifys.item(i).doNotify(aborted);
         }
     }
-
-
 };
 
 
@@ -1160,16 +1138,6 @@ public:
         queryCoven().sendRecv(mb,RANK_RANDOM,MPTAG_DALI_SESSION_REQUEST,SESSIONREPLYTIMEOUT);
     }
 
-    virtual void refreshWhiteList() override
-    {
-        throwUnexpectedX("refreshWhiteList called on client");
-    }
-
-    virtual StringBuffer &getWhiteList(StringBuffer &out) const override
-    {
-        throwUnexpectedX("getWhiteList called on client");
-    }
-
     void onClose(SocketEndpoint &ep)
     {
         CHECKEDCRITICALBLOCK(sessmanagersect,60000);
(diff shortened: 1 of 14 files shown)


jakesmith:
Some notes:
1) Since whitelisting is now done at the MP connect level, and rejected clients are closed at that time, there are changes here so that MP clients provide their 'role' during connect (in a backward/forward compatible way), and there is corresponding client code to handle the exception response.
2) I will add a corresponding 7.0.x point change, so that Dali clients provide a role (at MP connect time), and handle the exception if rejected.
3) Legacy clients (prior to this change and/or prior to (2)), will not provide this role, by default if their IP is whitelisted they will be permitted to connected (behaviour can be changed by WhiteList/allowAnonRoles)
4) Legacy clients (prior to this change and/or prior to (2)), who are rejected will see their MP socket closed, this will result in a MP link closed exception at the client side. 
5) MP clients have always been able to connect to Dali directly (without registration as a session and without providing a role), that includes foreign:: access from other environments. This change means those foreign 'clients' will need whitelisting too.


@richardkchapman @mckellyln @ghalliday - please review.

dali/base/dasess.cpp
```diff
@@ -74,8 +72,19 @@ static std::unordered_map<std::string, DaliClientRole> daliClientRoleMap = {
...
     { "Monitoring", DCR_Monitoring },
+    { "DaliStop", DCR_DaliStop },
 };
 
+DaliClientRole queryRole(const char *roleStr)
+{
+    const auto &it = daliClientRoleMap.find(roleStr);
+    if (it == daliClientRoleMap.end())
+        return DCR_Unknown;
+    StringBuffer ipStr;
```
ghalliday: unused

dali/server/daserver.cpp
```diff
@@ -136,6 +139,199 @@ void usage(void)
...
+    {
+        if (!whiteListTree->getPropBool("@enabled", true)) // on by default
+            return false;
+        // Default for now is to allow clients that send no role (legacy) to connect if their IP is whitelisted.
+        whiteListHandler.setAllowAnonRoles(whiteListTree->getPropBool("@allowAnonRoles", true));
+    }
+
+    std::unordered_map<std::string, std::string> machineMap;
+
+    auto populateMachineMap = [&machineMap](IPropertyTree &environment)
```
ghalliday: I prefer the previous code structure - as members of a WhiteListClass (which could implement IWhiteLIstHandler).  Not sure what I think of use of lambda functions like this.  (Not used to it, may be a good idea, but is it more confusing than helpful?)

dali/server/daserver.cpp
```diff
@@ -136,6 +139,199 @@ void usage(void)
...
+    {
+        const auto &it = machineMap.find(compName);
+        if (it == machineMap.end())
+            return defaultValue;
+        IpAddress ip(it->second.c_str());
+        if (ip.isNull())
+            return defaultValue;
+        return ip.getIpText(result);
+    };
+    auto addRoles = [&whiteListHandler, &resolveComputer](const IPropertyTree &component, const std::vector<unsigned __int64> &roles)
```
ghalliday: My suspicion is that this will be expanded each time it is called, rather than generated once.

dali/server/daserver.cpp
```diff
@@ -136,6 +139,199 @@ void usage(void)
...
+            if (ip)
+            {
+                for (auto &role: roles)
+                    whiteListHandler.add(ip, role);
+            }
+        }
+    };
+
+    populateMachineMap(*conn->queryRoot());
+    enum SoftwareComponentType
```
ghalliday: These seem like they should belong in a public header somewhere.

system/jlib/jsocket.cpp
```diff
@@ -6794,3 +6805,130 @@ int wait_write_multiple(UnsignedArray &socks,       //IN   sockets to be checked
...
+        std::size_t operator () (std::pair<T1, T2> const &pair) const
+        {
+            std::size_t h1 = std::hash<T1>()(pair.first);
+            std::size_t h2 = std::hash<T2>()(pair.second);
+            return h1 ^ h2;
+        }
+    };
+
+    using WhiteListHT = std::unordered_set<std::pair<std::string, unsigned __int64>, PairHasher>;
+    WhiteListPopulateFunction populateFunc;
```
ghalliday: The use of callbacks seems unnecessarily complex

system/jlib/jsocket.hpp
```diff
@@ -629,5 +629,20 @@ extern jlib_decl int wait_write_multiple(UnsignedArray  &socks,     //IN   socke
 extern jlib_decl void throwJSocketException(int jsockErr);
 extern jlib_decl IJSOCK_Exception* createJSocketException(int jsockErr, const char *_msg);
 
+interface IWhiteListHandler : extends IInterface
+{
+    virtual bool isWhiteListed(const char *ip, unsigned __int64 role) const = 0;
+    virtual StringBuffer &getWhiteList(StringBuffer &out) const = 0;
+    virtual void add(const char *ip, unsigned __int64 role) = 0;
```
ghalliday: I wouldn't expect this in the public interface

system/jlib/jsocket.hpp
```diff
@@ -629,5 +629,20 @@ extern jlib_decl int wait_write_multiple(UnsignedArray  &socks,     //IN   socke
...
+    virtual StringBuffer &formatRole(StringBuffer &out, unsigned __int64 role) = 0;
```
ghalliday: can be const I think

dali/base/dasess.cpp
```diff
@@ -557,35 +565,8 @@ class CSessionRequestServer: public Thread
                 const SocketEndpoint &peerIP = coven.queryComm().queryChannelPeerEndpoint(mb.getSender());
                 Owned<INode> servernode(deserializeINode(mb));  // hopefully me, but not if forwarded
                 int role=0;
-                if (mb.length()-mb.getPos()>=sizeof(role)) { // a capability block present
+                if (mb.length()-mb.getPos()>=sizeof(role)) // a capability block present
                     mb.read(role);
-                    if (!manager.authorizeConnection(peerIP, (DaliClientRole) role))
```
ghalliday: What happens if the client does not call initClientProcess?  Will it fail on the first connection?  E.g. some legacy clients.

ghalliday (commented):
@jakesmith a few comments, mainly about the structure rather than the functionality

dali/server/daserver.cpp
jakesmith: The reason I changed it from a class to a single function,
was that I wanted the 'populate' to be single callback function and not part of the interface.

The implementation doesn't know about how to populate, but calls the supplied function to do so on demand.

(the lamda function could just be outside, but seems tidier to encapsulate the function here).

system/jlib/jsocket.cpp
jakesmith: Really all of this should be in the environment.* (as comment says somewhere), but I can't make libenvironment.so (that's true before this change too),
but that causes a circular dependency..

system/jlib/jsocket.hpp
jakesmith: agreed

dali/base/dasess.cpp
jakesmith: the whitelist when it connects (ultimately MP does via verifyConnection, that calls connect()).

The role is supplied via MP start up with this change, legacy clients (i.e. pre-7.0.x) will not supply a role and so the 'role' can't be screen for, with the default settings (allowAnonRoles) it will verify using IP only. NB: This is covered in my notes in 1st message of this PR.

This the same as clients that didn't call initClientProcess before, which happens for example when a foreign components makes direct Dali connections.

dali/base/dasess.cpp
jakesmith: agreed, will remove.

dali/server/daserver.cpp
jakesmith: Really all of this should be in the environment.* (as comment says somewhere), but I can't make libenvironment.so (that's true before this change too),
but that causes a circular dependency..

system/mp/mpcomm.cpp
```diff
@@ -900,15 +932,35 @@ protected: friend class CMPPacketReader;
...
+                     * will see reply as success, so no clean error,
+                     * but will fail shortly afterwards since server connection is closed
+                     */
+                    if (rd > sizeof(rd)) // legacy clients will only ever send a reply of 0 or 4, if greater, then new client is replying with an exception
+                    {
+                        MemoryBuffer mb;
+                        mb.setBuffer(rd, replyMem, false);
+                        size32_t len;
+                        mb.read(len); // exception length
+                        if (len)
```
ghalliday: Doesn't really matter, but I don't think len can ever be null - e.g., a null exception would add a single byte to the output.

ghalliday (commented):
@jakesmith looks good.  A few minor comments

system/jlib/jsocket.cpp
```diff
@@ -6794,3 +6805,140 @@ int wait_write_multiple(UnsignedArray &socks,       //IN   sockets to be checked
...
+    {
+    }
+// IWhiteListHandler impl.
+    virtual bool isWhiteListed(const char *ip, unsigned __int64 role, StringBuffer *responseText) const override
+    {
+        CriticalBlock block(populatedCrit);
+        ensurePopulated();
+        if (0 == role) // unknown, can only check ip
+        {
+            if (allowAnonRoles)
```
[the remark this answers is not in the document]
jakesmith: Yes, as before, but we need this option until we weed out legacy clients that have no role I think.
And anyway, even without anon, a malicious client on a trusted IP can lie about it's role..
Malicious use from a trusted IP can still send whatever role they like..
So the role only protects against misuse from trusted folks on trusted IP's.

dali/server/daserver.cpp
```diff
@@ -136,6 +139,199 @@ void usage(void)
...
+    {
+        Owned<IPropertyTreeIterator> instanceIter = component.getElements("Instance");
+        ForEach(*instanceIter)
+        {
+            const char *compName = instanceIter->query().queryProp("@computer");
+            StringBuffer ipSB;
+            const char *ip = resolveComputer(compName, component.queryProp("@netAddress"), ipSB);
+            if (ip)
+            {
+                for (auto &role: roles)
```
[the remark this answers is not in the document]
jakesmith: agree const, no harm in the reference type though (even if no gain) ?

dali/server/daserver.cpp
```diff
@@ -136,6 +139,199 @@ void usage(void)
...
+    {
+        const auto &it = machineMap.find(compName);
+        if (it == machineMap.end())
+            return defaultValue;
+        IpAddress ip(it->second.c_str());
+        if (ip.isNull())
+            return defaultValue;
+        return ip.getIpText(result);
+    };
+    auto addRoles = [&writer, &resolveComputer](const IPropertyTree &component, const std::vector<unsigned __int64> &roles)
```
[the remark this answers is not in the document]
jakesmith: Haven't used a std::initializer_list yet, but the std::vector is by passed by reference, so.. there should be no temporary copy.. so shouldn't be any more efficient?

system/jlib/jsocket.cpp
```diff
@@ -6794,3 +6805,140 @@ int wait_write_multiple(UnsignedArray &socks,       //IN   sockets to be checked
...
+        CriticalBlock block(populatedCrit);
+        enabled = true;
+        whiteList.clear();
+        IPOnlyWhiteList.clear();
+        populated = false;
+    }
+// IWhiteListWriter impl.
+    virtual void add(const char *ip, unsigned __int64 role) override
+    {
+        CriticalBlock block(populatedCrit);
```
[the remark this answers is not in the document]
jakesmith: you're right, I'll remove.

system/mp/mpcomm.cpp
jakesmith: Yeah I knew it couldn't, but I added check deliberately for possible future backward compatibility, e.g. no exception sent, but reply extended to send something else.

dali/server/daserver.cpp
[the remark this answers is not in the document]
jakesmith: In fact having read about std::initializer_list, I think it will be more expensive (than currently simply passing by reference).
A initializer_list will create a temporary array:
{quote}
The underlying array is a temporary array of type const T[N], in which each element is copy-initialized (except that narrowing conversions are invalid) from the corresponding element of the original initializer list.
{quote}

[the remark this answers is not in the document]
ghalliday: ok.  I suspect the initialiser is better - because it won't allocate anything on the heap.  I suspect it will generate a constant array, so have no creation overhead.

It will be interesting to compare the generated code.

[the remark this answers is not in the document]
ghalliday: BTW The code you have will already be creating an initialiser - because that's passed as the parameter to the vector constructor.

jakesmith:
@ghalliday - I think is ready and was supposed to be back to you.

ghalliday:
@richardkchapman I think this is ready to merge

richardkchapman:
@jakesmith Should target 7.4.x

jakesmith:
@richardkchapman - yes, have retargeted to 7.4.x

PR #12760, opened 2019-07-04
HPCC-18377 Add the audience field to the logging

shamser:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/initfiles/etc/DIR_NAME/environment.conf.in
+++ b/initfiles/etc/DIR_NAME/environment.conf.in
@@ -7,7 +7,7 @@ classpath=${INSTALL_DIR}/classes
 runtime=${RUNTIME_PATH}
 lock=${LOCK_PATH}
 # Supported logging fields: AUD,CLS,DET,MID,TIM,DAT,PID,TID,NOD,JOB,USE,SES,COD,MLT,MCT,NNT,COM,QUO,PFX,ALL,STD
-logfields=TIM+DAT+MLT+MID+PID+TID+COD+QUO+PFX
+logfields=TIM+DAT+MLT+MID+PID+TID+COD+QUO+PFX+AUD
 pid=${PID_PATH}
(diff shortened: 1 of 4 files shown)


shamser:
TODO: Where should this change be documented?

shamser:
@ghalliday Please can you review.

ghalliday:
@richardkchapman should this target 7.4.x or 7.6.0?
@shamser I think jmisc.cpp openLogFile also needs changing.  Should that be using MSGFIELD_STANDARD?

shamser:
@ghalliday @richardkchapman  Shall I re-target to 7.6.0?

shamser:
@ghalliday ECL Watch is not display the log files correctly. There are ECLWatch code that is hard-coded to expect the log files to be in the old format.  Please see WsWuInfo::parseLogLine in ws_workunitHelpers.cpp.  The problem is the old format is assumed for displaying log files in ECL Watch and for search for log for particular workunits.

So, the options I can think of are:
1) Make use of the format specified in environment.conf.   However, if we do that then it won't display/search older older log files correctly.
2) Create a function that "guesses" what fTIM+DAT+MLT+MID+PID+TID+COD+QUO+PFXormat has been written out by examine a few lines of the log file.  This should be easy enough when it's just to determine whether or not the audience has been included.  However, it would be hard to guess every column correctly. (columns such as process id, thread id  and node id are all numbers).
3) Create a header line in every new log file with the format (when missing, assume the format is in the old format).  For example, add something like:
#TIM+DAT+MLT+MID+PID+TID+COD+QUO+PFX

shamser:
A header line is added to record the log format.

PR #12763, opened 2019-07-08
HPCC-22468 allow envgen2 to generate bindings for the new esp cluster 

xwang2713:
This fix "missing esp bindgins" when run "envgen2 -add-node esp#<name>@ip=<ip>"
Also enable DynamicESDL in genenvrules.conf to allow envgen to create DynamicESDL service
Provide default DynamicESDL service name to match default enviornment.xml

Signed-off-by: Xiaoming Wang <xiaoming.wang@lexisnexi.com> 

@kenrowland please review

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

## Testing:

Build: http://10.240.32.243/view/custom7.x/job/CE-JIRA-HPCC-22468/
Tested on Ubuntu 16.04: 
/opt/HPCCSystems/sbin/envgen2 -env-in env.xml -env-out env2.xml -add-node esp#e1@ip=10.176.152.75

### DIFF
--- a/deployment/configenv/xml_jlibpt/SWProcess.cpp
+++ b/deployment/configenv/xml_jlibpt/SWProcess.cpp
@@ -41,22 +41,25 @@ SWProcess::SWProcess(const char* name, EnvHelper * envHelper):SWComponentBase(na
 
 IPropertyTree * SWProcess::addComponent(IPropertyTree *params)
 {
+  IPropertyTree * pCompTree;
   const char* clone = params->queryProp("@clone");
   if (clone)
   {
-     return SWComponentBase::cloneComponent(params);
+     pCompTree = SWComponentBase::cloneComponent(params);
   }
-
-  IPropertyTree * pCompTree = SWComponentBase::addComponent(params);
-  if (pCompTree->hasProp("@daliServers") &&  !strcmp(pCompTree->queryProp("@daliServers"), ""))
+  else
   {
-     IPropertyTree * envTree = m_envHelper->getEnvTree();
-     StringBuffer xpath;
-     xpath.clear().appendf(XML_TAG_SOFTWARE "/DaliServerProcess/@name");
-     const char *daliName = envTree->queryProp(xpath.str());
-     if (daliName)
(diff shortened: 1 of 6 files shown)


deployment/configenv/xml_jlibpt/SWProcess.cpp
```diff
@@ -41,22 +41,25 @@ SWProcess::SWProcess(const char* name, EnvHelper * envHelper):SWComponentBase(na
 
 IPropertyTree * SWProcess::addComponent(IPropertyTree *params)
 {
+  IPropertyTree * pCompTree;
   const char* clone = params->queryProp("@clone");
   if (clone)
   {
-     return SWComponentBase::cloneComponent(params);
+     pCompTree = SWComponentBase::cloneComponent(params);
```
kenrowland: The clone here seems to be redundant with the check for a clone in the new code above that calls cloneComponent (lines 33-37 in SWEspProcess.cpp. In keeping with OO, it would seem that the addComponent and cloneComponent methods should do exactly what their names imply and that a decision to call cloneComponent should be made prior to calling.

kenrowland (commented):
@xwang2713 Please see comment.

deployment/configenv/xml_jlibpt/SWProcess.cpp
xwang2713: Since envgen2 only expose "add" but not "clone" action later is included in the first.
"add" action has three scenarios:
1) the component already exists. In this case "add" will add/update attributes as well as possible add a sub-element based user input in "add" action
2) the component doesn't exist but user want to clone an existing component by keeping the setting in that component
3) the component doesn't exist and user want to create a new component based xsd default values and attributes provided in "add" action.

For most software component "addComponent" function is enough. But for EspProcess extra bindings need applied which is why it is overriding here.

xwang2713:
@richardkchapman this is ready to merge.

PR #12770, opened 2019-07-10
HPCC-22486 Fix wait_multiple return num vs readySocks discrepency

jakesmith:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jsocket.cpp
+++ b/system/jlib/jsocket.cpp
@@ -6742,6 +6742,7 @@ int wait_multiple(bool isRead,               //IN   true if wait read, false it 
 #ifdef _DEBUG
         DBGLOG("%s",dbgSB.str());
 #endif
+        res = readySocks.ordinality();
     }
     else if (res == SOCKET_ERROR)
     {


jakesmith:
@mckellyln - please review

system/jlib/jsocket.cpp
```diff
@@ -6742,6 +6742,7 @@ int wait_multiple(bool isRead,               //IN   true if wait read, false it
 #ifdef _DEBUG
         DBGLOG("%s",dbgSB.str());
 #endif
+        res = readySocks.ordinality();
```
mckellyln: I think line 6731 also needs changing to match this, to remove the check for POLLNVAL.

jakesmith: it was ignoring POLLNVAL before and after this change.

Except before it would have ignore it (not filled in readySocks),
and return >0.
The caller would then hit an assert looking at which of the readySocks had changed.

So same issue as the incident I hit (not with POLLNVAL) but with revents=0.
Namely the returned number of ready socks msmatched the number filled in readySocks.

I think you're right that POLLNVAL should not be ignored,
but if we are going to start returning it (without causing assert to be hit), then it should a separate JIRA/PR so we can track it, in case it has any consequences.

jakesmith: @mckellyln - can you link the 2 JIRA's together to more easily find them later? Thanks.

mckellyln:
I want to also make sure wait_multiple() returns > 0 with readySocks items in the POLLNVAL case, so that any follow-on read/write gets an appropriate error.
We can make that change here or in a separate issue.

jakesmith:
I agree, the change to ignore it a while back seems to be wrong.
But changing it no longer ignore POLLNVAL is a separate issue, please open a new JIRA.

mckellyln:
ok, I will create a new JIRA issue for excluding POLLNVAL check.
ok, I approve this change.

mckellyln (approved):
Since I will create new JIRA and PR for the POLLNVAL change I will approve this.

jakesmith:

PR #12776, opened 2019-07-11
HPCC-10559 Coverity: httpproxy.cpp(679) out of bound access

mayx:
- Use a separate buffer for the read in question to make it simple
- Read multiple bytes at a time instead of 1
- Remove unnecessary socket shutdowns (shutdowns are called automatically
  by jsocket close)

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Premature optimization

## Smoketest:

### DIFF
--- a/esp/tools/soapplus/httpproxy.cpp
+++ b/esp/tools/soapplus/httpproxy.cpp
@@ -297,7 +297,6 @@ public:
             }
             else 
             {
-                m_w->shutdown();
                 m_w->close();
                 break;
             }
@@ -660,12 +659,11 @@ public:
             char peername[256];
             int clientport = m_client->peer_name(peername, 256);
 
-            char inbuf[1024];
-            char outbuf[1024];
-            memset(inbuf, 0, 1024);
-            memset(outbuf, 0, 1024);
+            char inbuf[8];
+            char outbuf[8];
+            memset(inbuf, 0, 8);
+            memset(outbuf, 0, 8);
 
-            unsigned int len = 0;
             unsigned int lenread = 0;
             m_client->read(inbuf, 8, 8, lenread);
             if(lenread != 8)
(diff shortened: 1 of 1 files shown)


mayx:
@mckellyln  please review...

esp/tools/soapplus/httpproxy.cpp
```diff
@@ -674,25 +672,28 @@ class CSocksProxyThread : public Thread
                 return -1;
             }
```
mckellyln: What are the first 8 bytes used for ?

mckellyln: ok, I see below it is the port, correct ?

esp/tools/soapplus/httpproxy.cpp
```diff
@@ -674,25 +672,28 @@ class CSocksProxyThread : public Thread
...
                     break;
-                }
-                char c = inbuf[len - 1];
-                username.append(c);
-                m_client->read(inbuf + len, 0, 1, lenread);
+                username.append(lenread, inbuf2);
+                m_client->read(inbuf2, 0, 16, lenread);
             }
-            
+
```
mckellyln: This loop is to read until you get a null byte and then you know you are finished ?

esp/tools/soapplus/httpproxy.cpp
mckellyln: Just a comment, perhaps it would be more clear if the port code was right after this read.

esp/tools/soapplus/httpproxy.cpp
```diff
@@ -674,25 +672,28 @@ class CSocksProxyThread : public Thread
                 return -1;
             }
 
-            len += lenread;
-            m_client->read(inbuf + len, 0, 1, lenread);
+            char inbuf2[16];
+            m_client->read(inbuf2, 0, 16, lenread);
             StringBuffer username;
             while(lenread > 0)
```
mckellyln: Is min size set to 0 to prevent throwing an exception on close ?

esp/tools/soapplus/httpproxy.cpp
```diff
@@ -674,25 +672,28 @@ class CSocksProxyThread : public Thread
                 return -1;
             }
 
-            len += lenread;
-            m_client->read(inbuf + len, 0, 1, lenread);
+            char inbuf2[16];
+            m_client->read(inbuf2, 0, 16, lenread);
```
mckellyln: Is min size set to 0 to prevent throwing an exception on close ?

esp/tools/soapplus/httpproxy.cpp
```diff
@@ -674,25 +672,28 @@ class CSocksProxyThread : public Thread
...
+                    {
+                        if (i > 0)
+                            username.append(i, inbuf2);
+                        done = true;
+                        break;
+                    }
                 }
-                if(inbuf[len - 1] == '\0')
-                {
+                if (done)
```
mckellyln: why this ?

esp/tools/soapplus/httpproxy.cpp
```diff
@@ -674,25 +672,28 @@ class CSocksProxyThread : public Thread
...
                     break;
```
mckellyln: Why check for done and break here ?

esp/tools/soapplus/httpproxy.cpp
mayx: To store the first 8 bytes, which are the header, includes version, command, destination ip and destination port.

esp/tools/soapplus/httpproxy.cpp
mayx: Yes.

esp/tools/soapplus/httpproxy.cpp
mayx: Seeing 0 means got all data expected.

AttilaVamos:
@mayx The previous Smoketest error was my fault.

esp/tools/soapplus/httpproxy.cpp
mayx: Yes, 0 would avoid the exception, and since read will block for data, so 0 or 1 won't make much difference here.

mayx:
@mckellyln made a change per your suggestion, please review again...

esp/tools/soapplus/httpproxy.cpp
mckellyln: But you checked for that already in the if block above

mayx: That's for the for loop, this check is to exit the outer while loop...

mckellyln: ok, yes now I understand - sorry I missed that its needed to exit the while

esp/tools/soapplus/httpproxy.cpp
```diff
@@ -674,50 +672,50 @@ class CSocksProxyThread : public Thread
...
-                len += lenread;
-                if(len >= 1023)
+                bool done = false;
+                for (int i = lenread - 1; i >= 0; i--)
                 {
-                    len = 0;
+                    if (inbuf2[i] == '\0')
+                    {
+                        if (i > 0)
+                            username.append(i, inbuf2);
```
mckellyln: should this be append(i+1, inbuf2) ?

mckellyln: ok, no, it is correct as is.

esp/tools/soapplus/httpproxy.cpp
mckellyln: should there be a check that the username is not empty ?

mckellyln: ok, a 1 byte null char empty username is allowed

mckellyln (approved):

mckellyln:
@mayx looks good to merge

mckellyln:
@richardkchapman also - good to merge

mckellyln:
@mayx was there a change pushed after the review ?

mayx:
@mckellyln no there wasn't. I accidentally made a change meant for another issue, but then reversed it back.

mckellyln:
@mayx ok thanks.
Good to merge

mckellyln:
@mayx, since its a coverity fix, can we re-target this to master ?

mayx:
@mckellyln will do.

mayx:
Opened another one based on master.

PR #12779, opened 2019-07-12
HPCC-22490 CStreamLineReader returns eof even if data returned

shamser:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jio.cpp
+++ b/system/jlib/jio.cpp
@@ -1387,42 +1387,23 @@ public:
                 {
                     ++currentPtr;
                     out.append(currentPtr-startPtr-(preserveEols?0:1), startPtr);
-                    if (currentPtr == endPtr)
-                        return !refill();
                     return false;
                 }
                 case '\r':
                 {
                     ++currentPtr;
-                    // check for \n
-                    if (currentPtr < endPtr)
-                    {
-                        if ('\n' == *currentPtr) // i.e. \r\n
-                        {
-                            ++currentPtr;
-                            out.append(currentPtr-startPtr-(preserveEols?0:2), startPtr);
-                        }
-                        else // i.e. \r only
-                            out.append(currentPtr-startPtr-(preserveEols?0:1), startPtr);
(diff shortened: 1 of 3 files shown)


shamser:
@ghalliday Please can  you review

system/jlib/jio.cpp
```diff
@@ -1387,42 +1387,23 @@ class CStreamLineReader : public CSimpleInterfaceOf<IStreamLineReader>
...
-                    }
-                    else // must output what we have and read 1 more byte to check for \n
-                    {
-                        out.append(currentPtr-startPtr-(preserveEols?0:1), startPtr);
-                        if (!refill())
-                            return true;
-                        else if ('\n' != *currentPtr)
-                            return false;
-                        ++currentPtr;
+                    out.append(currentPtr-startPtr-(preserveEols?0:1), startPtr);
```
ghalliday: this change introduces a subtle change - lines that terminate in \r\n will write all but 1 byte, and then append the \n.  That *might* cause the target buffer to reallocate.  I doubt it is significant though.

ghalliday (commented):
@shamser the change looks ok - except I think the unit test in jlibtests also needs updating.  Smoketest should pick it up.

system/jlib/jio.cpp
shamser: There doesn't seem to be any need to preserve line terminators in any of the current code that uses CStreamLineReader.  We could set all of them to preserveEols==false? (I can also add a comment to state the point about the in the code about this so someone can modify if the need arises)

shamser:
@ghalliday I've updated the smoketests.   Shall I squash?

ghalliday:
@shamser yes, please squash - I'll merge

shamser:
@ghalliday I've squashed.

PR #12782, opened 2019-07-12
HPCC-22296 SoapCall has extra / in first line of request

mayx:
- Clean the path part of the url by removing extra slashes when parsing
  the url.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Premature optimization

## Smoketest:

### DIFF
--- a/common/thorhelper/thorsoapcall.cpp
+++ b/common/thorhelper/thorsoapcall.cpp
@@ -170,6 +170,24 @@ private:
         return;
     }
 
+    void cleanPath(const char *src, StringBuffer &output)
+    {
+        const char *srcPtr = src;
+        char *dst = output.reserveTruncate(strlen(src));
+        char *dstPtr = dst;
+        while (*srcPtr)
+        {
+            *dstPtr++ = *srcPtr;
+            if ('/' == *srcPtr++)
+            {
+                while ('/' == *srcPtr)
+                    ++srcPtr;
+            }
+        }
+        *dstPtr = '\0';
+        output.setLength(dstPtr-dst);
+    }
+
 public:
     Url(char *urltext)
     {
@@ -206,7 +224,7 @@ public:
             host.set(urltext);
 
             if ((p = strchr(p, '/')) != NULL)
-                path.append(p);
+                cleanPath(p, path);
             else
                 path.append("/");
         }
@@ -223,10 +241,9 @@ public:
             if ((p = strchr(urltext, '/')) != NULL)
             {
                 *p = 0;
-                p++;
-
                 host.set(urltext);
-                path.append("/").append(p);
+                *p = '/';
+                cleanPath(p, path);
             }
             else
             {


mayx:
@jakesmith please review...

common/thorhelper/thorsoapcall.cpp
```diff
@@ -170,6 +170,23 @@ class Url : public CInterface, implements IInterface
         return;
     }
 
+    void cleanPath()
+    {
+        int len = path.length();
```
jakesmith: picky: should be an unsigned (size32_t)

common/thorhelper/thorsoapcall.cpp
```diff
@@ -170,6 +170,23 @@ class Url : public CInterface, implements IInterface
...
+        if (len <= 1)
+            return;
+        int gap = 0;
+        for (int i = 1; i < len; i++)
```
jakesmith: picky: int's should be unsigned

common/thorhelper/thorsoapcall.cpp
```diff
@@ -170,6 +170,23 @@ class Url : public CInterface, implements IInterface
         return;
     }
 
+    void cleanPath()
```
jakesmith: It looks like cleanPath() is trying to walk the whole string and remove any duplicated '/' 's?
e.g. convert : "POST //WsPackageProcess/myxpath//sub HTTP/1.1"
to "POST /WsPackageProcess/myxpath/sub HTTP/1.1

Is that ok/right?

Could do with brief comment here.

common/thorhelper/thorsoapcall.cpp
```diff
@@ -170,6 +170,23 @@ class Url : public CInterface, implements IInterface
...
+        int len = path.length();
+        if (len <= 1)
+            return;
+        int gap = 0;
+        for (int i = 1; i < len; i++)
+        {
+            if (path.charAt(i) == '/' && path.charAt(i-1-gap) == '/')
+                gap++;
+            else if (gap > 0)
+                path.setCharAt(i-gap, path.charAt(i));
```
jakesmith: may not be critical, but calling setCharAt per character is not particularly efficient.
If expected to be hit a lot, would be better to deal with underlying string and reset 'path' to new string when done.
e.g. could detach() and work on const char *writeptr

jakesmith:
@mayx - a couple of comments/questions.

common/thorhelper/thorsoapcall.cpp
mayx: Yes all duplicate slashes should be removed, at the beginning, in the middle or at the end of the path...

mayx:
@jakesmith changes made per suggestion, please review again...

common/thorhelper/thorsoapcall.cpp
```diff
@@ -170,6 +170,24 @@ class Url : public CInterface, implements IInterface
...
+    {
+        size32_t len = path.length();
+        if (len <= 1)
+            return;
+        size32_t gap = 0;
+        char* buf = path.detach();
+        //Remove all duplicate slashes
+        for (size32_t i = 1; i < len; i++)
+        {
+            if (buf[i] == '/' && buf[i-1-gap] == '/')
```
jakesmith: minor: [i-1-gap] is always the 1st '/' in a sequence, so you could code this more efficiently as a while loop.

Personally I'd code this routine something like this:
```
void cleanPath(const char *src, StringBuffer &output)
{
    const char *srcPtr = src;
    char *dst = output.reserveTruncate(strlen(src));
    char *dstPtr = dst;
    while (*srcPtr)
    {
        *dstPtr++ = *srcPtr;
        if ('/' == *srcPtr++)
        {
            while ('/' == *srcPtr)
                ++srcPtr;
        }
    }
    *dstPtr = '\0';
    output.setLength(dstPtr-dst);
}

```

You could pass a single StringBuffer as both the input and output, and detach() instead, but given the input in the Uri ctor isn't originally a StringBuffer (i.e. 'p' was just appended before calling cleanPath)...

jakesmith:
@mayx - 1 comment/suggestion, but okay to be merged as is.

mayx:
@jakesmith changes made per suggestion, please review again...

jakesmith:
@mayx - looks fine.

PR #12785, opened 2019-07-15
HPCC-22520 Fix problems passing string containing chars>127 to macros

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/rtl/eclrtl/eclrtl.cpp
+++ b/rtl/eclrtl/eclrtl.cpp
@@ -3652,6 +3652,22 @@ int rtlSingleUtf8ToCodepage(char * out, unsigned inlen, char const * in, char co
     return static_cast<int>(trailbytes); //cast okay as is certainly 0--3
 }
 
+bool rtlCodepageToCodepage(StringBuffer & out, unsigned maxoutlen, unsigned inlen, char const * in, char const * outcodepage, char const * incodepage)
+{
+    UConverter * inconv = queryRTLUnicodeConverter(incodepage)->query();
+    UConverter * outconv = queryRTLUnicodeConverter(outcodepage)->query();
+    UErrorCode err = U_ZERO_ERROR;
+
+    //GH->PG is there a better way of coding this with out temporary buffer?
+    size32_t oldLength = out.length();
+    char * tempBuffer = out.reserve(maxoutlen);
+    char * target = tempBuffer;
+    ucnv_convertEx(outconv, inconv, &target, tempBuffer+maxoutlen, &in, in+inlen, NULL, NULL, NULL, NULL, TRUE, TRUE, &err);
+    unsigned len = target - tempBuffer;
+    out.setLength(oldLength + len);
+    return U_SUCCESS(err) != FALSE;
+}
+
 #else
 void rtlUnicodeToEscapedStrX(unsigned & outlen, char * & out, unsigned inlen, UChar const * in)
 {
@@ -3681,6 +3697,13 @@ int rtlSingleUtf8ToCodepage(char * out, unsigned inlen, char const * in, char co
 {
     rtlThrowNoUnicode();
 }
+
+bool rtlCodepageToCodepage(StringBuffer & out, unsigned maxoutlen, unsigned inlen, char const * in, char const * outcodepage, char const * incodepage)
+{
+    out.append(inlen, in);
+    return true;
+}
+
 
 #endif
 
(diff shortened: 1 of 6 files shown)


ghalliday:
@shamser please review.
The actual fix is very small.  The other functions are not called, but were written for a first fix of the issue - they are generally useful (I have needed at least one of them writing recent code), so it would be useful to include them as well.

rtl/eclrtl/eclrtl.cpp
```diff
@@ -3682,6 +3698,13 @@ int rtlSingleUtf8ToCodepage(char * out, unsigned inlen, char const * in, char co
     rtlThrowNoUnicode();
 }
 
+bool rtlCodepageToCodepage(StringBuffer & out, unsigned maxoutlen, unsigned inlen, char const * in, char const * outcodepage, char const * incodepage)
```
shamser: Shouldn't it throw an exception here?

system/jlib/junicode.cpp
```diff
@@ -713,3 +713,26 @@ extern jlib_decl bool appendUtfXmlName(MemoryBuffer & target, UtfReader::UtfForm
 {
     return replaceUtf(utfXmlNameReplacementFunc, target, type, sourceLength, source);
 }
+
+extern jlib_decl bool containsOnlyAscii(const char * source)
```
shamser: Where is containsOnlyAscii used?

rtl/eclrtl/eclrtl.cpp
```diff
@@ -3652,6 +3652,22 @@ int rtlSingleUtf8ToCodepage(char * out, unsigned inlen, char const * in, char co
     return static_cast<int>(trailbytes); //cast okay as is certainly 0--3
 }
 
+bool rtlCodepageToCodepage(StringBuffer & out, unsigned maxoutlen, unsigned inlen, char const * in, char const * outcodepage, char const * incodepage)
+{
+    UConverter * inconv = queryRTLUnicodeConverter(incodepage)->query();
+    UConverter * outconv = queryRTLUnicodeConverter(outcodepage)->query();
+    UErrorCode err = U_ZERO_ERROR;
```
shamser: Is it worth copying skipping the conversion when incodepage and out codepage is the same? I know you've mentioned that it's not currently used by the generated code but it could be included for some purposed in the future.

shamser (commented):
@ghalliday

rtl/eclrtl/eclrtl.cpp
ghalliday: It probably should - I'll check and change

system/jlib/junicode.cpp
ghalliday: Not currently used

rtl/eclrtl/eclrtl.cpp
ghalliday: I can't imagine this ever being called with the same source and target codepage, so not worth skipping the conversion.

rtl/eclrtl/eclrtl.cpp
ghalliday: On reflection it is probably best to leave as it is - since it will work for strings that only contain ascii characters.
Really it would be better to have some lightweight code that converted from latin-1 to utf8 (it would only require a character mapping table and using the functions in junicode), but that is outside the scope of this fix.

ghalliday:
@shamser

shamser (approved):
@richardkchapman Code reviewed

system/jlib/junicode.hpp
```diff
@@ -81,6 +81,10 @@ typedef MemoryBuffer & (*utfReplacementFunc)(MemoryBuffer & target, UTF32 match,
 extern jlib_decl bool replaceUtf(utfReplacementFunc func, MemoryBuffer & target, UtfReader::UtfFormat type, unsigned sourceLength, const void * source);
 extern jlib_decl bool appendUtfXmlName(MemoryBuffer & target, UtfReader::UtfFormat type, unsigned sourceLength, const void * source);
 
+//Does the string only contain characters < 128 (and non need any work when converteding to utf8)
```
richardkchapman: trivial: typo in comment

PR #12797, opened 2019-07-17
HPCC-22540 Handle secure and rowservice socket accept exceptions

jakesmith:
NB: this is similar to HPCC-22525

Previously if accept threw an exception (e.g. due to running out
of handles), it would cause dafilesrv to exit, which caused
clients with established connection to break too.

Catch trace and continue

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/fs/dafsserver/dafsserver.cpp
+++ b/fs/dafsserver/dafsserver.cpp
@@ -4299,6 +4299,7 @@ public:
 
         selecthandler->start();
 
+        Owned<IException> exception;
         for (;;)
         {
             Owned<ISocket> sock;
@@ -4362,9 +4363,13 @@ public:
                     }
                     catch (IException *e)
                     {
-                        EXCLOG(e,"CRemoteFileServer");
-                        e->Release();
-                        continue;
+                        exception.setown(e);
+                    }
+                    if (exception)
+                    {
+                        EXCLOG(exception, "CRemoteFileServer");
+                        exception.clear();
+                        sockavail = false;
                     }
                 }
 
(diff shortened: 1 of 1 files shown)


jakesmith:
@ghalliday - please review (NB: this is similar to HPCC-22525)

fs/dafsserver/dafsserver.cpp
```diff
@@ -4396,18 +4401,20 @@ class CRemoteFileServer : implements IRemoteFileServer, public CInterface
                     catch (IJSOCK_Exception *e)
```
ghalliday: minor: This specialised catch could now be deleted.

fs/dafsserver/dafsserver.cpp
```diff
@@ -4362,9 +4363,13 @@ class CRemoteFileServer : implements IRemoteFileServer, public CInterface
                     }
                     catch (IException *e)
                     {
-                        EXCLOG(e,"CRemoteFileServer");
-                        e->Release();
-                        continue;
+                        exception.setown(e);
+                    }
+                    if (exception)
```
ghalliday: You could include this code within the try and then avoid assigning to exception.  I don't mind either way.

fs/dafsserver/dafsserver.cpp
jakesmith: agree, will remove.

fs/dafsserver/dafsserver.cpp
jakesmith: I'll leave this one as is, partly because in another PR, I want to call something here, and don't want to call in catch

PR #12802, opened 2019-07-18
HPCC-22538 Add a helper function that dumps file and socket info

jakesmith:
Based on lsof output, output which files a process has open,
and the sockets (and their endpoints) the process is using.

Call in dafilesrv when it runs out of handles for diagnostic
purposes.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/fs/dafsserver/dafsserver.cpp
+++ b/fs/dafsserver/dafsserver.cpp
@@ -3209,6 +3209,62 @@ class CRemoteFileServer : implements IRemoteFileServer, public CInterface
     unsigned targetActiveThreads;
     Linked<IPropertyTree> keyPairInfo;
 
+    class CHandleTracer
+    {
+        CTimeMon timer;
+        CriticalSection crit;
+        Owned<IFile> stdIOIFile;
+        std::vector<Owned<IFileIO>> reservedHandles;
+        unsigned handlesToReserve = 3; // need a few for pipe process to succeed
+
+        void reserveHandles()
+        {
+            if (stdIOIFile)
+            {
+                for (unsigned r=0; r<handlesToReserve; r++)
+                {
+                    IFileIO *iFileIO = stdIOIFile->open(IFOread);
+                    if (iFileIO)
+                        reservedHandles.push_back(iFileIO);
+                }
+            }
+        }
+        void releaseHandles()
+        {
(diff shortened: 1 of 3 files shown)


jakesmith:
@mckellyln - please review the last commit.

The 1st commit is merged already (but not upmerged yet).
The 2nd commit is in a separate PR.

mckellyln:
@jakesmith I never like when code calls system("command") (ie pipe...) but in this case lsof command is very powerful and difficult to code natively.

fs/dafsserver/dafsserver.cpp
```diff
@@ -3209,6 +3209,62 @@ class CRemoteFileServer : implements IRemoteFileServer, public CInterface
...
+        {
+            reservedHandles.clear();
+        }
+    public:
+        CHandleTracer()
+        {
+            /* Reserve handles, so that when we run out, we hope to release them
+             * and thereby have enough to use when reading current state.
+             */
+            stdIOIFile.setown(createIFile("stdin:"));
```
mckellyln: Perhaps stdout or some other expected file descriptor is a better choice than stdin ?
I have seen in other applications stdin is closed when running as a daemon.

jakesmith: yes, as per offline discussion, I agree, think stdout would be safer, in case stdin has deliberately been closed.
Will change.

mckellyln:
@jakesmith looks good but for one question.

jakesmith:
>@jakesmith I never like when code calls system("command") (ie pipe...) but in this case lsof command is very powerful and difficult to code natively.

Generally agree, but no point re-inventing the wheel here, lsof decodes the sockets, builds a map of UDP and TCP sockets (that are coded in other parts of /proc/ on Linux) and then maps the local and peer endpoints together, amongst other things.
So could be done, but lsof provides the functionally wanted (and we use pipe for similar things in jdebug for similar reasons).

mckellyln:
@jakesmith looks good.

mckellyln (approved):

PR #12803, opened 2019-07-18
HPCC-11962 Monitor Dali and kill (after timeout) process

jakesmith:
Forcibly kill eclagent process if Dali has closed and process
doesn't exit gracefully.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

## Testing:

Tested by:
1) running system
2) stopping thor,
3) running Thor job (so blocked on Thor).
4) stopping system.

That left the eclagent process orphaned.

With this PR, it spotted Dali had gone down, and after the pre-requisite 60 second timeout,
it killed itself.

### DIFF
--- a/ecl/eclagent/eclagent.cpp
+++ b/ecl/eclagent/eclagent.cpp
@@ -3460,12 +3460,15 @@ extern int HTHOR_API eclagent_main(int argc, const char *argv[], StringBuffer * 
             wuXML->kill();  // free up text as soon as possible.
         }
 
+        Owned<IConnectionMonitor> daliDownMonitor;
+
         Owned<IUserDescriptor> standAloneUDesc;
         if (daliServers.length())
         {
+            SocketEndpoint daliEp(daliServers, DALI_SERVER_PORT);
             {
                 MTIME_SECTION(queryActiveTimer(), "SDS_Initialize");
-                Owned<IGroup> serverGroup = createIGroup(daliServers.str(), DALI_SERVER_PORT);
+                Owned<IGroup> serverGroup = createIGroup(1, &daliEp);
                 initClientProcess(serverGroup, DCR_EclAgent, 0, NULL, NULL, MP_WAIT_FOREVER);
             }
 #ifdef MONITOR_ECLAGENT_STATUS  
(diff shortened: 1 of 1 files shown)


jakesmith:
@mckellyln - please review

ecl/eclagent/eclagent.cpp
```diff
@@ -3460,19 +3460,54 @@ extern int HTHOR_API eclagent_main(int argc, const char *argv[], StringBuffer *
...
+                    {
+                        WARNLOG("Dali has closed. Waiting 1 minute for process to exit gracefully before killing..");
+
+                        if (sem.wait(60000)) // see if process shutsdown gracefully before killing
+                            return;
+                        WARNLOG("Killing process");
+#ifdef _WIN32
+                        TerminateProcess(GetCurrentProcess(), 1);
+#else
+                        kill(getpid(), SIGKILL);
```
mckellyln: just being picky, I prefer raise() if sending to oneself.

jakesmith: I don't think we use raise() anywhere at the moment.
Above is copy and paste of code that we use in quite a few place.

I thought about adding a wrapper to remove the platform #ifdef's here, but didn't.
We can do that another time and change to raise()  (so everywhere uses).

mckellyln:
@jakesmith looks good.  Have you tested it ?

jakesmith:
>Have you tested it ?

Yes, have updated the Testing section above.

mckellyln (approved):

mckellyln:
@jakesmith looks good to merge.

jakesmith:

ecl/eclagent/eclagent.cpp
richardkchapman: Seems like this is likely to block the very thread that is trying to close cleanly?

Also, isn't there a guillotine thread running that (a) should have killed eventually anyway and (b) could check periodically that dali connection is still valid.

richardkchapman:
@jakesmith Some questions

jakesmith:
>Also, isn't there a guillotine thread running that (a) should have killed eventually anyway and (b) could check periodically that dali connection is still valid.

There is a guillotine mechanism, set via workunit option "maxRunTime", but it's not enabled by default, probably because eclagent processes can be blocked by various things for undefined periods.

jakesmith:
>Seems like this is likely to block the very thread that is trying to close cleanly?

This onClose is called by a connection monitor thread inside MP, it will be only be invoked when Dali has closed. When it does, it will do a sem.wait for up to 1 minute, if the eclagent process is already existing then the dtor will signal and this thread will do nothing, if 1 minute expires (after being told Dali went down), then the process will be forcibly killed.

jakesmith:
@richardkchapman - added some responses to questions.

richardkchapman:
@ghalliday Please sanity-check

ghalliday:
@jakesmith is there a danger that this will cause deadlock/race e.g.:

- a connection monitor which is registered after this one will never get called.  (e.g. querySessionManager() called late for some reason)
- An (unlikely) race condition if the semaphore is signalled, but onClose() times out - WARNLOG will crash if the process has exited.
-

jakesmith:
>a connection monitor which is registered after this one will never get called. (e.g. querySessionManager() called late for some reason)

I don't think that's a problem, it's likely querySessionManager() will have been called earlier, but even it wasn't, if it was use late (after signal that Dali had gone had been received), then it couldn't be used, or rather those uses would stall or error. 

>An (unlikely) race condition if the semaphore is signalled, but onClose() times out - WARNLOG will crash if the process has exited.

Very unlikely I think, it has a [relatively] long way to go before logging is defunct., it should have got past that line by that time.
But yes theoretically there's a window, I'll change so it avoids it.

jakesmith:
@ghalliday  - Actually, I take it back, I don't think there is a window.
If the it's signaled and process is winding up (and in or about to do WARNLOG), it must be before it has stopped MP, which will have stopped (joined) the thread that calls the IConnectionMonitor::onClose methods.
So afaics, it cannot be in onClose dealing with tracing after it is invalid

ghalliday:
@richardkchapman I'll let you take a last look and merge

jakesmith:
@richardkchapman - agree, have switched it to master.

PR #12807, opened 2019-07-19
Start next stage of disk read refactoring

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/common/thorhelper/csvsplitter.cpp
+++ b/common/thorhelper/csvsplitter.cpp
@@ -71,9 +71,21 @@ void CSVSplitter::addTerminator(const char * text)
     matcher.addEntry(text, TERMINATOR);
 }
 
+void CSVSplitter::addItem(MatchItem item, const char * text)
+{
+    if (text)
+        matcher.addEntry(text, item);
+}
+
 void CSVSplitter::addEscape(const char * text)
 {
     matcher.queryAddEntry((size32_t)strlen(text), text, ESCAPE);
+}
+
+void CSVSplitter::addWhitespace()
+{
+    matcher.queryAddEntry(1, " ", WHITESPACE);
+    matcher.queryAddEntry(1, "\t", WHITESPACE);
 }
 
 void CSVSplitter::reset()
@@ -156,11 +168,9 @@ void CSVSplitter::init(unsigned _maxColumns, ICsvParameters * csvInfo, const cha
 
     //MORE Should this be configurable??
     if (!(flags & ICsvParameters::preserveWhitespace))
-    {
-        matcher.queryAddEntry(1, " ", WHITESPACE);
-        matcher.queryAddEntry(1, "\t", WHITESPACE);
-    }
+        addWhitespace();
 }
+
 
 void CSVSplitter::init(unsigned _maxColumns, size32_t _maxCsvSize, const char *quotes, const char *separators, const char *terminators, const char *escapes, bool preserveWhitespace)
 {
@@ -181,11 +191,8 @@ void CSVSplitter::init(unsigned _maxColumns, size32_t _maxCsvSize, const char *q
     if (escapes)
         addActionList(matcher, escapes, ESCAPE);
 
-    if (preserveWhitespace)
-    {
-        matcher.queryAddEntry(1, " ", WHITESPACE);
-        matcher.queryAddEntry(1, "\t", WHITESPACE);
-    }
+    if (!preserveWhitespace)
+        addWhitespace();
 }
 
 void CSVSplitter::setFieldRange(const byte * start, const byte * end, unsigned curColumn, unsigned quoteToStrip, bool unescape)
--- a/common/thorhelper/thorfile.cpp
+++ b/common/thorhelper/thorfile.cpp
@@ -86,6 +86,16 @@ public:
         memcpy(rowBuilder.ensureCapacity(size, NULL), src, size);
         return size;
     }
+    virtual const char * queryFormat() override
+    {
+        return "thor";
+    }
+    virtual void getFormatOptions(IXmlWriter & options) override
+    {
+    }
+    virtual void getFormatDynOptions(IXmlWriter & options) override
+    {
+    }
 
 protected:
     StringAttr filename;
--- a/common/thorhelper/thorfile.cpp
+++ b/common/thorhelper/thorfile.cpp
@@ -88,7 +88,7 @@ public:
     }
     virtual const char * queryFormat() override
     {
-        return "thor";
+        return "flat";
     }
     virtual void getFormatOptions(IXmlWriter & options) override
     {
--- a/common/thorhelper/thorread.cpp
+++ b/common/thorhelper/thorread.cpp
@@ -25,6 +25,7 @@
 #include "rtlfield.hpp"
 #include "rtlds_imp.hpp"
 #include "rtldynfield.hpp"
+#include "roxiemem.hpp"
 
 #include "rmtclient.hpp"
 #include "rmtfile.hpp"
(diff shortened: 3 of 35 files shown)


ghalliday:
@jakesmith @richardkchapman for review and discussion

There are a couple of failures that will be trivial to tidy up.

ghalliday:
The code needs a little bit of cleaning up, but I am particularly interested in whether the structure is correct, and whether it will be easy to extend to other file formats.
The failures are because i) remote file reading doesn't support csv with the normal disk read activity and ii) a bug in roxie localfilepositions.

@jpmcmu @jakesmith @richardkchapman please review.  Hopefully we can discuss on Friday.

common/thorhelper/csvsplitter.cpp
```diff
@@ -181,11 +191,8 @@ void CSVSplitter::init(unsigned _maxColumns, size32_t _maxCsvSize, const char *q
     if (escapes)
         addActionList(matcher, escapes, ESCAPE);
 
-    if (preserveWhitespace)
```
richardkchapman: You have inverted the test compared to the original code, I think - was the original wrong?

common/thorhelper/thorfile.cpp
```diff
@@ -86,6 +86,16 @@ class DiskWorkUnitReadArg : public CThorDiskReadArg
         memcpy(rowBuilder.ensureCapacity(size, NULL), src, size);
         return size;
     }
+    virtual const char * queryFormat() override
+    {
+        return "thor";
```
richardkchapman: Is this preferred to "flat" ?

common/thorhelper/thorread.cpp
```diff
@@ -25,13 +25,152 @@
...
+    unsigned projectedCrc;
+    Linked<IOutputMetaData> actualMeta;
+    Linked<IOutputMetaData> expectedMeta;
+    Linked<IOutputMetaData> projectedMeta;
+    Linked<const IPropertyTree> options;
+    mutable Owned<const IDynamicTransform> translator;
+    mutable Owned<const IKeyTranslator> keyedTranslator;
+};
+
+static CriticalSection translatorCS;
```
richardkchapman: Bit worried about a single shared critsec like this - especially when every call to queryTranslator will be blocked if someone is creating a translator (even if checkedTranslators is true). Is there some way that could be improved?

common/thorhelper/thorread.cpp
```diff
@@ -25,13 +25,152 @@
...
+    {
+        translator.setown(createRecordTranslatorViaCallback(projectedRecord, sourceRecord, type_utf8));
+    }
+    else
+    {
+        if ((projectedMeta != sourceMeta) && (projectedCrc != sourceCrc))
+            translator.setown(createRecordTranslator(projectedRecord, sourceRecord));
+    }
+
+    if (translator)
```
richardkchapman: Should put some (optional?) tracing of the translator in here, to match recent changes in master branch.

common/thorhelper/thorread.cpp
```diff
@@ -25,13 +25,152 @@
...
+    {
+        if (!translator->canTranslate())
+            throw MakeStringException(0, "Untranslatable record layout mismatch detected for file %s", filename);
+
+        if (translator->needsTranslate())
+        {
+            if (sourceMeta != expectedMeta)
+            {
+                Owned<const IKeyTranslator> _keyedTranslator = createKeyTranslator(sourceMeta->queryRecordAccessor(true), expectedMeta->queryRecordAccessor(true));
+                //MORE: What happens if the key filters cannot be translated?
```
richardkchapman: I don't know, but probably should work out before finally enabling this code... Does it need a JIRA to revisit?

common/thorhelper/thorread.cpp
```diff
@@ -556,11 +703,458 @@ size32_t BinaryDiskRowReader::getFixedDiskRecordSize()
 
 //---------------------------------------------------------------------------------------------------------------------
 
+class ExternalFormatDiskRowReader : public LocalDiskRowReader
```
richardkchapman: What does "External format" mean? Is it worth commenting some of these classes?

common/thorhelper/thorread.cpp
```diff
@@ -556,11 +703,458 @@ size32_t BinaryDiskRowReader::getFixedDiskRecordSize()
...
+{
+public:
+    virtual bool first() override { return false; }
+    virtual bool next() override { return false; }
+    virtual bool isValid() override { return false; }
+    virtual IDynamicFieldValueFetcher &query() override
+    {
+        throwUnexpected();
+    }
+} nullNestedRowIterator;
```
richardkchapman: Should this variable be static?

common/thorhelper/thorread.cpp
```diff
@@ -556,11 +703,458 @@ size32_t BinaryDiskRowReader::getFixedDiskRecordSize()
...
+            return projectedFilter.matches(row);
+        }
+        else
+            return true;
+    }
+
+    size32_t getFixedDiskRecordSize();
+
+protected:
+    constexpr static unsigned defaultMaxCsvRowSize = 10; // MB
```
richardkchapman: Might be inclined to call the variable defaultMaxCsvRowSizeMB

common/thorhelper/thorread.cpp
```diff
@@ -556,11 +703,458 @@ size32_t BinaryDiskRowReader::getFixedDiskRecordSize()
...
+    size32_t fixedDiskRecordSize = actualDiskMeta->getFixedSize();
+    if (fixedDiskRecordSize && grouped)
+        fixedDiskRecordSize += 1;
+    return fixedDiskRecordSize;
+}
+
+
+//---------------------------------------------------------------------------------------------------------------------
+
+class CompoundProjectRowReader : extends CInterfaceOf<IAllocRowStream>, implements IRawRowStream, implements IDiskRowReader
```
richardkchapman: Again a comment on the purpose of the class might be useful

common/thorhelper/thorread.cpp
```diff
@@ -556,11 +703,458 @@ size32_t BinaryDiskRowReader::getFixedDiskRecordSize()
...
+    }
+
+    virtual const void *nextRow() override
+    {
+        size32_t rawInputSize;
+        const void * next = rawInputStream->nextRow(rawInputSize);
+        if (isSpecialRow(next))
+            return next;
+
+        size32_t size = translator->translate(allocatedBuilder.ensureRow(), unexpectedCallback, (const byte *)next);
```
richardkchapman: Should you release next?

ecl/hthor/hthor.cpp
```diff
@@ -10835,12 +10890,13 @@ bool CHThorNewDiskReadBaseActivity::openFilePart(ILocalOrDistributedFile * local
...
                            projectedDiskMeta->queryTypeInfo()->canInterpret() && projectedDiskMeta->queryTypeInfo()->canSerialize();
 
+
     /*
      * If a file part can be accessed local, then read it locally
      * If a file part supports a remote stream, then use that
      * Otherwise failover to the legacy remote access.
      */
-    const char * format = "thor";   // more - should extract from the current file (could even mix flat and csv...)
+    const char * format = helper.queryFormat();   // more - should extract from the current file (could even mix flat and csv...)
```
richardkchapman: Is the comment still relevant?

rtl/eclrtl/rtldynfield.cpp
```diff
@@ -1492,6 +1496,50 @@ class GeneralRecordTranslator : public CInterfaceOf<IDynamicTransform>
...
+            //MORE: Could special case casting from utf8 to utf8 similar to strings above
+        case type_unicode:
+        case type_varunicode:
+        {
+            offset = destType.buildString(builder, offset, field, srcSize, source);
+            break;
+        }
+        case type_set:
+        {
+            UNIMPLEMENTED; // JCS->GH - but perhaps can/should translate using iterator too?
```
richardkchapman: DO we support set in CSV files?

rtl/include/eclhelper.hpp
```diff
@@ -2394,6 +2395,9 @@ struct IHThorDiskReadBaseArg : extends IHThorCompoundBaseArg
     virtual unsigned getProjectedFormatCrc() = 0;
     virtual void getEncryptKey(size32_t & keyLen, void * & key) = 0;
     virtual void setCallback(IThorDiskCallback * callback) = 0;
+    virtual const char * queryFormat() = 0;
```
richardkchapman: This requires a change to the eclhelper version, I think (as other classes are derived from this)

common/thorhelper/thorfile.cpp
ghalliday: no.  It is one of the things I plan to change.

common/thorhelper/thorread.cpp
ghalliday: yes, that would be better.

common/thorhelper/thorread.cpp
ghalliday: I agree.  I think this code was copied.

common/thorhelper/thorread.cpp
ghalliday: no, because it is calling the interface that returns a row that is only valid until the next call.

ecl/hthor/hthor.cpp
ghalliday: yes, because this is getting it from the activity, rather than the dali meta info

rtl/eclrtl/rtldynfield.cpp
ghalliday: no - neither sets or datasets are supported.

rtl/include/eclhelper.hpp
ghalliday: yes, and it is not possible to make it backward compatible.  So potentially problematic.

ghalliday:
pushed to rebase and after some refactoring to compare smoketest results.

ghalliday:
Opened https://github.com/hpcc-systems/HPCC-Platform/pull/12934 as a replacement

common/thorhelper/thorread.cpp
```diff
@@ -444,108 +602,414 @@ const void *BinaryDiskRowReader::nextRow()
...
+    if (fixedDiskRecordSize && grouped)
+        fixedDiskRecordSize += 1;
+    return fixedDiskRecordSize;
+}
+
+
+//---------------------------------------------------------------------------------------------------------------------
+
+/*
+ * base class for reading a non-binary local file that uses IDynamicFieldValueFetcher to extract the values from
```
jpmcmu: Is there a reason the external file format would have to be non-binary? IE if I wanted to add parquet support would this be the base class I would use?

common/thorhelper/thorread.hpp
```diff
@@ -27,19 +27,56 @@
...
+public:
+    // Accessor functions to provide the basic information from the disk read
+    virtual const char * queryFormat() const = 0;
+    virtual unsigned getActualCrc() const = 0;
+    virtual unsigned getExpectedCrc() const = 0;
+    virtual unsigned getProjectedCrc() const = 0;
+    virtual IOutputMetaData * queryActualMeta() const = 0;
+    virtual IOutputMetaData * queryExpectedMeta() const = 0;
+    virtual IOutputMetaData * queryProjectedMeta() const = 0;
+    virtual const IPropertyTree * queryOptions() const = 0;
```
jpmcmu: Based on the CSV option test below it looks like generic options would be supported and are injected here. Is that correct?

jpmcmu (commented):
@ghalliday Looks good. I left a few questions for you on usage of the code. I also have some questions, that are beyond the scope of this PR, about how this code could be extended to support "foreign" data sources like HDFS. As an example being able to query HDFS about # of partitions and partition locality would be beneficial. I will follow up offline.

PR #12819, opened 2019-07-24
HPCC-22580 Improve error reporting in embedded Java code

richardkchapman:
Dump stack traces for any Java exceptions to log file.

Also provide a way to output to standard HPCC log from within embedded java
code.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

## Testing:
Added a test in regression suite. Tested exception output manually while debugging HPCC-22579

### DIFF
--- a/plugins/javaembed/javaembed.cpp
+++ b/plugins/javaembed/javaembed.cpp
@@ -86,6 +86,10 @@ static void UNSUPPORTED(const char *feature)
 namespace javaembed {
 
 static jmethodID throwable_toString;
+static jmethodID throwable_getStackTrace;
+static jmethodID throwable_getCause;
+static jmethodID frame_toString;
+
 
 static void forceGC(class CheckedJNIEnv* JNIenv);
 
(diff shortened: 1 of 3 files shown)


richardkchapman:
@rpastrana Please review

testing/regress/ecl/javaembed.ecl
```diff
@@ -27,10 +27,15 @@ string jcat(string a, string b) := EMBED(java)
 ENDEMBED;
 
 integer jadd(integer a, integer b) := EMBED(java)
+import com.HPCCSystems.HpccUtils;
```
rpastrana: as an fyi we've been using org.hpccsystems as our namespace

testing/regress/ecl/javaembed.ecl
```diff
@@ -27,10 +27,15 @@ string jcat(string a, string b) := EMBED(java)
...
+class myClass
```
rpastrana: very minor, classes tend to be Capitalized

plugins/javaembed/javaembed.cpp
```diff
@@ -4929,6 +4985,16 @@ JNIEXPORT jclass JNICALL Java_com_HPCCSystems_HpccClassLoader_defineClassForEmbe
 
 }
 
+JNIEXPORT void JNICALL Java_com_HPCCSystems_HpccUtils_log(JNIEnv *JNIenv, jclass, jstring msg)
+{
+    if (msg)
+    {
+        const char *text = JNIenv->GetStringUTFChars(msg, 0);
+        DBGLOG("javaembed: user: %s", text);
```
rpastrana: Is "user" in this case the audience?

plugins/javaembed/javaembed.cpp
```diff
@@ -120,12 +124,58 @@ class CheckedJNIEnv : private JNIEnv
...
+        ignoreException();
+        const char *text = JNIEnv::GetStringUTFChars(msg, 0);
+        DBGLOG("javaembed: exception: %s", text);
+        JNIEnv::ReleaseStringUTFChars(msg, text);
+        ignoreException();
+        JNIEnv::DeleteLocalRef(msg);
+
+        jobjectArray frames = (jobjectArray) JNIEnv::CallObjectMethod(exception, throwable_getStackTrace);
+        ignoreException();
+        jsize length = JNIEnv::GetArrayLength(frames);
```
rpastrana: are we guaranteed frames will be valid if we ignored a possible exception? It might not matter if JNIEnv::GetArrayLength sets length<= 0

rpastrana (commented):
@richardkchapman left a few comments/questions

plugins/javaembed/javaembed.cpp
richardkchapman: user just means this tracing came from the user's code not the platform code

plugins/javaembed/javaembed.cpp
richardkchapman: Unless there's a logic error, getStackTrace should always succeed. the main reason for the "ignoreException" is that there is a mode of running JNI code that detects whether you have checked the exception flag after each call. I could throw an internal error inside ignoreException if it detects an error.

testing/regress/ecl/javaembed.ecl
richardkchapman: You may have, but the com.HPCCSystems namespace as used by the embedded Java code predates your usage, I suspect...

It's probably too late to unify the two usages now.

richardkchapman:
@rpastrana Repushed

rpastrana (approved):
@richardkchapman looks good.

PR #12826, opened 2019-07-29
HPCC-22112 Provide tracing when Record Translation has taken effect

richardkchapman:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Premature optimization

## Smoketest:

## Testing:
Observed logs after smoketest run

### DIFF
--- a/initfiles/componentfiles/configschema/xsd/roxie.xsd
+++ b/initfiles/componentfiles/configschema/xsd/roxie.xsd
@@ -274,6 +274,9 @@
                     <xs:attribute name="soapTraceLevel" type="xs:nonNegativeInteger" hpcc:displayName="Soap trace Level"
                                   hpcc:presetValue="1"
                                   hpcc:tooltip="Level of detail in reporting SOAPCALL information(set to 0 for none, 1 for normal, >1 or more for extended)"/>
+                    <xs:attribute name="traceTranslations" type="xs:boolean" hpcc:displayName="Trace record layout translations"
+                                  hpcc:presetValue="true"
+                                  hpcc:tooltip="Trace record layout translations to log file"/>
                     <xs:attribute name="traceEnabled" type="xs:boolean" hpcc:displayName="Trace Enabled"
                                   hpcc:presetValue="false"
                                   hpcc:tooltip="TRACE activity output enabled by default (can be overridden in workunit or query)"/>
--- a/common/thorhelper/thorcommon.cpp
+++ b/common/thorhelper/thorcommon.cpp
@@ -2089,6 +2089,8 @@ static bool getTranslators(Owned<const IDynamicTransform> &translator, Owned<con
         if ((projectedFormat != sourceFormat) && (projectedCrc != sourceCrc))
         {
             translator.setown(createRecordTranslator(projectedFormat->queryRecordAccessor(true), sourceFormat->queryRecordAccessor(true)));
+            DBGLOG("Record layout translator created for %s", tracing);
+            translator->describe();
 
             if (!translator->canTranslate())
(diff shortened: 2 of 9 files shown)


richardkchapman:
@jakesmith Please review (especially thor changes)

jakesmith:
A couple of questions:
1) Should DBGLOG's be PROGLOG's ? (and in doDescribe in GeneralRecordTranslator also)
2) Do we need a 'traceTranslations' option in hthor and thor too, or is in roxie only ok?

@richardkchapman

richardkchapman:
I didn't bother making the tracing optional in thor/hthor as I did not expect the volume of tracing to be an issue in those platforms.

As for DBGLOG vs PROGLOG - @shamser can you tell us the current rules?

initfiles/componentfiles/configschema/xsd/roxie.xsd
```diff
@@ -274,6 +274,9 @@
                     <xs:attribute name="soapTraceLevel" type="xs:nonNegativeInteger" hpcc:displayName="Soap trace Level"
                                   hpcc:presetValue="1"
                                   hpcc:tooltip="Level of detail in reporting SOAPCALL information(set to 0 for none, 1 for normal, >1 or more for extended)"/>
+                    <xs:attribute name="traceTranslations" type="xs:boolean" hpcc:displayName="Trace record layout translations"
+                                  hpcc:presetValue="true"
```
shamser: Should this tracing be enabled by default?  As it's debug tracing, shouldn't this be disabled by default?

shamser (commented):
@richardkchapman

initfiles/componentfiles/configschema/xsd/roxie.xsd
richardkchapman: It's tracing the users want to look out for... Does that mean it should be a PROGLOG?

I've never been particularly clear on the rules for using the various logging macros.

richardkchapman:

jakesmith:
I thought the idea was that at some point we'd disabled DBGLOG in release builds.
So if the tracing is still wanted in release, we should use something else (like PROGLOG).

initfiles/componentfiles/configschema/xsd/roxie.xsd
shamser: DBGLOG is temporary logging used for debugging/investigation purposes (ithe plan is to have DBGLOG disabled by default.).  For user information, it should be LOG(MCuserInfo, format,...).  There is an explaination of the difference types near the top of jlog.hpp.

ghalliday:
@richardkchapman  is this suitable for a 7.4 release?
The tracing could get very large - e.g. some nasty dataset with nested complex datasets.  Should there be a compact description of what is being translated?

richardkchapman:
@ghalliday I have tweaked the tracing a bit to bake it more compact and also more useful.

I don't want to change the DBGLOGs at this time:

1. None of the message categories listed in jlog seem more appropriate. It's not "progress", nor is it a warning or error. It's information, and DBGLOG is closest to covering that. 
2. There are literally 1000's of uses of DBGLOG throughout the code using the same assumption.
3. If we need to look at all DBGLOG and PROGLOG to determine which are progress, which are info, and which are "temporary to be removed in release mode" then that's a separate task, but I'm not sure the last of those is a useful concept (we can, and do, put DBGLOG's inside of #ifdef's)

ghalliday:
@richardkchapman first scan looks good.  Can you squash and I'll recheck it.

richardkchapman:
@ghalliday Squashed

ghalliday:
Slight concern about the size of the output, but we can monitor.

PR #12831, opened 2019-07-30
HPCC-22601 RuntimeStatistic are not being updated in a thread safe way

shamser:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jstats.cpp
+++ b/system/jlib/jstats.cpp
@@ -2091,7 +2091,35 @@ protected:
 
 void CRuntimeStatistic::merge(unsigned __int64 otherValue, StatsMergeAction mergeAction)
 {
-    value = mergeStatisticValue(value, otherValue, mergeAction);
+    switch (mergeAction)
+    {
+    case StatsMergeKeepNonZero:
+        if (otherValue && !value)
+        {
+            unsigned __int64 zero = 0;
+            value.compare_exchange_strong(zero, otherValue);
+        }
+        break;
+    case StatsMergeAppend:
+    case StatsMergeReplace:
+        value = otherValue;
+        break;
+    case StatsMergeSum:
+        addAtomic(otherValue);
+        break;
+    case StatsMergeMin:
+        value.store_min(otherValue);
+        break;
+    case StatsMergeMax:
+        value.store_max(otherValue);
+        break;
+    default:
+#ifdef _DEBUG
+        throwUnexpected();
+#else
+        value = otherValue;
+#endif
+    }
 }
 
 //--------------------------------------------------------------------------------------------------------------------


shamser:

system/jlib/jstats.cpp
```diff
@@ -2091,7 +2091,35 @@ class TreeNodeStatisticGatherer : public CInterfaceOf<IStatisticGatherer>
...
 void CRuntimeStatistic::merge(unsigned __int64 otherValue, StatsMergeAction mergeAction)
 {
-    value = mergeStatisticValue(value, otherValue, mergeAction);
+    switch (mergeAction)
+    {
+    case StatsMergeKeepNonZero:
+        if (otherValue)
+        {
+            unsigned __int64 zero = 0;
+            value.compare_exchange_weak(zero, otherValue);
```
jakesmith: this means that if otherValue is non zero, but something else has already set value, this will fail, i.e. the new otherValue will be ignored.

Prior to this change, sequential attempts to merge with StatsMergeKeepNonZero would succeed, and result in latest value being kept.

I think this should just set value (if otherValue is non-zero), to ensure latest is kept.

jakesmith:
@shamser - 1 comment, but would be good if @ghalliday looked as well.

system/jlib/jstats.cpp
shamser: @jakesmith Originally in mergeStatisticValue, it had: 
    case StatsMergeKeepNonZero:
        if (prevValue)
            return prevValue;
        return newValue;

So, it only updated to the newValue if the prevValue was zero.

shamser:
@ghalliday Please can you have a look too?

system/jlib/jstats.cpp
jakesmith: Ah yes, ok. 
'StatsMergeKeepNonZero' sounds like it will keep the new value if non-zero..

@ghalliday - is it correct to ignore new value with this mode?

ghalliday: Should be compare_exchange_strong.   compare_exchange_weak is for use in loops (details of the difference are connected to how how cas is implemented on non intel architectures).

ghalliday: Yes, I think that is correct.

ghalliday (commented):
@shamser one comment

shamser:
@ghalliday

system/jlib/jstats.cpp
```diff
@@ -2091,7 +2091,35 @@ class TreeNodeStatisticGatherer : public CInterfaceOf<IStatisticGatherer>
 
 void CRuntimeStatistic::merge(unsigned __int64 otherValue, StatsMergeAction mergeAction)
 {
-    value = mergeStatisticValue(value, otherValue, mergeAction);
+    switch (mergeAction)
+    {
+    case StatsMergeKeepNonZero:
+        if (otherValue)
```
ghalliday: @shamser I wonder if this should say

if (otherValue && !value)

to avoid the more expensive cas if the value has already been set.
I am happy to merge as it is.

shamser:
@ghalliday I've added the extra check to avoid cas and squashed.

richardkchapman:
Spurious apostrophe in commit title

shamser:
@richardkchapman Apostrophe removed

PR #12865, opened 2019-08-08
HPCC-22599 Disable Data Patterns for unsupported file types

GordonSmith:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/src/src/DataPatternsWidget.ts
+++ b/esp/src/src/DataPatternsWidget.ts
@@ -49,6 +49,8 @@ type _TabContainerWidget = {
     addChild(tabItem: any, pos: number);
 };
 
+export const supportedFileType = (contentType: string): boolean => ["flat", "csv", "thor"].indexOf((contentType || "").toLowerCase()) >= 0;
+
 export interface DataPatternsWidget extends _TabContainerWidget {
 }
 
--- a/esp/src/eclwatch/LFDetailsWidget.js
+++ b/esp/src/eclwatch/LFDetailsWidget.js
@@ -34,6 +34,7 @@ define([
     "hpcc/FileBelongsToWidget",
     "src/FileSpray",
     "hpcc/FileHistoryWidget",
+    "src/DataPatternsWidget",
 
     "dojo/text!../templates/LFDetailsWidget.html",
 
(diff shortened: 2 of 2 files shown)


GordonSmith:
@miguelvazq please review

miguelvazq (approved):

esp/src/src/DataPatternsWidget.ts
```diff
@@ -49,6 +49,8 @@ type _TabContainerWidget = {
     addChild(tabItem: any, pos: number);
 };
 
+export const supportedFileType = (contentType: string): boolean => ["flat", "csv", "thor"].indexOf((contentType || "").toLowerCase()) >= 0;
```
richardkchapman: What about XML?

richardkchapman:
@dcamper Please review

esp/src/src/DataPatternsWidget.ts
dcamper: The script submits a workunit containing generated ECL.  In that ECL is something like:

ds := DATASET(filePath, RECORDOF(filePath, LOOKUP), flat);

In other words, it relies on compile-time record definition lookup.  Is that available/correct with XML files?  My understand is that it was not (always) available and, if not, XML files should be ignored.

GordonSmith: @dcamper I didn't realize that was the issue - in theory I do have the actual RECORD available to me - if I used that, does it mean XML would also be supported (this is the ECL record as per the Logical File details page).

dcamper: DataPatterns.Profile() doesn't actually care about the underlying file format at all.  You can even profile a file on the landing zone with a properly-crafted DATASET pathname and record definition.  Bottom line, if you can determine the record layout then by all means, let's use it!  And please give me a hint offline, as I have some other code that needs modifying....

dcamper: FYI:  I have created a new Jira for the XML case.  As per a discussion I had with Gordon, this PR should be focused on only the undefined filetype case.

dcamper:
@GordonSmith Back to you for XML file type adjustment, whatever that may be.

PR #12882, opened 2019-08-14
HPCC-22584 Support new log format in WsTopology

wangkx:
1. Add readLogMessageFields() to read the information about the
log fields in new log format;
2. Rewrite the functions of reading last log time to support new
log format;
3. Clean findTimestampAndLT(), readALogLine() and addALogLine().

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/services/ws_topology/ws_topologyService.cpp
+++ b/esp/services/ws_topology/ws_topologyService.cpp
@@ -43,6 +43,7 @@ static const char* MACHINE_URL = "MachineInfoAccess";
 static const unsigned THORSTATUSDETAILS_REFRESH_MINS = 1;
 static const long LOGFILESIZELIMIT = 100000; //Limit page size to 100k
 static const long AVERAGELOGROWSIZE = 10000;
+static const unsigned LOG_DATE_TIME_LENGTH = 19;
 const char* TEMPZIPDIR = "tempzipfiles";
 
 void CWsTopologyEx::init(IPropertyTree *cfg, const char *process, const char *service)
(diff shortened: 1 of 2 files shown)


wangkx:
@afishbeck please review.

afishbeck (approved):
@wangkx looks fine

esp/services/ws_topology/ws_topologyService.cpp
```diff
@@ -539,64 +540,118 @@ bool CWsTopologyEx::findTimestampAndLT(const char *logname, IFile* rFile, ReadLo
...
-    if (!hasAnotherLine || (bytesRemaining < 28))
-        return true;
+    if (finger > contentStart)
+    {//Try to read LogMessageFields from the first log line
+        StringBuffer logLine;
+        logLine.append(finger - contentStart, contentStart);
+        readLogReq.logfields = getMessageFieldsFromHeader(logLine);
+        if (readLogReq.logfields > 0)
+        {
+            readLogReq.logDatePos = getPositionOfField(readLogReq.logfields, MSGFIELD_date);
```
ghalliday: I think this is assuming that the positions are offsets within the field, (which they were for a little while), but they now return field numbers.
@shamser would it make sense to have a utility function that filled in an array with a set of pointers to the start and end of each field, which could then be used by functions that are parsing log files to quickly extract the different values.

esp/services/ws_topology/ws_topologyService.cpp
```diff
@@ -539,64 +540,118 @@ bool CWsTopologyEx::findTimestampAndLT(const char *logname, IFile* rFile, ReadLo
...
-        CDateTime dt;
-        pTr = readALogLine(pTr, bytesRemaining, readLogReq.ltBytes, logLine, hasAnotherLine);
-        if ((logLine.length() > 27) && readLogTime((char*) logLine.str(), 9, 19, dt))
-            latestLogTime = dt;
+        size32_t bytesRead = rIO->read(readFrom, readSize, contentBuffer.clear().reserve(readSize));
+        if (bytesRead != readSize)
+            throw MakeStringException(ECLWATCH_CANNOT_READ_FILE, "Failed to read file %s.", logName);
+            
+        if (!previousPartialLine.isEmpty())
+            contentBuffer.append(previousPartialLine); //Add the leftover from previous contentBuffer
```
ghalliday: Isn't this round the wrong way?  partial should come before the new content.  More efficient would be to read the data onto the end of the previous - or use a ISerialStream

esp/services/ws_topology/ws_topologyService.cpp
```diff
@@ -539,64 +540,118 @@ bool CWsTopologyEx::findTimestampAndLT(const char *logname, IFile* rFile, ReadLo
...
+        readFrom = fileSize - readSize; //Search from the last chuck
+
+    StringBuffer contentBuffer, previousPartialLine;
+    while (1)
     {
-        CDateTime dt;
-        pTr = readALogLine(pTr, bytesRemaining, readLogReq.ltBytes, logLine, hasAnotherLine);
-        if ((logLine.length() > 27) && readLogTime((char*) logLine.str(), 9, 19, dt))
-            latestLogTime = dt;
+        size32_t bytesRead = rIO->read(readFrom, readSize, contentBuffer.clear().reserve(readSize));
```
ghalliday: This looks like it will read the whole of the rest of the file, just to read the last log line?  If so that seems inefficient - it could be better to read near the end of the file, and work backwards (assuming there are new newlines within log lines - although I'm not sure this code would cope with newlines in log lines anyway).

ghalliday (commented):
@shamser please can you check this code.

esp/services/ws_topology/ws_topologyService.cpp
shamser: getPositionOfField returns the column number.  (Although it appears that the the fields are fixed width, it isn't - PID and TID may be wider to allow for values being larger than digits. See https://track.hpccsystems.com/browse/HPCC-22651)

esp/services/ws_topology/ws_topologyService.cpp
```diff
@@ -539,64 +540,118 @@ bool CWsTopologyEx::findTimestampAndLT(const char *logname, IFile* rFile, ReadLo
...
-        return true;
+    if (finger > contentStart)
+    {//Try to read LogMessageFields from the first log line
+        StringBuffer logLine;
+        logLine.append(finger - contentStart, contentStart);
+        readLogReq.logfields = getMessageFieldsFromHeader(logLine);
+        if (readLogReq.logfields > 0)
+        {
+            readLogReq.logDatePos = getPositionOfField(readLogReq.logfields, MSGFIELD_date);
+            readLogReq.logTimePos = getPositionOfField(readLogReq.logfields, MSGFIELD_milliTime);
```
shamser: There's no guarantee that the system will be configured to log millitime.  It's possible to configure to log micro time or just time.

esp/services/ws_topology/ws_topologyService.cpp
```diff
@@ -539,64 +540,118 @@ bool CWsTopologyEx::findTimestampAndLT(const char *logname, IFile* rFile, ReadLo
...
+    return false;
+}
 
-    return true;
+const char* CWsTopologyEx::readLogTimeFromDataBuffer(StringBuffer& dataBuffer, ReadLog& readLogReq, CDateTime& latestLogTime)
+{
+    const char* startPtr = dataBuffer.str();
+    const char* lineEndPtr = startPtr + dataBuffer.length();
+    const char* finger = lineEndPtr - 1;
+    unsigned limit = readLogReq.logDatePos + LOG_DATE_TIME_LENGTH;
```
shamser: readLogReq.logDatePos will contain the column position, so limit isn't going to have the expected value.

esp/services/ws_topology/ws_topologyService.cpp
```diff
@@ -539,64 +540,118 @@ bool CWsTopologyEx::findTimestampAndLT(const char *logname, IFile* rFile, ReadLo
...
-    bool hasAnotherLine = false;
-    pTr = readALogLine(pTr, bytesRemaining, readLogReq.ltBytes, logLine, hasAnotherLine);
-    if (!hasAnotherLine || (bytesRemaining < 28))
-        return true;
+    if (finger > contentStart)
+    {//Try to read LogMessageFields from the first log line
+        StringBuffer logLine;
+        logLine.append(finger - contentStart, contentStart);
+        readLogReq.logfields = getMessageFieldsFromHeader(logLine);
+        if (readLogReq.logfields > 0)
```
shamser: If readLogReq.logfields==0, then readLogReq.logfields should be set to MSGFIELD_LEGACY to use the legacy format

esp/services/ws_topology/ws_topologyService.cpp
```diff
@@ -798,6 +853,9 @@ void CWsTopologyEx::readTpLogFileRequest(IEspContext &context, const char* fileN
     readLogReq.prevPage = -1;
     readLogReq.nextPage = -1;
     readLogReq.lastHours = 0;
+    readLogReq.logfields = 0;//legacy log format
+    readLogReq.logDatePos = 9;//legacy log format
+    readLogReq.logTimePos = 20;//legacy log format
```
shamser: See the comment in Live 574 regarding using MSGFIELD_LEGACY. The legacy column number shouldn't be hard coded.

shamser (changes requested):
@wangkx I found a few problems.  Also, this code may simplified substantially by using CStreamLineReader (defined in system/jlib/jio.cpp).  You won't have to handle the processing of end of line characters or splitting the input stream into line sized chunks. It is used in WsWuInfo::readWorkunitLog (esp/services/ws_workunits/ws_workunitHelpers.cpp ~line2054) to process thor log files.

wangkx:
@shamser I need to read the last date/time in a log file. That is why I check the last content chuck of a log file first. If no data/time, I go to previous chucks...  Are you still suggesting that I should use the CStreamLineReader?

shamser:
@wangkx No,  I missed that part where you were reading blocks from the end.  It would nice to have the reading the of the rows separate from from the processing (e.g. readLogTimeFromDataBuffer). How would you feel about implementing CStreamReverseLineReader()?  It would probably be useful for other uses too.

esp/services/ws_topology/ws_topologyService.cpp
shamser: We have log format configurable by modifying environment.conf.  Ideally, it should support any format that is configurable.  However,  I think @richardkchapman mentioned that if it was going to be a lot of work, not to bother trying to support non-standard format.

esp/services/ws_topology/ws_topologyService.cpp
shamser: Sorry, about the change.  The change was necessary following https://track.hpccsystems.com/browse/HPCC-22651. (It wasn't being used anywhere else at the time.)

esp/services/ws_topology/ws_topologyService.cpp
shamser: Sorry, about the change. The change was necessary following https://track.hpccsystems.com/browse/HPCC-22651. I didn't see it being used anywhere else at the time.

wangkx:
I only need one date/time (the latest one). We may create another JIRA for implementing CStreamReverseLineReader().

esp/services/ws_topology/ws_topologyService.cpp
wangkx: @shamser if a system is configured to log millitime, what will the getPositionOfField(logfields, MSGFIELD_time) return? How to know which time format is used in a log file without checking system configuration?

wangkx: How to know a log field is not used in a log file without checking system configuration?

shamser: If time field is not configured to be logged, then getPositionOfField(logfields, MSGFIELD_time) will return 0.  To check if milliTime is being used:  
if (readLogReq.logfields & MSGFIELD_milliTime)

wangkx: Thanks.

wangkx:
A new PR will be issued for adding the changes due to new getPositionOfField() functionality.

PR #12888, opened 2019-08-15
HPCC-22021 Enhance log agent identification information

timothyklemm:
Read optional type and group values for any log agent.
Enable type and group values to be propagated for WsLogServiceESPAgent agents.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/logging/logginglib/loggingagentbase.cpp
+++ b/esp/logging/logginglib/loggingagentbase.cpp
@@ -17,6 +17,7 @@
 
 #include "LoggingErrors.hpp"
 #include "loggingagentbase.hpp"
+#include <set>
 
 static const char* const defaultTransactionTable = "transactions";
 static const char* const defaultTransactionAppName = "accounting_log";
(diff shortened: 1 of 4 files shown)


timothyklemm:
@wangkx Please review.

esp/logging/logginglib/loggingagentbase.cpp
```diff
@@ -330,6 +331,103 @@ IEspUpdateLogRequestWrap* CLogContentFilter::filterLogContent(IEspUpdateLogReque
...
+}
+
+bool CLogAgentBase::CVariantIterator::isValid()
+{
+    return (m_variantIt != m_agent->agentVariants.end());
+}
+
+const IEspLogAgentVariant& CLogAgentBase::CVariantIterator::query()
+{
+    return *(m_variantIt->get());
```
wangkx: Could the get() return an nullptr?

esp/logging/logginglib/loggingagentbase.cpp
```diff
@@ -330,6 +331,103 @@ IEspUpdateLogRequestWrap* CLogContentFilter::filterLogContent(IEspUpdateLogReque
...
+    CriticalBlock block(poolLock);
+    PooledStrings::iterator it = pool.find(tmp);
+
+    if (it != pool.end())
+        return it->getLink();
+    pool.insert(tmp);
+    return tmp.getClear();
+}
+
+bool CLogAgentBase::initVariants(IPropertyTree * cfg)
```
wangkx: Remove the extra space in the front of '*'.

esp/logging/logginglib/loggingagentbase.hpp
```diff
@@ -249,14 +250,67 @@ class CUpdateLogRequestWrap : implements IEspUpdateLogRequestWrap, public CInter
...
+struct IEspLogAgentVariantComparator
+{
+    // Implement alternative to std::less for potential set ordering
+    bool operator () (const Owned<const IEspLogAgentVariant>& lhs, const Owned<const IEspLogAgentVariant>& rhs) const
+    {
+        return (*this)(lhs.get(), rhs.get());
+    }
+    bool operator () (const IEspLogAgentVariant* lhs, const IEspLogAgentVariant* rhs) const
+    {
+        return compare(lhs, rhs) < 0;
```
wangkx: Why < 0?

esp/logging/logginglib/loggingagentbase.hpp
```diff
@@ -340,6 +394,39 @@ class LOGGINGCOMMON_API CLogAgentBase : public CInterface, implements IEspLogAge
 
     CLogAgentBase() { services[0] = LGSTterm; };
     virtual ~CLogAgentBase() {};
+
+public:
+    bool initVariants(IPropertyTree * cfg) override;
```
wangkx: Remove the extra space in the front of '*'.

esp/logging/logginglib/loggingagentbase.hpp
```diff
@@ -340,6 +394,39 @@ class LOGGINGCOMMON_API CLogAgentBase : public CInterface, implements IEspLogAge
...
+    Variants agentVariants;
+    class CVariantIterator : implements CInterfaceOf<const IEspLogAgentVariantIterator>
+    {
+    public:
+        CVariantIterator(const CLogAgentBase& agent);
+        ~CVariantIterator();
+        bool first() override;
+        bool next() override;
+        bool isValid() override;
+        const IEspLogAgentVariant & query() override;
```
wangkx: Remove the extra space in the front of '&'.

wangkx (commented):
@timothyklemm several minor comments

esp/logging/logginglib/loggingagentbase.cpp
timothyklemm: Technically, an Owned object can return nullptr but that would signal a logic error elsewhere in this change because nullptr should never be inserted. The more likely scenario, which is a usage error, is being called when isValid returns false. In this case, the problem is attempting to get a pointer from an undefined Owned instance.

esp/logging/logginglib/loggingagentbase.hpp
timothyklemm: Sets determine ordering using an equivalent of '<' to decide if lhs should precede rhs. The compare method, like stricmp that it calls, returns a negative number when lhs precedes rhs and the operator result is true when a negative number is returned.

wangkx (approved):
@looks fine.

timothyklemm:

esp/logging/logginglib/loggingagentbase.cpp
```diff
@@ -330,6 +331,108 @@ IEspUpdateLogRequestWrap* CLogContentFilter::filterLogContent(IEspUpdateLogReque
...
+        throw MakeStringException(0, "CVariantIterator::query called in invalid state");
+    IEspLogAgentVariant* entry = m_variantIt->get();
+    if (nullptr == entry)
+        throw MakeStringException(0, "CVariantIterator::query encountered a NULL variant");
+    return *entry;
+}
+
+CLogAgentBase::CVariant::CVariant(const char* name, const char* type, const char* group)
+{
+    m_name.setown(normalize(name));
```
ghalliday: What is the advantage in normalizing the strings?  It complicates the code and doesn't seem to be required.

ghalliday (commented):
@timothyklemm one question.  Looks ok to merge.

esp/logging/logginglib/loggingagentbase.cpp
timothyklemm: It is an attempt to reduce persistent and redundant memory buffers. The overall feature is required in limited situations where the additional overhead may be acceptable. My concern involves the potential for this to be used in a manner other than expected. When I considered the possibility that the ability to add two sets of unrelated agents to a manager means somebody could add 10, or 20, or more, to a single manager. In such a case, I see the strings being replicated many times and I wanted to avoid it without artificially limiting the user.

I would want to keep the trim and case insensitivity, but the pooling of strings could be removed if I've overthought it.

PR #12889, opened 2019-08-15
HPCC-17261 New feature #$ for location of currently expanding macro

ghalliday:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/ecl/hql/hqlatoms.cpp
+++ b/ecl/hql/hqlatoms.cpp
@@ -210,6 +210,7 @@ IAtom * _graphLocal_Atom;
 IAtom * groupAtom;
 IAtom * groupedAtom;
 IAtom * hashAtom;
+IAtom * _hash_dollar_Atom;
 IAtom * headingAtom;
 IAtom * _hidden_Atom;
 IAtom * hintAtom;
@@ -685,6 +686,7 @@ MODULE_INIT(INIT_PRIORITY_HQLATOM)
     MAKEATOM(group);
     MAKEATOM(grouped);
     MAKEATOM(hash);
+    MAKESYSATOM(hash_dollar);
     MAKEATOM(heading);
     MAKESYSATOM(hidden);
     MAKEATOM(hint);
--- a/ecl/hql/hqlatoms.hpp
+++ b/ecl/hql/hqlatoms.hpp
@@ -214,6 +214,7 @@ extern HQL_API IAtom * _graphLocal_Atom;
 extern HQL_API IAtom * groupAtom;
 extern HQL_API IAtom * groupedAtom;
 extern HQL_API IAtom * hashAtom;
+extern HQL_API IAtom * _hash_dollar_Atom;
 extern HQL_API IAtom * headingAtom;
 extern HQL_API IAtom * _hidden_Atom;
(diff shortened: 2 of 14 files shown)


ghalliday:
Initial push to check smoketest etc.  Semantics may change - requesting feedback in the jira.

ghalliday:
@shamser please review

shamser (approved):
@ghalliday Code reviewed. Documentation needs to be updated.

ghalliday:
@dcamper please can you double check the semantics.  Essentially #$ inside a macro has the same meaning as $ at the same place if it was not in a macro.

dcamper:
Gavin, I'm installing your branch to give this feature a test drive.

dcamper:
@ghalliday Enclosing a directory of ECL code I used for initial testing.

- When the module is simple (does not have arguments), the syntax for using #$ is different depending on whether the module is file-based or directory-based.  See FileMod.AddValue() vs DirMod.AddValue(), kicked off via BWR_Test_[File|Dir]Mod.ecl.  That kind of makes sense, but not entirely.  Is it working as designed?
- When the module accepts arguments, I cannot get #$ to work at all.  I tried a number of techniques.  In the enclosed code, I tried re-instantiating the module with the arguments that should have been visible to the function macro, but that did not work.  BWR_Test_[File|Dir]Mod2.ecl contains what I tried.

I think the ideal usage would be that the symbol #$ is translated into the immediately enclosing module, whether that module is file- or directory-based, and whether it is simple or has arguments.  That would simplify the usage of the symbol (and probably complicate your work immensely).  Is that possible?

[FeatureTesting.zip](https://github.com/hpcc-systems/HPCC-Platform/files/3525407/FeatureTesting.zip)

ghalliday:
@dcamper 
The idea was for the functionality to match $, and $ for a file currently means the directory containing the file, so #$ means the same for a macro.

If you want to pass anything into a macro I think you need to do it via a paramater e.g.
```
        factor := (string)NUM_CONSTANT;
        EXPORT AddMultipliedValue(inFile, fieldName, scale = factor) := FUNCTIONMACRO
```
Although I suspect that will only work for a small subset of the things you would like to use it for.

I can see why it be useful to have something to access the context of the macro, but I can't immediately see how it could be implemented - the information doesn't currently exist.

The aim of this change was to allow all imports to be relative - which I think this provides.  Does it solve that problem?

richardkchapman:
@dcamper any comment?

dcamper:
The requested functionality is something I asked for as a like-to-have, but others have concrete projects that depend on it.  I've requested feedback from that group.

kev77log:
I am reasonably certain that this will provide the functionality that was originally requested in https://track.hpccsystems.com/browse/HPCC-22528.

dcamper:
@ghalliday Kevin's project has the priority here.  If he says that this change will work for him, then we can move forward with this implementation.  Thanks!

ghalliday:
@richardkchapman

PR #12914, opened 2019-08-23
HPCC-22682 Limit access to files marked as Restricted

shamser:
## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dali/base/dadfs.cpp
+++ b/dali/base/dadfs.cpp
@@ -269,6 +269,8 @@ public:
             return str.append(": Lookup connection timeout: ").append(errstr);
         case DFSERR_FailedToDeleteFile:
             return str.append(": Failed to delete file: ").append(errstr);
+        case DFSERR_RestrictedFileAccessDenied:
+            return str.append(": Access to restricted file denied: ").append(errstr);
         }
         return str.append("Unknown DFS Exception");
     }
@@ -1044,8 +1046,8 @@ public:
 
     IDistributedFile *dolookup(CDfsLogicalFileName &logicalname, IUserDescriptor *user, bool writeattr, bool hold, bool lockSuperOwner, IDistributedFileTransaction *transaction, unsigned timeout);
 
-    IDistributedFile *lookup(const char *_logicalname, IUserDescriptor *user, bool writeattr, bool hold, bool lockSuperOwner, IDistributedFileTransaction *transaction, unsigned timeout);
-    IDistributedFile *lookup(CDfsLogicalFileName &logicalname, IUserDescriptor *user, bool writeattr, bool hold, bool lockSuperOwner, IDistributedFileTransaction *transaction, unsigned timeout);
+    IDistributedFile *lookup(const char *_logicalname, IUserDescriptor *user, bool writeattr, bool hold, bool lockSuperOwner, IDistributedFileTransaction *transaction, bool privilegedUser, unsigned timeout) override;
+    IDistributedFile *lookup(CDfsLogicalFileName &logicalname, IUserDescriptor *user, bool writeattr, bool hold, bool lockSuperOwner, IDistributedFileTransaction *transaction, bool privilegedUser, unsigned timeout) override;
 
     /* createNew always creates an unnamed unattached distributed file
      * The caller must associated it with a name and credentials when it is attached (attach())
@@ -1056,7 +1058,7 @@ public:
     IDistributedSuperFile *createNewSuperFile(IPropertyTree *tree, const char *optionalName=nullptr);
     void removeSuperFile(const char *_logicalname, bool delSubs, IUserDescriptor *user, IDistributedFileTransaction *transaction);
 
-    IDistributedFileIterator *getIterator(const char *wildname, bool includesuper,IUserDescriptor *user);
+    IDistributedFileIterator *getIterator(const char *wildname, bool includesuper,IUserDescriptor *user,bool isPrivilegedUser);
     IDFAttributesIterator *getDFAttributesIterator(const char *wildname, IUserDescriptor *user, bool recursive, bool includesuper,INode *foreigndali,unsigned foreigndalitimeout);
     IPropertyTreeIterator *getDFAttributesTreeIterator(const char *filters, DFUQResultField* localFilters, const char *localFilterBuf,
         IUserDescriptor *user, bool recursive, bool& allMatchingFilesReceived, INode *foreigndali,unsigned foreigndalitimeout);
@@ -1377,7 +1379,7 @@ public:
         for (;;)
         {
             // Transaction files have already been unlocked at this point, delete all remaining files
-            Owned<IDistributedFile> file = queryDistributedFileDirectory().lookup(lfn, user, true, false, true, NULL, SDS_SUB_LOCK_TIMEOUT);
+            Owned<IDistributedFile> file = queryDistributedFileDirectory().lookup(lfn, user, true, false, true, nullptr, defaultPrivilegedUser, SDS_SUB_LOCK_TIMEOUT);
             if (!file.get())
                 return;
             StringBuffer reason;
@@ -1672,7 +1674,7 @@ public:
             return LINK(ret);
         else
         {
-            ret = queryDistributedFileDirectory().lookup(name, udesc, false, false, false, this, timeout);
+            ret = queryDistributedFileDirectory().lookup(name, udesc, false, false, false, this, defaultPrivilegedUser, timeout);
             if (ret)
                 queryCreate(name, ret, true);
             return ret;
@@ -2497,11 +2499,12 @@ class CDistributedFileIterator: public CDistributedFileIteratorBase<IDistributed
     IDistributedFileDirectory *parent;
     Linked<IUserDescriptor> udesc;
     Linked<IDistributedFileTransaction> transaction;
+    bool isPrivilegedUser = false;
 
     bool set()
     {
         while (isValid()) {
-            cur.setown(parent->lookup(queryName(),udesc));
+            cur.setown(parent->lookup(queryName(),udesc, false, false, false, nullptr, isPrivilegedUser));
             if (cur)
                 return true;
             index++;
@@ -2510,8 +2513,8 @@ class CDistributedFileIterator: public CDistributedFileIteratorBase<IDistributed
     }
 
 public:
-    CDistributedFileIterator(IDistributedFileDirectory *_dir,const char *wildname,bool includesuper,IUserDescriptor *user,IDistributedFileTransaction *_transaction=NULL)
-        : transaction(_transaction)
+    CDistributedFileIterator(IDistributedFileDirectory *_dir,const char *wildname,bool includesuper,IUserDescriptor *user,bool _isPrivilegedUser, IDistributedFileTransaction *_transaction=NULL)
+        : isPrivilegedUser(_isPrivilegedUser), transaction(_transaction)
     {
         setUserDescriptor(udesc,user);
         if (!wildname||!*wildname)
@@ -4688,7 +4691,7 @@ class CDistributedSuperFile: public CDistributedFileBase<IDistributedSuperFile>
 
     CDistributedFilePartArray partscache;
     FileClusterInfoArray clusterscache;
-
+    bool containsRestrictedSubfile = false;
     /**
      * Adds a sub-file to a super-file within a transaction.
      */
@@ -5121,16 +5124,17 @@ protected:
                 if (!orderedSubFiles[i])
                     ThrowStringException(-1, "CDistributedSuperFile: SuperFile %s: missing subfile part number %d of %d", logicalName.get(), i+1, n);
             }
+            containsRestrictedSubfile = false;
             // Now try to resolve them all (file/superfile)
             for (unsigned f=0; f<n; f++)
             {
                 IPropertyTree &sub = *(orderedSubFiles[f]);
                 sub.getProp("@name",subname.clear());
                 Owned<IDistributedFile> subfile;
-                subfile.setown(transaction?transaction->lookupFile(subname.str(),timeout):parent->lookup(subname.str(), udesc, false, false, false, transaction, timeout));
+                subfile.setown(transaction?transaction->lookupFile(subname.str(),timeout):parent->lookup(subname.str(), udesc, false, false, false, transaction, defaultPrivilegedUser, timeout));
                 if (!subfile.get())
                     subfile.setown(transaction?transaction->lookupSuperFile(subname.str(),timeout):parent->lookupSuperFile(subname.str(),udesc,transaction,timeout));
-
+                containsRestrictedSubfile = containsRestrictedSubfile || subfile->isRestrictedAccess();
                 // Some files are ok not to exist
                 if (!subfile.get())
                 {
@@ -6120,6 +6124,13 @@ public:
             if (numSubfiles != subFileCount)
                 throw MakeStringException(-1, "The value of @numsubfiles (%d) is not equal to the number of SubFile items (%d)!",numSubfiles, subFileCount);
         }
+    }
+    virtual bool isRestrictedAccess() override
+    {
+        // This ensures restriction applies even if superfile is unrestricted but subfiles are.
+        // However, this also means that the superfile will show as "Restricted" in ECL Watch and whenever the user tries to unset the flag
+        // it will appear to reset to Restricted.
+        return containsRestrictedSubfile;
     }
 
 private:
@@ -7618,11 +7629,11 @@ INamedGroupStore  &queryNamedGroupStore()
 
 // --------------------------------------------------------
 
-IDistributedFile *CDistributedFileDirectory::lookup(const char *_logicalname, IUserDescriptor *user, bool writeattr, bool hold, bool lockSuperOwner, IDistributedFileTransaction *transaction, unsigned timeout)
+IDistributedFile *CDistributedFileDirectory::lookup(const char *_logicalname, IUserDescriptor *user, bool writeattr, bool hold, bool lockSuperOwner, IDistributedFileTransaction *transaction, bool privilegedUser, unsigned timeout)
 {
     CDfsLogicalFileName logicalname;
     logicalname.set(_logicalname);
-    return lookup(logicalname, user, writeattr, hold, lockSuperOwner, transaction, timeout);
+    return lookup(logicalname, user, writeattr, hold, lockSuperOwner, transaction, privilegedUser, timeout);
 }
 
 IDistributedFile *CDistributedFileDirectory::dolookup(CDfsLogicalFileName &_logicalname, IUserDescriptor *user, bool writeattr, bool hold, bool lockSuperOwner, IDistributedFileTransaction *transaction, unsigned timeout)
@@ -7723,9 +7734,13 @@ IDistributedFile *CDistributedFileDirectory::dolookup(CDfsLogicalFileName &_logi
     return NULL;
 }
 
-IDistributedFile *CDistributedFileDirectory::lookup(CDfsLogicalFileName &logicalname, IUserDescriptor *user, bool writeattr, bool hold, bool lockSuperOwner, IDistributedFileTransaction *transaction, unsigned timeout)
+IDistributedFile *CDistributedFileDirectory::lookup(CDfsLogicalFileName &logicalname, IUserDescriptor *user, bool writeattr, bool hold, bool lockSuperOwner, IDistributedFileTransaction *transaction, bool privilegedUser, unsigned timeout)
 {
-    return dolookup(logicalname, user, writeattr, hold, lockSuperOwner, transaction, timeout);
+    Owned <IDistributedFile>distributedFile = dolookup(logicalname, user, writeattr, hold, lockSuperOwner, transaction, timeout);
+    // Restricted access is currently designed to stop users viewing sensitive information. It is not designed to stop users deleting or overwriting existing restricted files
+    if (writeattr==false && distributedFile && distributedFile->isRestrictedAccess() && !privilegedUser)
+        throw new CDFS_Exception(DFSERR_RestrictedFileAccessDenied,logicalname.get());
+    return distributedFile.getClear();
 }
 
 IDistributedSuperFile *CDistributedFileDirectory::lookupSuperFile(const char *_logicalname,IUserDescriptor *user,IDistributedFileTransaction *transaction, unsigned timeout)
@@ -7764,7 +7779,9 @@ bool CDistributedFileDirectory::exists(const char *_logicalname,IUserDescriptor 
     external = dlfn.isExternal();
     foreign = dlfn.isForeign();
     if (foreign) {
-        Owned<IDistributedFile> file = lookup(_logicalname, user, false, false, false, NULL, defaultTimeout);
+        // Restricted access is currently designed to stop users viewing sensitive information. Assuming privileged user rights to allow
+        // exists() operation to succeed regardless of user rights
+        Owned<IDistributedFile> file = lookup(_logicalname, user, false, false, false, NULL, defaultPrivilegedUser, defaultTimeout);
         if (file.get()==NULL)
             return false;
         if (file->querySuperFile()) {
@@ -7799,7 +7816,9 @@ bool CDistributedFileDirectory::exists(const char *_logicalname,IUserDescriptor 
 
 bool CDistributedFileDirectory::existsPhysical(const char *_logicalname, IUserDescriptor *user)
 {
-    Owned<IDistributedFile> file = lookup(_logicalname, user, false, false, false, NULL, defaultTimeout);
+    // Restricted access is currently designed to stop users viewing sensitive information. Assuming privileged user rights to allow
+    // existsPhysical() operation to succeed regardless of user rights
+    Owned<IDistributedFile> file = lookup(_logicalname, user, false, false, false, NULL, defaultPrivilegedUser, defaultTimeout);
     if (!file)
         return false;
     return file->existsPhysicalPartFiles(0);
@@ -8498,9 +8517,9 @@ void CDistributedFileDirectory::addEntry(CDfsLogicalFileName &dlfn,IPropertyTree
     }
 }
 
-IDistributedFileIterator *CDistributedFileDirectory::getIterator(const char *wildname, bool includesuper, IUserDescriptor *user)
+IDistributedFileIterator *CDistributedFileDirectory::getIterator(const char *wildname, bool includesuper, IUserDescriptor *user,bool isPrivilegedUser)
 {
-    return new CDistributedFileIterator(this,wildname,includesuper,user);
+    return new CDistributedFileIterator(this,wildname,includesuper,user,isPrivilegedUser);
 }
 
 GetFileClusterNamesType CDistributedFileDirectory::getFileClusterNames(const char *_logicalname,StringArray &out)
@@ -11177,8 +11196,8 @@ DistributedFileCompareResult CDistributedFileDirectory::fileCompare(const char *
     StringBuffer msg;
     try
     {
-        Owned<IDistributedFile> file1 = lookup(lfn1, user, false, false, false, NULL, defaultTimeout);
-        Owned<IDistributedFile> file2 = lookup(lfn2, user, false, false, false, NULL, defaultTimeout);
+        Owned<IDistributedFile> file1 = lookup(lfn1, user, false, false, false, NULL, defaultPrivilegedUser, defaultTimeout);
+        Owned<IDistributedFile> file2 = lookup(lfn2, user, false, false, false, NULL, defaultPrivilegedUser, defaultTimeout);
         if (!file1)
         {
             errstr.appendf("File %s not found",lfn1);
@@ -11325,7 +11344,7 @@ DistributedFileCompareResult CDistributedFileDirectory::fileCompare(const char *
 bool CDistributedFileDirectory::filePhysicalVerify(const char *lfn, IUserDescriptor *user, bool includecrc, StringBuffer &errstr)
 {
     bool differs = false;
-    Owned<IDistributedFile> file = lookup(lfn, user, false, false, false, NULL, defaultTimeout);
+    Owned<IDistributedFile> file = lookup(lfn, user, false, false, false, NULL, defaultPrivilegedUser, defaultTimeout);
     if (!file)
     {
         errstr.appendf("Could not find file: %s",lfn);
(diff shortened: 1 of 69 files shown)


shamser:
Note: documentation needs to be updated:https://track.hpccsystems.com/browse/HPCC-22742

shamser:
@ghalliday @jakesmith Please can you review.

dali/base/dadfs.cpp
```diff
@@ -4683,7 +4685,7 @@ class CDistributedSuperFile: public CDistributedFileBase<IDistributedSuperFile>
 
     CDistributedFilePartArray partscache;
     FileClusterInfoArray clusterscache;
-
+    bool subfileRestricted;
```
jakesmith: should initialize to false here.
It may not be possible to reference before initialized at the moment, but safer, more readable and inconsequential cost.

Generally better to initialize all members at member declaration time.

dali/base/dadfs.cpp
```diff
@@ -5116,16 +5119,19 @@ class CDistributedSuperFile: public CDistributedFileBase<IDistributedSuperFile>
                 if (!orderedSubFiles[i])
                     ThrowStringException(-1, "CDistributedSuperFile: SuperFile %s: missing subfile part number %d of %d", logicalName.get(), i+1, n);
             }
+            bool superfileRestricted = root->getPropBool("@restricted", false);
```
jakesmith: Needs changing, inlight of other PR-12940. But should perhaps either call isRestrictedAccess() for that reason, or an inline version of it.

dali/base/dadfs.cpp
```diff
@@ -5116,16 +5119,19 @@ class CDistributedSuperFile: public CDistributedFileBase<IDistributedSuperFile>
...
             {
                 IPropertyTree &sub = *(orderedSubFiles[f]);
                 sub.getProp("@name",subname.clear());
                 Owned<IDistributedFile> subfile;
-                subfile.setown(transaction?transaction->lookupFile(subname.str(),timeout):parent->lookup(subname.str(), udesc, false, false, false, transaction, timeout));
+                subfile.setown(transaction?transaction->lookupFile(subname.str(),timeout):parent->lookup(subname.str(), udesc, false, false, false, transaction, true, timeout));
                 if (!subfile.get())
                     subfile.setown(transaction?transaction->lookupSuperFile(subname.str(),timeout):parent->lookupSuperFile(subname.str(),udesc,transaction,timeout));
-
+                subfileRestricted = subfileRestricted || subfile->isRestrictedAccess();
```
jakesmith: trivial: could be:
```
subfileRestricted |= subfile->isRestrictedAccess();
```

dali/base/dadfs.cpp
```diff
@@ -6116,6 +6122,13 @@ class CDistributedSuperFile: public CDistributedFileBase<IDistributedSuperFile>
                 throw MakeStringException(-1, "The value of @numsubfiles (%d) is not equal to the number of SubFile items (%d)!",numSubfiles, subFileCount);
         }
     }
+    virtual bool isRestrictedAccess() const override
+    {
+        // This ensures restriction applies even if superfile is unrestricted but subfile are.
+        // However, this also means that the superfile will show as "Restricted" in ECL Watch and whenever the user tries to unset the flag
+        // it will appear to reset to Restricted. (TODO)
+        return subfileRestricted || root->getPropBool("@restricted", false);
```
jakesmith: Needs changing, inlight of other PR-12940.
How could subfileRestricted be false, but the getPropBool be true here?

dali/base/dadfs.cpp
```diff
@@ -7608,9 +7621,12 @@ IDistributedFile *CDistributedFileDirectory::dolookup(CDfsLogicalFileName &_logi
     return NULL;
 }
 
-IDistributedFile *CDistributedFileDirectory::lookup(CDfsLogicalFileName &logicalname, IUserDescriptor *user, bool writeattr, bool hold, bool lockSuperOwner, IDistributedFileTransaction *transaction, unsigned timeout)
+IDistributedFile *CDistributedFileDirectory::lookup(CDfsLogicalFileName &logicalname, IUserDescriptor *user, bool writeattr, bool hold, bool lockSuperOwner, IDistributedFileTransaction *transaction, bool privilegedUser, unsigned timeout)
 {
-    return dolookup(logicalname, user, writeattr, hold, lockSuperOwner, transaction, timeout);
+    Owned <IDistributedFile> distributedFile = dolookup(logicalname, user, writeattr, hold, lockSuperOwner, transaction, timeout);
+    if (writeattr==false && distributedFile && distributedFile->isRestrictedAccess() && !privilegedUser)
+        throw new CDFS_Exception(DFSERR_RestrictedFileAccessDenied,logicalname.get());
```
jakesmith: I think it's deliberate, but writing to a file that is restricted by a non-privileged user - is allowed, correct?

If so, I think should add a comment here though to explain.

dali/base/dadfs.cpp
```diff
@@ -7608,9 +7621,12 @@ IDistributedFile *CDistributedFileDirectory::dolookup(CDfsLogicalFileName &_logi
     return NULL;
 }
 
-IDistributedFile *CDistributedFileDirectory::lookup(CDfsLogicalFileName &logicalname, IUserDescriptor *user, bool writeattr, bool hold, bool lockSuperOwner, IDistributedFileTransaction *transaction, unsigned timeout)
+IDistributedFile *CDistributedFileDirectory::lookup(CDfsLogicalFileName &logicalname, IUserDescriptor *user, bool writeattr, bool hold, bool lockSuperOwner, IDistributedFileTransaction *transaction, bool privilegedUser, unsigned timeout)
 {
-    return dolookup(logicalname, user, writeattr, hold, lockSuperOwner, transaction, timeout);
+    Owned <IDistributedFile> distributedFile = dolookup(logicalname, user, writeattr, hold, lockSuperOwner, transaction, timeout);
```
jakesmith: trivial/formatting: don't normally have a space after template name.

dali/base/dadfs.cpp
```diff
@@ -7684,7 +7701,7 @@ bool CDistributedFileDirectory::exists(const char *_logicalname,IUserDescriptor
 
 bool CDistributedFileDirectory::existsPhysical(const char *_logicalname, IUserDescriptor *user)
 {
-    Owned<IDistributedFile> file = lookup(_logicalname, user, false, false, false, NULL, defaultTimeout);
+    Owned<IDistributedFile> file = lookup(_logicalname, user, false, false, false, NULL, true, defaultTimeout);
```
jakesmith: is privilegedUser=true always correct here, or is this is a stop gab, before tightening up later?

dali/daliadmin/daliadmin.cpp
```diff
@@ -590,7 +590,7 @@ static void setdfspartattr(const char *lname, unsigned partNum, const char *attr
         throw MakeStringException(0, "External file not supported");
     if (lfn.isForeign()) 
         throw MakeStringException(0, "Foreign file not supported");
-    Owned<IDistributedFile> file = queryDistributedFileDirectory().lookup(lname, userDesc);
+    Owned<IDistributedFile> file = queryDistributedFileDirectory().lookup(lname, userDesc, false, false, false, nullptr, true);
     if (nullptr == file.get())
```
jakesmith: normally just:
```
if (!file)
```

(as is the case elsewhere in this file)

ecl/hthor/hthor.cpp
```diff
@@ -10436,11 +10441,15 @@ CHThorNewDiskReadBaseActivity::CHThorNewDiskReadBaseActivity(IAgentContext &_age
...
     projectedDiskMeta = helper.queryProjectedDiskRecordSize();
     readerOptions.setown(createPTree());
+    isCodeSigned = false;
     if (_node)
     {
         const char *recordTranslationModeHintText = _node->queryProp("hint[@name='layoutTranslation']/@value");
         if (recordTranslationModeHintText)
             recordTranslationModeHint = getTranslationMode(recordTranslationModeHintText);
+        const char *signedBy =_node->queryProp("signedBy");
+        if (signedBy && *signedBy!='\0')
```
jakesmith: clearer to use !isEmptyString(signedBy)

ecl/hthor/hthor.cpp
```diff
@@ -158,14 +158,16 @@ const void * CRowBuffer::next()
...
+ILocalOrDistributedFile *resolveLFNFlat(IAgentContext &agent, const char *logicalName, const char *errorTxt, bool optional, bool privilegedUser)
 {
     Owned<ILocalOrDistributedFile> ldFile = agent.resolveLFN(logicalName, errorTxt, optional);
     if (!ldFile)
         return nullptr;
     IDistributedFile *dFile = ldFile->queryDistributedFile();
     if (dFile && isFileKey(dFile))
         throw MakeStringException(0, "Attempting to read index as a flat file: %s", logicalName);
+    if (dFile->isRestrictedAccess() && !privilegedUser)
+        throw MakeStringException(0, "Access denied to restricted file: %s", logicalName);
```
jakesmith: why is this checking here? It should already have been checked and fired exception if necessary in DFS lookup shouldn't it?

jakesmith: also dFile could be null here.

testing/unittests/dalitests.cpp
```diff
@@ -1731,9 +1731,9 @@ class CDaliDFSStressTests : public CppUnit::TestFixture
             ASSERT(strcmp(sfile3->querySubFile(0).queryLogicalName(), "regress::trans::sub3") == 0 && "promote failed, wrong name for sub3");
             ASSERT(outlinked.length() == 1 && "promote failed, outlinked expected only one item");
             ASSERT(strcmp(outlinked.popGet(), "regress::trans::sub2") == 0 && "promote failed, outlinked expected to be sub2");
-            Owned<IDistributedFile> sub1 = dir.lookup("regress::trans::sub1", user, false, false, false, NULL, timeout);
+            Owned<IDistributedFile> sub1 = dir.lookup("regress::trans::sub1", user, false, false, false, NULL, false, timeout);
             ASSERT(sub1.get() && "promote failed, sub1 was physically deleted");
-            Owned<IDistributedFile> sub2 = dir.lookup("regress::trans::sub2", user, false, false, false, NULL, timeout);
+            Owned<IDistributedFile> sub2 = dir.lookup("regress::trans::sub2", user, false, false, false, NULL, false, timeout);
             ASSERT(sub2.get() && "promote failed, sub2 was physically deleted");
```
jakesmith: Can you add restricted access tests somewhere here (in dalitests.cpp) ?

thorlcr/graph/thgraph.cpp
```diff
@@ -335,7 +335,10 @@ CGraphElementBase::CGraphElementBase(CGraphBase &_owner, IPropertyTree &_xgmml)
     ownerId = xgmml->getPropInt("att[@name=\"_parentActivity\"]/@value", 0);
     onCreateCalled = prepared = haveCreateCtx = nullAct = false;
     onlyUpdateIfChanged = xgmml->getPropBool("att[@name=\"_updateIfChanged\"]/@value", false);
-
+    isCodeSigned = false;
+    const char *signedBy = xgmml->queryProp("att[@name=\"signedBy\"]/@value");
+    if (signedBy && *signedBy!='\0')
```
jakesmith: trivial: clearer to use:
```
if (!isEmptyString(signedBy))
```

thorlcr/graph/thgraph.hpp
```diff
@@ -372,6 +373,7 @@ class graph_decl CGraphElementBase : public CInterface, implements IInterface
     virtual CActivityBase *factory() { return factory(getKind()); }
     virtual CActivityBase *factorySet(ThorActivityKind kind) { CActivityBase *_activity = factory(kind); activity.setown(_activity); return _activity; }
     virtual ICodeContext *queryCodeContext();
+    virtual bool queryIsCodeSigned() { return isCodeSigned; }
```
jakesmith: could/should be const method

ghalliday:
JIRA: https://track.hpccsystems.com/browse/HPCC-22682

thorlcr/activities/thdiskbase.cpp
```diff
@@ -46,7 +46,7 @@ void CDiskReadMasterBase::init()
     fileName.set(expandedFileName);
     reInit = 0 != (helper->getFlags() & (TDXvarfilename|TDXdynamicfilename));
 
-    Owned<IDistributedFile> file = queryThorFileManager().lookup(container.queryJob(), helperFileName, 0 != ((TDXtemporary|TDXjobtemp) & helper->getFlags()), 0 != (TDRoptional & helper->getFlags()), true);
+    Owned<IDistributedFile> file = queryThorFileManager().lookup(container.queryJob(), helperFileName, 0 != ((TDXtemporary|TDXjobtemp) & helper->getFlags()), 0 != (TDRoptional & helper->getFlags()), true, container.queryIsCodeSigned());
```
jakesmith: what about the other activities that read?

jakesmith (changes requested):

jakesmith:
@shamser - have done 1st review. 
it looks like there's a number of places that should be conditionally passing privilegeUser that aren't at the moment (e.g. in Thor).
But also there are a few places that are hard-wired to privilegeUser=true afaics, e.g. code in dfs in the transaction code. It would mean I think that users can manipulate restricted files to some extent, from non-signed code.

Also, I know there was an offline conversation about it (https://www.flowdock.com/app/richard-chapman/design-descussions/threads/qTzeF2kStrK4Oldhw2NEI3hKHQp) to some extent,
but as it stands, I believe it's still possible for non-signed code to read signed code indirectly by calling esp services?

dali/base/dadfs.cpp
shamser: Isn't that bitwise 'or' rather than logical 'or'?

jakesmith: it is indeed, I was just testing :)

ghalliday (commented):
Generally looks ok.  

Main global comment is that you should distinguish the cases where you should definitely pass true for privileged user, and others where it is a default.  Should possibly have multiple symbolic defaults - e.g. espIsPrivilegedUser, defaultIsPrivilegedUser.  We will need to revisit what happens when trying to view these files from esp, and it would greatly help finding those cases later.

Slightly picky, the name privilegedUser suggests to me that it could be a username, rather than isPrivilegedUser, which is obviously a boolean.  However the same goes for all the existing boolean parameters, so probably best to leave as it is.

ecl/hthor/hthorkey.cpp
```diff
@@ -2136,11 +2136,18 @@ class CHThorThreadedActivityBase : public CHThorActivityBase, implements IThread
...
-    CHThorThreadedActivityBase (IAgentContext &_agent, unsigned _activityId, unsigned _subgraphId, IHThorArg &_arg, IHThorFetchContext &_fetch, ThorActivityKind _kind, IRecordSize *diskSize)
+    CHThorThreadedActivityBase (IAgentContext &_agent, unsigned _activityId, unsigned _subgraphId, IHThorArg &_arg, IHThorFetchContext &_fetch, ThorActivityKind _kind, IRecordSize *diskSize, IPropertyTree *_node)
         : CHThorActivityBase(_agent, _activityId, _subgraphId, _arg, _kind), fetch(_fetch)
     {
         exception = NULL;
         rowLimit = 0;
+        if (_node)
+        {
+            const char *signedBy =_node->queryProp("signedBy");
+            bool isCodeSigned = false;
```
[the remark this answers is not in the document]
jakesmith: yes, I spotted that - but forgot to double check and comment.

dali/base/dadfs.cpp
```diff
@@ -1667,7 +1669,7 @@ class CDistributedFileTransaction: implements IDistributedFileTransactionExt, pu
             return LINK(ret);
         else
         {
-            ret = queryDistributedFileDirectory().lookup(name, udesc, false, false, false, this, timeout);
+            ret = queryDistributedFileDirectory().lookup(name, udesc, false, false, false, this, true, timeout);
```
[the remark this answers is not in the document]
shamser: Is there any functionlity in CDistributedFileTransaction to read the superfiles?  It looks like you could add, remove and delete  subfiles but nothing to read the contents.. is that right?

dali/base/dadfs.cpp
```diff
@@ -2496,7 +2498,7 @@ class CDistributedFileIterator: public CDistributedFileIteratorBase<IDistributed
     bool set()
     {
         while (isValid()) {
-            cur.setown(parent->lookup(queryName(),udesc));
+            cur.setown(parent->lookup(queryName(),udesc, false, false, false, nullptr, true));
```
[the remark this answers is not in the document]
shamser: CDistributedFileIterator comment shows that it is used to list files in directory -nothing there to read contents.  Should be ok?

ecl/hthor/hthor.ipp
```diff
@@ -2254,6 +2254,7 @@ protected:
     MemoryAttr encryptionkey;
     bool persistent;
     bool grouped;
+    bool isCodeSigned;
```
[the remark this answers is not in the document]
shamser: Didn't set default as none of the other member variables has defaults.

dali/base/dadfs.cpp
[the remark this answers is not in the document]
jakesmith: it's not just superfiles, it's potentially any logical file involved (looked up) within the context of a super file transaction  (between a StartSuperFileTransaction() and FinishSuperFileTransaction()).
But it looks like at the moment, nothing that uses a transaction, goes on to call IDistributedFileTransaction::lookupFile() or IDistributedFileTransaction::lookupSuperFile() - and proceed to read it's content.
But that is how it is supposed to be used, so code could.

dali/base/dadfs.cpp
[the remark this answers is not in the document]
jakesmith: the iterator returns IDistributeFile's so any code that uses a IDistributedFileIterator - could use one and get a IDistributeFile and read content.

Whether there's any paths to that, through ECL code, I don't know, but doesn't feel like it should always default to give unsigned access.
e.g. if any engines uses one of these iterators to get files, it will bypass the signedness check.

The solution may be to pass in the isprivileged flag when creating one of these and ensure all uses pass in whether they are trusted or not.

dali/base/dadfs.cpp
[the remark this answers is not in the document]
shamser: We can revisit this later: https://track.hpccsystems.com/browse/HPCC-22759

dali/daliadmin/daliadmin.cpp
```diff
@@ -558,7 +558,7 @@ static void dfsfile(const char *lname,IUserDescriptor *userDesc, UnsignedArray *
         outln(str.str());
     }
     else {
-        Owned<IDistributedFile> file = queryDistributedFileDirectory().lookup(lname,userDesc);
+        Owned<IDistributedFile> file = queryDistributedFileDirectory().lookup(lname,userDesc,false,false,false,nullptr,true);
```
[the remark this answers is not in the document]
shamser: Daliadmin access to restricted probably needs to be more sophisticated. I have created a jira to discuss this and re-design in the future: https://track.hpccsystems.com/browse/HPCC-22765.  For the first release, we can leave as is.

dali/daliadmin/daliadmin.cpp
shamser: Unrelated change but I'll go ahead and change.

dali/base/dadfs.cpp
shamser: I have already defined isRestrictedAccess to return true if subfile is restricted or superfile is restricted.  I'll modify inlight of PR-12940

dali/base/dadfs.cpp
shamser: Yes, it can be and yes they'd probably be able access the subfile directly if that was the case.  Should we make it so that whenever the superfile is marked as restricted, all subfiles are marked as restricted?

dali/base/dadfs.cpp
shamser: Comment added

dali/base/dadfs.cpp
```diff
@@ -7649,7 +7665,8 @@ bool CDistributedFileDirectory::exists(const char *_logicalname,IUserDescriptor
     external = dlfn.isExternal();
     foreign = dlfn.isForeign();
     if (foreign) {
-        Owned<IDistributedFile> file = lookup(_logicalname, user, false, false, false, NULL, defaultTimeout);
+        // SATODO:
```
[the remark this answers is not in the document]
shamser: Comment added to code.

dali/base/dadfs.cpp
shamser: Yes it's always true because just checking if it exists shouldn't cause an sensitive information to be read.  Add comment in code about this.

dali/base/dadfs.cpp
jakesmith: If a super file is not restricted, but all subs are, then it is fully restricted (from reading any content) anyway.
If super is restricted, but some subs are not, then they can always bypass the super and go to the subs - so don't think restricting the super adds anything, *unless* reading a subfile (looking it up in DFS), involved seeing if was owned by a super that was restricted. Which is kind of similar to trying to delete a subfile that is owned by a super (you get an error).

I would not do anything special for supers for now.

dali/base/dadfs.cpp
```diff
@@ -11007,8 +11024,9 @@ DistributedFileCompareResult CDistributedFileDirectory::fileCompare(const char *
     StringBuffer msg;
     try
     {
-        Owned<IDistributedFile> file1 = lookup(lfn1, user, false, false, false, NULL, defaultTimeout);
-        Owned<IDistributedFile> file2 = lookup(lfn2, user, false, false, false, NULL, defaultTimeout);
+        // SATODO:
```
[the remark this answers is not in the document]
shamser: Here's a jira to revisit this: https://track.hpccsystems.com/browse/HPCC-22759

dali/base/dadfs.cpp
```diff
@@ -12702,7 +12720,7 @@ bool CDistributedFileDirectory::removePhysicalPartFiles(const char *logicalName,
  */
 extern da_decl void removeLogical(const char *fname, IUserDescriptor *user) {
     if (queryDistributedFileDirectory().exists(fname, user)) {
-        Owned<IDistributedFile> file = queryDistributedFileDirectory().lookup(fname, user, true);
+        Owned<IDistributedFile> file = queryDistributedFileDirectory().lookup(fname, user, true, false, false, nullptr, true);
```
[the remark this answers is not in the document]
shamser: Yes and also this is only used in unit tests.

testing/unittests/dalitests.cpp
shamser: I was thinking of creating regression queries in testing/regress so that the handling of the restricted files by the different activities may be checked - both by signed and unsigned queries.  I'm not sure how useful calling lookup with PrivilegedUser=true and PrivilegedUser=false would be

shamser:
@jakesmith @richardkchapman Please can you review. There's Roxie and Thor changes. @ghalliday has started reviewing

roxie/ccd/ccdserver.cpp
```diff
@@ -418,7 +418,7 @@ class CRoxieServerActivityFactoryBase : public CActivityFactory, implements IRox
     bool optUnstableInput = false;  // is the input forced to unordered?
     bool optUnordered = false; // is the output specified as unordered?
     unsigned heapFlags;
-
+    bool isCodeSigned = false;
```
ghalliday: minor: moving this before heapFlags would reduce memory consumption by 8 bytes (since it would pack in with the other booleans)

roxie/ccd/ccdserver.cpp
```diff
@@ -21841,6 +21843,7 @@ class CRoxieServerDiskReadBaseActivity : public CRoxieServerActivity, implements
     bool isLocal;
     bool forceRemote;
     bool isGrouped = false;
+    bool isCodeSigned = false;
```
ghalliday: unfortunately this pushes up the size of all disk read activities by 8 bytes.

@richardkchapman would this be better to not store here, and use factory->isCodeSigned() instead?

ghalliday (commented):
@shamser a few relatively minor comments.  I was also going to comment that some test cases would be good.

dali/base/dadfs.cpp
```diff
@@ -5121,16 +5124,20 @@ class CDistributedSuperFile: public CDistributedFileBase<IDistributedSuperFile>
...
                 sub.getProp("@name",subname.clear());
                 Owned<IDistributedFile> subfile;
-                subfile.setown(transaction?transaction->lookupFile(subname.str(),timeout):parent->lookup(subname.str(), udesc, false, false, false, transaction, timeout));
+                subfile.setown(transaction?transaction->lookupFile(subname.str(),timeout):parent->lookup(subname.str(), udesc, false, false, false, transaction, defaultPrivilegedUser, timeout));
                 if (!subfile.get())
                     subfile.setown(transaction?transaction->lookupSuperFile(subname.str(),timeout):parent->lookupSuperFile(subname.str(),udesc,transaction,timeout));
-
+                containsRestrictedSubfile = containsRestrictedSubfile || subfile->isRestrictedAccess();
+                if (subfile->isRestrictedAccess() != isRestrictedSuperfile)
+                    UWARNLOG("CDistributedSuperFile: SuperFile '%s' and subfile '%s' have different restricted flag", logicalName.get(), subname.str());
```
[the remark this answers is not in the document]
shamser: There are 2 potential issues
1) if the superfile is marked as restricted but subfiles are not, it would be possible to create a new superfile to access the data.
2) if the subfiles are restricted but superfile isn't, it would not be possible to unrestrict the superfile with the ECLWatch
The warning is to let the users know so they are can address the issue if they wish. It won't cause any other issues.  I've commented in the jira with some questions regarding how superfiles and subfile restrictions should be handled.

ecl/hthor/hthorkey.cpp
```diff
@@ -1087,9 +1087,16 @@ ILocalOrDistributedFile *resolveLFNIndex(IAgentContext &agent, const char *logic
 
 extern HTHOR_API IHThorActivity *createIndexReadActivity(IAgentContext &_agent, unsigned _activityId, unsigned _subgraphId, IHThorIndexReadArg &arg, ThorActivityKind _kind, IPropertyTree *_node)
 {
+    bool isCodeSigned = false;
```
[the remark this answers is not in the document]
shamser: I've created a Jira for the filename being variable issue:https://track.hpccsystems.com/browse/HPCC-22797

roxie/ccd/ccdstate.hpp
```diff
@@ -171,4 +171,11 @@ extern void stopDelayedReleaser();
 extern void createDelayedReleaser();
 extern void stopDelayedReleaser();
 
+
+inline bool isActivityCodeSigned(IPropertyTree &graphNode)
+{
+    if (!isEmptyString(graphNode.queryProp("att[@name=\"signedBy\"]/@value")))
```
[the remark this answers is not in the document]
shamser: We'd need to examine if multiple signers are needed and, if so, how it may be supported.  Current version does not support multiple signers.

roxie/ccd/ccdcontext.cpp
```diff
@@ -2487,22 +2487,22 @@ class CSlaveContext : public CRoxieContextBase
         return NULL;
     }
 
-    virtual const IResolvedFile *resolveLFN(const char *filename, bool isOpt)
+    virtual const IResolvedFile *resolveLFN(const char *filename, bool isOpt, bool isPrivilegedUser)
     {
         CDateTime cacheDate; // Note - this is empty meaning we don't know...
         return querySlaveDynamicFileCache()->lookupDynamicFile(*this, filename, cacheDate, 0, header, isOpt, false);
```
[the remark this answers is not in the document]
shamser: I've create a Jira for this: https://track.hpccsystems.com/browse/HPCC-22798

roxie/ccd/ccdserver.cpp
richardkchapman: I suspect there are quite a few flags that would be better off stored in the factory rather than being copied into the activity (especially as we seem to always pass as parameters to the constructor too).

shamser: I'm using the flag from the factory.  It's made the code some much simpler.

shamser:
@ghalliday I've addressed the comments.  Final review?

dali/base/dadfs.cpp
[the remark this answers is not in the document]
jakesmith: My comment from JIRA:
```
My feeling is that it should allow mixed, but reading a super that has any restricted subs should only be possible if authorized.
In that scheme, there would be no separate super restricted property, it would be derived at super load time, or when you add / remove a sub.
```

dali/base/dadfs.cpp
```diff
@@ -6121,6 +6128,13 @@ class CDistributedSuperFile: public CDistributedFileBase<IDistributedSuperFile>
                 throw MakeStringException(-1, "The value of @numsubfiles (%d) is not equal to the number of SubFile items (%d)!",numSubfiles, subFileCount);
         }
     }
+    virtual bool isRestrictedAccess() override
```
jakesmith: minor: could be const

shamser: Unfortunately, I had to remove const from this function as it uses queryAttributes

dali/base/dadfs.cpp
```diff
@@ -6121,6 +6128,13 @@ class CDistributedSuperFile: public CDistributedFileBase<IDistributedSuperFile>
...
+    {
+        // This ensures restriction applies even if superfile is unrestricted but subfiles are.
+        // However, this also means that the superfile will show as "Restricted" in ECL Watch and whenever the user tries to unset the flag
+        // it will appear to reset to Restricted.
+        return containsRestrictedSubfile || queryAttributes().getPropBool("@restricted");
```
jakesmith: It might get confusing if a subfile implies super is restricted, but also the super has a restricted attribute of it's own..
And if super has restricted flag but subs don't there's nothing to stop the user looking at the super (if that doesn't count as reading?) and then reading the unprotected subs directly.

I think I wouldn't have a separate super restricted attribute.

ecl/eclagent/eclagent.cpp
```diff
@@ -1456,7 +1456,7 @@ bool EclAgent::fileExists(const char *name)
     StringBuffer lfn;
     expandLogicalName(lfn, name);
 
-    Owned<IDistributedFile> f = queryDistributedFileDirectory().lookup(lfn.str(),queryUserDescriptor());
+    Owned<IDistributedFile> f = queryDistributedFileDirectory().lookup(lfn.str(),queryUserDescriptor(), false, false, false, nullptr, defaultPrivilegedUser);
```
jakesmith: I know this is not reading content, but is there any benefits from passing defaultPrivilegedUser here, when isPrivilegedUser is available?

jakesmith: actually forget that, fileExists() is only used internally.

dali/base/dadfs.cpp
jakesmith: right yes annoying.

jakesmith:
@shamser - added a few comments.

dali/base/dadfs.cpp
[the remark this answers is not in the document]
ghalliday: I think I agree with Jake, but it can be revisited once this PR is merged

ghalliday:
@richardkchapman I think ready to merge.  Do we need to change the target branch to 7.6.0?

PR #12933, opened 2019-09-04
HPCC-18819 Add drop zone(s) discovery/listing features to fileservices for ECL programmers.

AttilaVamos:
Implement GetDefaultDropZone() and GetDropZones() functions into Std.File

Update spray_test.ecl to use GetDefaultDropZone() function to eliminate hardwired path usage.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

## Testing:

Tested manually.


AttilaVamos:
@ghalliday @jakesmith  please review. There are two parameters: username and password, but not used in the implementation yet. Please advise me they are really needed and if yes, how suppose to be used them.

testing/regress/ecl/dropzone_query_test.ecl
```diff
@@ -0,0 +1,7 @@
+//nokey
+
+import Std.File AS FileServices;
+
+output(FileServices.GetDefaultDropZone());
+
+output(FileServices.GetDropZones());
```
jakesmith: There doesn't appear to be an accompanying key file (dropzone_query_test.xml) ?
But if that's the case, why isn't the PR failing?

plugins/fileservices/fileservices.cpp
```diff
@@ -2644,4 +2648,33 @@ FILESERVICES_API char * FILESERVICES_CALL fsGetEspURL(const char *username, cons
...
+{
+    Owned<IEnvironmentFactory> envFactory = getEnvironmentFactory(true);
+    Owned<IConstEnvironment> constEnv = envFactory-> openEnvironment();
+    Owned<IConstDropZoneInfoIterator> dropZoneIt = constEnv->getDropZoneIterator();
+    SCMStringBuffer dropZoneDir;
+    if (dropZoneIt->first())
+        dropZoneIt->query().getDirectory(dropZoneDir);
+
+    return strdup(dropZoneDir.str());
+
```
jakesmith: trivial: spurious extra new line

plugins/fileservices/fileservices.cpp
```diff
@@ -2644,4 +2648,33 @@ FILESERVICES_API char * FILESERVICES_CALL fsGetEspURL(const char *username, cons
     return strdup("");
 }
 
+FILESERVICES_API char * FILESERVICES_CALL fsGetDefaultDropZone(const char *username, const char *userPW)
+{
+    Owned<IEnvironmentFactory> envFactory = getEnvironmentFactory(true);
+    Owned<IConstEnvironment> constEnv = envFactory-> openEnvironment();
+    Owned<IConstDropZoneInfoIterator> dropZoneIt = constEnv->getDropZoneIterator();
```
jakesmith: Is it documented (or JIRA to document) that default dropzone means 1st one listed in environment?

Might be worth making default explicit, e.g. an attribute on the default DropZone, so don't need to reorder.

plugins/fileservices/fileservices.cpp
```diff
@@ -2644,4 +2648,33 @@ FILESERVICES_API char * FILESERVICES_CALL fsGetEspURL(const char *username, cons
...
+    MemoryBuffer mb;
+    Owned<IEnvironmentFactory> envFactory = getEnvironmentFactory(true);
+    Owned<IConstEnvironment> constEnv = envFactory-> openEnvironment();
+    Owned<IConstDropZoneInfoIterator> dropZoneIt = constEnv->getDropZoneIterator();
+    ForEach(*dropZoneIt)
+    {
+        SCMStringBuffer dropZoneDir;
+        dropZoneIt->query().getDirectory(dropZoneDir);
+        size32_t sz = dropZoneDir.length();
+        mb.append(sz).append(sz,dropZoneDir.str());
```
jakesmith: isn't this concatenating them altogether without separators, such that the caller wouldn't be able to tell where new one begins?

jakesmith: it may be better to return as a dataset, 1 row per dropzone

jakesmith:
>There are two parameters: username and password, but not used in the implementation yet. Please advise me they are really needed and if yes, how suppose to be used them.

No, I think should be removed. The connection for this meta info. is from Dali at the moment, and it can't authorize, aside from the whitelisting protection.
The Dropzones should be protected from actually being used by LDAP if necessary.

jakesmith:
@AttilaVamos - added a few comments

plugins/fileservices/fileservices.cpp
AttilaVamos: I can't recall it is documented somewhere. @JamesDeFabia do you know?
Ok, but I think it can be another JIRA.

plugins/fileservices/fileservices.cpp
AttilaVamos: There is a "separator", because the mb content is : <len1><record1><len2><record2>...
@ghalliday Is this the right format a dataset returned by a File Service function?

jakesmith: Sorry I misread, it is returning a dataset(FsDropZoneRecord).
Does it output a sensible result when there are multiple DropZones defined?

I think format is ok, but will let @ghalliday confirm.

plugins/fileservices/fileservices.cpp
JamesDeFabia: I don't think the concept of a default drop zone has ever been mentioned in docs. I agree with Jake that it would be better to explicitly declare one to be the default rather than letting the first one be the default.

AttilaVamos:
@jakesmith username and psw are removed. Please re-review.

plugins/fileservices/fileservices.cpp
```diff
@@ -60,6 +60,8 @@ static const char * EclDefinition =
 "export integer4 RECFMVB_RECSIZE := -1; // special value for SprayFixed record size \n"
 "export integer4 PREFIX_VARIABLE_RECSIZE := -3; // special value for SprayFixed record size \n"
 "export integer4 PREFIX_VARIABLE_BIGENDIAN_RECSIZE := -4; // special value for SprayFixed record size \n"
+"export FsDropZone := string{maxlength(255)}; \n"
```
jakesmith: is the maxlength necessary? ... and a path _could_ be longer

plugins/fileservices/fileservices.cpp
jakesmith: Agreed as a sensible, but not for this JIRA/PR.

jakesmith:
@AttilaVamos - looks fine, one minor comment/question re. reason for maxlength.

plugins/fileservices/fileservices.cpp
AttilaVamos: I saw this value at the fsLogicalFileName and I thought the drop zone path has some similarity with logical file name. So that's why I use it.

jakesmith: default linux filename length is 255.
But this is a path, and the default is 4096, so 255 may well be too small  in some cases, so I think it should be changed.

However, is it  (and the other uses of maxlength in fileservices) actually needs/helpful at all. @ghalliday ?

ghalliday: I would remove it (and just use a type of string in the record - it was needed when the original code was written, but maxlength has not been needed for many years.

plugins/fileservices/fileservices.cpp
ghalliday: The code and format both look sensible.

ecllibrary/std/File.ecl
```diff
@@ -958,5 +962,31 @@ EXPORT PromoteSuperFileList(set of varstring superNames, varstring addHead='', b
...
 EXPORT varstring GetEspURL(const varstring username = '', const varstring userPW = '') :=
     lib_fileservices.FileServices.GetEspURL(username, userPW);
+    
+    
+ /**
+ * Returns the full URL to the default Drop Zone
+ *
+ *
+ * @return              A string containing the path to the default DropZone. 
+ *						If more than one Drop Zone
```
ghalliday: tabs used for indentation causing problems.

ecllibrary/std/File.ecl
```diff
@@ -958,5 +962,31 @@ EXPORT PromoteSuperFileList(set of varstring superNames, varstring addHead='', b
...
+ */
+EXPORT varstring GetDefaultDropZone() :=
+    lib_fileservices.FileServices.GetDefaultDropZone();
+
+ /**
+ * Returns the a dataset with full URL to all Drop Zones
+ *
+ *
+ * @return              A dataset containing all defined Drop Zone paths. 
+ *						Will return an empty dataset if a Drop Zone
```
ghalliday: tabs again causing problems.

ghalliday (commented):
@AttilaVamos looks very close to ready.  A few final comments.

AttilaVamos:
@jakesmith updated, please re-review.

ecllibrary/std/File.ecl
```diff
@@ -959,4 +963,30 @@ EXPORT PromoteSuperFileList(set of varstring superNames, varstring addHead='', b
...
+ *                      If more than one Drop Zone
+ *                      process is defined then the first found will
+ *                      be returned; will return an empty string if a Drop Zone
+ *                      cannot be found
+ */
+EXPORT varstring GetDefaultDropZone() :=
+    lib_fileservices.FileServices.GetDefaultDropZone();
+
+ /**
+ * Returns the a dataset with full path to all Drop Zones
```
jakesmith: >Returns the a dataset with full path to all Drop Zones

Returns a dataset with full paths to all Drop Zones

jakesmith:
@AttilaVamos - looks good, one v. minor typo issue in comment.
Can you correct, and squash the commits?

richardkchapman:
@AttilaVamos please retarget to 7.6.0

AttilaVamos:
Close because there is new, re-targeted version: https://github.com/hpcc-systems/HPCC-Platform/pull/13006

PR #12934, opened 2019-09-04
HPCC-21462 Start refactoring disk read code

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/common/thorhelper/thorread.cpp
+++ b/common/thorhelper/thorread.cpp
@@ -25,6 +25,7 @@
 #include "rtlfield.hpp"
 #include "rtlds_imp.hpp"
 #include "rtldynfield.hpp"
+#include "roxiemem.hpp"
 
 #include "rmtclient.hpp"
 #include "rmtfile.hpp"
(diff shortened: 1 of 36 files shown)


common/thorhelper/thorxmlwrite.cpp
```diff
@@ -149,3 +150,147 @@ extern thorhelper_decl void convertRowToJSON(size32_t & lenResult, char * & resu
...
+    //Horrible that there is no interface for adding a (len, data) string
+    StringBuffer temp;
+    temp.append(size, value);
+    root->addProp(fieldname, temp.str());
+}
+
+
+void PropertyTreeXmlWriter::outputInlineXml(const char *text)
+{
+    UNIMPLEMENTED;
```
richardkchapman: Are these going to be implemented later (in which case is there a Jira), or should they be "throwUnexpected()" ?

ecl/hqlcpp/hqlsource.cpp
```diff
@@ -3274,8 +3362,10 @@ ABoundActivity * HqlCppTranslator::doBuildActivityDiskRead(BuildCtx & ctx, IHqlE
     if (isPiped)
         return info.buildActivity(ctx, expr, TAKpiperead, "PipeRead", NULL);
     ensureDiskAccessAllowed(tableExpr);
-    if (modeOp == no_csv)
+    if ((modeOp == no_csv) && !info.genericDiskRead)
         return info.buildActivity(ctx, expr, TAKcsvread, "CsvRead", NULL);
+    if (info.genericDiskRead)
```
richardkchapman: Minor, but feels like code would be clearer if these two lines were moved above the modeOp==no_csv test

richardkchapman:
@jakesmith Please review

ghalliday:
Pushed another commit responding to Richard's comments, and fixing problems when the option was enabled (caused by last phase of refactoring).  Disabled the test covered by bug HPCC-22629.

AttilaVamos:
The problem was this:
```
CMake Error at cmake/Modules/DownloadLcbDep.cmake:18 (MESSAGE):
  Failed to download
  https://packages.couchbase.com/clients/c/mock/CouchbaseMock-1.5.19.jar
Call Stack (most recent call first):
  tests/CMakeLists.txt:70 (DOWNLOAD_LCB_DEP)
```
This PR test rescheduled.

common/thorhelper/thorread.hpp
```diff
@@ -27,19 +27,56 @@
 #include "jrowstream.hpp"
 #include "rtlkey.hpp"
 
+//The following is constant for the life of a disk read activity
+interface IDiskReadOutputMapping : public IInterface
```
jakesmith: style question: I use 'extends' in this context. We should be consistent, what what is preferred?

common/thorhelper/thorxmlwrite.cpp
```diff
@@ -149,3 +150,147 @@ extern thorhelper_decl void convertRowToJSON(size32_t & lenResult, char * & resu
...
+
+void PropertyTreeXmlWriter::outputLiteralString(size32_t size, const char *value, const char *fieldname)
+{
+    //Horrible that there is no interface for adding a (len, data) string
+    StringBuffer temp;
+    temp.append(size, value);
+    root->addProp(fieldname, temp.str());
+}
+
+
```
jakesmith: trivial: extra newline

common/thorhelper/thorread.hpp
richardkchapman: I prefer to avoid non-standard things like extends - they just make the code less accessible for new readers.

ghalliday: yes, extends would be better

common/thorhelper/thorread.cpp
```diff
@@ -25,33 +25,184 @@
...
+        return mode == other->queryTranslationMode() && streq(format, other->queryFormat()) &&
+                ((actualCrc && actualCrc == other->getActualCrc()) || (actualMeta == other->queryActualMeta())) &&
+                ((expectedCrc && expectedCrc == other->getExpectedCrc()) || (expectedMeta == other->queryExpectedMeta())) &&
+                ((projectedCrc && projectedCrc == other->getProjectedCrc()) || (projectedMeta == other->queryProjectedMeta())) &&
+                areMatchingPTrees(options, other->queryOptions());
+    }
+
+    virtual bool expectedMatchesProjected() const
+    {
+        return (expectedCrc && expectedCrc == projectedCrc) || (expectedMeta == projectedMeta);
```
jakesmith: operator precedence ok, but I think clearer with braces:
```
  (expectedCrc && (expectedCrc == projectedCrc)) || (expectedMeta == projectedMeta
```

common/thorhelper/thorread.cpp
```diff
@@ -25,33 +25,184 @@
...
+    mutable Owned<const IKeyTranslator> keyedTranslator;
+    mutable SpinLock translatorLock; // use a spin lock since almost certainly not going to contend
+};
+
+void DiskReadMapping::ensureTranslators() const
+{
+    if (checkedTranslators.load())
+        return;
+    SpinBlock block(translatorLock);
+    if (checkedTranslators.load())
```
jakesmith: am curious why use load() here and above?, and not just:
```
if (checkedTranslators)
```

common/thorhelper/thorread.cpp
```diff
@@ -25,33 +25,184 @@
...
+}
+
 
 //---------------------------------------------------------------------------------------------------------------------
 
 constexpr size32_t defaultReadBufferSize = 0x10000;
 
-class DiskRowReader : extends CInterfaceOf<IAllocRowStream>, implements IRawRowStream, implements IDiskRowReader, implements IThorDiskCallback
+/*
+ * The base class for reading rows from an external file.  Each activity will have an instance of a disk reader for
```
jakesmith: super picky: 2 spaces before 'Each'

common/thorhelper/thorread.cpp
```diff
@@ -25,33 +25,184 @@
...
+ * each actual file format.
+ */
+class DiskRowReader : extends CInterfaceOf<IDiskRowStream>, implements IDiskRowReader, implements IThorDiskCallback
```
jakesmith: style: similar to other q. - I have always thought 'extends' was something we introduced to be coupled with 'interface', in this context I'd use public .. ?

common/thorhelper/thorread.cpp
```diff
@@ -61,21 +212,18 @@ class DiskRowReader : extends CInterfaceOf<IAllocRowStream>, implements IRawRowS
 
 
 protected:
-    offset_t getLocalOffset();
+    virtual offset_t getLocalOffset();
```
jakesmith: can/should this be a const method?

common/thorhelper/thorread.cpp
```diff
@@ -61,21 +212,18 @@ class DiskRowReader : extends CInterfaceOf<IAllocRowStream>, implements IRawRowS
...
-    Owned<const IKeyTranslator> keyedTranslator;
-    Linked<IOutputMetaData> expectedDiskMeta = nullptr;
-    Linked<IOutputMetaData> projectedDiskMeta = nullptr;
-    Linked<IOutputMetaData> actualDiskMeta = nullptr;
-    unsigned expectedCrc = 0;
-    unsigned projectedCrc = 0;
-    unsigned actualCrc = 0;
+    RtlDynamicRowBuilder allocatedBuilder;
+    const IDynamicTransform * translator;
+    const IKeyTranslator * keyedTranslator;
```
jakesmith: style: is there actual overhead if these are initialzed to null here? If not seems preferable for readability/future safety.

common/thorhelper/thorread.cpp
```diff
@@ -444,108 +602,414 @@ const void *BinaryDiskRowReader::nextRow()
...
+        if (likely(fieldFilterMatch(next))) // NOTE - keyed fields are checked pre-translation
+            return processor(sizeRead, next);
+    }
+}
+
+
+//Implementation of IAllocRowStream, return a row allocated with roxiemem
+const void *BinaryDiskRowReader::nextRow()
+{
+    return inlineNextRow([this](size32_t sizeRead, const byte * next){
```
jakesmith: personally, I think more readable if lambda function defined on a separate line.

jakesmith:
@ghalliday - looks good afaics, I added a few minor comments.

common/thorhelper/thorread.cpp
ghalliday: I always use two spaces after a '.' or '?' at the start of a new sentence.

common/thorhelper/thorread.cpp
ghalliday: Because I wanted to make it clear that it was an atomic - and therefore wouldn't suffer from the problem of not using an atomic in this situation.

common/thorhelper/thorread.hpp
ghalliday: Should we remove the use of extends from within our code?

common/thorhelper/thorread.cpp
ghalliday: yes, I think we should require all members are initialised - unless there is a very good reason not to

common/thorhelper/thorread.cpp
ghalliday: In general I'm not sure it can - because calculating it can cause some non const functions to be called.  I think it started off const.

common/thorhelper/thorxmlwrite.cpp
ghalliday: changed

ecl/hqlcpp/hqlsource.cpp
ghalliday: changed

common/thorhelper/thorread.cpp
ghalliday: Like this?
```
    return inlineNextRow(
        [this](size32_t sizeRead, const byte * next)
        {
            if (needToTranslate)
            {
                size32_t size = translator->translate(allocatedBuilder.ensureRow(), *this, next);
                return allocatedBuilder.finalizeRowClear(size);
            }
            else
            {
                size32_t allocatedSize;
                void * result = outputAllocator->createRow(sizeRead, allocatedSize);
                memcpy(result, next, sizeRead);
                return (const void *)outputAllocator->finalizeRow(sizeRead, result, allocatedSize);
            }
        }
    );

```

common/thorhelper/thorread.hpp
```diff
@@ -27,19 +27,56 @@
...
+public:
+    // Accessor functions to provide the basic information from the disk read
+    virtual const char * queryFormat() const = 0;
+    virtual unsigned getActualCrc() const = 0;
+    virtual unsigned getExpectedCrc() const = 0;
+    virtual unsigned getProjectedCrc() const = 0;
+    virtual IOutputMetaData * queryActualMeta() const = 0;
+    virtual IOutputMetaData * queryExpectedMeta() const = 0;
+    virtual IOutputMetaData * queryProjectedMeta() const = 0;
+    virtual const IPropertyTree * queryOptions() const = 0;
```
jpmcmu: Based on the CSV option test below it looks like generic options would be supported and are injected here. Is that correct?

ghalliday:
@jakesmith see responses and update to the PR

jpmcmu (commented):
@ghalliday Looks good to me. I had a question on support for generic options for file formats, and a few questions that are probably beyond the scope of this PR about how this code could be extended to support "foreign" data sources like HDFS. As an example being able to query HDFS about # of partitions and partition locality would be beneficial. I will follow up offline.

common/thorhelper/thorread.hpp
jakesmith: 'interface' is non-standard too.. I feel it (and personally 'extends') clarifies the intended pattern.

common/thorhelper/thorread.cpp
jakesmith: I think that's clearer, but actually what I meant was to declare it with an auto.
So :
```
        auto processorFunc = [this](size32_t sizeRead, const byte * next)
        {
            if (needToTranslate)
            {
                size32_t size = translator->translate(allocatedBuilder.ensureRow(), *this, next);
                return allocatedBuilder.finalizeRowClear(size);
            }
            else
            {
                size32_t allocatedSize;
                void * result = outputAllocator->createRow(sizeRead, allocatedSize);
                memcpy(result, next, sizeRead);
                return (const void *)outputAllocator->finalizeRow(sizeRead, result, allocatedSize);
            }
        };

        return inlineNextRow(processorFunc);

```

common/thorhelper/thorread.hpp
ghalliday: Yes, there is support for arbitrary options in the interfaces - but no compiler support yet.

ghalliday:
Once this is merged I can try and work on the next stage in the compiler and else where of supporting arbitrary files.
There are two parts to that problem
a) getting the metadata for the file - e.g., number of parts etc.
b) reading a single part of the file.
This PR only really covers the second of those.

ghalliday:
@richardkchapman please look at the further changes.  I can squash once you are happy.

richardkchapman:
@ghalliday Please squash

common/thorhelper/thorread.cpp
ghalliday: ok, I may change to that in a subsequent PR

PR #12963, opened 2019-09-11
HPCC-21820 Add options to generate record formats to the workunit

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/ecl/hqlcpp/hqlhtcpp.cpp
+++ b/ecl/hqlcpp/hqlhtcpp.cpp
@@ -2197,7 +2197,7 @@ void ActivityInstance::createGraphNode(IPropertyTree * defaultSubGraph, bool alw
                 addAttribute(WaMetaGroupSortOrder, getExprECL(groupSortOrder, s.clear(), true).str());
         }
 
-        if (options.noteRecordSizeInGraph)
+        if (options.noteRecordSizeInGraph || options.generateActivityFormats)
         {
             LinkedHqlExpr record = activityExpr->queryRecord();
             if (!record && (getNumChildTables(activityExpr) == 1))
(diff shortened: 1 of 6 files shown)


ghalliday:
@jakesmith fyi
@shamser please review

ecl/hqlcpp/hqlhtcpp.cpp
```diff
@@ -10226,6 +10232,14 @@ void HqlCppTranslator::buildRecordEcl(BuildCtx & subctx, IHqlExpression * record
 }
 
 
+void HqlCppTranslator::addFormatAttribute(ActivityInstance & instance, WuAttr attr, IHqlExpression * record)
+{
+    //MORE: Better would be to add a section to the workunit and only generate the record formats once, and reference them from here
```
ghalliday: HPCC-22805 has been added to come back to this, but I doubt it will be worth it unless there are more reasons to make the change.

shamser (commented):
@ghalliday Just one concern: record formats may be quite large.  Can large property values cause problems?  Will it even be viewable in graph viewer?  Shouldn't something like this be logged in thor/roxie rather than stored in the graph?

ghalliday:
@shamser yes I agree this is not an option that could be enabled by default.  This is primarily here at the moment so Jake can find out the format of a spill file so he can construct a small example.

The format can't be logged in thor/roxie because they do not have that information - this is providing an initial way to export it from the code generator.  If it was enabled all the time it would need to use a different mechanism - to store the values, and to ensure duplicates are commoned up.

shamser (approved):
@richardkchapman Code reviewed.  Looks good.

PR #12979, opened 2019-09-13
HPCC-22265 Support FieldFilter syntax field!=[sets] for != matching

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/rtl/eclrtl/rtlnewkey.cpp
+++ b/rtl/eclrtl/rtlnewkey.cpp
@@ -94,7 +94,7 @@ static void readSubLength(size32_t & subLength, const char * &in)
 
 void readFieldFromFieldFilter(StringBuffer & fieldText, const char * & src)
 {
-    readUntilTerminator(fieldText, src, "=*:");
+    readUntilTerminator(fieldText, src, "=*:!");
 }
 
 void deserializeSet(ISetCreator & creator, const char * filter)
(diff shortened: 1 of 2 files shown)


ghalliday:
@rpastrana fyi
@richardkchapman please review once smoke test has completed.

richardkchapman:
@richardkchapman

rtl/eclrtl/rtlnewkey.cpp
```diff
@@ -2058,24 +2068,13 @@ IFieldFilter * deserializeFieldFilter(unsigned fieldId, const RtlTypeInfo & type
         {
             char * end;
             size32_t subLength = strtoul(src+1, &end, 10);
+            //This could support !, but this is a legacy syntax : is now specified as part of the set
             if (*end != '=')
                 UNIMPLEMENTED_X("Expected =");
 
-            //Created a truncated type
```
richardkchapman: This refactoring seems unrelated to the feature being implemented? I assume it is using code (subStrinfFieldFilter) that was not available when this was first implemented?

richardkchapman: Somewhat concerned it's using a different type from what it used to

ghalliday: I think the reason was that the test code now calls this function: deserializeFieldFilter, and the way this function handled substrings didn't match the way they were handled in the test code.  I thought the test code was correct, but I will double check again.

ghalliday: Without the change the regression test fails.  I still need to check whether the change is correct though...

ghalliday: I think I would keep this change - this code is used when string filters are processed in dafilesrv - which means it should not assume it has been serialized by a fieldvalue.  (Without the change some of the test generate an internal error.)  It is also in a deprecated syntax, so unlikely to cause any issues.

ghalliday:
@richardkchapman would take a is.

PR #12981, opened 2019-09-13
HPCC-22802 Heuristic to disable remote index streaming inside CQ's

jakesmith:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/thorlcr/activities/indexread/thindexreadslave.cpp
+++ b/thorlcr/activities/indexread/thindexreadslave.cpp
@@ -77,6 +77,7 @@ protected:
     bool rowLimitSkips = false;
     rowcount_t keyedProcessed = 0;
     rowcount_t rowLimit = RCMAX;
+    bool useRemoteStreaming = false;
 
 
     class TransformCallback : implements IThorIndexCallback , public CSimpleInterface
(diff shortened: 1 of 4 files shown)


jakesmith:
@ghalliday - please review

thorlcr/activities/indexread/thindexreadslave.cpp
```diff
@@ -168,15 +168,65 @@ class CIndexReadSlaveBase : public CSlaveActivity
...
+                     */
+
+                    return ((size * numParts) * compressionMultiple) < (keyCacheSize * cacheSizeMultiple);
+                };
+
+                if (0 != (helper->getFlags() & TIRusesblob))
+                    tryRemoteStream = false;
+                else if (localMerge)
+                    tryRemoteStream = false;
+                else if (inChildQuery() && !getOptBool(THOROPT_STREAM_REMOTE_SMALL_INDEX)) // option turns of heuristic
```
ghalliday: Shouldn't this be:

  if (inChildQuery())
  {
     if (getOptBool())
     {...code below }
     else
          tryRemoteStream = false;

?
Currently the option is forceStreamRemoteSmallIndex, I had assumed it would be allowStreamRemoteSmallIndex from the description.

Perhaps renaming it would make it clear.  Or the option could use the heuristic as the default allowing it to be overridden either way??

thorlcr/activities/indexread/thindexreadslave.cpp
```diff
@@ -168,15 +168,65 @@ class CIndexReadSlaveBase : public CSlaveActivity
...
+                     * If in a CQ, it is counterproductive to use an index read stream per CQ execution if the index
+                     * involved is relatively small.
+                     * Because, if the index is small and direct reading (and caching) key node pages, it is likely
+                     * that repeated executions will not read any (or few) new key pages (i.e. cache hit).
+                     *
+                     * Example: small 1-way key being remotely read by the whole cluster.
+                     * If it is small it will fit (or mostly fit) in node key cache, and thus mostly read from memory vs over the network etc.
+                     *
+                     */
+                    offset_t logicalFileSize = part.queryOwner().queryProperties().getPropInt64("@size"); // NB: size is compressed size
```
ghalliday: Shouldn't this be calculated once before reading the indexes, rather than on each file part?

thorlcr/activities/indexread/thindexreadslave.cpp
```diff
@@ -203,7 +253,22 @@ class CIndexReadSlaveBase : public CSlaveActivity
                             }
                         }
                         else
-                            remoteCandidates.push_back(copy);
+                        {
+                            if ((0 == avgPartSize) && inChildQuery() && !getOptBool(THOROPT_STREAM_REMOTE_SMALL_INDEX)) // if file size from meta was unavailable
```
ghalliday: I think this might mean some parts are streamed remote and some are executed locally? It doesn't feel like the decision is in the correct place.

thorlcr/graph/thgraph.cpp
```diff
@@ -2729,12 +2734,12 @@ void CJobBase::startJob()
     setPerformanceMonitorHook(perfmonhook);
     PrintMemoryStatusLog();
     logDiskSpace();
-    unsigned keyNodeCacheMB = (unsigned)getWorkUnitValueInt("keyNodeCacheMB", DEFAULT_KEYNODECACHEMB * queryJobChannels());
-    unsigned keyLeafCacheMB = (unsigned)getWorkUnitValueInt("keyLeafCacheMB", DEFAULT_KEYLEAFCACHEMB * queryJobChannels());
-    unsigned keyBlobCacheMB = (unsigned)getWorkUnitValueInt("keyBlobCacheMB", DEFAULT_KEYBLOBCACHEMB * queryJobChannels());
-    setNodeCacheMem(keyNodeCacheMB * 0x100000);
-    setLeafCacheMem(keyLeafCacheMB * 0x100000);
-    setBlobCacheMem(keyBlobCacheMB * 0x100000);
+    keyNodeCacheMB = (unsigned)getWorkUnitValueInt("keyNodeCacheMB", DEFAULT_KEYNODECACHEMB * queryJobChannels()) * 0x100000;
```
ghalliday: No longer stored in MB, and not unreasonable to overflow 4G on some systems, so unsigned is too small.

ghalliday (commented):
@jakesmith various comments.

thorlcr/activities/indexread/thindexreadslave.cpp
```diff
@@ -168,15 +168,65 @@ class CIndexReadSlaveBase : public CSlaveActivity
...
+                    if (logicalFileSize)
+                    {
+                        unsigned numKeyParts = part.queryOwner().numParts();
+                        if ((numKeyParts>1) && !part.queryOwner().queryProperties().getPropBool("@local"))
+                            numKeyParts--; // TLK
+                        avgPartSize = logicalFileSize / numKeyParts;
+
+                        // NB: partDescs.ordinality() == <number of indexparts this slave is handling>
+
+                        if (keyCacheRatioHeuristic(avgPartSize, partDescs.ordinality(), 10, 4))
```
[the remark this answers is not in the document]
jakesmith: >Why partDescs.ordinality() instead of numKeyParts?

no because numKeyParts is total # of index [data] parts, and partDescs is just those this slave is dealing with.

>Not clear where the factor of 10 comes from. Can it be deduced from the index meta data,. and if not should we add it?

It was a very rough guess.
The expanded size isn't available in the published meta data. Otherwise, yes I could have based it on that.

thorlcr/activities/indexread/thindexreadslave.cpp
jakesmith: It is 'forceStreamRemoteSmallIndex' and yes that would be better name for it's current meaning.

I might change it to "minimumCQStreamedIndexSizeMB" , default it to 0, and for that to mean use internal rough heuristic multiples..
If defined, it will be absolute size check, allowing someone to force smaller or larger indexes to be streamed / not-streamed if wanted.

thorlcr/activities/indexread/thindexreadslave.cpp
jakesmith: yes agree, I'll move the calculation so done once.

thorlcr/activities/indexread/thindexreadslave.cpp
jakesmith: agree, will move so calculated once.
Although the existing code still does allow for some to be streamed and some not, depending on whether the parts have different formats and are/are not translatable. e.g. could be a super with different formats, where some are are translatable and therefore streamable, and others are not.

thorlcr/graph/thgraph.cpp
jakesmith: agree, will fix.

jakesmith:
@ghalliday - please re-review

thorlcr/activities/indexread/thindexreadslave.cpp
```diff
@@ -558,6 +560,85 @@ class CIndexReadSlaveBase : public CSlaveActivity
...
+                    else // no min. size to stream set, so use a heuristic
+                    {
+                        /*
+                         * Rough heuristic.
+                         *
+                         * If (([average compressed part size] * [# parts handling] * [compressionMultiple] * [cacheSizeFitPercentage%]) > [keyCacheSize])
+                         *     then useRemoteStreaming = true
+                         *
+                         * i.e. if the [cacheSizeFitPercentage] % of total size of the compressed index data this slave is handling multiplied
+                         * by a rough compression multiplier [compressionMultiple] if larger than the cache size [keyCacheSize], then use streaming.
```
ghalliday: if->is and missing ) on the next line

thorlcr/activities/indexread/thindexreadslave.cpp
```diff
@@ -558,6 +560,85 @@ class CIndexReadSlaveBase : public CSlaveActivity
...
+                        // anything larger is streamed, anything smaller is read directly
+                        if (logicalFileSize > (minRemoteCQIndexSizeMb * 0x100000))
+                            useRemoteStreaming = true;
+                    }
+                    else // no min. size to stream set, so use a heuristic
+                    {
+                        /*
+                         * Rough heuristic.
+                         *
+                         * If (([average compressed part size] * [# parts handling] * [compressionMultiple] * [cacheSizeFitPercentage%]) > [keyCacheSize])
```
ghalliday: What is the reason for multiplying by #parts handling rather than #parts?
I would have expected the test to be "does a good proportion of this index fit into the local index cache?"

ghalliday (commented):
@jakesmith a few minor comments and one question

thorlcr/activities/indexread/thindexreadslave.cpp
```diff
@@ -558,6 +560,85 @@ class CIndexReadSlaveBase : public CSlaveActivity
         statsArr = _statsArr.getArray();
         lastSeeks = lastScans = 0;
         localMerge = (localKey && partDescs.ordinality()>1) || seekGEOffset;
+
+        if (parts)
```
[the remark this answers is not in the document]
jakesmith: yes, the logical filename has been bound by the time the slave activities init() calls are made.
If a dynamic filename is used in a child query. The activities that need it (see needReInit()), will be reinitialized, i.e. the new parts will be serialized/deserliazed here.

thorlcr/activities/indexread/thindexreadslave.cpp
jakesmith: because the cache is per slave, and this slave is only loading key node pages from the parts it has, not the whole key
So the heuristic here is, does a good proportion of the sum [approx.] f the parts I'm handling (often just 1 part), fit in to the cache. In a CQ - it's dealing with all parts, so same thing as whole index.

thorlcr/activities/indexread/thindexreadslave.cpp
jakesmith: will fix

thorlcr/activities/indexread/thindexreadslave.cpp
jakesmith: I've added a comment.

jakesmith:
@ghalliday - please see replies/minor changes.

thorlcr/activities/indexread/thindexreadslave.cpp
ghalliday: ok, I was confusing it with a keyed join!

PR #13011, opened 2019-09-19
HPCC-22475 ESP service to sign ECL code

mayx:
- Create ecm and other source files for the service
- Add deployment and configuration scripts for the service
- Implement the Sign functions with gpg
- Add xslt file to display sign response in a more readable format

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Premature optimization

## Smoketest:


mayx:
@shamser please review...

esp/scm/ws_codesign.ecm
```diff
@@ -0,0 +1,44 @@
...
+{
+    gpg ("gpg")
+};
+
+ESPrequest SignRequest
+{
+    ESPenum SigningMethodType SigningMethod("gpg");
+    [cols(20)] string KeyID;
+    [password, cols(20)] string KeyPass;
+    [cols(70), rows(20)] string Text;
```
shamser: I assume "cols" and "rows"  modifies does not limit the maximum size of the source file submitted (Text)

esp/services/ws_codesign/ws_codesignService.cpp
```diff
@@ -0,0 +1,92 @@
...
+    if(keylen != 16 && keylen != 40)
+    {
+        resp.setErrMsg("Invalid KeyID");
+        return false;
+    }
+
+    StringBuffer cmd, output, errmsg;
+
+    cmd.appendf("gpg --list-keys %s", keyid);
+    int ret = runExternalCommand(output, errmsg, cmd.str(), "");
```
shamser: I understand there are a number of libraries that should allow the functionality to be accessed directly, rather than access via an external command.  Here's one that I found after a quick google:https://www.gnupg.org/documentation/manuals/gpgme/.  There may be better ones out there.

esp/services/ws_codesign/ws_codesignService.cpp
```diff
@@ -0,0 +1,92 @@
...
+    if (ret != 0 || strstr(output.str(), keyid) == nullptr)
+    {
+        resp.setErrMsg("Key not found");
+        return false;
+    }
+
+    output.clear();
+    errmsg.clear();
+    runExternalCommand(output, errmsg, "gpg-connect-agent", "RELOADAGENT");
```
shamser: If the passphrase is being provided, why is it necessary to use gpg agents? Is it to unload cached secrets?

esp/services/ws_codesign/ws_codesignService.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+    output.clear();
+    errmsg.clear();
+    cmd.clear().appendf("gpg --clearsign --default-key %s --pinentry-mode loopback --yes --batch", keyid);
```
shamser: This is changing the default key for the user every-time it's executed.    You probably want the "-u" option to specify the key id to use.  (This should also mean that you do not need to call gpg-connect-agent reload)

esp/services/ws_codesign/ws_codesignService.cpp
```diff
@@ -0,0 +1,92 @@
...
+    ret = runExternalCommand(output, errmsg, cmd.str(), input.str());
+    if (ret != 0 || output.length() == 0)
+    {
+        UERRLOG("gpg clearsign error: [%d] %s\nOutput: n%s", ret, errmsg.str(), output.str());
+        resp.setErrMsg("Failed to sign text, please check service log for details");
+        return false;
+    }
+
+    resp.setRetCode(0);
+    resp.setSignedText(output.str());
```
shamser: Does ESP automatically encode the response so it's valid in json or xml?

shamser (commented):
@mayx I have some questions.  Please can you message on flowdock when you'd like me to have another look. Thanks.

esp/scm/ws_codesign.ecm
mayx: Right it only controls the display of the forms.

esp/services/ws_codesign/ws_codesignService.cpp
mayx: Right I saw the gpgme lib as well. But since we're probably going to re-implement this service with PKI when that is in place, I don't think it's worth it to add the gpgme dependency at this moment.

esp/services/ws_codesign/ws_codesignService.cpp
mayx: gpg always uses agent, which caches the passphrase for 10 minutes. Yes this step is to clear the cached secret.

esp/services/ws_codesign/ws_codesignService.cpp
mayx: Agree -u option is better, I'll change it.

esp/services/ws_codesign/ws_codesignService.cpp
mayx: Yes it does. But as a result, you need to xml-decode the result before using it. I added xslt transform to the response page GUI, from which the signed text can be copy-pasted directly from the web interface.

mayx:
@shamser change made per suggestion, please review again...

esp/services/ws_codesign/ws_codesignService.cpp
shamser: I would use the --no-use-agent to stop the passphrase being cached.  Repeatedly asking the agent to reload is not a good idea: (1) there could be something else using the agent and this code will be affecting that process too (2) it wouldn't be very efficient.

shamser (changes requested):
@mayx Please can you see the comment regarding agent

mayx:
@shamser the --no-use-agent is deprecated:
"gpg: WARNING: "--no-use-agent" is an obsolete option - it has no effect"
Also I timed RELOADAGENT, it takes several milliseconds so is very fast. I can do some stress test and see if there's any race conditions for multiple threads, but I doubt it.

mayx:
@shamser I tested with 5 concurrent threads and no issue was found...

mayx:
And I tested with the gpg-connect-agent command commented out, the difference in overall performance is very small.

shamser:
@afishbeck It seems the agent has having to be reloaded.  Do you know a better way?  @mayx At the very least, we should be invalidating just the keyid's in question not all the ids.

mayx:
@shamser I don't see a way from gpg to just clear one key. What I can do is to cache a hash of the passphrase on ESP for 10 minutes, and only call gpg reload upon cache timeout. This should dramatically reduces the calls to the reload. What do you think?

mayx:
@shamser after some digging, I came across a way to clear one key only, will test more and check in a fix.

mayx:
@shamser change checked in to clear passphrase of only one cached key, please review again...

esp/services/ws_codesign/ws_codesignService.cpp
```diff
@@ -0,0 +1,100 @@
...
+    int keylen = strlen(keyid);
+    if(keylen != 16 && keylen != 40)
+    {
+        resp.setErrMsg("Invalid KeyID");
+        return false;
+    }
+
+    StringBuffer cmd, output, errmsg;
+
+    cmd.appendf("gpg --list-secret-keys --with-keygrip %s", keyid);
```
ghalliday: security: Possibly being pedantic, but passing a user parameter unchecked is a potential security hole.  E.g. I could pass an extra options that I wanted, as long as I pad to 40 characters

esp/services/ws_codesign/ws_codesignService.cpp
```diff
@@ -0,0 +1,100 @@
...
+    resp.setRetCode(-1);
+
+    const char* keyid = req.getKeyID();
+    const char* text = req.getText();
+    if (!keyid || !*keyid || !text || !*text)
+    {
+        resp.setErrMsg("Please provide both KeyID and Text");
+        return false;
+    }
+    int keylen = strlen(keyid);
```
ghalliday: picky, should be type size_t

esp/services/ws_codesign/ws_codesignService.cpp
```diff
@@ -0,0 +1,100 @@
...
+    cmd.appendf("gpg --list-secret-keys --with-keygrip %s", keyid);
+    int ret = runExternalCommand(output, errmsg, cmd.str(), "");
+    if (ret != 0 || strstr(output.str(), keyid) == nullptr)
+    {
+        resp.setErrMsg("Key not found");
+        return false;
+    }
+
+    StringBuffer keygrip;
+    if (auto kgptr = strstr(output.str(), "Keygrip = "))
```
ghalliday: style: I think this would be clearer, if the assignment was on a separate line from an if, and less similar to an error prone assignment within an if condition.

esp/services/ws_codesign/ws_codesignService.cpp
ghalliday: The --with-keygrip option is only supported on gpg > 2.1, which does not include the default installation for ubuntu 16.04.

esp/services/ws_codesign/ws_codesignService.cpp
```diff
@@ -0,0 +1,100 @@
...
+        output.clear();
+        errmsg.clear();
+        cmd.clear().appendf("gpg-connect-agent \"clear_passphrase --mode=normal %s\" /bye", keygrip.str());
+        runExternalCommand(output, errmsg, cmd.str(), "");
+    }
+
+    output.clear();
+    errmsg.clear();
+    cmd.clear().appendf("gpg --clearsign -u %s --pinentry-mode loopback --yes --batch", keyid);
+    StringBuffer passbuf(req.getKeyPass()), input;
```
ghalliday: style: A bit unusual to have multiple declarations on the same line with a constructor.  Separate lines would be easier to scan.

esp/services/ws_codesign/ws_codesignService.cpp
```diff
@@ -0,0 +1,100 @@
...
+    {
+        output.clear();
+        errmsg.clear();
+        cmd.clear().appendf("gpg-connect-agent \"clear_passphrase --mode=normal %s\" /bye", keygrip.str());
+        runExternalCommand(output, errmsg, cmd.str(), "");
+    }
+
+    output.clear();
+    errmsg.clear();
+    cmd.clear().appendf("gpg --clearsign -u %s --pinentry-mode loopback --yes --batch", keyid);
```
ghalliday: What happens if the keyid contains a space?

esp/services/ws_codesign/ws_codesignService.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+    const char* keyid = req.getKeyID();
+    const char* text = req.getText();
+    if (!keyid || !*keyid || !text || !*text)
+    {
+        resp.setErrMsg("Please provide both KeyID and Text");
+        return false;
+    }
+    int keylen = strlen(keyid);
+    if(keylen != 16 && keylen != 40)
```
ghalliday: Why is keylen restricted to 16 or 40 characters? From looking at the code I thought it was the username that was associated with the key/value pair.  That could be any length.

ghalliday (commented):
@mayx a few comments.

I'm also not 100% sure what the keyid refers to.  I would expect it to be a user readable descriptive string.

esp/services/ws_codesign/ws_codesignService.cpp
```diff
@@ -0,0 +1,100 @@
...
+{
+}
+
+void Cws_codesignEx::init(IPropertyTree *cfg, const char *process, const char *service)
+{
+    if(cfg == nullptr)
+        throw MakeStringException(-1, "Cannot initialize Cws_codesignEx, cfg is NULL");
+
+    StringBuffer xpath;
+    xpath.appendf("Software/EspProcess[@name=\"%s\"]/EspService[@name=\"%s\"]", process, service);
```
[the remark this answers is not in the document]
mayx: Yes agree, I'll create a Jira for this.

esp/services/ws_codesign/ws_codesignService.cpp
mayx: Should check for spaces in keyid, also I'll check the version and with older version I'll use the other (not as good method) to clear the passphrase.

esp/services/ws_codesign/ws_codesignService.cpp
mayx: I will check for spaces.

esp/services/ws_codesign/ws_codesignService.cpp
mayx: The Key can be identified by the user, or the ID. The ID is either 16 or 40 chars long for gpg. But  on second thought it probably is better to also allow user as the identifier. So I'll rename KeyID to KeyIdentifier, remove length restriction, and check for spaces.

mayx: Actually space is allowed if the identifier is "user <user@mail.com>". I'll instead check for quote, and then enclose the value with quotes.

ghalliday: Yes, when I was testing my userid was "Gavin Halliday"

esp/services/ws_codesign/ws_codesignService.cpp
mayx: Actually gpg 1.x doesn't cache passphrase so that step can be spared.

mayx:
Checked in the following changes:
- Change KeyID to KeyIdentifier
- Enclose key identifier in quotes on the command line so spaces won't break it
- Make the commands backward compatible with version 1
- Other misc changes suggested by gavin
@ghalliday please review again...

esp/scm/ws_codesign.ecm
```diff
@@ -0,0 +1,44 @@
...
+};
+
+ESPresponse [exceptions_inline] SignResponse
+{
+    int RetCode;
+    string ErrMsg;
+    string SignedText;
+};
+
+ESPservice [auth_feature("WsCodeSignAccess:ACCESS"), version("1.0"), default_client_version("1.0"), exceptions_inline("./smc_xslt/exceptions.xslt")] ws_codesign
```
richardkchapman: If Russ changes the name as I suggested in my review of his PR then you will need to change these lines, I think?

esp/services/ws_codesign/ws_codesignService.cpp
```diff
@@ -0,0 +1,110 @@
...
+        auto kgptr = strstr(output.str(), "Keygrip = ");
+        if (kgptr)
+            keygrip.append(40, kgptr+10);
+
+        if (keygrip.length() > 0)
+        {
+            output.clear();
+            errmsg.clear();
+            cmd.clear().appendf("gpg-connect-agent \"clear_passphrase --mode=normal %s\" /bye", keygrip.str());
+            runExternalCommand(output, errmsg, cmd.str(), "");
```
shamser: If we really want to be secure, this should be done after signing.

shamser (commented):
@mayx Reviewed.  Just one comment.

esp/services/ws_codesign/ws_codesignService.cpp
mayx: Let me do it before and after to remain on the safe side. Command line users might be a bit annoyed but that's not our targeted use case.

PR #13016, opened 2019-09-24
HPCC-22839 Thor master does not end after job abort

mckellyln:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

## Testing:

Smoketest and manual stopping Thor master in various ways

### DIFF
--- a/thorlcr/master/thmastermain.cpp
+++ b/thorlcr/master/thmastermain.cpp
@@ -441,9 +441,63 @@ bool checkClusterRelicateDAFS(IGroup &grp)
 
 static bool auditStartLogged = false;
 
+class CThorEndHandler : public CSimpleInterface, implements IThreaded
+{
+    CThreaded threaded;
+    unsigned timeout = 30000;
+    std::atomic<bool> started{false};
+    std::atomic<bool> stopped{false};
+    Semaphore sem;
+public:
+    CThorEndHandler() : threaded("CThorEndHandler")
+    {
+        threaded.init(this);
+    }
+    ~CThorEndHandler()
+    {
+        stop();
+        threaded.join(timeout);
+    }
+    void start(unsigned timeoutSecs)
+    {
+        bool expected = false;
+        if (started.compare_exchange_strong(expected, true));
+        {
+            timeout = timeoutSecs * 1000; // sem_post and sem_wait are mem_barriers
+            sem.signal();
+        }
+    }
+    void stop()
+    {
+        bool expected = false;
+        if (stopped.compare_exchange_strong(expected, true));
+            sem.signal();
+    }
+    virtual void threadmain() override
+    {
+        // wait to be signalled to start timer
+        sem.wait();
+        if (stopped)
+            return;
+        if (!sem.wait(timeout))
+        {
+            // if it wasn't set by now then it's -1 and Thor restarts ...
+            int eCode = queryExitCode();
+            _exit(eCode);
+        }
+    }
+};
+// start thread now
+static CThorEndHandler thorEndHandler;
+
 static bool firstCtrlC = true;
 bool ControlHandler(ahType type)
 {
+    // MCK - NOTE: this routine may make calls to non-async-signal safe functions
+    //             (such as malloc) that really should not be made if we are called
+    //             from a signal handler - start end handler timer to always end
+    thorEndHandler.start(120);
+
     if (ahInterrupt == type)
     {
         if (firstCtrlC)
@@ -470,14 +524,7 @@ bool ControlHandler(ahType type)
                     queryServerStatus().queryProperties()->queryProp("@queue"));
             }
             queryLogMsgManager()->flushQueue(10*1000);
-#ifdef _WIN32
-            TerminateProcess(GetCurrentProcess(), 1);
-#else
-            //MORE- verify this
-            // why not just raise(SIGKILL);  ?
-            kill(getpid(), SIGKILL);
-#endif
-            _exit(1);
+            _exit(TEC_CtrlC);
         }
     }
     // ahTerminate
@@ -842,11 +889,16 @@ int main( int argc, char *argv[]  )
         FLLOG(MCexception(e), thorJob, e,"ThorMaster");
         e->Release();
     }
+
+    // cleanup handler to be sure we end
+    thorEndHandler.start(30);
+
+    PROGLOG("Thor closing down 5");
     stopPerformanceMonitor();
     disconnectLogMsgManagerFromDali();
     closeThorServerStatus();
-    if (globals) globals->Release();
-    PROGLOG("Thor closing down 5");
+    if (globals)
+        globals->Release();
     PROGLOG("Thor closing down 4");
     closeDllServer();
     PROGLOG("Thor closing down 3");


mckellyln:
@jakesmith please review

thorlcr/master/thmastermain.cpp
```diff
@@ -441,9 +441,67 @@ bool checkClusterRelicateDAFS(IGroup &grp)
...
+    {
+        // wait to be signalled to start timer
+        sem.wait();
+        if (stopped)
+            return;
+        if (!sem.wait(timeout))
+            raise(SIGKILL);
+    }
+};
+CThorEndHandler thorendhdlr;
```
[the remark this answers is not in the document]
jakesmith: I think it's okay, but I don't know anywhere else we start a thread like like this, i.e. as the module loads/constructs.

thorlcr/master/thmastermain.cpp
```diff
@@ -441,9 +441,67 @@ bool checkClusterRelicateDAFS(IGroup &grp)
...
+
 static bool firstCtrlC = true;
 bool ControlHandler(ahType type)
 {
+    // MCK - NOTE: this routine may make calls to non-async-signal safe functions
+    //             (such as malloc) that really should not be made if we are called
+    //             from a signal handler - start end handler timer to always end
+    thorendhdlr.start(120);
```
jakesmith: Is the reason for pre-constructing and pre-starting the thread on line 495, to avoid creating a new thread here inside a signal handler?

I think so and makes sense, but the pre-existing code is also not safe (it calls malloc for example when calling mp cancel()). So all (or most) this code should move out of the signal handler I suspect and be called on a thread.

jakesmith:
@mckellyln - I think ok as is with some minor changes. The code that shouldn't really be called in a signal handler is a separate/pre-existing issue, and could be tackled separately.

thorlcr/master/thmastermain.cpp
mckellyln: Correct, we really should move the bulk of the signal handler code into a separate thread and the handler just signals that thread to go on.  But that is a separate and larger issue.  I added the thorendhdlr.start(120) here as it seemd good insurance for Thor to always end.

thorlcr/master/thmastermain.cpp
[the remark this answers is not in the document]
mckellyln: updated

thorlcr/master/thmastermain.cpp
```diff
@@ -441,9 +441,67 @@ bool checkClusterRelicateDAFS(IGroup &grp)
...
+        }
+    }
+    void start(unsigned timeoutsecs)
+    {
+        if (!started)
+        {
+            timeout = timeoutsecs * 1000;
+            start();
+        }
+    }
```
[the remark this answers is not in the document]
mckellyln: updated

mckellyln:
@jakesmith updated PR as per your comments, thanks.  Please re-review.

thorlcr/master/thmastermain.cpp
```diff
@@ -441,9 +441,61 @@ bool checkClusterRelicateDAFS(IGroup &grp)
...
 
+class CThorEndHandler : public CSimpleInterface, implements IThreaded
+{
+    CThreaded threaded;
+    unsigned timeout;
+    std::atomic<bool> started{false};
+    std::atomic<bool> stopped{false};
+    Semaphore sem;
+public:
+    CThorEndHandler() : threaded("CThorEndHandler"), timeout(30000)
```
jakesmith: picky/trivial: better to initialize on member declaration.

thorlcr/master/thmastermain.cpp
```diff
@@ -441,9 +441,61 @@ bool checkClusterRelicateDAFS(IGroup &grp)
...
+            sem.signal();
+        }
+    }
+    void stop()
+    {
+        bool expected = false;
+        if (stopped.compare_exchange_strong(expected, true));
+        {
+            sem.signal();
+        }
```
jakesmith: style: normally we don't put 1 line blocks in braces.

jakesmith:
@mckellyln - 2 trivial comments, please squash ready for merge.

jakesmith:
I'm unsure, but as not a new issue, I think this should retarget 7.6

thorlcr/master/thmastermain.cpp
mckellyln: done

thorlcr/master/thmastermain.cpp
mckellyln: done

mckellyln:
Squashed.
Since it is an ongoing issue I thought should target 7.4.x but up to @jakesmith / @richardkchapman

jakesmith:
Will defer to @richardkchapman ..

thorlcr/master/thmastermain.cpp
```diff
@@ -441,9 +441,59 @@ bool checkClusterRelicateDAFS(IGroup &grp)
...
+            sem.signal();
+    }
+    virtual void threadmain() override
+    {
+        // wait to be signalled to start timer
+        sem.wait();
+        if (stopped)
+            return;
+        if (!sem.wait(timeout))
+            raise(SIGKILL);
```
richardkchapman: Does this compile on Windows, I wonder?

richardkchapman:
@mckellyln 7.4.x is fine - but needs checking re Windows builds

thorlcr/master/thmastermain.cpp
jakesmith: I don't think it will, the other similar case (in ControlHandler in this file) has #ifdef Win32 to call TerminateProcess.

jakesmith: Would _exit be a good alternative here?
If not, might be worth adding a TerminateSelf or similar to platform.h - so can use here/other places platform independently.

mckellyln:
@jakesmith @richardkchapman I will change raise() to _exit() and update for Windows.

thorlcr/master/thmastermain.cpp
```diff
@@ -480,7 +480,13 @@ class CThorEndHandler : public CSimpleInterface, implements IThreaded
         if (stopped)
             return;
         if (!sem.wait(timeout))
-            raise(SIGKILL);
+        {
+#ifdef _WIN32
+            TerminateProcess(GetCurrentProcess(), 1);
+#else
+            _exit(1);
```
jakesmith: wouldn't _exit be equally applicable to Windows?

mckellyln: Yes I think so.  I copied code from somewhere else.  Let's just use _exit(1) for both.

jakesmith:
>Let's just use _exit(1) for both.

@mckellyln - weren't you going to remove #ifdef WIN32 conditional code for this (in both places)?

mckellyln:
@jakesmith updated commit \#2

jakesmith:
@mckellyln - looks please squash ready for merge.

mckellyln:
@jakesmith squashed.
@richardkchapman ready for merge.

PR #13060, opened 2019-10-07
HPCC-22948 Fire record format error if mismatch and can't translate

jakesmith:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/thorlcr/activities/diskread/thdiskread.cpp
+++ b/thorlcr/activities/diskread/thdiskread.cpp
@@ -44,8 +44,31 @@ public:
             Owned<IException> e = MakeActivityWarning(&container, TE_GroupMismatch, "DFS and code generated group info. differs: DFS(%s), CodeGen(%s), using DFS info", isGrouped?"grouped":"ungrouped", codeGenGrouped?"grouped":"ungrouped");
             queryJobChannel().fireException(e);
         }
-        if (RecordTranslationMode::None == getTranslationMode(*this))
+        RecordTranslationMode translationMode = getTranslationMode(*this);
+        if (RecordTranslationMode::None != translationMode)
         {
+            /* Turn off translation if not "flat" or not RecordTranslationMode::AlwaysDisk
+             * NB: translation is turned off in slaves in this case too,
+             * turning off here, ensures format crc's are checked and any mismatch spotted
+             * with sensible error.
+             */
+
+            const char *kind = queryFileKind(file);
+            if (!strisame(kind, "flat") && (RecordTranslationMode::AlwaysDisk != translationMode))
+                translationMode = RecordTranslationMode::None;
+        }
+        if (RecordTranslationMode::None == translationMode)
+        {
+            unsigned dfsCrc;
+            if (file->getFormatCrc(dfsCrc) && helper->getDiskFormatCrc() != dfsCrc)
+            {
+                VStringBuffer fileStr("File: %s", file->queryLogicalName());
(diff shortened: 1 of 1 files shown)


richardkchapman:
@mckellyln Please review

thorlcr/activities/diskread/thdiskread.cpp
```diff
@@ -44,8 +44,31 @@ class CDiskReadMasterVF : public CDiskReadMasterBase
...
+            if (!strisame(kind, "flat") && (RecordTranslationMode::AlwaysDisk != translationMode))
+                translationMode = RecordTranslationMode::None;
+        }
+        if (RecordTranslationMode::None == translationMode)
+        {
+            unsigned dfsCrc;
+            if (file->getFormatCrc(dfsCrc) && helper->getDiskFormatCrc() != dfsCrc)
+            {
+                VStringBuffer fileStr("File: %s", file->queryLogicalName());
+                Owned<IThorException> e = MakeActivityException(this, TE_FormatCrcMismatch, "Automatic translation not possible, and layout does not match published layout. %s", fileStr.str());
```
mckellyln: trivial: perhaps "Automatic translation not enabled or possible" ?

richardkchapman: Automatic translation not available?

thorlcr/activities/diskread/thdiskread.cpp
```diff
@@ -44,8 +44,31 @@ class CDiskReadMasterVF : public CDiskReadMasterBase
...
+                if (!queryContainer().queryJob().getWorkUnitValueInt("skipFileFormatCrcCheck", 0))
+                    throw e.getClear();
+                e->setAction(tea_warning);
+                fireException(e);
```
mckellyln: If we do not throw() above then should we fireException here ?  Maybe yes because it is just tea_warning ?

mckellyln:
Looks fine.
I suppose it should be documented that translation will not occur for non-flat files unless its set to AlwaysDisk ?

thorlcr/activities/diskread/thdiskread.cpp
mckellyln: Yes that fits a little better.

jakesmith: Agree, will change "possible" to "available"

thorlcr/activities/diskread/thdiskread.cpp
jakesmith: yes, this is the mechanism used to propagate the warning to the workunit.

jakesmith:
@JamesDeFabia - is there existing documentation re. record translation capabilities ?
It would be adding some notes to say that:
When reading flat files, translation will only occur if the file type matches (i.e. it is reading a file that has been published as "flat"), or if translation from the published format is forced with "AlwaysDisk" (or "disk").

NB: Unless the flat file is fixed width, you'll get a file type mismatch error before this point if the file it's trying to read has been published as a different type (e.g. "csv"), since it's not possible to translate a variable length record format from another format.

jakesmith:
@mckellyln - updated error message from "possible" to "available".

mckellyln:
Looks good to merge.

mckellyln (approved):
@jakesmith approved.

PR #13064, opened 2019-10-08
HPCC-22955 Add caching of public/private key files

RussWhitehead:
Crypto Plugin should cache public/private key files,
since it is costly to load them every time

## Type of change:
- [ x] This change improves the code (refactor or other change that does not change the functionality)

## Checklist:
- [ x] My code follows the code style of this project.
  - [x ] My code does not create any new warnings from compiler, build system, or lint.
- [x ] The commit message is properly formatted and free of typos.
  - [x ] The commit message title makes sense in a changelog, by itself.
  - [x ] The commit is signed.
- [x ] I have read the CONTRIBUTORS document.
- [x ] The change has been fully tested:
  - [x ] All new and existing tests passed.
- [x ] I have given due consideration to all of the following potential concerns:
  - [x ] Scalability
  - [ x] Performance
  - [ ] Security
  - [ x] Thread-safety
  - [ ] Premature optimization
- [ x] There are no similar instances of the same problem that should be addressed

## Smoketest:


RussWhitehead:
@rpastrana Please review

jakesmith:
These may not be an immediate concerns, but:
+ What happens if we need to rollover the keys from time to time, i.e. load new versions?
+ Should there be a way to expunge a cached key?
+ Should there be a configurable lifetime? (e.g. if cache hit, but key older than threshold, then throws away and reloads)

jakesmith:
Ah ok, I had assumed for general use, and I think it should be so that others can use it too.
Esp and dafilesrv will currently be reloading the keys frequently, so it would be good to put this cache in front of those use cases too (but using the cache can be a separate PR).

RussWhitehead:
Opened https://track.hpccsystems.com/browse/HPCC-22957 to look into caching for Cryptohelper library.
@richardkchapman Please merge

plugins/cryptolib/cryptolib.cpp
```diff
@@ -527,6 +527,54 @@ CRYPTOLIB_API bool CRYPTOLIB_CALL clPKIVerifySignatureBuff(const char * pkalgori
...
+{
+private:
+    CriticalSection csKeyCache;//guards serial modifications to the cache map
+    typedef std::unordered_map<string, Owned<CLoadedKey>> KeyCache;
+    KeyCache keyCache;
+
+public:
+    CLoadedKey * getInstance(bool isPublic, const char * keyFS, const char * keyBuff, const char * passphrase)//
+    {
+        VStringBuffer searchKey("%s_%s_%s", isEmptyString(keyFS) ? "" : keyFS, isEmptyString(keyBuff) ? "" : keyBuff, isEmptyString(passphrase) ? "" : passphrase);
```
rpastrana: Shouldn't we bail out and avoid entering the CritBlock if keyFS and KeyBuff are empty?

plugins/cryptolib/cryptolib.cpp
```diff
@@ -527,6 +527,54 @@ CRYPTOLIB_API bool CRYPTOLIB_CALL clPKIVerifySignatureBuff(const char * pkalgori
...
+            ret = (*it).second;
+        }
+        else
+        {
+            CLoadedKey * pCLK = nullptr;
+            if (!isEmptyString(keyFS))
+            {
+                //Create CLoadedKey from filespec
+                if (isPublic)
+                    pCLK = loadPublicKeyFromFile(keyFS, passphrase);
```
rpastrana: it seems we could load the keys outside of the critblock and we should only protect the lookup and the insert

rpastrana (changes requested):
@RussWhitehead left a few comments

plugins/cryptolib/cryptolib.cpp
```diff
@@ -527,6 +527,54 @@ CRYPTOLIB_API bool CRYPTOLIB_CALL clPKIVerifySignatureBuff(const char * pkalgori
...
+//-----------------------------------------------------------------
+class CKeyCache
+{
+private:
+    CriticalSection csKeyCache;//guards serial modifications to the cache map
+    typedef std::unordered_map<string, Owned<CLoadedKey>> KeyCache;
+    KeyCache keyCache;
+
+public:
+    CLoadedKey * getInstance(bool isPublic, const char * keyFS, const char * keyBuff, const char * passphrase)//
```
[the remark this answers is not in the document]
RussWhitehead: Good catch !

plugins/cryptolib/cryptolib.cpp
RussWhitehead: Done

RussWhitehead:
@rpastrana Please rereview

rpastrana (approved):
@RussWhitehead approved

richardkchapman:
This should target 7.6.x (or possibly master), not 7.6.2

richardkchapman:
Checkboxes are messed up.

richardkchapman:
@ghalliday Thoughts? Security implications?

plugins/cryptolib/cryptolib.cpp
```diff
@@ -570,7 +620,7 @@ CRYPTOLIB_API void CRYPTOLIB_CALL clPKIEncrypt(size32_t & __lenResult,void * & _
                                             size32_t lenInputdata,const void * inputdata)
 {
     verifyPKIAlgorithm(pkalgorithm);
-    Owned<CLoadedKey> publicKey = loadPublicKeyFromFile(publickeyfile, passphrase);
+    Owned<CLoadedKey> publicKey = g_KeyCache.getInstance(true, publickeyfile, nullptr, passphrase);
```
ghalliday: Are there any threading issues with this keys being shared between multiple callers?  I would expect it to work, but double checking...

plugins/cryptolib/cryptolib.cpp
```diff
@@ -527,6 +527,56 @@ CRYPTOLIB_API bool CRYPTOLIB_CALL clPKIVerifySignatureBuff(const char * pkalgori
...
+            else
+            {
+                //Create CLoadedKey from key contents
+                if (isPublic)
+                    pCLK = loadPublicKeyFromMemory(keyBuff, passphrase);
+                else
+                    pCLK = loadPrivateKeyFromMemory(keyBuff, passphrase);
+            }
+
+            keyCache.insert(pair<string, CLoadedKey*>(searchKey.str(), pCLK));
```
ghalliday: For efficient to use emplace, and you can use a pair of the matching type then.

ghalliday (commented):
@RussWhitehead a couple of minor comments.  Did you see any performance boost with the change in place?

PR #13072, opened 2019-10-09
HPCC-22960 Check mismatched file type reads

jakesmith:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/thorlcr/thorutil/thormisc.cpp
+++ b/thorlcr/thorutil/thormisc.cpp
@@ -805,7 +805,15 @@ void reportExceptionToWorkunit(IConstWorkUnit &workunit,IException *e, ErrorSeve
         else
             we->setSeverity(severity);
     }
-} 
+}
+
+void reportExceptionToWorkunitCheckIgnore(IConstWorkUnit &workunit, IException *e, ErrorSeverity severity)
+{
+    ErrorSeverity mappedSeverity = workunit.getWarningSeverity(e->errorCode(), severity);
+    if (SeverityIgnore == mappedSeverity)
+        return;
+    reportExceptionToWorkunit(workunit, e, mappedSeverity);
+}
 
 StringBuffer &getCompoundQueryName(StringBuffer &compoundName, const char *queryName, unsigned version)
 {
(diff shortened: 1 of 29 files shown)


jakesmith:
@ghalliday please review.

NB: I have not changed NewDiskRead, that may need revisiting for similar checks when it is generalized to handle csv, xml etc.

ghalliday:
Was this triggered by a particular problem that was hit?  I'm wondering how this change would be generalized when we have pluggable file formats.

jakesmith:
I user hit it, incorrectly trying to read a CSV as a FLAT file.
They targeted Thor, which would have given a mismatch error, but because it was a simple test and they weren't using pickBestEngine=false, it ran on HThor and failed with a deserialization error, which didn't mean much to them.

jakesmith:
>I'm wondering how this change would be generalized when we have pluggable file formats.

In general there should be a match.
But we have a couple of exception at the moment:

1) Reading any file format as a CSV
2) Reading any file format if FLAT and fixed width.
NB: it won't try to translate by default unless file type is also flat.

Perhaps new pluggable file formats might lend themselves to read arbitrary data too, in same way as CSV can be applied. So perhaps whether strict (must match read type) or not should be an option of the pluggable file reader interface.

jakesmith:
There was a couple of issues exposed by the regression.
One was that there the warning reported to the workunit wasn't consistent, both in terms of error number and severity level (should have been warning). One of the tests that was failing had a commented out #onwarning, probably because it looked like it was necessary, but the warning was being reported as info only and not in the output.

I've made these warnings consistent now.
I've opened a separate JIRA re. consolidating error code numbers in 1 place for all engines for common errors like this: https://track.hpccsystems.com/browse/HPCC-22976

jakesmith:
@ghalliday - please re-review (perhaps after regression suite passes)

thorlcr/thorutil/thormisc.cpp
```diff
@@ -805,7 +805,15 @@ void reportExceptionToWorkunit(IConstWorkUnit &workunit,IException *e, ErrorSeve
...
     }
-} 
+}
+
+void reportExceptionToWorkunitCheckIgnore(IConstWorkUnit &workunit, IException *e, ErrorSeverity severity)
+{
+    ErrorSeverity mappedSeverity = workunit.getWarningSeverity(e->errorCode(), SeverityWarning);
+    if (SeverityIgnore == mappedSeverity)
+        return;
+    reportExceptionToWorkunit(workunit, e, severity);
```
ghalliday: Should this pass mappedSeverity instead of severity?

ghalliday:
@jakesmith one minor question

thorlcr/thorutil/thormisc.cpp
jakesmith: agree, have changed, and squashed change in.

jakesmith:
@ghalliday - see addressed comment.

PR #13074, opened 2019-10-09
HPCC-22955 Add caching of public/private key files

RussWhitehead:
Crypto Plugin should cache public/private key files,
since it is costly to load them every time

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [ ] Security
  - [x] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/plugins/cryptolib/CMakeLists.txt
+++ b/plugins/cryptolib/CMakeLists.txt
@@ -33,6 +33,8 @@ include_directories (
          ${HPCC_SOURCE_DIR}/system/include
          ${HPCC_SOURCE_DIR}/system/jlib
          ${HPCC_SOURCE_DIR}/rtl/include
+         ${HPCC_SOURCE_DIR}/dali/base
+         ${HPCC_SOURCE_DIR}/system/mp
          ${HPCC_SOURCE_DIR}/system/security/cryptohelper
          ${OPENSSL_INCLUDE_DIR}
     )
--- a/plugins/cryptolib/cryptolib.cpp
+++ b/plugins/cryptolib/cryptolib.cpp
@@ -611,7 +611,6 @@ static void doPKIDecrypt(size32_t & __lenResult,void * & __result,
 }
 
 
-
 //encryption functions that take filespecs of key files
 CRYPTOLIB_API void CRYPTOLIB_CALL clPKIEncrypt(size32_t & __lenResult,void * & __result,
                                             const char * pkalgorithm,
--- a/plugins/cryptolib/cryptolib.cpp
+++ b/plugins/cryptolib/cryptolib.cpp
@@ -19,6 +19,7 @@
 #include "jexcept.hpp"
 #include "digisign.hpp"
 #include "ske.hpp"
+#include "dasess.hpp"
 #include <openssl/pem.h>
 #include <openssl/err.h>
 #include <openssl/rand.h>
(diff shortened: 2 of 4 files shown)


RussWhitehead:
@richardkchapman Your concern about a rogue ECL query specifying the PKE file that is already cached seems valid and we should discuss
@ghalliday The documentation for OpenSSL is bad, but in searches I could find anything that indicates these CLoadedKey files cannot be shared. Furthermore my testing did not find any problems.
Please rereview

plugins/cryptolib/CMakeLists.txt
```diff
@@ -33,6 +33,8 @@ include_directories (
          ${HPCC_SOURCE_DIR}/system/include
          ${HPCC_SOURCE_DIR}/system/jlib
          ${HPCC_SOURCE_DIR}/rtl/include
+         ${HPCC_SOURCE_DIR}/dali/base
```
ghalliday: These dependencies don't look right.  Why are these now required?

ghalliday:
I'm not sure what benefit is of adding the user to the hash key.  If they passed the same key and passphrase then they will load the key anyway, so I don't think it will provide any extra protection.  On the other hand it is likely to be the same user on calls, so won't make the caching worse.

The main concern would be over whether caching the keys in memory makes it more likely that they could be extracted from the running process.

ghalliday:
@RussWhitehead I forgot to mention you to reassign.

RussWhitehead:
@richardkchapman @ghalliday @jakesmith 
I believe the real problem with this implementation is that, because the key file is opened and read by the execution engine (HPCC User), ANY user that can run a workunit can access any key file accessible to HPCC User. The changes I introduced above will not fix this problem. 
Not sure how to fix this other than removing the ability for an ECL user to specify a filespec, and force them to manage the key files with their own confidential store.
Comments?

jakesmith:
The issue of being able to point to a physical file, and for any user to be able to access it,
yes that's an issue, but not really related to this PR.
That's a problem already (without caching being added).

Either specifying a file path should be prohibited, or it should be validated for authority.
We _could_ turn it into a file::<location> sepc., which can already be tied down via LDAP scope checking.
Or similarly, we could have a separate scheme which allows named key pairs to be referenced and fetched (after passing LDAP checks).

Not sure if WsStore could fit in here somewhere(?) - could it allow groups to share secrets? @rpastrana ?

@RussWhitehead - there should be a separate JIRA covering this issue as not directly to do with this new caching. Can you open a separate JIRA to discuss further?

Re. caching - I think there was concern that it opened up another vulnerability, in that the keys are sitting in memory and in theory an unauthorized user could use them.

Was the main incentive behind the key caching?
Is there 1 area that particularly gains from this time saving?

rpastrana:
Re: WsStore group based restrictions. It wasn't in the design, but it sounds like a reasonable idea. The service would be dependent on the ESP/security manager to provide group info. @RussWhitehead  @jakesmith

jakesmith:
Further discussion is needed in the new JIRA specifically for this issue (of securely getting/sharing secrets/keys), @RussWhitehead - if u can open one.

Re. caching - we can continue to discuss here, @RussWhitehead can u comment on the questions above re. caching.

RussWhitehead:
@jakesmith  Opened https://track.hpccsystems.com/browse/HPCC-23082 to address key file Accessibility

RussWhitehead:
@ghalliday @jakesmith This change restricts the cached key files to the user/key/passphrase that originally referenced it. Any user that subsequently tries to reference the same key file will have to go through the scope check.  Once we add key file scoping, this will make even more sense.

@ghalliday You are correct, having these files cached in memory would make them accessible to snooping. We should discuss the tradeoffs between run time security and performance

plugins/cryptolib/CMakeLists.txt
RussWhitehead: @ghalliday These dependencies were introduced with adding the ICodeContext to the plugin API call

ghalliday:
@RussWhitehead back to you to follow up on the discussion and switch to using thread local variables for the cache.

RussWhitehead:
Replaced by https://github.com/hpcc-systems/HPCC-Platform/pull/13204

PR #13076, opened 2019-10-10
HPCC-22978 Make it easier to extract an archive from a compiled dll

richardkchapman:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

## Testing:
Used to get archive from xml I was investigating...

### DIFF
--- a/tools/wuget/wuget.cpp
+++ b/tools/wuget/wuget.cpp
@@ -16,6 +16,7 @@
 ############################################################################## */
 
 #include "jlib.hpp"
+#include "jfile.hpp"
 #include "thorplugin.hpp"
 #include "workunit.hpp"
 
@@ -25,6 +26,7 @@ void usage(const char *progname)
     printf("Options:\n");
     printf("  -m     Display manifest information\n");
     printf("  -w     Display embedded workunit\n");
+    printf("  -a     Display embedded archive\n");
     printf("/nIf no options specified, -w is assumed\n");
 }
 
@@ -33,6 +35,7 @@ int main(int argc, char **argv)
     InitModuleObjects();
     bool doManifestInfo = false;
     bool doWorkunit = false;
+    bool doArchive = false;
     for (int i = 1; i < argc; i++)
     {
         if (argv[i][0]=='-')
@@ -41,6 +44,8 @@ int main(int argc, char **argv)
                 doManifestInfo = true;
             else if (strcmp(argv[i], "-w")==0)
                 doWorkunit = true;
+            else if (strcmp(argv[i], "-a")==0)
+                doArchive = true;
             else
             {
                 usage(argv[0]);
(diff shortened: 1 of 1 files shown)


tools/wuget/wuget.cpp
```diff
@@ -57,14 +64,30 @@ int main(int argc, char **argv)
...
             {
                 filesSeen++;
-                if (doWorkunit || !doManifestInfo)
+                if (doWorkunit || doArchive)
                 {
+                    const char *filename = argv[i];
+                    const char *ext = pathExtension(filename);
                     StringBuffer xml;
-                    if (getWorkunitXMLFromFile(argv[i], xml))
+                    if (ext && strieq(ext, ".xml"))
```
jakesmith: trivial: could be:
```
 if (strisame(ext, ".xml"))
```

tools/wuget/wuget.cpp
```diff
@@ -57,14 +64,30 @@ int main(int argc, char **argv)
...
                     {
                         Owned<ILocalWorkUnit> wu = createLocalWorkUnit(xml);
-                        exportWorkUnitToXML(wu, xml.clear(), true, false, true);
-                        printf("%s\n", xml.str());
+                        if (doWorkunit)
+                        {
+                            exportWorkUnitToXML(wu, xml.clear(), true, false, true);
+                            printf("%s\n", xml.str());
+                        }
+                        if (doArchive)
```
jakesmith: can't be true if doWorkunit is true, so could be:
```
else if (doArchive)

File: tools/wuget/wuget.cpp [hunk 3]
```diff
@@ -57,14 +64,30 @@ int main(int argc, char **argv)
...
+                        {
+                            Owned<IConstWUQuery> query=wu->getQuery();
```
jakesmith: pedantic/formatting: space either side of = , to be consistent

jakesmith (approved):

jakesmith:
@richardkchapman - looks fine, just some trivial/pedantic comments.

tools/wuget/wuget.cpp
richardkchapman: I think it can, if someone specified -w -a (because they wanted both). May be unusual usage but...

jakesmith: it can't at the moment, because it is either or in the parameter passing:
```
            else if (strcmp(argv[i], "-w")==0)
                doWorkunit = true;
            else if (strcmp(argv[i], "-a")==0)
                doArchive = true;
```

richardkchapman: Which is inside a for loop...

jakesmith: sorry yeah misreading.

richardkchapman:
@ghalliday I force-pushed to address Jake's concerns

PR #13091, opened 2019-10-21
HPCC-22908 Return cached Activity while rebuilding new state

wangkx:
1. ESP begins to build Activity cache when ESP starts and is
attached to dali.
2. User always gets current cached copy back instantly if the
cache is available.
3. If user asks the Activity info before the 1st cache is
available, ESP returns the cache after the cache is available.
4. When idle, ESP automaticaly rebuilds the cache if a pre-
configured timeout exceeds and swaps in new version when ready.
5. If user makes a request and the cache is older than a pre-
configured timeout, ESP returns the cache and informs the cache
builder to refresh the cache.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/services/ws_smc/ws_smcService.cpp
+++ b/esp/services/ws_smc/ws_smcService.cpp
@@ -55,7 +55,6 @@ static const char* SMC_ACCESS_DENIED = "Access Denied";
 static const char* QUEUE_ACCESS_DENIED = "Failed to access the queue functions. Permission denied.";
 
 const char* PERMISSIONS_FILENAME = "espsmc_permissions.xml";
-const unsigned DEFAULTACTIVITYINFOCACHETIMEOUTSECOND = 10;
 
 void AccessSuccess(IEspContext& context, char const * msg,...) __attribute__((format(printf, 2, 3)));
 void AccessSuccess(IEspContext& context, char const * msg,...)
@@ -142,10 +141,14 @@ void CWsSMCEx::init(IPropertyTree *cfg, const char *process, const char *service
         m_PortalURL.append(portalURL);
 
     xpath.setf("Software/EspProcess[@name=\"%s\"]/EspService[@name=\"%s\"]/ActivityInfoCacheSeconds", process, service);
-    activityInfoCacheSeconds = cfg->getPropInt(xpath.str(), DEFAULTACTIVITYINFOCACHETIMEOUTSECOND);
+    unsigned activityInfoCacheSeconds = cfg->getPropInt(xpath.str(), DEFAULTACTIVITYINFOCACHEFORCEBUILDSECOND);
     xpath.setf("Software/EspProcess[@name=\"%s\"]/EspService[@name=\"%s\"]/LogDaliConnection", process, service);
     if (cfg->getPropBool(xpath.str()))
         querySDS().setConfigOpt("Client/@LogConnection", "true");
+
+    xpath.setf("Software/EspProcess[@name=\"%s\"]/EspService[@name=\"%s\"]/ActivityInfoCacheAutoRebuildSeconds", process, service);
+    unsigned activityInfoCacheAutoRebuildSeconds = cfg->getPropInt(xpath.str(), DEFAULTACTIVITYINFOCACHEAUTOREBUILDSECOND);
+    activityInfoReader.setown(new CActivityInfoReader(activityInfoCacheAutoRebuildSeconds, activityInfoCacheSeconds));
 }
 
 struct CActiveWorkunitWrapper: public CActiveWorkunit
@@ -1194,7 +1197,9 @@ bool CWsSMCEx::onActivity(IEspContext &context, IEspActivityRequest &req, IEspAc
         if (version >= 1.06)
             setBannerAndChatData(version, resp);
 
-        Owned<CActivityInfo> activityInfo = getActivityInfo(context);
+        Owned<CActivityInfo> activityInfo = activityInfoReader->getActivityInfo();
+        if (!activityInfo)
+            throw MakeStringException(ECLWATCH_INTERNAL_ERROR, "Failed to get Activity Info. Please try later.");
         setActivityResponse(context, activityInfo, req, resp);
     }
     catch(IException* e)
@@ -1203,29 +1208,6 @@ bool CWsSMCEx::onActivity(IEspContext &context, IEspActivityRequest &req, IEspAc
     }
 
     return true;
-}
-
-void CWsSMCEx::clearActivityInfoCache()
-{
-    CriticalBlock b(getActivityCrit);
-    activityInfoCache.clear();
-}
-
-CActivityInfo* CWsSMCEx::getActivityInfo(IEspContext &context)
-{
-    CriticalBlock b(getActivityCrit);
-
-    if (activityInfoCache && activityInfoCache->isCachedActivityInfoValid(activityInfoCacheSeconds))
-        return activityInfoCache.getLink();
-
-    DBGLOG("CWsSMCEx::getActivityInfo - rebuild cached information");
-    {
-        EspTimeSection timer("createActivityInfo");
-        activityInfoCache.setown(new CActivityInfo());
-        activityInfoCache->createActivityInfo(context);
-    }
-
-    return activityInfoCache.getLink();
 }
 
 void CWsSMCEx::addWUsToResponse(IEspContext &context, const IArrayOf<IEspActiveWorkunit>& aws, IEspActivityResponse& resp)
@@ -1279,6 +1261,12 @@ void CWsSMCEx::setActivityResponse(IEspContext &context, CActivityInfo* activity
     double version = context.getClientVersion();
     const char* sortBy = req.getSortBy();
     bool descending = req.getDescending();
+    if (version >= 1.22)
+    {
+        StringBuffer s;
+        resp.setActivityTime(activityInfo->queryTimeCached(s));
+        resp.setDaliDetached(activityInfoReader->isDaliDetached());
+    }
     if (version >= 1.16)
     {
         IArrayOf<IEspTargetCluster> thorClusters;
@@ -1419,7 +1407,7 @@ bool CWsSMCEx::onMoveJobDown(IEspContext &context, IEspSMCJobRequest &req, IEspS
             }
         }
         AccessSuccess(context, "Changed job priority %s",req.getWuid());
-        clearActivityInfoCache();
+        activityInfoReader->rebuild();
         resp.setRedirectUrl("/WsSMC/");
     }
     catch(IException* e)
@@ -1448,7 +1436,7 @@ bool CWsSMCEx::onMoveJobUp(IEspContext &context, IEspSMCJobRequest &req, IEspSMC
             }
         }
         AccessSuccess(context, "Changed job priority %s",req.getWuid());
-        clearActivityInfoCache();
+        activityInfoReader->rebuild();
         resp.setRedirectUrl("/WsSMC/");
     }
     catch(IException* e)
@@ -1494,7 +1482,7 @@ bool CWsSMCEx::onMoveJobBack(IEspContext &context, IEspSMCJobRequest &req, IEspS
             }
         }
         AccessSuccess(context, "Changed job priority %s",req.getWuid());
-        clearActivityInfoCache();
+        activityInfoReader->rebuild();
         resp.setRedirectUrl("/WsSMC/");
     }
     catch(IException* e)
@@ -1541,7 +1529,7 @@ bool CWsSMCEx::onMoveJobFront(IEspContext &context, IEspSMCJobRequest &req, IEsp
         }
 
         AccessSuccess(context, "Changed job priority %s",req.getWuid());
-        clearActivityInfoCache();
+        activityInfoReader->rebuild();
         resp.setRedirectUrl("/WsSMC/");
     }
     catch(IException* e)
@@ -1571,7 +1559,7 @@ bool CWsSMCEx::onRemoveJob(IEspContext &context, IEspSMCJobRequest &req, IEspSMC
             }
         }
         AccessSuccess(context, "Removed job %s",req.getWuid());
-        clearActivityInfoCache();
+        activityInfoReader->rebuild();
         resp.setRedirectUrl("/WsSMC/");
     }
     catch(IException* e)
@@ -1593,7 +1581,7 @@ bool CWsSMCEx::onStopQueue(IEspContext &context, IEspSMCQueueRequest &req, IEspS
             queue->stop(createQueueActionInfo(context, "stopped", req, info));
         }
         AccessSuccess(context, "Stopped queue %s", req.getCluster());
-        clearActivityInfoCache();
+        activityInfoReader->rebuild();
         double version = context.getClientVersion();
         if (version >= 1.19)
             getStatusServerInfo(context, req.getServerType(), req.getCluster(), req.getNetworkAddress(), req.getPort(), resp.updateStatusServerInfo());
@@ -1619,7 +1607,7 @@ bool CWsSMCEx::onResumeQueue(IEspContext &context, IEspSMCQueueRequest &req, IEs
             queue->resume(createQueueActionInfo(context, "resumed", req, info));
         }
         AccessSuccess(context, "Resumed queue %s", req.getCluster());
-        clearActivityInfoCache();
+        activityInfoReader->rebuild();
         double version = context.getClientVersion();
         if (version >= 1.19)
             getStatusServerInfo(context, req.getServerType(), req.getCluster(), req.getNetworkAddress(), req.getPort(), resp.updateStatusServerInfo());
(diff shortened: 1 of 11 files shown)


wangkx:
@jakesmith please review.

esp/services/ws_smc/ws_smcService.cpp
```diff
@@ -142,10 +141,15 @@ void CWsSMCEx::init(IPropertyTree *cfg, const char *process, const char *service
...
-    activityInfoCacheSeconds = cfg->getPropInt(xpath.str(), DEFAULTACTIVITYINFOCACHETIMEOUTSECOND);
+    unsigned activityInfoCacheSeconds = cfg->getPropInt(xpath.str(), DEFAULTACTIVITYINFOCACHEFORCEBUILDSECOND);
     xpath.setf("Software/EspProcess[@name=\"%s\"]/EspService[@name=\"%s\"]/LogDaliConnection", process, service);
     if (cfg->getPropBool(xpath.str()))
         querySDS().setConfigOpt("Client/@LogConnection", "true");
+
+    xpath.setf("Software/EspProcess[@name=\"%s\"]/EspService[@name=\"%s\"]/ActivityInfoCacheAutoRebuildSeconds", process, service);
+    unsigned activityInfoCacheAutoRebuildSeconds = cfg->getPropInt(xpath.str(), DEFAULTACTIVITYINFOCACHEAUTOREBUILDSECOND);
+    activityInfoReader.setown(new CActivityInfoReader(activityInfoCacheAutoRebuildSeconds * 1000, activityInfoCacheSeconds));
+    activityInfoReader->start();
```
jakesmith: see comment re. contained thread.

esp/services/ws_smc/ws_smcService.cpp
```diff
@@ -2621,3 +2605,50 @@ bool CWsSMCEx::onLockQuery(IEspContext &context, IEspLockQueryRequest &req, IEsp
...
+    while (!stopping)
+    {
+        if (!detached)
+        {
+            try
+            {
+                if (waitTimeMillies == (unsigned)-1)
+                {
+                    PROGLOG("WsSMC CActivityInfoReader Thread Re-started.");
+                    waitTimeMillies = autoRebuildMillSeconds;
```
jakesmith: There looks like there might be are potential race conditions between something setting detached and the timer being set.
if gets !detached just before here, then it could wait indefinitely.

You don't really to change the timer, cycling around ever couple minutes (due to default auto rebuild timeout) to recheck detached would be fine, and code would be simpler.

esp/services/ws_smc/ws_smcService.cpp
```diff
@@ -2621,3 +2605,50 @@ bool CWsSMCEx::onLockQuery(IEspContext &context, IEspLockQueryRequest &req, IEsp
...
+            {
+                IERRLOG("Unknown exception in WsSMC CActivityInfoReader::run");
+            }
+        }
+        else
+        {
+            OWARNLOG("Detached from DALI, WsSMC CActivityInfoReader interrupted");
+            waitTimeMillies = (unsigned)-1;
+        }
+        sem.wait(waitTimeMillies);
```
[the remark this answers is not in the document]
jakesmith: See comment(s) elsewhere, you need to be careful you're not signalling too much, meaning there could be bunch of pending signals and this keeps rebuilding when unnecessary.
It should only signal / rebuild it not doing so already.

I think the simplest/safest way might be to change code here to use an atomic like this:
```
        waiting = true;
        if (!sem.wait(autoRebuildMillSeconds)
        {
            bool expected = true;
            waiting.compare_exchange_strong(expected, false);
        }
```

then similar waiting.exchange code in rebuild() to only signal is waiting.

esp/services/ws_smc/ws_smcService.hpp
```diff
@@ -154,15 +157,69 @@ class CActivityInfo : public CInterface, implements IInterface
...
+    }
+
+    virtual int run();
+    CActivityInfo* getActivityInfo()
+    {
+        while (!stopping)
+        {
+            {
+                CriticalBlock b(crit);
+                if (activityInfoCache)
```
jakesmith: see other comment below (after line 199)
If run has been changed to block if no existing activityInfoCache entry, then you can remove condition here (just assert), you can also remove the sleep and loop.

jakesmith (changes requested):

jakesmith:
@wangkx - looks good in general, but the signal/blocking needs looking at.

esp/services/ws_smc/ws_smcService.hpp
```diff
@@ -154,15 +157,69 @@ class CActivityInfo : public CInterface, implements IInterface
...
+    Owned<CActivityInfo> activityInfoCache;
+    Semaphore sem;
+    CriticalSection crit;
+public:
+    CActivityInfoReader(unsigned _autoRebuildMillSeconds, unsigned _forceRebuildSeconds)
+        : autoRebuildMillSeconds(_autoRebuildMillSeconds), forceRebuildSeconds(_forceRebuildSeconds) {};
+
+    ~CActivityInfoReader()
+    {
+        stopping = true;
```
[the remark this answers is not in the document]
wangkx: Will try the CThreaded.

esp/services/ws_smc/ws_smcService.hpp
```diff
@@ -154,15 +157,69 @@ class CActivityInfo : public CInterface, implements IInterface
...
+};
+
+class CWsSMCEx;
+
+class CActivityInfoReader : public Thread
+{
+    bool stopping = false;
+    bool detached = false;
+    unsigned autoRebuildMillSeconds = 1000*DEFAULTACTIVITYINFOCACHEAUTOREBUILDSECOND;
+    unsigned forceRebuildSeconds = DEFAULTACTIVITYINFOCACHEFORCEBUILDSECOND;
```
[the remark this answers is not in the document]
wangkx: Will change the ms to seconds.

esp/services/ws_smc/ws_smcService.hpp
```diff
@@ -154,15 +157,69 @@ class CActivityInfo : public CInterface, implements IInterface
...
+            {
+                CriticalBlock b(crit);
+                if (activityInfoCache)
+                {
+                    if (!activityInfoCache->isCachedActivityInfoValid(forceRebuildSeconds))
+                        rebuild();
+                    return activityInfoCache.getLink();
+                }
+            } //release crit so that the activityInfoCache can be set.
+            sleep(1);
```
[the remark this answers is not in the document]
wangkx: Yes, it is in case 1st time and not yet built.

esp/services/ws_smc/ws_smcService.hpp
```diff
@@ -196,6 +254,18 @@ class CWsSMCEx : public CWsSMC
     bool onSetBanner(IEspContext &context, IEspSetBannerRequest &req, IEspSetBannerResponse& resp);
     bool onNotInCommunityEdition(IEspContext &context, IEspNotInCommunityEditionRequest &req, IEspNotInCommunityEditionResponse &resp);
 
+    bool attachServiceToDali() override
```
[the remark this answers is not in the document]
wangkx: Will add the 'virtual'.

esp/services/ws_smc/ws_smcService.hpp
```diff
@@ -154,15 +157,69 @@ class CActivityInfo : public CInterface, implements IInterface
...
+    void setDetachedState(bool _detached)
+    {
+        if (detached != _detached)
+        {
+            detached = _detached;
+            if (!detached)
+                rebuild();
+        }
+    }
+    bool isDaliDetached() { return detached; }
```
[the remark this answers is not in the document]
wangkx: Will change it to const method.

esp/services/ws_smc/ws_smcService.cpp
```diff
@@ -2255,10 +2242,7 @@ void CWsSMCEx::getStatusServerInfo(IEspContext &context, const char *serverType,
     if (!serverType || !*serverType)
         throw MakeStringException(ECLWATCH_MISSING_PARAMS, "Server type not specified.");
 
-    Owned<CActivityInfo> activityInfo = getActivityInfo(context);
-    if (!activityInfo)
-        throw MakeStringException(ECLWATCH_INTERNAL_ERROR, "Failed to get Activity Info cache.");
-
+    Owned<CActivityInfo> activityInfo = activityInfoReader->getActivityInfo();
```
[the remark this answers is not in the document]
wangkx: Will add a check. If null, returns.

esp/services/ws_smc/ws_smcService.cpp
```diff
@@ -2621,3 +2605,50 @@ bool CWsSMCEx::onLockQuery(IEspContext &context, IEspLockQueryRequest &req, IEsp
...
+                PROGLOG("WsSMC CActivityInfoReader: collecting ActivityInfo.");
+                EspTimeSection timer("createActivityInfo");
+                Owned<IEspContext> espContext =  createEspContext();
+                Owned<CActivityInfo> activityInfo = new CActivityInfo();
+                activityInfo->createActivityInfo(*espContext);
+                PROGLOG("WsSMC CActivityInfoReader: ActivityInfo collected.");
+
+                CriticalBlock b(crit);
+                activityInfoCache.setown(activityInfo.getClear());
+                PROGLOG("WsSMC CActivityInfoReader: ActivityInfo cached.");
```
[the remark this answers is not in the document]
wangkx: Will remove  the 2nd and 4th.

esp/services/ws_smc/ws_smcService.hpp
```diff
@@ -154,15 +157,69 @@ class CActivityInfo : public CInterface, implements IInterface
...
+    CActivityInfo* getActivityInfo()
+    {
+        while (!stopping)
+        {
+            {
+                CriticalBlock b(crit);
+                if (activityInfoCache)
+                {
+                    if (!activityInfoCache->isCachedActivityInfoValid(forceRebuildSeconds))
+                        rebuild();
```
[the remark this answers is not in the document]
wangkx: Will add a flag.

esp/services/ws_smc/ws_smcService.cpp
wangkx: Will remove this line.

esp/services/ws_smc/ws_smcService.cpp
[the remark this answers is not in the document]
wangkx: Thanks for the code. Will do.

esp/services/ws_smc/ws_smcService.cpp
jakesmith: You may still want a 'start' type method called here, but it won't be the Thread::start (as contained a CThread object)

esp/services/ws_smc/ws_smcService.hpp
wangkx: Will do.

esp/services/ws_smc/ws_smcService.cpp
wangkx: Will not change the timer.

esp/services/ws_smc/ws_smcService.hpp
```diff
@@ -154,15 +157,69 @@ class CActivityInfo : public CInterface, implements IInterface
...
+                    return activityInfoCache.getLink();
+                }
+            } //release crit so that the activityInfoCache can be set.
+            sleep(1);
+        }
+        return nullptr;
+    }
+    void rebuild()
+    {
+        sem.signal();
```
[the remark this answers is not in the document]
wangkx: Will check the waiting before calling the signal().

esp/services/ws_smc/ws_smcService.cpp
wangkx: Not sure what is needed in the 'start' type method. Will see.

wangkx:
@jakesmith Thanks for your review comments. I hope I understand all of them. Please review my changes.

esp/services/ws_smc/ws_smcService.cpp
```diff
@@ -2255,9 +2242,8 @@ void CWsSMCEx::getStatusServerInfo(IEspContext &context, const char *serverType,
     if (!serverType || !*serverType)
         throw MakeStringException(ECLWATCH_MISSING_PARAMS, "Server type not specified.");
 
-    Owned<CActivityInfo> activityInfo = getActivityInfo(context);
-    if (!activityInfo)
-        throw MakeStringException(ECLWATCH_INTERNAL_ERROR, "Failed to get Activity Info cache.");
+    Owned<CActivityInfo> activityInfo = activityInfoReader->getActivityInfo();
+    assert(activityInfo);
```
jakesmith: Should use assertex instead,
*but*, It could legitimately be null here, if stopping before ever got started, so you should not assert..

esp/services/ws_smc/ws_smcService.hpp
```diff
@@ -154,15 +157,67 @@ class CActivityInfo : public CInterface, implements IInterface
...
+        stopping = true;
+        sem.signal();
+        threaded.join();
+    }
+
+    virtual void threadmain() override;
+    CActivityInfo* getActivityInfo()
+    {
+        CriticalBlock b(crit);
+        assert(activityInfoCache);
```
jakesmith: in fact, it could still be null by this time, if the thread hasn't got started...
you could hold the crit before the thread starts, and release after 1st build.

esp/services/ws_smc/ws_smcService.hpp
```diff
@@ -154,15 +157,67 @@ class CActivityInfo : public CInterface, implements IInterface
...
+        if (waiting.compare_exchange_strong(expected, false))
+            sem.signal();
+    }
+    void setDetachedState(bool _detached)
+    {
+        if (detached != _detached)
+        {
+            detached = _detached;
+            if (!detached)
+                rebuild();
```
jakesmith: I don't think there's any point in signalling now..?
It it's blocked - then it will timeout and the thread will see !detached.
It it's not blocked - there's nothing to do.
So I think this code should simply set detached to false.

esp/services/ws_smc/ws_smcService.hpp
jakesmith: But also, what happens if detached and this method is called?
I think in 1st commit - it would loop continuously until re-attached.

What is the expected behaviour here when detached? Is there anything to prevent it being called?

jakesmith (changes requested):

jakesmith:
@wangkx - I think there are still some race conditions and undefined behaviour involving detached and calls to getActivityInfo.

esp/services/ws_smc/ws_smcService.hpp
jakesmith: yes I misread it as calling when detached which would be pointless.
If was detached and now isn't, yes sensible to signal.

If this can be called from multiple threads, it's not thread safe though.

esp/services/ws_smc/ws_smcService.hpp
jakesmith: so conclusion from offline chat, is that it could be detached at any time, including before the service actually starts, i.e. before any activity info is available.

- If detached and no activityInfoCache available, it must return null and callers need to handle that (not assert)

- If not detached it should block until there's an activityInfoCache instance available OR until detached becomes true.
You can could block it, by holding the crit early at CActivityInfoReader time and ensuring released after the 1st creation.

Alternatively, you could check if 'first' and only block if activityInfoCache is unavailable, until it is

esp/services/ws_smc/ws_smcService.hpp
wangkx: Should I add a new CriticalSection before line 209 to make it thread safe?

esp/services/ws_smc/ws_smcService.cpp
wangkx: Will throw an exception to inform ESP client that Activity Info is not available at this time.

esp/services/ws_smc/ws_smcService.hpp
wangkx: Got it

wangkx:
@jakesmith please review.

esp/services/ws_smc/ws_smcService.hpp
```diff
@@ -154,15 +157,94 @@ class CActivityInfo : public CInterface, implements IInterface
...
+        if (detached != _detached)
+        {
+            detached = _detached;
+            if (detached)
+            {
+                // NB: first will still be true, signal and getActivityInfo() will return null
+                if (firstBlocked)
+                {
+                    firstBlocked = false;
+                    firstSem.signal(); // NB: first still false
```
jakesmith: should say // NB: first still true

jakesmith:
@wangkx - apart from minor fix to 1 comment, looks ok I think.

Please squash.
Have you tried to test detaching/re-attaching Dali whilst in various delayed state, to double check the logic?

wangkx:
@jakesmith I just fixed the comment. The commits are squashed. I did multiple tests, including datach/re-attach dali in/after 1st build. Please review.

jakesmith:
But perhaps this is probably more suited to 7.8?

richardkchapman:
@wangkx Agree, this feels more like 7.8?

wangkx:
@AttilaVamos I am not sure that the smoketest errors are related to this PR. Could you please double check?

wangkx:
@richardkchapman this PR is ready to be merged to master.

PR #13097, opened 2019-10-24
HPCC-15207 Serialize ESP TxSummary items based on a log level

wangkx:
1. Always to store a TXSummary item with the log level even if
the log level is greater than preconfigured ESP TxSummaryLevel.
2. Add logLevel as an input of CTxSummary::serialize() and only
serialize a TxSummary item if its logLevel is equal to or lower
than the input.
3. Use preconfigured ESP TxSummaryLevel to serialize TxSummary
items when ESP writes TxSummary into ESP log.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/platform/espcontext.cpp
+++ b/esp/platform/espcontext.cpp
@@ -114,6 +114,8 @@ public:
     ~CEspContext()
     {
         flushTraceSummary();
+        if (m_txSummary)
+            m_txSummary->log(getTxSummaryLevel());
     }
     virtual void addOptions(unsigned opts){options|=opts;}
     virtual void removeOptions(unsigned opts){opts&=~opts;}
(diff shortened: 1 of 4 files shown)


wangkx:
@timothyklemm please review.

esp/platform/espcontext.cpp
```diff
@@ -87,6 +87,19 @@ class CEspContext : public CInterface, implements IEspContext
 
     StringAttr   m_transactionID;
 
+    void flushTraceSummary()
```
timothyklemm: This method was public, and only manipulated the in-memory content without updating the log file for a specific reason. At some point, for legacy compatibility, we may need to generate the current state of the TxSummary prior to destruction of the context. The specific use case is an ESP request option that causes the TxSummary to be added to the response output. In this case we would want the data to be as correct as possible, with the knowledge that some values might change slightly between putting the data in the response and destroying the context.

What did this refactoring get us?

esp/platform/txsummary.hpp
```diff
@@ -51,21 +52,23 @@ class CTxSummary : extends CInterface
     virtual unsigned getElapsedTime() const;
 
     // Appends all summary entries to the given buffer.
-    virtual void serialize(StringBuffer& buffer) const;
+    virtual void serialize(StringBuffer& buffer, const LogLevel logLevel = LogMin) const;
```
timothyklemm: This, and all of the interface changes in this file, are breaking changes for my security manager plugin. Is this appropriate for a point release, or should it be a minor release?

timothyklemm (changes requested):

esp/platform/espcontext.cpp
wangkx: Based on our discussion today, I will leave the flushTraceSummary() to be public so that it can be called outside. I will move the log() to the ~CEspContext().

I will remove the check 'if (getTxSummaryLevel() >= LogMin)' and move the LogMin to both the set and the append. Also do the similar inside the updateTraceSummaryHeader().

esp/platform/txsummary.hpp
```diff
@@ -51,21 +52,23 @@ class CTxSummary : extends CInterface
...
 
     // Adds the unique key and value to the end of the summary.
     // Returns true if the key value pair are added to the summary. Returns
     // false if the key is NULL, empty, or not unique within the summary.
-    virtual bool append(const char* key, const char* value);
-    template <typename TValue, typename TSuffix = const char*, class TSerializer = TokenSerializer>
-    bool append(const char* key, const TValue& value, const TSuffix& suffix = "", const TSerializer& serializer = TSerializer());
+    virtual bool append(const char* key, const char* value, const LogLevel logLevel = LogMin);
+    template <typename TValue, typename TSuffix = const char*, class TSerializer = TokenSerializer, typename TLogLevel = LogLevel>
+    bool append(const char* key, const TValue& value, const TSuffix& suffix = "", const TSerializer& serializer = TSerializer(), const TLogLevel& logLevel = LogMin);
```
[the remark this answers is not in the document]
wangkx: Will the log level from template parameter.

esp/platform/txsummary.hpp
wangkx: I can change it to a minor release.

wangkx:
@timothyklemm please review my changes.

timothyklemm (approved):

wangkx:

PR #13098, opened 2019-10-25
HPCC-22240 Add auth_feature security declarations to dynamic bindings

timothyklemm:
--------------------------------------------------------------------------------
Recognize auth_feature attributes specified on Binding/Definition and
Binding/Definition/Methods/Method elements. The syntax used is the same as is
used in ESDL definitions. Where an ESDL definition recognizes EsdlService and
EsdlMethod as security fragment scopes, the binding supports BindingService and
BindingMethod. BindingMethod takes precedence over BindingService, which takes
precedence over EsdlMethod, which takes precedence over EsdlService.

Add defaultFeatureAuth to the DynamicESDL Esp Service configuration element. The
default value for this attribute, which will be applied if it is either missing
empty, is ${service}Access:full. In addition to supporting the auth_feature
notation, the value NO_DEFAULT may be used to require that security be declared
by either the ESDL definition, the dynamic binding, or a combination of both.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/services/esdl_svc_engine/esdl_binding.cpp
+++ b/esp/services/esdl_svc_engine/esdl_binding.cpp
@@ -38,9 +38,72 @@
 #include "build-config.h"
 #include "jsmartsock.ipp"
 #include "esdl_monitor.hpp"
+#include "EsdlAccessMapGenerator.hpp"
 
 #include "loggingagentbase.hpp"
 #include "httpclient.ipp"
+
+class EsdlSvcReporter : public EsdlDefReporter
+{
+protected:
+    void reportSelf(Flags flag, const char* component, const char* level, const char* msg) const override
+    {
+        const LogMsgCategory* catPtr = nullptr;
+        switch (flag & ReportCategoryMask)
+        {
+        case ReportDisaster:  catPtr = &MCdisaster; break;
+        case ReportAError:    catPtr = &MCauditError; break;
+        case ReportIError:    catPtr = &MCinternalError; break;
+        case ReportOError:    catPtr = &MCoperatorError; break;
+        case ReportUError:    catPtr = &MCuserError; break;
+        case ReportIWarning:  catPtr = &MCinternalWarning; break;
+        case ReportOWarning:  catPtr = &MCoperatorWarning; break;
+        case ReportUWarning:  catPtr = &MCuserWarning; break;
+        case ReportDProgress: catPtr = &MCdebugProgress; break;
+        case ReportOProgress: catPtr = &MCoperatorProgress; break;
+        case ReportUProgress: catPtr = &MCuserProgress; break;
+        case ReportDInfo:     catPtr = &MCdebugInfo; break;
+        case ReportOInfo:     catPtr = &MCoperatorInfo; break;
+        case ReportUInfo:     catPtr = &MCuserInfo; break;
+        case ReportStats:     catPtr = &MCstats; break;
+        default: return;
+        }
+        LOG(*catPtr, "%s [%s]: %s", level, component, msg);
+    }
+};
+
+static const char* Scope_EsdlService = "EsdlService";
+static const char* Scope_EsdlMethod = "EsdlMethod";
+static const char* Scope_BindingService = "BindingService";
+static const char* Scope_BindingMethod = "BindingMethod";
+static const char* Var_Service = "service";
+static const char* Var_Method = "method";
+#define ANNOTATION_AUTH_FEATURE "auth_feature"
+
+// The scope mapper and level mapper instances could be static service members, but enabling the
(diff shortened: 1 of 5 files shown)


timothyklemm:
@afishbeck and @asselitx Please review.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -38,10 +38,73 @@
...
+static const char* Scope_EsdlMethod = "EsdlMethod";
+static const char* Scope_BindingService = "BindingService";
+static const char* Scope_BindingMethod = "BindingMethod";
+static const char* Var_Service = "service";
+static const char* Var_Method = "method";
+#define ANNOTATION_AUTH_FEATURE "auth_feature"
+
+// The scope mapper and level mapper instances could be static service members, but enabling the
+// declaration of these types in the header requires inclusion of headers that unnecessarily affect
+// other projects that include this class' header.
```
asselitx: What effect do you see on other projects when they're static service members?

timothyklemm: To make these instances static members of the service class would require the header included on line 41 to be included from the service header file. That change requires all projects that reference the header to change the CMakeLists file to include the hidl directory to get the file included by the include. Otherwise those projects won't compile.

I think static members might appear to be a bit "cleaner", but I decided the apparent benefit was outweighed by the cost to change unrelated projects. I added this and the following comment to address that and to preempt questions about static versus instance/local usage.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -205,6 +268,12 @@ void EsdlServiceImpl::init(const IPropertyTree *cfg,
 
         m_usesURLNameSpace = startsWith(m_serviceNameSpaceBase.str(), "http://") ? true : false;
 
+        const char* defaultFeatureAuth = srvcfg->queryProp("@defaultFeatureAuth");
+        if (isEmptyString(defaultFeatureAuth))
+            m_defaultFeatureAuth.append("${service}Access:full");
+        else if (!strieq(defaultFeatureAuth, "no_default"))
```
asselitx: This behavior looks good, but I want to confirm my understanding - the 'no_default' value would be a special value for the 'defaultFeatureAuth' attribute in the Service config element?

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -456,13 +531,37 @@ void EsdlServiceImpl::configureTargets(IPropertyTree *cfg, const char *service)
...
+            EsdlAccessMapReporter  reporter(*authMap.get(), g_reporterHelper.getLink());
+            EsdlAccessMapGenerator generator(g_scopeMapper, g_levelMapper, reporter);
+            generator.setVariable(Var_Service, service);
+            generator.setVariable(Var_Method, method);
+            generator.insertScope(Scope_EsdlService, svcAuthDef);
+            generator.insertScope(Scope_EsdlMethod, mthDef->queryMetaData(ANNOTATION_AUTH_FEATURE));
+            generator.insertScope(Scope_BindingService, svcAuthBnd);
+            generator.insertScope(Scope_BindingMethod, methodCfg.queryProp("@" ANNOTATION_AUTH_FEATURE));
+            if (!m_defaultFeatureAuth.isEmpty())
+                generator.setDefaultSecurity(m_defaultFeatureAuth);
```
asselitx: What is the effective generator DefaultSecurity if m_defaultFeatureAuth isEmpty()?

asselitx:
@afishbeck my review is complete- please review.

esp/services/esdl_svc_engine/esdl_binding.cpp
timothyklemm: The default behavior of the base class, TAccessMapGenerator, is to require security but specify no default. The existing use cases of this class set a default. This use case will set the same default, unless the user explicitly changes it in the static service configuration. And NO_DEFAULT can be used.

It is an error to require security and not specify security. For hidl, errors are not fatal. For ESDL, errors are fatal. The ESDL reporter handles enforcement by throwing an exception when the error is reported.

afishbeck (approved):
@timothyklemm looks fine.

richardkchapman:
@afishbeck @timothyklemm Just checking that 7.8 is the correct target for this change

timothyklemm:
@richardkchapman This builds on the changes made in HPCC-22458, which was designated for 7.8.

PR #13111, opened 2019-10-30
HPCC-22136 Import WU from ZAP report

wangkx:
1. Upgrade the _createWorkUnit() to support wuXMLPTree.
2. Add importWorkUnit() to IWorkUnitFactory:
   Unzip a ZAP report into unique esp data folder;
   Update WU XML;
   Calls the  _createWorkUnit() with the WU XML.
3. Add WORKUNIT_API importWorkunitFromZAPFile() which calls the
importWorkUnit().
4. Add WsWorkunit code to accept uploaded ZAP report and call
importWorkunitFromZAPFile().
5. Update WsWuInfo::getWorkunitThorLogInfo() to get Thor Log
Information from ZAP report.
6. Add the code to read log files (eclagent log and thor logs)
from ZAP files.

Also:
1. Store ESP name when WsWorkunits service is loaded. The name
is used to find out the esp data folder from getConfigurationDirectory().
2. Modify CHttpRequest::createUploadFile: If no netAddress is specified
for the upload target, create a local file as the upload target.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:


wangkx:
@jakesmith please review

common/workunit/workunit.cpp
```diff
@@ -5389,15 +5556,44 @@ class CDaliWorkUnitFactory : public CWorkUnitFactory, implements IDaliClientShut
...
+        }
+        else
+        {
+            Owned<IPropertyTree> wuXMLPTree(wuXML);
+
+            IPropertyTree *root = conn->queryRoot();
+            const char *wuidUnique = root->queryName();
+            wuXMLPTree->renameProp("/", wuidUnique);
+
+            Owned<IPropertyTree> progressPTree = pruneBranch(wuXMLPTree, "GraphProgress[1]");
```
jakesmith: The [1] shouldn't be necessary, it implies there could be >1 GraphProgress node..

common/workunit/workunit.cpp
```diff
@@ -5060,6 +5062,171 @@ bool CWorkUnitFactory::restoreWorkUnit(const char *base, const char *wuid, bool
...
+            unzipDirWithIP.append(localIP).append(unzipDir).append(PATHSEPSTR);
+
+            //update log entries in WU XML;
+            updateWUProcessLogAttrs("Process/EclAgent/*", unzipDirWithIP);
+            updateWUProcessLogAttrs("Process/Thor/*", unzipDirWithIP);
+
+            //update QueryAssociatedFiles in WU XML;
+            updateWUQueryAssociatedFilesAttrs(localIP);
+
+            readGraphProgressFileToWUPTree();
```
jakesmith: it's a little inefficient to load the file and set it into the wuTree.
For it later to be removed (pruned) then added separately to /GraphProgress, when really it was separate in the 1st place.

I think this would better as a separate step. (but still within CWorkUnitFactory::importWorkUnit) after the creation of the wuid and import of it's xml (see comments above), then add progress tree.

e.g. something like:
```
    CImportWorkUnitHelper helper(zapReportFileName, user);
    Owned<IWorkUnit> newWuid = createWorkUnit(app, user, secMgr, secUser);
    helper.setUNZIPDir(directories, component, instance, newWuid->queryWuid());
    helper.unzipZAPReport(zapReportFilePath, zapReportPassword);   
    helper.buildWUPTreesFromZAPReport(); // keeps progress tree separately

    Owned<IPropertyTree> wuidTree = helper.getWorkUnitTree();
    newWuid->import(wuidTree);
    Owned<IPropertyTree> progressTree = helper.getPropgressTree();
    importProgress(newWuid->queryWuid(), progressTree);
```

esp/bindings/http/platform/httptransport.cpp
```diff
@@ -1971,6 +1971,10 @@ IFile* CHttpRequest::createUploadFile(const char * netAddress, const char* fileP
     }
     tmpFileName.appendf("%s/%s.part", filePath, fileName.str());
 
+    //If no netAddress is specified, create a local file.
+    if (isEmptyString(netAddress))
+        return createIFile(tmpFileName);
```
jakesmith: Not new, but looking at the code above this..
It looks rather odd, reversing twice, strchr and writing inside StringBuffer..
What is it trying to do exactly?

it's called tmpFileName, should it be using createUniqueFile instead?

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -2051,6 +2145,20 @@ void WsWuInfo::getWorkunitThorSlaveLog(IPropertyTree* directories, const char *p
...
+{
+    StringBuffer zapWUFolder;
+    getWorkunitProcessLogPath("Thor", zapWUFolder);
+
+    VStringBuffer logFileName("%s%c%s_thorslave.%u.%s.log", zapWUFolder.str(), PATHSEPCHAR, thorProcess, slaveNum, logDate);
+    Owned<IFile> logfile = createIFile(logFileName);
+    if (!logfile)
+        throw MakeStringException(ECLWATCH_CANNOT_OPEN_FILE, "Cannot open %s.", logFileName.str());
+
+    getWorkunitLogFromZAPFile(logfile, logFileName, buf, outFile);
```
jakesmith: I haven't looked at changes here and above very carefully yet,
but we do plan to have logs per workunit.
So it would better if possible, there was little or no mention of "ZAP" in related changed of code that is looking at workunits whether they are 'original' or imported.

It should if at all possible be agnostic and the code look as similar as possible.

As I say, I haven't fully looked at all the changes involved here,
and perhaps there are some cases where the workunit has references to clusters (for example) that don't exist in the environment they've been imported into, and the current code relies on them to some extent.

Let's have an offline discussion re. viewing/using imported workunits, vs existing, to explore what differences are needed..

jakesmith:
@wangkx - I've given this a 1st pass, and made various comments.
In general it looks good, but I have some questions, and I think it could be restructured a little.

I did not review the code relating to viewing imported files (that refers to them a ZAP workunits), I got the feeling that there was too much special casing, which I would hope there wouldn't.
Perhaps necessary for backward compatibility, if old code relies on lots of lookups into (for example) clusters in the environment.. I'm not sure.
Let's discuss that particular aspect offline.

Did you test the Cassandra implementation?

common/workunit/workunit.cpp
wangkx: @jakesmith In fact, I DO like this approach better because it is much easy to understand and to implement. The existing approach is based on your comment: 'I would be better/more efficient to do it in one call. I would add a new createWorkUnit method that takes a IPropertyTree. Load the archive workunit xml to a ptree, patch up where necessary, than add via the new createWorkUnit, that copies the content into the newly created uniquely named wuid.' As your comment said, the existing approach has only one call to Dali WUs Tree. Is the existing approach more efficient than this approach?

jakesmith: @wangkx - efficiency is important, whilst not a critical area, it could be time consuming and is easy to avoid adding then removing the progress tree here.
a new createWorkUnit method (as opposed to altering the existing) would have worked too, if you had passed the workunit tree and progress tree separately.
The issue here is you add it to a loaded workunit xml , but then prune it later, before making the new connection for the GraphProgress.

The both new and approach, must make 2 connections - one for the /WorkUnit/... and one for the GraphProgress, there is no difference in efficiency wrt as far as that is concerned, in both cases it has to copy 2 trees. So as far a Dali traffic is concerned there is no difference.
It's the local copying/pruning of the GraphProgress that can be avoided with this approach.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -1975,7 +2038,32 @@ void WsWuInfo::getWorkunitThorLog(const char* fileName, MemoryBuffer& buf, const
...
+{
+    //This is a WU from ZAP report. The thor log only contains the log lines related to this WU.
+    //The whole log file should be sent back. 
+    if (!isEmptyString(outFile))
+    {
+        OwnedIFile oFile = createIFile(outFile);
+        if (!oFile)
+            throw MakeStringException(ECLWATCH_CANNOT_OPEN_FILE, "Cannot open %s.", outFile);
+
+        copyFile(oFile, iFile);
```
jakesmith: Currently getWorkunitLogFromZAPFile is both writing to the file and copying to buffer.
There's a:
```
return;
```
missing here

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -1049,25 +1049,81 @@ void WsWuInfo::getInfo(IEspECLWorkunit &info, unsigned long flags)
...
         {
-            IWARNLOG("Cannot find TargetClusterInfo for workunit %s", cw->queryWuid());
-            return countThorLog;
+            numberOfSlaveLogs = getNumberOfSlavesFromZAPFiles();
+            thorMasterLogDateSearchString.set("_thormaster.");
+        }
+        else
+        {
+            numberOfSlaveLogs = getNumberOfSlavesByClusterName();
+            thorMasterLogDateSearchString.set("/thormaster.");
```
jakesmith: can u always deduce # of slaves from the log (names) if number of slaves is required?
What if slave logs aren't available, because zap report didn't contain them?

What is numberOfSlaveLogs needed for?
I see the:
```
                thorLog->setNumberSlaves(numberOfSlaveLogs);
```
..below. Is that so Eclwatch can go to a particular slave log?

If so, could deduce # of slaves from slave log names if available in imported helpers.
Just a thought, perhaps workunit should always, contain group and pattern to slave logs, so that doesn't have to go lookup stuff in environment which may be out of date anyway.
Then an imported workunit would not be particularly different, would also contain group and pattern in same way.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -1049,25 +1049,81 @@ void WsWuInfo::getInfo(IEspECLWorkunit &info, unsigned long flags)
...
-        Owned<IConstWUClusterInfo> clusterInfo = getTargetClusterInfo(clusterName.str());
-        if (!clusterInfo)
+        unsigned numberOfSlaveLogs = 0;
+        StringAttr thorMasterLogDateSearchString;
+        if (debugImportedStr.length()) //WU imported from a ZAP report
         {
-            IWARNLOG("Cannot find TargetClusterInfo for workunit %s", cw->queryWuid());
-            return countThorLog;
+            numberOfSlaveLogs = getNumberOfSlavesFromZAPFiles();
+            thorMasterLogDateSearchString.set("_thormaster.");
```
jakesmith: Isn't there a directory per workunit?
If so, can't pattern be the same, i.e. "thormaster." ?

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -1049,25 +1049,81 @@ void WsWuInfo::getInfo(IEspECLWorkunit &info, unsigned long flags)
...
     {
-        StringAttr clusterName(cw->queryClusterName());
-        if (!clusterName.length()) //Cluster name may not be set yet
-            return countThorLog;
-
-        Owned<IConstWUClusterInfo> clusterInfo = getTargetClusterInfo(clusterName.str());
-        if (!clusterInfo)
+        unsigned numberOfSlaveLogs = 0;
+        StringAttr thorMasterLogDateSearchString;
+        if (debugImportedStr.length()) //WU imported from a ZAP report
```
jakesmith: As discussed offline, would be good to remove references to 'improted' and ZAP where possible in new code, and instead distinguish where necessary as 'logsPerWorkUnit' flag or similar.
Since we plan to have workunits per workunit in general anyway.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -1122,7 +1178,7 @@ unsigned WsWuInfo::getWorkunitThorLogInfo(IArrayOf<IEspECLHelpFile>& helpers, IE
                     continue;
 
                 const char* pStr = logName.str();
-                const char* ppStr = strstr(pStr, "/thormaster.");
+                const char* ppStr = strstr(pStr, thorMasterLogDateSearchString.get());
```
jakesmith: why is this necessary?
To get these logs, it has called cw->getLogs("Thor", ...)
so they explicitly thor log files already (no matter what name they have)

esp/services/ws_workunits/ws_workunitsHelpers.cpp
jakesmith: There's:
```
1142:            StringBuffer groupName;
1143:            getClusterThorGroupName(groupName, processName.str());
```

that will likely not make much sense for an imported version.
i.e. the processName will not exist in the local environment.

what is groupName needed for?

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -1896,6 +1952,13 @@ void WsWuInfo::getWorkunitEclAgentLog(const char* fileName, const char* agentPid
     Owned<IFile> rFile = createIFile(fileName);
     if(!rFile)
         throw MakeStringException(ECLWATCH_CANNOT_OPEN_FILE, "Cannot open file %s.", fileName);
+
+    if (debugImportedStr.length()) //WU imported from a ZAP report
+    {
+        getWorkunitLogFromZAPFile(rFile, fileName, buf, outFile);
```
jakesmith: can we switch from debugImportedStr to a flag like e.g. logsPerWorkunit,
and from getWorkunitLogFromZAPFile and similar to getWorkunitLogSingleFile

esp/services/ws_workunits/ws_workunitsHelpers.cpp
jakesmith: Also note:
```
ppStr += 12
```
would be wrong if thorMasterLogDateSearchString varies. Should avoid constants like this

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -2051,6 +2145,20 @@ void WsWuInfo::getWorkunitThorSlaveLog(IPropertyTree* directories, const char *p
     getWorkunitThorSlaveLog(nodeGroup, ipAddress, logDate, logDir.str(), slaveNum, buf, outFile, forDownload);
 }
 
+void WsWuInfo::getWorkunitThorSlaveLogFromZAPFile(const char* thorProcess, const char* logDate, int slaveNum,
+    MemoryBuffer& buf, const char* outFile)
+{
+    StringBuffer zapWUFolder;
```
jakesmith: should avoid referring to as 'zap'

esp/services/ws_workunits/ws_workunitsHelpers.hpp
```diff
@@ -147,12 +149,19 @@ class WsWuInfo
         const char* sourceAlias, MemoryBuffer &mb, bool forDownload);
     void copyContentFromRemoteFile(const char* sourceFileName, const char* sourceIPAddress,
         const char* sourceAlias, const char *outFileName);
+    void readDebugImportedStr()
+    {
+        Owned<IStringIterator> debugValues(&cw->getDebugValues("imported"));
+        if (debugValues->first())
+            debugValues->str(debugImportedStr);
```
jakesmith: I think instead do:
```
   SCMStringBuffer imported;
   cw->getDebugValue(imported);
   logsPerWorkUnit = !isEmptyString(imported.str());
```

jakesmith (changes requested):

jakesmith:
@wangkx - added some more comments.

Because per workunits will eventually be the norm.
It would be best to common up as much as possible and remove references where possible of 'ZAP' and 'imported', and instead distinguish between the 2 flavours only by whether logs per workunit, or filtered logging.

Need to check everything works as expected when the ZAP file is imported and has no relation to the current environment, i.e. the thor process, group, cluster do not not exist.
So the code in ws_workunitsHelper must where possible not call into the environment to lookup info (like # of slaves), and instead rely on info in the workunit.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
wangkx: > .below. Is that so Eclwatch can go to a particular slave log?
Yes, ECLWatch has to display a number which a user may choose to download a slave log.  
> what is groupName needed for?
The groupName is used for standard (not imported) WU to get slave log.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
wangkx: When a ZAP report is created, the '_' is added in the front of the 'thormaster.'

esp/services/ws_workunits/ws_workunitsHelpers.cpp
jakesmith: So what if instead there were methods in the workunit like:
```
IGroup *IConstWorkUnit::getThorGroup();
```
and:
```
const char *IConstWorkUnit::getSlaveLogPattern();
```

and Thor could published the group and the slave log pattern to the workunit.
For legacy workunits (that didn't have published group/pattern), it could as you do, get the group based on getClusterThorGroupName, and the slave pattern from the environment directories.

Then this code could be agnostic to where workunit came from

esp/services/ws_workunits/ws_workunitsHelpers.cpp
jakesmith: ok, you _could_ when you unzip it, normalize it so filenames are same..
I think a directory per workunit with consistently named logs is what we want.
Not sure why ZAP file isn't like that too, any reason it couldn't be?

wangkx: When creating a ZAP report, all of WU files are retrieved/created into one folder with one layer. A WU may have thor logs from >1 thor instances. To identify them, thor instance name is used as a prefix of a thor master log, like: analyt_thor400_90_a_thormaster.2019_02_14

esp/services/ws_workunits/ws_workunitsHelpers.cpp
wangkx: Should the coming import() function insert a 'logsPerWorkUnit' flag into the WU tree if not exist?

esp/services/ws_workunits/ws_workunitsHelpers.cpp
jakesmith: right ok, yes that makes sense.
In that case could use a regular pattern in both cases "*thormaster*.log" or similar..
To reduce amount of conditional code?

esp/services/ws_workunits/ws_workunitsHelpers.cpp
jakesmith: for now I think the 'imported' string you added in the publish info is fine.
But internally refer to that once, and use a flag everywhere for conditional code where necessary.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
wangkx: The code here is to find out the log date. Some WU may run for > 1 days. ECLWatch has to display a dialog box for a user to choose a thor slave using the log date, as well as thor process name and a slave ID.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
wangkx: We may switch from debugImportedStr to a flag like e.g. logsPerWorkunit if we do not need the flag to find out the log date and # of slaves from ZAP.

common/workunit/workunit.cpp
```diff
@@ -5389,15 +5556,44 @@ class CDaliWorkUnitFactory : public CWorkUnitFactory, implements IDaliClientShut
...
+            IPropertyTree *root = conn->queryRoot();
+            const char *wuidUnique = root->queryName();
+            wuXMLPTree->renameProp("/", wuidUnique);
+
+            Owned<IPropertyTree> progressPTree = pruneBranch(wuXMLPTree, "GraphProgress[1]");
+            if (progressPTree)
+            {
+                VStringBuffer xpath("/GraphProgress/%s", wuidUnique);
+                Owned<IRemoteConnection> progressConn = querySDS().connect(xpath, myProcessSession(), RTM_LOCK_WRITE|RTM_CREATE_QUERY, SDS_LOCK_TIMEOUT);
+                if (progressConn)
```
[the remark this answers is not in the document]
wangkx: Will throw an exception

common/workunit/workunit.cpp
```diff
@@ -5389,15 +5556,44 @@ class CDaliWorkUnitFactory : public CWorkUnitFactory, implements IDaliClientShut
...
+                {
+                    IPropertyTree *progressRoot = progressConn->queryRoot();
+                    if (progressRoot->hasChildren())
+                        WARNLOG("_createWorkUnit WUID %s graphprogress already exists, replacing", wuidUnique);
```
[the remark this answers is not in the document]
wangkx: Will remove the check.

common/workunit/workunit.cpp
```diff
@@ -5389,15 +5556,44 @@ class CDaliWorkUnitFactory : public CWorkUnitFactory, implements IDaliClientShut
...
+                    progressRoot->setPropTree(nullptr, progressPTree.getClear());
+                }
+            }
+
+            if (root->hasChildren())
+                WARNLOG("_createWorkUnit WUID %s replacing wuXML", wuidUnique);
```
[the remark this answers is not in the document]
wangkx: Will remove the check.

common/workunit/workunit.cpp
```diff
@@ -5389,15 +5556,44 @@ class CDaliWorkUnitFactory : public CWorkUnitFactory, implements IDaliClientShut
...
+            conn->queryRoot()->setPropInt("@wuidVersion", WUID_VERSION);
+            conn->queryRoot()->setProp("@totalThorTime", "");
+        }
+        else
+        {
+            Owned<IPropertyTree> wuXMLPTree(wuXML);
+
+            IPropertyTree *root = conn->queryRoot();
+            const char *wuidUnique = root->queryName();
+            wuXMLPTree->renameProp("/", wuidUnique);
```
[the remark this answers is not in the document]
wangkx: ok

common/workunit/workunit.cpp
```diff
@@ -5060,6 +5062,171 @@ bool CWorkUnitFactory::restoreWorkUnit(const char *base, const char *wuid, bool
...
+        IPTree *queryWUPTree()
+        {
+            return wuTree;
+        }
+        void setUNZIPDir(const IPropertyTree *directories, const char *component, const char *instance)
+        {   //Set a unique unzip folder inside the component's data folder
+            getConfigurationDirectory(directories, "data", component, instance, unzipDir);
+            unzipDir.append(PATHSEPSTR).append(zapReportFileName.length() - 4, zapReportFileName);
+            if (!user.isEmpty())
+                unzipDir.append("_").append(user);
```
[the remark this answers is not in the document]
wangkx: Will do

common/workunit/workunit.cpp
wangkx: ok

esp/bindings/http/platform/httptransport.cpp
wangkx: The code may be replaced by pathTail(). Will do.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
wangkx: Will add back.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
wangkx: Will do

esp/services/ws_workunits/ws_workunitsHelpers.cpp
wangkx: Will do

esp/services/ws_workunits/ws_workunitsHelpers.hpp
wangkx: Not sure it should be called 'logsPerWorkUnit' because the debugImportedStr is also used for # of thor slaves and log date.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
wangkx: Here, I need: ThorGroup, # of slaves, and the dates of the thor logs. The getThorGroup() returns the ThorGroup. What will getSlaveLogPattern() return?

wangkx:
@jakesmith please review my changes and let me know if I miss anything.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
jakesmith: But should not use constants like '12'.
Use a const string and strlen if necessary for readability/maintainability.

esp/services/ws_workunits/ws_workunitsHelpers.hpp
jakesmith: Where?
As far as I can see, it is only tested to see if set (not it's content), e.g.:
```
if (debugImportedStr.length()) //WU imported from a ZAP report
```

So is essentially a simple flag

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -1084,7 +1140,9 @@ unsigned WsWuInfo::getWorkunitThorLogInfo(IArrayOf<IEspECLHelpFile>& helpers, IE
             uniqueProcesses.setValue(processName.str(), true);
 
             StringBuffer groupName;
-            getClusterThorGroupName(groupName, processName.str());
+            cw->getThorGroup(groupName);
```
jakesmith: see other comment.
I know I suggested it, but as you can see now you've implemented, it doesn't make sense, because the group can be different depending on 'processName', i.e. Thor can run on >1 group.
Need to remove getThorGroup/setThorGroup.

jakesmith (changes requested):

jakesmith:
@wangkx - haven't been through it all, but it doesn't feel encapsulated enough yet. i.e. there's too much conditional code depending on where the workunit came from.
Ideally the esp code would call methods in the workunit to get all the necessary info., such that esp code dealing with workunits regardless of whether they're imported or not, was the same.
Imported workunits can have extra info. added, to help the workunit methods return the correct info.

Standard workunits can do what they used to do, like lookup clusters in the environment, lookup groups etc., but it would be good to try to get the esp code not to manually do this, it should be encapsulated in workunit.* as much as possible.

esp/services/ws_workunits/ws_workunitsHelpers.hpp
wangkx: It is a simple flag. How bout changing the debugImportedStr to hasDebugImportedStr? The flag is not only used for getting logs, also used for finding numberOfSlaves for standard WU and for Imported WU.

esp/bindings/http/platform/httptransport.cpp
```diff
@@ -1957,19 +1957,18 @@ bool CHttpRequest::readUploadFileName(CMimeMultiPart* mimemultipart, StringBuffe
...
-        pStr = strchr(str, '/');
-    if (pStr)
-    {
-        pStr[0] = 0;
-        fileName.clear().append(str).reverse();
-    }
-    tmpFileName.appendf("%s/%s.part", filePath, fileName.str());
+    StringBuffer tmpFileName(filePath);
+    if (tmpFileName.charAt(tmpFileName.length() - 1) != PATHSEPCHAR)
+        tmpFileName.append(PATHSEPCHAR);
```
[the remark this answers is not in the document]
wangkx: Will do

esp/bindings/http/platform/httptransport.cpp
```diff
@@ -1957,19 +1957,18 @@ bool CHttpRequest::readUploadFileName(CMimeMultiPart* mimemultipart, StringBuffe
...
+    tmpFileName.append(pathTail(fileName));
+    fileNameWithPath.set(tmpFileName);
```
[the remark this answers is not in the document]
wangkx: OK

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -1049,25 +1049,81 @@ void WsWuInfo::getInfo(IEspECLWorkunit &info, unsigned long flags)
...
+    StringBuffer zapWUFolderPath;
+    getWorkunitProcessLogPath("Thor", zapWUFolderPath);
+
+    unsigned numberOfSlaves = 0;
+    Owned<IFile> zapFileFolder = createIFile(zapWUFolderPath);
+    Owned<IDirectoryIterator> thorSlaveFiles = zapFileFolder->directoryFiles("*_thorslave.*");
+    ForEach(*thorSlaveFiles)
+    {
+        const char* pStr = strstr(thorSlaveFiles->query().queryFilename(), "_thorslave.");
+        pStr += 11;
```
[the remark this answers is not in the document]
wangkx: Will change it to strlen.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -1049,25 +1049,81 @@ void WsWuInfo::getInfo(IEspECLWorkunit &info, unsigned long flags)
...
+        const char* slaveNumStrEnd = strchr(pStr, '.');
+        if (!slaveNumStrEnd)
+            continue;
+
+        StringBuffer slaveNumStr;
+        slaveNumStr.append(slaveNumStrEnd - pStr, pStr);
+        unsigned slaveNum = atoi(slaveNumStr);
+        if (slaveNum > numberOfSlaves)
+            numberOfSlaves = slaveNum;
```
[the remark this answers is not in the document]
wangkx: The format of the slave log in ZAP report is not n_of_m. For example, this is a slave log: thor02_160way_thorslave.9.2019_11_07

esp/services/ws_workunits/ws_workunitsHelpers.cpp
wangkx: Will remove

wangkx: Can one WU use >1 thor group?

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -1049,25 +1049,81 @@ void WsWuInfo::getInfo(IEspECLWorkunit &info, unsigned long flags)
...
+    unsigned numberOfSlaves = 0;
+    Owned<IFile> zapFileFolder = createIFile(zapWUFolderPath);
+    Owned<IDirectoryIterator> thorSlaveFiles = zapFileFolder->directoryFiles("*_thorslave.*");
+    ForEach(*thorSlaveFiles)
+    {
+        const char* pStr = strstr(thorSlaveFiles->query().queryFilename(), "_thorslave.");
+        pStr += 11;
+        const char* slaveNumStrEnd = strchr(pStr, '.');
+        if (!slaveNumStrEnd)
+            continue;
```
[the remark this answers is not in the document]
wangkx: In the ZAP report, the name of the slave log file does not contain 'of'.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
wangkx: If we have the logPerWorkunit, I still need to know the number of slaves (for example, 50) so that ECLwatch can display a list of numbers (from 1 to 50) for a user to download a specific slave log. So, Should the pattern be: /var/log/HPCCSystems/mythor/thorslave.*_of_50.2019_11_1[2-3].log or /var/log/HPCCSystems/mythor/W20191113-140205/thorslave.*_of_50.log ?

wangkx:

plugins/cassandra/cassandrawu.cpp
```diff
@@ -3259,6 +3374,7 @@ class CCasssandraWorkUnitFactory : public CWorkUnitFactory, implements ICassandr
                 wuXML->setProp("@xmlns:xsi", "http://www.w3.org/1999/XMLSchema-instance");
                 wuXML->setPropInt("@wuidVersion", WUID_VERSION);  // we implement the latest version.
                 wuXML->setProp("@totalThorTime", ""); // must be non null, otherwise sorting by thor time excludes the values
+
```
jakesmith: trivial: unintentional extra line added?

esp/services/ws_workunits/ws_workunitsHelpers.cpp
jakesmith: yes.
A workunit can run on any of the Thor's attached to the queue, and there can be any number of Thors, which may be on same group, or independent groups.

jakesmith: in fact I was wrong, all Thor's in a logical cluster must be the same size *at the moment*.
It would be best to try not to assume that's true though, as it may well change.

jakesmith:
@wangkx - looks better, I think getting closer, have added some new comments.
May be worth discussing some of the points in flowdock to clarify.

plugins/cassandra/cassandrawu.cpp
wangkx: Will remove

esp/services/ws_workunits/ws_workunitsService.cpp
```diff
@@ -4442,6 +4444,46 @@ int CWsWorkunitsSoapBindingEx::onGetForm(IEspContext &context, CHttpRequest* req
...
+            if ((accessOwn != SecAccess_Full) || (accessOthers != SecAccess_Full))
+                throw MakeStringException(-1, "Permission denied.");
+    
+            StringBuffer password, importQueryAssociatedFiles;
+            request->getParameter("importQueryAssociatedFiles", importQueryAssociatedFiles);
+            request->getParameter("Password", password);
+
+            request->readContentToFiles(nullptr, zipFolder, fileNames);
+            if (!fileNames.ordinality())
+                throw MakeStringException(-1, "Failed to read upload content.");
```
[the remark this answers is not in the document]
wangkx: Will loop through each file name and import > 1 WUs if available.

common/workunit/workunit.cpp
```diff
@@ -13778,3 +14153,23 @@ bool isValidMemoryValue(const char *memoryUnit)
...
+    if (isEmptyString(component))
+        throw MakeStringException(WUERR_InvalidUserInput, "Empty Component name");
+    if (isEmptyString(instance))
+        throw MakeStringException(WUERR_InvalidUserInput, "Empty Instance name");
+
+    Owned<IEnvironmentFactory> envFactory = getEnvironmentFactory(true);
+    Owned<IConstEnvironment> env = envFactory->openEnvironment();
+    Owned<IPropertyTree> root = &env->getPTree();
+
+    Owned<IWorkUnitFactory> wuFactory = getWorkUnitFactory();
```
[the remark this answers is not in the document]
wangkx: I may call the getWorkUnitFactory() when initializing WsWorkunits service and call the wuFactory->importWorkUnit(...) directly. Then, I may remove this WORKUNIT_API: importWorkunitFromZAPFile().

common/workunit/workunit.cpp
```diff
@@ -13778,3 +14153,23 @@ bool isValidMemoryValue(const char *memoryUnit)
...
+extern WORKUNIT_API IWorkUnit * importWorkunitFromZAPFile(const char *zapReportFileName, const char *zapReportFilePath,
+    const char *zapReportPassword, const char *component, const char *instance, const char *app, const char *user,
+    ISecManager *secMgr, ISecUser *secUser)
+{
+    if (isEmptyString(zapReportFileName))
+        throw MakeStringException(WUERR_InvalidUserInput, "Empty ZAP report name");
+    if (isEmptyString(component))
+        throw MakeStringException(WUERR_InvalidUserInput, "Empty Component name");
+    if (isEmptyString(instance))
+        throw MakeStringException(WUERR_InvalidUserInput, "Empty Instance name");
```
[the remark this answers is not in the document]
wangkx: Will change it to makeStringException().

common/workunit/workunit.cpp
```diff
@@ -5060,6 +5087,174 @@ bool CWorkUnitFactory::restoreWorkUnit(const char *base, const char *wuid, bool
...
+            if (isEmptyString(zapReportPassword))
+                zipCommand.appendf("%s -d %s", zapReportFileName.get(), unzipDir.str());
+            else
+                zipCommand.appendf("%s -P %s -d %s", zapReportFileName.get(), zapReportPassword, unzipDir.str());
+            return (system(zipCommand.str()));
+        }
+        void buildPTreesFromZAPReport()
+        {
+            readWUXMLFileToPTree();
+            wuTree->renameProp("/", wuid.get()); //renameProp(nullptr, wuid.get()) not work: renameProp: cannot rename self, renameProp has to rename in context of a parent
```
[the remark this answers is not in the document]
wangkx: Will remove.

common/workunit/workunit.cpp
```diff
@@ -5060,6 +5087,174 @@ bool CWorkUnitFactory::restoreWorkUnit(const char *base, const char *wuid, bool
...
+            setImportDebugAttribute();
```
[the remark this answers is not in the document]
wangkx: Ok, will call the existing API.

common/workunit/workunit.cpp
```diff
@@ -5060,6 +5087,174 @@ bool CWorkUnitFactory::restoreWorkUnit(const char *base, const char *wuid, bool
...
+        void setImportDebugAttribute()
+        {
+            StringBuffer str("Import ");
+            str.append(wuid.get());
+            if (!isEmptyString(user))
+                str.append(" by ").append(user);
+            str.append(" at ").append(unzipDateTime);
+            str.append(" from ").append(zapReportFileName);
+            wuTree->setProp("Debug", "");
+            wuTree->setProp("Debug/imported", str);
```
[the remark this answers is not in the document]
wangkx: The code for unzipDateTime may be removed by a mistake. I will add it back.
The scope stores who originally created the WU in the ZAP report. The intention for adding user is to record who imports the WU.  @jakesmith do you think there is no need to record who imports the WU?

common/workunit/workunit.cpp
```diff
@@ -6200,6 +6402,40 @@ class CStringPTreeAttrIterator : implements IStringIterator, public CInterface
     virtual bool isValid() { return it->isValid(); }
     virtual IStringVal & str(IStringVal &s) { s.set(it->query().queryProp(name)); return s; }
 };
+
+//==========================================================================================
+
+class CWUThorLogInfo: public CSimpleInterface, implements IConstWUThorLogInfo
```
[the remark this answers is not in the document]
wangkx: Will change it.

common/workunit/workunit.cpp
```diff
@@ -10013,6 +10249,145 @@ IWUGraphStats *CLocalWorkUnit::updateStats(const char *graphName, StatisticCreat
...
+            const char *pStr = logName.str();
+            const char *datePtr = strstr(pStr, thorMasterLogDateSearchString.get());
+            if (!datePtr)
+            {
+                IWARNLOG("Invalid thorlog entry in workunit xml: %s", logName.str());
+                continue;
+            }
+
+            datePtr += thorMasterLogDateSearchString.length();
+            if (strlen(datePtr) < 10) //log single file
```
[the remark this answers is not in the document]
wangkx: Will add a comment

common/workunit/workunit.cpp
```diff
@@ -10013,6 +10249,145 @@ IWUGraphStats *CLocalWorkUnit::updateStats(const char *graphName, StatisticCreat
...
+                Owned<CWUThorLogInfo> thorLog = new CWUThorLogInfo(logName.str(), processName.str(), groupName, "");
+                wuThorLogs.append(*thorLog.getLink());
+                continue;
+            }
+
+            StringBuffer logDate;
+            logDate.append(10, datePtr);
+
+            Owned<CWUThorLogInfo> thorLog = new CWUThorLogInfo(logName.str(), processName.str(), groupName, logDate);
+            wuThorLogs.append(*thorLog.getLink());
```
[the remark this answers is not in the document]
wangkx: Will change

common/workunit/workunit.cpp
```diff
@@ -10013,6 +10249,145 @@ IWUGraphStats *CLocalWorkUnit::updateStats(const char *graphName, StatisticCreat
...
+    return numberOfThorSlaves;
+}
+
+void CLocalWorkUnit::getWUThorLogInfo(IArrayOf<IConstWUThorLogInfo> &wuThorLogs) const
+{
+    StringAttr thorMasterLogDateSearchString;
+    if (hasDebugValue("imported")) //WU imported from a ZAP report
+        thorMasterLogDateSearchString.set("_thormaster.");
+    else
+        thorMasterLogDateSearchString.set("/thormaster.");
```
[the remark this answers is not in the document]
wangkx: How about adding 'thorMasterLogSearchString=_thormaster.' in Debug/imported?

common/workunit/workunit.cpp
```diff
@@ -5060,6 +5087,174 @@ bool CWorkUnitFactory::restoreWorkUnit(const char *base, const char *wuid, bool
...
+        }
+        void readGraphProgressFileToPTree()
+        {
+            StringBuffer fileName;
+            if (!findZAPFile("ZAPReport_*.graphprogress", true, fileName))
+                return;
+            graphProgressTree.setown(createPTreeFromXMLFile(fileName));
+            if (!graphProgressTree)
+                throw MakeStringException(WUERR_InvalidUserInput, "Failed to retrieving ZAPReport_*.graphprogress.");
+        }
```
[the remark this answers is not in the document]
wangkx: Will change.

common/workunit/workunit.cpp
```diff
@@ -5060,6 +5087,174 @@ bool CWorkUnitFactory::restoreWorkUnit(const char *base, const char *wuid, bool
...
+        {
+            return wuTree;
+        }
+        IPTree *querygraphProgressPTree()
+        {
+            return graphProgressTree;
+        }
+        void setUNZIPDir(const IPropertyTree *directories, const char *component, const char *instance)
+        {   //Set a unique unzip folder inside the component's data folder
+            getConfigurationDirectory(directories, "data", component, instance, unzipDir);
```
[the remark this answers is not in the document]
wangkx: Will change similar to getWorkUnitFactory().

common/workunit/workunit.cpp
```diff
@@ -5060,6 +5087,174 @@ bool CWorkUnitFactory::restoreWorkUnit(const char *base, const char *wuid, bool
...
+
+            //update QueryAssociatedFiles in WU XML;
+            updateWUQueryAssociatedFilesAttrs(localIP);
+
+            readGraphProgressFileToPTree();
+        }
+    };
+
+    CImportWorkUnitHelper helper(zapReportFileName, user);
+    Owned<IWorkUnit> newWU = createWorkUnit(app, user, secMgr, secUser);
```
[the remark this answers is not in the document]
wangkx: Please see my other comment.

common/workunit/workunit.cpp
```diff
@@ -10013,6 +10249,145 @@ IWUGraphStats *CLocalWorkUnit::updateStats(const char *graphName, StatisticCreat
...
+        return true;
+    return false;
+}
+
+unsigned CLocalWorkUnit::getNumberOfThorSlaves() const
+{
+    unsigned numberOfThorSlaves = 0;
+    StringBuffer slaveLogPattern;
+    getSlaveLogPattern(slaveLogPattern);
+    if (!slaveLogPattern.isEmpty())
```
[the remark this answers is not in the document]
wangkx: Yes, the getNumberOfThorSlaves is all I need for now. I will modify addProcess() to add 'slaves=n'.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -2051,6 +2061,20 @@ void WsWuInfo::getWorkunitThorSlaveLog(IPropertyTree* directories, const char *p
     getWorkunitThorSlaveLog(nodeGroup, ipAddress, logDate, logDir.str(), slaveNum, buf, outFile, forDownload);
 }
 
+void WsWuInfo::getWorkunitThorSlaveLogSingleFile(const char* thorProcess, const char* logDate, int slaveNum,
+    MemoryBuffer& buf, const char* outFile)
+{
+    StringBuffer logPath;
+    getWorkunitProcessLogPath("Thor", logPath);
+
+    VStringBuffer logFileName("%s%c%s_thorslave.%u.%s.log", logPath.str(), PATHSEPCHAR, thorProcess, slaveNum, logDate);
```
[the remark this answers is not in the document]
wangkx: How about adding 'thorSlaveLogSearchString=_thorslave' under Debug/imported.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -2985,6 +3009,7 @@ bool addToQueryString(StringBuffer &queryString, const char *name, const char *v
 int WUSchedule::run()
 {
     PROGLOG("ECLWorkunit WUSchedule Thread started.");
+
```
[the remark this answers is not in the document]
wangkx: Will remove

esp/services/ws_workunits/ws_workunitsHelpers.hpp
```diff
@@ -137,6 +137,8 @@ struct WsWUExceptions
 
 class WsWuInfo
 {
+    bool logSingleFile = false;
```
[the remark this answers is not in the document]
wangkx: Will change

common/workunit/workunit.hpp
```diff
@@ -1286,6 +1295,10 @@ interface IConstWorkUnit : extends IConstWorkUnitInfo
     virtual void clearGraphProgress() const = 0;
     virtual IStringVal & getAbortBy(IStringVal & str) const = 0;
     virtual unsigned __int64 getAbortTimeStamp() const = 0;
+    virtual StringBuffer & getSlaveLogPattern(StringBuffer & str) const = 0;
+    virtual bool logSingleFile() const = 0;
```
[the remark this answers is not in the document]
wangkx: Will change

common/workunit/workunit.cpp
```diff
@@ -4041,6 +4056,18 @@ class CLockedWorkUnit : implements ILocalWorkUnit, implements IExtendedWUInterfa
             { return c->getAbortBy(str); }
     virtual unsigned __int64 getAbortTimeStamp() const
             { return c->getAbortTimeStamp(); }
+    virtual StringBuffer &getSlaveLogPattern(StringBuffer &str) const
+            { return c->getSlaveLogPattern(str); }
+    virtual void setSlaveLogPattern(const char *pattern)
+            { return c->setSlaveLogPattern(pattern); }
```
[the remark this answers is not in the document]
wangkx: No need for setSlaveLogPattern/getSlaveLogPattern. Will remove.

common/workunit/workunit.cpp
```diff
@@ -10013,6 +10249,145 @@ IWUGraphStats *CLocalWorkUnit::updateStats(const char *graphName, StatisticCreat
...
+    Owned<IStringIterator> thorInstances = getProcesses("Thor");
+    ForEach (*thorInstances)
+    {
+        SCMStringBuffer processName;
+        thorInstances->str(processName);
+        if (processName.length() < 1)
+            continue;
+
+        bool *found = uniqueProcesses.getValue(processName.str());
+        if (found && *found)
```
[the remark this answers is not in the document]
wangkx: Ok

common/workunit/workunit.cpp
```diff
@@ -5060,6 +5087,174 @@ bool CWorkUnitFactory::restoreWorkUnit(const char *base, const char *wuid, bool
...
+        }
+    };
+
+    CImportWorkUnitHelper helper(zapReportFileName, user);
+    Owned<IWorkUnit> newWU = createWorkUnit(app, user, secMgr, secUser);
+    helper.setWUID(newWU->queryWuid());
+    helper.setUNZIPDir(directories, component, instance);
+    helper.unzipZAPReport(zapReportFilePath, zapReportPassword);
+    helper.buildPTreesFromZAPReport();
+    newWU->import(helper.queryWUPTree(), helper.querygraphProgressPTree());
```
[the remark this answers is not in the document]
wangkx: Should I also remove the ZAP report file (the zip file being uploaded)?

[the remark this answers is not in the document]
jakesmith: yes I think should

common/workunit/workunit.cpp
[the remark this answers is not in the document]
jakesmith: >The scope stores who originally created the WU in the ZAP report.

I'd have to check, but I thought what you were passing through to create thew new workunit was the current user, so same as user importing, and therefore original owner was lost?
If that's so, then no need to store separately in Debug/@imported.

common/workunit/workunit.cpp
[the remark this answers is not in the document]
wangkx: @jakesmith I am adding the code to set a 'pattern for slave log' when importing a WU. Should I also add some code to set the 'pattern for slave log' when a standard thor WU is created? The existing thgraphmanager.cpp calls addProcess() to create the /Progress/Thor section in the CJobManager::updateWorkUnitLog(). I am not sure that is the place to set the 'pattern for slave log' (no enough information about slave log?). Any suggestion?

esp/services/ws_workunits/ws_workunitsHelpers.cpp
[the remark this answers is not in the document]
jakesmith: Relating to comments elsewhere, my feeling is that this would be cleaner if all pertanent info. was captured in the workunit under Process/ including whether @singleLog.

You could then expect IConstWorkUnit::getLogs() to retrieve any log, including slave logs.
But you might want to extend it, so it can give you a specific slave log, e.g:
```
IStringIterator *CLocalWorkUnit::getLogs(const char *type, const char *instance, unsigned n) const
{
     // See if Process/<type>/<instance> has a @pattern .. If has @pattern, should also have @max (for # slaves)
     // if has pattern and n>0, construct log using pattern and n. Error if n>max.
     // if has pattern and n undefined/default, construct logs for all up to @max.
}
```

```
Owned<IStringIterator> logs = workunit->getLogs("Thor", "mythor", 5);
```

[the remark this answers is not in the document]
jakesmith: Also, the introduction of IConstWUThorLogInfo may be sensible, but it's questionable why there is now 2 different ways to get log info getWUThorLogInfo() and getLogs()

I suspect you can simplify, so that everything uses getWUThorLogInfo ?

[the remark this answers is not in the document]
jakesmith: And I think when simplified, your code WsWuInfo::getWorkunitThorSlaveLogSingleFile code may go away.
With everything universally using IConstWUThorLogInfo (which will encode whether @snigle or not too), then your code based on IConstWUThorLogInfo::isSingleLog(), calling your getWorkunitLogSingleFile

common/workunit/workunit.cpp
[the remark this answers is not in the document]
jakesmith: Yes, see other comments, I think if you capture enough info (including #slaves in Thor case) via addProcess, then things become a lot cleaner.
Thor knows the pattern of the slave log (which will change at some point), so yes updateWorkUnitLog is a reasonable starting point.

I think I'd add @max (for #slaves), and @pattern for when retrieving/building slave logs, e.g. with:
```
void addProcess(const char *type, const char *instance, unsigned pid, const char *log, unsigned max, const char *pattern)
```

jakesmith (changes requested):

jakesmith:
@wangkx - have added more comments.

There is still probably some significant changes to make this simpler and more encapsulated.
This PR is getting a bit bloated with comments etc.

Can you look at latest comments and close this PR and open a new one when ready?

common/workunit/workunit.cpp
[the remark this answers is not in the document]
wangkx: OK, I may change the scope in the old WU XML PTree to be the userid of the importing user before importing it.

common/workunit/workunit.cpp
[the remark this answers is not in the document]
wangkx: I am changing the WsWorkunits code to depr_ver the log date from Thor Log Info.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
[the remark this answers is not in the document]
wangkx: Will check

wangkx:
I am closing this PR. A new PR will be opened after more changes to make this simpler and more encapsulated.

PR #13112, opened 2019-10-30
HPCC-23050  Write location and file name with analyzer issues

shamser:
## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/common/wuanalysis/anacommon.cpp
+++ b/common/wuanalysis/anacommon.cpp
@@ -14,6 +14,7 @@
     See the License for the specific language governing permissions and
     limitations under the License.
 ############################################################################## */
+#include <string.h>
 #include "anacommon.hpp"
 #include "workunit.hpp"
 
@@ -36,7 +37,12 @@ void PerformanceIssue::print() const
 {
     StringBuffer out;
     formatStatistic(out, cost, SMeasureTimeNs);
-    printf("[%s] E%d %s: %s\n", out.str(), errorCode, scope.str(), comment.str());
+    printf("[%s] E%d \"%s\" %s ", out.str(), errorCode,  comment.str(), scope.str());
+    if (filename.length()>0)
+        printf("%s", filename.str());
+    if (line>0 && column>0)
+        printf("(%u,%u)", line, column);
+    printf("\n");
 }
 
 void PerformanceIssue::createException(IWorkUnit * wu)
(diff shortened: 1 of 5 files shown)


shamser:
@ghalliday

common/wuanalysis/anacommon.cpp
```diff
@@ -46,7 +47,12 @@ void PerformanceIssue::createException(IWorkUnit * wu)
     we->setSeverity(SeverityInformation);
     we->setScope(scope.str());
     we->setPriority((unsigned) statUnits2msecs(cost));
-    we->setExceptionMessage(comment.str());
+    we->setExceptionLineNo(line);
+    we->setExceptionColumn(column);
+    we->setExceptionFileName(filename);
```
ghalliday: i think this will set it to blank rather than null if it is not known.  Is that a problem?

common/wuanalysis/anawu.cpp
```diff
@@ -247,7 +247,7 @@ void WuScope::getAttr(StringBuffer & result, WuAttr attr) const
 {
     StringBuffer name;
     name.append('@').append(queryWuAttributeName(attr));
-    attrs->getProp(result, name);
+    result.set(attrs->queryProp(name));
```
ghalliday: out of interest, how does this change the behaviour?

ghalliday (commented):
@shamser one comment and 1 question.

common/wuanalysis/anacommon.cpp
shamser: I don't think it would cause a problem but I'll modify to only set if non-empty string.

common/wuanalysis/anawu.cpp
shamser: getProp isn't returning anything (empty string). Could there be aproblem with getProp?  Eventually, I modified to use queryProp and it is working.

common/wuanalysis/anacommon.cpp
shamser: getProp isn't returning anything (empty string). Could there be a problem with getProp?  Eventually, I modified to use queryProp and it is working.

common/wuanalysis/anawu.cpp
shamser: getProp isn't returning any value...I've created a jira for this: https://track.hpccsystems.com/browse/HPCC-23072.  I'll investigate later.

shamser:
@ghalliday Code modified to not write filename if empty or line/column if either one is 0

common/wuanalysis/anawu.cpp
shamser: Jake has found the problem. I had getProp parameters the wrong way round

ghalliday: That's counterintuitive!  I would have a made the same mistake.

ghalliday:
@shamser, all looks good.  Please squash and I will merge.

PR #13116, opened 2019-10-31
HPCC-23067 Fix issues with strcmp/stricmp

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/common/workunit/pkgimpl.hpp
+++ b/common/workunit/pkgimpl.hpp
@@ -57,7 +57,7 @@ public:
     {
         ForEachItemIn(idx, subFiles)
         {
-            if (stricmp(subFiles.item(idx).queryProp("@value"), subname))
+            if (strieq(subFiles.item(idx).queryProp("@value"), subname))
                 return idx;
         }
         return NotFound;
--- a/dali/base/dafdesc.cpp
+++ b/dali/base/dafdesc.cpp
@@ -82,7 +82,7 @@ bool getCrcFromPartProps(IPropertyTree &fileattr,IPropertyTree &props, unsigned 
     }
     // NB: old @crc keys and compressed were not crc of file but of data within.
     const char *kind = props.queryProp("@kind");
-    if (kind&&strcmp(kind,"key")) // key part
+    if (strsame(kind,"key")) // key part
         return false;
     bool blocked;
     if (isCompressed(fileattr,&blocked)) {
--- a/esp/services/ws_dfu/ws_dfuService.cpp
+++ b/esp/services/ws_dfu/ws_dfuService.cpp
@@ -516,7 +516,7 @@ bool CWsDfuEx::onDFUSpace(IEspContext &context, IEspDFUSpaceRequest & req, IEspD
         else
         {
             Owned<IEspSpaceItem> item64 = createSpaceItem();
-            if (stricmp(countby, COUNTBY_OWNER))
+            if (!strieq(countby, COUNTBY_OWNER))
             {
                 if (scopeName)
                     item64->setName(scopeName);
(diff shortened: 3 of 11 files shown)


ghalliday:
@richardkchapman please review.  I suspect these fixes should target earlier branches, but here for discussion.

dali/base/dafdesc.cpp
```diff
@@ -82,7 +82,7 @@ bool getCrcFromPartProps(IPropertyTree &fileattr,IPropertyTree &props, unsigned
     }
     // NB: old @crc keys and compressed were not crc of file but of data within.
     const char *kind = props.queryProp("@kind");
-    if (kind&&strcmp(kind,"key")) // key part
+    if (kind&&streq(kind,"key")) // key part
```
ghalliday: @jakesmith what effect would this have?

esp/services/ws_dfu/ws_dfuService.cpp
```diff
@@ -516,7 +516,7 @@ bool CWsDfuEx::onDFUSpace(IEspContext &context, IEspDFUSpaceRequest & req, IEspD
         else
         {
             Owned<IEspSpaceItem> item64 = createSpaceItem();
-            if (stricmp(countby, COUNTBY_OWNER))
+            if (stricmp(countby, COUNTBY_OWNER))  // MORE: this looks like it should be strieq
```
ghalliday: @wangkx please can you check this

common/workunit/pkgimpl.hpp
```diff
@@ -57,7 +57,7 @@ class CPackageSuperFileArray : implements ISimpleSuperFileEnquiry, public CInter
     {
         ForEachItemIn(idx, subFiles)
         {
-            if (stricmp(subFiles.item(idx).queryProp("@value"), subname))
+            if (strieq(subFiles.item(idx).queryProp("@value"), subname))
```
richardkchapman: What would the effect of this be? @afishbeck ?

esp/services/ws_dfu/ws_dfuService.cpp
wangkx: The logic here is: if countby is not COUNTBY_OWNER. You may change it to !strieq.

common/workunit/pkgimpl.hpp
afishbeck: Looks like this is probably never called from here... it's emulating functionality that exists in the dali interface.  If it were called it wouldn't have worked, and this would be a fix.

afishbeck: @richardkchapman

richardkchapman:
@ghalliday I think this is back with you

dali/base/dafdesc.cpp
jakesmith: it was the wrong way around.
The effect is that it would not be returning the crc values here when it should have been (should have been for non key types).
This is called from DFS in a few places, I think it going to mean if crc checking was on, it wouldn't work due to this.

jakesmith: Might be worth changing to:
```
if (strsame(kind, "key"))
```

ghalliday:
I've updated to take jake and kevins comments.  @richardkchapman please review.

richardkchapman:
@ghalliday Looks reasonable. Only question is should we split out cases that fix significant bugs (e.g. httpcall non-functional, if I'm reading the fix correctly) and target them to earlier builds/add test cases...

PR #13118, opened 2019-11-01
HPCC-22909 WUDetails - metrics - large percentages

shamser:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jstats.cpp
+++ b/system/jlib/jstats.cpp
@@ -377,7 +377,14 @@ static StringBuffer & formatLoad(StringBuffer & out, unsigned __int64 value)
 static StringBuffer & formatSkew(StringBuffer & out, unsigned __int64 value)
 {
     //Skew stored as 10000 = perfect, display as percentage
-    return out.appendf("%.2f%%", ((double)(__int64)value) / 100.0);
+    const __int64 sval = (__int64) value;
+    const double percent = ((double)sval) / 100.0;
+    if (sval >= 10000 || sval <= -10000) // For values >= 100%, whole numbers
+        return out.appendf("%.0f%%", percent);
+    else if (sval >= 1000 || sval <= -1000) // For values >= 10%, 1 decimal point
+        return out.appendf("%.1f%%", percent);
+    else                                 // Anything < 10%, 2 decimal points
+        return out.appendf("%.2f%%", percent);
 }
 
 static StringBuffer & formatIPV4(StringBuffer & out, unsigned __int64 value)


shamser:
@ghalliday

system/jlib/jstats.cpp
```diff
@@ -377,7 +377,13 @@ static StringBuffer & formatLoad(StringBuffer & out, unsigned __int64 value)
 static StringBuffer & formatSkew(StringBuffer & out, unsigned __int64 value)
 {
     //Skew stored as 10000 = perfect, display as percentage
-    return out.appendf("%.2f%%", ((double)(__int64)value) / 100.0);
+    const double percent = ((double)(__int64)value) / 100.0;
+    if (value > 1000)      // > 10% => display whole numbers only
```
ghalliday: @jakesmith what do you think?
@shamser really picky, but are these the correct thresholds?  It means that you have at most 2 digits, and often only 1 and a bit significant digits.  E.g., 105 is displayed the same as 114.  I think that loses too much informatin, so I would have an extra digit, (and use >= rather than >).

shamser: I'm not sure what you mean by 105 is displayed same as 114.  Value 105 will be displayed as 1.0% and 114 will be displayed as 1.1%... should we have higher precision for small numbers?

ghalliday: 105 will also be displayed as 1.1% (since it is floating point and will round up).  If it was truncated then 1.10% and 1.19% would be displayed the same.  Having an accuracy of only 10% for some numbers isn't really high enough.

shamser: Sorry, my mistake.  Should we have 2 digits of precision for anything less than 10%?

ghalliday: Yes, I think multiply all your test values by 10, but also make the test >= so that the number of digits stays consistent from 9.99% to 10.0%

jakesmith: I'm not sure it matters much either way (I can't think where I've looked at graph/skew and cared about such a small skew, or that marginally different).

But don't have a problem with 2 decimal places consistently either. 
I think the original request was on silly large numbers 10000's% with decimals, which leads to not nice display issues.
So the main issue is that they get long with no benefit.
I suspect trying to limit to 5 digits in total is good enough, so >=10000% whole num, >=1000, to 1 decimal digit, < 1000 = 2 decimal digits.

shamser:
@ghalliday @jakesmith

system/jlib/jstats.cpp
```diff
@@ -377,7 +377,14 @@ static StringBuffer & formatLoad(StringBuffer & out, unsigned __int64 value)
 static StringBuffer & formatSkew(StringBuffer & out, unsigned __int64 value)
 {
     //Skew stored as 10000 = perfect, display as percentage
-    return out.appendf("%.2f%%", ((double)(__int64)value) / 100.0);
+    const __int64 & sval = (__int64) value;
```
ghalliday: @shamser Why the reference?  That looks like a bug to me - assigning a temporary value to a reference.

shamser: I'm using it as a quick way to cast from unsigned to signed.  I need the signed for both percent calculation and for comparing the range.  I can do it without using reference but there'll be 2 copies of the value.

jakesmith: because of the if/else >= , <= would calling:
```
unsigned __int64 aval := labs((__int64)value);
```
then you only need e.g. if (aval >= 10000)

system/jlib/jstats.cpp
```diff
@@ -377,7 +377,14 @@ static StringBuffer & formatLoad(StringBuffer & out, unsigned __int64 value)
...
+    const double percent = ((double)sval) / 100.0;
+    if (sval >= 10000 || sval <= -10000)
```
jakesmith: for the code reader, I think adding back comments that say // >= 100% or similar helps.

system/jlib/jstats.cpp
ghalliday: if you are doing that, then you need to use std::abs(x).  long may be smaller than __int64.

shamser: I could take the absolute value for the comparison but I need the -ve value for display.

shamser:
@ghalliday @jakesmith Please can you review

jakesmith:
@shamser - I think I'd have used std::abs() to simplify conditions, but looks fine as is.

PR #13123, opened 2019-11-02
HPCC-23077 Pipe crash in run if pipethread exits (race condition)

mckellyln:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

## Testing:

Smoke test and manual pipefail runs.

### DIFF
--- a/system/jlib/jthread.cpp
+++ b/system/jlib/jthread.cpp
@@ -2092,13 +2092,15 @@ public:
         title.clear();
         prog.set(_prog);
         dir.set(_dir);
-        if (_title) {
+        if (_title)
+        {
             title.set(_title);
             PROGLOG("%s: Creating PIPE program process : '%s' - hasinput=%d, hasoutput=%d stderrbufsize=%d", title.get(), prog.get(),(int)hasinput, (int)hasoutput, stderrbufsize);
         }
         CheckAllowedProgram(prog,allowedprogs);
         retcode = 0;
-        if (forkthread) {
+        if (forkthread)
+        {
             {
                 CriticalUnblock unblock(sect); 
                 forkthread->join();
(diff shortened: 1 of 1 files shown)


mckellyln:
@jakesmith please review.

jakesmith:
Looks good.

richardkchapman:
Not sure I see why it would crash before?

jakesmith:
I thought I had, but now can't.
I had thought (as per Mark's comments in JIRA), the "forkthread.clear()", could destroy the running thread just before it called closeOutput().

But looking again, it should remain alive afaics, Thread start links itself and _threadmain will ultimately release/destroy it, if nobody else has a link.
So afaics, it should always be valid for the duration of Thread::run()

@mckellyln - can you elaborate on how the crash happens?

mckellyln:
I thought closeOutput() was being called when forkthread was no longer valid.
I will try yo reproduce and report more info.

jakesmith:
'forkthread' will be null/invalid after the clear, but the object it referenced should still be valid afaics.

mckellyln:
It may not be from the forkthread.clear().  It sure seems CLinuxPipeProcess can go out of scope before closeOutput() is called.  Could the pipewriter end before this thread calls closeOutout() ?

jakesmith:
>Could the pipewriter end before this thread calls closeOutput() ?

PipeWriterThread? I suspect it can, but don't see how related?

mckellyln:
It is because of this -
```
    ~CPipeThroughSlaveActivity()
    {
        ::Release(pipeWriter);
    }
```
That the CLinuxPipeProcess goes out of scope - and thus the closeOutput() call causes the segv.

ghalliday:
I think the fix is correct, but the reason is slightly different.  If retcode has been set but the thread is still running then forkthread will be cleared.  That will prevent the destructor of the pipe object from joining the thread.  When the thread finally wakes up the object will be destroyed leading to a crash.

I suspect only likely on systems with very few (1) cores.

richardkchapman:
@mckellyln Should this target 7.6.x ?

ghalliday:
I think it probably should.  Any "safe" fixes for crashes should target the next point release.

mckellyln:
ok, I'll retarget 7.6.x

PR #13124, opened 2019-11-04
HPCC-22566 Avoid race causing pipe write exception

jakesmith:
If the pipe program fails to start, it should report an
exec failure, but due to a race condition with the pipe writer
thread, it could have reported a pipe write exception instead.
The pipe write exception still occurs, but only after the pipe
exec has already been reported (thrown). However the pipe
activity was _also_ reporting the exception from pipe write if
it got there early enough. It should have been ignored,
because the verifyPipe had already thrown.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/thorlcr/activities/piperead/thprslave.cpp
+++ b/thorlcr/activities/piperead/thprslave.cpp
@@ -38,6 +38,7 @@ protected:
     bool eof, pipeFinished;
     unsigned retcode;
     unsigned flags;
+    Owned<IException> verifyPipeException;
 
 protected:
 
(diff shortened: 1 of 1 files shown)


richardkchapman:
@ghalliday Please review

jakesmith:
@mckellyln - please also review.

thorlcr/activities/piperead/thprslave.cpp
```diff
@@ -522,6 +520,14 @@ PipeWriterThread::PipeWriterThread(CPipeThroughSlaveActivity & _activity)
     inputStream = activity.inputStream;
 }
 
+IException *PipeWriterThread::joinExc()
+{
+    Thread::join();
+    if (activity.verifyPipeException) // if verifyPipe failure, ignore [follow-on] pipe write failure.
```
ghalliday: Will verify pipe have been called before this function?  It looks to me that it is called after (in CPipeThroughSlaveActivity::stop).
Possibly a cleaner fix is

        if (wrexc && !verifyPipeException)

in CPipeThroughSlaveActivity::stop, since the writer shouldn't really need to access activity.

ghalliday (commented):
@jakesmith one question/comment

thorlcr/activities/piperead/thprslave.cpp
jakesmith: verifyPipe will [normally] have been called before this on error, via CPipeStream::read() (via nextRow).
The call to verifyPipe in stop() will normally be a nop (pipeFinished already true).
But if nextRow is never called, then verifyPipe is called afterwards, but then the exception it generates would be thrown directly (before wrexc).

I agree though that testing verifyPipeException in stop() would be better.

jakesmith:
@ghalliday - please re-review

ghalliday:
@jakesmith change looks good.  Any reason this shouldn't go in 7.6.x.?

jakesmith:
Not sure why I targeted master.. I will retarget.

jakesmith:
@ghalliday - rebased/retargeted to 7.6.x

richardkchapman:
@ghalliday Please review

PR #13134, opened 2019-11-07
HPCC-23080 Fix various undefined behaviours caught by sanitize option

ghalliday:
Never pass nullptr to memset/memcmp/memcpy:
  Note, most of the new gurad conditions check if the length is zero, rather
  than if the pointers are null.  That is because the pointers can only be
  null if th length is zero, and I think the test is more intuitive.

Never shift -ve integers left

Fix potential issues with overflow of signed addition/subtraction

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/plugins/stringlib/stringlib.cpp
+++ b/plugins/stringlib/stringlib.cpp
@@ -421,7 +421,7 @@ STRINGLIB_API void STRINGLIB_CALL slStringSubsOut(unsigned & tgtLen, char * & tg
     }
     else
     {
-        memcpy(tgt, src, srcLen);
+        memcpy_iflen(tgt, src, srcLen);
     }
 
     tgtLen = srcLen;
@@ -452,7 +452,7 @@ STRINGLIB_API void STRINGLIB_CALL slStringSubs(unsigned & tgtLen, char * & tgt, 
     }
     else
     {
-        memcpy(tgt, src, srcLen);
+        memcpy_iflen(tgt, src, srcLen);
     }
 
     tgtLen = srcLen;
@@ -503,7 +503,7 @@ STRINGLIB_API void STRINGLIB_CALL slStringRepad(unsigned & tgtLen, char * & tgt,
         if (!tgt)
             rtlThrowOutOfMemory(0, "In StringLib.StringRepad");
         tgtLen = tLen;
-        memcpy(tgt,base,srcLen);
+        memcpy_iflen(tgt,base,srcLen);
         memset(tgt+srcLen,' ',tLen-srcLen);
     }
     else
@@ -610,7 +610,7 @@ STRINGLIB_API void STRINGLIB_CALL slStringExtract(unsigned & tgtLen, char * & tg
         if ( finger[len] == ',' )
             break;
     tgt = (char *)CTXMALLOC(parentCtx, len);
-    memcpy(tgt,finger,len);
+    memcpy_iflen(tgt,finger,len);
     tgtLen = len;
 }
 
@@ -847,7 +847,7 @@ STRINGLIB_API void STRINGLIB_CALL slStringFindReplace (unsigned & tgtLen, char *
     if ( srcLen < stokLen || stokLen == 0)
     {
         tgt = (char *) CTXMALLOC(parentCtx, srcLen);
-        memcpy(tgt, src, srcLen);
+        memcpy_iflen(tgt, src, srcLen);
         tgtLen = srcLen;
     }
     else
@@ -1129,8 +1129,8 @@ STRINGLIB_API void STRINGLIB_CALL slStringExcludeNthWord(unsigned & tgtLen, char
     if (len)
     {
         tgt = (char *)CTXMALLOC(parentCtx, len);
-        memcpy(tgt,src,startLast);
-        memcpy(tgt+startLast,src+idx,(srcLen - idx));
+        memcpy_iflen(tgt,src,startLast);
+        memcpy_iflen(tgt+startLast,src+idx,(srcLen - idx));
     }
     else
         tgt = NULL;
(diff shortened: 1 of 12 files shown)


ghalliday:
@richardkchapman please review.  Easier to compare when ignoring whitespace.
Note, there is also a PR against the LN repo: https://github.com/hpcc-systems/LN/pull/950

ecl/hql/hqlexpr.cpp
```diff
@@ -9873,6 +9873,8 @@ IHqlScope * CHqlVirtualScope::deriveConcreteScope()
             IHqlExpression *cur = symbols.mapToValue(&iter.query());
             OwnedHqlExpr mapped = replacer.transform(cur);
             OwnedHqlExpr rebound = cloneFunction(mapped);
+            if (rebound->isFunction() && rebound->queryChild(1)->numChildren() != 0 && (mapped == rebound), false)
```
richardkchapman: This change does not look like the others (and the && ... , false is rather odd?

plugins/stringlib/stringlib.cpp
```diff
@@ -610,7 +613,8 @@ STRINGLIB_API void STRINGLIB_CALL slStringExtract(unsigned & tgtLen, char * & tg
         if ( finger[len] == ',' )
             break;
     tgt = (char *)CTXMALLOC(parentCtx, len);
-    memcpy(tgt,finger,len);
+    if (likely(len))
+        memcpy(tgt,finger,len);
```
richardkchapman: trivial and not new, but spaces after commas would be nice

rtl/eclrtl/eclrtl.cpp
```diff
@@ -1287,16 +1310,22 @@ bool rtlDataToBool(unsigned len, const void * _src)
...
 void rtlBoolToData(unsigned tlen, void * tgt, bool src)
 {
-    memset(tgt, 0, tlen);
-    if (src)
-        ((char *)tgt)[tlen-1] = 1;
+    if (likely(tlen))
+    {
+        memset(tgt, 0, tlen);
+        if (src)
+            ((char *)tgt)[tlen-1] = 1;
```
richardkchapman: This would potentially have corrupted before, I think?

system/jlib/jbuff.cpp
```diff
@@ -174,7 +174,10 @@ MemoryAttr::MemoryAttr(const MemoryAttr & src)
 
 void MemoryAttr::set(size_t _len, const void * _ptr)
 {
-    memcpy(allocate(_len), _ptr, _len);
+    if (_len)
```
richardkchapman: Does this change the existing behaviour?

richardkchapman:
An alternative approach would have been to add an inline function memcpy0() that checked for zero length. Would it have made code clearer?

richardkchapman:
@ghalliday A few minor comments

ecl/hql/hqlexpr.cpp
ghalliday: no, it wasn't meant to be included.  It was an internal consistency check part of HPCC-23091.

system/jlib/jbuff.cpp
ghalliday: I've just double checked, and it will behave the same.

rtl/eclrtl/eclrtl.cpp
ghalliday: I guess it would - if you had cast a boolean to a string0.

richardkchapman:
There's a typo (gurad) in the commit message

rtl/eclrtl/eclrtl.cpp
richardkchapman: Somebody must have done so in your suite for you to have caught this (assuming it was changed because you saw an issue from sanitize)

ghalliday: When I hit an error with memset/memcmp/memcpy in any file I then searched the file for other potential issues.  This is almost certainly one of those.

ghalliday:
@richardkchapman I quite like the idea of null-safe versions of the functions.  Should I switch to using that for at least some of these?  No reason this has to be rushed.

richardkchapman:
@ghalliday Up to you

ghalliday:
@richardkchapman please re-review

rtl/eclrtl/eclrtl.cpp
```diff
@@ -5562,14 +5599,17 @@ ECLRTL_API void rtlCreateUnicodeRange(size32_t & outlen, UChar * & out, unsigned
...
+    }
     else
     {
-        memcpy(out, str, len * sizeof(UChar));
+        memcpy_iflen(out, str, len * sizeof(UChar));
         while (len != compareLen)
             out[len++] = ' ';
     }
-    memset(out + compareLen, fill, (fieldLen-compareLen) * sizeof(UChar));
+    if (fieldLen != compareLen)
```
richardkchapman: Greater-than feels more appropriate than != here...

richardkchapman:
@ghalliday One minor comment. Also, why no memcmp_iflen ?

Is 7.6.x the best target here?

ghalliday:
@richardkchapman added memcmp_iflen.
I think 7.6.x is the best target.  I don't think any of the issues would ever cause a bug - except casting a boolean to a string0, so no real need to target an earlier build.

richardkchapman:
@ghalliday Please squash. And if you felt like changing that != at the same time...

rtl/eclrtl/eclrtl.cpp
ghalliday: @richardkchapman 
I have changed it - although it couldn't have been < compareLen (see the test at the head of the function).  Please check I've got it correct.

PR #13137, opened 2019-11-07
HPCC-22498 Filter WUs on Total Cluster Time

wangkx:
Similar function has been implemented in cassandrawu.

One bug is fixed in the fetchDataByThorTime() of the cassandrawu.cpp: 
the 'where fieldValue' in the existing line 4302 should be changed to 
'and fieldValue'.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/common/workunit/workunit.cpp
+++ b/common/workunit/workunit.cpp
@@ -5705,6 +5705,12 @@ public:
                         query.append("=?~\"").append(fv).append('\"');
                     query.append("]");
                 }
+                else if (subfmt==WUSFtotalthortime)
+                {
+                    query.append("[@totalThorTime>=\"");
+                    formatTimeCollatable(query, milliToNano(atoi(fv)), false);
+                    query.append("\"]");
+                }
                 else if (!*fv)
                 {
                     unknownAttributes.append(getEnumText(subfmt,workunitSortFields));
--- a/esp/scm/ws_workunits.ecm
+++ b/esp/scm/ws_workunits.ecm
@@ -507,6 +507,7 @@ ESPrequest [nil_remove] WUQueryRequest
     [min_ver("1.57")] ESParray<ESPstruct ApplicationValue> ApplicationValues;
     [min_ver("1.72")] string BeforeWU;
     [min_ver("1.72")] string AfterWU;
+    [min_ver("1.77")] unsigned TotalClusterTimeThresholdMilliSec(0);
 
     [depr_ver("1.02")] string After; //Not used since 1.02
     [depr_ver("1.02")] string Before; //Not used since 1.02
--- a/esp/services/ws_workunits/ws_workunitsService.cpp
+++ b/esp/services/ws_workunits/ws_workunitsService.cpp
@@ -1751,6 +1751,17 @@ bool addWUQueryFilterTime(WUSortField *filters, unsigned short &count, MemoryBuf
     return true;
 }
 
+bool addWUQueryFilterTotalClusterTime(WUSortField *filters, unsigned short &count, MemoryBuffer &filterBuf, unsigned milliseconds, WUSortField value)
+{
+    if (milliseconds == 0)
+        return false;
+
+    VStringBuffer vBuf("%u", milliseconds);
+    filters[count++] = value;
+    filterBuf.append(vBuf);
+    return true;
+}
+
 bool addWUQueryFilterApplication(WUSortField *filters, unsigned short &count, MemoryBuffer &buff, const char *appname, const char *appkey, const char *appdata)
 {
     if (isEmpty(appname))
@@ -1857,6 +1868,8 @@ void doWUQueryWithSort(IEspContext &context, IEspWUQueryRequest & req, IEspWUQue
     addWUQueryFilter(filters, filterCount, filterbuf, req.getOwner(), (WUSortField) (WUSFuser | WUSFnocase));
     addWUQueryFilter(filters, filterCount, filterbuf, req.getJobname(), (WUSortField) (WUSFjob | WUSFnocase));
     addWUQueryFilter(filters, filterCount, filterbuf, req.getECL(), (WUSortField) (WUSFecl | WUSFwild));
+
+    addWUQueryFilterTotalClusterTime(filters, filterCount, filterbuf, req.getTotalClusterTimeThresholdMilliSec(), WUSFtotalthortime);
 
     addWUQueryFilterTime(filters, filterCount, filterbuf, req.getStartDate(), WUSFwuid);
     addWUQueryFilterTime(filters, filterCount, filterbuf, req.getEndDate(), WUSFwuidhigh);
--- a/plugins/cassandra/cassandrawu.cpp
+++ b/plugins/cassandra/cassandrawu.cpp
@@ -3417,6 +3417,22 @@ public:
         bool sortByThorTime = (baseSort == WUSFtotalthortime);
         bool needsPostSort = (baseSort != WUSFwuid && baseSort != WUSFtotalthortime);
         bool sortDescending = (sortorder & WUSFreverse) || needsPostSort;
+
+        bool keepThorTimeFilter = sortByThorTime;
+        if (!keepThorTimeFilter)
+        {
+            const WUSortField *filterPtr = filters;
+            while (filterPtr && *filterPtr)
+            {
+                WUSortField field = (WUSortField) (*filterPtr & 0xff);
+                if (field == WUSFtotalthortime)
+                {
+                    keepThorTimeFilter = true;
+                    break;
+                }
+            }
+        }
+
         if (!result)
         {
             Owned<CassMultiIterator> merger = new CassMultiIterator(needsPostSort ? NULL : cached, 0, 0, sortDescending); // We always merge by wuid (except when we merge by thor time... we turn the compare off then to make it an appender)
@@ -3426,7 +3442,7 @@ public:
                 unsigned found = cached->lookupStartRow(startWuid, thorTimeThreshold, startOffset);
                 if (found)
                 {
-                    if (!sortByThorTime)
+                    if (!keepThorTimeFilter)
                     {
                         if (sortDescending)
                             startWuid.setCharAt(startWuid.length()-1, startWuid.charAt(startWuid.length()-1)-1);  // we want to find the last wuid BEFORE
@@ -3492,7 +3508,7 @@ public:
                 case WUSFwuid: // Acts as wuidLo when specified as a filter
                 case WUSFwuidhigh:
                     // Wuid filters can be added to good and poor filters, and to remoteWild if they are done via merged sets rather than ranges...
-                    if (sortByThorTime)
+                    if (keepThorTimeFilter)
                         remoteWildFilters.append(*new PostFilter(field, fv, true));
                     else
                         mergeFilter(wuidFilters, field, fv);
@@ -3554,7 +3570,7 @@ public:
                     merger->addResult(*new CassandraResult(fetchDataForWuid(workunitInfoMappings, value, true)));
                 }
             }
-            else if (sortByThorTime)
+            else if (sortByThorTime || !thorTimeThreshold.isEmpty())
             {
                 merger->addPostFilters(goodFilters, 0);
                 merger->addPostFilters(poorFilters, 0);
@@ -4299,7 +4315,7 @@ private:
         getFieldNames(searchMappings+3, names, tableName);  // mappings+3 means we don't return the key columns (xpath, upper(keyPrefix), upper(key))
         VStringBuffer selectQuery("select %s from %s where xpath=? and fieldPrefix=?", names.str()+1, tableName.str());
         if (threshold && *threshold)
-            selectQuery.appendf(" where fieldValue >= ?");
+            selectQuery.appendf(" and fieldValue >= ?");
         if (descending)
             selectQuery.append(" ORDER BY fieldValue DESC, wuid ASC");
         else


wangkx:
@afishbeck please review.

esp/scm/ws_workunits.ecm
```diff
@@ -507,6 +507,7 @@ ESPrequest [nil_remove] WUQueryRequest
     [min_ver("1.57")] ESParray<ESPstruct ApplicationValue> ApplicationValues;
     [min_ver("1.72")] string BeforeWU;
     [min_ver("1.72")] string AfterWU;
+    [min_ver("1.77")] unsigned TotalClusterTimeThresholdMilliSeconds(0);
```
afishbeck: It's interesting I see 4 examples of "milliseconds" in this file:
TimeoutMilliSec
ElapsedTimeMs
TotalClusterTimeThresholdMilliSeconds
MS
I wonder if "TotalClusterTimeThresholdMs" is obvious enough?  If not, could standardize on the other one that already exists "TotalClusterTimeThresholdMilliSec"?

esp/services/ws_workunits/ws_workunitsService.cpp
```diff
@@ -1751,6 +1751,19 @@ bool addWUQueryFilterTime(WUSortField *filters, unsigned short &count, MemoryBuf
...
 }
 
+bool addWUQueryFilterTotalClusterTime(WUSortField& sortorder, WUSortField *filters, unsigned short &count, MemoryBuffer &buff, unsigned milliseconds, WUSortField value)
+{
+    if (milliseconds == 0)
+        return false;
+
+    VStringBuffer vBuf("%d", milliseconds);
+    filters[count++] = value;
+    buff.append(vBuf);
```
afishbeck: Might be better to have buff as "filterbuf" like the caller, just to make it obvious.

afishbeck (commented):
@wangkx A couple thoughts on naming, otherwise looks good.

wangkx:

afishbeck (approved):
@wangkx looks good.

wangkx:

richardkchapman:
@afishbeck @jakesmith Any efficiency concerns?

esp/services/ws_workunits/ws_workunitsService.cpp
```diff
@@ -1751,6 +1751,19 @@ bool addWUQueryFilterTime(WUSortField *filters, unsigned short &count, MemoryBuf
...
+bool addWUQueryFilterTotalClusterTime(WUSortField& sortorder, WUSortField *filters, unsigned short &count, MemoryBuffer &filterBuf, unsigned milliseconds, WUSortField value)
+{
+    if (milliseconds == 0)
+        return false;
+
+    VStringBuffer vBuf("%d", milliseconds);
+    filters[count++] = value;
+    filterBuf.append(vBuf);
+
+    sortorder = WUSFtotalthortime;//Cassandra WUs can be filtered by TotalClusterTime only if sorting by TotalClusterTime.
```
jakesmith: @wangkx - does this mean it will be unnecessarily sorting in the Dali case?

It may not be significant timing wise, but it would be good to avoid steps that are only applicable to one implementation.

plugins/cassandra/cassandrawu.cpp
```diff
@@ -4299,7 +4299,7 @@ class CCasssandraWorkUnitFactory : public CWorkUnitFactory, implements ICassandr
         getFieldNames(searchMappings+3, names, tableName);  // mappings+3 means we don't return the key columns (xpath, upper(keyPrefix), upper(key))
         VStringBuffer selectQuery("select %s from %s where xpath=? and fieldPrefix=?", names.str()+1, tableName.str());
         if (threshold && *threshold)
-            selectQuery.appendf(" where fieldValue >= ?");
+            selectQuery.appendf(" and fieldValue >= ?");
```
jakesmith: @richardkchapman - not really sure re. reasons/implications of this change - is it correct/ok?

jakesmith:
@wangkx - One question re. performance

plugins/cassandra/cassandrawu.cpp
richardkchapman: It looks like a bugfix to me (selectQuery already has a where clause, see 2 lines earlier)

esp/services/ws_workunits/ws_workunitsService.cpp
wangkx: @jakesmith without this line,  the 'filtered by TotalClusterTime' will not work for CassandraWU. Should I change that logic in cassandrawu.cpp?

jakesmith: @richardkchapman - can you check the logic here re. cassandra impl. needing to add a sort order in this case particularly?

esp/services/ws_workunits/ws_workunitsService.cpp
```diff
@@ -1858,6 +1871,8 @@ void doWUQueryWithSort(IEspContext &context, IEspWUQueryRequest & req, IEspWUQue
     addWUQueryFilter(filters, filterCount, filterbuf, req.getJobname(), (WUSortField) (WUSFjob | WUSFnocase));
     addWUQueryFilter(filters, filterCount, filterbuf, req.getECL(), (WUSortField) (WUSFecl | WUSFwild));
 
+    addWUQueryFilterTotalClusterTime(sortorder, filters, filterCount, filterbuf, req.getTotalClusterTimeThresholdMilliSec(), WUSFtotalthortime);
```
jakesmith: @wangkx - it looks to me, that if there is any other sort order read above (via readWUQuerySortOrder), then it will be lost in favour of WUSFtotalthortime.
Is that correct, and isn't that confusing from a UI perspective vs other fields?

At a minimum I think needs some comments here as to why this field is treated differently from others.

wangkx: > if there is any other sort order read above (via readWUQuerySortOrder), then it will be lost in favour of WUSFtotalthortime. Is that correct, and isn't that confusing from a UI perspective vs other fields?
Yes. It is correct. UI may have to add some text around the Input box of a FilterByTotalThorTime. 
>At a minimum I think needs some comments here as to why this field is treated differently from others.
There is a comment at the line 1763.

jakesmith: I guess my question was:
>//Cassandra WUs can be filtered by TotalClusterTime only if sorting by TotalClusterTime.

Why is that when not for other fields?

jakesmith: >Yes. It is correct. UI may have to add some text around the Input box of a FilterByTotalThorTime.

If the UI needs changes to clarify loss of sort order, is there a separate JIRA?
It's still not obvious from reading the code though, so there should be commend explaining the logic, and reasons behind it.

wangkx: > Why is that when not for other fields?
In fact, I am not familiar with CassandraWU code. I have not checked how other sort goes. 
 
> If the UI needs changes to clarify loss of sort order, is there a separate JIRA?
There is a JIRA for UI.

jakesmith: HPCC-23098 ? 

I might be misunderstanding.
When you filter by total cluster time, does it effect the display sort order?
That's going to be confusing if so, and it's not clear (to me) yet why (and not necessary in Dali case). And not mentioned in HPCC-23098 either.

wangkx: Yes HPCC-23098. I haven't mentioned this behavior there because I am not sure we will accept this behavior (but, I did mention it with both Miguel and Kanul). It will be clear if this PR is accepted.

> When you filter by total cluster time, does it effect the display sort order?

Yes. this is the behavior I am talking about.

wangkx:
@jakesmith @richardkchapman please review.

plugins/cassandra/cassandrawu.cpp
```diff
@@ -3554,6 +3554,13 @@ class CCasssandraWorkUnitFactory : public CWorkUnitFactory, implements ICassandr
                     merger->addResult(*new CassandraResult(fetchDataForWuid(workunitInfoMappings, value, true)));
                 }
             }
+            else if (!thorTimeThreshold.isEmpty())
```
jakesmith: this doesn't look right to me, because in lines 3569/3570, it is also using thorTimeThreshold,
but now, with this change, if set, it will never reach there, even if sortByThorTime is true.

jakesmith (changes requested):

jakesmith:
@wangkx - back to you

plugins/cassandra/cassandrawu.cpp
wangkx: @jakesmith the lines 3569/3570 will be reached if sortByThorTime is true, but, the thorTimeThreshold is empty.

common/workunit/workunit.cpp
```diff
@@ -5704,6 +5704,12 @@ class CDaliWorkUnitFactory : public CWorkUnitFactory, implements IDaliClientShut
                         query.append("=?~\"").append(fv).append('\"');
                     query.append("]");
                 }
+                else if (subfmt==WUSFtotalthortime)
+                {
+                    StringBuffer t;
+                    formatTimeCollatable(t, milliToNano(atoi(fv)), false);
```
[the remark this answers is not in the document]
wangkx: Will change to 'query'.

esp/services/ws_workunits/ws_workunitsService.cpp
```diff
@@ -1751,6 +1751,17 @@ bool addWUQueryFilterTime(WUSortField *filters, unsigned short &count, MemoryBuf
     return true;
 }
 
+bool addWUQueryFilterTotalClusterTime(WUSortField& sortorder, WUSortField *filters, unsigned short &count, MemoryBuffer &filterBuf, unsigned milliseconds, WUSortField value)
+{
+    if (milliseconds == 0)
+        return false;
+
+    VStringBuffer vBuf("%d", milliseconds);
```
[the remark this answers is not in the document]
wangkx: Will use '%u'

esp/services/ws_workunits/ws_workunitsService.cpp
```diff
@@ -1751,6 +1751,17 @@ bool addWUQueryFilterTime(WUSortField *filters, unsigned short &count, MemoryBuf
     return true;
 }
 
+bool addWUQueryFilterTotalClusterTime(WUSortField& sortorder, WUSortField *filters, unsigned short &count, MemoryBuffer &filterBuf, unsigned milliseconds, WUSortField value)
```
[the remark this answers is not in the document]
wangkx: Will remove

plugins/cassandra/cassandrawu.cpp
jakesmith: @wangkx , I was talking about the case where thorTimeThreshold is set and sortByThorTime is true.

With this change it will not follow the continuation code inside the 'if (wuidFilters.length())' block

wangkx: Make sense. I may remove my changes (line 3557-3563) and change line 3564 to:
else if (sortByThorTime || !thorTimeThreshold.isEmpty())
The wuidFilters should be used when !thorTimeThreshold.isEmpty() even if sortByThorTime  is false. Any objection?

wangkx:
@jakesmith please review.

plugins/cassandra/cassandrawu.cpp
```diff
@@ -3554,14 +3554,7 @@ class CCasssandraWorkUnitFactory : public CWorkUnitFactory, implements ICassandr
...
             }
-            else if (!thorTimeThreshold.isEmpty())
-            {
-                merger->addPostFilters(goodFilters, 0);
-                merger->addPostFilters(poorFilters, 0);
-                merger->addPostFilters(remoteWildFilters, 0);
-                merger->addResult(*new CassandraResult(fetchDataByThorTime(thorTimeThreshold, sortDescending, merger->hasPostFilters() ? 0 : pageSize+startOffset)));
-            }
-            else if (sortByThorTime)
+            else if (sortByThorTime || !thorTimeThreshold.isEmpty())
```
jakesmith: I notice (on line 3435) that thorTimeThreshold is cleared if !sortByThorTime ..

So still not sure about the logic here.
Has it been tested?

@wangkx

wangkx: The code is used for the pages after the first page (line 3423) and there is WU inside the cache (line 3427). Looks like that the code is to clear old thorTimeThreshold. After the code, the thorTimeThreshold is set as needed at line 3507/3508.  I am going to move line 3435 after line 3436.

wangkx:

jakesmith:
@wangkx - back to you re. Cassandra smoketest issue.

wangkx:
@AttilaVamos the HPCCSmoketest started 2 days ago, but, has not finished. Please help.

AttilaVamos:
@wangkx Sorry for this late reply. I don't know (yet) where the result went missing, but it was clean. I rescheduled this PR.

wangkx:
@jakesmith you may review now.

jakesmith:
>One bug is fixed.

@wangkx - can you update the PR and JIRA description to describe the bug fixed?

wangkx:
@jakesmith I just updated the PR description. I cannot change the JIRA description. But, I have added a comment on the JIRA.

jakesmith:
@wangkx - looks good, please squash and rebase onto latest candidate-7.6.x if needed.

wangkx:
Squash and rebased. @richardkchapman this PR is ready to be meged.

PR #13139, opened 2019-11-08
HPCC-23075 XML Spray loses header length if there is not footer present.

AttilaVamos:
Fix header size loosing bug.

Add regression test for
- Check xml header and footer is right in size when dataset write out as XML file (HPCC-23079)
- Check the despray doesn't lose any record if they are smaller than headers (HPCC-23046)
- Check the spray creates same file as the original was with proper header/footer size

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

## Testing:

Tested manually

### DIFF
--- a/dali/ft/filecopy.cpp
+++ b/dali/ft/filecopy.cpp
@@ -1757,7 +1757,21 @@ void FileSprayer::derivePartitionExtra()
 void FileSprayer::displayPartition()
 {
     ForEachItemIn(idx, partition)
+    {
         partition.item(idx).display();
+
+#ifdef _DEBUG
+        if ((partition.item(idx).whichInput >= 0) && (partition.item(idx).whichInput < sources.ordinality()) )
+            LOG(MCdebugInfoDetail, unknownJob,
+                     "   Header size: %" I64F "u, XML header size: %" I64F "u, XML footer size: %" I64F "u",
+                     sources.item(partition.item(idx).whichInput).headerSize,
+                     sources.item(partition.item(idx).whichInput).xmlHeaderLength,
+                     sources.item(partition.item(idx).whichInput).xmlFooterLength
+            );
+        else
+            LOG(MCdebugInfoDetail, unknownJob,"   No source file for this partition");
+#endif
+    }
 }
 
 
@@ -3000,27 +3014,19 @@ void FileSprayer::spray()
 bool FileSprayer::isSameSizeHeaderFooter()
 {
     bool retVal = true;
-    unsigned whichHeaderInput = 0;
-    bool isEmpty = true;
-    headerSize = 0;
-    footerSize = 0;
 
     if (sources.ordinality() == 0)
         return retVal;
+
+    unsigned whichHeaderInput = 0;
+    headerSize = sources.item(whichHeaderInput).xmlHeaderLength;
+    footerSize = sources.item(whichHeaderInput).xmlFooterLength;
 
     ForEachItemIn(idx, partition)
     {
         PartitionPoint & cur = partition.item(idx);
         if (cur.inputLength && (idx+1 == partition.ordinality() || partition.item(idx+1).whichOutput != cur.whichOutput))
         {
-            if (isEmpty)
-            {
-                headerSize = sources.item(whichHeaderInput).xmlHeaderLength;
-                footerSize = sources.item(cur.whichInput).xmlFooterLength;
-                isEmpty = false;
-                continue;
-            }
-
             if (headerSize != sources.item(whichHeaderInput).xmlHeaderLength)
             {
                 retVal = false;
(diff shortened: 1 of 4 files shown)


AttilaVamos:
@jakesmith please review.

dali/ft/filecopy.cpp
```diff
@@ -1757,7 +1757,21 @@ void FileSprayer::derivePartitionExtra()
 void FileSprayer::displayPartition()
 {
     ForEachItemIn(idx, partition)
+    {
         partition.item(idx).display();
+
+#ifdef _DEBUG
+        if ((partition.item(idx).whichInput >= 0) && (partition.item(idx).whichInput < sources.ordinality()) )
+            LOG(MCdebugInfoDetail, unknownJob,
+                     "   Header size: %d, XML header size: %" I64F "d, XML footer size: %" I64F "d",
```
jakesmith: trivial: all are unsigned so should be %u.

Incidentally why is headerSize an unsigned but xmlHeaderLength a offset_t - should be consistent I'd have thought?

dali/ft/filecopy.cpp
```diff
@@ -3000,26 +3014,22 @@ void FileSprayer::spray()
 bool FileSprayer::isSameSizeHeaderFooter()
```
jakesmith: what is the role of isSameSizeHeaderFooter() ?
I see some code that sets FPheaderLength and FPfooterLength as meta properties of each part:
```
                if (!sameSizeHeaderFooter)
                {
                    FilePartInfo & curHeaderSource = sources.item(whichHeaderInput);
                    curProps.setPropInt64(FPheaderLength, curHeaderSource.xmlHeaderLength);

                    FilePartInfo & curFooterSource = sources.item(cur.whichInput);
                    curProps.setPropInt64(FPfooterLength, curFooterSource.xmlFooterLength);

                    if ( idx+1 != partition.ordinality() )
                        whichHeaderInput = partition.item(idx+1).whichInput;
                }
```

so if not same as 1st part, then sets into part,
but I don't think I see where these per part properties are used?

dali/ft/filecopy.cpp
```diff
@@ -3000,26 +3014,22 @@ void FileSprayer::spray()
...
     if (sources.ordinality() == 0)
         return retVal;
 
+    unsigned whichHeaderInput = 0;
+
+    headerSize = sources.item(whichHeaderInput).xmlHeaderLength;
+    footerSize = sources.item(whichHeaderInput).xmlFooterLength;
+
+
+
```
jakesmith: trivial: a few too many blank lines

dali/ft/filecopy.cpp
```diff
@@ -3000,26 +3014,22 @@ void FileSprayer::spray()
...
         PartitionPoint & cur = partition.item(idx);
         if (cur.inputLength && (idx+1 == partition.ordinality() || partition.item(idx+1).whichOutput != cur.whichOutput))
         {
-            if (isEmpty)
-            {
-                headerSize = sources.item(whichHeaderInput).xmlHeaderLength;
-                footerSize = sources.item(cur.whichInput).xmlFooterLength;
-                isEmpty = false;
-                continue;
-            }
```
jakesmith: trivial: unnecessary blank line.

jakesmith:
@AttilaVamos - a few trivial comments + 1 question.

AttilaVamos:
@jakesmith updated.

dali/ft/filecopy.cpp
jakesmith: @AttilaVamos - was there an answer to this question?

dali/ft/filecopy.cpp
jakesmith: >trivial: all are unsigned so should be %u.

@AttilaVamos - looks like this one hasn't been addressed.

jakesmith:
@AttilaVamos - there 1 trivial issue and a question that wasn't addressed/answered from before.

dali/ft/filecopy.cpp
AttilaVamos: It seems my previous answer went missing. 
This implementation was part of the https://track.hpccsystems.com/browse/HPCC-13331 more than 4 years ago. I think only the despray uses this information.

dali/ft/filecopy.cpp
AttilaVamos: It has been changed in my last commit/push.

AttilaVamos:
@jakesmith The trivial issue already addressed in my last commit/push and the question answered.

jakesmith:
>trivial: all are unsigned so should be %u.

I still see the %d in the PR/branch, see: https://github.com/hpcc-systems/HPCC-Platform/blob/bf9af6ee4ca35b98437ad8d6b8d12e637586dc44/dali/ft/filecopy.cpp

@AttilaVamos

dali/ft/filecopy.cpp
jakesmith: Ok thanks.

AttilaVamos:
@jakesmith updated, please recheck.

jakesmith:
Looks good.

PR #13141, opened 2019-11-11
HPCC-23105 Fix errors in Thor generated by -Wall option in gcc 8.3

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jsocket.cpp
+++ b/system/jlib/jsocket.cpp
@@ -7001,3 +7001,5 @@ IWhiteListHandler *createWhiteListHandler(WhiteListPopulateFunction populateFunc
     return new CWhiteListHandler(populateFunc, roleFormatFunc);
 }
 
+static_assert(sizeof(IpAddress) == 16, "check size of IpAddress");
+static_assert(sizeof(SocketEndpoint) == 20, "check size of SocketEndpoint");
--- a/system/jlib/jsocket.hpp
+++ b/system/jlib/jsocket.hpp
@@ -76,9 +76,9 @@ enum JSOCKET_ERROR_CODES {
 //
 class jlib_decl IpAddress
 {
-    unsigned netaddr[4];
+    unsigned netaddr[4] = { 0, 0, 0, 0 };
 public:
-    IpAddress()                                         { ipset(NULL); }
+    IpAddress() = default;
     IpAddress(const IpAddress& other)                   { ipset(other); }
     explicit IpAddress(const char *text)                { ipset(text); }
     
@@ -140,11 +140,11 @@ extern jlib_decl IpAddress &localHostToNIC(IpAddress &ip);
 class jlib_decl SocketEndpoint : extends IpAddress
 {
 public:
-    SocketEndpoint()                                            { set(NULL,0); };
+    SocketEndpoint() = default;
     SocketEndpoint(const char *name,unsigned short _port=0)     { set(name,_port); };
     SocketEndpoint(unsigned short _port)                        { setLocalHost(_port); };
     SocketEndpoint(unsigned short _port, const IpAddress & _ip) { set(_port,_ip); };          
-    SocketEndpoint(const SocketEndpoint &other) : IpAddress(other) { port = other.port; }
+    SocketEndpoint(const SocketEndpoint &other) = default;
 
     void deserialize(MemoryBuffer & in);
     void serialize(MemoryBuffer & out) const;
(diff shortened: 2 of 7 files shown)


ghalliday:
@jakesmith please review carefully.....

system/jlib/jsocket.hpp
```diff
@@ -140,11 +140,11 @@ extern jlib_decl IpAddress &localHostToNIC(IpAddress &ip);
 class jlib_decl SocketEndpoint : extends IpAddress
 {
 public:
-    SocketEndpoint()                                            { set(NULL,0); };
+    SocketEndpoint() = default;
     SocketEndpoint(const char *name,unsigned short _port=0)     { set(name,_port); };
     SocketEndpoint(unsigned short _port)                        { setLocalHost(_port); };
     SocketEndpoint(unsigned short _port, const IpAddress & _ip) { set(_port,_ip); };          
-    SocketEndpoint(const SocketEndpoint &other) : IpAddress(other) { port = other.port; }
+    SocketEndpoint(const SocketEndpoint &other) = default;
```
jakesmith: Without this there'd be a default copy constructor implicitly defined I think, so this line not strictly necessary?
But probably nice to declare anyway to show it is expected.

thorlcr/master/mawatchdog.cpp
```diff
@@ -294,7 +295,9 @@ class CMasterWatchdogUDP : public CMasterWatchdogBase
             Owned<ISocket> sock = ISocket::udp_connect(getFixedPort(masterEp.port, TPORT_watchdog), ipStr.str());
             // send empty packet, stopped set, will cease reading
             HeartBeatPacketHeader hb;
-            memset(&hb, 0, sizeof(hb));
+            //clear struct, to avoid spurious warnings when serializing raw [unpacked] struct.
+            //Cast is to avoid warning about clearing an object with non trivial copy assignment
+            memset(reinterpret_cast<void *>(&hb), 0, sizeof(hb));
```
jakesmith: would it not be better to set member defaults?

jakesmith: ah I see you have, so this memset to 0, isn't necessary any more I think?

jakesmith: >to avoid spurious warnings when serializing raw [unpacked] struct..

Valgrind... yes I think you were right, due to alignment issues, if the whole struct isn't cleared, then there will be uninitialized data, and I think that causes false positives in valgrind.

Perhaps it would be better to pad the struct, or use a pragma to pack it..

jakesmith:
@ghalliday - looks fine as is, but a couple of questions.

thorlcr/master/mawatchdog.cpp
ghalliday: yes, but I think the issue it is trying to solve is to clear the bytes between the fields e.g.between a short and the next unsigned.

system/jlib/jsocket.hpp
ghalliday: I think it probably would be, but I wanted to play it safe..

ghalliday:
@jakesmith please look at the alternative change

system/jlib/jsocket.cpp
```diff
@@ -7001,3 +7001,5 @@ IWhiteListHandler *createWhiteListHandler(WhiteListPopulateFunction populateFunc
     return new CWhiteListHandler(populateFunc, roleFormatFunc);
 }
 
+static_assert(sizeof(IpAddress) == 16, "check size of IpAddress");
+static_assert(sizeof(SocketEndpoint) == 20, "check size of SocketEndpoint");
```
jakesmith: Perhaps add comment here, saying that checking because struct is padded to ensure alignment when included elsewhere?

jakesmith:
@ghalliday - one minor comment (re. adding comment), but looks fine.

ghalliday:
@richardkchapman

system/jlib/jsocket.cpp
```diff
@@ -7001,3 +7001,5 @@ IWhiteListHandler *createWhiteListHandler(WhiteListPopulateFunction populateFunc
     return new CWhiteListHandler(populateFunc, roleFormatFunc);
 }
 
+static_assert(sizeof(IpAddress) == 16, "check size of IpAddress");
```
richardkchapman: Probably not true if we ever turn off IPV6 flag...

richardkchapman:
@ghalliday One minor comment

system/jlib/jsocket.cpp
ghalliday: I think the size is always 16bytes at the moment - even with ipv6 turned off.  If it changed we would want to know.  I will open a jira to remove the code that directly writes the memory rather than serializing it first.

ghalliday:
@richardkchapman added https://track.hpccsystems.com/browse/HPCC-23115

PR #13159, opened 2019-11-18
HPCC-22949 Ensure local time is updated intermittently

shamser:
## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [x] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/common/thorhelper/thorcommon.cpp
+++ b/common/thorhelper/thorcommon.cpp
@@ -42,6 +42,8 @@
 
 #define ROWAGG_PERROWOVERHEAD (sizeof(AggregateRowBuilder))
 
+defaultFlushTimerInterval = nanosec_to_cycle(1000000000); // Cycles
+
 void AggregateRowBuilder::Link() const { LinkRoxieRow(this); }
 bool AggregateRowBuilder::Release() const { ReleaseRoxieRow(this); return false; }  // MORE - return value is iffy
 
--- a/common/thorhelper/thorcommon.cpp
+++ b/common/thorhelper/thorcommon.cpp
@@ -42,7 +42,7 @@
 
 #define ROWAGG_PERROWOVERHEAD (sizeof(AggregateRowBuilder))
 
-defaultFlushTimerInterval = nanosec_to_cycle(1000000000); // Cycles
+const cycle_t defaultFlushTimerInterval = nanosec_to_cycle(1000000000); // Cycles
 
 void AggregateRowBuilder::Link() const { LinkRoxieRow(this); }
 bool AggregateRowBuilder::Release() const { ReleaseRoxieRow(this); return false; }  // MORE - return value is iffy
--- a/common/thorhelper/thorcommon.hpp
+++ b/common/thorhelper/thorcommon.hpp
@@ -32,7 +32,7 @@
 static unsigned const defaultDaliResultOutputMax = 2000; // MB
 static unsigned const defaultDaliResultLimit = 10; // MB
 static unsigned const defaultMaxCsvRowSize = 10; // MB
-extern jlib_decl const cycle_t defaultFlushTimerInterval;
+extern THORHELPER_API const cycle_t defaultFlushTimerInterval;
 
 
 #define OPT_OUTPUTLIMIT_LEGACY    "outputLimit"             // OUTPUT Mb limit (legacy property name, renamed to outputLimitMb in 5.2)
--- a/common/thorhelper/thorcommon.hpp
+++ b/common/thorhelper/thorcommon.hpp
@@ -32,6 +32,7 @@
 static unsigned const defaultDaliResultOutputMax = 2000; // MB
 static unsigned const defaultDaliResultLimit = 10; // MB
 static unsigned const defaultMaxCsvRowSize = 10; // MB
+static cycle_t defaultFlushTimerInterval = nanosec_to_cycle(1000000000); // Cycles
 
 
 #define OPT_OUTPUTLIMIT_LEGACY    "outputLimit"             // OUTPUT Mb limit (legacy property name, renamed to outputLimitMb in 5.2)
@@ -256,7 +257,6 @@ public:
 };
 
 #ifdef TIME_ACTIVITIES
-
 class ActivityTimer
 {
     unsigned __int64 startCycles;
@@ -281,8 +281,7 @@ public:
         else
             startCycles = 0;
     }
-
-    ~ActivityTimer()
+    void flushTimes()
     {
         if (enabled)
         {
@@ -290,9 +289,19 @@ public:
             accumulator.endCycles = nowCycles;
             cycle_t elapsedCycles = nowCycles - startCycles;
             accumulator.totalCycles += elapsedCycles;
+            startCycles = nowCycles;
             if (isFirstRow)
                 accumulator.firstExitCycles = nowCycles;
         }
+    }
+    inline void flushTimesIntermittantly()
+    {
+        if (get_cycles_now() - (cycle_t) startCycles > defaultFlushTimerInterval)
+            flushTimes();
+    }
+    ~ActivityTimer()
+    {
+        flushTimes();
     }
 };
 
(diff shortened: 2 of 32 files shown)


shamser:
@ghalliday @jakesmith Please can you review.

ghalliday (commented):
@shamser the other general comment is when is this an issue?  I think it is only when an activity is likely to do a large amount of local processing, or blocking, before it returns.  I suspect that is a smaller set of activities than the calls below.
Also this solution doesn't solve the problem where the activity is currently blocked.

common/thorhelper/thorcommon.cpp
```diff
@@ -42,6 +42,8 @@
 
 #define ROWAGG_PERROWOVERHEAD (sizeof(AggregateRowBuilder))
 
+defaultFlushTimerInterval = nanosec_to_cycle(1000000000); // Cycles
```
ghalliday: Needs to be:

    const cycle_t defaultFlushTimerInterval = nanosec_to_cycle(1000000000); // Cycles

common/thorhelper/thorcommon.hpp
```diff
@@ -32,6 +32,7 @@
 static unsigned const defaultDaliResultOutputMax = 2000; // MB
 static unsigned const defaultDaliResultLimit = 10; // MB
 static unsigned const defaultMaxCsvRowSize = 10; // MB
+extern jlib_decl const cycle_t defaultFlushTimerInterval;
```
ghalliday: I told you the wrong qualifier.  It needs to be:

`extern THORHELPER_API const cycle_t defaultFlushTimerInterval;
`

ghalliday:
@shamser with the two changes above it compiles on windows.

thorlcr/activities/keyedjoin/thkeyedjoinslave.cpp
```diff
@@ -2554,6 +2554,7 @@ class CKeyedJoinSlave : public CSlaveActivity, implements IJoinProcessor
                         waitingForDoneGroups = true;
                     }
                     waitingForDoneGroupsSem.wait();
+                    t.flushTimesIntermittantly();  // Need this even though time is spent blocked
```
jakesmith: not sure I understand this comment..

As you say, this is blocked time really, so I think should be updating that here if anything.

common/thorhelper/thorcommon.cpp
```diff
@@ -42,6 +42,8 @@
 
 #define ROWAGG_PERROWOVERHEAD (sizeof(AggregateRowBuilder))
 
+const cycle_t defaultFlushTimerInterval = nanosec_to_cycle(1000000000); // Cycles
```
jakesmith: for ease of reading should comment that 1 second.

But doing this every second seem overkill.. The stats aren't collected that often..

Not sure what sensible value is, is 1 minute too long?

jakesmith: Btw, if useful there's a oneSecInCycles constant defined in jdebug.hpp

jakesmith: >accessing defaultFlushTimerInterval is potentially problematic - it will be an indirect reference because of the inter-dll symbol resolution, and it needs to be exported otherwise the code will not link on windows (or with hidden symbols on linux).

@ghalliday - perhaps defaultFlushTimerInterval value should be copied into ActivityTimeAccumulator as a member, for speedier access?

common/thorhelper/thorcommon.hpp
```diff
@@ -281,19 +281,28 @@ class ActivityTimer
...
             cycle_t nowCycles = get_cycles_now();
             accumulator.endCycles = nowCycles;
             cycle_t elapsedCycles = nowCycles - startCycles;
             accumulator.totalCycles += elapsedCycles;
+            startCycles = nowCycles;
             if (isFirstRow)
                 accumulator.firstExitCycles = nowCycles;
         }
     }
+    inline void flushTimesIntermittantly()
```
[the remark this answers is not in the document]
jakesmith: How much does this slow a simple query down, which hits the flushTimesIntermittantly() per record on a sequence of activities?

I wonder whether a countdown counter wouldn't be better / more efficient.
e.g. approximate a starting number, update cycles when hits zero.
Could keep adjusting the starting number if elapsed time is much shorter or longer than desired update time.

common/thorhelper/thorcommon.hpp
```diff
@@ -281,15 +281,29 @@ class ActivityTimer
...
+    }
+    inline void flushTimesIntermittantly()
     {
         if (enabled)
         {
             cycle_t nowCycles = get_cycles_now();
-            accumulator.endCycles = nowCycles;
-            cycle_t elapsedCycles = nowCycles - startCycles;
-            accumulator.totalCycles += elapsedCycles;
+            if (nowCycles - (cycle_t) startCycles > defaultFlushTimerInterval)
```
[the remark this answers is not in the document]
shamser: startCycles is unsigned. cycle_t types is signed and in most places we are using cycle_t.  We can make nowCycles unsigned but it'd be better to make startCycles cycle_t type. Shall I change startCycles to cycle_t type so that it is consistant and casting is not necessary?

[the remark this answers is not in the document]
jakesmith: ok, we didn't cast before, but I thought startCycles was a cycle_t, I see it is not now in this context, but is in other classes in this header (e.g. the startTimer member of ActivityTimeAccumulator).
I would make startCycles a cycle_t for consistency.

common/thorhelper/thorcommon.hpp
```diff
@@ -281,15 +281,29 @@ class ActivityTimer
...
+            {
+                flushTimes(nowCycles);
+                startCycles = nowCycles;
```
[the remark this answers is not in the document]
shamser: To improve the code, we could store the subtraction to be used in both the comparison and assignment later. We could eliminate flushTimes and put that code in flushTimesIntermittantly, if we were not sure that compiler would handle the inline specifier as efficiently.

As far as the suggestion to eliminate the comparison altogether, it will mean with every call to flushTimesIntermittantly, we would have:
1) extra work in the way of 2 assignments (accumulator.endCycles and startCycles) and 1 addition (calculate new value for totalCycles)
3) but save a little time by not needing to do the comparison.

It would be simpler but I would think it will be a little slower especially.

thorlcr/activities/keyedjoin/thkeyedjoinslave.cpp
shamser: The local execute time is calculated from both the blocked time and the total cycles.  If we don't update the total cycles when the activity is blocked, then the execute time will not be calculated correctly.  In fact, we could have the situation of the totalCycles < blockedCycles, which doesn't make sense. (I assume the stats are updated intermittantly even if the activity is blocked or executing slowly?)

common/thorhelper/thorcommon.cpp
shamser: How frequently are the stats collected?  Could we use the same time period to update the stats? (Thanks for the info regardiong oneSecInCycles)

shamser (commented):
@jakesmith @ghalliday

common/thorhelper/thorcommon.hpp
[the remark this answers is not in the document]
jakesmith: probably slower to remove the condition/always assign vs calc. diff once + condition.
Might be worth timing running each in a tight loop.

But a simple counter as in other comment might be best, because you really want to try to do as close to nothing if possible in vast majority of calls given how many billions of times this will be called.

@ghalliday - what do you think of a self-adjusting countdown counter ?

ghalliday:
I'm not sure about this approach, particularly the cost of calling get_cycles_now() that frequently.

The benefit is that when the graph is running the timings will be up to date  - but it does not affect the timings once everything is completed, which is when most of the timings will be looked at.

As Jake says, timings are only collected every 30s, so updating approx every 10s would be sufficient.

This only needs to be done for activities that are likely to spend a long time before returning a row (rather than potentially).  I suspect that means we should only use it for aggregates/rollups, not for filters.

In those cases unconditionally updating the times every (10K?) input rows would have lower overhead, and likely to update frequently enough.

Note, this approach does not extend to blocked times, or stranded activities (since updating the timings is not thread safe), so it has limited application.

ghalliday:
@shamser I think the conclusion is this probably isn't the best way to fix it.  See the jira for more comments and discussion.

PR #13165, opened 2019-11-19
HPCC-23140 Refactor/speed up verifyAll and add per connect timeout

jakesmith:
And pass shorter per connection timeout, such that Thor fails
earlier if slaves have not come up in a timely way on startup.

Used to have to wait for watchdog and consequently wait approx.
30 minutes, in some circumstances, if slaves on channels failed
to start up.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/mp/mpcomm.cpp
+++ b/system/mp/mpcomm.cpp
@@ -24,6 +24,9 @@
     look at all timeouts
 */
 
+#include <future>
+#include <vector>
+
 #include "platform.h"
 #include "portlist.h"
 #include "jlib.hpp"
@@ -82,7 +85,6 @@
 
 #define _TRACING
 
-static  CriticalSection verifysect;
 static  CriticalSection childprocesssect;
 static  UnsignedArray childprocesslist;
 
@@ -2604,6 +2606,7 @@ void CMPServer::notifyClosed(SocketEndpoint &ep, bool trace)
 class CInterCommunicator: public IInterCommunicator, public CInterface
 {
     CMPServer *parent;
+    CriticalSection verifysect;
 
 public:
     IMPLEMENT_IINTERFACE;
@@ -2817,6 +2820,7 @@ class CCommunicator: public ICommunicator, public CInterface
     CMPServer *parent;
     bool outer;
     rank_t myrank;
+    CriticalSection verifysect;
 
     const SocketEndpoint &queryEndpoint(rank_t rank)
     {
@@ -2945,46 +2949,98 @@ public:
         return channel->verifyConnection(tm,true);
     }
 
-    bool verifyAll(bool duplex, unsigned timeout)
+    bool verifyAll(bool duplex, unsigned totalTimeout, unsigned perConnectionTimeout)
     {
         CriticalBlock block(verifysect);
-        CTimeMon tm(timeout);
+        CTimeMon totalTM(totalTimeout);
+
+        Semaphore sem;
+        sem.signal(getAffinityCpus());
+        std::atomic<bool> abort{false};
+        
+        auto verifyConnWithConnect = [&](unsigned rank, unsigned timeout)
         {
-            ForEachNodeInGroup(rank,*group) {
-                bool doverify;
-                if (duplex)
-                    doverify = (rank!=myrank);
-                else if ((rank&1)==(myrank&1))
-                    doverify = (myrank>rank);
-                else
-                    doverify = (myrank<rank);
-                if (doverify) {
-                    Owned<CMPChannel> channel = getChannel(rank);
+            CTimeMon tm(timeout);
+            Owned<CMPChannel> channel = getChannel(rank);
+            return channel->verifyConnection(tm, true);    
+        };
+
+        auto verifyConnWithoutConnect = [&](unsigned rank, unsigned timeout)
+        {
+            CTimeMon tm(timeout);
+            while (true)
+            {
+                Owned<CMPChannel> channel = getChannel(rank);
+                if (channel->verifyConnection(tm, false))
+                    return true;
+                if (abort || tm.timedout())
+                    return false;
+                Sleep(100);
+            }
+        };
+
+        auto threadedVerifyConnectFunc = [&](rank_t rank, std::function<bool (unsigned rank, unsigned timeout)> connectFunc)
+        {
+            // NB: running because took (via wait()) a semaphore slot, restore it at end of scope
+            struct RestoreSlot
+            {
+                Semaphore &sem;
+                RestoreSlot(Semaphore &_sem) : sem(_sem) { }
+                ~RestoreSlot() { sem.signal(); }
+            } restoreSlot(sem);
+
+            unsigned timeoutMs;
+            if (totalTM.timedout(&timeoutMs) || abort)
+                return false;
+            if (perConnectionTimeout && (perConnectionTimeout < timeoutMs))
+                timeoutMs = perConnectionTimeout;
+
+            if (!connectFunc(rank, timeoutMs))
+            {
+                abort = true; // ensure verifyFunc knows before release slot, to prevent other thread being launched
+                return false;
+            }
+            return true;
+        };
+
+        auto verifyFunc = [&](std::function<bool (unsigned rank)> isRankToVerifyFunc, std::function<bool (unsigned rank, unsigned timeout)> connectFunc)
+        {
+            std::vector<std::future<bool>> results;
+            for (rank_t rank=0; rank<group->ordinality(); rank++)
+            {
+                if (isRankToVerifyFunc(rank))
+                {
+                    // check timeout before and after sem.wait
+                    // NB: sem.wait if successful, takes a slot which is restored by the thread when it is done
                     unsigned remaining;
-                    if (tm.timedout(&remaining)) {
-                        return false;
+                    if (totalTM.timedout(&remaining) || !sem.wait(remaining) || totalTM.timedout(&remaining))
+                    {
+                        abort = true;
+                        break;
                     }
-                    if (!channel->verifyConnection(tm,true))
-                        return false;
+                    else if (abort)
+                        break;
+                    results.push_back(std::async(std::launch::async, threadedVerifyConnectFunc, rank, connectFunc));
                 }
             }
-        }
-        if (!duplex) {
-            ForEachNodeInGroup(rank,*group) {
-                bool doverify = ((rank&1)==(myrank&1))?(myrank<rank):(myrank>rank);
-                if (doverify) {
-                    Owned<CMPChannel> channel = getChannel(rank);
-                    while (!channel->verifyConnection(tm,false)) {
-                        unsigned remaining;
-                        if (tm.timedout(&remaining))
-                            return false;
-                        CriticalUnblock unblock(verifysect);
-                        Sleep(100);
-                    }
-                }
+            bool res = true;
+            for (auto &f: results)
+            {
+                if (!f.get())
+                    res = false;
             }
+            return res && !abort;
+        };
+
+        if (duplex)
+            return verifyFunc([this](rank_t rank) { return rank != myrank; }, verifyConnWithConnect);
+        else
+        {
+            if (!verifyFunc([this](rank_t rank) { return ((rank&1)==(myrank&1)) ? (myrank > rank) : (myrank < rank); }, verifyConnWithConnect))
+                return false;
+
+            return verifyFunc([this](rank_t rank) { return ((rank&1)==(myrank&1)) ? (myrank < rank) : (myrank > rank); }, verifyConnWithoutConnect);
         }
-        return true;
     }
 
     unsigned probe(rank_t srcrank, mptag_t tag, rank_t *sender, unsigned timeout=0)
--- a/dali/base/dacoven.cpp
+++ b/dali/base/dacoven.cpp
@@ -291,10 +291,10 @@ public:
     }
 
 
-    virtual bool verifyAll(bool duplex=false, unsigned timeout=1000*60*30)
+    virtual bool verifyAll(bool duplex=false, unsigned timeout=1000*60*30, unsigned perConnectionTimeout=0)
     {
         assertex(comm);
-        return comm->verifyAll(duplex,timeout);
+        return comm->verifyAll(duplex, timeout, perConnectionTimeout);
     }
     
     // receive, returns senders rank or false if no message available in time given or cancel called
--- a/system/mp/mpcomm.hpp
+++ b/system/mp/mpcomm.hpp
@@ -54,7 +54,7 @@ interface ICommunicator: extends IInterface
     virtual void flush  (mptag_t tag) = 0;    // flushes pending buffers
 
     virtual bool verifyConnection(rank_t rank, unsigned timeout=1000*60*5) = 0; // verifies connected to rank
-    virtual bool verifyAll(bool duplex=false, unsigned timeout=1000*60*30) = 0;
+    virtual bool verifyAll(bool duplex=false, unsigned timeout=1000*60*30, unsigned perConnectionTimeout=0) = 0;
     virtual void disconnect(INode *node) = 0;
     virtual void barrier() = 0;
     virtual const SocketEndpoint &queryChannelPeerEndpoint(const SocketEndpoint &sender) const = 0;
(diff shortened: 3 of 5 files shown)


system/mp/mpcomm.cpp
```diff
@@ -2817,6 +2820,7 @@ class CCommunicator: public ICommunicator, public CInterface
     CMPServer *parent;
     bool outer;
     rank_t myrank;
+    CriticalSection verifysect;
```
jakesmith: NB: moved from global scope to instance scope.
It protects verifyAll/verifyConnection/disconnect from one another, I can't see why it was at global scope, and my test involved multiple overlapping verifyAll's in fact, I think it was only luck that worked before because of the way that verifyAll backs off for 100ms in a loop.

But please look/consider any possible issues here.

jakesmith:
@mckellyln - please review

richardkchapman:
Typo (conncetion) in commit message

jakesmith:
Fixed typo in commit message.

@mckellyln - please review.

system/mp/mpcomm.cpp
```diff
@@ -2945,46 +2949,102 @@ class CCommunicator: public ICommunicator, public CInterface
...
-                    doverify = (rank!=myrank);
-                else if ((rank&1)==(myrank&1))
-                    doverify = (myrank>rank);
-                else
-                    doverify = (myrank<rank);
-                if (doverify) {
+            CTimeMon tm(timeout);
+            CriticalBlock block(verifysect);
+            Owned<CMPChannel> channel = getChannel(rank);
+            return channel->verifyConnection(tm, true);
```
mckellyln: This routine has a CS in it which may serialize execution.

jakesmith: yes, in previous draft of this PR, I had a CS per rank,
but in fact originally the CS was held at the start of verifyAll, with some nasty logic to periodically yield if it failed. That was because the CS was global and multiple communicators (e.g. belonging to different channels) could clash and deadlock each other.
The CS is now inside the communicator and it is correct that it is held for the duration of verifyAll, to exclude other calls to verifyAll or disconnect.
(and that means that the threads no longer block each other)

jakesmith:
@mckellyln resolved issue with badly placed CS.

system/mp/mpcomm.cpp
mckellyln: I still do not see new threads or any parallelism

jakesmith: there's no CS now. So not sure what you mean, can you elaborate?

mckellyln: Doesn't the f.get() block ?

mckellyln: Should the f.get() be a short wait_for and future_status check is ready ?

jakesmith: Shouldn't be necessary, by the time it's in the loop calling f.get() the last N will be running (they will have gained a slot), and will be in the threadedVerifyConnectFunc. They must timeout after approx the overall timeout amount or per connection timeout if shorter.

We need to wait for them to finish either way before leaving verifyAll, so I think f.get() is good.

system/mp/mpcomm.cpp
```diff
@@ -2945,46 +2949,99 @@ class CCommunicator: public ICommunicator, public CInterface
...
+                    if (totalTM.timedout(&remaining) || !sem.wait(remaining) || totalTM.timedout(&remaining))
+                    {
+                        abort = true;
+                        break;
                     }
-                    if (!channel->verifyConnection(tm,true))
-                        return false;
+                    else if (abort)
+                        break;
+                    threads.push_back(std::async(threadedVerifyConnectFunc, rank, connectFunc));
```
mckellyln: Should this be: ?
threads.push_back(std::async(std::launch::async,threadedVerifyConnectFunc, rank, connectFunc));
Without the std::launch::async I don't think new threads are always created ??

system/mp/mpcomm.cpp
mckellyln: It may be related to std::launch::async ?  See other comment about this.

mckellyln: Unrelated, but also if we could eliminate the connectsect CS held for connect() it might offer a potential speedup.

system/mp/mpcomm.cpp
jakesmith: not sure, the default behaviour is std::launch::async | std::launch::deferred, which does mean it's implementation dependent, so might limit the number of threads, e.g. if resources are tight.
That may not be a bad thing though.
But I'm not seeing that, I see up to 32 concurrent threads (based on my hardware_concurrency() limit).

Are you not seeing threads created?

I doubt matters much either way here, but given there's a semaphore limiting max here too, it could be changed to std::launch::async.

But I'm curious if you're not seeing thread and why.

mckellyln: Correct, no new threads, even with delays added, unless I add std::launch::async.

mckellyln: Would it be a compiler version issue ?

mckellyln: Or that there are already several threads running and my limit is lower (8) ?  I will try with a forced larger value to see.

mckellyln: I tried a larger value (32) and it did not make any difference.

jakesmith: I'm not sure why we're seeing very different behviour, I see the max number of threads running in parallel, you see 1.
But as the semantics mean it is guaranteed, and this impl. is self limiting anyway, I will change to std::launch::async

jakesmith:
@mckellyln - please check after change to std::launch::async

system/mp/mpcomm.cpp
```diff
@@ -2945,46 +2949,99 @@ class CCommunicator: public ICommunicator, public CInterface
...
+                return false;
+            }
+            return true;
+        };
+
+        auto verifyFunc = [&](std::function<bool (unsigned rank)> conditionFunc, std::function<bool (unsigned rank, unsigned timeout)> connectFunc)
+        {
+            std::vector<std::future<bool>> threads;
+            for (rank_t rank=0; rank<group->ordinality(); rank++)
+            {
```
mckellyln: I don't think this works if the group->ordinality() is larger than std::thread::hardware_concurrency() ?

system/mp/mpcomm.cpp
```diff
@@ -2945,46 +2949,99 @@ class CCommunicator: public ICommunicator, public CInterface
...
-                        if (tm.timedout(&remaining))
-                            return false;
-                        CriticalUnblock unblock(verifysect);
-                        Sleep(100);
-                    }
-                }
+            bool res = true;
+            for (auto &f: threads)
+            {
+                if (!f.get())
```
mckellyln: It sort of seems that the above loop over ranks could be a while and interleave in that loop the check of each thread's completion status and if complete then a get() ?

mckellyln: I think its ok as is, but since get() blocks it does seem like we are waiting on them in order.

jakesmith: get() on future is effectively a join() and then getting the result from the thread.

The threads will run asynchronously, yes you could be blocked on 1st get() (joining), whilst 2->n had already finished.

But you have to wait on all anyway (have to join all), so there's no value in timing out and joining others instead.
If were waiting blocked (joining) on 1 and others had finished, it would join all others as soon as became unblocked.

I don't think there's a better way of coding it or any gain.
But I could be missing something in what you mean.
@mckellyln  - do you have some pseudo code of what you mean?

mckellyln: See my comment above this one.  I think there may be a problem if group->ordinality() > std::thread::hardware_concurrency() ?  And then to change it to a while loop to handle that __suggests__ to me that one might interleave the checking for complete in some threads while still waiting for others.  But yes I agree and stated above, its ok to get() afterwards - as long as we started all we need to.

system/mp/mpcomm.cpp
jakesmith: @mckellyln - why? (and I have tested with large groups)

system/mp/mpcomm.cpp
jakesmith: >one might interleave the checking for complete in some threads while still waiting for others

still not sure what point would be. But an example may help me understand benefits.

richardkchapman:
Seems to include an unwanted submodule change

jakesmith:
Removed accidental update to submodule.

system/mp/mpcomm.cpp
mckellyln: ok, I agree I thin it is ok.

system/mp/mpcomm.cpp
mckellyln: I'll agree it is ok as coded.

mckellyln:
@jakesmith, one last question - do you think as a part of this the connectsect CS held for connect() in verifyConnection() could be looked at ?

jakesmith:
>do you think as a part of this the connectsect CS held for connect() in verifyConnection() could be looked at ?

Possible, but not that related. The connectsect prevents contention dealing with the same CMPChannel.
To avoid that CS, e.g. when multiple threads send to avoid packet clashing etc., is a different problem. And I suspect the CS is reasonable there.

The verifyAll CS (verifysect) is preventing multiple verifyAll's or disconnects clashing on the same communicator (prior to this change, it was preventing any running concurrently, even if different communicators).

I'm not sure looking at the connect() CS usage is worth it, but if you have some further thoughts on it, please open a separate JIRA.

@mckellyln

mckellyln:
@jakesmith I am good with this PR.  ok to merge after squash.

mckellyln (approved):
good to merge after squash

jakesmith:

richardkchapman:

system/mp/mpcomm.cpp
```diff
@@ -2945,46 +2949,99 @@ class CCommunicator: public ICommunicator, public CInterface
...
+                    if (totalTM.timedout(&remaining) || !sem.wait(remaining) || totalTM.timedout(&remaining))
+                    {
+                        abort = true;
+                        break;
                     }
-                    if (!channel->verifyConnection(tm,true))
-                        return false;
+                    else if (abort)
+                        break;
+                    threads.push_back(std::async(std::launch::async, threadedVerifyConnectFunc, rank, connectFunc));
```
ghalliday: Out of interest, why did you choose to use the standard library functionality, rather than the CAsyncFor?

system/mp/mpcomm.cpp
```diff
@@ -2945,46 +2949,99 @@ class CCommunicator: public ICommunicator, public CInterface
...
+            {
+                abort = true; // ensure verifyFunc knows before release slot, to prevent other thread being launched
+                return false;
+            }
+            return true;
+        };
+
+        auto verifyFunc = [&](std::function<bool (unsigned rank)> conditionFunc, std::function<bool (unsigned rank, unsigned timeout)> connectFunc)
+        {
+            std::vector<std::future<bool>> threads;
```
ghalliday: Probably not significant in this context, but this will be a vector of fairly large objects.   They will contain synchronization objects.  A quick look at the headers shows them using shared_ptr to shared the state.  They take up about 48 bytes each on my machine.

ghalliday (commented):
@jakesmith various questions and comments.

system/mp/mpcomm.cpp
```diff
@@ -2945,46 +2949,99 @@ class CCommunicator: public ICommunicator, public CInterface
...
+
+            if (!connectFunc(rank, timeoutMs))
+            {
+                abort = true; // ensure verifyFunc knows before release slot, to prevent other thread being launched
+                return false;
+            }
+            return true;
+        };
+
+        auto verifyFunc = [&](std::function<bool (unsigned rank)> conditionFunc, std::function<bool (unsigned rank, unsigned timeout)> connectFunc)
```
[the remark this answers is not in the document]
jakesmith: I'll change it to "isRankToVerifyFunc"

system/mp/mpcomm.cpp
```diff
@@ -2945,46 +2949,99 @@ class CCommunicator: public ICommunicator, public CInterface
...
     }
 
-    bool verifyAll(bool duplex, unsigned timeout)
+    bool verifyAll(bool duplex, unsigned totalTimeout, unsigned perConnectionTimeout)
     {
         CriticalBlock block(verifysect);
-        CTimeMon tm(timeout);
+        CTimeMon totalTM(totalTimeout);
+
+        std::vector<std::future<bool>> threads;
```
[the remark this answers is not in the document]
jakesmith: yes, I'll remove

system/mp/mpcomm.cpp
jakesmith: CAsyncFor didn't quite fit.

I want to iterate through ranks, adding threads if room in the pool, block if not.
I could do it with a CAsyncFor, but it would immediately launch N threads for each rank, then return immediately for the ones that do nothing, or similar.
Lambda's and closures also helper to capture the variables in use here, which would have not been so clean if using a a derived class for CAsyncFor.

std::future is also a nice paradigm that works well here.

system/mp/mpcomm.cpp
```diff
@@ -2945,46 +2949,99 @@ class CCommunicator: public ICommunicator, public CInterface
...
+        Semaphore sem;
+        sem.signal(std::thread::hardware_concurrency());
+        bool abort = false;
```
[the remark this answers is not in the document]
jakesmith: ok, I thought because byte it was atomic,
but I didn't realize because it's a local var, that compiler may assume it isn't changed.
Now you point out, I wonder if because it's captured and used in a lambda - the compiler is wise enough to know it might be used/changed by another thread.

Also didn't realize that making it an atomic had no overhead if only reading and writing.
I agree it is clearer as an atomic.

Will change it.

system/mp/mpcomm.cpp
```diff
@@ -2945,46 +2949,99 @@ class CCommunicator: public ICommunicator, public CInterface
...
-    bool verifyAll(bool duplex, unsigned timeout)
+    bool verifyAll(bool duplex, unsigned totalTimeout, unsigned perConnectionTimeout)
     {
         CriticalBlock block(verifysect);
-        CTimeMon tm(timeout);
+        CTimeMon totalTM(totalTimeout);
+
+        std::vector<std::future<bool>> threads;
+        Semaphore sem;
+        sem.signal(std::thread::hardware_concurrency());
```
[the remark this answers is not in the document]
jakesmith: will change to getAffinityCpus()

system/mp/mpcomm.cpp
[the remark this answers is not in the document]
jakesmith: I may be missing what you're getting at..

If the rank loop that's launching threads (creating std::future<std::async>>'s), stops early,
then this loop will have less future's to iterate through, but all that were created and add to 'threads', will be running at this stage. So these future::get()'s will block until they finish.

jakesmith:
@ghalliday - please see replies and changes.

system/mp/mpcomm.cpp
ghalliday: Should really be called results rather than threads.

ghalliday (commented):
@jakesmith @richardkchapman I think this is fine and ready to merge.  (Up to Jake if he renames the threads variable.)

I have some hesitation about not using CAsyncFor (and possibly extending that with a filter function) since this code is very differently structured to other code we have that is performing async operations.  std::futures are neat, but I would be hesitant to use them in contexts that are time or memory critical - I think it will mean there are more synchronization operations is bound to affect performance.

A general concern (which I have also hit when writing code that uses vectors) is that mixing jlib classes and std classes can be very confusing.  I'm not sure what the solution is, but something we should discuss properly (at the offsite?) rather than drifting into it.

jakesmith:
renamed threads to results, and squashed.

@ghalliday @richardkchapman -ready to merge

ghalliday:
@richardkchapman

richardkchapman:
@jakesmith  Feels a little large for a point release?

jakesmith:
yes I agree, I'll move to master/7.8

jakesmith:
@richardkchapman - retargeted to master.

PR #13166, opened 2019-11-19
HPCC-23084 Aborting of Thor jobs creates core files

mckellyln:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

## Testing:

Smoke test and manual regression Thor suite and manual aborting.

### DIFF
--- a/system/jlib/jexcept.cpp
+++ b/system/jlib/jexcept.cpp
@@ -1101,6 +1101,12 @@ static void throwSigSegV()
 }
 #endif
 
+static std::atomic<bool> processAborted { false };
+void jlib_decl setProcessAborted(bool _abortVal)
+{
+    processAborted = _abortVal;
+}
+
 NO_SANITIZE("alignment") void excsighandler(int signum, siginfo_t *info, void *extra)
 {
     static byte nested=0;
@@ -1354,7 +1360,10 @@ NO_SANITIZE("alignment") void excsighandler(int signum, siginfo_t *info, void *e
         if ( SEHHandler->fireException(new CSEHException(signum,s.str())) )
             return;
     }
-    raise(signum);
+    if ( (SIGABRT == signum) && (processAborted) )
+        _exit(128 + SIGABRT);
+    else
+        raise(signum);
 #endif
     nested--;
 }
--- a/thorlcr/graph/thgraph.cpp
+++ b/thorlcr/graph/thgraph.cpp
@@ -1727,6 +1727,13 @@ void CGraphBase::abort(IException *e)
 
         abortException.set(e);
         aborted = true;
+
+        if (!job.getOptBool("coreOnAbort", false))
+        {
+            // prevent dtors throwing ... __verbose_terminate_handler()->abort()->raise()->core files
+            setProcessAborted(true);
+        }
+
         graphCancelHandler.cancel(0);
 
         if (0 == containers.count())
--- a/thorlcr/msort/tsorts.cpp
+++ b/thorlcr/msort/tsorts.cpp
@@ -798,12 +798,21 @@ public:
     }
     ~CThorSorter()
     {
-        stop();
-        ActPrintLog(activity, "Joining Sort Slave Server");
-        verifyex(threaded.join(10*60*1000));
-        myendpoint.set(NULL,0);
-        rowArray.kill();
-        ActPrintLog(activity, "~CThorSorter");
+        try
+        {
+            stop();
+            ActPrintLog(activity, "Joining Sort Slave Server");
+            verifyex(threaded.join(10*60*1000));
+            myendpoint.set(NULL,0);
+            rowArray.kill();
+            ActPrintLog(activity, "~CThorSorter");
+        }
+        catch (IException *e)
+        {
+            e->Release();
+        }
+        // do we ignore other exceptions (...) here ?
+        // if so, we may have std::terminate call abort() ...
     }
 // ISortSlaveMP
     virtual bool Connect(unsigned _partno, unsigned _numnodes)
--- a/thorlcr/slave/slavmain.cpp
+++ b/thorlcr/slave/slavmain.cpp
@@ -1645,10 +1645,19 @@ public:
     }
     ~CJobListener()
     {
-        for (unsigned sc=1; sc<channelsPerSlave; sc++)
-            mpServers.item(sc).stop();
-        mpServers.kill();
-        stop();
+        try
+        {
+            for (unsigned sc=1; sc<channelsPerSlave; sc++)
+                mpServers.item(sc).stop();
+            mpServers.kill();
+            stop();
+        }
+        catch (IException *e)
+        {
+            e->Release();
+        }
+        // do we ignore other exceptions (...) here ?
+        // if so, we may have std::terminate call abort() ...
     }
     void stop()
     {
@@ -1867,6 +1876,10 @@ public:
                         }
                         jobs.removeExact(job);
                         PROGLOG("QueryDone, removed %s from jobs", key.get());
+
+                        // reset for next job
+                        setProcessAborted(false);
+
                         if (exception)
                             throw exception.getClear(); // NB: this will cause exception to be part of the reply to master
 
--- a/system/jlib/jexcept.hpp
+++ b/system/jlib/jexcept.hpp
@@ -129,6 +129,7 @@ void jlib_decl *setSEHtoExceptionHandler(IExceptionHandler *handler); // sets ha
 void jlib_decl setTerminateOnSEHInSystemDLLs(bool set=true);
 void jlib_decl setTerminateOnSEH(bool set=true);
 
+void jlib_decl setProcessAborted(bool _abortVal);
 
 __declspec(noreturn) void jlib_decl throwUnexpectedException(const char * file, unsigned line) __attribute__((noreturn));
 __declspec(noreturn) void jlib_decl throwUnexpectedException(const char * where, const char * file, unsigned line) __attribute__((noreturn));
--- a/thorlcr/master/thgraphmanager.cpp
+++ b/thorlcr/master/thgraphmanager.cpp
@@ -696,6 +696,9 @@ void CJobManager::run()
         { 
             reply(workunit, wuid, MakeStringException(0, "Unknown exception"), agentep, false); 
         }
+
+        // reset for next job
+        setProcessAborted(false);
     }
     delete dp;
     jobq.clear();


mckellyln:
@ghalliday please review; see notes in JIRA.

thorlcr/msort/tsorts.cpp
```diff
@@ -798,12 +798,21 @@ class CThorSorter : public CSimpleInterface, implements IThorSorter, implements
     }
     ~CThorSorter()
```
mckellyln: Should I add noexcept ?

thorlcr/slave/slavmain.cpp
```diff
@@ -1645,10 +1645,19 @@ class CJobListener : public CSimpleInterface
     }
     ~CJobListener()
```
mckellyln: noexcept ?

mckellyln:
Catching exceptions in these two dtors now.

mckellyln:
Would it still be possible an STL exception was thrown and our catch(IException *e) didn't catch it ?  And thus we should still not raise(SIGABRT) in the signal handler ?

thorlcr/msort/tsorts.cpp
ghalliday: I'm not sure.  What do we want to do about unexpected exceptions?
Adding noexcept would make it a bit more resilient, but probably not much.
The only exceptions which are likely to be thrown are those derived from std::exception.  std::bad_alloc is a possibility, and I wonder if we should call std::set_new_handler to change the behaviour in that situation.  (I have been debating a change in eclcc since my regression suite keeps hitting it).

thorlcr/graph/thgraph.cpp
```diff
@@ -1174,6 +1174,7 @@ void CGraphBase::init()
 {
     bool log = queryJob().queryForceLogging(queryGraphId(), (NULL == queryOwner()) || isGlobal());
     setLogging(log);
+    setJobAborted(0);
```
ghalliday: @jakesmith  Is this the correct place to clear it, or should it be done once abort processing has finished?

jakesmith: yes, I think what we want is to turn on this behaviour (prevent cores), once an aborting condition has started, ideally after we know the original error has been reported.
Then turn it off again only when we know it's been handled.

GraphBase::init() is per subgraph, so will be called [too] often during a single [outer] graph.

It's hard to know when the aborting processing has completely finished (globally).
But I think the setJobAborted(0) statement should be called, when the job is completely finished, which is after is at end of CJobBase::endJob()

thorlcr/graph/thgraph.cpp
```diff
@@ -1727,6 +1728,10 @@ void CGraphBase::abort(IException *e)
 
         abortException.set(e);
         aborted = true;
+
+        // prevent dtors throwing ... __verbose_terminate_handler()->abort()->raise()->core files
+        setJobAborted(1);
```
jakesmith: I wouldn't set in clean(), because we still want to see truly unexpected crashes and their cores, e.g. if job is finishing in an otherwise successful state and crashes as it's winding up.

If on the other hand, it's winding up after a reported aborting condition and throws an unhandled exception that leads to a core, we care much less.

So I would turn this on, only when know there's a abort type scenario in motion.
There may not be 1 place that covers all bases though, some places where it would be good to disable potential follow-on unhanded exceptions are:

+ When the slave receives a GraphAbort: message (slavmain.cpp).
+ When the slave has reported an exception to the master. There are 2 places that happens a) CJobSlaveChannel::fireException, and b) in slavmain.cpp CThreadExceptionCatcher::fireException
+ When the Thor CJobManger::stop() has been called.

system/jlib/jexcept.cpp
```diff
@@ -1101,6 +1101,12 @@ static void throwSigSegV()
 }
 #endif
 
+static std::atomic<unsigned> jobAborted { 0 };
+void jlib_decl setJobAborted(unsigned _abortVal)
```
jakesmith: minor: given this is in jlib and not necessarily in a processes that are running 'jobs'.
perhaps setProcessAborted would be better?

mckellyln: ok, I will change the name.

thorlcr/graph/thgraph.cpp
mckellyln: ok, I will move this to endJob().

thorlcr/graph/thgraph.cpp
jakesmith: I misread the PR, thought above change was in clean(), I see it's in abort() now.
Which is good catch all place for any abort handling, so I would leave as is.
It might be a little late in some cases, but I think good enough.

mckellyln: After offline discussion leaving this in abort() is ok.

thorlcr/msort/tsorts.cpp
mckellyln: I'll leave as is.

mckellyln:
@jakesmith please review.

system/jlib/jexcept.cpp
```diff
@@ -1354,7 +1360,10 @@ NO_SANITIZE("alignment") void excsighandler(int signum, siginfo_t *info, void *e
         if ( SEHHandler->fireException(new CSEHException(signum,s.str())) )
             return;
     }
-    raise(signum);
+    if ( (SIGABRT == signum) && (processAborted) )
+        _exit(255);
```
jakesmith: perhaps should be setProcessAbortedCode , and return that code here instead of 255 ?

jakesmith:
@mckellyln - should probably make this an option, could be per job (getOptBool), but default can be on.
So can easily pursue secondary crashes.

system/jlib/jexcept.cpp
mckellyln: We have to exit at this point and the init wrapper will not restart Thor for some exit values.

mckellyln:
ok, I can add an option to disable this as a way to help debug future similar cases.

system/jlib/jexcept.cpp
jakesmith: sure but this in jlib, not Thor specific.
Thor's use can still use a code appropriate to it, e.g. 1

mckellyln: Default from before would be 128 + SIGABRT (6) so thats 134.
Are you suggesting:
```
_exit(processAborted);
```

jakesmith: So before it was 134 and it caused the Thor init scripts recycled(?)
As it stands it's hard-coded as 255 , and presumably that works with the thor init scripts and it still recycles?
I haven't checked the init script, but if 255 is what seemed appropriate, then you can have the calls to setProcessAbortCode set it to 255, in Thor code,
and then yes, here in jlib do:
```
_exit(processAbortCode)
```

thorlcr/graph/thgraph.cpp
```diff
@@ -1727,6 +1727,13 @@ void CGraphBase::abort(IException *e)
 
         abortException.set(e);
         aborted = true;
+
+        if (!job.getOptBool("coreOnAbort", false))
```
mckellyln: Does this seem reasonable to get the option ?

system/jlib/jexcept.cpp
```diff
@@ -1354,7 +1360,10 @@ NO_SANITIZE("alignment") void excsighandler(int signum, siginfo_t *info, void *e
         if ( SEHHandler->fireException(new CSEHException(signum,s.str())) )
             return;
     }
-    raise(signum);
+    if ( (SIGABRT == signum) && (processAborted) )
+        _exit(128 + SIGABRT);
```
mckellyln: I made it exit exactly like before, just without core.

mckellyln:
@jakesmith one more time around when you have time.

system/jlib/jexcept.cpp
mckellyln: Yes, 134 and 255 both would cause recycle.

system/jlib/jexcept.cpp
jakesmith: not a big deal either way, as you say the behaviour will be the same.
But there's no intention to use the value (i.e. just treated as a switch), then it would be better if it and the setting function were boolean based.

thorlcr/graph/thgraph.cpp
jakesmith: looks fine. Don't actually need the optional default value of false, since false is the default anyway.

jakesmith:
@mckellyln - 2 minor comments but could be merged as is after squashing.

system/jlib/jexcept.cpp
mckellyln: Changed to bool

thorlcr/graph/thgraph.cpp
mckellyln: left as is to know the default easily.

mckellyln:
Changed unsigned to bool and squashed.

mckellyln:
ok, to merge

richardkchapman:
@ghalliday Are you ok with this now?

thorlcr/graph/thgraph.cpp
```diff
@@ -2808,6 +2815,9 @@ void CJobBase::endJob()
     }
     if (exceptions && exceptions->ordinality())
         throw exceptions.getClear();
+
+    // reset for next job
+    setProcessAborted(false);
```
ghalliday: @mckellyln @jakesmith should this be before the throw?

jakesmith: well maybe this is too early to clear it actually.
If an exception is fired there, it could still catch and clear up cleanly ready for the next job, so it should be cleared a bit later.

The last interaction with the job the slave has is at the end of the QueryDone statement.
And the last interaction in the master is at the end of the while (!stopped) loop in CJobManager::run() (before it loops around to listen for the next job).

@mckellyln - I think should move the setProcessAborted(false) to those 2 places instead.

mckellyln: I don't think so.  If this throw is not explicitly caught and env setting coreOnAbort is not set to true then we would get a traceback in the log but no core file.  If this call was moved to before the throw, the only difference I think, would be that if its not caught we would always get a core file.

mckellyln: ok, I can move it to those two places.  Are those still better than in init() ?

jakesmith: logically, yes I think so, because init() is per subgraph, so occurs many times during a job.

mckellyln: ok, if it threw and Thor didn't end then I agree it may not get reset properly for the next job.

mckellyln:
@jakesmith please review two new locations for reset.

jakesmith:
Afaict, it looks good and will work as intended.
I'll let @ghalliday have 1 final look.

mckellyln:
@ghalliday squashed

PR #13168, opened 2019-11-20
HPCC-23148 Add a flag to indicate if an index filename is constant for a query

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/ecl/hqlcpp/hqlcpputil.cpp
+++ b/ecl/hqlcpp/hqlcpputil.cpp
@@ -442,3 +442,14 @@ bool worthGeneratingRowAsSingleActivity(IHqlExpression * expr)
         expr = expr->queryChild(0);
     }
 }
+
+bool isNonConstantAndQueryInvariant(IHqlExpression * expr)
+{
+    if (expr->isConstant())
+        return false;
+    if (!expr->isIndependentOfScope())
+        return false;
+    if (isContextDependent(expr, true, false))
+        return false;
+    return true;
+}
--- a/ecl/hqlcpp/hqlckey.cpp
+++ b/ecl/hqlcpp/hqlckey.cpp
@@ -1338,6 +1338,8 @@ ABoundActivity * HqlCppTranslator::doBuildActivityKeyedJoinOrDenormalize(BuildCt
             flags.append("|FFvarfilename");
         if (hasDynamicFilename(info.queryFile()))     
             flags.append("|FFdynamicfilename");
+        if (isNonConstantAndQueryInvariant(info.queryFileFilename()))
+            flags.append("|FFinvariantfilename");
     }
 
     if (flags.length())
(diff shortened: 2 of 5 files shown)


ghalliday:
@jakesmith fyi.  Is it worth introducing this for files as well?
@shamser please review.

ghalliday:
As coded this probably isn't backward compatible - because testing the flag will only be true for new queries.  That may mean that the inverse flag needs to be added - which will result in far more differences.
See modfied childindex and buildbug1/glheureux7 as examples which change.

richardkchapman:
You could add two flags - one to say whether the other can be trusted...

If we don't add code in Roxie that checks the flag the compatibilty is less significant.

shamser (approved):

ghalliday:
@ghalliday I will need to investigate why hthor crashes.  I suspect it is a bug that has been there all along that is now being tested.

jakesmith:
@ghalliday - have incorporated the childindex.ecl change into my PR (https://github.com/hpcc-systems/HPCC-Platform/pull/13175)

jakesmith:
Missed your earlier question above:
>fyi. Is it worth introducing this for files as well?

Yes think so, whenever a computed filename is involved in a child query, it requires the slave to reach out to the master to resolve the filename, and serialize the meta info. 
So if in fact it's constant for lifetime of the job, it would be good to avoid this in diskread, but also e.g. keyed join and fetch.

ghalliday:
@jakesmith new version that also includes the flags for disk read and fetch.  I didn't implement join because there were no spare flags, so need to think about whether it is worth adding a new function.

ecl/hqlcpp/hqlcpputil.cpp
```diff
@@ -442,3 +442,14 @@ bool worthGeneratingRowAsSingleActivity(IHqlExpression * expr)
         expr = expr->queryChild(0);
     }
 }
+
+bool isNonConstantAndQueryInvariant(IHqlExpression * expr)
+{
+    if (expr->isConstant())
```
jakesmith: can this be true, and hasDynamicFilename(info.queryFile()) also be true?

jakesmith:
@ghalliday - 1 question, but looks fine.

ecl/hqlcpp/hqlcpputil.cpp
ghalliday: @jakesmith yes I think so.  DYNAMIC(filename) means resolve the filename each time - even if it is constant.  Is that ok?

jakesmith: @ghalliday, yes I think so, I will only assume invariant as this logic implies, when the FFinvariantfilename is present.

ghalliday:
@richardkchapman please review/check comments above

ecl/hqlcpp/hqlcpputil.cpp
richardkchapman: Seems like a weird condition to want to know. Why are constant filenames not considered context-independent?

ghalliday: @richardkchapman  Not sure.  I think it was purely to minimize the number of changes in the helper files - since almost all examples are constant (and simplify my comparison when regression checking).  I think we could equally include constant expressions - I can change if you prefer.

richardkchapman:
Missing some checkboxes

PR #13174, opened 2019-11-21
HPCC-22971 Avoid creating compound disk reads for alien types

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/ecl/hql/hqlattr.cpp
+++ b/ecl/hql/hqlattr.cpp
@@ -4069,3 +4069,82 @@ CHqlMetaProperty * queryMetaProperty(IHqlExpression * expr)
     CHqlExprMeta::addProperty(body, EPmeta, info);
     return info;
 }
+
+//--------------------------------------------------------------------------------------------------------------------
+
+inline bool isSelfSelect(IHqlExpression * expr)
+{
+    if (expr->getOperator() != no_select)
+        return false;
+    if (expr->queryChild(0)->getOperator() != no_selfref)
+        return false;
+    return true;
+}
+
+//This should possibly cache the results for a record using an attribute....
+bool canDefinitelyProcessWithTranslator(IHqlExpression * record)
+{
+    dbgassertex(record->getOperator() == no_record);
+    ForEachChild(i, record)
+    {
+        IHqlExpression * cur = record->queryChild(i);
+        switch (cur->getOperator())
+        {
+        case no_field:
+        {
+            ITypeInfo * type = cur->queryType();
+            switch (type->getTypeCode())
+            {
+            case type_alien:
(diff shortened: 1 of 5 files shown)


ghalliday:
@jakesmith @richardkchapman do we agree this is a good idea?  See jira for details of examples that change.

jakesmith:
>If we never created compound activities for datasets with alien data types and complex ifblocks then it would allow a very simple disk activity to be kept for the case where alien datatypes are mixed with virtual fields.

I'm not clear on what you mean here.
If you prevent compound activities being generated in alien/ifblock cases, that do you mean by it "would allow a very simple disk activity to be kept for the case where alien datatypes are mixed with virtual fields" ?

ghalliday:
Response to Jake:
Going forward the fileposition needs to be handled by the translation code (e.g. to support remote reading and filtering), but alien data types are not supported by the translation code - so we need another approach or to disallow the functionality.  But it is legitimate to want to use complex ifs/alien data types and and filepositions when reading external data files..

It would be a pain (if not impossible) to retain the current disk reading code with all its complexity including updating a local fileposition.
However it would be reasonable to have a disk read activity which only supports an absolute minimum of functionality which can be used for this case (and selected with a flag supplied by the code generator).  This allows the general case to be as efficient as possible, and the code for this weird special case to be as minimal as possible.

jakesmith:
Are there real world cases in use, where alien data types and virtual fields are used together?

ghalliday:
Yes there are (e.g. using an alien data type to represent an image which is fetched from a file using fetch).  In that case they should not be using an alien data type, but I think there are also other more legitimate cases when it is an external data file.

ecl/hql/hqlattr.cpp
```diff
@@ -4069,3 +4069,82 @@ CHqlMetaProperty * queryMetaProperty(IHqlExpression * expr)
...
+            case type_row:
+                if (hasReferenceModifier(type))  // Never currently generated
+                    return false;
+                if (!canDefinitelyProcessWithTranslator(cur->queryRecord()))
+                    return false;
+                break;
+            case type_dictionary:
+            case type_groupedtable:
+            case type_table:
+                {
```
jakesmith: pedantic/formatting: inconsistent indentation of brace vs other case statement (line 4094).

jakesmith:
@ghalliday - looks sensible to me

ghalliday:
@richardkchapman please review

PR #13187, opened 2019-11-26
HPCC-23170 DFU server should copy most file part attributes, except those it has to set, like size and filecrc

AttilaVamos:
Copy all file parts attributes which "not has to set" .

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

## Testing:

Tested manually.

### DIFF
--- a/dali/ft/filecopy.cpp
+++ b/dali/ft/filecopy.cpp
@@ -3154,21 +3154,24 @@ void FileSprayer::updateTargetProperties()
 
                     curProps.setProp("@modified", temp.getString(timestr).str());
                 }
-                if (replicate && (distributedSource != distributedTarget) )
+                if ((distributedSource != distributedTarget) && (cur.whichInput != (unsigned)-1))
                 {
-                    assertex(cur.whichInput != (unsigned)-1);
                     FilePartInfo & curSource = sources.item(cur.whichInput);
                     if (curSource.properties)
                     {
                         Owned<IAttributeIterator> aiter = curSource.properties->getAttributes();
-                        //At the moment only clone the topLevelKey indicator (stored in kind), but make it easy to add others.
-                        ForEach(*aiter) {
+                        ForEach(*aiter)
+                        {
                             const char *aname = aiter->queryName();
-                            if (strieq(aname,"@kind")
-                                ) {
-                                if (!curProps.hasProp(aname))
-                                    curProps.setProp(aname,aiter->queryValue());
-                            }
+                            if ( !( strieq(aname,"@fileCrc") ||
+                                    strieq(aname,"@modified") ||
+                                    strieq(aname,"@node") ||
+                                    strieq(aname,"@num")  ||
+                                    strieq(aname,"@size") ||
+                                    strieq(aname,"@name") ) ||
+                                    ( strieq(aname,"@recordCount") && (sources.ordinality() == targets.ordinality()) )
+                               )
+                                curProps.setProp(aname,aiter->queryValue());
                         }
                     }
                 }
--- a/testing/regress/ecl/split_test.ecl
+++ b/testing/regress/ecl/split_test.ecl
@@ -511,8 +511,6 @@ res14 := gatherCounts(dst14[1].targetFilename, origRecCount, dst14[1].noSplit);
 
 sequential (
 #if (VERBOSE = 1)
-    output(pad),
-    output(length(pad)),
     output(dropzonePath, NAMED('dropzonePath')),
 #end
 


AttilaVamos:
@jakesmith please review.

dali/ft/filecopy.cpp
```diff
@@ -3162,13 +3162,11 @@ void FileSprayer::updateTargetProperties()
...
+                        ForEach(*aiter)
+                        {
                             const char *aname = aiter->queryName();
-                            if (strieq(aname,"@kind")
-                                ) {
-                                if (!curProps.hasProp(aname))
-                                    curProps.setProp(aname,aiter->queryValue());
-                            }
+                            if (!curProps.hasProp(aname))
+                                curProps.setProp(aname,aiter->queryValue());
```
jakesmith: @AttilaVamos  Are you sure this won't copy properties when not valid? e.g. if repartions and number of parts in source and destination aren't same, then properties like size , and fileCrc will not be correct.

AttilaVamos: @jakesmith This is the final step of copy(spray, etc) procedure, so all attributes related the target parts like: size, num, crc, etc has been already set. By my best intention this will copy the extras (kind, recordCount, <future_stuff>) only.

jakesmith: I'm not sure.
Looking at the code before this, it doesn't necessarily (for example) set compressedSize. It depends on whether 'compressOutput' is set or not.
So if it is not, then it will not be set per part (correct),
but the code you've change here, not will copy the property from the source, if it's present (incorrect).

Similar for fileCrc I think, if calcCRC() is false, the proceeding code will not set it per part, but this copy attribute code will carry it over from the source which would be wrong.

jakesmith: Also this block of code is in this conditional block:
```
if (replicate && (distributedSource != distributedTarget) )
```

so only copies attributes if replicate is true?
Meaning still wrong (e.g. missing "kind") if replicate=false ...

jakesmith: @AttilaVamos  back to you

AttilaVamos: Yes, you are right, I should rework this.

AttilaVamos:
@jakesmith please check this attempt.

dali/ft/filecopy.cpp
jakesmith: Are the number of parts of target and source always the same at this point?
If they are, then copying recordCount is good, if not, then it will be wrong.

dali/ft/filecopy.cpp
```diff
@@ -3154,21 +3154,24 @@ void FileSprayer::updateTargetProperties()
 
                     curProps.setProp("@modified", temp.getString(timestr).str());
                 }
-                if (replicate && (distributedSource != distributedTarget) )
+                if ((distributedSource != distributedTarget) && (cur.whichInput != (unsigned)-1))
```
jakesmith: When can cur.whichInput be -1 ?

dali/ft/filecopy.cpp
```diff
@@ -3154,21 +3154,24 @@ void FileSprayer::updateTargetProperties()
...
                 {
-                    assertex(cur.whichInput != (unsigned)-1);
+                    //assertex(cur.whichInput != (unsigned)-1);
```
jakesmith: should remove this line

testing/regress/ecl/split_test.ecl
```diff
@@ -511,8 +511,6 @@ res14 := gatherCounts(dst14[1].targetFilename, origRecCount, dst14[1].noSplit);
 
 sequential (
 #if (VERBOSE = 1)
-    output(pad),
-    output(length(pad)),
```
jakesmith: is change related/intentional ?

jakesmith:
@AttilaVamos - may be ready, a couple of questions

dali/ft/filecopy.cpp
AttilaVamos: When there is no (physical) input file like superfile/-key copy or spray empty directory.

dali/ft/filecopy.cpp
AttilaVamos: Removed

testing/regress/ecl/split_test.ecl
AttilaVamos: This test failed in Smoketest during rework and executed in local environment. When verbose was enabled this test didn't compile based on that missing reference (pad).

dali/ft/filecopy.cpp
AttilaVamos: That is a good point. No, it isn't (only for key or keydiff files). I will push an update soon.

AttilaVamos:
@jakesmith please re-review.

jakesmith:
Looks good.

PR #13210, opened 2019-12-09
HPCC-22504 Ensure NumStarts matches NumStops in Thor

shamser:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [x] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/thorlcr/graph/thgraphmaster.cpp
+++ b/thorlcr/graph/thgraphmaster.cpp
@@ -2958,19 +2958,22 @@ CTimingInfo::CTimingInfo(CJobBase &ctx) : CThorStats(ctx, StTimeLocalExecute)
 
 ProgressInfo::ProgressInfo(CJobBase &ctx) : CThorStats(ctx, StNumRowsProcessed)
 {
-    startcount = stopcount = 0;
+    startCount = stopCount = 0;
 }
 void ProgressInfo::processInfo() // reimplement as counts have special flags (i.e. stop/start)
 {
     reset();
-    startcount = stopcount = 0;
+    startCount = stopCount = 0;
     ForEachItemIn(n, counts)
     {
         unsigned __int64 thiscount = counts.item(n);
-        if (thiscount & THORDATALINK_STARTED)
-            startcount++;
         if (thiscount & THORDATALINK_STOPPED)
-            stopcount++;
+        {
+            startCount++;
+            stopCount++;
+        }
+        else if (thiscount & THORDATALINK_STARTED)
+            startCount++;
         thiscount = thiscount & THORDATALINK_COUNT_MASK;
         tallyValue(thiscount, n+1);
     }
@@ -2982,8 +2985,8 @@ void ProgressInfo::getStats(IStatisticGatherer & stats)
     CThorStats::getStats(stats, true);
     stats.addStatistic(kind, tot);
     stats.addStatistic(StNumSlaves, counts.ordinality());
-    stats.addStatistic(StNumStarts, startcount);
-    stats.addStatistic(StNumStops, stopcount);
+    stats.addStatistic(StNumStarts, startCount);
+    stats.addStatistic(StNumStops, stopCount);
 }
 
 
--- a/thorlcr/graph/thgraphslave.hpp
+++ b/thorlcr/graph/thgraphslave.hpp
@@ -78,7 +78,8 @@ public:
 
     inline void dataLinkStop()
     {
-        count = (count & THORDATALINK_COUNT_MASK) | THORDATALINK_STOPPED;
+        if (hasStarted())
+            count = (count & THORDATALINK_COUNT_MASK) | THORDATALINK_STOPPED;
 #ifdef _TESTING
         owner.ActPrintLog("ITDL output %d stopped, count was %" RCPF "d", outputId, getDataLinkCount());
 #endif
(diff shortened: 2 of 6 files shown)


shamser:
This PR make sure that the NumStart matches the NumStops.  It doesn't actually track the exact number of starts and stops.  The work to track the exact number of starts & stops will be done in this: https://track.hpccsystems.com/browse/HPCC-23201.

@jakesmith  Please can you review.

thorlcr/graph/thgraphslave.hpp
```diff
@@ -70,15 +70,15 @@ class CEdgeProgress
             if (!hasStopped())
                 throw MakeActivityException(&owner, 0, "Starting without being stopped 1st");
             else
-                throw MakeActivityException(&owner, 0, "Started and stopped states both set");
+                count &= THORDATALINK_COUNT_MASK;
```
jakesmith: this line is not necessary, as will fall through to line 76 that masks and sets.

thorlcr/graph/thgraphslave.hpp
```diff
@@ -70,15 +70,15 @@ class CEdgeProgress
...
         }
         icount = 0;
         count = (count & THORDATALINK_COUNT_MASK) | THORDATALINK_STARTED;
     }
 
     inline void dataLinkStop()
     {
-        count = (count & THORDATALINK_COUNT_MASK) | THORDATALINK_STOPPED;
+        count |= THORDATALINK_STOPPED;
```
jakesmith: I'm a bit concerned this changes the current semantics of hasStarted().
Currently (before this change), it would be false after stop, after this change, it will remain true.
So now means "has ever started".
There is a lot of code that is dependent on hasStarted(), and this may well cause that code to be recalled when it should not be.

Currently, it's impossible to track CQ activities start/stopped state - because of the single bit.
This change would allow it in theory to track the 'has it ever started' state, but that's not the same, and this change may break other things.

jakesmith:
@shamser - see comments re. concern of change of semantics.
Was there a problem with approach we discussed, that if flagged stopped, then you know started ? 

Also the new wucheckstartstops.py file, looks like it has a lot in common with the existing python wutest.py file ? It would be good to avoid duplicating, by moving the common code to a python module.

thorlcr/graph/thgraphslave.hpp
shamser: @jakesmith Ok.  I understand the change is going to cause a problem.

I have examined the other ways of doing this. If we were to serialize the start and stop state separately, we could either have the master use the last recieved value or sum the value. If it were to use the last recieved value, it would always eventually end-up with start=false, stop=true.  If we can't really sum the values, as the NumStarts is likely to be serialized repeatedly for long-running activities.

The only thing I can think of is separating NumStart/NumStops from hasStarted().  Could we create a separate variable for hasStarted?

shamser:
> @shamser - see comments re. concern of change of semantics.
> Was there a problem with approach we discussed, that if flagged stopped, then you know started ?
> 
> Also the new wucheckstartstops.py file, looks like it has a lot in common with the existing python wutest.py file ? It would be good to avoid duplicating, by moving the common code to a python module.

shamser:
> @shamser - see comments re. concern of change of semantics.
> Was there a problem with approach we discussed, that if flagged stopped, then you know started ?
So, on the master if numStopped>0 & numStarted==0, set numStarted=numStarted?

> 
> Also the new wucheckstartstops.py file, looks like it has a lot in common with the existing python wutest.py file ? It would be good to avoid duplicating, by moving the common code to a python module.
I'll common up this code.

jakesmith:
>So, on the master if numStopped>0 & numStarted==0, set numStarted=numStarted?

Yes.
One change you may want to make, is in dataLinkStop(). It currently sets THORDATALINK_STOPPED unconditionally. 
stop() (and dataLinkStop()) can be called without the activity ever having starts - this is intentional, so that upstream activities are aware that they are not needed (e.g. some arms of splitter not going to be started).

It think it would be safe to change this so that dataLinkStop() only sets THORDATALINK_STOPPED, if hasStarted(). This would prevent spuriously showing started in the progress info., when in fact it had only been stopped and never started.
NB: if you do this, would want to do similar for HPCC-23201 when dealing with separate started/stopped counts.

shamser:
@jakesmith I've modified as suggested.  Please can you review.

thorlcr/graph/thgraphmaster.cpp
```diff
@@ -2961,10 +2961,12 @@ void ProgressInfo::processInfo() // reimplement as counts have special flags (i.
     ForEachItemIn(n, counts)
     {
         unsigned __int64 thiscount = counts.item(n);
-        if (thiscount & THORDATALINK_STARTED)
-            startcount++;
         if (thiscount & THORDATALINK_STOPPED)
+        {
+            startcount++;
```
jakesmith: trivial: perhaps change to camelCase whilst changing..

thorlcr/graph/thgraphmaster.cpp
```diff
@@ -2961,10 +2961,12 @@ void ProgressInfo::processInfo() // reimplement as counts have special flags (i.
...
             stopcount++;
+        } else if (thiscount & THORDATALINK_STARTED)
```
jakesmith: style: else on newline to be consistent with rest of class/file (and preferred style anyway).

jakesmith:
The python changes, I've looked through and haven't spotted any issues.

@shamser - couple of minor/trivial comments.

shamser:
@jakesmith

jakesmith:
@shamser - looks fine please squash, should this be master or 7.6.x?

shamser:
@jakesmith Ready to merge?

jakesmith:

PR #13232, opened 2019-12-18
HPCC-23230 Rationalize createValue() and createDataset()

ghalliday:
## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:


ghalliday:
@shamser please review.
Note, some of the persist crcs changed.  I think that is probably ok for a minor/major release.

ghalliday:
note, before merging, we should double check that changing the crcs for ECL (not for records) is acceptable.

ghalliday:
Note a few examples do change - e.g. aslow1/vlad18 because now that the expressions have consistent types they are commoned up.

common/fileview2/fvwugen.cpp
```diff
@@ -54,12 +54,12 @@ MODULE_INIT(INIT_PRIORITY_STANDARD)
 IHqlExpression * addFilter(IHqlExpression * dataset, IHqlExpression * limitField)
 {
     IHqlExpression * lower = createConstant(limitField->queryType()->castFrom(true, (__int64)0));
-    lower = createValue(no_colon, lower, createValue(no_stored, createConstant(LOWER_LIMIT_ID)));
+    lower = createValue(no_colon, lower->getType(), lower, createValue(no_stored, makeVoidType(), createConstant(LOWER_LIMIT_ID)));
```
shamser: Why does createConstant(LOWER_LIMIT_ID) have void type?

ecl/hql/hqlgram.y
```diff
@@ -2091,7 +2091,8 @@ conditionalAssignmentElseClause
                             parser->normalizeExpression($2);
                             parser->ensureBoolean($2);
                             //normalizeExpression($2, type_boolean, false);
-                            OwnedHqlExpr map = createValue(no_mapto, $2.getExpr(), $4.getExpr());
+                            OwnedHqlExpr other = $4.getExpr();
+                            OwnedHqlExpr map = createValue(no_mapto, other->getType(), $2.getExpr(), LINK(other));
```
shamser: Previously, wouldn't the expression type be $2->getType()?

ecl/hql/hqlgram.y
```diff
@@ -7341,7 +7342,7 @@ scopeFunctionWithParameters
                             OwnedHqlExpr parms = $4.getExpr();
                             //NB: Do not call createComma() incase the first argument is a dataset
                             if (parms)
-                                $$.setExpr(createValue(no_comma, $1.getExpr(), parms.getClear()), $1);
+                                $$.setExpr(createValue(no_comma, nullptr, $1.getExpr(), parms.getClear()), $1);
```
shamser: Why is the type $1.getType()?

common/fileview2/fvwugen.cpp
```diff
@@ -112,7 +112,7 @@ IHqlExpression * buildDiskFileViewerEcl(const char * logicalName, IHqlExpression
     OwnedHqlExpr newRecord = createRecord(fields);
     newRecord.setown(createSymbol(createIdAtom("_SourceRecord_"), newRecord.getLink(), ob_private));
 
-    OwnedHqlExpr dataset = createNewDataset(createConstant(logicalName), newRecord.getLink(), createValue(no_thor), NULL, NULL, NULL);
+    OwnedHqlExpr dataset = createNewDataset(createConstant(logicalName), newRecord.getLink(), createValue(no_thor, makeNullType()), NULL, NULL, NULL);
```
shamser: Should NULL be replaced by nullptr?

common/fileview2/fvwugen.cpp
```diff
@@ -182,7 +182,7 @@ IHqlExpression * PositionTransformer::createTransformed(IHqlExpression * _expr)
                 {
                     IHqlExpression * newTarget = createField(child->queryId(), child->getType(), LINK(child), insertedAttr.getLink());
                     fields.append(*newTarget);
-                    assigns.append(*createValue(no_assign, makeVoidType(), newTarget, createSelectExpr(createValue(no_left), LINK(newTarget))));
+                    assigns.append(*createValue(no_assign, makeVoidType(), newTarget, createSelectExpr(createValue(no_left, makeNullType()), LINK(newTarget)))); // MORE: This probably doesn't work
```
shamser: Should probably comment why this doesn't work.

ecl/hql/hqlgram2.cpp
```diff
@@ -1688,13 +1688,15 @@ void HqlGram::addAssignment(attribute & target, attribute &source)
         {
             case type_record:
             case type_row:
+            case type_null: // Assign no_null
```
shamser: Why is this allowed?  When would the source be type_null?

shamser (commented):
@ghalliday Can all instances of createDatasetF by the new createDataset with std::initializer_list?

common/fileview2/fvwugen.cpp
ghalliday: It doesn't. It is the stored workflow action that has the void type, with a parameter of LOWER_LIMIT_ID.   E.g. see equivalent entries in the grammar.

ecl/hql/hqlgram.y
ghalliday: Yes, and it was wrong.  The type of   expr1=>expr2 should be the type of expr2.

ecl/hql/hqlgram.y
ghalliday: I think the type is actually a nullptr in this case.

common/fileview2/fvwugen.cpp
ghalliday: Long term yes, but it should be a separate PR to replace NULL with nullptr.  Possibly throughout the source.

common/fileview2/fvwugen.cpp
ghalliday: ok

ecl/hql/hqlgram2.cpp
ghalliday: When it is a no_null.  I.e.  [ ]

ghalliday:
> @ghalliday Can all instances of createDatasetF by the new createDataset with std::initializer_list?

I think so - it should be safer since you don't have to remember the extra null on the end.  All new code should use it.

common/fileview2/fvwugen.cpp
ghalliday: Updated comment - but note, code is deprecated and should really be deleted.

ghalliday:
@shamser some replies

shamser (approved):
@richardkchapman Reviewed.

ecl/hql/hqlexpr.hpp
```diff
@@ -1276,20 +1278,24 @@ struct OwnedHqlExprItem : public CInterface
...
-extern HQL_API IHqlExpression *createValue(node_operator op, IHqlExpression *p1, IHqlExpression *p2);
 extern HQL_API IHqlExpression *createOpenValue(node_operator op, ITypeInfo *type);
 extern HQL_API IHqlExpression *createOpenNamedValue(node_operator op, ITypeInfo *type, IIdAtom * id);
 extern HQL_API IHqlExpression *createNamedValue(node_operator op, ITypeInfo *type, IIdAtom * id, HqlExprArray & args);
 extern HQL_API IHqlExpression *createId(IIdAtom * id);
 
+extern HQL_API IHqlExpression *createValue(node_operator op);
+extern HQL_API IHqlExpression *createValue(node_operator op, IHqlExpression *p1, IHqlExpression *p2);
+extern HQL_API IHqlExpression* createValue(node_operator op, HqlExprArray& operands);
+extern HQL_API IHqlExpression *createValue(node_operator op, IHqlExpression *p1);
```
richardkchapman: picky, but feels more natural for this to be two lines higher...

ecl/hql/hqlexpr.cpp
```diff
@@ -5188,6 +5188,24 @@ CHqlExpression *CHqlExpressionWithType::makeExpression(node_operator _op, ITypeI
...
+{
+    HqlExprArray args;
+    args.ensure(operands.size());
+    for (auto & cur : operands)
+    {
+        //Skip null entries
+        if (cur)
+        {
+#ifdef _DEBUG
+            assertex(QUERYINTERFACE(cur, IHqlExpression));
```
richardkchapman: Why not dbgassertex (as you used below)

ecl/hql/hqlexpr.cpp
```diff
@@ -5188,6 +5188,24 @@ CHqlExpression *CHqlExpressionWithType::makeExpression(node_operator _op, ITypeI
     return (CHqlExpression *)e->closeExpr();
 }
 
+CHqlExpression* CHqlExpressionWithType::makeExpression(node_operator op, ITypeInfo *type, const std::initializer_list<IHqlExpression *> &operands)
+{
+    HqlExprArray args;
+    args.ensure(operands.size());
```
richardkchapman: Does the fact that you skip nulls make this ensure() less desirable?

ecl/hql/hqlexpr.cpp
```diff
@@ -11510,6 +11528,27 @@ extern IHqlExpression *createDatasetF(node_operator op, ...)
     return createDataset(op, children);
 }
 
+IHqlExpression *createDataset(node_operator op, const std::initializer_list<IHqlExpression *> &operands)
+{
+    HqlExprArray args;
+    for (auto & cur : operands)
```
richardkchapman: No call to ensure() here?

ecl/hql/hqlgram2.cpp
```diff
@@ -7204,13 +7204,13 @@ IHqlExpression * HqlGram::createBuildIndexFromIndex(attribute & indexAttr, attri
 
     IHqlExpression * select;
     if (sourceDataset)
-        select = createDatasetF(no_newusertable, LINK(sourceDataset), LINK(record), LINK(transform), NULL); //createUniqueId(), NULL);
+        select = createDataset(no_newusertable, { LINK(sourceDataset), LINK(record), LINK(transform) }); //createUniqueId(), NULL);
```
richardkchapman: Comment is now inaccurate I think?

ecl/hql/hqlgram2.cpp
```diff
@@ -7204,13 +7204,13 @@ IHqlExpression * HqlGram::createBuildIndexFromIndex(attribute & indexAttr, attri
...
     else if (transform)
-        select = createDatasetF(no_newusertable, LINK(dataset), LINK(record), LINK(transform), NULL); //createUniqueId(), NULL);
+        select = createDataset(no_newusertable, { LINK(dataset), LINK(record), LINK(transform) }); //createUniqueId(), NULL);
```
richardkchapman: ditto

ecl/hql/hqlgram2.cpp
```diff
@@ -7204,13 +7204,13 @@ IHqlExpression * HqlGram::createBuildIndexFromIndex(attribute & indexAttr, attri
...
     else
     {
         IHqlExpression * newRecord = checkBuildIndexRecord(LINK(record), errpos);
-        select = createDatasetF(no_selectfields, LINK(dataset), newRecord, NULL); //createUniqueId(), NULL);
+        select = createDataset(no_selectfields, { LINK(dataset), newRecord }); //createUniqueId(), NULL);
```
richardkchapman: ditto

richardkchapman:
@ghalliday Some minor comments

ecl/hql/hqlexpr.cpp
ghalliday: I think because the code was copied and modified.  I will change.

ecl/hql/hqlexpr.cpp
ghalliday: Well spotted.  I debated the same point, and thought it is was probably ok.  I could count them first..

ecl/hql/hqlexpr.cpp
ghalliday: No - I also debated this pone.  It currently isn't because some could be commas and therefore exapanded, and some could be null.

ecl/hql/hqlexpr.hpp
ghalliday: I agree, moved in subsequent commit.

PR #13233, opened 2019-12-18
HPCC-23232 Eclagent's use of QueryTerminationCleanup wrong.

jakesmith:
QueryTerminationCleanup at the end process, must clearup
before the query dll is unloaded.
If it waits until atexit, any hooks installed by the query
dll of dependent dlls, will be called to late and could crash.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/ecl/eclagent/eclagent.cpp
+++ b/ecl/eclagent/eclagent.cpp
@@ -1877,7 +1877,7 @@ void EclAgent::doProcess()
         {
             MTIME_SECTION(queryActiveTimer(), "Process");
             Owned<IEclProcess> process = loadProcess();
-            QueryTerminationCleanup threadCleanup;
+            QueryTerminationCleanup threadCleanup(false);
 
             if (checkVersion && (process->getActivityVersion() != eclccCodeVersion))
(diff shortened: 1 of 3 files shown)


jakesmith:
@ghalliday - please review

ghalliday:
@richardkchapman @jakesmith this change looks reasonable.

As a summary it is trying to fix the problems with gcc < 2.18 where thread termination functions can be called after dlls are unloaded.  If the change is needed for hthor then I suspect thor, roxie (and anything else?) should also be explicitly calling the hook destructors on the main thread before exiting.

jakesmith:
I think you're right, but have only seen this being hit on eclagent so far.
Shall we merge this one, which bit us, and open a separate JIRA for otehrs?
@ghalliday @richardkchapman

mckellyln:
Perhaps not seen in thor, roxie because eclagent/hthor is started/stopped much more often.

richardkchapman:
Roxie is never stopped other than via a kill... and it's main thread never calls any of the plugins that might add a hook.

Thor may also not ever use its main thread to call plugins?

PR #13235, opened 2019-12-19
HPCC-23234 Fix problems if KJ in CQ and stopped early.

jakesmith:
If a KJ is inside a CQ and it's stopped early (e.g. due to a
downstream CHOOSEN), then it can lead to spurious results,
because it is not reset fully on restart.

Also problems with the limiters not being reset, which effectively
meant that it had an unbound lookahead.
That could cause memory issues, probably more of an issue if
KJ is inside a LOOP where each execution would be expected to
create larger outputs than if inside a CQ.

NB: commit also includes some minor tidying (e.g. moving const variables together).

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/thorlcr/activities/keyedjoin/thkeyedjoinslave.cpp
+++ b/thorlcr/activities/keyedjoin/thkeyedjoinslave.cpp
@@ -695,9 +695,7 @@ class CJoinGroupList
     CJoinGroup *head = nullptr, *tail = nullptr;
     unsigned count = 0;
 
-public:
-    CJoinGroupList() { }
-    ~CJoinGroupList()
+    void removeAll()
     {
         while (head)
         {
@@ -705,6 +703,17 @@ public:
             head->Release();
             head = next;
         }
+    }
+public:
+    CJoinGroupList() { }
+    ~CJoinGroupList()
+    {
+        removeAll();
+    }
+    void clear()
+    {
+        removeAll();
+        head = tail = nullptr;
     }
     inline unsigned queryCount() const { return count; }
     inline CJoinGroup *queryHead() const { return head; }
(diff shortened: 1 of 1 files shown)


thorlcr/activities/keyedjoin/thkeyedjoinslave.cpp
```diff
@@ -706,6 +704,17 @@ class CJoinGroupList
             head = next;
         }
     }
+public:
+    CJoinGroupList() { }
+    ~CJoinGroupList()
+    {
+        removeAll();
+    }
+    void clear()
```
ghalliday: The split looks sensible, but this function doesn't seem to be called.

ghalliday: I take it back, my search wasn't finding it.

jakesmith: called on [new] lines 3048-3049

ghalliday:
@jakesmith change looks sensible.  Any chance of a test case?

jakesmith:
@ghalliday I discovered these issues, whilst developing a fix for HPCC-23211.
I've got a test case for that in the PR (https://github.com/hpcc-systems/HPCC-Platform/pull/13238), that exposes and tests this issue, as part of testing the fix.

so if we can wait for a test case, until HPCC-23211 is ready?

ghalliday:
@jakesmith as long as there will be one eventually that is fine by me.  Is 7.6.x the correct target?

jakesmith:
>Is 7.6.x the correct target?

I think so, it hasn't actually been reported, or new.

PR #13244, opened 2020-01-02
HPCC-23202 Update Kafka plugin to use latest released Kafka client library

dcamper:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

## Testing:

Manual testing against latest Kafka server and client libraries.

### DIFF
--- a/plugins/kafka/kafka.cpp
+++ b/plugins/kafka/kafka.cpp
@@ -130,12 +130,15 @@ namespace KafkaPlugin
 
     KafkaStreamedDataset::~KafkaStreamedDataset()
     {
-        if (consumedRecCount > 0)
+        if (consumerPtr)
         {
-            consumerPtr->commitOffset(lastMsgOffset);
-        }
+            if (consumedRecCount > 0)
+            {
+                consumerPtr->commitOffset(lastMsgOffset);
+            }
 
-        delete(consumerPtr);
+            delete(consumerPtr);
+        }
     }
 
     const void* KafkaStreamedDataset::nextRow()
@@ -145,7 +148,7 @@ namespace KafkaPlugin
         __int32 timeoutWait = 100;  //!< Amount of time to wait between retries
         __int32 attemptNum = 0;
 
-        if (maxRecords <= 0 || consumedRecCount < maxRecords)
+        if (consumerPtr && (maxRecords <= 0 || consumedRecCount < maxRecords))
         {
             RdKafka::Message* messageObjPtr = NULL;
             bool messageConsumed = false;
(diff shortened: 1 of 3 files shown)


dcamper:
@ghalliday Please review or reassign.  Thanks!

plugins/kafka/kafka.cpp
```diff
@@ -506,8 +526,17 @@ namespace KafkaPlugin
             consumerPtr->stop(topicPtr, partitionNum);
         }
 
-        delete(topicPtr.load());
-        delete(consumerPtr);
+        if (topicPtr)
```
ghalliday: Why this change?  You don't need to protect delete of a null pointer, and this is within the destructor so the values will never be accessed.

plugins/kafka/kafka.cpp
```diff
@@ -506,8 +526,17 @@ namespace KafkaPlugin
...
+        {
+            delete(topicPtr.load());
+            topicPtr = NULL;
+        }
+
+        if (consumerPtr)
```
ghalliday: same here

plugins/kafka/kafka.cpp
```diff
@@ -591,45 +624,41 @@ namespace KafkaPlugin
...
             }
         }
-        else
+        catch (...)
         {
-            throw MakeStringException(-1, "Kafka: Failed to start Consumer read for %s:%d @ %s; error: %d", topic.c_str(), partitionNum, brokers.c_str(), startErr);
+            // The current object is allocated on the heap and was supposed to be managed by the
+            // IRowStream subclass we're returning; since we've encounted an error, we need
+            // to self-destruct to avoid memory leaks
+            delete(this);
```
ghalliday: This rings some alarm bells.  Better would be to have the catch in the global getMessageDataset() function and delete consumerObjPtr.  (or use std::unique_ptr)

ghalliday:
@dcamper a few comments/questions

plugins/kafka/kafka.cpp
dcamper: Will revert.

plugins/kafka/kafka.cpp
dcamper: Will revert.

plugins/kafka/kafka.cpp
dcamper: Moving to using a catch in the calling function.

dcamper:
Changes pushed.  @ghalliday Please recheck.  Thanks!

ghalliday:
looks good to me.  @richardkchapman letting you double check before merging.

PR #13258, opened 2020-01-07
HPCC-22136 Import WU from ZAP report

wangkx:
This PR is created to replace https://github.com/hpcc-systems/HPCC-Platform/pull/13111. Please see https://github.com/hpcc-systems/HPCC-Platform/pull/13111#issuecomment-562689543.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/common/workunit/workunit.cpp
+++ b/common/workunit/workunit.cpp
@@ -59,6 +59,8 @@
 #include <string>
 #include <algorithm>
 
+#define ECLAGENTLOGSEARCHSTR        "eclagent."
+
 using namespace cryptohelper;
 
 static int workUnitTraceLevel = 1;
@@ -172,6 +174,37 @@ void doDescheduleWorkkunit(char const * wuid)
     Owned<IPropertyTree> root = conn->getRoot();
     bool more;
     do more = root->removeProp(xpath.str()); while(more);
+}
+
+StringBuffer & getProcessLogDateFromLogFileName(const char * processType, const char * logFileNameWithPath, StringBuffer & logDate)
+{
+    //The existing thor master stores the information about a WU into thor master logs
+    //which are named like '...thormaster.yyyy_mm_dd.log'. The datePtr is pointed at
+    //the end of the 'thormaster.'. After that, there is the 'yyyy_mm_dd' part which 
+    //is 10 bytes long. The strlen(datePtr) should >= 10.
+    //If thor master stores all of its log lines for a WU into one log file, there is
+    //no need for the log file name to have the 'yyyy_mm_dd' part. The log name may
+    //look like '...thormaster.log'. The datePtr is pointed at the end of the 'thormaster.'.
+    //The strlen(datePtr) should = 3.
+    //So, if (strlen(datePtr) >= 10), we read the logDate (10 bytes) at the datePtr.
+    //ECLAgent log is similar.
+    const char * searchStr = nullptr;
+    if (strieq(processType, "Thor"))
+        searchStr = THORMASTERLOGSEARCHSTR;
+    else
+        searchStr = ECLAGENTLOGSEARCHSTR;
+
+    const char * logFileName = pathTail(logFileNameWithPath);
+    const char * datePtr = strstr(logFileName, searchStr);
+    if (!datePtr)
+        return logDate;
+
+    datePtr += strlen(searchStr);
+    if (strlen(datePtr) < 10) //10 bytes for yyyy_mm_dd
+        return logDate;
+
+    logDate.append(10, datePtr);
+    return logDate;
 }
 
 //======================================================
@@ -3829,6 +3862,21 @@ public:
     {
         return new CDaliWuGraphStats(getWritableProgressConnection(graphName, _wfid), creatorType, creator, _wfid, graphName, subgraph);
     }
+    virtual void import(IPropertyTree *wuTree, IPropertyTree *graphProgressTree)
+    {
+        connection->queryRoot()->setPropTree(nullptr, LINK(wuTree));
+        loadPTree(connection->getRoot());
+
+        if (!graphProgressTree)
+            return;
+
+        VStringBuffer xpath("/GraphProgress/%s", queryWuid());
+        Owned<IRemoteConnection> progressConn = querySDS().connect(xpath, myProcessSession(), RTM_LOCK_WRITE|RTM_CREATE, SDS_LOCK_TIMEOUT);
+        if (!progressConn)
+            throw MakeStringException(0, "Failed to access %s.", xpath.str());
+
+        progressConn->queryRoot()->setPropTree(nullptr, LINK(graphProgressTree));
+    }
 
 protected:
     IRemoteConnection *getProgressConnection() const
@@ -4118,6 +4166,14 @@ public:
             { return c->getAbortBy(str); }
     virtual unsigned __int64 getAbortTimeStamp() const
             { return c->getAbortTimeStamp(); }
+    virtual bool usingDedicatedLogFiles(const char* processType, const char* processName, const char* logSpec) const
+            { return c->usingDedicatedLogFiles(processType, processName, logSpec); }
+    virtual StringBuffer & getSlaveLogFileName(const char * thorProcess, int slaveNum, const char *ipAddress, StringBuffer & logFileName, bool withPath) const
+            { return c->getSlaveLogFileName(thorProcess, slaveNum, ipAddress, logFileName, withPath); }
+    virtual void getWUProcessLogInfo(const char *processType, const char *processName, IArrayOf<IConstWUProcessLogInfo> &processLogs) const
+            { c->getWUProcessLogInfo(processType, processName, processLogs); }
+    virtual void import(IPropertyTree *wuTree, IPropertyTree *graphProgressTree)
+            { return c->import(wuTree, graphProgressTree); }
 
 
     virtual void clearExceptions(const char *source=nullptr)
@@ -4126,8 +4182,8 @@ public:
             { c->commit(); }
     virtual IWUException * createException()
             { return c->createException(); }
-    virtual void addProcess(const char *type, const char *instance, unsigned pid, const char *log)
-            { c->addProcess(type, instance, pid, log); }
+    virtual void addProcess(const char *type, const char *instance, unsigned pid, unsigned max, const char *pattern, bool singleLog, const char *log)
+            { c->addProcess(type, instance, pid, max, pattern, singleLog, log); }
     virtual void protect(bool protectMode)
             { c->protect(protectMode); }
     virtual void setAction(WUAction action)
@@ -5113,6 +5169,303 @@ bool CWorkUnitFactory::restoreWorkUnit(const char *base, const char *wuid, bool 
     return true;
 }
 
+void CWorkUnitFactory::importWorkUnit(const char *zapReportFileName, const char *zapReportFilePath,
+    const char *zapReportPassword, const char *importDir, const char *app, const char *user, ISecManager *secMgr, ISecUser *secUser)
+{
+    class LogEntry : public CInterface
+    {
+    public:
+        LogEntry(const char *_log, const char *_process) : log(_log), process(_process) {}
+
+        StringAttr process, log;
+        StringBuffer slaveLogPattern;
+        unsigned numberOfSlaves;
+    };
+    class CImportWorkUnitHelper
+    {
+        StringAttr zapReportFileName, zapReportFilePath, zapReportPassword, importDir, user, wuid, fromWUID, scope;
+        StringBuffer unzipDir, unzipDateTime;
+        Owned<IPTree> wuTree, graphProgressTree;
+        CIArrayOf<LogEntry> thorLogEntries;
+        CIArrayOf<LogEntry> eclAgentLogEntries;
+
+        bool findZAPFile(const char *mask, bool optional, StringBuffer &fileName)
+        {
+            Owned<IFile> d = createIFile(unzipDir);
+            if (!d->exists())
+                throw MakeStringException(WUERR_InvalidUserInput, "%s not found.", unzipDir.str());
+
+            Owned<IDirectoryIterator> di = d->directoryFiles(mask, false, false);
+            if (!di->first())
+            {
+                if (optional)
+                    return false;
+                throw MakeStringException(WUERR_InvalidUserInput, "Failed to found %s in %s.", mask, unzipDir.str());
+            }
+
+            fileName.set(unzipDir).append(PATHSEPSTR);
+            di->getName(fileName);
+            if (di->next())
+                throw MakeStringException(WUERR_InvalidUserInput, "More than 1 %s files found in %s.", mask, unzipDir.str());
+            return true;
+        }
+        IPTree *readWUXMLFile(const char *pattern, bool optional)
+        {
+            StringBuffer fileName;
+            if (!findZAPFile(pattern, optional, fileName))
+                return nullptr;
+
+            Owned<IPTree> tree = createPTreeFromXMLFile(fileName);
+            if (!tree)
+                throw MakeStringException(WUERR_InvalidUserInput, "Failed to retrieving %s.", pattern);
+
+            Owned<IFile> f = createIFile(fileName);
+            f->remove();
+            return tree.getClear();
+        }
+        void updateWUProcessLogAttrs(const char *unzipDirWithIP)
+        {
+            IPropertyTree *processTreeRoot = wuTree->queryPropTree("Process");
+            IPropertyTree *oldThorProcessBranch = processTreeRoot->queryPropTree("Thor");
+            IPropertyTree *oldEclAgentProcessBranch = processTreeRoot->queryPropTree("EclAgent");
+
+            VStringBuffer eclAgentLogSearchStr("_%s", ECLAGENTLOGSEARCHSTR);
+            VStringBuffer thorMasterLogSearchStr("_%s", THORMASTERLOGSEARCHSTR);
+            Owned<IFile> unzipDir = createIFile(unzipDirWithIP);
+            Owned<IDirectoryIterator> files = unzipDir->directoryFiles("*.log");
+            ForEach(*files)
+            {
+                const char* file = files->query().queryFilename();
+                StringBuffer name(pathTail(file)); //ProcessName_thormaster.Date or ProcessName_eclagent.Date
+                const char *ptr = strstr(name, eclAgentLogSearchStr);
+                if (ptr)
+                {
+                    name.setLength(ptr - name.str()); //ProcessName
+                    eclAgentLogEntries.append(*new LogEntry(file, name));
+                }
+                else
+                {
+                    ptr = strstr(name, thorMasterLogSearchStr);
+                    if (!ptr)
+                        continue;
+
+                    name.setLength(ptr - name.str()); //ProcessName
+                    Owned<LogEntry> logEntry = new LogEntry(file, name);
+
+                    IPropertyTree *proc = nullptr;
+                    if (oldThorProcessBranch)
+                        proc = oldThorProcessBranch->queryPropTree(name);
+
+                    logEntry->slaveLogPattern.set(name).append("_");
+                    if (proc && proc->hasProp("@pattern"))
+                    {
+                        proc->getProp("@pattern", logEntry->slaveLogPattern);
+                        logEntry->numberOfSlaves = proc->getPropInt("@max");
+                    }
+                    else
+                    {   //legacy WU
+                        getSlaveLogPatternForLegacyWU(file, logEntry->slaveLogPattern);
+                        logEntry->numberOfSlaves = findNumberOfThorSlavesFromSlaveLogs(unzipDirWithIP);
+                    }
+                    thorLogEntries.append(*logEntry.getClear());
+                }
+            }
+            if (oldThorProcessBranch)
+                processTreeRoot->removeTree(oldThorProcessBranch);
+            if (oldEclAgentProcessBranch)
+                processTreeRoot->removeTree(oldEclAgentProcessBranch);
+        }
+        StringBuffer &getSlaveLogPatternForLegacyWU(const char *logSpec, StringBuffer &slaveLogPattern)
+        {
+            StringBuffer logDate;
+            getProcessLogDateFromLogFileName("Thor", logSpec, logDate);
+
+            slaveLogPattern.append(THORSLAVELOGSEARCHSTR).append(SLAVEIDSTR);
+            if (!logDate.isEmpty()) //Should not happen
+                slaveLogPattern.append(".").append(logDate);
+            slaveLogPattern.append(".log");
+            return slaveLogPattern;
+        }
+        unsigned findNumberOfThorSlavesFromSlaveLogs(const char *zapWUFolderPath)
+        {
+            //This is used for legacy WU where the /Process/THORNAME/@max is not set. 
+            //From the log names of thor slaves, find out the largest thor slave ID as
+            //the NumberOfThorSlaves. Inside a ZAP reports, the log names of thor slaves 
+            //has the format: 'xxx_thorslave.SLAVE_ID.xxx'. The pStr is pointed at
+            //the end of the '_thorslave.'. The slaveNumStrEnd is pointed at the end
+            //of a SLAVE_ID.
+            unsigned numberOfThorSlaves = 0;
+            VStringBuffer xPath("*%s*", THORSLAVELOGSEARCHSTR);
+            Owned<IFile> zapFileFolder = createIFile(zapWUFolderPath);
+            Owned<IDirectoryIterator> thorSlaveFiles = zapFileFolder->directoryFiles(xPath);
+            ForEach(*thorSlaveFiles)
+            {
+                const char *pStr = strstr(thorSlaveFiles->query().queryFilename(), THORSLAVELOGSEARCHSTR);
+                pStr += strlen(THORSLAVELOGSEARCHSTR);
+                const char *slaveNumStrEnd = strchr(pStr, '.');
+                if (!slaveNumStrEnd)
+                    continue;
+
+                StringBuffer slaveNumStr;
+                slaveNumStr.append(slaveNumStrEnd - pStr, pStr);
+                unsigned slaveNum = atoi(slaveNumStr);
+                if (slaveNum > numberOfThorSlaves)
+                    numberOfThorSlaves = slaveNum;
+            }
+            return numberOfThorSlaves;
+        }
+        void updateWUQueryAssociatedFilesAttrs(const char *localIP)
+        {
+            StringArray filesNotInZAP;
+            IPropertyTree *queryAssociatedTreeRoot = wuTree->queryPropTree("Query/Associated");
+            Owned<IPropertyTreeIterator> itr = queryAssociatedTreeRoot->getElements("File");
+            ForEach (*itr)
+            {
+                IPropertyTree &fileTree = itr->query();
+                const char *fileName = fileTree.queryProp("@filename");
+
+                StringBuffer fileNameWithPath(unzipDir);
+                fileNameWithPath.append(PATHSEPSTR).append(pathTail(fileName));
+                if (checkFileExists(fileNameWithPath)) //Check if the ZAP report contains this file.
+                {
+                    fileTree.setProp("@ip", localIP);
+                    fileTree.setProp("@filename", fileNameWithPath);
+                }
+                else
+                    filesNotInZAP.append(fileName);
+            }
+            ForEachItemIn(i, filesNotInZAP)
+            {
+                VStringBuffer xPath("File[@filename='%s']", filesNotInZAP.item(i));
+                queryAssociatedTreeRoot->removeTree(queryAssociatedTreeRoot->queryPropTree(xPath));
+            }
+        }
+        void getWUAttributes(IWorkUnit *workunit)
+        {
+            wuid.set(workunit->queryWuid());
+            scope.set(workunit->queryWuScope());
+        }
+        IPTree *queryWUPTree() const
+        {
+            return wuTree;
+        }
+        IPTree *queryGraphProgressPTree() const
+        {
+            return graphProgressTree;
+        }
+        void setUNZIPDir()
+        {   //Set a unique unzip folder inside the component's data folder
+            unzipDir.append(importDir).append(PATHSEPSTR).append(wuid.get());
+        }
+        int unzipZAPReport()
+        {
+            Owned<IFile> unzipFolder = createIFile(unzipDir);
+            if (!unzipFolder->exists())
+                unzipFolder->createDirectory();
+            else
+            {//This should not happen. Just in case.
+                Owned<IDirectoryIterator> iter = unzipFolder->directoryFiles(nullptr, false, false);
+                ForEach(*iter)
+                {
+                    OwnedIFile thisFile = createIFile(iter->query().queryFilename());
+                    if (thisFile->isFile() == foundYes)
+                        thisFile->remove();
+                }
+            }
+
+            CDateTime dt;
+            dt.setNow();
+            dt.getString(unzipDateTime);
+
+            //Unzip ZAP Report
+            StringBuffer zipCommand("unzip ");
+            if (!isEmptyString(zapReportFilePath))
+                zipCommand.append(zapReportFilePath);
+            zipCommand.appendf("%s -d %s", zapReportFileName.get(), unzipDir.str());
+            if (!isEmptyString(zapReportPassword))
+                zipCommand.appendf(" -P %s", zapReportPassword.get());
+            return (system(zipCommand.str()));
+        }
+        void buildPTreesFromZAPReport()
+        {
+            wuTree.setown(readWUXMLFile("ZAPReport_W*.xml", false));
+            fromWUID.set(wuTree->queryName());
+            wuTree->setProp("@scope", scope);
+
+            StringBuffer localIP, unzipDirWithIP("//");
+            IpAddress ipaddr = queryHostIP();
+            ipaddr.getIpText(localIP);
+            unzipDirWithIP.append(localIP).append(unzipDir).append(PATHSEPSTR);
+
+            //update log entries in WU XML;
+            updateWUProcessLogAttrs(unzipDirWithIP);
+
+            //update QueryAssociatedFiles in WU XML;
+            updateWUQueryAssociatedFilesAttrs(localIP);
+
+            graphProgressTree.setown(readWUXMLFile("ZAPReport_*.graphprogress", true));
+        }
+        void updateJobName(IWorkUnit *workunit)
+        {
+            StringBuffer newJobName;
+            const char *jobName = workunit->queryJobName();
+            if (isEmptyString(jobName))
+                newJobName.appendf("IMPORTED from %s", fromWUID.get());
+            else
+                newJobName.appendf("IMPORTED: %s", jobName);
+            workunit->setJobName(newJobName);
+        }
+        void setImportDebugAttribute(IWorkUnit *workunit)
+        {
+            StringBuffer attr;
+            attr.append("FromWUID=").append(fromWUID).append(",");
+            attr.append("ImportDT=").append(unzipDateTime).append(",");
+            attr.append("ZAPReport=").append(zapReportFileName);
+            workunit->setDebugValue("imported", attr, true);
+        }
+        void addProcesses(IWorkUnit *workunit)
+        {
+            ForEachItemIn(j, eclAgentLogEntries)
+            {
+                LogEntry &entry = eclAgentLogEntries.item(j);
+                workunit->addProcess("EclAgent", entry.process, 0, 0, nullptr, true, entry.log);
+            }
+            ForEachItemIn(i, thorLogEntries)
+            {
+                LogEntry &entry = thorLogEntries.item(i);
+                workunit->addProcess("Thor", entry.process, 0, entry.numberOfSlaves, entry.slaveLogPattern, true, entry.log);
+            }
+        }
+    public:
+        CImportWorkUnitHelper(const char *_zapReportFileName, const char *_zapReportFilePath,
+                const char *_zapReportPassword, const char *_importDir, const char *_user)
+            : zapReportFileName(_zapReportFileName), zapReportFilePath(_zapReportFilePath),
+            zapReportPassword(_zapReportPassword), importDir(_importDir), user(_user) { };
+
+        void import(IWorkUnit *workunit)
+        {
+            getWUAttributes(workunit);
+            setUNZIPDir();
+            unzipZAPReport();
+            buildPTreesFromZAPReport();
+            workunit->import(queryWUPTree(), queryGraphProgressPTree());
+            setImportDebugAttribute(workunit);
+            updateJobName(workunit);
+            addProcesses(workunit);
+
+            StringBuffer zapReportToRemove;
+            if (!isEmptyString(zapReportFilePath))
+                zapReportToRemove.append(zapReportFilePath);
+            zapReportToRemove.append(zapReportFileName);
+            Owned<IFile> zapReportFile = createIFile(zapReportToRemove);
+            zapReportFile->remove();
+        }
+    };
+
+    Owned<IWorkUnit> newWU = createWorkUnit(app, user, secMgr, secUser);
+    CImportWorkUnitHelper helper(zapReportFileName, zapReportFilePath, zapReportPassword, importDir, user);
+    helper.import(newWU);
+}
 
 int CWorkUnitFactory::setTracingLevel(int newLevel)
 {
@@ -6126,6 +6479,11 @@ public:
     {
         return baseFactory->restoreWorkUnit(base, wuid, restoreAssociated);
     }
+    virtual void importWorkUnit(const char *zapReportFileName, const char *zapReportFilePath, const char *zapReportPassword,
+        const char *importDir, const char *app, const char *user, ISecManager *secMgr, ISecUser *secUser)
+    {
+        baseFactory->importWorkUnit(zapReportFileName, zapReportFilePath, zapReportPassword, importDir, app, user, secMgr, secUser);
+    }
     virtual IWorkUnit * getGlobalWorkUnit(ISecManager *secMgr, ISecUser *secUser)
     {
         if (!secMgr) secMgr = defaultSecMgr.get();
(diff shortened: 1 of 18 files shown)


wangkx:
@jakesmith could you please take a look?

jakesmith:
@wangkx - so this can be used once merged (I don't think anything currently calls importWorkUnit), can you add a 'importzap' option to wutool that takes a zap file?

common/workunit/workunit.cpp
```diff
@@ -5060,6 +5114,232 @@ bool CWorkUnitFactory::restoreWorkUnit(const char *base, const char *wuid, bool
...
+                {   //legacy WU
+                    setSlaveLogPatternForLegacyWU(logSpec, slaveLogPattern);
+                }
+                proc.setProp("@pattern", slaveLogPattern);
+
+                if (!proc.hasProp("@max"))
+                    proc.setPropInt("@max", findNumberOfThorSlavesFromSlaveLogs(unzipDirWithIP));
+            }
+        }
+        void setSlaveLogPatternForLegacyWU(const char *logSpec, StringBuffer &slaveLogPattern)
```
jakesmith: this is more of a getXX method really:
```
StringBuffer &getSlaveLogPatternForLegacyWU(const char *logSpec, StringBuffer &slaveLogPattern)
```

thorlcr/master/thgraphmanager.cpp
```diff
@@ -327,13 +327,16 @@ void CJobManager::fatal(IException *e)
...
+    StringBuffer log, logUrl, slaveLogPattern;
     logHandler->getLogName(log);
     createUNCFilename(log, logUrl, false);
-    workunit.addProcess("Thor", globals->queryProp("@name"), 0, logUrl.str());
-}
-
 
+    //Not sure whether this is the correct way to set the slaveLogPattern or not:
+    slaveLogPattern.set("thorslave.").append(SLAVEIDSTR);
+    slaveLogPattern.append(strstr(log, "/thormaster.") + strlen("/thormaster"));
```
jakesmith: not ideal burning in "thormaster" here and having to make sure matches in workunit code.
Ideally should be a constant defined somewhere, included and used.
But I'm not sure where is best, not jlib (I would also not put SLAVEIDSTR there).

system/jlib/jlog.hpp
```diff
@@ -1107,6 +1107,8 @@ extern jlib_decl void UseSysLogForOperatorMessages(bool use=true);
 #define SYSLOG querySysLogEventLogger()->log
 #define AUDIT SYSLOG                               // bwd compatibility
 
+#define SLAVEIDSTR "SLAVEID" //Not sure this should be here
```
jakesmith: really need to move this somewhere common to workunit and thor code, but this is too specific to be jlib.
Also true of "thormaster" and similar strings used between components.
There may be a convenient thor include to use somewhere, perhaps thor.hpp.

thorlcr/master/thgraphmanager.cpp
```diff
@@ -327,13 +327,16 @@ void CJobManager::fatal(IException *e)
...
 {
-    StringBuffer log, logUrl;
+    StringBuffer log, logUrl, slaveLogPattern;
     logHandler->getLogName(log);
     createUNCFilename(log, logUrl, false);
-    workunit.addProcess("Thor", globals->queryProp("@name"), 0, logUrl.str());
-}
-
 
+    //Not sure whether this is the correct way to set the slaveLogPattern or not:
```
jakesmith: it may be ok. The proper way is to probably to use the IComponentLogFileCreator interface, which the logHandler came from and use the date + thorslave string to create a filename for the slave log.

common/workunit/workunit.cpp
```diff
@@ -167,6 +167,29 @@ void doDescheduleWorkkunit(char const * wuid)
...
+{
+    //The existing thor master stores the information about a WU into thor master logs
+    //which are named like '...thormaster.yyyy_mm_dd.log'. The datePtr is pointed at
+    //the end of the 'thormaster.'. After that, there is the 'yyyy_mm_dd' part which 
+    //is 10 bytes long. The strlen(datePtr) should >= 10.
+    //If thor master stores all of its log lines for a WU into one log file, there is
+    //no need for the log file name to have the 'yyyy_mm_dd' part. The log name may
+    //look like '...thormaster.log'. The datePtr is pointed at the end of the 'thormaster.'.
+    //The strlen(datePtr) should = 3.
+    //So, if (strlen(datePtr) >= 10), we read the logDate (10 bytes) at the datePtr.
```
jakesmith: am I correct in thinking this method is only used for legacy workunits, that don't have the new @pattern ?

common/workunit/workunit.cpp
```diff
@@ -5060,6 +5114,232 @@ bool CWorkUnitFactory::restoreWorkUnit(const char *base, const char *wuid, bool
...
+                    proc.getProp("@pattern", slaveLogPattern);
+                }
+                else
+                {   //legacy WU
+                    setSlaveLogPatternForLegacyWU(logSpec, slaveLogPattern);
+                }
+                proc.setProp("@pattern", slaveLogPattern);
+
+                if (!proc.hasProp("@max"))
+                    proc.setPropInt("@max", findNumberOfThorSlavesFromSlaveLogs(unzipDirWithIP));
```
jakesmith: could this go under the //legacy WU branch , as it only applies if legacy wuid?

common/workunit/workunit.cpp
```diff
@@ -5060,6 +5114,232 @@ bool CWorkUnitFactory::restoreWorkUnit(const char *base, const char *wuid, bool
...
+        {
+            StringBuffer logDate;
+            getThorLogDateFromLogFileName(logSpec, "/thormaster.", logDate);
+
+            slaveLogPattern.append("thorslave.").append(SLAVEIDSTR);
+            if (!logDate.isEmpty()) //Should not happen
+                slaveLogPattern.append(".").append(logDate);
+            slaveLogPattern.append(".log");
+        }
+        unsigned findNumberOfThorSlavesFromSlaveLogs(const char *zapWUFolderPath)
```
jakesmith: is this just for legacy wuids?
If so, could do with comment to that effect here.

jakesmith (changes requested):

jakesmith:
@wangkx - have done 1st pass, added comments/questions.

common/workunit/workunit.cpp
```diff
@@ -5060,6 +5114,232 @@ bool CWorkUnitFactory::restoreWorkUnit(const char *base, const char *wuid, bool
...
+    helper.getWUAttributes(newWU);
+    helper.setUNZIPDir(importDir);
+    helper.unzipZAPReport(zapReportFilePath, zapReportPassword);
+    helper.buildPTreesFromZAPReport();
+    newWU->import(helper.queryWUPTree(), helper.querygraphProgressPTree());
+    helper.setImportDebugAttribute(newWU);
+
+    Owned<IFile> zapReportFile = createIFile(zapReportFileName);
+    zapReportFile->remove();
+    return newWU.getClear();
```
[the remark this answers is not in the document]
wangkx: At this moment, it is not used. What I am doing is to similate the CWorkUnitFactory::createWorkUnit() which returns IWorkUnit*.

common/workunit/workunit.cpp
```diff
@@ -6200,6 +6485,51 @@ class CStringPTreeAttrIterator : implements IStringIterator, public CInterface
...
+    unsigned numberOfThorSlaves = 0;
+    bool singleLog = false;
+
+public:
+    CWUThorLogInfo(const char *_logName, const char *_processName, const char *_groupName,
+        const char *_logDate, const char *_pattern, unsigned _numberOfThorSlaves, bool _singleLog)
+        : logName(_logName), processName(_processName), groupName(_groupName), logDate(_logDate),
+          pattern(_pattern), numberOfThorSlaves(_numberOfThorSlaves), singleLog(_singleLog) {};
+
+    virtual const char *getLogName() const
```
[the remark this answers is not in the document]
wangkx: Will do.

esp/bindings/http/platform/httptransport.cpp
```diff
@@ -1957,19 +1957,18 @@ bool CHttpRequest::readUploadFileName(CMimeMultiPart* mimemultipart, StringBuffe
...
-        pStr = strchr(str, '/');
-    if (pStr)
-    {
-        pStr[0] = 0;
-        fileName.clear().append(str).reverse();
-    }
-    tmpFileName.appendf("%s/%s.part", filePath, fileName.str());
+    StringBuffer tmpFileName(filePath);
+    if (tmpFileName.charAt(tmpFileName.length() - 1) != PATHSEPCHAR)
+        addPathSepChar(tmpFileName);
```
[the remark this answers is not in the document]
wangkx: Will remove line 1963

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -1102,44 +1076,32 @@ unsigned WsWuInfo::getWorkunitThorLogInfo(IArrayOf<IEspECLHelpFile>& helpers, IE
...
-                thorLogList.append(*thorLog.getLink());
+            Owned<IEspThorLogInfo> thorLog = createThorLogInfo();
+            thorLog->setProcessName(logInfo.getProcessName());
+            if (version < 1.78)
+            {
+                thorLog->setClusterGroup(logInfo.getGroupName());
+                thorLog->setLogDate(logInfo.getLogDate());
             }
+            thorLog->setNumberSlaves(logInfo.getNumberOfThorSlaves());
+            thorLogList.append(*thorLog.getLink());
```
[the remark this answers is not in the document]
wangkx: Will do.

common/workunit/workunit.cpp
wangkx: Will do.

thorlcr/master/thgraphmanager.cpp
wangkx: How about '#define MASTERLOGSEARCHSTR "/thormaster"' in thorhelper.hpp?

system/jlib/jlog.hpp
wangkx: How about move it to thorhelper.hpp?

thorlcr/master/thgraphmanager.cpp
wangkx: I guess that many changes is needed to use the IComponentLogFileCreator interface. How about leave it as is for now?

common/workunit/workunit.cpp
wangkx: Will do.

common/workunit/workunit.cpp
wangkx: Will add a comment.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -1056,44 +1056,18 @@ unsigned WsWuInfo::getWorkunitThorLogInfo(IArrayOf<IEspECLHelpFile>& helpers, IE
...
-            IWARNLOG("Cannot find TargetClusterInfo for workunit %s", cw->queryWuid());
-            return countThorLog;
-        }
-
-        unsigned numberOfSlaveLogs = clusterInfo->getNumberOfSlaveLogs();
+        IArrayOf<IConstWUThorLogInfo> wuThorLogs;
+        if (version < 1.78)
+            cw->getWUThorLogInfo(wuThorLogs);
+        else
+            cw->getWUThorLogInfoLW(wuThorLogs);
```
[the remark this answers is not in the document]
wangkx: The code here is to call getWUThorLogInfo() for old ESP client version (< 1.78). The getWUThorLogInfo() returns groupName and logDate. The getWUThorLogInfoLW() does not. ESP tries not to change old behaviour in case of breaking old ESP client. 

I may encapsulate the logic into one IConstWorkUnit call, like getWUThorLogInfo(bool returnGroupName, bool returnLogDate, ...) . Please comment.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -1102,44 +1076,32 @@ unsigned WsWuInfo::getWorkunitThorLogInfo(IArrayOf<IEspECLHelpFile>& helpers, IE
...
-                thorLog->setClusterGroup(groupName.str());
-                thorLog->setLogDate(logDate.str());
-                thorLog->setNumberSlaves(numberOfSlaveLogs);
-                thorLogList.append(*thorLog.getLink());
+            Owned<IEspThorLogInfo> thorLog = createThorLogInfo();
+            thorLog->setProcessName(logInfo.getProcessName());
+            if (version < 1.78)
+            {
+                thorLog->setClusterGroup(logInfo.getGroupName());
+                thorLog->setLogDate(logInfo.getLogDate());
```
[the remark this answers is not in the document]
wangkx: The logDate is correct. The groupName may not be correct for imported WU. I may skip that if Debug Value "imported" is set. Please comment.

common/workunit/workunit.cpp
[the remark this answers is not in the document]
jakesmith: ok, with create it makes sense - because you are about to fill in details.
Not sure it does when importing from a zap though.
I'd remove for now,
and add ability to use/test this to wutool.

wangkx:
I can add an 'importzap' option to wutool after this PR. The wutool has to upload a ZAP file (copy a file to ESP working dir) and 
calls a WsWorkunits method (small code changes in WsWorkunits also) to import the ZAP file.

common/workunit/workunit.cpp
```diff
@@ -10013,6 +10416,93 @@ IWUGraphStats *CLocalWorkUnit::updateStats(const char *graphName, StatisticCreat
     return new CWuGraphStats(LINK(p), creatorType, creator, _wfid, graphName, subgraph);
 }
 
+bool CLocalWorkUnit::usingDedicatedLogFiles() const
+{
+    if (hasDebugValue("imported"))
```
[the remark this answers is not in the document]
jakesmith: ok, what will currently happen (as this PR stands) with those other places, if the workunit originally came from a zap report?

esp/services/ws_workunits/ws_workunitsHelpers.cpp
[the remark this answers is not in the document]
jakesmith: I think what's confusing is that the getWUThorLogInfoLW vs getWUThorLogInfo conditional code - there should be 1 interface method, it can fill-in/deduce more than is required by newer clients.

This code may be fine though, simply using some fields that may be present from the IConstWUThorLogInfo instances.

thorlcr/master/thgraphmanager.cpp
jakesmith: yes that may be fine.

system/jlib/jlog.hpp
jakesmith: yep, that's fine.

thorlcr/master/thgraphmanager.cpp
jakesmith: yes, I'd leave this as is for now.

common/workunit/workunit.cpp
jakesmith: I think that logic should be rolled into getWUThorLogInfo, i.e. it should deduce by whatever info. available regardless of whether imported or non-imported, and should make the log date available from IConstWUThorLogInfo.

jakesmith:
>I can add an 'importzap' option to wutool after this PR. The wutool has to upload a ZAP file (copy a file to ESP working dir) and
calls a WsWorkunits method (small code changes in WsWorkunits also) to import the ZAP file.

Why?
Code should be oranized so anyone can call IWorkUnitFactory::importWorkunit with the zap report directly. Not just via esp.

wangkx:
I misunderstood your 'importzap' option. I thought that a wutool calls an esp WsWorkunits method which calls the importWorkunit(). You are correct. A wutool may call the importWorkunit() directly.

common/workunit/workunit.cpp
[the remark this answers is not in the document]
wangkx: If the workunit originally came from a zap report, its XML has the Debug  Value "imported". Both thor log and an eclagent log can be retrieved by a log file name. To use @singlelog, the thor or eclagent process name has to be passed around.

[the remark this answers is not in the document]
jakesmith: in fact it's not clear why usingDedicatedLogFiles() (or Debug/@imported) is needed at all.
It is not called anywhere afaics.

Now that the individual Process/ entries direct any code that reads the logs whether it is @singleLog or not.

However, I see code that doesn't call usingDedicatedLogFiles() but calls hasDebugValue("imported") directly, i.e.:
```
    if (hasDebugValue("imported")) //WU imported from a ZAP report
        thorMasterLogDateSearchString.set("_thormaster.");
    else
        thorMasterLogDateSearchString.set("/thormaster.");
```

that should be able to deduce this from the IConstWUThorLogInfo information, without need for specifically looking for  "improted".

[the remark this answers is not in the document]
wangkx: Let me confirm: 1. the singlelog attribute inside the Process/ entries (not the "imported") should be used to decide single log file or not.  2. the IConstWUThorLogInfo information (the pattern attribute) should be used to build the thorMasterLogDateSearchString without checking the "imported".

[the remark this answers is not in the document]
jakesmith: yes I think so, can you foresee a problem with that approach?

That way there is no special handling/conditinal code based on whether it is imported, only differential handling (handled by the IConstWUThorLogInfo impl.) based on whether single log or not.
Which we will want anyway at some point even for standard (non imported) workunits.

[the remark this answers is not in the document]
wangkx: I do not foresee any problem for the 1st item (just some code changes as I mentioned before). For the second item,  instead of using the thorMasterLogDateSearchString, I may use the pathTail on the log attribute (not the pattern attribute) from the IConstWUThorLogInfo information.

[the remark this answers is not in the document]
jakesmith: > I may use the pathTail on the log attribute (not the pattern attribute) from the IConstWUThorLogInfo information.

yes, could be done within CWUThorLogInfo ctor?

common/workunit/workunit.cpp
```diff
@@ -8427,9 +8758,81 @@ void CLocalWorkUnit::addProcess(const char *type, const char *instance, unsigned
...
+        else
+        {
+            //legacy wuid: an IP address inside the name of thor slave.
+            //ex. thorslave.10.239.219.6_20100.2012_05_23.log
+            logFileName.appendf("thorslave.%s*", ipAddress);
+        }
+
+        //Now, add the 'yyyy_mm_dd' (10 bytes) into the logFileName.
+        StringBuffer logDate;
+        getThorLogDateFromLogFileName(logSpec, "/thormaster.", logDate);
```
jakesmith: related to other discussion:
feels like getSlaveLogFileName should getting a IConstWUThorLogInfo for the correct process,
then asking it for a thor slave log given a slave number.

common/workunit/workunit.cpp
```diff
@@ -5060,6 +5114,232 @@ bool CWorkUnitFactory::restoreWorkUnit(const char *base, const char *wuid, bool
...
+
+                StringBuffer slaveLogPattern;
+                slaveLogPattern.set(process).append("_");
+                if (proc.hasProp("@pattern"))
+                {
+                    proc.getProp("@pattern", slaveLogPattern);
+                }
+                else
+                {   //legacy WU
+                    setSlaveLogPatternForLegacyWU(logSpec, slaveLogPattern);
```
jakesmith: this is readlly a getXX method

common/workunit/workunit.cpp
wangkx: After my new changes, this function (getThorLogDateFromLogFileName()) will be called by 2 palces: 1. getWUThorLogInfo(); 2. getSlaveLogPatternForLegacyWU() when importing a ZAP.

common/workunit/workunit.cpp
wangkx: Will do.

common/workunit/workunit.cpp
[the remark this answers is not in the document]
wangkx: I may enhance the IConstWUThorLogInfo  to IConstWUProcessLogInfo because similar functions may be used for eclagent log.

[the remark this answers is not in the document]
jakesmith: That sounds sensible, hopefully the outcome will come simpler.

wangkx:
@AttilaVamos could you please check the Smoketest error? This PR does not change anything related to regex.

AttilaVamos:
@wangkx Please resolve the conflicting file and push a new commit.  
I think the previous regex related failure didn't related to this PR.

wangkx:
@jakesmith please review

common/workunit/workunit.cpp
```diff
@@ -5044,6 +5104,237 @@ bool CWorkUnitFactory::restoreWorkUnit(const char *base, const char *wuid, bool
...
+                if (proc.hasProp("@pattern"))
+                {
+                    proc.getProp("@pattern", slaveLogPattern);
+                }
+                else
+                {   //legacy WU
+                    getSlaveLogPatternForLegacyWU(logSpec, slaveLogPattern);
+                    proc.setPropInt("@max", findNumberOfThorSlavesFromSlaveLogs(unzipDirWithIP));
+                }
+                proc.setProp("@pattern", slaveLogPattern);
```
jakesmith, line 5180: Re. related to comment on line 5162.

I would expect you to be able to clear the existing entries, and call workunit.addProcess() to be called for each new log being imported.

common/workunit/workunit.cpp
```diff
@@ -4049,6 +4095,20 @@ class CLockedWorkUnit : implements ILocalWorkUnit, implements IExtendedWUInterfa
             { return c->getAbortBy(str); }
     virtual unsigned __int64 getAbortTimeStamp() const
             { return c->getAbortTimeStamp(); }
+    virtual StringBuffer & getProcessLogPattern(const char * type, const char *instance, StringBuffer & pattern) const
+            { return c->getProcessLogPattern(type, instance, pattern); }
+    virtual void setProcessLogPattern(const char * type, const char *instance, const char * pattern)
+            { return c->setProcessLogPattern(type, instance, pattern); }
```
jakesmith, line 4101: it looks like neither of these are actually used?
I think the log @pattern - is (and should be) only for internal use, so there shouldn't be exposed interface methods.

[not sure if that also applies to some of the other methods below]

jakesmith (changes requested):

jakesmith:
@wangkx - have mostly gone through it and I think getting closer.
But have some concerns (see comments/questions).

wangkx:
@jakesmith I made the changes based on our discussions. Please review. 

In new updateWUProcessLogAttrs(), I still call both getProp("@pattern", patternString) and setProp("@pattern", patternString). The patternString has an initial value before the getProp is called.

The https://github.com/hpcc-systems/HPCC-Platform/pull/13258#discussion_r401584303 is related to an existing issue. When a job runs overnight, there are >1 thor log items in the Process. The existing code always returns the first one. To fix the issue, ECLWatch UI has to accept a date input and ESP code has to pass the date to this getSlaveLogFileName(). Should I create a separate JIRA?

jakesmith:
jirabot doesn't have this PR attributed to anyone, well "pattern":
```
HPCC | #13258 | master | HPCC-22136 Import WU from ZAP report | pattern | wangkx | 89 | 3
-- | -- | -- | -- | -- | -- | -- | --
```

@wangkx - was this ready for review? If so, can you resolve the conflict and I'll take a look.

wangkx:
It was ready for review. You were not notified due to the '@pattern'. Today, I removed the @ . But, the conflict occurs. I will resolve it.

wangkx:
@jakesmith this is ready for review

common/workunit/workunit.cpp
```diff
@@ -5044,6 +5100,262 @@ bool CWorkUnitFactory::restoreWorkUnit(const char *base, const char *wuid, bool
...
+            wuTree->setProp("@distributedAccessToken", distributedAccessToken.str());
+
+            StringBuffer localIP, unzipDirWithIP("//");
+            IpAddress ipaddr = queryHostIP();
+            ipaddr.getIpText(localIP);
+            unzipDirWithIP.append(localIP).append(unzipDir).append(PATHSEPSTR);
+
+            //update log entries in WU XML;
+            updateWUProcessLogAttrs(unzipDirWithIP, false);
+            updateWUProcessLogAttrs(unzipDirWithIP, true);
```
jakesmith, line 5320: @wangkx - don't think this is quite right, based on our discussion, the intention was to clear the existing Process entries, then use standard methods, i.e. IWorkUnit::addProcess(), to add in the new entries, rather than trying to patch up the existing IPT.
Previous comment: https://github.com/hpcc-systems/HPCC-Platform/pull/13258#discussion_r401600423

wangkx: From our discussion, I may pass 0 as PID when calling addProcess(). The Process Instance Name can be parsed from the log file name. Number of thor slaves? For new WU, the existing process entry is the place to store the number of slaves. Should I read from there? For legacy WU, I guess I have to use the findNumberOfThorSlavesFromSlaveLogs().

common/workunit/workunit.cpp
```diff
@@ -9397,6 +9827,112 @@ IWUGraphStats *CLocalWorkUnit::updateStats(const char *graphName, StatisticCreat
...
+        const char *processName = proc.queryName();
+        if (isEmptyString(processName) || logSpec.isEmpty())
+            continue;
+
+        unsigned numberOfThorSlaves = 0, processID = 0;
+        StringBuffer groupName;
+        if (strieq(processType, "Thor"))
+        {
+            getClusterThorGroupName(groupName, processName);
+            numberOfThorSlaves = getNumberOfThorSlaves(processName);
```
jakesmith, line 9919: getProcesses() is called (line 9903) and then you are iterating through them.
getNumberOfThorSlaves also calls getProcesses() but with the processName it got from the instance the iterator is on, so it is re-iterating to find a match of something it already has afaics?

shouldn't this pickup @max if present... and use the legacy code if not.
And since this is the only place getNumberOfThorSlaves is called afaics, then there's no point in getNumberOfThorSlaves calling getProceses().

And since getNumberOfThorSlaves isn't used other than in this context (but IConstWUProcessLogInfo::getNumberOfThorSlaves is), it would be better remove this method.

common/workunit/workunit.hpp
```diff
@@ -1257,6 +1273,10 @@ interface IConstWorkUnit : extends IConstWorkUnitInfo
     virtual void clearGraphProgress() const = 0;
     virtual IStringVal & getAbortBy(IStringVal & str) const = 0;
     virtual unsigned __int64 getAbortTimeStamp() const = 0;
+    virtual StringBuffer & getSlaveLogFileName(const char *thorProcess, int slaveNum, const char *ipAddress, StringBuffer &logFileName, bool withPath) const = 0;
+    virtual bool usingDedicatedLogFiles(const char* processType, const char* processName, const char* logSpec) const = 0;
+    virtual unsigned getNumberOfThorSlaves(const char * processName) const = 0;
```
jakesmith, line 1278: see other comment, only called in one context, and it is where it is already iterating through processes.
Better to remove.

common/workunit/workunit.cpp
```diff
@@ -9397,6 +9827,112 @@ IWUGraphStats *CLocalWorkUnit::updateStats(const char *graphName, StatisticCreat
...
+bool CLocalWorkUnit::usingDedicatedLogFiles(const char *processType, const char *processName, const char *logSpec) const
+{
+    if (isEmptyString(processType) || (isEmptyString(processName) && isEmptyString(logSpec)))
+        return false;
+
+    IArrayOf<IConstWUProcessLogInfo> processLogs;
+    if (!isEmptyString(processName))
+        getWUProcessLogInfo(processType, processName, processLogs);
+    else
+    {
```
jakesmith, line 9839: When does this apply?
i.e. when is usingDedicatedLogFiles() called without a processName ?

should add a comment to explain.

common/workunit/workunit.cpp
```diff
@@ -9397,6 +9827,112 @@ IWUGraphStats *CLocalWorkUnit::updateStats(const char *graphName, StatisticCreat
...
+
+unsigned CLocalWorkUnit::getNumberOfThorSlaves(const char *processName) const
+{
+    Owned<IPropertyTreeIterator> procs = getProcesses("Thor", processName);
+    if (procs->first())
+    {
+        unsigned numberOfThorSlaves = procs->query().getPropInt("@max", 0);
+        if (numberOfThorSlaves > 0)
+            return numberOfThorSlaves;
+    }
```
jakesmith, line 9880: is it correct to continue here, if there is no the request "Thor" processName is not part of the workunit?

jakesmith (changes requested):

jakesmith:
@wang - added some comments/questions.

common/workunit/workunit.cpp
```diff
@@ -5145,12 +5156,8 @@ void CWorkUnitFactory::importWorkUnit(const char *zapReportFileName, const char
...
         void updateWUProcessLogAttrs(const char *unzipDirWithIP, bool thorCluster)
         {
-            const char *processBranchName = thorCluster ? "Thor" : "EclAgent";
             IPropertyTree *processTreeRoot = wuTree->queryPropTree("Process");
-            IPropertyTree *oldProcessBranch = processTreeRoot->queryPropTree(processBranchName);
-            Owned<IPropertyTree> newProcessBranch = createPTree(processBranchName);
-
-            bool processLogFound = false;
+            IPropertyTree *oldProcessBranch = processTreeRoot->queryPropTree(thorCluster ? "Thor" : "EclAgent");
             VStringBuffer mask("*_%s*", thorCluster ? THORMASTERLOGSEARCHSTR : ECLAGENTLOGSEARCHSTR);
```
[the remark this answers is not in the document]
wangkx: What if there are many slave logs which should be skipped?

common/workunit/workunit.cpp
```diff
@@ -5307,6 +5302,7 @@ void CWorkUnitFactory::importWorkUnit(const char *zapReportFileName, const char
         void buildPTreesFromZAPReport()
         {
             readWUXMLFileToPTree(wuTree, "ZAPReport_W*.xml", false);
+            fromWUID.set(wuTree->queryName());
             wuTree->setProp("@scope", scope);
             wuTree->setProp("@distributedAccessToken", distributedAccessToken.str());
```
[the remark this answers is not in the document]
wangkx: It was copied from the CWorkUnitFactory::createNamedWorkUnit(). Not sure whether it is needed here or not. I can remove it.

common/workunit/workunit.cpp
```diff
@@ -5044,6 +5100,269 @@ bool CWorkUnitFactory::restoreWorkUnit(const char *base, const char *wuid, bool
...
+                throw MakeStringException(WUERR_InvalidUserInput, "Failed to found %s in %s.", mask, unzipDir.str());
+            }
+
+            fileName.set(unzipDir).append(PATHSEPSTR);
+            di->getName(fileName);
+            if (di->next())
+                throw MakeStringException(WUERR_InvalidUserInput, "More than 1 %s files found in %s.", mask, unzipDir.str());
+            return true;
+        }
+        void readWUXMLFileToPTree(Owned<IPTree> &tree, const char *pattern, bool optional)
```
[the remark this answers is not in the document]
wangkx: Ok

common/workunit/workunit.cpp
```diff
@@ -5044,6 +5100,269 @@ bool CWorkUnitFactory::restoreWorkUnit(const char *base, const char *wuid, bool
...
+            if (!isEmptyString(zapReportFilePath))
+                zipCommand.append(zapReportFilePath);
+            zipCommand.appendf("%s -d %s", zapReportFileName.get(), unzipDir.str());
+            if (!isEmptyString(zapReportPassword))
+                zipCommand.appendf(" -P %s", zapReportPassword.get());
+            return (system(zipCommand.str()));
+        }
+        void buildPTreesFromZAPReport()
+        {
+            readWUXMLFileToPTree(wuTree, "ZAPReport_W*.xml", false);
```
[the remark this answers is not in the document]
wangkx: wuTree.setown(readWUXMLFile("ZAPReport_W*.xml", false));

common/workunit/workunit.cpp
```diff
@@ -5044,6 +5100,269 @@ bool CWorkUnitFactory::restoreWorkUnit(const char *base, const char *wuid, bool
...
+            {
+                const char* file = files->query().queryFilename();
+                StringBuffer name(pathTail(file)); //ProcessName_thormaster.Date or ProcessName_eclagent.Date
+                const char *ptr = strstr(name, thorCluster ? THORMASTERLOGSEARCHSTR : ECLAGENTLOGSEARCHSTR) - 1;
+                name.setLength(ptr - name.str()); //ProcessName
+
+                if (!thorCluster)
+                {
+                    eclAgentLogEntry.setown(new LogEntry(file, name));
+                    break; //We should only have 1 eclagent log.
```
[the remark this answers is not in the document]
wangkx: Giving it more thought for the WU continued over to the next day, ZAP report should be able to collect the information from multiple logs into one log file.

common/workunit/workunit.cpp
```diff
@@ -5044,6 +5100,269 @@ bool CWorkUnitFactory::restoreWorkUnit(const char *base, const char *wuid, bool
...
+            {
+                IPropertyTree &fileTree = itr->query();
+
+                StringBuffer fileNameWithPath(unzipDir);
+                fileNameWithPath.append(PATHSEPSTR).append(pathTail(fileTree.queryProp("@filename")));
+                if (checkFileExists(fileNameWithPath)) //Check if the ZAP report contains this file.
+                {
+                    fileTree.setProp("@ip", localIP);
+                    fileTree.setProp("@filename", fileNameWithPath);
+                }
```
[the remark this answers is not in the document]
wangkx: I should remove the file tree from the Query/Associated/File.

common/workunit/workunit.cpp
```diff
@@ -5044,6 +5100,269 @@ bool CWorkUnitFactory::restoreWorkUnit(const char *base, const char *wuid, bool
...
+        {
+            wuid.set(workunit->queryWuid());
+            scope.set(workunit->queryWuScope());
+            workunit->getWorkunitDistributedAccessToken(distributedAccessToken);
+        }
+        IPTree *queryWUPTree()
+        {
+            return wuTree;
+        }
+        IPTree *querygraphProgressPTree()
```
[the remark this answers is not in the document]
wangkx: Will change

common/workunit/workunit.cpp
wangkx: Good catch.

common/workunit/workunit.hpp
wangkx: Will remove.

common/workunit/workunit.cpp
wangkx: The existing ECLWatch retrieves thor master  log using log file name (with full path and no processName). I will add a comment.

common/workunit/workunit.cpp
wangkx: Could this happen when the addProcess() has not be called? This is in getNumberOfThorSlaves() which will be removed.

wangkx:
@jakesmith please review

common/workunit/workunit.cpp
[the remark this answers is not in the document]
jakesmith: should they?
@wangkx - what happens to the slave logs as it stands, how do you access them after you've imported a zap report?

common/workunit/workunit.cpp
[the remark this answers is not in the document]
jakesmith: it **may** make sense if it was in the same environment, but it's only used at runtime I think, and won't make any sense when importing the zap into any other environment.

I would remove it, it more likely wrong than right.

common/workunit/workunit.cpp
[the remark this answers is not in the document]
jakesmith: ok, fair enough - but if you were iterating through log files once, you wouldn't' need to special case it, and could add all log entries even if >1.

jakesmith:
@wangkx - a few last questions.

common/workunit/workunit.cpp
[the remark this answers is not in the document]
wangkx: I can access the slave logs using the information (slaveLogPattern, numberOf Slaves, logSpec) in Process/Thor/ProcessName (the entry has the information for both master and slaves).

common/workunit/workunit.cpp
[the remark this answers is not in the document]
wangkx: Removed

common/workunit/workunit.cpp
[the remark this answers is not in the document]
wangkx: The code has been changed to iterate through log files once.

common/workunit/workunit.cpp
```diff
@@ -5153,48 +5153,57 @@ void CWorkUnitFactory::importWorkUnit(const char *zapReportFileName, const char
...
-            IPropertyTree *oldProcessBranch = processTreeRoot->queryPropTree(thorCluster ? "Thor" : "EclAgent");
-            VStringBuffer mask("*_%s*", thorCluster ? THORMASTERLOGSEARCHSTR : ECLAGENTLOGSEARCHSTR);
+            IPropertyTree *oldThorProcessBranch = processTreeRoot->queryPropTree("Thor");
+            IPropertyTree *oldEclAgentProcessBranch = processTreeRoot->queryPropTree("EclAgent");
+
+            VStringBuffer eclAgentLogSearchStr("_%s", ECLAGENTLOGSEARCHSTR);
+            VStringBuffer thorMasterLogSearchStr("_%s", THORMASTERLOGSEARCHSTR);
             Owned<IFile> unzipDir = createIFile(unzipDirWithIP);
-            Owned<IDirectoryIterator> files = unzipDir->directoryFiles(mask);
+            Owned<IDirectoryIterator> files = unzipDir->directoryFiles("*");
```
jakesmith, line 5165: should this be *.log?

wangkx: Sorry, it should be. Will fix.

jakesmith:
@wangkx - could you also go ahead and squash and rebase onto latest candidate-7.8.x ?

wangkx:
OK, I will.

wangkx:
@jakesmith The PR is squashed and rebased onto latest candidate-7.8.x.

common/workunit/workunit.cpp
```diff
@@ -5113,6 +5167,288 @@ bool CWorkUnitFactory::restoreWorkUnit(const char *base, const char *wuid, bool
...
+        void getWUAttributes(IWorkUnit *workunit)
+        {
+            wuid.set(workunit->queryWuid());
+            scope.set(workunit->queryWuScope());
+        }
+        IPTree *queryWUPTree()
+        {
+            return wuTree;
+        }
+        IPTree *queryGraphProgressPTree()
```
jakesmith, line 5350: minor: should this be a const method? and should it return const IPTree ?

anywhere that declares a IPTree/IPropertyTree (or other reference/pointer) where it is expected not to change, should be declared so really..

common/workunit/workunit.cpp
```diff
@@ -7808,9 +8193,43 @@ void CLocalWorkUnit::addProcess(const char *type, const char *instance, unsigned
...
+
+StringBuffer &CLocalWorkUnit::getSlaveLogFileName(const char *thorProcess, int slaveNum,
+    const char *ipAddress, StringBuffer &logFileName, bool withPath) const
+{
+    IArrayOf<IConstWUProcessLogInfo> wuThorLogs;
+    getWUProcessLogInfo("Thor", thorProcess, wuThorLogs);
+    if (wuThorLogs.ordinality() == 0)
+        return logFileName;
+
+    return wuThorLogs.item(0).getSlaveLogFileName(slaveNum, ipAddress, logFileName, withPath);
```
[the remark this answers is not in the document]
jakesmith: I think this is ok, because there cannot be >1 wuThorLogs returned from getWUProcessLogInfo, correct?
Should add comment and assert that though.

common/workunit/workunit.cpp
```diff
@@ -9476,6 +9923,103 @@ IWUGraphStats *CLocalWorkUnit::updateStats(const char *graphName, StatisticCreat
...
+        }
+        else
+            return false; //Should not happen
+
+        getWUProcessLogInfo(processType, process, processLogs);
+    }
+    if (processLogs.ordinality() == 0)
+        return false;
+
+    return processLogs.item(0).getSingleLog();
```
jakesmith, line 9967: I think similar to previous comment, that processLogs.ordinality() should never be >1.
Should assert that's true and add comment though.

common/workunit/workunit.hpp
```diff
@@ -65,6 +65,8 @@ typedef unsigned short UChar;
 #define QUERRREG_UNSUSPEND          QUERYREGISTRY_ERROR_START+9
 #define QUERRREG_COMMENT            QUERYREGISTRY_ERROR_START+10
 
+#define ECLAGENTLOGSEARCHSTR        "eclagent."
```
jakesmith, line 68: only internal, only used by implementation in workunit.cpp?

esp/services/ws_workunits/ws_workunitsService.cpp
```diff
@@ -59,6 +59,7 @@
 #endif
 
 #define ESP_WORKUNIT_DIR "workunits/"
+const char* zipFolder = "tempzipfiles" PATHSEPSTR;
```
jakesmith, line 62: trivial: should be static really, and preferrable for const like this to be a constexpr, so:
```
static constexpr const char* zipFolder = ..
```

esp/services/ws_workunits/ws_workunitsService.hpp
```diff
@@ -203,6 +203,8 @@ class CWsWorkunitsEx : public CWsWorkunits
     void getGraphsByQueryId(const char *target, const char *queryId, const char *graphName, const char *subGraphId, IArrayOf<IEspECLGraphEx>& ECLGraphs);
     void checkAndSetClusterQueryState(IEspContext &context, const char* cluster, const char* querySetId, IArrayOf<IEspQuerySetQuery>& queries, bool checkAllNodes);
     void checkAndSetClusterQueryState(IEspContext &context, const char* cluster, StringArray& querySetIds, IArrayOf<IEspQuerySetQuery>& queries, bool checkAllNodes);
+    IWorkUnitFactory *queryWUFactory() { return wuFactory; };
+    const char *getDataDirectory() { return dataDirectory.str(); };
```
jakesmith, line 207: trivial: could be const method.

jakesmith:
@wangkx - I think this is close, added a couple of minor/trivial comments.

I also did some testing, and it looks good but noticed a few small issues:

1. It left the uploaded zap report in tempzipfiles. I presume that is not intentional.
2. Once imported it's not easy to tell which workunit is the one you just imported.
To make it easier, I think it could be worth prepending "IMPORTED: " to the job name, so it's visialbe and you can tell when viewing the workunit list.
4. Seeing an odd artefact related to Number of slaves (see Teams IM).
5. Not directly related to this - wouldn't it make more sense to have slave log helper access under Helpers ?
6. Is there a linked JIRA for the Eclwatch UI changes that are needed to access this feature?

common/workunit/workunit.cpp
wangkx: In fact, if the WU is not imported and run overnight, the processLogs.ordinality() will > 1. Could the SingleLog flags be different from one processLog to another processLog?

common/workunit/workunit.hpp
wangkx: For now, yes.

common/workunit/workunit.cpp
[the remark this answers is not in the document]
wangkx: As I commented days ago, this part of the code may need to be revisited. The existing code of reading thor logs does not support a WU running overnight. This function may need to return a StringArray (SlaveLogFileNames). I will add some comment.

wangkx:
> It left the uploaded zap report in tempzipfiles. I presume that is not intentional.
Will check and fix.
> prepending "IMPORTED: " to the job name
Good idea.
> Not directly related to this - wouldn't it make more sense to have slave log helper access under Helpers ?
For a 400 node thor, not sure we want to show so many items under the Helpers.
> Is there a linked JIRA for the Eclwatch UI changes that are needed to access this feature?
I will create one.

jakesmith:
>For a 400 node thor, not sure we want to show so many items under the Helpers.

I did not mean list all, I meant that something similar to the existing mechanism, but moved under Helpers.

wangkx:
'''moved under Helpers.'''
Will ask for a UI change.

common/workunit/workunit.cpp
wangkx: It is fine to be a const method. But, returning const IPTree does not work for the LINK(graphProgressTree) inside the import(IPropertyTree *wuTree, IPropertyTree *graphProgressTree).

jakesmith:
>Seeing an odd artefact related to Number of slaves (see Teams IM).

@wangkx - see MSTeams IM's/screenshot re. this issue.

wangkx:
```Seeing an odd artefact related to Number of slaves (see Teams IM).```
Finally, I think I reproduced the problem. https://track.hpccsystems.com/browse/HPCC-23904. Sound likes a UI issue.

**Comment by wangkx** [comment]
@jakesmith please review

**Comment by jakesmith** [comment]
@wangkx - I think this is ready, but probably should not go in a 7.8.x point build now.
Can you link the new associated UI JIRA (HPCC-23903) to this one,
and can you squash and rebase to master please?

**Comment by wangkx** [comment]
@jakesmith this PR has been squashed and rebased to master. The UI JIRA (HPCC-23903) has been closed as a duplicate of https://track.hpccsystems.com/browse/HPCC-22137. The HPCC-22137 is under the same parent task as the JIRA of this PR.

File: common/workunit/workunit.hpp (hunk 35 above)
Comment (jakesmith) [reply common/workunit/workunit.hpp:68]: ok, I think should remain an implementation detail only, and so should really be moved out of header into implementation.

File: esp/services/ws_workunits/ws_workunitsService.cpp [hunk 38]
```diff
@@ -4472,6 +4475,48 @@ int CWsWorkunitsSoapBindingEx::onGetForm(IEspContext &context, CHttpRequest* req
...
+        if (strieq(method, "ImportWUZAPFile"))
+        {
+            SecAccessFlags accessOwn, accessOthers;
+            getUserWuAccessFlags(ctx, accessOwn, accessOthers, false);
+            if ((accessOwn != SecAccess_Full) || (accessOthers != SecAccess_Full))
+                throw MakeStringException(-1, "Permission denied.");
+    
+            StringBuffer password, importQueryAssociatedFiles;
+            request->getParameter("importQueryAssociatedFiles", importQueryAssociatedFiles);
+            request->getParameter("Password", password);
```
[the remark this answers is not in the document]
jakesmith: looks like this one was missed?

jakesmith:
@wangkx - could merge as is, but 2 minor comments

common/workunit/workunit.hpp
wangkx: I will move this line to workunit.cpp.

esp/services/ws_workunits/ws_workunitsService.cpp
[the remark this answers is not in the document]
wangkx: @jakesmith the importQueryAssociatedFiles has been removed since it is never used. The 'Password' is capitalized because it is a part of the ESP request. It is read into the 'password' which is a c++ variable and is camelCase.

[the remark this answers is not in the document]
jakesmith: @wangkx - ok understood.

wangkx:
@jakesmith I just moved the 'deine ECLAGENTLOGSEARCHSTR' from workunit.hpp to workunit.cpp.

jakesmith:

richardkchapman:
@ghalliday Please sanity check

wangkx:
I am closing this PR because it will be broken into separate PR's.

PR #13271, opened 2020-01-10
HPCC-23284 Prevent final remote copy stalling progress/watchdog

jakesmith:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/thorlcr/activities/thdiskbaseslave.cpp
+++ b/thorlcr/activities/thdiskbaseslave.cpp
@@ -430,10 +430,13 @@ void CDiskWriteSlaveActivityBase::close()
             outraw.clear();
         }
 
+        Owned<IFileIO> tmpFileIO;
         {
             CriticalBlock block(statsCs);
             mergeStats(fileStats, outputIO);
-            outputIO.clear();
+
+            // ensure it is released/destroyed after releasing crit, since the IFileIO might involve a final copy and take considerable time.
+            tmpFileIO.setown(outputIO.getClear());
         }
 
         if (!rfsQueryParallel && dlfn.isExternal() && !lastNode())


jakesmith:
@mckellyln - please review

thorlcr/activities/thdiskbaseslave.cpp
```diff
@@ -430,10 +430,13 @@ void CDiskWriteSlaveActivityBase::close()
...
         }
 
+        Owned<IFileIO> tmpFileIO;
         {
             CriticalBlock block(statsCs);
             mergeStats(fileStats, outputIO);
-            outputIO.clear();
+
+            // ensure it is released/destroyed after releasing crit, since the IFileIO might involve a final copy and take considerable time.
+            tmpFileIO.setown(outputIO.getClear());
```
mckellyln: This seems fine to get out from under the statsCs.  But does it mean the timing info in the stats may not be correct ?  Is there a way to wait for the final copy to end and then do the mergeStats() ?

mckellyln:
@jakesmith one question.

thorlcr/activities/thdiskbaseslave.cpp
jakesmith: as it was / as it stands - it won't make any difference.
Because the way the statistics are collected here, it will be getting the stats for 1 IFileIO, which is the primary file written, and disregards the timing for the final copy.

So this change doesn't fix that, but you're right that it should
I'll open a separate JIRA to look at how fix that, it will be slightly more complicated than this fix.
@mckellyln

mckellyln: ok, sounds good.

jakesmith: For reference new JIRA: https://track.hpccsystems.com/browse/HPCC-23283

mckellyln:
@jakesmith approved for merge.

mckellyln (approved):

jakesmith:

jakesmith:
@ghalliday @richardkchapman - perhaps this should target 7.4.x?

thorlcr/activities/thdiskbaseslave.cpp
```diff
@@ -430,10 +430,13 @@ void CDiskWriteSlaveActivityBase::close()
...
         }
 
         if (!rfsQueryParallel && dlfn.isExternal() && !lastNode())
```
ghalliday: @jakesmith Should this be done before or after the file is destroyed?  If after then there should be a tmpFileIO.clear()

ghalliday:
@jakesmith looks good.  One question.

ghalliday:
I think it would be safe to include in an earlier version.

thorlcr/activities/thdiskbaseslave.cpp
jakesmith: so this is only in the case where the file is external (file:://) and all nodes are extending the same temporary output (on node 1). The last node will do the rename or copy to the final destination.

Triggering (to the next node) asap is preferrable, but it won't actually make any difference whether before or after, since it's only the last node that will do anything on the destruction of tmpFileIO.

@ghalliday - I'll rebase to 7.4.x

jakesmith:
@ghalliday - rebased. Probably worth waiting for smoketest to re-pass before merging.

PR #13289, opened 2020-01-23
HPCC-23082 Crypto Plugin needs key file access control

RussWhitehead:
Add new ECL module to load key files from HPCC LFN
Deprecate ECL module that loads key files from file system
Add new testcases for LFN load of key files

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [x] Performance
  - [x] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/plugins/cryptolib/cryptolib.cpp
+++ b/plugins/cryptolib/cryptolib.cpp
@@ -17,14 +17,14 @@
 #include "jexcept.hpp"
 #include "digisign.hpp"
 #include "ske.hpp"
-#include <openssl/pem.h>
-#include <openssl/err.h>
+#include "dadfs.hpp"
+#include "dautils.hpp"
+
 #include <openssl/rand.h>
-#include <string>
 #include <unordered_map>
-#include "jthread.hpp"
 
 #include "cryptolib.hpp"
+
 
 using namespace std;
 
@@ -177,20 +177,20 @@ void verifySymmetricAlgorithm(const char * algorithm, size32_t len)
      if (strieq(algorithm, "aes-128-cbc"))
      {
          if (len != 16)
-             throw MakeStringException(-1, "Invalid Key Length %d specified for algorithm %s, try 16", len, algorithm);
+             throw makeStringExceptionV(-1, "Invalid Key Length %d specified for algorithm %s, try 16", len, algorithm);
      }
      else if (strieq(algorithm, "aes-192-cbc"))
      {
          if (len != 24)
-             throw MakeStringException(-1, "Invalid Key Length %d specified for algorithm %s, try 24", len, algorithm);
+             throw makeStringExceptionV(-1, "Invalid Key Length %d specified for algorithm %s, try 24", len, algorithm);
      }
      else if (strieq(algorithm, "aes-256-cbc"))
      {
          if (len != 32)
-             throw MakeStringException(-1, "Invalid Key Length %d specified for algorithm %s, try 32", len, algorithm);
+             throw makeStringExceptionV(-1, "Invalid Key Length %d specified for algorithm %s, try 32", len, algorithm);
      }
      else
-         throw MakeStringException(-1, "Unsupported symmetric algorithm (%s) specified", algorithm);
+         throw makeStringExceptionV(-1, "Unsupported symmetric algorithm (%s) specified", algorithm);
 }
 
 CRYPTOLIB_API void CRYPTOLIB_CALL clSymmetricEncrypt(size32_t & __lenResult, void * & __result,
@@ -262,7 +262,7 @@ void throwHashError(const char * str)
     unsigned long err = ERR_get_error();
     char errStr[1024];
     ERR_error_string_n(err, errStr, sizeof(errStr));
-    throw MakeStringException(-1, "%s : ERROR %ld (0x%lX) : %s", str, err, err, errStr);
+    throw makeStringExceptionV(-1, "%s : ERROR %ld (0x%lX) : %s", str, err, err, errStr);
 }
 
 CRYPTOLIB_API void CRYPTOLIB_CALL clHash(size32_t & __lenResult, void * & __result,
@@ -362,7 +362,7 @@ CRYPTOLIB_API void CRYPTOLIB_CALL clHash(size32_t & __lenResult, void * & __resu
     }
     else
     {
-        throw MakeStringException(-1, "Unsupported hash algorithm '%s' specified", algorithm);
+        throw makeStringExceptionV(-1, "Unsupported hash algorithm '%s' specified", algorithm);
     }
 }
 
@@ -400,16 +400,62 @@ void doPKISign(size32_t & __lenResult, void * & __result,
     }
     else
     {
-        throw MakeStringException(-1, "Unable to create Digital Signature");
+        throw makeStringException(-1, "Unable to create Digital Signature");
     }
 }
 
 void verifyPKIAlgorithm(const char * pkAlgorithm)
 {
     if (!strieq(pkAlgorithm, "RSA"))
-        throw MakeStringException(-1, "Unsupported PKI algorithm (%s) specified", pkAlgorithm);
+        throw makeStringExceptionV(-1, "Unsupported PKI algorithm (%s) specified", pkAlgorithm);
 }
 
+
+//---------------------------------------------------------
+// Helper that reads the given logical file into a string buffer
+// Throws if user does not have read permission or if FNF
+//---------------------------------------------------------
+void loadLFS(const char * lfs, IUserDescriptor * user, StringBuffer &sb)
+{
+    CDfsLogicalFileName lfn;
+    lfn.set(lfs);
+    try
+    {
+        Owned<IDistributedFile> df = queryDistributedFileDirectory().lookup(lfn, user, false, false, false, nullptr, defaultPrivilegedUser);//scope checks
+        if (!df)
+        {
+            throw makeStringExceptionV(-1, "File %s Not Found", lfs);
+        }
+
+        if (df->numParts() == 0)
+        {
+            throw makeStringExceptionV(-1, "File %s is Empty", lfs);
+        }
+
+        IDistributedFilePart &part = df->queryPart(0);
+
+        RemoteFilename rfn;
+        Owned<IFile> file = createIFile(part.getFilename(rfn));
+        if (!file->exists())
+        {
+            throw makeStringExceptionV(-1, "File %s Not Found", lfs);
+        }
+
+        Owned<IFileIO> io = file->open(IFOread);
+        offset_t len = file->size();
+        char * buff = sb.reserveTruncate(len);
+        size32_t read = io->read(0, len, buff);
+        assertex(read == len);
+        //IFileIO dtor closes the file
+    }
+    catch(IException * e)
+    {
+        StringBuffer s;
+        VStringBuffer sb("Error accessing Key file '%s' : %s", lfs, e->errorMessage(s).str());
+        e->Release();
+        throw makeStringException(-1, sb.str());
+    }
+}
 
 //-----------------------------------------------------------------
 //  Simple cache for instances of Digital Signature Managers
@@ -422,7 +468,7 @@ private:
     DSMCache dsmCache;
 
 public:
-    IDigitalSignatureManager * getInstance(const char * algo, const char * pubKeyFS, const char * pubKeyBuff, const char * privKeyFS, const char * privKeyBuff, const char * passphrase)
+    IDigitalSignatureManager * getInstance(const char * algo, const char * pubKeyFS, const char * pubKeyBuff, const char * privKeyFS, const char * privKeyBuff, const char * passphrase, bool isLFN, IUserDescriptor * user)
     {
         VStringBuffer searchKey("%s_%s_%s_%s_%s_%s", algo, isEmptyString(pubKeyFS) ? "" : pubKeyFS, isEmptyString(pubKeyBuff) ? "" : pubKeyBuff,
                                 isEmptyString(privKeyFS) ? "" : privKeyFS, isEmptyString(privKeyBuff) ? "" : privKeyBuff, passphrase);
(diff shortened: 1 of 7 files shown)


RussWhitehead:
@jakesmith Please review

richardkchapman:
This should probably target 7.8 at this point?

plugins/cryptolib/cryptolib.cpp
```diff
@@ -411,6 +411,52 @@ void verifyPKIAlgorithm(const char * pkAlgorithm)
...
+void loadLFS(const char * lfs, IUserDescriptor * user, StringBuffer &sb)
+{
+    CDfsLogicalFileName lfn;
+    lfn.set(lfs);
+    try
+    {
+        Owned<IDistributedFile> df = queryDistributedFileDirectory().lookup(lfn, user, false, false, false, nullptr, defaultPrivilegedUser);//scope checks
+        if (!df)
+        {
+            throw MakeStringException(-1, "File Not Found");
```
jakesmith: minor: preferable to use makeStringExceptioV and output lfn name in exception message.

plugins/cryptolib/cryptolib.cpp
```diff
@@ -411,6 +411,52 @@ void verifyPKIAlgorithm(const char * pkAlgorithm)
...
+        }
+
+        unsigned numParts = df->numParts();
+        if (numParts != 1)
+        {
+            throw MakeStringException(-1, "Part count (%d) is incorrect", numParts);
```
jakesmith: minor: would be good to mention the lfn in all these exception message.

plugins/cryptolib/cryptolib.cpp
```diff
@@ -411,6 +411,52 @@ void verifyPKIAlgorithm(const char * pkAlgorithm)
...
+        }
+
+        IDistributedFilePart *part = df->getPart(0);
```
jakesmith: leaks, use queryPart() to avoid linking/leaking.

plugins/cryptolib/cryptolib.cpp
```diff
@@ -411,6 +411,52 @@ void verifyPKIAlgorithm(const char * pkAlgorithm)
...
+        RemoteFilename rfn;
+        Owned<IFile> file = createIFile(part->getFilename(rfn));
+        if (!file->exists())
+        {
+            throw MakeStringException(-1, "File Not Found");
+        }
+
+        Owned<IFileIO> io = file->open(IFOread);
+        offset_t len = file->size();
+        char * buff = sb.reserve(len);
```
jakesmith: trivial: slightly better to use reserveTruncate to avoid over allocation.

plugins/cryptolib/cryptolib.cpp
```diff
@@ -411,6 +411,52 @@ void verifyPKIAlgorithm(const char * pkAlgorithm)
...
+        size32_t read = io->read(0, len, buff);
+        assertex(read == len);
+        io->close();
```
jakesmith: trivia: don't need to close() fileio dtor will automatically close.

plugins/cryptolib/cryptolib.cpp
```diff
@@ -411,6 +411,52 @@ void verifyPKIAlgorithm(const char * pkAlgorithm)
...
+    }
+    catch(IException * e)
+    {
+        StringBuffer s;
+        VStringBuffer sb("Error accessing Key file '%s' : %s", lfs, e->errorMessage(s).str());
+        throw MakeStringException(-1, "%s", sb.str());
```
jakesmith: leaks 'e'. Should Release() before throwing new exception

plugins/cryptolib/cryptolib.cpp
```diff
@@ -436,7 +482,19 @@ class CDSMCache
...
+                bool isPublic = isEmptyString(privKeyFS);
+                const char * fs = isPublic ? pubKeyFS : privKeyFS;;
+
+                if (isLFN)
+                {
+                    StringBuffer sb;
+                    loadLFS(fs, user, sb);//read key file into StringBuffer
+                    ret = createDigitalSignatureManagerInstanceFromKeys(isPublic ? sb.str() : nullptr, isPublic ? nullptr : sb.str(), passphrase);
+                }
+                else
```
jakesmith: could do with comment saying only to support legacy direct files (unless we can remove completely).

plugins/cryptolib/cryptolib.cpp
```diff
@@ -484,14 +561,37 @@ CRYPTOLIB_API void CRYPTOLIB_CALL clPKISignBuff(size32_t & __lenResult, void * &
...
+
+CRYPTOLIB_API bool CRYPTOLIB_CALL clPKIVerifySignatureLFN(ICodeContext * ctx,
+                                                    const char * pkalgorithm,
+                                                    const char * publickeyLFN,
+                                                    const char * passphrase,
+                                                    size32_t lenSignature,const void * signature,
+                                                    size32_t lenSigneddata,const void * signeddata)
+{
+    verifyPKIAlgorithm(pkalgorithm);
+    Linked<IUserDescriptor> udesc = ctx->queryUserDescriptor();
```
jakesmith: trivial: no point in linking/releasing via a Linked object here, because it must be valid on entry/exit, and not passing ownership etc.

jakesmith (changes requested):

jakesmith:
The regression failure (with error "regress::certificates::privkey.pem&apos; : Part count (4)"), is because if run on Thor, files will have same # of parts as slaves, so code can't assume logical file has 1 part. But it's probably sufficient to just try to load the key from the 1st part.

jakesmith:
@RussWhitehead - do we need to keep the deprecated methods which would mean continuing to expose ECL functions that allow direct unchecked access to physical files ?

plugins/cryptolib/cryptolib.hpp
```diff
@@ -66,6 +66,15 @@ CRYPTOLIB_API void CRYPTOLIB_CALL clPKISign(size32_t & __lenResult,void * & __re
 CRYPTOLIB_API bool CRYPTOLIB_CALL clPKIVerifySignature(const char * pkalgorithm,const char * publickeyfile,const char * passphrase,size32_t lenSignature,const void * signature,size32_t lenSigneddata,const void * signeddata);
 
 
+
```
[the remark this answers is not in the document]
RussWhitehead: @jakesmith I do not know if anyone is using these features, but because it is documented and part of the STD library I think we need to go through the pains of deprecating

plugins/cryptolib/cryptolib.cpp
RussWhitehead: A while back in another review RKC didn't want me to display the (private) file spec because anyone with logfile access could find its location.  Perhaps not an issue so much now, since it is a LFN and might be scope protected.  Will do as you requested

plugins/cryptolib/cryptolib.cpp
RussWhitehead: Done. Must have missed that feature in the JLib docs :)

plugins/cryptolib/cryptolib.hpp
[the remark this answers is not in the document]
ghalliday: I agree with Russ - (and Dan's comment in the Jira).  For the situation where a system only has one user (or doesn't even have any users configured), deploying the keys locally to the nodes makes a lot of sense, and doesn't introduce a security problem.
It also means the file can be read locally, rather than having to use a strange logical file syntax (which II suspect may still read remotely).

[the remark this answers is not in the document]
jakesmith: right, hadn't seen or forgotten Dan's comment - I agree that's valid reason for wanting it.

But that means on systems that are multi user and they are trying to lock down access, the LDAP logical file validation can be bypassed by the versions that read from a file path.

Perhaps should instead allow direct paths, but validate users have permission via LDAP (if enabled) via another convention/ldap key, not via a LFN.

jakesmith:
@RussWhitehead - it does sound like there is a legitimate use case for the direct path version, meaning we will have to keep them for now.

plugins/cryptolib/cryptolib.cpp
```diff
@@ -466,14 +524,33 @@ CRYPTOLIB_API void CRYPTOLIB_CALL clPKISign(size32_t & __lenResult,void * & __re
...
+CRYPTOLIB_API void CRYPTOLIB_CALL clPKISignLFN(ICodeContext * ctx,
+                                        size32_t & __lenResult,void * & __result,
+                                        const char * pkalgorithm,
+                                        const char * privatekeyLFN,
+                                        const char * passphrase,
+                                        size32_t lenInputdata,const void * inputdata)
+{
+    verifyPKIAlgorithm(pkalgorithm);
+    Linked<IUserDescriptor> udesc = ctx->queryUserDescriptor();
+    Owned<IDigitalSignatureManager> pDSM = g_DSMCache.getInstance(pkalgorithm, nullptr, nullptr, privatekeyLFN, nullptr, passphrase, true, udesc);
```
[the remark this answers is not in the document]
RussWhitehead: I opened https://track.hpccsystems.com/browse/HPCC-23534 to address this change

plugins/cryptolib/cryptolib.hpp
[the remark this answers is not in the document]
RussWhitehead: So we will keep the local file system interface. I will remove all references to deprecation, and add to the Jira that we should elaborate on the risks of local keys, and highlight in the new DFS interfaces that users should lock them down with file scopes

plugins/cryptolib/cryptolib.cpp
RussWhitehead: Good catch (haha, see what I did there)

plugins/cryptolib/cryptolib.cpp
RussWhitehead: Done

RussWhitehead:
@jakesmith Please rereview based on your comments

richardkchapman:
This should target master for 7.8

RussWhitehead:
@richardkchapman Target changed to MASTER

RussWhitehead:
@jakesmith Please rereview

plugins/cryptolib/cryptolib.cpp
jakesmith: @RussWhitehead - look like this one has been missed

plugins/cryptolib/cryptolib.cpp
jakesmith: trivial: makeStringExceptionV preferred

plugins/cryptolib/cryptolib.cpp
```diff
@@ -448,14 +501,35 @@ class CDSMCache
...
     verifyPKIAlgorithm(pkalgorithm);//TODO extend cryptohelper to support more algorithms
-    Owned<IDigitalSignatureManager> pDSM = g_DSMCache.getInstance(pkalgorithm, nullptr, nullptr, privatekeyfile, nullptr, passphrase);
+    Owned<IDigitalSignatureManager> pDSM = g_DSMCache.getInstance(pkalgorithm, nullptr, nullptr, privatekeyfile, nullptr, passphrase, false, nullptr);
+    if (pDSM)
+    {
+        doPKISign(__lenResult, __result, pDSM, lenInputdata, inputdata);
+    }
+    else
+    {
+        throw MakeStringException(-1, "Unable to create Digital Signature Manager");
```
jakesmith: trivial: makeStringException preferred

plugins/cryptolib/cryptolib.cpp
ghalliday: or makeStringException(-1, sb.str())

plugins/cryptolib/cryptolib.cpp
```diff
@@ -400,17 +400,58 @@ void doPKISign(size32_t & __lenResult, void * & __result,
...
+        size32_t read = io->read(0, len, buff);
+        assertex(read == len);
+        //IFileIO dtor closes the file
+    }
+    catch(IException * e)
+    {
+        StringBuffer s;
+        VStringBuffer sb("Error accessing Key file '%s' : %s", lfs, e->errorMessage(s).str());
+        e->Release();
+        throw makeStringExceptionV(-1, sb.str());
```
ghalliday: need to remove the V

RussWhitehead:
@ghalliday Removed the unwanted V

ghalliday (commented):
@RussWhitehead one minor comment about a potential crash.  The commits should also be squashed.

plugins/cryptolib/cryptolib.cpp
```diff
@@ -400,17 +400,58 @@ void doPKISign(size32_t & __lenResult, void * & __result,
...
+    lfn.set(lfs);
+    try
+    {
+        Owned<IDistributedFile> df = queryDistributedFileDirectory().lookup(lfn, user, false, false, false, nullptr, defaultPrivilegedUser);//scope checks
+        if (!df)
+        {
+            throw makeStringExceptionV(-1, "File %s Not Found", lfs);
+        }
+
+        IDistributedFilePart &part = df->queryPart(0);
```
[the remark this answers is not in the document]
jakesmith: Yes that's a possibility.
@RussWhitehead I would check df->numParts() and throw an error saying empty if 0.

RussWhitehead:
@jakesmith @ghalliday I added that check as you recommended. But you should be aware there are many examples of calls throughout the platform to queryPart() that do not check for this error

plugins/cryptolib/cryptolib.cpp
```diff
@@ -428,6 +428,10 @@ void loadLFS(const char * lfs, IUserDescriptor * user, StringBuffer &sb)
         }
 
         IDistributedFilePart &part = df->queryPart(0);
+        if (!part)
```
jakesmith: @RussWhitehead - can you change it to test df->numParts()==0 instead?

It's probably illegal (or undefined) testing a reference for null.
NB: queryPart should probably thrown an exception rather than try and return a dodgy null reference, but don't change as part of this PR.

RussWhitehead:
@jakesmith @ghalliday Added "numParts() == 0" check and squashed

richardkchapman:
@RussWhitehead Does not appear to be squashed

RussWhitehead:
@richardkchapman Squashed

RussWhitehead:
@richardkchapman This appears ready for merge

PR #13300, opened 2020-02-02
HPCC-23433 Fix locking race condition with RTM_DELETE_ON_DISCONNECT

jakesmith:
When deleting the root node on a RTM_DELETE_ON_DISCONNECT, a
mutex was being released prematurely, allowing another thread
to delete the root of the connection at the same time.
This in turn caused a 'Transaction to orphaned server node'
exception to be thrown.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jmutex.cpp
+++ b/system/jlib/jmutex.cpp
@@ -15,7 +15,6 @@
     limitations under the License.
 ############################################################################## */
 
-
 #include "platform.h"
 #include "jmutex.hpp"
 #include "jsuperhash.hpp"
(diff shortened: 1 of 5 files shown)


jakesmith:
NB: a lot of the changes in this PR are cosmetic refactoring, resuling from the trivial CLCReadLockBlock, CLCWriteLockBlock classes being deleted in favour of their base class, so as to make it simpler to call the new changeToWrite method in the ReadWriteLock.

@mckellyln - please review.

system/jlib/jmutex.hpp
```diff
@@ -586,67 +586,88 @@ class jlib_decl Monitor: public Mutex
...
+    {
+        cs.enter();
+        if (writeLocks)
+        {
+            cs.leave();
+            return false; // failed because already write locked
+        }
+        if (readLocks)
+            readLocks--;
+        return doLockWrite(timed, timeout); // returns false if timed out
```
mckellyln: Should we try the doLockWrite() and then if ok decrement readLocks ?

system/jlib/jmutex.hpp
```diff
@@ -586,67 +586,88 @@ class jlib_decl Monitor: public Mutex
...
+            readLocks++;
+            cs.leave();
+        }
+        else
+        {
+            readWaiting++;
+            cs.leave();
+            if (timed)
+            {
+                if (!readSem.wait(timeout)) {
```
mckellyln: Might as well move these parens as well

mckellyln:
@jakesmith one question and one comment.

system/jlib/jmutex.hpp
jakesmith: yep agree, will change formatting here, and below.

system/jlib/jmutex.hpp
jakesmith: yes, but in fact that introduces a whole in itself. It should not release/reduce the readLocskc count here 1st.
Unfortunately, that makes the implementation trickier in that there may be something in a new waiting state.
I'll revisit.

jakesmith:
@ghalliday @mckellyln - more complicated than I had initially anticipated (in ReadWriteLock)

Please review.

system/jlib/jmutex.hpp
```diff
@@ -661,35 +710,74 @@ class jlib_decl ReadWriteLock
...
-                                cs.leave();
-                            }
+        exclWriteOwner = 0;
+#endif
+        return true;
+    }
+    void unlock()
+    {
+        cs.enter(); 
+        if (readLocks)
```
ghalliday: Could do with some comments.  E.g. if (readLocks is non zero then a read lock must be being released)

system/jlib/jmutex.hpp
```diff
@@ -661,35 +710,74 @@ class jlib_decl ReadWriteLock
...
+        changeToWrite(false, 0);
+    }
+    bool changeToWrite(unsigned timeout)
+    {
+        return changeToWrite(true, timeout);
+    }
+    bool changeToRead()
+    {
+        cs.enter();
+        if (!writeLocks)
```
ghalliday: minor - this could be checked outside the critical section.  Should it be an internal error rather than returning false?

system/jlib/jmutex.hpp
```diff
@@ -586,71 +586,120 @@ class jlib_decl Monitor: public Mutex
...
+        }
+#ifdef _DEBUG
+        exclWriteOwner = GetCurrentThreadId();
+#endif
+        return true;
+    }
+    bool changeToWrite(bool timed, unsigned timeout)
+    {
+        cs.enter();
+        if (writeLocks || (0 == readLocks))
```
ghalliday: more: should this  throw an internal error instead of returning false?  ((readLocks==0) could be checked before the cs.

ghalliday: or is the intention that it is ok to call this if it already has a write lock.  In which case the comment on the return is misleading.

ghalliday (commented):
@jakesmith I have looked at it and the change looks good.  (Although the code in jlib could do with more comments, and I'm not convinced it is good for it to all be inline).
@mckellyln I think it is worth you checking carefully as well.

system/jlib/jmutex.hpp
jakesmith: it's supposed to fail if already has a write lock, but a soft failure (return false) as opposed to an exception - in keeping with pre-exisitng timed lock read / write calls that return false on failure to get lock.

jakesmith: >((readLocks==0) could be checked before the cs.

should it though ? it's an edge case, not generally expected, and I think code clearer when checked inside the crit.

system/jlib/jmutex.hpp
jakesmith: I think a boolean result indicating not locked failure, similar to the failure on timed lock read/write calls.

>this could be checked outside the critical section

I'm not sure this edge case makes it worth moving the test, and clearer semantics if inside crit ?

system/jlib/jmutex.hpp
jakesmith: will add some comments.

jakesmith:
@ghalliday - thanks, added some replies and questions back re. whether some of these conditions really should be moved before locking the crit.

jakesmith:
.. and @mckellyln - please also review.

system/jlib/jmutex.hpp
ghalliday: I don't really mind if it is inside or outside.
If the function should never be called in that situation then it should be an internal error.  If it is expected then there should be a comment e.g.
//If the lock is already a read lock then this function will return false.

Similarly for changeToWrite - if the lock is already a write lock then it will return false.
Is that the correct response though?  If there is a timeout you can't distinguish between timedout and couldn't get the lock, and already had the write lock.

I don't have a strong view on which semantics you choose, but it should have a comment documenting what they are.

system/jlib/jmutex.hpp
ghalliday: The comment implies it is not legal to call it when not read locked.  If that is kind-of allowed the comment should be changed.  E.g.
//if called when not readlocked then fail immediately.

system/jlib/jmutex.hpp
jakesmith: >If there is a timeout you can't distinguish between timedout and couldn't get the lock, and already had the write lock.

ok yes I agree, on reflection I think these unexpected use cases should throw an exception instead. Will change.

jakesmith:
@mckellyln @ghalliday - changed invalid use cases to exceptions and added comments.

Have also moved some inline methods out of header into cpp, partly because use of jexcept inside header was causing circular header dependency issues.

mckellyln:
@jakesmith ok, reviewiing still

system/jlib/jmutex.cpp
```diff
@@ -321,7 +321,191 @@ void Monitor::notifyAll()
...
+            writeWaiting--;
+            writeLocks++;
+            writeSem.signal();
+        }
+    }
+    else // implies this unlock() is paired with a previous write lock
+    {
+        writeLocks--;
+        dbgassertex(writeLocks == 0);
+        if (readWaiting)
```
mckellyln: So we are ok with readers always acquiring before any writers ?
Perhaps the check for writers should come first ?  Or we could rotate the order of this check on each unlock ?

jakesmith: not sure, but that is not new semantics, ReadWriteLock has always given priority to pending blocked readers over blocked writers.

Worth considering, but as a separate task/discussion/JIRA.

jakesmith:
>@jakesmith ok, reviewiing still

@mckellyln - re-tagging

mckellyln (approved):
I am ok with this, but I do think writers should get priority over readers.

mckellyln:
@jakesmith ok for merge.

jakesmith:
>I am ok with this, but I do think writers should get priority over readers.

as I say, worth considering, but wouldn't want change the pre-existing semantics in this PR.
Please open new JIRA to look at / further discussion.

jakesmith:
@ghalliday - will squash ready for merge.

jakesmith:
@ghalliday - please merge.

system/jlib/jmutex.cpp
```diff
@@ -321,7 +321,191 @@ void Monitor::notifyAll()
...
+#ifdef _DEBUG
+    exclWriteOwner = GetCurrentThreadId();
+#endif
+    return true;
+}
+
+bool ReadWriteLock::changeToRead()
+{
+    cs.enter();
+    if (!writeLocks) // NB: If the lock is already a read lock then this function will return false.
```
ghalliday: I'm not sure the comment is correct

ghalliday (commented):
@jakesmith one minor comment on a comment

system/jlib/jmutex.cpp
jakesmith: no, should have been removed.

jakesmith:
@ghalliday - removed rogue comment.

PR #13306, opened 2020-02-04
HPCC-23428 Avoid the need to backpatch indexes

richardkchapman:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [x] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [x] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

## Testing:
Smoketest/unit tests

### DIFF
--- a/ecl/hthor/hthor.cpp
+++ b/ecl/hthor/hthor.cpp
@@ -1129,7 +1129,18 @@ void CHThorIndexWriteActivity::execute()
         Owned<IPropertyTree> metadata;
         buildUserMetadata(metadata);
         buildLayoutMetadata(metadata);
-        unsigned nodeSize = metadata ? metadata->getPropInt("_nodeSize", NODESIZE) : NODESIZE;
+        unsigned nodeSize = NODESIZE;
+        if (metadata)
+        {
+            nodeSize = metadata->getPropInt("_nodeSize", NODESIZE);
+            if (metadata->getPropBool("_noSeek", false))
+                flags |= TRAILING_HEADER_ONLY;
+            if (metadata->getPropBool("_useTrailingHeader", true))
+                flags |= USE_TRAILING_HEADER;
+        }
+        else
+            flags |= USE_TRAILING_HEADER;
+
         size32_t keyMaxSize = helper.queryDiskRecordSize()->getRecordSize(NULL);
         if (hasTrailingFileposition(helper.queryDiskRecordSize()->queryTypeInfo()))
             keyMaxSize -= sizeof(offset_t);
--- a/ecl/hthor/hthor.cpp
+++ b/ecl/hthor/hthor.cpp
@@ -1129,16 +1129,10 @@ void CHThorIndexWriteActivity::execute()
         Owned<IPropertyTree> metadata;
         buildUserMetadata(metadata);
         buildLayoutMetadata(metadata);
-        unsigned nodeSize = NODESIZE;
-        if (metadata)
-        {
-            nodeSize = metadata->getPropInt("_nodeSize", NODESIZE);
-            if (metadata->getPropBool("_noSeek", false))
-                flags |= TRAILING_HEADER_ONLY;
-            if (metadata->getPropBool("_useTrailingHeader", true))
-                flags |= USE_TRAILING_HEADER;
-        }
-        else
+        unsigned nodeSize = metadata->getPropInt("_nodeSize", NODESIZE);
+        if (metadata->getPropBool("_noSeek", false))
+            flags |= TRAILING_HEADER_ONLY;
+        if (metadata->getPropBool("_useTrailingHeader", true))
             flags |= USE_TRAILING_HEADER;
 
         size32_t keyMaxSize = helper.queryDiskRecordSize()->getRecordSize(NULL);
@@ -1335,7 +1329,7 @@ void CHThorIndexWriteActivity::buildUserMetadata(Owned<IPropertyTree> & metadata
     {
         StringBuffer name(nameLen, nameBuff);
         StringBuffer value(valueLen, valueBuff);
-        if(*nameBuff == '_' && strcmp(name, "_nodeSize") != 0)
+        if(*nameBuff == '_' && !checkReservedMetadataName(name))
         {
             OwnedRoxieString fname(helper.getFileName());
             throw MakeStringException(0, "Invalid name %s in user metadata for index %s (names beginning with underscore are reserved)", name.str(), fname.get());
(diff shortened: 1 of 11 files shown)


richardkchapman:
@ghalliday Please review once smoketest completes

system/jhtree/jhtree.cpp
```diff
@@ -3405,8 +3420,10 @@ class IKeyManagerTest : public CppUnit::TestFixture
     void testKeys()
     {
         ASSERT(sizeof(CKeyIdAndPos) == sizeof(unsigned __int64) + sizeof(offset_t));
-        testKeys(false);
-        testKeys(true);
+        for (bool var : { false, true })
```
ghalliday: neat

ecl/hthor/hthor.cpp
```diff
@@ -1129,7 +1129,18 @@ void CHThorIndexWriteActivity::execute()
         Owned<IPropertyTree> metadata;
         buildUserMetadata(metadata);
         buildLayoutMetadata(metadata);
-        unsigned nodeSize = metadata ? metadata->getPropInt("_nodeSize", NODESIZE) : NODESIZE;
+        unsigned nodeSize = NODESIZE;
+        if (metadata)
+        {
```
ghalliday: This code should really be common rather than duplicated, but it will probably get refactored up in the future.

system/jhtree/keybuild.cpp
```diff
@@ -201,8 +204,13 @@ class CKeyBuilderBase : public CInterface
...
         {
             out->flush();
-            out->seek(0, IFSbegin);
-            keyHdr->write(out, crc);
+            if (keyHdr->getKeyType() & USE_TRAILING_HEADER)
+                keyHdr->write(out, crc);  // write a copy at end too, for use on systems that can't seek
+            if (!(keyHdr->getKeyType() & TRAILING_HEADER_ONLY))
+            {
+                out->seek(0, IFSbegin);
+                keyHdr->write(out, crc);
```
ghalliday: If the trailing header isn't written then the crc from the original header write needs to be used.

system/jhtree/keybuild.cpp
```diff
@@ -201,8 +204,13 @@ class CKeyBuilderBase : public CInterface
         if (out)
         {
             out->flush();
-            out->seek(0, IFSbegin);
-            keyHdr->write(out, crc);
+            if (keyHdr->getKeyType() & USE_TRAILING_HEADER)
```
ghalliday: Is there a problem with updating the crcs? 
This may need to update a different crc variable. since it is the crc of an extra block, rather than the header that is being written.

ghalliday (commented):
@jakesmith please can you cast your eye over the changes.
@richardkchapman looks clean, but I think there is an issue with crc calculation.

system/jhtree/keybuild.cpp
richardkchapman: Worse than that - also need to include the crc of the trailing header.

richardkchapman:
@jakesmith Please review

system/jhtree/keybuild.cpp
jakesmith: the key header contains the crc for the whole file, it's just returning the crc of the KeyHdr node here, to roll it up with the crc's of the node data, to return the complete file crc for the file to the caller to publish it in Dali.

I think if you calc. once (pass null into one of the write calls if both are called), you can then merge in this crc before and after the node data crc in finish().

jakesmith:
@richardkchapman - couple of questions/comments.

thorlcr/activities/indexwrite/thindexwriteslave.cpp
```diff
@@ -171,7 +171,17 @@ class IndexWriteSlaveActivity : public ProcessSlaveActivity, public ILookAheadSt
             flags |= HTREE_TOPLEVEL_KEY;
         buildUserMetadata(metadata);                
         buildLayoutMetadata(metadata);
-        unsigned nodeSize = metadata ? metadata->getPropInt("_nodeSize", NODESIZE) : NODESIZE;
+        unsigned nodeSize = NODESIZE;
+        if (metadata)
```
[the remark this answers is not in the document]
richardkchapman: I will fix

thorlcr/activities/indexwrite/thindexwriteslave.cpp
```diff
@@ -171,7 +171,17 @@ class IndexWriteSlaveActivity : public ProcessSlaveActivity, public ILookAheadSt
...
+            if (metadata->getPropBool("_noSeek", false))
+                flags |= TRAILING_HEADER_ONLY;
+            if (metadata->getPropBool("_useTrailingHeader", true))
+                flags |= USE_TRAILING_HEADER;
+        }
+        else
+            flags |= USE_TRAILING_HEADER;
         builder.setown(createKeyBuilder(out, flags, maxDiskRecordSize, nodeSize, helper->getKeyedSize(), isTopLevel ? 0 : totalCount, helper, !isTlk, isTlk));
     }
     void buildUserMetadata(Owned<IPropertyTree> & metadata)
```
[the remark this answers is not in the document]
richardkchapman: Yes, they do. Good catch

system/jhtree/keybuild.cpp
richardkchapman: @jakesmith It's worse than that. I think I would have to return crc info for the original head info if I don't overwrite it, or new head if I do (to merge it in properly), and also ensure crc info for trailing header (if I write one) is folded in

richardkchapman:
@ghalliday @jakesmith I have tried to put in a proper fix for the CRCs, as well as addressing other comments.

CRCs are complex - we need to treat the trailing header as we would treat any other node write, and handle the leading one with its original contents if we don't overwrite it, or its new one if we do.

richardkchapman:
I'm not sure if this change is actually sensible, given there seem to be other places in index build code that require ability to seek (more investigation needed).

jakesmith:
>I'm not sure if this change is actually sensible, given there seem to be other places in index build code (such as blobs) that look like they require ability to seek (more investigation needed).

@richardkchapman - tagging you for now.

thorlcr/activities/indexwrite/thindexwriteslave.cpp
```diff
@@ -171,7 +171,12 @@ class IndexWriteSlaveActivity : public ProcessSlaveActivity, public ILookAheadSt
             flags |= HTREE_TOPLEVEL_KEY;
         buildUserMetadata(metadata);                
         buildLayoutMetadata(metadata);
-        unsigned nodeSize = metadata ? metadata->getPropInt("_nodeSize", NODESIZE) : NODESIZE;
+        // NOTE - if you add any more flags here, be sure to update checkReservedMetadataName
+        unsigned nodeSize = metadata->getPropInt("_nodeSize", NODESIZE);
```
ghalliday: minor: This code should really be a shared function, rather than duplicated.

ghalliday:
I will merge as is, so that the reading code will definitely be in 7.8.0.  New jira to be created to revisit creating blobs.

PR #13323, opened 2020-02-06
HPCC-23062 Compilation fails when cache file cannot be updated

shamser:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/ecl/hql/hqlexpr.cpp
+++ b/ecl/hql/hqlexpr.cpp
@@ -1112,7 +1112,17 @@ bool HqlParseContext::createCache(const char * simplifiedEcl, bool isMacro)
         writeStringToStream(*stream, "</Cache>\n");
         stream->flush();
     }
-    cacheFile->move(filename);
+    try
+    {
+        cacheFile->rename(filename);
+    }
+    catch (IException * e)
+    {
+        // compilation shouldn't fail just because the cache couldn't be copied over
+        // (Copy over may fail when another process has created cache and using cache already)
+        DBGLOG(e);
+        e->Release();
+    }
     return true;
 }
 


ecl/hql/hqlexpr.cpp
```diff
@@ -1112,7 +1112,15 @@ bool HqlParseContext::createCache(const char * simplifiedEcl, bool isMacro)
         writeStringToStream(*stream, "</Cache>\n");
         stream->flush();
     }
-    cacheFile->move(filename);
+    try
+    {
+        cacheFile->rename(filename);
```
shamser: File move is suitable for renames across file systems and always retries on failures.  Rename designed to only works within same file system and does not attempt retries (so, it's quicker).  

Rename is more suitable here as the source & target file is in same directory and retries are unnecessary.

ecl/hql/hqlexpr.cpp
```diff
@@ -1112,7 +1112,15 @@ bool HqlParseContext::createCache(const char * simplifiedEcl, bool isMacro)
...
+    }
+    catch(...)
+    {
+        // compilation shouldn't fail just because the cache couldn't be copied over
```
richardkchapman: You'll leak an exception here...

shamser: Fixed.

shamser:
@richardkchapman Please can you review.

richardkchapman:
Change looks fine, but we are not generally targeting 7.2.x for any but the most serious of issues (which this does not feel like it is). Can you retarget to 7.4.x?

shamser:
The issue was reported for 7.4.28 .  I'll re-target to 7.4.x.

PR #13330, opened 2020-02-10
HPCC-23352 Include docker files in our build sources

richardkchapman:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [x] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dockerfiles/buildall.sh
+++ b/dockerfiles/buildall.sh
@@ -1,0 +1,66 @@
+#!/bin/bash
+
+#todo - trim the cmake artefacts out of the base build
+ 
+# Work out which candidate branch this one was probably forked from
+
+git_parent() {
+  best=1000000
+  parent=
+  for candidate in `git branch | grep ' candidate-' | sed 's/\*//'` master ; do
+    delta=$(git rev-list --count --no-merges $candidate...HEAD)
+    if [ $delta -lt $best ] ; then
+      best=$delta
+      parent=$candidate
+    fi
+  done
+  if [ -z $parent ] ; then
+    exit 2
+  fi
+  echo $parent
+}
+
+#eval $(minikube docker-env)
+
+BASE_VER=7.8
+BUILD_VER=master-2020-02-10
+HEAD=$(git rev-parse --short HEAD)
+PARENT=`git_parent`
+PUSH=1
+
+set -e
+
+#docker rm $(docker ps -q -f 'status=exited')
+#docker rmi $(docker images -q -f "dangling=true")
+
+if ! docker pull hpccsystems/platform-build-base:${BASE_VER} ; then
+  docker image build -t hpccsystems/platform-build-base:${BASE_VER} platform-build-base/ 
+  if [ "$PUSH" = "1" ] ; then
+    docker push hpccsystems/platform-build-base:${BASE_VER}
+  fi
+fi
+
+if ! docker pull hpccsystems/platform-build:${BUILD_VER} ; then
+  docker image build -t hpccsystems/platform-build:${BUILD_VER} --build-arg BASE_VER=${BASE_VER} --build-arg BUILD_VER=${BUILD_VER} platform-build/  
+  if [ "$PUSH" = "1" ] ; then
+    docker push hpccsystems/platform-build:${BUILD_VER}
+  fi
+fi
+
+if ! docker pull hpccsystems/platform-core:${BUILD_VER} ; then
+  docker image build -t hpccsystems/platform-core:${BUILD_VER} --build-arg BUILD_VER=${BUILD_VER} platform-core/  
+  if [ "$PUSH" = "1" ] ; then
+    docker push hpccsystems/platform-core:${BUILD_VER}
+  fi
+fi
+
+#more - only do this if head does not match BUILD_VER
+docker image build -t hpccsystems/platform-build:${HEAD} --build-arg BUILD_VER=${BUILD_VER} --build-arg COMMIT=${HEAD} platform-build-incremental/
+docker image build -t hpccsystems/platform-core:${HEAD} --build-arg BUILD_VER=${HEAD} platform-core/  
+
+docker image build -t hpccsystems/roxie:${HEAD} --build-arg BUILD_VER=${HEAD} roxie/
+docker image build -t hpccsystems/dali:${HEAD} --build-arg BUILD_VER=${HEAD} dali/
+docker image build -t hpccsystems/esp:${HEAD} --build-arg BUILD_VER=${HEAD} esp/
+docker image build -t hpccsystems/eclccserver:${HEAD} --build-arg BUILD_VER=${HEAD} eclccserver/
+docker image build -t hpccsystems/eclagent:${HEAD} --build-arg BUILD_VER=${HEAD} eclagent/
+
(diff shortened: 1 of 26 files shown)


richardkchapman:
@ghalliday for initial review

dockerfiles/platform-build-base/Dockerfile
```diff
@@ -0,0 +1,86 @@
+FROM ubuntu:18.04 as build
+ENV DEBIAN_FRONTEND=noninteractive
```
ghalliday: I think I read somewhere it wasn't recommended to set this globally, rather than on a particular command, since any install in the container would also have it set.  That is only likely to be developers, so may not be a big issue.

dockerfiles/platform-build-base/Dockerfile
```diff
@@ -0,0 +1,86 @@
...
+  rsync \
+  psmisc \
+  libpopt0 \
+  zlib1g-dev
+
+RUN apt-get install -y \
+  bash-completion \
+  expect \
+  git \
+  nano \
```
ghalliday: Should also install vi(m) - then would cover most developers.

dockerfiles/platform-core/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+FROM ubuntu:18.04
+ENV DEBIAN_FRONTEND=noninteractive
+RUN apt clean && \
+    apt autoclean && \
+    apt install -f && \
+    apt autoremove && \
+    apt-get update
+
+RUN apt-get install -y \
+  xterm \
```
ghalliday: why xterm?

dockerfiles/roxie/Dockerfile
```diff
@@ -0,0 +1,9 @@
+ARG BUILD_VER
+FROM hpccsystems/platform-core:${BUILD_VER}
+
+USER hpcc
+RUN mkdir -p /var/lib/HPCCSystems/roxie
+WORKDIR /var/lib/HPCCSystems/roxie
+RUN roxie --init > roxie.json
```
ghalliday: Possibly ok for a demo, but I don't think you'll want to do this long term.

dockerfiles/startall.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
```
ghalliday: Add a comment describing what the script does.

ghalliday (commented):
@richardkchapman a few minor comments, but matches what I was expecting.

dockerfiles/buildall.sh
```diff
@@ -0,0 +1,66 @@
...
+  echo $parent
+}
+
+#eval $(minikube docker-env)
+
+BASE_VER=7.8
+BUILD_VER=master-2020-02-10
+HEAD=$(git rev-parse --short HEAD)
+PARENT=`git_parent`
+PUSH=1
```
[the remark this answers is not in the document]
richardkchapman: It should be a parameter. But I would expect most people using this code to never reach this point, as they will pull a pre-existing image for any tagged build

dockerfiles/dali/daliconf.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
richardkchapman: The plan would be that Dali should not need a config file by default, and that all components should look more like the roxie case. I think dali needs one at present because of the need to disable whitelist

dockerfiles/roxie/Dockerfile
richardkchapman: Depends whether we think that we want to supply the "default" config in text form for inspection/documentation or not.

dockerfiles/platform-build-base/Dockerfile
richardkchapman: It's only an issue if you expect derived containers to be used to launch interactive shells. Which we don't

dockerfiles/platform-core/Dockerfile
richardkchapman: It's in our documentation as a requirement...

(Unlike most of the dependencies in the next section, which are NOT listed as requirements, but should be!)

dockerfiles/roxie/Dockerfile
ghalliday: I thought we said the embedded config would have the defaults, and the shipped config would have them all commented out - so they wouldn't match.

dockerfiles/platform-core/Dockerfile
ghalliday: I haven't ever seen it as a requirement, I don't see it on the wiki.

dockerfiles/roxie/Dockerfile
richardkchapman: Yes, but I'm __think__ that the output from --init  should quite possibly be the latter rather than the former...

dockerfiles/platform-core/Dockerfile
richardkchapman: Hmmmm - maybe it was just in the docker file I used as my starting point... and the reasons for it being in there may be lost in the mists of time

I can try taking it out and see what happens.

dockerfiles/dali/daliconf.xml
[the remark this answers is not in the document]
richardkchapman: daserver is dali

dockerfiles/buildall.sh
```diff
@@ -0,0 +1,66 @@
...
+#more - only do this if head does not match BUILD_VER
+docker image build -t hpccsystems/platform-build:${HEAD} --build-arg BUILD_VER=${BUILD_VER} --build-arg COMMIT=${HEAD} platform-build-incremental/
+docker image build -t hpccsystems/platform-core:${HEAD} --build-arg BUILD_VER=${HEAD} platform-core/  
+
+docker image build -t hpccsystems/roxie --build-arg BUILD_VER=${HEAD} roxie/
+docker image build -t hpccsystems/dali --build-arg BUILD_VER=${HEAD} dali/
+docker image build -t hpccsystems/esp --build-arg BUILD_VER=${HEAD} esp/
+docker image build -t hpccsystems/eclccserver --build-arg BUILD_VER=${HEAD} eclccserver/
+docker image build -t hpccsystems/eclagent --build-arg BUILD_VER=${HEAD} eclagent/
+
```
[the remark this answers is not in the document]
richardkchapman: These component images should also be published, alongside the platform-core one. This script does not do so (yet)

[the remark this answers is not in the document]
xwang2713: I see the update which gives a tag for each Docker image. We probably still need build and push latest tags as well. The reason is that Docker pull/run will not pull new image if the same tag image already exists locally even a new image with the same tag pushed.. But for latest tag it always pull the image.

Also later we probably should allow user input a tag, for example "7.6.24-1" in buildall.sh. The tag from "git rev-parse --short HEAD" is hard for users to pick the Docker image.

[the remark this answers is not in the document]
xwang2713: I can build and deploy (startall.sh) on my local Linux.  ECL sample run successfully on EclWatch.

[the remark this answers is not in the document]
richardkchapman: I have a new PR in preparation that will automatically build and push (using github actions) each time a new tag is pushed

The use of HEAD here is intended for local debugging purposes only

PR #13334, opened 2020-02-11
HPCC-15610 Reduce optimization level for very complicated functions

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/ecl/hqlcpp/hqlhtcpp.cpp
+++ b/ecl/hqlcpp/hqlhtcpp.cpp
@@ -266,9 +266,9 @@ MemberFunction::MemberFunction(HqlCppTranslator & _translator, BuildCtx & classc
 {
     stmt = ctx.addQuotedFunction(text, (flags & MFdynamicproto) != 0);
     if (flags & MFoptimize)
-        stmt->setOptimize(true);
+        stmt->setForceOptimize(true);
     else if (flags & MFnooptimize)
-        stmt->setOptimize(false);
+        stmt->setForceOptimize(false);
 }
 
 MemberFunction::~MemberFunction() noexcept(false)
@@ -279,10 +279,12 @@ MemberFunction::~MemberFunction() noexcept(false)
     {
         //This estimate is before the peephole optimize is applied, so may be a large over estimate
         unsigned estimatedSize = calcTotalChildren(stmt);
+        //If the function is smaller than the optimize threshold then force optimization on
         if (maxOptimizeSize && (estimatedSize <= maxOptimizeSize))
-            stmt->setOptimize(true);
+            stmt->setForceOptimize(true);
+        //If the function is larger than the no-optimize threshold then force optimization off
         else if (minNoOptimizeSize && (estimatedSize >= minNoOptimizeSize))
-            stmt->setOptimize(false);
+            stmt->setForceOptimize(false);
     }
 
     //Do not process the aliases if we are aborting from an error
--- a/ecl/hqlcpp/hqlhtcpp.cpp
+++ b/ecl/hqlcpp/hqlhtcpp.cpp
@@ -265,10 +265,26 @@ MemberFunction::MemberFunction(HqlCppTranslator & _translator, BuildCtx & classc
 MemberFunction::MemberFunction(HqlCppTranslator & _translator, BuildCtx & classctx, const char * text, unsigned _flags) : translator(_translator), ctx(classctx), flags(_flags)
 {
     stmt = ctx.addQuotedFunction(text, (flags & MFdynamicproto) != 0);
+    if (flags & MFoptimize)
+        stmt->setOptimize(true);
+    else if (flags & MFnooptimize)
+        stmt->setOptimize(false);
 }
 
 MemberFunction::~MemberFunction() noexcept(false)
 {
+    unsigned maxOptimizeSize = translator.queryOptions().maxOptimizeSize;
+    unsigned minNoOptimizeSize = translator.queryOptions().minNoOptimizeSize;
+    if (stmt && (maxOptimizeSize || minNoOptimizeSize))
+    {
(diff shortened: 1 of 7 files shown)


ghalliday:
@shamser please review.
For an example look at wyoung1.eclxml.  Generate time is approximately 15s.  Compile time before this change ~8minutes, after this change - which only actually modifies the optimization level for a single function < 3minutes.  (Search for NOOPTIMIZE in the generated source.)

ecl/hqlcpp/hqlstmt.ipp
```diff
@@ -45,21 +45,21 @@ public:
...
             void                    setPriority(unsigned _prio) { priority = _prio; }
     virtual void                    finishedFramework() { throwUnexpected(); }
 
 protected:
     bool hasChildren() const;
+    StringBuffer & appendTextExtra(StringBuffer & out) const;
 
 protected:
     unsigned short                      priority;       //64bit: pack with link count in CInterface
-#ifdef _DEBUG
```
ghalliday: remove the conditional code by using an explicit base type for the enum.

ghalliday:
Shamser I had a few failures on a local test.  I need to investigate the cause.  Possibly this can only be used for member functions, and not for static functions.  If so I'll need to take that into account.

ghalliday:
Richard.  Also slightly concerned by the comment:

"This attribute should be used for debugging purposes only. It is not suitable in production code."

But I don't think it is likely to cause a problem - it is primarily a hint rather than a requirement.

ecl/hqlcpp/hqlhtcpp.cpp
```diff
@@ -265,10 +265,26 @@ MemberFunction::MemberFunction(HqlCppTranslator & _translator, BuildCtx & classc
...
 MemberFunction::~MemberFunction() noexcept(false)
 {
+    unsigned maxOptimizeSize = translator.queryOptions().maxOptimizeSize;
+    unsigned minNoOptimizeSize = translator.queryOptions().minNoOptimizeSize;
+    if (stmt && (maxOptimizeSize || minNoOptimizeSize))
+    {
+        //This estimate is before the peephole optimize is applied, so may be a large over estimate
+        unsigned estimatedSize = calcTotalChildren(stmt);
+        if (maxOptimizeSize && (estimatedSize <= maxOptimizeSize))
+            stmt->setOptimize(true);
```
shamser: It looks like this overrides the optimize flag provided in the constructor.  So, if define this:
`MemberFunction  x(...,MFnooptimize); `

then here it forces the function to optimize.

ecl/hqlcpp/hqlhtcpp.cpp
```diff
@@ -265,10 +265,26 @@ MemberFunction::MemberFunction(HqlCppTranslator & _translator, BuildCtx & classc
 MemberFunction::MemberFunction(HqlCppTranslator & _translator, BuildCtx & classctx, const char * text, unsigned _flags) : translator(_translator), ctx(classctx), flags(_flags)
 {
     stmt = ctx.addQuotedFunction(text, (flags & MFdynamicproto) != 0);
+    if (flags & MFoptimize)
+        stmt->setOptimize(true);
+    else if (flags & MFnooptimize)
+        stmt->setOptimize(false);
```
shamser: This looks like it's setting a boolean value (not 2 boolean values) and, so, it looks like optimize has in indeterminate value when neither MFoptimize and MFnooptimize is set.  I think it would be clearer to separate out the setter functions for 'optimize' and 'nooptimize'.

ecl/hqlcpp/hqlstmt.cpp
```diff
@@ -1301,6 +1313,12 @@ IHqlExpression * HqlStmt::queryExpr(unsigned index) const
     return NULL;
 }
 
+void HqlStmt::setOptimize(bool value)
+{
+    optimize = value;
+    noOptimize = !value;
```
shamser: No way to set both to false.

ecl/hqlcpp/hqlstmt.ipp
```diff
@@ -45,21 +45,21 @@ public:
...
     StmtKind                            kind;
-#else
-    unsigned char                       kind;
-#endif
     bool                                incomplete:1;
     bool                                included:1;
+    bool                                optimize:1;     // Should really be in a derived class, but this avoids extra memory
+    byte                                noOptimize:1;   // Should really be in a derived class, but this avoids extra memory
```
shamser: Worth making this bool for consistency?

shamser (changes requested):
@ghalliday

ecl/hqlcpp/hqlhtcpp.cpp
ghalliday: I've renamed the function to hopefully make it clearer.  It is a 3 state - default/optimize/nooptimize and it will never be reset to the default.

ghalliday:
@shamser I have addressed most of your concerns, and fixed the bug when the flag was added to static functions.

ecl/hqlcpp/hqlhtcpp.cpp
shamser: If estimateSize>maxOptimizeSize, then it does change the value for optimize here - it does not set it to false. This means the value of optimize will be the value as set in the  MemberFunction's constructor (lines 268-269).  i.e. if the flag & optimize=true, the optimize will not be set to false even if the estimated size is too big.

shamser (approved):
@richardkchapman Reviewed.

PR #13368, opened 2020-02-17
HPCC-23354 Supply an example helm chart and other docs

richardkchapman:
Work in progress, but I think the concept is promising enough to review and
merge as a basis for future development.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [x] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

## Testing:
Manual testing. Note that some other PRs opened today are required for the helm chart to work.


richardkchapman:
@xwang2713 Please review

dockerfiles/hpcc/files/roxie.yaml
```diff
@@ -0,0 +1,3 @@
+poo: "{{ $.Values.global.image.version }}"
```
richardkchapman: This is a test file only... should be replaced with something better once we support yaml reading

dockerfiles/hpcc/templates/_util.tpl
```diff
@@ -0,0 +1,86 @@
...
+
+{{- /* Add dali arg for a component */ -}}
+{{- define "hpcc.utils.daliArg" -}}
+"--daliServers={{ (index .Values.dali 0).name }}"
+{{- end -}}
+
+{{- /* Add image attributes for a component */ -}}
+{{- /* Pass in a dictionary with root and imagename defined */ -}}
+{{- define "hpcc.utils.addImageAttrs" -}}
+image: "hpccsystems/{{ .imagename }}:{{ .root.Values.global.image.version }}"
```
xwang2713: image repo first part probably can be a variable as well. It will be convenient for non Docker Hub hosting or some other cases in which the images are not hosted under hpccsystems in Docker Hub.

dockerfiles/hpcc/templates/dllserver-claim.yaml
```diff
@@ -8,4 +8,4 @@ spec:
     - ReadWriteMany
```
xwang2713: This is for local hostPath or NFS type, right?  For dynamic EBS or similar we probably should use StatefullSet instead which probably will be hard to make both "Deployment" and "StatefulSet" implementation in one Helm/Charts

dockerfiles/cleanup.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+# clean up old images
+
+docker rm $(docker ps -q -f 'status=exited')
```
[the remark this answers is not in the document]
richardkchapman: This is a utility script for developers - if they don't want to run it, they don't have to...

dockerfiles/hpcc/templates/dllserver-claim.yaml
richardkchapman: At the moment I'm only trying to support local hostPath.

I suspect for other types of location we can make entirely different dllserver/dllserver-claim templates.

dockerfiles/cleanup.sh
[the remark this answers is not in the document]
richardkchapman: I added a flag to control this

dockerfiles/hpcc/templates/_util.tpl
richardkchapman: I will allow it to be overridden

xwang2713: Actually I wonder users may want to change one component Docker image, for example, Dali. Should we provide an image variable for each component in values.yaml. If this is empty just do the normal path, otherwise users can use whatever Docker image they want, most likely with "helm upgrade".

xwang2713 (commented):
Looks good

dockerfiles/hpcc/templates/_util.tpl
richardkchapman: I think we will want to be able to override individual versions at some point - but I'll open a separate Jira for that.

PR #13380, opened 2020-02-19
HPCC-23378 Add YAML read and loadConfiguration support to jptree

afishbeck:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -46,3 +46,6 @@
 [submodule "system/aeron"]
 	path = system/aeron
 	url = https://github.com/hpcc-systems/aeron.git
+[submodule "system/libyaml"]
+	path = system/libyaml
+	url = https://github.com/hpcc-systems/libyaml.git
--- a/system/jlib/jptree.cpp
+++ b/system/jlib/jptree.cpp
@@ -24,6 +24,7 @@
 #include "jlzw.hpp"
 #include "jregexp.hpp"
 #include "jstring.hpp"
+#include "yaml.h"
 
 #ifdef __APPLE__
 #include <crt_externs.h>
@@ -7732,10 +7733,7 @@ static IPropertyTree * loadConfiguration(const char * filename, const char * com
     const char * ext = pathExtension(filename);
     Owned<IPropertyTree> configTree;
     if (strieq(ext, ".yaml"))
-    {
-        throw makeStringExceptionV(99, "YAML Configuration file %s not yet supported", filename);
-        //MORE: Translate Tags to attributes if they start with a lower case letter
-    }
+        configTree.setown(createPTreeFromYAMLFile(filename, 0, ptr_ignoreWhiteSpace, nullptr));
     else if (strieq(ext, ".json"))
     {
         configTree.setown(createPTreeFromJSONFile(filename, 0, ptr_ignoreWhiteSpace, nullptr));
@@ -7804,7 +7802,7 @@ IPropertyTree * createPTreeFromYAML(const char * yaml)
     if (*yaml == '{')
         return createPTreeFromJSONString(yaml, 0, ptr_ignoreWhiteSpace, nullptr);
 
-    throw makeStringExceptionV(99, "Someone needs to implement a YAML parser");
+    return createPTreeFromYAMLString(yaml, 0, ptr_ignoreWhiteSpace, nullptr);
 }
 
 static const char * extractOption(const char * option, const char * cur)
@@ -7880,11 +7878,13 @@ jlib_decl IPropertyTree * loadConfiguration(const char * defaultYaml, const char
         componentDefault.set(defaultConfig->queryPropTree(componentTag));
         if (!componentDefault)
             throw makeStringExceptionV(99, "Default configuration does not contain the tag %s", componentTag);
+        if (*defaultYaml=='{')
+            componentDefault.setown(mapJsonToXmlConfig(componentDefault));
     }
     else
         componentDefault.setown(createPTree(componentTag));
 
-    Owned<IPropertyTree> config(mapJsonToXmlConfig(componentDefault));
+    Linked<IPropertyTree> config(componentDefault);
     const char * optConfig = nullptr;
     bool outputConfig = false;
     for (const char * * pArg = argv; *pArg; pArg++)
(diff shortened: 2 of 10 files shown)


afishbeck:
@richardkchapman Yaml part 1, who should review?
Note that switching the defaultJson to defaultYaml could be considered an initial test case
Who can review submodule and cmake chnages?

to test:
roxie --outputconfig --config=roxie.yaml
agentexec  --outputconfig --config=config.yaml
eclccserver --outputconfig --config=eclccserver.yaml

afishbeck:
converted to direct source for now.  there was a timing issue building the submodule with multiple make threads.  Trying to pass a setting variable that wasn't recognized if the thread count was too high.

system/jlib/jptree.cpp
```diff
@@ -7881,11 +7879,13 @@ jlib_decl IPropertyTree * loadConfiguration(const char * defaultYaml, const char
         componentDefault.set(defaultConfig->queryPropTree(componentTag));
         if (!componentDefault)
             throw makeStringExceptionV(99, "Default configuration does not contain the tag %s", componentTag);
+        if (*defaultYaml=='{')
+            componentDefault.setown(mapJsonToXmlConfig(componentDefault));
     }
     else
         componentDefault.setown(createPTree(componentTag));
 
-    Owned<IPropertyTree> config(mapJsonToXmlConfig(componentDefault));
```
richardkchapman: Not sure about this change... I think you are trying to do the mapping if json passed in but not if yaml is, but I would think the mapping is needed in both cases

richardkchapman: Looks like you did the mapping (#text, lowercase implies attribute) in the yaml reader instead.

Not sure what I think about that - @ghalliday will probably need to comment.

afishbeck: We do something similar in JSON.  Although I just realized we don't use "#text" we use "#value".  There are things that we can represent in PTREE and xml that we can't in JSON and YAML.  Supporting this natively allows us to.

XML -> JSON -> YAML -> XML without loosing as much information.
In this case its:
<tag at="1">text or value/<tag>
If fromYAML and toYAML both recognize #value or #text or whatever then that info can be preserved without a user realizing.
Internal XPATH users would  currently realize though, but we could probably add support somehow if we add support complex tag names.

afishbeck: The native lower case to attribute mapping would have one side effect:
xml -> yaml -> xml could leave you with more as attributes than the original xml.

afishbeck: Really all named scalars could be attributes when reading from Yaml, but:
  - no way to prevent really large content attributes.. how much does that matter?  Doesn't look great serializing to XML.
  - would be even less symmetric when translating back and forth

afishbeck:
@richardkchapman updated

Couldn't get libyaml to work using saving and setting "BUILD_SHARED_LIBS" so modified my local fork and opened a PR on the HPCC-SYSTEM libyaml fork.  We can move to that if you merge that PR.

Note about YAML to PTREE mappings (no specific config mapping should be necessary):
Sequence items (arrays) are always elements.

By default I now load ALL YAML scalars as attributes.  ToYAML should write all attributes as scalars.
```
x:
  a: "zzz"
  b: true

is loaded as:
<x a="zzz" b="true"/>


But To store as an element I've used a local tag (a custom type tagged "!el" or "!element"):
x:
  a: zzz
  b: !el true

is loaded as:
<x a="zzz">
  <b>true</b>
</x>

To store as the text of the parent element use local tag and name the scalar ^:
x:
  a: zzz
  b: true
  ^: !el how now  #could add support without !el tag but wanted to be explicit first pass

is loaded as:
<x a="zzz" b="true">how now</x>

```

richardkchapman:
I merged your PR into our libyaml fork

.gitmodules
```diff
@@ -46,3 +46,6 @@
 [submodule "system/aeron"]
 	path = system/aeron
 	url = https://github.com/hpcc-systems/aeron.git
+[submodule "system/libyaml"]
+	path = system/libyaml
+	url = https://github.com/afishbeck/libyaml.git
```
richardkchapman: Needs to refer to hpcc-systems

system/jlib/jptree.cpp
```diff
@@ -7961,3 +7961,262 @@ jlib_decl IPropertyTree * loadConfiguration(const char * defaultYaml, const char
...
+            iEvent->beginNode(tagname, parser.offset);
+
+        yaml_event_t event;
+        yaml_event_type_t eventType = YAML_NO_EVENT;
+
+        while (eventType!=YAML_MAPPING_END_EVENT)
+        {
+            eventType = nextEvent(event, YAML_MAPPING_END_EVENT, YAML_SCALAR_EVENT, "expected map to start with scalar name");
+            if (eventType==YAML_MAPPING_END_EVENT)
+                continue;
```
ghalliday: minor: Should this also call yaml_event_delete(&event)?  I suspect it will have no effect,

system/jlib/jptree.cpp
```diff
@@ -7961,3 +7961,262 @@ jlib_decl IPropertyTree * loadConfiguration(const char * defaultYaml, const char
...
+                    else
+                    {
+                        iEvent->beginNode(name, parser.offset);
+                        iEvent->endNode(name, event.data.scalar.length, (const void *) event.data.scalar.value, false, parser.offset);
+                    }
+                }
+                else //by default all named scalars are ptree attributes
+                {
+                    name.insert(0, '@');
+                    StringAttr value((const char *) event.data.scalar.value, event.data.scalar.length);
```
ghalliday: minor: If typical length is < 16 characters then a local StringBuffer may be more efficient.  Even more efficient would be a way of avoiding creating a temporary.

system/jlib/jptree.cpp
```diff
@@ -7881,11 +7879,13 @@ jlib_decl IPropertyTree * loadConfiguration(const char * defaultYaml, const char
...
         if (!componentDefault)
             throw makeStringExceptionV(99, "Default configuration does not contain the tag %s", componentTag);
+        if (*defaultYaml=='{')
+            componentDefault.setown(mapJsonToXmlConfig(componentDefault));
     }
     else
         componentDefault.setown(createPTree(componentTag));
 
-    Owned<IPropertyTree> config(mapJsonToXmlConfig(componentDefault));
+    Owned<IPropertyTree> config(componentDefault);
```
ghalliday: This should be Linked rather than Owned, since it isn't taking ownership.  (I think for some strange reason this code may  work with Owned.)

system/jlib/jptree.cpp
```diff
@@ -7961,3 +7961,262 @@ jlib_decl IPropertyTree * loadConfiguration(const char * defaultYaml, const char
...
+                loadMap(name);
+                break;
+            case YAML_SEQUENCE_START_EVENT:
+                loadSequence(name);
+                break;
+            case YAML_SCALAR_EVENT:
+            {
+                //!el or !element will be our local tag (custom schema type) for an element
+                //ptree toYAML should set this for element scalars, and parent text content
+                if (event.data.scalar.tag && (streq((const char *)event.data.scalar.tag, "!el") || streq((const char *)event.data.scalar.tag, "!element")))
```
ghalliday: Worth having an example of the syntax.  More broadly it would be worth having a comment (in the header) that documents how the yaml content is mapped to the IPropertyTree.

ghalliday (commented):
Looks good.
Is there a separate jira for generating a YAML file from an IPropertyTree?

afishbeck:
Rebased.  This area is a moving target.  Will address Gavin's comments shortly.

system/jlib/jptree.cpp
afishbeck: Agreed.  No leaks detected but could change.

system/jlib/jptree.cpp
afishbeck: A bit unfortunate that setAttribute() requires the '@' since it is implied.  I may be able to avoid the insert.

AttilaVamos:
Those wutool test errors related to  FSDiskFullWriteError are not related to this PR. They are result of a some kind of environment issues, but I think the disk is not full.

system/jlib/jptree.cpp
afishbeck: actually value does seem to be null terminated, so should be able to use directly.

system/jlib/jptree.cpp
afishbeck: actually value does seem to be null terminated, so should be able to use directly.

afishbeck:
@ghalliday updated, addressed each of your comments.

PR #13383, opened 2020-02-20
HPCC-23543 Rationalize eclagent command line parameters

richardkchapman:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [x] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/ecl/eclagent/eclagent.cpp
+++ b/ecl/eclagent/eclagent.cpp
@@ -70,8 +70,6 @@ using roxiemem::OwnedRoxieString;
 
 #define MONITOR_ECLAGENT_STATUS     
  
-static const char XMLHEADER[] = "<?xml version=\"1.0\" encoding=\"UTF-8\"?>";
-
 //#define ROOT_DRIVE      "c:"
 
 //#define DEFAULT_REALTHOR_HOST "localhost"
@@ -90,6 +88,7 @@ constexpr LogMsgCategory MCresolve = MCprogress(100);       // Category used to 
 constexpr LogMsgCategory MCrunlock = MCprogress(100);      // Category used to inform about run lock progress
 
 Owned<IPropertyTree> agentTopology;
+Owned<IProperties> cmdLineArgs;
 
 MODULE_INIT(INIT_PRIORITY_STANDARD)
 {
@@ -98,6 +97,7 @@ MODULE_INIT(INIT_PRIORITY_STANDARD)
 MODULE_EXIT()
 {
     agentTopology.clear();
+    cmdLineArgs.clear();
 }
 
 //-----------------------------------------------------------------------------------------------------
@@ -156,6 +156,7 @@ public:
     CHThorDebugSocketListener(CHThorDebugContext * _debugContext) 
         : Thread("CHThorDebugSocketListener"), debugContext(_debugContext)
     {
+        port = 0;
         running = false;
         suspended = false;
         unsigned poolSize = 10;//MORE : What is a good threadcoutn?
@@ -509,8 +510,8 @@ public:
 
 //=======================================================================================
 
-EclAgent::EclAgent(IConstWorkUnit *wu, const char *_wuid, bool _checkVersion, bool _resetWorkflow, bool _noRetry, char const * _logname, const char *_allowedPipeProgs, IPropertyTree *_queryXML, IProperties *_globals, IPropertyTree *_config, ILogMsgHandler * _logMsgHandler)
-    : wuRead(wu), wuid(_wuid), checkVersion(_checkVersion), resetWorkflow(_resetWorkflow), noRetry(_noRetry), allowedPipeProgs(_allowedPipeProgs), globals(_globals), config(_config), logMsgHandler(_logMsgHandler)
+EclAgent::EclAgent(IConstWorkUnit *wu, const char *_wuid, bool _checkVersion, bool _resetWorkflow, bool _noRetry, char const * _logname, const char *_allowedPipeProgs, IPropertyTree *_queryXML, ILogMsgHandler * _logMsgHandler)
+    : wuRead(wu), wuid(_wuid), checkVersion(_checkVersion), resetWorkflow(_resetWorkflow), noRetry(_noRetry), allowedPipeProgs(_allowedPipeProgs), logMsgHandler(_logMsgHandler)
 {
     isAborting = false;
     isStandAloneExe = false;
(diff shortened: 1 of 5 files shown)


richardkchapman:
@shamser For initial review (not yet tested!)

ecl/eclagent/eclagent.cpp
```diff
@@ -2020,12 +2012,6 @@ void EclAgent::doProcess()
                 UWARNLOG(errCode, "%s", rmMsg.str());
         }
 
-        if (globals->getPropBool("DUMPFINALWU", false))
```
shamser: Why is this option being removed?

ecl/eclagent/eclagent.cpp
```diff
@@ -3546,10 +3492,10 @@ extern int HTHOR_API eclagent_main(int argc, const char *argv[], StringBuffer *
                 IExtendedWUInterface * extendedWu = queryExtendedWU(daliWu);
                 extendedWu->copyWorkUnit(standAloneWorkUnit, true, true);
                 wuid.set(daliWu->queryWuid());
-                globals->setProp("WUID", wuid.str());
+                cmdLineArgs->setProp("WUID", wuid.str());  // So it can be retrieved by getenv
```
shamser: Should this be lowercase "wuid"?

ecl/eclagent/eclagent.cpp
```diff
@@ -3623,19 +3569,21 @@ extern int HTHOR_API eclagent_main(int argc, const char *argv[], StringBuffer *
 
             if (w)
             {
-                EclAgent agent(w, wuid.str(), globals->getPropInt("IGNOREVERSION", 0)==0, globals->getPropBool("WFRESET", false), globals->getPropBool("NORETRY", false), logfilespec.str(), globals->queryProp("allowedPipePrograms"), query.getClear(), globals, agentTopology, logMsgHandler);
-                const bool isRemoteWorkunit = (daliServers.length() != 0);
-                const bool resolveFilesLocally = standAloneExe && (!isRemoteWorkunit || globals->getPropBool("USELOCALFILES", false));
-                const bool writeResultsToStdout = standAloneExe && (!isRemoteWorkunit || globals->getPropBool("RESULTSTOSTDOUT", true));
+                EclAgent agent(w, wuid.str(), agentTopology->getPropBool("@ignoreVersion", false), agentTopology->getPropBool("@resetWorkflow", false), agentTopology->getPropBool("@noretry", false), logfilespec.str(),
```
shamser: should it be "@noRetry"?

shamser (changes requested):
@richardkchapman Few minors.  Also:  The same command line arguments is being processed in 2 different places: eclagent_main (writing into cmdLineArgs) and loadConfiguration (writing into agentTopology) .  It may get confusing which one to use.

Would it make sense to have command line args processed only in loadConfiguration and getting rid of cmdLineArgs completely?

ecl/eclagent/eclagent.cpp
richardkchapman: It was undocumented and not especially useful (IMHO)

ecl/eclagent/eclagent.cpp
richardkchapman: If there was any ECL code calling getenv to get it, it would previously have been there as WUID, so I left it.

getenv should probably be made case-insensitive, but that's a different (existing) bug.

ecl/eclagent/eclagent.cpp
richardkchapman: Arguably, yes. None of these options are documented anyway, and are only ever used by developers after looking at the source, so it's not that critical...

shamser (approved):
@richardkchapman Reviewed.  Not sure why the build is failing.

richardkchapman:

ecl/eclagent/eclagent.cpp
```diff
@@ -3307,6 +3268,18 @@ int myhook(int alloctype, void *, size_t nSize, int p1, long allocSeq, const uns
 }
 #endif
 //--------------------------------------------------------------
+
+void usage()
+{
+    printf("USAGE: eclagent --wuid=wuid options\n"
+           "       eclagent PROCESS=dllname options\n"
```
ghalliday: where is this option implemented?

ecl/eclagent/eclagent.ipp
```diff
@@ -43,6 +43,8 @@ class EclGraph;
 typedef unsigned __int64 graphid_t;
 typedef unsigned __int64 activityid_t;
 
+extern Owned<IPropertyTree> agentTopology;
```
ghalliday: Would be cleaner having agentTopology as a member of eclagent, rather than exposing it as a global variable.

ghalliday (commented):
@richardkchapman A few minor comments

ecl/eclagent/eclagent.cpp
```diff
@@ -90,15 +88,7 @@ constexpr LogMsgCategory MCresolve = MCprogress(100);       // Category used to
...
 
 Owned<IPropertyTree> agentTopology;
-
-MODULE_INIT(INIT_PRIORITY_STANDARD)
-{
-    return true;
-}
-MODULE_EXIT()
-{
-    agentTopology.clear();
```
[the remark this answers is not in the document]
richardkchapman: Fairly. This change could only potentially make the object be destroyed earlier.

I suspect it was not always Owned, and that this code was changed from a delete to a clear when the object was changed from a pointer to Owned...

ecl/eclagent/eclagent.cpp
richardkchapman: It isn't, I suspect. The usage functions(s) are very inconsistent and out of date. I fixed some issues in them but looks like I didn't get all of them (this is old code that was moved)

ecl/eclagent/eclagent.ipp
richardkchapman: I didn't want to refactor to that extent. Plus you can argue that global options belong in a global variable...

It used to be both, which was very confusing.

ecl/eclagent/eclagent.cpp
```diff
@@ -90,15 +88,7 @@ constexpr LogMsgCategory MCresolve = MCprogress(100);       // Category used to
 constexpr LogMsgCategory MCrunlock = MCprogress(100);      // Category used to inform about run lock progress
 
 Owned<IPropertyTree> agentTopology;
```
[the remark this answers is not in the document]
richardkchapman: Possibly, but the name is not new.

ecl/eclagent/eclagent.cpp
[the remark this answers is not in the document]
ghalliday: I think it will actually mean it is called later (when exe is quite, rather than when ExitModuleObjects() is called).

PR #13409, opened 2020-02-27
HPCC-23515 Create Docker and Helm charts for Thor

jakesmith:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dockerfiles/hpcc/templates/_util.tpl
+++ b/dockerfiles/hpcc/templates/_util.tpl
@@ -90,3 +90,23 @@ volumeMounts:
 image: "{{ include "hpcc.utils.imageName" . }}"
 imagePullPolicy: {{ .root.Values.global.image.pullPolicy }}
 {{- end -}}
+
+{{- /* A kludge to ensure host mounted storage (e.g. for minikube or docker for desktop) has correct permissions for PV */ -}}
+{{- define "hpcc.utils.changeMountPerms" -}}
+initContainers:
+# This is a bit of a hack, to ensure that the persistent storage mounted
+# is writable. This is not something we would want to do if using anything other than
+# local storage (which is only sensible on single-node systems).
+# NB: uid=999 and gid=1000 are the uid/gid of the hpcc user, built into platform-core
+{{- $permCmd := printf "chown -R 999:1000 %s" .volumePath }}
+- name: volume-mount-hack
+  image: busybox
+  command: [
+             "sh",
+             "-c",
+             "{{ $permCmd }}"
+           ]
+  volumeMounts:
+    - name: {{ .volumeName | quote}}
+      mountPath: {{ .volumePath | quote }}
+{{- end }}
--- a/dockerfiles/hpcc/templates/thor.yaml
+++ b/dockerfiles/hpcc/templates/thor.yaml
@@ -1,0 +1,126 @@
+{{ range $thor := $.Values.thor -}}
+{{- $masterPort := $thor.masterport | default 20000 -}}
+{{- $slavePort := $thor.slaveport | default 20100 -}}
+{{- $pvName  := printf "%s-pv" .name }}
+{{- $pvcName := printf "%s-pvc" .name }}
+{{- $slaveName := printf "%s-slave" .name }}
+{{- $serviceName := printf "%s-svc" .name }}
+
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ .name | quote }}
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      run: {{ .name | quote }}
+  template:
+    metadata:
+      labels:
+        run: {{ .name | quote }}
+    spec:
+      containers:
+      - name: {{ .name | quote }}
+        ports:
+          - containerPort: {{ $masterPort }}
+        args: [
+                {{ include "hpcc.utils.configArg" . }},
+                {{ include "hpcc.utils.daliArg" $ }},
+                --masterport={{ $masterPort }},
+                --numSlaves={{ $thor.numSlaves }}
+              ]
+{{ include "hpcc.utils.addImageAttrs" (dict "root" $ "imagename" "thormaster") | indent 8 }}
+{{ include "hpcc.utils.addVolumeMounts" . | indent 8 }}
+{{ include "hpcc.utils.addVolumes" . | indent 6 }}
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ $slaveName | quote }}
+spec:
+  replicas: {{ $thor.numSlaves }}
+  selector:
+    matchLabels:
+      run: {{ $slaveName | quote }}
+  template:
+    metadata:
+      labels:
+        run: {{ $slaveName | quote }}
+    spec:
+      {{- if $.Values.global.singleNode | default false }}
+{{ include "hpcc.utils.changeMountPerms" (dict "volumeName" "mythorstorage" "volumePath" "/var/lib/HPCCSystems/hpcc-data") | indent 6 }}
+      {{- end }}
+      containers:
+        - name: {{ $slaveName | quote }}
+          args: [
+                  {{ include "hpcc.utils.configArg" . }},
+                  {{ include "hpcc.utils.daliArg" $ }},
+                  --slaveport={{ $slavePort }},
+                  --master={{ printf "%s:%v" $serviceName $masterPort }}
+                ]
+{{ include "hpcc.utils.addImageAttrs" (dict "root" $ "imagename" "thorslave") | indent 10 }}
+          ports:
+            - containerPort: {{ $slavePort }}
+{{ include "hpcc.utils.addVolumeMounts" . | indent 10 }}
+          - name: "mythorstorage"
+            mountPath: "/var/lib/HPCCSystems/hpcc-data"
+        {{- if $thor.startDafilesrv | default false }}
+        - name: {{ printf "%s-dafilesrv" $slaveName | quote }}
+{{ include "hpcc.utils.addImageAttrs" (dict "root" $ "imagename" "thorslave") | indent 10 }}
+          command: ["dafilesrv"]
+          ports:
+            - containerPort: 7100
+          volumeMounts:
+            - name: "mythorstorage"
+              mountPath: "/var/lib/HPCCSystems/hpcc-data"
+        {{- end }}
+{{ include "hpcc.utils.addVolumes" . | indent 6 }}
+      - name: "mythorstorage"
+        persistentVolumeClaim:
+          claimName: {{ $pvcName | quote }}
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ $serviceName | quote }}
+spec:
+  ports:
+  - port: {{ $masterPort }}
+    protocol: TCP
+    targetPort: {{ $masterPort }}
+  selector:
+    run: {{ .name | quote }}
+  type: ClusterIP
+---
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  name: {{ $pvName | quote }}
+  labels:
+    type: local
+spec:
+  storageClassName: manual
+  capacity:
+    storage: {{ $thor.storageSize }}
+  accessModes:
+    - ReadWriteMany
+  persistentVolumeReclaimPolicy: Retain
+  hostPath:
+    path: {{ printf "%s/hpcc-data" $.Values.global.hostMountPath | quote }}
+    type: DirectoryOrCreate
+---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: {{ $pvcName | quote }}
+spec:
+  storageClassName: manual
+  accessModes:
+    - ReadWriteMany
+  resources:
+    requests:
+      storage: {{ $thor.storageSize }}
+---
+{{- include "hpcc.utils.generateConfigMap" (dict "root" $ "me" .) -}}
+{{- end }}
--- a/dockerfiles/hpcc/values.yaml
+++ b/dockerfiles/hpcc/values.yaml
@@ -18,6 +18,10 @@ global:
   
   singleNode: true
 
+  # For singleMode only. Host mount point root for persistent storage.
+  hostMountPath: "/docker-host-mnt"
+
+
 dali:
 - name: mydali
 
@@ -50,5 +54,10 @@ roxie:
       Roxie:
         useAeron: true
 
-
-    
+thor:
+  - name: thor1
+    masterport: 20000
+    numSlaves: 2
+    slaveport: 20100
+    startDafilesrv: true # Temporary until non attached storage functionality added
+    storageSize: 100G
--- a/system/jlib/jptree.cpp
+++ b/system/jlib/jptree.cpp
@@ -7895,6 +7895,17 @@ IPropertyTree & queryGlobalConfig()
     return *globalConfiguration;
 }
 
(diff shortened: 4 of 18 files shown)


jakesmith:

richardkchapman:
@jakesmith Needs rebasing

dockerfiles/hpcc/templates/thor.yaml
```diff
@@ -0,0 +1,128 @@
...
+    matchLabels:
+      run: {{ printf "%s-slave" .name | quote }}
+  template:
+    metadata:
+      labels:
+        run: {{ printf "%s-slave" .name | quote }}
+    spec:
+      {{ if $.Values.global.singleNode | default false -}}
+      initContainers:
+      # This is a bit of a hack, to ensure that the persistent storage mounted for dllserver
```
richardkchapman: Comment is wrong - this is not for dllserver

dockerfiles/hpcc/templates/thor.yaml
```diff
@@ -0,0 +1,128 @@
...
+      # is writable. This is not something we would want to do if using anything other than
+      # local storage (which is only sensible on single-node systems).
+      - name: volume-mount-hack
+        image: busybox
+        command: ["sh", "-c", "chown -R 999:1000 /var/lib/HPCCSystems/hpcc-data"]
```
richardkchapman: Should do chmod 755 rather than chown, I suspect. That way don't need to rely on magic user/gruop ids that may or may not be right in the future.

Should also consider templatizing as we are going to need this for any place we want persistent storage.

dockerfiles/hpcc/templates/thor.yaml
```diff
@@ -0,0 +1,128 @@
...
+                  --slaveport={{ $thor.slaveport }},
+                  --master={{ printf "%s-svc:%v" .name $thor.masterport }}
+                ]
+{{ include "hpcc.utils.addImageAttrs" (dict "root" $ "imagename" "thorslave") | indent 10 }}
+          ports:
+            - containerPort: {{ $thor.slaveport }}
+{{ include "hpcc.utils.addVolumeMounts" . | indent 10 }}
+          - mountPath: /var/lib/HPCCSystems/hpcc-data
+            name: mythorstorage
+        {{ if $thor.startDafilesrv | default false -}}
```
richardkchapman: Feels like this should be {{- and not itself indented

dockerfiles/hpcc/templates/thor.yaml
```diff
@@ -0,0 +1,128 @@
...
+        - name: {{ printf "%s-slave-dafilesrv" .name | quote }}
+{{ include "hpcc.utils.addImageAttrs" (dict "root" $ "imagename" "thorslave") | indent 10 }}
+          command: ["/opt/HPCCSystems/bin/dafilesrv"]
```
richardkchapman: images put /opt/HPCCSystems/bin on the path, so better not to put it here I think

dockerfiles/hpcc/values.yaml
```diff
@@ -71,5 +75,10 @@ notroxie:
     Xconfig: |
       hello there
 
-
-    
\ No newline at end of file
+thor:
+  - name: thor1
+    masterport: 20000
```
richardkchapman: Why would we want users to tell us these ports?

system/jlib/jptree.cpp
```diff
@@ -7869,6 +7869,17 @@ jlib_decl StringBuffer & regenerateConfig(StringBuffer &jsonText, IPropertyTree
     return jsonText;
 }
 
+jlib_decl IPropertyTree * loadArgsIntoConfiguration(IPropertyTree *config, const char * * argv)
```
richardkchapman: Not sure if this is compatible with Gavin's recent changes. Also not sure why you are supplying some more command-line args later after the first call ?

dockerfiles/hpcc/templates/thor.yaml
```diff
@@ -0,0 +1,128 @@
...
+    - ReadWriteMany
+  persistentVolumeReclaimPolicy: Retain
+  hostPath:
+    path: {{ printf "%s/hpcc-data" $.Values.global.hostMountPath }}
+    type: DirectoryOrCreate
+---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: {{ printf "%s-pvc" .name | quote }}
```
richardkchapman: I might be inclined to use printf once at top to save the required string into a variable - just for clarity and consistency with how it was done in roxie.yaml

richardkchapman (commented):
Some minor comments. Mostly looks sensible

richardkchapman:
@ghalliday Please review, especially wrt jptree

dockerfiles/hpcc/values.yaml
jakesmith: don't know really, and wondered if should remove,

I think the template should specify ports, even though strictly it doesn't have to, to work, because guides suggest the k8s yaml should defined the ports to make it clear which ports the pods use.

Not sure whether I should hardwire to the defaults, or whether I should allow these port options, but make them optional with defaults.

AttilaVamos:
The Roxie Schedule2 error is not related to this RP.

system/jlib/jptree.cpp
jakesmith: this is used by the slaves. They don't load a configuration file, but instead have the Thor configuration serialized to them, but they still have a least 1 command line parameter, which tell it what the --master is. In bare-metal case there are few other options e.g. --slavenum

dockerfiles/hpcc/templates/thor.yaml
jakesmith: ok will change.

dockerfiles/hpcc/templates/thor.yaml
jakesmith: agree, whitespace removal at end (e.g. newline) caused me a bit of template error debugging confusion

dockerfiles/hpcc/templates/thor.yaml
jakesmith: I've templatized it, and will check chmod 755 works as expected

dockerfiles/hpcc/templates/thor.yaml
jakesmith: will make comment generic when move to _util.tpl template definition.

dockerfiles/hpcc/templates/thor.yaml
jakesmith: yep, will do for this and a few others.

dockerfiles/hpcc/templates/thor.yaml
```diff
@@ -0,0 +1,124 @@
...
+      run: {{ .name | quote }}
+  template:
+    metadata:
+      labels:
+        run: {{ .name | quote }}
+    spec:
+      containers:
+      - name: {{ .name | quote }}
+        ports:
+          - containerPort: {{ $slavePort }}
```
ghalliday: ignorance: What is this containerPort for?

dockerfiles/hpcc/templates/thor.yaml
```diff
@@ -0,0 +1,124 @@
...
+                ]
+{{ include "hpcc.utils.addImageAttrs" (dict "root" $ "imagename" "thorslave") | indent 10 }}
+          ports:
+            - containerPort: {{ $slavePort }}
+{{ include "hpcc.utils.addVolumeMounts" . | indent 10 }}
+          - mountPath: /var/lib/HPCCSystems/hpcc-data
+            name: mythorstorage
+        {{- if $thor.startDafilesrv | default false }}
+        - name: {{ printf "%s-slave-dafilesrv" .name | quote }}
+{{ include "hpcc.utils.addImageAttrs" (dict "root" $ "imagename" "thorslave") | indent 10 }}
```
ghalliday: Should there be a different image for dafilesrv?

dockerfiles/hpcc/templates/thor.yaml
```diff
@@ -0,0 +1,124 @@
...
+    type: local
+spec:
+  storageClassName: manual
+  capacity:
+    storage: {{ $thor.storageSize }}
+  accessModes:
+    - ReadWriteMany
+  persistentVolumeReclaimPolicy: Retain
+  hostPath:
+    path: {{ printf "%s/hpcc-data" $.Values.global.hostMountPath }}
```
ghalliday: | quote? in case it has spaces etc.

dockerfiles/hpcc/templates/thor.yaml
```diff
@@ -0,0 +1,124 @@
+{{ range $thor := $.Values.thor -}}
+{{- $masterPort := $thor.masterport | default 20000 -}}
+{{- $slavePort := $thor.slaveport | default 20100 -}}
+{{- $slaveDeploymentName := printf "%s-slave" .name | quote }}
+{{- $serviceName := printf "%s-svc" .name }}
```
ghalliday: | quote for consistency

thorlcr/slave/thslavemain.cpp
```diff
@@ -123,7 +118,11 @@ static bool RegisterSelf(SocketEndpoint &masterEp)
             assertex(mySlaveNum == configSlaveNum);
 
         globals.setown(createPTree(msg));
-        mergeCmdParams(globals); // cmd line
+
+        /* NB: preserve command line option overrides
+         * Not sure if any cmdline options are actually needed by this stage..
+         */
+        loadArgsIntoConfiguration(globals, cmdArgs);
```
ghalliday: All components (including client tools I suspect) need to call loadConfiguration.  If the slave calls it, I don't think this code is needed.

ghalliday (commented):
A few comments and questions.

dockerfiles/hpcc/templates/thor.yaml
jakesmith: could be, but put here, because it is temporary, we don't want to keep dafilesrv in the loop, but for now, until the 'nas' option/functionality is available, it is necessary so that the system can 'see' the Thor file parts.

dockerfiles/hpcc/templates/thor.yaml
jakesmith: yep, will change.

dockerfiles/hpcc/templates/thor.yaml
jakesmith: has changed a bit in most recent commit, don't quote any of these vars here anymore, because some of them are used in other strings, so quote them when used.

thorlcr/slave/thslavemain.cpp
jakesmith: slaves do not have a configuration file, the master serializes the configuration from the master.
They do however have at least one command line param --master=, and in base metal case a few others.

@ghalliday  Is loadConfiguration appropriate without a configuration file?

dockerfiles/hpcc/templates/thor.yaml
jakesmith: From https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/:
```
List of ports to expose from the container. Exposing a port here gives the system additional information about the network connections a container uses, but is primarily informational. Not specifying a port here DOES NOT prevent that port from being exposed. Any port which is listening on the default "0.0.0.0" address inside a container will be accessible from the network. Cannot be updated.
```

jakesmith:
@ghalliday @richardkchapman - if you could review changes.

dockerfiles/hpcc/templates/thor.yaml
```diff
@@ -30,75 +37,66 @@ spec:
...
-      # This is a bit of a hack, to ensure that the persistent storage mounted for dllserver
-      # is writable. This is not something we would want to do if using anything other than
-      # local storage (which is only sensible on single-node systems).
-      - name: volume-mount-hack
-        image: busybox
-        command: ["sh", "-c", "chown -R 999:1000 /var/lib/HPCCSystems/hpcc-data"]
-        volumeMounts:
-        - mountPath: "/var/lib/HPCCSystems/hpcc-data"
-          name: mythorstorage
+      {{- if $.Values.global.singleNode | default false }}
```
richardkchapman: Leading spaces here probably best removed (they will be trimmed anyway by the {{- )

dockerfiles/hpcc/templates/thor.yaml
```diff
@@ -30,75 +37,66 @@ spec:
...
+{{ include "hpcc.utils.changeMountPerms" (dict "volumeName" "mythorstorage" "volumePath" "/var/lib/HPCCSystems/hpcc-data") | indent 6 }}
       {{- end }}
```
richardkchapman: same here

dockerfiles/hpcc/templates/_util.tpl
```diff
@@ -90,3 +90,22 @@ volumeMounts:
 image: "{{ include "hpcc.utils.imageName" . }}"
 imagePullPolicy: {{ .root.Values.global.image.pullPolicy }}
 {{- end -}}
+
```
richardkchapman: should change eclccserver.yaml to use this too

jakesmith: true, yes, I meant to.. will change

dockerfiles/hpcc/templates/thor.yaml
jakesmith: where can indent without effect, I thought looked clearer, but I'll change

system/jlib/jptree.cpp
ghalliday: @richardkchapman 
In the container world is there any advantage to the slaves not accessing the same configuration files locally?
Also, we now have a global file which is loaded, and likely to be accessed by jlib to get at global information.  How does that fit in?  (And does this suggest that having a separate global file, rather than a section of the component file might not be the best approach?)

richardkchapman: There's no reason slaves should not see the same config file that the master does

I'm not sure it has any impact on global

richardkchapman: (note - there's already precendence for sharing config files- e.g. between agentexec and eclagent, or between eclccserver and the job it launches

jakesmith: Currently thormaster/thorslave is a little different - in that the master loads the config, augements it with extra information, and serializes it to the slaves.

The slaves also need to pass different config options to the command line, in a container setup they only need --master= , in bare-metal they need to know a couple of other values, e.g. --slavenum

The slaves could instead see the same configuration file as the master, load it, and then the master could send only what additional config options it wants/needs to.
NB: I wouldn't be surprised if the properties the master sets now, could or should go away, either because they should already be set or defaulted before the master (or slave) sees them, or they just don't apply to container setup (e.g. @masterBuildTag).

Further changes like that should be left for a separate PR though.

What I needed now is a way to parse command line options (--xx), then slave starts, and to merge in the configuration being serialized from the master.

jakesmith:
@richardkchapman - changed eclccserver.yaml to use template for permissions and changed platform-core to specify uid/gid so know chown is consistent.

@ghalliday - leaving loadArgsIntoConfiguration in slave as it is for now. Can revisit if master->slave configuration serialization changes.

@ghalliday @richardkchapman Please let me know when to squash.

ghalliday (approved):
I think this is ready to merge, but we should create a jira to revisit the configuration as a separate PR.

jakesmith:
Opened JIRA to revisit thormaster->slave config serializatio: https://track.hpccsystems.com/browse/HPCC-23602

PR #13430, opened 2020-03-02
HPCC-23364 Off-node storage

jakesmith:
Whilst continuing to use DFS and published groups, detected
containerized/cloud setup and use pseudo groups populated with
localhost.

Also move pv/pvc to be shared (not per Thor).

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dockerfiles/hpcc/templates/dllserver.yaml
+++ b/dockerfiles/hpcc/templates/dllserver.yaml
@@ -7,9 +7,9 @@ metadata:
 spec:
   storageClassName: manual
   capacity:
-    storage: 10Gi
+    storage: {{ .Values.global.dllserver.storageSize }}
   accessModes:
     - ReadWriteMany
   hostPath:
-    path: {{ printf "%s/queries" $.Values.global.hostMountPath }}
+    path: {{ printf "%s/queries" $.Values.global.dataStorage.hostMountPath }}
     type: DirectoryOrCreate
--- a/dockerfiles/hpcc/templates/roxie.yaml
+++ b/dockerfiles/hpcc/templates/roxie.yaml
@@ -64,6 +64,9 @@ spec:
       labels:
         run: {{ $servername | quote }}
     spec:
+      {{- if $.Values.global.singleNode | default false }}
+{{ include "hpcc.utils.changeMountPerms" (dict "volumeName" "datastorage-pv" "volumePath" "/var/lib/HPCCSystems/hpcc-data") | indent 6 }}
+      {{- end }}
       containers:
       - name: {{ $servername | quote }}
         args: [
@@ -72,7 +75,8 @@ spec:
                 "--numChannels={{ $numChannels }}",
                 "--serverPorts={{ template "hpcc.utils.portListToCommas" $roxie.ports }}", 
                 "--topologyServers={{ $toponame }}:{{ $roxie.topoport }}",
-                "--localSlave=false"
+                "--localSlave=false", 
+                "--resolveLocally=false"
               ]
 {{ include "hpcc.utils.addImageAttrs" (dict "root" $ "me" $roxie "imagename" "roxie") | indent 8 }}
 {{ include "hpcc.utils.addVolumeMounts" . | indent 8 }}
--- a/dockerfiles/hpcc/values.yaml
+++ b/dockerfiles/hpcc/values.yaml
@@ -5,6 +5,12 @@ global:
   
   dllserver:
     storageSize: 3Gi
+  
+  dataStorage:
+    storageSize: 1G
+    # For singleMode only. Host mount point root for persistent storage.
+    hostMountPath: "/docker-host-mnt"
+    startDafilesrv: false # Temporary until non attached storage functionality added
 
   image:
     # This should probably say "latest" or "stable" in the example script, but set to a specific version in production systems
@@ -18,8 +24,6 @@ global:
   
   singleNode: true
 
-  # For singleMode only. Host mount point root for persistent storage.
-  hostMountPath: "/docker-host-mnt"
 
 
 dali:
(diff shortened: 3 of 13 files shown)


jakesmith:

NB: I will open a separate JIRA to investigate what changes might be needed for DFU.

dockerfiles/hpcc/values.yaml
```diff
@@ -5,6 +5,12 @@ global:
   
   dllserver:
     storageSize: 3Gi
+  
+  dataStorage:
+    storageSize: 100G
```
richardkchapman: This may be too large for some minikube systems

dockerfiles/hpcc/templates/roxie.yaml
```diff
@@ -72,7 +75,8 @@ spec:
                 "--numChannels={{ $numChannels }}",
                 "--serverPorts={{ template "hpcc.utils.portListToCommas" $roxie.ports }}", 
                 "--topologyServers={{ $toponame }}:{{ $roxie.topoport }}",
-                "--localSlave=false"
+                "--localSlave=false", 
+                "--resolveLocally=false"
```
richardkchapman: Need a similar change in localroxie.yaml

thorlcr/mfilemanager/thmfilemanager.cpp
```diff
@@ -374,6 +381,19 @@ class CFileManager : public CSimpleInterface, implements IThorFileManager
 
     IFileDescriptor *create(CJobBase &job, const char *logicalName, StringArray &groupNames, IArrayOf<IGroup> &groups, bool overwriteok, unsigned helperFlags=0, bool nonLocalIndex=false, unsigned restrictedWidth=0)
     {
+        if (isCloud())
```
richardkchapman: Would #ifdef _CONTAINERIZED be more sensible?

richardkchapman:
@jakesmith A few minor comments

thorlcr/mfilemanager/thmfilemanager.cpp
jakesmith: I think I'll leave this as is for now.
It should really be a runtime isNas() I think, where normally true if _CONTAINERIZED, but in future if we support bare metal, me may want attached storage so same container with isNas()==false.

dockerfiles/hpcc/templates/dllserver.yaml
```diff
@@ -7,9 +7,9 @@ metadata:
 spec:
   storageClassName: manual
   capacity:
-    storage: 10Gi
+    storage: {{ .Values.global.dllserver.storageSize }}
```
jakesmith: @richardkchapman - not directly related - but fixed.

dockerfiles/hpcc/templates/roxie.yaml
jakesmith: Added

jakesmith:
@richardkchapman have made the small changes above and squashed in.

jakesmith:
@richardkchapman - meant to say, I looked at removing storageSize storage.capacity, but it's required afaics:
```
Capacity
Generally, a PV will have a specific storage capacity. 
```
from : https://kubernetes.io/docs/concepts/storage/persistent-volumes/


Please review/merge.

dockerfiles/hpcc/templates/dllserver.yaml
richardkchapman: Not 100% sure this is right. There's a subtle distinction between the volume size when setting up a local mount and the size of a pv request. Do we always want them to match?

richardkchapman:
From the link you mention, I see:

Writing Portable Configuration

If you’re writing configuration templates or examples that run on a wide range of clusters and need persistent storage, it is recommended that you use the following pattern:
Include PersistentVolumeClaim objects in your bundle of config (alongside Deployments, ConfigMaps, etc).
Do not include PersistentVolume objects in the config, since the user instantiating the config may not have permission to create PersistentVolumes.
Give the user the option of providing a storage class name when instantiating the template.
If the user provides a storage class name, put that value into the persistentVolumeClaim.storageClassName field. This will cause the PVC to match the right storage class if the cluster has StorageClasses enabled by the admin.
If the user does not provide a storage class name, leave the persistentVolumeClaim.storageClassName field as nil. This will cause a PV to be automatically provisioned for the user with the default StorageClass in the cluster. Many cluster environments have a default StorageClass installed, or administrators can create their own default StorageClass.
In your tooling, watch for PVCs that are not getting bound after some time and surface this to the user, as this may indicate that the cluster has no dynamic storage support (in which case the user should create a matching PV) or the cluster has no storage system (in which case the user cannot deploy config requiring PVCs).

which looks like it could be relevant to us. Probably a topic for another Jira though

dockerfiles/hpcc/templates/dllserver.yaml
```diff
@@ -7,9 +7,9 @@ metadata:
...
   accessModes:
     - ReadWriteMany
   hostPath:
-    path: {{ printf "%s/queries" $.Values.global.hostMountPath }}
+    path: {{ printf "%s/queries" $.Values.global.dataStorage.hostMountPath }}
```
richardkchapman: Dubious. dataStorage and dllServer are separate global sections, referring to separate storage needs. Should this say dllserver.hostMountPath?

dockerfiles/hpcc/values.yaml
```diff
@@ -5,6 +5,12 @@ global:
   
   dllserver:
     storageSize: 3Gi
+  
+  dataStorage:
+    storageSize: 1G
```
richardkchapman: Let's be consistent about Gi vs G

richardkchapman:
@jakesmith Some minor quivvles/questions, which would not stop me merging but which I'd like to talk about first

PR #13482, opened 2020-03-10
HPCC-23691 Allow eclserver to listen on certain queues only

richardkchapman:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/ecl/eclccserver/eclccserver.cpp
+++ b/ecl/eclccserver/eclccserver.cpp
@@ -33,6 +33,8 @@
 #ifndef _CONTAINERIZED
 #include "dalienv.hpp"
 #endif
+#include <unordered_map>
+#include <string>
 
 static Owned<IPropertyTree> globals;
 static const char * * globalArgv = nullptr;
@@ -526,6 +528,7 @@ public:
             if (strcmp(newClusterName, clusterName.str()) != 0)
             {
 #ifdef _CONTAINERIZED
+                // MORE?
 #else
                 clusterInfo.setown(getTargetClusterInfo(newClusterName));
                 if (!clusterInfo)
@@ -817,7 +820,6 @@ int main(int argc, const char *argv[])
         OERRLOG("Failed to load configuration");
         return 1;
     }
-
     const char * processName = globals->queryProp("@name");
     setStatisticsComponentName(SCTeclcc, processName, true);
     if (globals->getPropBool("@enableSysLog",true))
@@ -854,14 +856,31 @@ int main(int argc, const char *argv[])
             if (optMonitorInterval)
                 startPerformanceMonitor(optMonitorInterval*1000, PerfMonStandard, nullptr);
 #ifdef _CONTAINERIZED
+            bool filtered = false;
+            std::unordered_map<std::string, bool> listenQueues;
+            Owned<IPTreeIterator> listening = globals->getElements("listen");
+            ForEach (*listening)
+            {
+                const char *lq = listening->query().queryProp(".");
+                if (lq)
+                {
+                    listenQueues[lq] = true;
+                    filtered = true;
+                }
+            }
+
             StringBuffer queueNames;
             Owned<IPTreeIterator> queues = globals->getElements("queues");
             ForEach(*queues)
             {
                 IPTree &queue = queues->query();
-                if (queueNames.length())
-                    queueNames.append(",");
-                getClusterEclCCServerQueueName(queueNames, queue.queryProp("@name"));
+                const char *qname = queue.queryProp("@name");
+                if (!filtered || listenQueues.count(qname))
+                {
+                    if (queueNames.length())
+                        queueNames.append(",");
+                    getClusterEclCCServerQueueName(queueNames, qname);
+                }
             }
 #else
             SCMStringBuffer queueNames;
--- a/dockerfiles/eclccserver/Dockerfile
+++ b/dockerfiles/eclccserver/Dockerfile
@@ -23,5 +23,4 @@ FROM hpccsystems/platform-core:${BUILD_LABEL}
 USER hpcc
 RUN mkdir -p /var/lib/HPCCSystems/eclccserver
 WORKDIR /var/lib/HPCCSystems/eclccserver
-RUN eclccserver --init > eclccserver.yaml
-ENTRYPOINT [ "eclccserver", "--config=eclccserver.yaml" ]
+ENTRYPOINT [ "eclccserver" ]


richardkchapman:
@ghalliday Please review

@afishbeck This causes output of yaml (e.g. using --outputconfig) to not match input - I tried to work out how to handle on output side but failed.

AttilaVamos:
I think the Wutool error is related to this JIRA issue: https://track.hpccsystems.com/browse/HPCC-23688

AttilaVamos:
@richardkchapman is this error
```
test: JlibIPTTest::testMarkup (F) line: 1403 /home/centos/smoketest/PR-13482/HPCC-Platform/testing/unittests/jlibtests.cpp
- Expression: areMatchingPTrees(xml, yaml)
```
can be related to your changes?

afishbeck:
@richardkchapman I may be missing something, but bear with me.

This is a very different ptree model for scalar arrays and I'm wondering what the benefits are?

One benefit I can see is you can always tell that something was an array rather than a single scalar, which may be what you meant about read and write not being the same.  But most places can treat "[xyz]" and "xyz" as equivalent.

The benefit of the existing model is it works well for using typical access paths.  PTree xpath, or dot notation.  For example xpath="listen[1]" would work for what is there now but for what you've added I think would be xpath="listen/_[1]".  

But toYAML in the current code would write single entry arrays without the [].  
In XPATH and other type paths, an potential most document models, that is probably still ok because "listen[1]" would still work.  

In terms of pure markup it looks different but could be considered equivalent:

```
eclccserver:
   listen: [xyz]

is written by toYAML as

eclccserver:
   listen: xyz

eclccserver:
   listen: [xyz, abc]

unless there is a bug should still be written as

eclccserver:
   listen: [xyz, abc]
```

Can we just consider for single entries that [xyz] and xyz are equivalent?
In most cases that is actually acceptable, but maybe not in your case?

afishbeck:
So with what I have there the line:
Owned<IPTreeIterator> listening = globals->getElements("listen/_");
would just be
Owned<IPTreeIterator> listening = globals->getElements("listen");

richardkchapman:
The problem with the existing model is that Gavin can't tell the difference between an array and a singleton in his config code, and thus decides to assume it's a singleton and merge all the values into one...

Personally I'm fine with your existing structure, and treating [x] and x as the same would actually be good in this situation - but only if we can find some other way to fix the config merging code...

@ghalliday over to you...

afishbeck:
If we do need a special indicator in the ptree then I’d rather it be on the singleton element then on the array.. remember we normally turn scalars into attributes so in reality v: x would have become an attribute.  You would have had to type “v: !el x” to create a real singleton element.  We could find a way of marking that special case in the ptree rather than the common case.  like a special attribute on the singleton element.

afishbeck:
Or an element “_” with a name attribute becomes a true singleton element.  The xpath syntax isn’t quite as elegant but either way the singleton is the special case.

afishbeck:
I think I would change it so:

1. toYaml writes a non special case ptree singleton element as v: [x].
2. A special attribute on the singleton is used when reading and writing  v: !el x.
3. v: x continues being handled as a ptree attribute.

richardkchapman:
I think we may be storing up trouble for ourselves trying to use ptree to represent yaml if it can't actually represent - in a consistent, reversible way - all the yaml we may want it to represent.

Having said I thought it was a feature that a: x and a: [x] could end up the same, I now think I have changed my mind.

I'm not a fan of relying on !el and similar to indicate how we have stored our information internally - it basically means there is some yaml that we cannot generate out of a ptree. Whether that's an issue, I don't know, as we currently only have one situation that we ever want to generate yaml, but...

If we drop the idea that the [ ] being optional on a singleton is a good thing, and that actually we WANT to be able to tell the difference (especially if we are trying to generate yaml for external consumption), I think the argument swings much more in favour of being able to differentiate, and use of _ as I did here is actually a standard way to represent this yaml as xml (see link I provided).

At present, we are in full control of what yaml we see, and what code reads it. We have not yet encountered a case in trying to translate existing configuration file schemas where the representation of repeated xml elements without any attributes was an issue - this case is for a new config setting so we COULD choose to represent it whatever way we want (or indeed to change the yaml to something taht we can represent less ambiguously). But the fact that we are so early in our yaml journey means we have a chance to get it right (if only we can agree what "right" is)

afishbeck:
Note that even now when I said x and [x] are equivalent I meant internally for the ptree/xpath.

In yaml if someone types v: x instead of v: [x] then internally you probably won’t find the value unless you look for either an attribute or element.

richardkchapman:
A summary of (some of) my view:

1. toYaml of yaml that was read into a ptree from yaml should create functionally identical yaml to what was input
2. A way of designing our yaml to be compatible with existing xpaths is desirable but not absolutely essential - we have the means to translate existing XML config files in legacy case to a new structure.
3. If two functionally-different yaml files generate the same ptree representation, then step 1 is unachieveable
4. Going forward, the Yaml representation of our config is the one that matters. So we need to be careful that we don't lock ourselves in to decisions now that will hurt

richardkchapman:
I'm ok with saying that if someone types v:x when we expected v:[x] that it doesn't work.

ideally in the fullness of time json schemas will detect and report such errors nice and early...

afishbeck:
If toYaml always wrote a singleton element as v: [x] and an attribute as v: x then I can’t at the moment think of when we ever really need v: !el x.

richardkchapman:
Never needing !el is good ...

But where does that leave us with the representation of v: [x,y,z] ?

richardkchapman:
Incidentally I have been playing with helm json schemas, and I have a file that will give an error if someone supplies listen: x instead of listen: [x]

ghalliday:
There are two slightly different questions.
1) Can we represent the YAML in the ptree and regenerate it.
It would be possible to do (1) by extending the ptree code - by indicating when a node was added whether it was an array element or not (maybe addPropTree is always an array element, and setPropTree is a signleton).  There would also need to be a way to specify an empty list.

2) Can we generate xml from the ptree and read it back in as equivalent?
I don't think we need this at all.  If the information is in the ptree then it would be possible to add a flag to toXML to translate, but there would also need to be something equivalent on reading.

afishbeck:
I don’t quite see the issue with the current representation of v: [x,y,z]

ghalliday:
The two issues are
i) What is the representation of v: [ ]
ii) how do you distinguish v: [x] and v: x
They matter if you are merging 
```
queue:
  a: hello
```
and
```
queue:
  b: there
```
you would expect
```
queue:
  a: hello
  b: there
```
But if you are merging
```
- queue:
    a: 'hello'
```
and
```
- queue:
    b: 'there'
```
You would either expect
```
- queue:
    a: 'hello'
- queue:
    b: 'there'
```
or
```
- queue:
    b: 'there'
```
i.e. appending list items or replacing them, but not merging them.

richardkchapman:
The trouble with your suggestion (if I've understood it) is that once in a ptree, there's no notion that it is representing yaml - it's only when you read or write that you know that...

afishbeck:
Ok yes for scalars attributes are a good way to represent singletons.

For child trees we still need a way to distinguish/ flag them internally.

afishbeck:
If it were a ptree level feature it would make json support better also.  Right now quantity is the only way to distinguish arrays from singletons.  Luckily JavaScript and other clients don’t usually care.

afishbeck:
What about a new ptree feature setPropTreeUnique which would flag it and if not unique would fail.  Any following append of the same name would also fail, unless forced which would clear the flag.

The ptree iterator would need to let you check the singleton flag.

afishbeck:
For this PR I think you can just treat scalar elements as array items that need to be merged and attributes as singletons.

Then open a jira about the bigger question?  I can try to keep the concise form but could switch to the nested array form if necessary.

richardkchapman:
@gavin had reasons why we could not (I forget what they were)

afishbeck:
So a possible implementation based on what already exists.

Purely from reading PTree code (so could be off) but the second time you call addPropertyTree for the same name, it looks like PTree creates an array object to hold the values.
So theoretically, if we:
1. add a way to use an array for even just 1 entry... "addPropertyTreeItem()" would always make sure an CTArray was used.
2. When in a YAML sequence or JSON array we call addPropertyTreeItem rather than addPropertyTree() or setPropertyTree().
3. we provide a way to check if a tag name references a CTArray.
4. check that when merging to determine if we treat it as a singleton or an array.
5. use it in toYAML and toJSON to properly output the [] even for single items.
6. may need to check what happens when an item is removed going from 2+ down to 1 and figure out how to stay an array.  Perhaps a flag on the array.

ghalliday:
@afishbeck I think the only other issue is 
7. a way of indicating a value that represents an empty list.
I think that is less important though.

richardkchapman:
So, @afishbeck are you going to make some jptree changes, and I'll revisit this PR after they are done?

afishbeck:
@richardkchapman yes I can start working on it now.

jakesmith:
```
1. add a way to use an array for even just 1 entry... "addPropertyTreeItem()" would always make sure an CTArray was used.
2. When in a YAML sequence or JSON array we call addPropertyTreeItem rather than addPropertyTree() or setPropertyTree().
```

For simplicity, would there be any downside in always creating the internal CPTArray, i.e. no need to new addPropertyTreeItem, but the 1st addPropertyTree (or setPropertyTree) would create an internal array.
I don't think that would effect the use of the IPT interface.
Would that effect anything in XML or JSON output?

All of this assumes that we want singletons merged rather than replaced, i.e. we want the former in Gavin's 'queue' example and not the latter.
Is that okay, or do we also need to represent replace in some way? And/or should that be functionality of the merge call, i.e. when merging you state if you want to merge to array, or replace, or perhaps need to list specific properties which override the default.

ghalliday:
I think there would be a downside - you still can't tell the difference between a YAML list and a nested item, you have only changed the default.  If you don't differentiate I think you would be better off leaving it as it is.

richardkchapman:
@ghalliday Please review

richardkchapman:
Not been able to test this yet due to some issues in tagging action

dockerfiles/hpcc/values.yaml
```diff
@@ -66,7 +66,7 @@ eclccserver:
 - name: myeclccserver
   replicas: 1
 ## Specify a list of queues to listen on if you don't want this eclccserver listening on all queues
-# listen: [ thor ]
+  listen: [ thor ]
```
richardkchapman: I should release this commented out (or including all ) when not testing

ecl/eclccserver/eclccserver.cpp
```diff
@@ -854,14 +856,31 @@ int main(int argc, const char *argv[])
             if (optMonitorInterval)
                 startPerformanceMonitor(optMonitorInterval*1000, PerfMonStandard, nullptr);
 #ifdef _CONTAINERIZED
+            bool filtered = false;
+            std::unordered_map<std::string, bool> listenQueues;
+            Owned<IPTreeIterator> listening = globals->getElements("listen");
```
ghalliday: Should there be a comment in a yaml file to indicate how these can be specified?

ghalliday:
@richardkchapman looks ok.  One comment.

ecl/eclccserver/eclccserver.cpp
richardkchapman: There is, in values.yaml (I guess it got merged in a previous PR)

richardkchapman: Looks like it was added in

HPCC-23706 Add json schema to helm chart

PR #13497, opened 2020-03-13
HPCC-23479 create proxies/*.eclmod from LN plugins

xwang2713:
Install *.eclmod based on cmake options 
 *eclmod files, including previous installed under plugins, will be installed under:
USE_PROXIES=ON (default):   plugins/
USE_EEPROXIES=ON (default OFF): plugins/eeproxies
USE_ECLCCPROXIES=ON (default OFF):  bin/eeproxies

*eclmod files will only be installed to one place. If multiple above options set to ON the priority order is USE_ECLCCPROXIES, USE_EEPROXIES, USE_PROXIES
Also lib_moxiekeylib.eclmod is not created since there is plugin/proxie/key_lib.ecllib has the same content.
Previously add lib_date.eclmod from LN/plugins/ is removed since it is irrelavent.

Signed-off-by: Xiaoming Wang <xiaoming.wang@lexisnexis.com>

@ghalliday please help review

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

## Testing:

Tested on local system with following scenarios:
Platform:  -DUSE_EEPROXIES=ON and -DUSE_ECLCCPROXIES=ON
Clienttools: -DUSE_ECLCCPROXIES=ON


cmake_modules/commonSetup.cmake
```diff
@@ -126,6 +126,10 @@ IF ("${COMMONSETUP_DONE}" STREQUAL "")
   option(WSSQL_SERVICE "Configure use of ws_sql service" ON)
   option(USE_DIGISIGN "Use digisign" ON)
   set(CUSTOM_PACKAGE_SUFFIX "" CACHE STRING "Custom package suffix to differentiate development builds")
+  option(USE_PROXIES  "Install default eclmod files" ON)
```
ghalliday: I don't think this is quite what is needed - I don't think we want to configure the target directory.
@GordonSmith I think we need your input.
What we want is for the installer for the client tools to have an option whether to install the proxies from the eeproxies directory to the plugins directory.  I think you probably want (the same?) option which controls whether the plugins are installed as part of make install.

GordonSmith: @ghalliday Is there any reason to not always install them?

ghalliday: If it wasn't for lib_date, then I would say no reason not to.  @richardkchapman maybe we should say that lib_date is not supported in 7.8.0 and following.

GordonSmith: Include it and embed some deprecated warnings for 7.8?

ghalliday: It is ancient code, which is unsupported, and will potentially pollute the global name space.  I would vote for it never seeing the light of day.

GordonSmith: In which case, don't include it and add some instructions to the red book to manually add it if really needed?

ghalliday: It may well already exist in the gitlab repo.  I'll try and check.

xwang2713: Let see if I understand correctly:
1) I still keep old proxies under plugins/proxies and new ones  under plugins/eeproxies/
2) For Platform build both old and new proxies will be installed to plugins/
3) For Clienttools old proxies will be installed to plugins and USE_EEPROXIES will decide if install new proxies (eeproxies)
4) not include lib_date in platform plugins/

ghalliday: It looks like there are a  reasonable number of references to lib_date in the exiting ECL code base.

xwang2713: OK. I will move it to platform plugins or plugins/eeproxies/

ghalliday:
@xwang2713 
Apologies for the delay in giving a definite answer.  I discussed with Richard again.  Our conclusion is that we should still build two versions of the client tools, but they are both built from the open source repo.

That means I think this PR is almost ready.  All it needs is a new option (INCLUDE_EE_PLUGINS or something similar).  This should default to false, and should be used to conditionally include the ee_proxies in the cmake build or not.  

We will then have a standard client build which does not set the option which is published to the portal etc., and another EE client build which does set the option, and which we publish internally (although still open source).

Please squash and rebase to target 7.8.0.  Thanks.

xwang2713:
OK. I will make the change
Also please help checking the LN one. I guess that one need be merged to make this work.

xwang2713:
@ghalliday 
1)  add INCLUDE_EE_PLUGINS option
2) only install EE eclmod files either it is PLATFORM or INCLUDE_EE_PLUGINS turn on (CLIENTTOOLS_ONLY).
3) rebase against candidate-7.8.0

xwang2713:
wait ...

xwang2713:
Fix cmake IF for build error.  rebase with candidate-7.8.0

cmake_modules/commonSetup.cmake
```diff
@@ -487,6 +488,11 @@ IF ("${COMMONSETUP_DONE}" STREQUAL "")
     endif ()
   endif ()
 
+  if (CMAKE_COMPILER_IS_GNUCXX)
```
ghalliday: This change has been included by mistake

plugins/CMakeLists.txt
```diff
@@ -41,6 +41,9 @@ add_subdirectory (exampleplugin)
 add_subdirectory (couchbase)
 add_subdirectory (sqs)
 add_subdirectory (spark)
+IF ( PLATFORM OR INCLUDE_EE_PLUGINS )
```
ghalliday: It should only be checking INCLUDE_EE_PLUGINS.  Standard platform builds will not include these either.

ghalliday (commented):
@xwang2713 looks very close.  Please fix those two issues, and then squash all the commits.

xwang2713:
OK

richardkchapman:
@xwang2713 Needs squashing

richardkchapman:
Merged and squashed manually

PR #13500, opened 2020-03-13
HPCC-23644 Do not log to component log files by default in container …

rpastrana:
…mode

- Conditionally removes creation of component log file logic
- Logging to continue on default stderr stream log handler
- Per container customizations are applied to stderr log handler
- Clients and test code continues to log to file

Signed-off-by: Pastrana, Rodrigo <rodrigo.pastrana@lexisnexisrisk.com>

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:


thorlcr/master/thmastermain.cpp
```diff
@@ -636,14 +636,21 @@ int main( int argc, const char *argv[]  )
...
             lf->setCreateAliasFile(false);
             logHandler = lf->beginLogging();
             createUNCFilename(lf->queryLogFileSpec(), logUrl, false);
+
+            LOG(MCdebugProgress, thorJob, "Opened log file %s", logUrl.str());
         }
-        LOG(MCdebugProgress, thorJob, "Opened log file %s", logUrl.str());
+#else
+        logHandler = queryStderrLogMsgHandler();
+        logUrl.set("stderr");
```
rpastrana: @jakesmith please comment on this change, the logUrl is written to "globals[@logURL]" but it's unclear how that value is ultimately consumed and the generic "stderr" might not be appropriate.
Similar code applies to thslavemain below. 

BTW, in this change, we're removing the use of per component log file in container mode.

jakesmith: @logUrl isn't used anywhere, must be legacy.
So feel free to delete it.

rpastrana:
@JamesDeFabia, this changes current bare-metal behavior,and might need to be documented somewhere in our containerized docs. HPCC components will no longer output logs to /var/log/HPCC/myComponent, but will continue to stream them to stderr. Which is the recommended pattern for app logging in kube/docker. Should this be logged in a documentation jira?

rpastrana:
@jakesmith please review.
As discussed off line, please note some clients will continue to log to file since their behavior cli params explicitly dictate logging to file. Also, dafilerv audit log continues to log to file, and thmaster and slave pass down the stderr log handler to workunit.

rpastrana:
@jakesmith bump

dali/dfu/dfuserver.cpp
```diff
@@ -149,9 +149,15 @@ int main(int argc, const char *argv[])
...
         removeSentinelFile(sentinelFile);
+        int maxdetail = 1000;
+#ifndef _CONTAINERIZED
         Owned<IComponentLogFileCreator> lf = createComponentLogFileCreator(globals, "dfuserver");
-        lf->setMaxDetail(1000);
+        lf->setMaxDetail(maxdetail);
         fileMsgHandler = lf->beginLogging();
+#else
+        Owned<ILogMsgFilter> filter = getCategoryLogMsgFilter(MSGAUD_all, MSGCLS_all, maxdetail);
+        queryLogMsgManager()->changeMonitorFilter(queryStderrLogMsgHandler(), filter);
```
jakesmith: When createComponentLogFileCreator is used, it automatically picks up some defaults, e.g. "logfields" (currently from environment.conf).

It would be good to preserve the ability to customize.
In container world you could instead pick up properties that are relevant to logging from the component configuration.

And get with something like:
```
queryComponentConfig().queryProp("log/logfields")
```

the logging level could also be there (may want to default to values from global: section)

Suspect this is a common enough pattern to put these lines in helper function that does it all this, i.e. so each component normally just calls e.g. startComponentLogging();

dali/sasha/saserver.cpp
```diff
@@ -322,6 +323,9 @@ int main(int argc, const char* argv[])
         lf->setMaxDetail(TopDetail);
         lf->beginLogging();
     }
+#else
+        PROGLOG("Saserver NOT logging to local file!");
```
jakesmith: should probably call the same common helper routine e.g. startComponentLogging()

dali/server/daserver.cpp
```diff
@@ -398,14 +398,17 @@ int main(int argc, char* argv[])
...
         ILogMsgHandler * fileMsgHandler;
         {
             Owned<IComponentLogFileCreator> lf = createComponentLogFileCreator(serverConfig, "dali");
             lf->setLogDirSubdir("server");//add to tail of config log dir
             lf->setName("DaServer");//override default filename
             fileMsgHandler = lf->beginLogging();
         }
-
+#else
+        PROGLOG("Daserver NOT logging to local file!");
```
jakesmith: all standard components should probably call same common helper routine e.g. startComponentLogging()

dali/server/daserver.cpp
```diff
@@ -587,8 +590,9 @@ int main(int argc, char* argv[])
         Owned<IMPServer> mpServer = getMPServer();
         Owned<IWhiteListHandler> whiteListHandler = createWhiteListHandler(populateWhiteListFromEnvironment, formatDaliRole);
         mpServer->installWhiteListCallback(whiteListHandler);
-
+#ifndef _CONTAINERIZED
         setMsgLevel(fileMsgHandler, serverConfig->getPropInt("SDS/@msgLevel", 100));
```
jakesmith: in container world, can make this property legacy, and if using standard routine, pick up a common property e.g. logging/loggingLevel

roxie/topo/toposerver.cpp
```diff
@@ -331,6 +333,9 @@ int main(int argc, const char *argv[])
             queryLogMsgManager()->enterQueueingMode();
             queryLogMsgManager()->setQueueDroppingLimit(512, 32);
         }
+#else
+        queryStderrLogMsgHandler()->setMessageFields(MSGFIELD_time | MSGFIELD_milliTime | MSGFIELD_thread | MSGFIELD_prefix);
```
jakesmith: if using a standard logging routine, this can be acheived by having a different "logfields" default in the components defaultYaml.

thorlcr/master/thmastermain.cpp
jakesmith: Separate issue, but currently Thor (and other components) add the IP + path to the workunit via workunit::addProcess.
That's going to need to change.

jakesmith (commented):
@rpastrana - makes sense. Have added some comments/suggestions:

I think it would be good to standardize on a single helper function that all platform container components call, such that they pick up loggiing preferences from their configuration files.

I'd leave the admin utilities, unittests, other CLI's untouched for now, they are likely to change and/or possibly be removed from standard containers altogether.

dali/dfu/dfuserver.cpp
rpastrana: @jakesmith I agree, and I like the idea. I didn't find queryComponentConfig in our code base. How are the other components accessing configuration properties?

dali/server/daserver.cpp
rpastrana: I think jlog's MODULE_INIT is essentially doing everything the startComponentLogging() would do. What are you envisioning it would do?

dali/dfu/dfuserver.cpp
jakesmith: @rpastrana - queryComponentConfig is in jptree.hpp.

As long as components have used loadConfiguration, they can/do use queryComponentConfig.
All components should call loadConfiguration, I think all the ones we have in containers do.

rpastrana:
@jakesmith please review (Note: I'm still testing some containerized code paths, but let's discuss the overall approach). Also, the default stderr logging handler is created in InitModuleObjs.

dali/datest/dfuwutest.cpp
```diff
@@ -944,7 +944,6 @@ int main(int argc, char* argv[])
         CDfsLogicalFileName dlfn;
         dlfn.setValidate("foreign::10.173.28.12:7070::thor_data400::in::uccv2::20061115::nyc::party",true);
 
-
```
jakesmith: super trivial: but unnecessary file touched.

system/jlib/jlog.cpp
```diff
@@ -2324,6 +2331,53 @@ MODULE_EXIT()
...
+void updateContainerizedLogMsgHandlerLogFields(unsigned fields)
+{
+	theStderrHandler->setMessageFields(fields);
+}
+
+void updateContainerizedLogMsgHandlerFromConfig()
+{
+	if (queryComponentConfig().hasProp("logging"))
+	{
+		updateContainerizedLogMsgHandler(&queryComponentConfig());
```
jakesmith: formatting: indentation looks off, tabs vs spaces?

system/jlib/jlog.cpp
```diff
@@ -2324,6 +2331,53 @@ MODULE_EXIT()
...
+	{
+		theManager->changeMonitorFilter(theStderrHandler, newFilter);
+	}
+}
+
+void updateContainerizedLogMsgHandler(IPropertyTree * tree)
+{
+	if (tree)
+	{
+		if (tree->hasProp("logging/logfields"))
```
jakesmith: We probably want this as @logFields, to be consistent with our config files/yaml.
Other properties below too should be attributes (lower case yaml scalar values imply attributes), and camelCased.

system/jlib/jlog.hpp
```diff
@@ -756,6 +756,11 @@ extern jlib_decl const LogMsgJobInfo unknownJob;
 
 extern jlib_decl ILogMsgManager * queryLogMsgManager();
 extern jlib_decl ILogMsgHandler * queryStderrLogMsgHandler();
+extern jlib_decl void updateContainerizedLogMsgHandlerDetail(LogMsgDetail logDetail);
+extern jlib_decl void updateContainerizedLogMsgHandlerLogFields(unsigned fields);
+extern jlib_decl void updateContainerizedLogMsgHandler(ILogMsgFilter * newFilter);
+extern jlib_decl void updateContainerizedLogMsgHandler(IPropertyTree * tree);
+extern jlib_decl void updateContainerizedLogMsgHandlerFromConfig();
```
jakesmith: I don't think need all these. It should get all from config and component defaults.
So a single installLogHandlerFromConfig() should be good enough. It is an error if a component config doesn't exist (i.e. queryComponentConfig() will throw an exception).

roxie/topo/toposerver.cpp
```diff
@@ -331,6 +333,9 @@ int main(int argc, const char *argv[])
             queryLogMsgManager()->enterQueueingMode();
             queryLogMsgManager()->setQueueDroppingLimit(512, 32);
         }
+#else
+        updateContainerizedLogMsgHandlerLogFields(MSGFIELD_time | MSGFIELD_milliTime | MSGFIELD_thread | MSGFIELD_prefix);
```
jakesmith: we really want these field types in the config and default (in defaultYaml)
e.g. in this component it can have:
```
    logFields: time, milliTime, thread, prefix
```

in it's defaultYaml - processed by loadConfiguration, and then picked up by the standard install log handler routine.

roxie/ccd/ccdmain.cpp
```diff
@@ -770,10 +770,15 @@ int STARTQUERY_API start_query(int argc, const char *argv[])
             removeLog();
         if (topology->hasProp("@logfile"))
         {
+#ifndef _CONTAINERIZED
             Owned<IComponentLogFileCreator> lf = createComponentLogFileCreator(topology, "roxie");
             lf->setMaxDetail(TopDetail);
             lf->beginLogging();
             logDirectory.set(lf->queryLogDir());
+#else
+            updateContainerizedLogMsgHandler(getCategoryLogMsgFilter(MSGAUD_all, MSGCLS_all, TopDetail));
```
jakesmith: I think should look like:
```
    installContainerLogHandler();
```

and logFields and detail level - should be attributes in defaultYaml (see line 527).

in installContainerLogHandler() it should look at the config (queryComponentConfig()) and pick up properties it needs, but probably should also default to something generic if missing.

services/runagent/frunssh.cpp
```diff
@@ -56,13 +56,17 @@ int main( int argc, char *argv[] )
...
+#ifndef _CONTAINERIZED
         StringBuffer logname;
         splitFilename(argv[0], NULL, NULL, &logname, NULL);
 
         Owned<IComponentLogFileCreator> lf = createComponentLogFileCreator("frunssh");
         lf->setCreateAliasFile(false);
         lf->setMsgFields(MSGFIELD_prefix);
         lf->beginLogging();
+#else
+        updateContainerizedLogMsgHandlerLogFields(MSGFIELD_prefix);
```
jakesmith: it's v. unlikely we're going to want frunssh in containerized world (we don't now).
So I'd leave this file untouched.

system/jlib/jlog.cpp
```diff
@@ -2295,7 +2295,14 @@ MODULE_INIT(INIT_PRIORITY_JLOG)
     thePassAllFilter = new PassAllLogMsgFilter();
     thePassLocalFilter = new PassLocalLogMsgFilter();
     thePassNoneFilter = new PassNoneLogMsgFilter();
-    theStderrHandler = new HandleLogMsgHandlerTable(stderr, MSGFIELD_STANDARD);
+    unsigned  msgFields = MSGFIELD_STANDARD;
+#ifdef _CONTAINERIZED
+    const char *logFields = queryEnvironmentConf().queryProp("logfields");
+    if (!isEmptyString(logFields))
+        msgFields = LogMsgFieldsFromAbbrevs(logFields);
+#endif
```
jakesmith: we shouldn't be environment.conf in _CONTAINERIZED mode, or rather we shouldn't be adding anything new to it.
So prob. don't want this change.
logging/@logFields will be a property picked up from the config queryComponentConfig() when the container log handler is installed.

system/jlib/jlog.cpp
```diff
@@ -2324,6 +2331,53 @@ MODULE_EXIT()
...
 
+void updateContainerizedLogMsgHandlerDetail(LogMsgDetail logDetail)
+{
+    updateContainerizedLogMsgHandler(getCategoryLogMsgFilter(MSGAUD_all, MSGCLS_all, logDetail));
+}
+
+void updateContainerizedLogMsgHandlerLogFields(unsigned fields)
+{
+	theStderrHandler->setMessageFields(fields);
+}
```
jakesmith: don't think you need this.
You can call this when installing the log handler with what you pick up from logging/@logFields, or default if empty.

thorlcr/master/thmastermain.cpp
```diff
@@ -636,16 +636,22 @@ int main( int argc, const char *argv[]  )
...
+        LOG(MCdebugProgress, thorJob, "Logging to Stderr!");
```
jakesmith: installContainerLogHandler() here.

dali/dfu/dfuserver.cpp
```diff
@@ -149,9 +149,14 @@ int main(int argc, const char *argv[])
...
         sentinelFile.setown(createSentinelTarget());
         removeSentinelFile(sentinelFile);
+        int maxdetail = 1000;
+#ifndef _CONTAINERIZED
         Owned<IComponentLogFileCreator> lf = createComponentLogFileCreator(globals, "dfuserver");
-        lf->setMaxDetail(1000);
+        lf->setMaxDetail(maxdetail);
         fileMsgHandler = lf->beginLogging();
+#else
+        updateContainerizedLogMsgHandler(getCategoryLogMsgFilter(MSGAUD_all, MSGCLS_all, maxdetail));
```
jakesmith: This is a component we haven't containerized.. 
In fact it's not so much that we haven't containerized, but that it hasn't been moved over to the new config file paradigm (i.e. it doesn't call loadConfiguration, so queryComponentConfig() would fire an exception).

And so for now for any component like this that hasn't been contaerized (and hasn't moved over to the new config), I would leave it untouched.

dali/dfuxref/dfuxrefmain.cpp
```diff
@@ -50,12 +50,17 @@ int main(int argc, char* argv[])
...
 
     {
+#ifndef _CONTAINERIZED
         Owned<IComponentLogFileCreator> lf = createComponentLogFileCreator("dfuxref");
         lf->setAppend(false);
         lf->setMaxDetail(TopDetail);
         lf->setMsgFields(MSGFIELD_STANDARD);
         lf->beginLogging();
+#else
+        updateContainerizedLogMsgHandler(getCategoryLogMsgFilter(MSGAUD_all, MSGCLS_all, TopDetail));
```
jakesmith: ditto to above, and it's doubtful XREF  ever will be in cloud/containerized env.

dali/fuse/dafuse.cpp
```diff
@@ -1046,6 +1047,9 @@ int main(int argc, char *_argv[])
         ls->setMaxDetail(TopDetail);
         lf->beginLogging();
     }
+#else
+    updateContainerizedLogMsgHandler(getCategoryLogMsgFilter(MSGAUD_all, MSGCLS_all, TopDetail));
```
jakesmith: dafuse is not in use and won't be ported to work in containers. So better left unaltered.

jakesmith:
@rpastrana - added some comments.

I think all components that have been containerized use the new config system,
If all containerized components should call 1 line :
```
   installContainerLogHandler()
```
then they can pick up their specific log requirements (e.g. custom fields, trace level), from the config file.
The defaults (that are currently hardcoded) can go in defaultYaml for that component.
installContainerLogHandler() can pickup logging settings from queryComponentConfig() , defaulting to standard setting if not defined.

rpastrana:
@jakesmith please review

system/jlib/jlog.cpp
```diff
@@ -2324,6 +2324,41 @@ MODULE_EXIT()
...
 
+#ifdef _CONTAINERIZED
+#define LOGFIELDS_ATT "@fields"
+#define LOGMSGDETAIL_ATT "@detail"
+#define LOGMSGAUDIENCES_ATT "@audiences"
+#define LOGMSGCLASSES_ATT "@classes"
+
+void setupContainerizedLogMsgHandlerFromConfig()
+{
+    IPropertyTree * logconfig = queryComponentConfig().queryPropTree("logging");
```
jakesmith, line 2335: if "logging" isn't defined, what is the default behaviour?
It still uses the stderr handler and all fields, and what trace level?

Can we define defaults here in the absence of settings, e.g.:
```
    // defaults
    unsigned msgFields = MSGFIELD_time|MSGFIELD_date|MSGFIELD_milliTime|MSGFIELD_msgID|MSGFIELD_process|MSGFIELD_thread|MSGFIELD_code|MSGFIELD_quote|MSGFIELD_prefix|MSGFIELD_audience;
    unsigned msgAudiences = MSGAUD_all;
    unsigned msgClasses = MSGAUD_all;
    LogMsgDetail logDetail = DefaultDetail;
```

then if (logging) and any of the above present, override these and then run the setMessageFields/changeMonitorFilter lines unconditionally.

system/jlib/jlog.cpp
```diff
@@ -2324,6 +2324,41 @@ MODULE_EXIT()
...
+    if (logconfig)
+    {
+        if (logconfig->hasProp(LOGFIELDS_ATT))
+        {
+            //Supported logging fields: AUD,CLS,DET,MID,TIM,DAT,PID,TID,NOD,JOB,USE,SES,COD,MLT,MCT,NNT,COM,QUO,PFX,ALL,STD
+            //default: TIM+DAT+MLT+MID+PID+TID+COD+QUO+PFX+AUD
+            const char *logFields = logconfig->queryProp("logfields");
```
jakesmith, line 2342: think "logfields" here is supposed to be LOGFIELDS_ATT

system/jlib/jlog.hpp
```diff
@@ -756,6 +756,9 @@ extern jlib_decl const LogMsgJobInfo unknownJob;
 
 extern jlib_decl ILogMsgManager * queryLogMsgManager();
 extern jlib_decl ILogMsgHandler * queryStderrLogMsgHandler();
+#ifdef _CONTAINERIZED
+extern jlib_decl void setupContainerizedLogMsgHandlerFromConfig();
```
jakesmith, line 760: since it's always by definition from config, and it's rather long method name is it worth the "FromConfig" ?

dali/sasha/saserver.cpp
```diff
@@ -322,6 +323,9 @@ int main(int argc, const char* argv[])
         lf->setMaxDetail(TopDetail);
         lf->beginLogging();
     }
+#else
+    setupContainerizedLogMsgHandlerFromConfig();
```
jakesmith, line 327: this is fine, just note to myself really, that if this is were called now, it would throw an exception in queryComponentConfig(), because Sasha doesn't use the new init system yet (but it should).
But since Sasha isn't yet containerized either - it will never be called.

It's fine though, since this is what we want when it is, so I'd leave it.

fs/dafilesrv/dafilesrv.cpp
```diff
@@ -788,12 +788,16 @@ int main(int argc,char **argv)
...
     }
+#ifndef _CONTAINERIZED
     {
         Owned<IComponentLogFileCreator> lf = createComponentLogFileCreator(logDir.str(), "DAFILESRV");
         lf->setCreateAliasFile(false);
         lf->setMaxDetail(TopDetail);
         lf->beginLogging();
     }
+#else
+        setupContainerizedLogMsgHandlerFromConfig();
```
jakesmith, line 799: formatting: indentation

jakesmith: Similar to comment in sasha - dafilesrv does not use the new config files yet (doesn't call loadConfiguration), so if this were called it would throw an exception in queryComponentConfig().
But dafilesrv isn't running in any of our containers yet, so it's fine.
When/if it is, we will need to ensure it uses loads a config.

roxie/ccd/ccdmain.cpp
```diff
@@ -770,10 +772,15 @@ int STARTQUERY_API start_query(int argc, const char *argv[])
             removeLog();
         if (topology->hasProp("@logfile"))
         {
+#ifndef _CONTAINERIZED
             Owned<IComponentLogFileCreator> lf = createComponentLogFileCreator(topology, "roxie");
             lf->setMaxDetail(TopDetail);
             lf->beginLogging();
             logDirectory.set(lf->queryLogDir());
+#else
+            setupContainerizedLogMsgHandlerFromConfig();
```
jakesmith, line 781: don't think it makes sense for this to be under if (topology->hasProp("@logfile")) now.
The #ifndef _CONTAINERIZED should be above that.

There are other potentially interesting properties here related to logging in general, like "logQueueLen" and "logQueueDrop".
Should also be properties of the generic logging: config, and picked up in setupContainerizedLogMsgHandlerFromConfig()

thorlcr/master/thmastermain.cpp
```diff
@@ -636,16 +636,23 @@ int main( int argc, const char *argv[]  )
...
             createUNCFilename(lf->queryLogFileSpec(), logUrl, false);
+
+            LOG(MCdebugProgress, thorJob, "Opened log file %s", logUrl.str());
         }
-        LOG(MCdebugProgress, thorJob, "Opened log file %s", logUrl.str());
+#else
+        setupContainerizedLogMsgHandlerFromConfig();
+        logHandler = queryStderrLogMsgHandler();
+        logUrl.set("stderr");
+        LOG(MCdebugProgress, thorJob, "Logging to Stderr!");
```
jakesmith, line 653: not worth keeping this line

jakesmith (changes requested):

jakesmith:
@rpastrana - looks like there's a few components that are missing a call to setupContainerizedLogMsgHandlerFromConfig()

Any component that currently calls loadConfiguration in main() is likely to want to call this.

Anything that doesn't that calls setupContainerizedLogMsgHandlerFromConfig() will fail if started because queryComponentConfig() will throw an exception.

system/jlib/jlog.cpp
rpastrana: We could define defaults already, but they're already defined in void setDefaults()
The logfields are only overwritten if the configuration contains @fields, no need to default.

If either  the audiences, classes, or details are specified, the other fields are defaulted to the same value as setDefaults()

rpastrana:
@jakesmith please review

system/jlib/jlog.cpp
```diff
@@ -2324,6 +2324,77 @@ MODULE_EXIT()
...
+#endif
+
+#define DEFAULT_QUEUELEN 512
+#define DEFAULT_QUEUEDROP 32
+#define DEFAULT_USE_SYSLOG false
+
+void setupContainerizedLogMsgHandler()
+{
+
+    Owned<IPropertyTreeIterator> iter = queryComponentConfig().getElements("*[1]/logging");
```
jakesmith, line 2351: we should only have one logging: section (ideally schema would validate that).
Not sure why you're iterating xpath "*[1]/logging" ?
Would expect it to still be:
```
IPropertyTree *logConfig = queryComponentConfig().queryPropTree("logging");
```

system/jlib/jlog.hpp
```diff
@@ -153,6 +160,20 @@ inline const char * LogMsgAudienceToFixString(LogMsgAudience audience)
         return("UNKNOWN  ");
     }
 }
+inline unsigned LogMsgAudFromAbbrev(char const * abbrev)
```
jakesmith, line 163: trivial: our normal casing for method would be camelCase with lower case start.
i.e. LogMsgFieldsFromAbbrevs() should really have been logMsgFieldsFromAbbrevs()

Would be a good time as you're adding new ones to make them and it cased normally.

system/jlib/jlog.hpp
```diff
@@ -347,39 +385,71 @@ inline unsigned LogMsgFieldFromAbbrev(char const * abbrev)
...
+// This function parses strings such as "DIS+ERR+WRN+INF" and "ALL+PRO-INF"
+inline unsigned LogMsgClassesFromAbbrevs(char const * abbrevs)
+{
+    return processAbbrevsString(abbrevs, LogMsgAttType_class);
+}
+
+// This function parses strings such as "AUD+CLS+DET+COD" and "STD+MIT-PID", and is used for fields attribute in XML handler descriptions
+inline unsigned LogMsgFieldsFromAbbrevs(char const * abbrevs)
+{
+    return processAbbrevsString(abbrevs, LogMsgAttType_fields);
```
jakesmith, line 452: trivial/not new: not sure any of these methods should be in the header, they are used internally to log.
Might be good to move to them (and thew new LogMsgAttType enumerated type) to jlog.cpp

ecl/eclcc/eclcc.cpp
```diff
@@ -503,6 +503,9 @@ static constexpr const char * defaultYaml = R"!!(
 version: "1.0"
 eclccserver:
     name: eclccserver
+    logging:
+      audiences: "USR+ADT"
```
jakesmith, line 507: feels like now in yaml, this should really be:
```
audiences: [ "USR", "ADT" ]
```

fs/dafilesrv/dafilesrv.cpp
```diff
@@ -788,12 +788,16 @@ int main(int argc,char **argv)
...
     }
+#ifndef _CONTAINERIZED
     {
         Owned<IComponentLogFileCreator> lf = createComponentLogFileCreator(logDir.str(), "DAFILESRV");
         lf->setCreateAliasFile(false);
         lf->setMaxDetail(TopDetail);
         lf->beginLogging();
     }
+#else
+        setupContainerizedLogMsgHandler();
```
jakesmith, line 799: formatting: indentation

roxie/topo/toposerver.cpp
```diff
@@ -276,6 +276,8 @@ static constexpr const char * defaultYaml = R"!!(
     topoport: 9004
     stdlog: true
     traceLevel: 1
+    logging:
+      fields: "TIM+MLT+TID+PFX"
```
jakesmith, line 280: we could do it later, but feels now in yaml like this should be:
```
    fields: [ "TIM", "MLT", "TID", "PFX" ]
```

system/jlib/jlog.cpp
```diff
@@ -2324,6 +2324,77 @@ MODULE_EXIT()
...
+
+#ifdef _DEBUG
+  #define DEFAULT_USELOGQUEUE false
+#else
+  #define DEFAULT_USELOGQUEUE true
+#endif
+
+#define DEFAULT_QUEUELEN 512
+#define DEFAULT_QUEUEDROP 32
+#define DEFAULT_USE_SYSLOG false
```
jakesmith, line 2346: trivial: more modern to use e.g.:
```
static constexpr const char *logFieldsAtt = "@fields";
```

(and compiler can validate declaration and possibly optimize some uses)

jakesmith (commented):
@rpastrana - mostly minor and trivial comments.
Main questions are re. use of "*[1]/logging" xpath, and whether this it would make sense to move to a more standard list syntax for the yaml config given these logging attributes are really lists.

system/jlib/jlog.cpp
rpastrana: The logging element is a child of the variable named "component" element, ie esp, roxie, etc.
I believe IPropertyTree *logConfig = queryComponentConfig().queryPropTree("logging"); targets the toplevel element. 
Also worth noting, I believe queryPropTree performs a getElements as well

jakesmith: There should only be 1 "logging:" section per component.
Each component has it's own config file.
queryComponentConfig() will return the component section from that config file.

So for example:
```
    version: 1.0
    dali:
      name: mydali
      storage:
        storageClass: ""
        storageSize: 1Gi
      logging:
         detail: 50
    global:
      imageVersion: "master-2020-02-10"
      singleNode: false
```

(added logging: as an example)

So queryComponentConfig() when running in daserver will return the "dali" section as the root tree returned (this is defined by the call to loadConfiguration()). You will not have access to other sections (e.g. global:) 

logging: would be a property under dali: and generated if present in values.yaml.

jakesmith:
@rpastrana - I think it's fine, can you squash and rebase - and I'll do another test.

rpastrana:
Closing in favor of https://github.com/hpcc-systems/HPCC-Platform/pull/13587

PR #13502, opened 2020-03-13
HPCC-23717 Change dali to use new config

jakesmith:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:


system/jlib/jptree.cpp
```diff
@@ -7928,6 +7920,16 @@ jlib_decl IPropertyTree * loadConfiguration(const char * defaultYaml, const char
...
 
+#ifdef _DEBUG
+    if (config->getPropBool("@hold"))
+    {
+        DBGLOG("Component paused for debugging purposes, attach and set 'held' to release");
+        bool held = true;
+        while (held)
+            Sleep(5);
+    }
+#endif
```
jakesmith: @richardkchapman - moved here so I could use via values.yaml - any objections?

dali/server/daserver.cpp
```diff
@@ -408,154 +415,163 @@ int main(int argc, char* argv[])
...
                 }
+                if (serverConfig->getPropBool("@daliServixCaching", true))
+                    setDaliServixSocketCaching(true);
+            }
+            catch (IException *e)
+            {
+                StringBuffer s("Failure whilst preparing dali backup location: ");
+                LOG(MCoperatorError, unknownJob, e, s.append(mirrorPath).append(". Backup disabled").str());
+                serverConfig->removeProp("SDS/@remoteBackupLocation");
+                e->Release();
```
jakesmith: Much of the above is different because indentation/scope change only.

jakesmith:

system/jlib/jptree.cpp
richardkchapman: It's already supported in jptree via values.yaml - see line 7881 or so.

richardkchapman: Ah - never mind I see what you did. The only downside of moving it here is that it will make it harder to debug a situation where it's crashing in config merging code. 

You could always leave it in both places...

jakesmith: >The only downside of moving it here is that it will make it harder to debug a situation where it's crashing in config merging code.

true, yes in _DEBUG why not. I'll change.

richardkchapman:
@jakesmith This is waiting for a change you promised...

jakesmith:
@richardkchapman - changed --hold so picked up in either place.

PR #13505, opened 2020-03-13
HPCC-23715 Guillotine thor job exceeding cost limit

shamser:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/common/workunit/workunit.cpp
+++ b/common/workunit/workunit.cpp
@@ -2487,8 +2487,75 @@ protected:
     IArrayOf<StatisticAggregator> aggregators;
 };
 
+// Aggregate costs excluding all hThor costs
+// Note: the only reason that this class is required is that it is not possible to determine the creator
+//       of a scope when iterating with IConstWUScopeIterator.  (When this functionlity becomes available, 
+//       consider filtering within aggregateCost and eliminating this class.)
+class CostAggregatorExcludeHThor : public CInterfaceOf<IWuScopeVisitor>
+{
+public:
+    virtual void noteStatistic(StatisticKind kind, unsigned __int64 value, IConstWUStatistic & extra) override
+    {
+        if (extra.getCreatorType() != SCThthor)
+            totalCost += value;
+    }
+    virtual void noteAttribute(WuAttr attr, const char * value) override { throwUnexpected(); }
+    virtual void noteHint(const char * kind, const char * value) override { throwUnexpected(); }
+    virtual void noteException(IConstWUException & exception) { throwUnexpected(); }
+    virtual cost_type getTotalCost() const { return totalCost; }
+protected:
+    cost_type totalCost = 0;
+};
 
-//To calculate aggregates, create a scope iterator, and an instance of a StatisticAggregator, and play the attributes through the interface
+// Aggregrate Thor or hThor costs
+cost_type aggregateCost(const IConstWorkUnit * wu, const char *scope, bool excludeHThor)
+{
+    // depth 1: workflow, depth 2: graph, depth 3: subgraph
+    WuScopeFilter filter;
+    if (scope && *scope) // All costs under specified scope (used to calculate all costs for a workflow)
+    {
+        filter.addScope(scope);
+        filter.setIncludeNesting(3);
+        filter.addOutputStatistic(StCostExecute);
+        filter.addRequiredStat(StCostExecute);
+    }
+    else
+        filter.addFilter("stat[CostExecute],depth[1..3],nested[0],where[CostExecute]");
+    filter.addSource("global");
+    filter.finishedFilter();
+    Owned<IConstWUScopeIterator> it = &wu->getScopeIterator(filter);
+    // Note: use of CostAggregatorExcludeHThor will be a slower, so use only when necessary
+    if (excludeHThor)
+    {
+        CostAggregatorExcludeHThor aggregator;
+        for (it->first(); it->isValid(); )
+        {
+            it->playProperties(aggregator);
+            stat_type value;
+            if (it->getStat(StCostExecute, value))
+                it->nextSibling();
+            else
+                it->next();
+        }
+        return aggregator.getTotalCost();
+    }
+    else
+    {
+        cost_type totalCost = 0;
+        for (it->first(); it->isValid(); )
+        {
+            stat_type value = 0.0;
+            if (it->getStat(StCostExecute, value))
+            {
+                totalCost += value;
+                it->nextSibling();
+            }
+            else
+                it->next();
+        }
+        return totalCost;
+    }
+}
 
 //---------------------------------------------------------------------------------------------------------------------
 
@@ -3883,6 +3950,8 @@ public:
             { return c->getDebugValueInt(propname, defVal); }
     virtual __int64 getDebugValueInt64(const char * propname, __int64 defVal) const
             { return c->getDebugValueInt64(propname, defVal); }
+    virtual double getDebugValueReal(const char * propname, double defVal) const
+            { return c->getDebugValueReal(propname, defVal); }
     virtual bool getDebugValueBool(const char * propname, bool defVal) const
             { return c->getDebugValueBool(propname, defVal); }
     virtual IStringIterator & getDebugValues() const 
@@ -7742,7 +7811,17 @@ __int64 CLocalWorkUnit::getDebugValueInt64(const char *propname, __int64 defVal)
     CriticalBlock block(crit);
     StringBuffer prop("Debug/");
     prop.append(lower);
-    return p->getPropInt64(prop.str(), defVal); 
+    return p->getPropInt64(prop.str(), defVal);
+}
+
+double CLocalWorkUnit::getDebugValueReal(const char *propname, double defVal) const
+{
+    StringBuffer lower;
+    lower.append(propname).toLowerCase();
+    CriticalBlock block(crit);
+    StringBuffer prop("Debug/");
+    prop.append(lower);
+    return p->getPropReal(prop.str(), defVal);
 }
 
 bool CLocalWorkUnit::getDebugValueBool(const char * propname, bool defVal) const
@@ -12987,15 +13066,15 @@ extern WORKUNIT_API void addTimeStamp(IWorkUnit * wu, StatisticScopeType scopeTy
     wu->setStatistic(queryStatisticsComponentType(), queryStatisticsComponentName(), scopeType, scopestr, kind, NULL, getTimeStampNowValue(), 1, 0, StatsMergeAppend);
 }
 
-extern WORKUNIT_API double calculateThorCost(__int64 timeNs, unsigned clusterWidth)
+extern WORKUNIT_API cost_type calculateThorCost(unsigned __int64 ms, unsigned clusterWidth)
 {
     IPropertyTree *costs = queryCostsConfiguration();
     if (costs)
     {
-        double thor_master_rate = costs->getPropReal("thor/@master", 0.0);
-        double thor_slave_rate = costs->getPropReal("thor/@slave", 0.0);
+        cost_type thor_master_rate = money2cost_type(costs->getPropReal("thor/@master"));
+        cost_type thor_slave_rate = money2cost_type(costs->getPropReal("thor/@slave"));
 
-        return calcCost(thor_master_rate, timeNs) + calcCost(thor_slave_rate, timeNs) * clusterWidth;
+        return calcCost(thor_master_rate, ms) + calcCost(thor_slave_rate, ms) * clusterWidth;
     }
     return 0;
 }
--- a/common/workunit/workunit.hpp
+++ b/common/workunit/workunit.hpp
@@ -1174,6 +1174,7 @@ interface IConstWorkUnit : extends IConstWorkUnitInfo
     virtual IStringVal & getDebugValue(const char * propname, IStringVal & str) const = 0;
     virtual int getDebugValueInt(const char * propname, int defVal) const = 0;
     virtual __int64 getDebugValueInt64(const char * propname, __int64 defVal) const = 0;
+    virtual double getDebugValueReal(const char * propname,  double defVal) const = 0;
     virtual bool getDebugValueBool(const char * propname, bool defVal) const = 0;
     virtual IStringIterator & getDebugValues() const = 0;
     virtual IStringIterator & getDebugValues(const char * prop) const = 0;
@@ -1646,6 +1647,7 @@ extern WORKUNIT_API void updateWorkunitStat(IWorkUnit * wu, StatisticScopeType s
 extern WORKUNIT_API void updateWorkunitTimings(IWorkUnit * wu, ITimeReporter *timer);
 extern WORKUNIT_API void updateWorkunitTimings(IWorkUnit * wu, StatisticScopeType scopeType, StatisticKind kind, ITimeReporter *timer);
 extern WORKUNIT_API void aggregateStatistic(StatsAggregation & result, IConstWorkUnit * wu, const WuScopeFilter & filter, StatisticKind search);
+extern WORKUNIT_API cost_type aggregateCost(const IConstWorkUnit * wu, const char *scope=nullptr, bool excludehThor=false);
 
 extern WORKUNIT_API const char *getTargetClusterComponentName(const char *clustname, const char *processType, StringBuffer &name);
 extern WORKUNIT_API void descheduleWorkunit(char const * wuid);
@@ -1658,7 +1660,7 @@ extern WORKUNIT_API const char * getWorkunitActionStr(WUAction action);
 extern WORKUNIT_API WUAction getWorkunitAction(const char * actionStr);
 
 extern WORKUNIT_API void addTimeStamp(IWorkUnit * wu, StatisticScopeType scopeType, const char * scope, StatisticKind kind, unsigned wfid=0);
-extern WORKUNIT_API double calculateThorCost(__int64 timeNs, unsigned clusterWidth);
+extern WORKUNIT_API cost_type calculateThorCost(unsigned __int64 ms, unsigned clusterWidth);
 
 extern WORKUNIT_API IPropertyTree * getWUGraphProgress(const char * wuid, bool readonly);
 
@@ -1687,9 +1689,7 @@ inline bool isGlobalScope(const char * scope) { return scope && (streq(scope, GL
 extern WORKUNIT_API bool isValidPriorityValue(const char * priority);
 extern WORKUNIT_API bool isValidMemoryValue(const char * memoryUnit);
 
-#define HourToSeconds(n) ((n)/3600)
-#define NanoSecondsToSeconds(n) ((double)(n)/1000000000)
-inline __int64 calcCost(double ratePerHour, __int64 timeNS) { return HourToSeconds(ratePerHour) * NanoSecondsToSeconds(timeNS) * 1e6; }
+inline cost_type calcCost(cost_type ratePerHour, unsigned __int64 ms) { return ratePerHour * ms / 1000 / 3600; }
 
 #ifdef _CONTAINERIZED
 extern WORKUNIT_API bool executeGraphOnLingeringThor(IConstWorkUnit &workunit, const char *graphName);
--- a/ecl/eclagent/eclagent.cpp
+++ b/ecl/eclagent/eclagent.cpp
@@ -79,6 +79,7 @@ using roxiemem::OwnedRoxieString;
 #define ABORT_CHECK_INTERVAL 30     // seconds
 #define ABORT_DEADMAN_INTERVAL (60*5)  // seconds
 #define MAX_FILE_READ_FAIL_COUNT 3
+#define CHECK_COST_INTERVAL 120 // How frequently workunit cost is checked against cost limit (seconds)
 
 typedef IEclProcess* (* EclProcessFactory)();
 
@@ -565,7 +566,20 @@ EclAgent::EclAgent(IConstWorkUnit *wu, const char *_wuid, bool _checkVersion, bo
         updateSuppliedXmlParams(w);
     }
     IPropertyTree *costs = queryCostsConfiguration();
-    agentMachineCost = costs ? costs->getPropReal("@agent", 0.0): 0.0;
+    if (costs)
+    {
+        agentMachineCost = money2cost_type(costs->getPropReal("@agent"));
+        if (agentMachineCost)
+        {
+            double softCostLimit = costs->getPropReal("@limit");
+            double guillotineCost = wu->getDebugValueReal("maxCost", softCostLimit);
+            double hardCostLimit = costs->getPropReal("@hardlimit");
+            if (hardCostLimit && ((guillotineCost == 0) || (guillotineCost > hardCostLimit)))
+                guillotineCost = hardCostLimit;
+            abortmonitor->setGuillotineCost(money2cost_type(guillotineCost));
+        }
+    }
+
 }
 
 EclAgent::~EclAgent()
@@ -1920,7 +1934,7 @@ void EclAgent::doProcess()
         const __int64 elapsedNs = elapsedTimer.elapsedNs();
         updateWorkunitStat(w, SSTglobal, NULL, StTimeElapsed, nullptr, elapsedNs);
 
-        const __int64 cost = calcCost(agentMachineCost, elapsedNs);
+        const cost_type cost = aggregateCost(w, nullptr, false);
         if (cost)
             w->setStatistic(queryStatisticsComponentType(), queryStatisticsComponentName(), SSTglobal, "", StCostExecute, NULL, cost, 1, 0, StatsMergeReplace);
 
@@ -2324,7 +2338,7 @@ void EclAgentWorkflowMachine::noteTiming(unsigned wfid, timestamp_type startTime
     updateWorkunitStat(wu, SSTworkflow, scope, StWhenStarted, nullptr, startTime, 0);
     updateWorkunitStat(wu, SSTworkflow, scope, StTimeElapsed, nullptr, elapsedNs, 0);
 
-    const __int64 cost = calcCost(agent.queryAgentMachineCost(), elapsedNs);
+    const cost_type cost = calcCost(agent.queryAgentMachineCost(), nanoToMilli(elapsedNs)) + aggregateCost(wu, scope, true);
     if (cost)
         wu->setStatistic(queryStatisticsComponentType(), queryStatisticsComponentName(), SSTworkflow, scope, StCostExecute, NULL, cost, 1, 0, StatsMergeReplace);
 }
@@ -3159,21 +3173,30 @@ void EclAgent::abortMonitor()
 {
     StringBuffer errorText;
     unsigned guillotineleft = 0;
-    for (;;) {
-        unsigned waittime = ABORT_CHECK_INTERVAL;   
-        if (abortmonitor->guillotinetimeout) {
-            if (guillotineleft==0) {
+    unsigned checkCostInterval = CHECK_COST_INTERVAL;
+    for (;;)
+    {
+        unsigned waittime = ABORT_CHECK_INTERVAL;
+        if (abortmonitor->guillotinetimeout)
+        {
+            if (guillotineleft==0)
+            {
                 guillotineleft = abortmonitor->guillotinetimeout;
                 DBGLOG("Guillotine set to %ds",guillotineleft);
             }
             if (guillotineleft<waittime)
                 waittime = guillotineleft;
         }
-        if (abortmonitor->sem.wait(waittime*1000) && abortmonitor->stopping) {
+        if ((abortmonitor->guillotineCost > 0) && (checkCostInterval < waittime))
+            waittime = checkCostInterval;
+        if (abortmonitor->sem.wait(waittime*1000) && abortmonitor->stopping)
+        {
             return;
         }
-        if (guillotineleft) {
-            if (abortmonitor->guillotinetimeout) {
+        if (guillotineleft)
+        {
+            if (abortmonitor->guillotinetimeout)
+            {
                 guillotineleft-=waittime;
                 if (guillotineleft==0) {
                     DBGLOG("Guillotine triggered");
@@ -3189,13 +3212,31 @@ void EclAgent::abortMonitor()
             if (queryWorkUnit()) // could be exiting
                 isAborting = queryWorkUnit()->aborting();
         }
-        if (isAborting) {
+        if (isAborting)
+        {
             DBGLOG("Abort detected");
             while (abortmonitor->sem.wait(ABORT_DEADMAN_INTERVAL*1000))
                 if (abortmonitor->stopping)
                     return; // stopped in time
             IERRLOG("EclAgent failed to abort within %ds - killing process",ABORT_DEADMAN_INTERVAL);
             break;
+        }
+        if (abortmonitor->guillotineCost)
+        {
+            if (checkCostInterval<=waittime)
+            {
+                cost_type totalCost = aggregateCost(queryWorkUnit());
+                if (totalCost > abortmonitor->guillotineCost)
+                {
+                    errorText.appendf("Workunit cost limit exceeded");
+                    break;
+                }
+                checkCostInterval=CHECK_COST_INTERVAL;
+            }
+            else
+            {
+                checkCostInterval-=waittime;
+            }
         }
     }
     fatalAbort(isAborting,errorText.str());
--- a/ecl/eclagent/eclgraph.cpp
+++ b/ecl/eclagent/eclgraph.cpp
@@ -886,7 +886,17 @@ void EclSubGraph::updateProgress()
             if (startGraphTime)
                 parent.updateWUStatistic(lockedwu, SSTsubgraph, subgraphid, StWhenStarted, nullptr, startGraphTime);
             if (elapsedGraphCycles)
-                parent.updateWUStatistic(lockedwu, SSTsubgraph, subgraphid, StTimeElapsed, nullptr, cycle_to_nanosec(elapsedGraphCycles));
+            {
+                unsigned __int64 elapsedTime = cycle_to_nanosec(elapsedGraphCycles);
+                parent.updateWUStatistic(lockedwu, SSTsubgraph, subgraphid, StTimeElapsed, nullptr, elapsedTime);
+                const cost_type cost = calcCost(agent->queryAgentMachineCost(), nanoToMilli(elapsedTime));
+                if (cost)
+                {
+                    StringBuffer scope;
+                    scope.append(WorkflowScopePrefix).append(agent->getWorkflowId()).append(":").append(subgraphid);
+                    lockedwu->setStatistic(queryStatisticsComponentType(), queryStatisticsComponentName(), SSTsubgraph, scope, StCostExecute, NULL, cost, 1, 0, StatsMergeReplace);
+                }
+            }
         }
     }
 }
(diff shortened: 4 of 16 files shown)


shamser:

common/workunit/workunit.cpp
```diff
@@ -2487,6 +2487,39 @@ class CompoundAggregator : implements StatisticAggregator
...
+    virtual void noteStatistic(StatisticKind kind, unsigned __int64 value, IConstWUStatistic & extra) override
+    {
+        summary.noteValue(value);
+    }
+
+protected:
+    StatsAggregation & summary;
+};
+
+extern WORKUNIT_API __int64 aggregateCost(IConstWorkUnit * wu)
```
jakesmith: minor: we do not normally declare the implementation with extern <..>. Normally we only declare extern in the header.
I know there are few others like this in this file, but majority elsewhere are only in header.

ecl/eclagent/eclagent.cpp
```diff
@@ -3185,6 +3187,21 @@ void EclAgent::abortMonitor()
             IERRLOG("EclAgent failed to abort within %ds - killing process",ABORT_DEADMAN_INTERVAL);
             break;
         }
+        if (checkCostInterval <= waittime )
```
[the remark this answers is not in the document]
jakesmith: If inital checkCostInterval was less than waittime, it means wouldn't check until waittime had expired.
Not sure that matters though, it means checkCostInterval doesn't work if < 30s

thorlcr/master/thdemonserver.cpp
```diff
@@ -180,6 +200,12 @@ class DeMonServer : public CSimpleInterface, implements IDeMonServer
     {
         lastReport = msTick();
         reportRate = globals->getPropInt("@watchdogProgressInterval", 30);
+        IPropertyTree *costs = queryCostsConfiguration();
```
jakesmith: why is this a global method in jptree?
I might have expected this to be queryComponentConfiguration().queryPropTree("cost"); 
or for that to be in a helper, but surprised cost built into jptree.

ecl/eclagent/eclgraph.cpp
```diff
@@ -1746,6 +1746,12 @@ void EclAgent::executeThorGraph(const char * graphName)
             {
                 throw deserializeException(reply);
             }
+            case DAMP_THOR_REPLY_COSTEXCEEDED:
+            {
+                Owned<IException> e = deserializeException(reply);
+                StringBuffer str;
+                throw new WorkflowException(e->errorCode(), e->errorMessage(str), 0, WorkflowException::ABORT, e->errorAudience());
```
jakesmith: was this required (new reply code)? or could you have sent back the exception as a DAMP_THOR_REPLY_ERROR ?

thorlcr/master/thdemonserver.cpp
```diff
@@ -180,6 +200,12 @@ class DeMonServer : public CSimpleInterface, implements IDeMonServer
...
+        if (costs)
+        {
+            thorMasterRate = costs->getPropReal("thor/@master", 0.0);
+            thorSlaveRate = costs->getPropReal("thor/@slave", 0.0);
```
jakesmith: minor: The other getPropX methods have defaults, I notice the new getPropReal doesn't.
Should probably be changed to be consistent, and would not need 0.0 passed here.

common/workunit/workunit.hpp
```diff
@@ -1689,7 +1690,10 @@ extern WORKUNIT_API bool isValidMemoryValue(const char * memoryUnit);
 
 #define HourToSeconds(n) ((n)/3600)
 #define NanoSecondsToSeconds(n) ((double)(n)/1000000000)
-inline __int64 calcCost(double ratePerHour, __int64 timeNS) { return HourToSeconds(ratePerHour) * NanoSecondsToSeconds(timeNS) * 1e6; }
+#define MilliSecondsToSeconds(n) ((double)(n)/1000)
+inline __int64 calcCost(double ratePerHour, __int64 timeNS) { return HourToSeconds(ratePerHour) * NanoSecondsToSeconds(timeNS) * 1e6; };
+inline __int64 calcCostMs(double ratePerHour, __int64 timeMS) { return HourToSeconds(ratePerHour) * MilliSecondsToSeconds(timeMS) * 1e6; };
```
jakesmith: trivial: extra trailing ; at end this line and one above.

common/workunit/workunit.cpp
```diff
@@ -2487,6 +2487,39 @@ class CompoundAggregator : implements StatisticAggregator
...
+{
+    const WuScopeFilter filter("stat[CostExecute],depth[0..3],nested[0]");
```
jakesmith: am not quite sure of the depth[0..3] semantics here.
Perhaps needs a comment added here to explain.

jakesmith (commented):
@shamser - looks good in general, a few minor comments only.

common/workunit/workunit.cpp
```diff
@@ -2487,6 +2487,39 @@ class CompoundAggregator : implements StatisticAggregator
...
+        {
+            unsigned __int64 value;
+            if (it->getStat(StCostExecute, value) && value!=0)
+                it->nextSibling();
+        }
+    }
+    return summary.getSum() / 1E6;
+}
 
 //To calculate aggregates, create a scope iterator, and an instance of a StatisticAggregator, and play the attributes through the interface
```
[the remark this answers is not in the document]
shamser: No, this shouldn't be here.

ecl/eclagent/eclagent.cpp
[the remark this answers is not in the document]
shamser: That's true.  I didn't think timeout check interval would be > checkCost interval.  However, I've updated  the code now so the the interval is lesser of timeout interval and checkCost interval.

ecl/eclagent/eclgraph.cpp
shamser: If I used DAMP_THOR_REPLY_ERROR, a generic error message is provided: 
image

shamser: With DAMP_THOR_REPLY_COSTEXCEEDED, I can structure the message better:
image
I was thinking about making this more generic.  However, I cannot think how else a generic version of this message would be used.

thorlcr/master/thdemonserver.cpp
shamser: queryCostsConfiguration() was introduced several pull requests ago and has been already merged.  It is already being used in a number of different places.  I could move queryCostsConfiguration() to workunit.hpp or somewhere else?

ecl/eclagent/eclgraph.cpp
jakesmith: what did it look like if you pass back the exception with a DAMP_THOR_REPLY_ERROR code?

thorlcr/master/thdemonserver.cpp
jakesmith: it's not a big deal - doesn't need to move now.
But feels like it doesn't belong in jlib (yes workunit would seem more appropriate).
@ghalliday - might want to comment.

common/workunit/workunit.hpp
```diff
@@ -1689,7 +1690,11 @@ extern WORKUNIT_API bool isValidMemoryValue(const char * memoryUnit);
 
 #define HourToSeconds(n) ((n)/3600)
 #define NanoSecondsToSeconds(n) ((double)(n)/1000000000)
-inline __int64 calcCost(double ratePerHour, __int64 timeNS) { return HourToSeconds(ratePerHour) * NanoSecondsToSeconds(timeNS) * 1e6; }
+#define MilliSecondsToSeconds(n) ((double)(n)/1000)
+inline unsigned __int64 calcCostSecs(double ratePerHour, double secs) { return HourToSeconds(ratePerHour) * secs * 1e6; }
+inline unsigned __int64 calcCostNs(double ratePerHour, unsigned __int64 timeNS) { return calcCostSecs(ratePerHour, NanoSecondsToSeconds(timeNS)); }
+inline unsigned __int64 calcCostMs(double ratePerHour, unsigned __int64 timeMS) { return calcCostSecs(ratePerHour, MilliSecondsToSeconds(timeMS)); }
```
jakesmith: do we ever need to calc. cost at NS precision ? Seems unlikely/excessive.
I would have thought per second would suffice for the foreseeable future.

ecl/eclagent/eclagent.cpp
```diff
@@ -564,7 +565,12 @@ EclAgent::EclAgent(IConstWorkUnit *wu, const char *_wuid, bool _checkVersion, bo
         updateSuppliedXmlParams(w);
     }
     IPropertyTree *costs = queryCostsConfiguration();
-    agentMachineCost = costs ? costs->getPropReal("@agent", 0.0): 0.0;
+    if (costs)
+    {
+        agentMachineCost = costs->getPropReal("@agent");
+        softCostLimit = costs->getPropInt("@limit", 0);
+        hardCostLimit = costs->getPropInt("@hardlimit", 0);
```
jakesmith: trivial: 0 is default on getPropInt.

ecl/eclagent/eclagent.cpp
```diff
@@ -2312,7 +2318,7 @@ void EclAgentWorkflowMachine::noteTiming(unsigned wfid, timestamp_type startTime
     updateWorkunitStat(wu, SSTworkflow, scope, StWhenStarted, nullptr, startTime, 0);
     updateWorkunitStat(wu, SSTworkflow, scope, StTimeElapsed, nullptr, elapsedNs, 0);
 
-    const __int64 cost = calcCost(agent.queryAgentMachineCost(), elapsedNs);
+    const __int64 cost = calcCostNs(agent.queryAgentMachineCost(), elapsedNs);
```
jakesmith: Having a calcCostNS seems overkill, unlikely to be practical to get that level of prevision.
Seconds are probably enough. I'd probably convert and settle on 1 cost method.

ecl/eclagent/eclagent.cpp
```diff
@@ -3177,14 +3190,30 @@ void EclAgent::abortMonitor()
...
             while (abortmonitor->sem.wait(ABORT_DEADMAN_INTERVAL*1000))
                 if (abortmonitor->stopping)
                     return; // stopped in time
             IERRLOG("EclAgent failed to abort within %ds - killing process",ABORT_DEADMAN_INTERVAL);
             break;
         }
+        if (checkCostInterval <= waittime)
+        {
+            __int64 totalCost = aggregateCost(queryWorkUnit());
+            if (totalCost > abortmonitor->guillotineCost)
```
jakesmith: can guillotineCost be unset / 0 / disabled ?
If so, the checkCostInterval and checking should be disabled.

ecl/eclagent/eclgraph.cpp
```diff
@@ -1805,6 +1811,13 @@ void EclAgent::executeGraph(const char * graphName, bool realThor, size32_t pare
             }
 
             activeGraph.setown(loadGraph(graphName, queryWorkUnit(), dll, false));
+            unsigned guillotineCost = queryWorkUnit()->getDebugValueInt("maxCost", softCostLimit);
+            if (guillotineCost)
+            {
+                if (hardCostLimit && guillotineCost > hardCostLimit)
+                    guillotineCost = hardCostLimit;
+                abortmonitor->setGuillotineCost(guillotineCost);
```
jakesmith: re. other question about whether guillotineCost can be 0.
Looks like if workunit setting "maxCost" could be 0, in which case abortmonitor->setGuillotineCost would not be called.
So I think the checkCostInterval logic and check needs to be disabled in that case.

thorlcr/master/thgraphmanager.cpp
```diff
@@ -838,8 +838,21 @@ void CJobManager::reply(IConstWorkUnit *workunit, const char *wuid, IException *
...
+        if (te)
+        {
+            switch (te->errorCode())
+            {
+            case TE_WorkUnitAborting:
+                replyMb.append((unsigned)DAMP_THOR_REPLY_ABORT);
+                break;
+            case TE_CostExceeded:
+                replyMb.append((unsigned)DAMP_THOR_REPLY_COSTEXCEEDED);
+                break;
```
jakesmith: I don't think you need a new reply code.

A user abort is thrown as:
```
MakeThorException(TE_AbortException, "THOR ABORT")
```

vs your exception is:
```
                LOG(MCwarning, thorJob, "ABORT job cost exceeds limit");
                Owned<IException> e = MakeThorException(TE_CostExceeded, "Job cost exceeds limit (cost: %" I64F "d, limit %u)", totalCost, costLimit);
```

But all failure exceptions from Thor will take this path, resulting in them being reported to the workunit with their text.
TE_WorkUnitAborting is special case - I'm not sure why it was introduced.

jakesmith (changes requested):
@shamser - back to you.

shamser:
@jakesmith Please can you have a look at the changes.

ecl/eclagent/eclagent.cpp
```diff
@@ -565,7 +565,12 @@ EclAgent::EclAgent(IConstWorkUnit *wu, const char *_wuid, bool _checkVersion, bo
         updateSuppliedXmlParams(w);
     }
     IPropertyTree *costs = queryCostsConfiguration();
-    agentMachineCost = costs ? costs->getPropReal("@agent", 0.0): 0.0;
+    if (costs)
+    {
+        agentMachineCost = costs->getPropReal("@agent",0.0);
```
jakesmith: trivial: but lines below use default, this one for consistency should

ecl/eclagent/eclgraph.cpp
```diff
@@ -1811,9 +1811,13 @@ void EclAgent::executeGraph(const char * graphName, bool realThor, size32_t pare
             }
 
             activeGraph.setown(loadGraph(graphName, queryWorkUnit(), dll, false));
-            unsigned guillotineCost = queryWorkUnit()->getDebugValueInt("maxCost", 0);
+            unsigned guillotineCost = queryWorkUnit()->getDebugValueInt("maxCost", softCostLimit);
             if (guillotineCost)
+            {
+                if (hardCostLimit && guillotineCost > hardCostLimit)
```
jakesmith: I think it's ok, but had to check operator precedence, it would clearer if you added brackets, i.e.
```
if (hardCostLimit && (guillotineCost > hardCostLimit))
```

thorlcr/master/thdemonserver.cpp
```diff
@@ -256,7 +256,17 @@ class DeMonServer : public CSimpleInterface, implements IDeMonServer
 
         IConstWorkUnit & wu =  graph->queryJob().queryWorkUnit();
         workunitCost = aggregateCost(&wu);
-        costLimit = wu.getDebugValueInt("maxCost", 0);
+
+        IPropertyTree *costs = queryCostsConfiguration();
```
jakesmith: trivial: probably could be const IPropertyTree *costs ?

thorlcr/master/thdemonserver.cpp
```diff
@@ -256,7 +256,17 @@ class DeMonServer : public CSimpleInterface, implements IDeMonServer
...
+        unsigned softLimit = 0, hardLimit = 0;
+        if (costs)
+        {
+            softLimit = costs->getPropInt("@limit");
+            hardLimit = costs->getPropInt("@hardlimit");
+        }
+        costLimit = wu.getDebugValueInt("maxCost", softLimit);
+        if (hardLimit && costLimit > hardLimit)
```
jakesmith: adding brackets would make it clearer.

common/workunit/workunit.cpp
```diff
@@ -13015,15 +13016,15 @@ extern WORKUNIT_API void addTimeStamp(IWorkUnit * wu, StatisticScopeType scopeTy
     wu->setStatistic(queryStatisticsComponentType(), queryStatisticsComponentName(), scopeType, scopestr, kind, NULL, getTimeStampNowValue(), 1, 0, StatsMergeAppend);
 }
 
-extern WORKUNIT_API double calculateThorCost(__int64 timeNs, unsigned clusterWidth)
+extern WORKUNIT_API double calculateThorCost(double elapsedTime, unsigned clusterWidth)
```
jakesmith: I think 'elapsedTime' units are seconds? But would be clearer if elapsedTimeSecs

common/workunit/workunit.hpp
```diff
@@ -1659,7 +1659,7 @@ extern WORKUNIT_API const char * getWorkunitActionStr(WUAction action);
 extern WORKUNIT_API WUAction getWorkunitAction(const char * actionStr);
 
 extern WORKUNIT_API void addTimeStamp(IWorkUnit * wu, StatisticScopeType scopeType, const char * scope, StatisticKind kind, unsigned wfid=0);
-extern WORKUNIT_API double calculateThorCost(__int64 timeNs, unsigned clusterWidth);
+extern WORKUNIT_API double calculateThorCost(double elapsedTime, unsigned clusterWidth);
```
jakesmith: I think 'elapsedTime' units are seconds? But would be clearer if elapsedTimeSecs

jakesmith:
@shamser - looks good, only a couple of new minor comments

jakesmith:
Please check and squash.

shamser:
@jakesmith Last few changes as suggested and squashed.

jakesmith (approved):
@shamser - looks good

shamser:
@richardkchapman This is ready for merging.

richardkchapman:

thorlcr/master/thdemonserver.cpp
```diff
@@ -38,6 +38,11 @@ class DeMonServer : public CSimpleInterface, implements IDeMonServer
     unsigned reportRate;
     CIArrayOf<CGraphBase> activeGraphs;
     UnsignedArray graphStarts;
+    double thorMasterRate = 0.0;
```
ghalliday: picky: Rate doesn't tell you what it is a rate of?  CostRate?

thorlcr/master/thdemonserver.cpp
```diff
@@ -38,6 +38,11 @@ class DeMonServer : public CSimpleInterface, implements IDeMonServer
...
+    double thorSlaveRate = 0.0;
+    unsigned costLimit = 0;
```
ghalliday: unsigned __int64.  I suspect using stats_type or a new cost_type would be clearer.

thorlcr/master/thdemonserver.cpp
```diff
@@ -84,6 +89,20 @@ class DeMonServer : public CSimpleInterface, implements IDeMonServer
         unsigned duration = msTick()-startTime;
         updateWorkunitStat(wu, SSTsubgraph, graphScope, StTimeElapsed, timer, milliToNano(duration));
 
+        if (costLimit || finished)
+        {
+            const unsigned clusterWidth = queryNodeClusterWidth();
+            const unsigned __int64 sgCost = (calcCost(thorMasterRate, ms2Secs(duration)) + calcCost(thorSlaveRate, ms2Secs(duration)) * clusterWidth);
+            const unsigned __int64 totalCost = workunitCost + sgCost/1E6;
```
ghalliday: Why the 1E6?  It isn't clear what the units are for each of your calculations.

thorlcr/master/thdemonserver.cpp
```diff
@@ -84,6 +89,20 @@ class DeMonServer : public CSimpleInterface, implements IDeMonServer
...
+            if (finished)
+                wu->setStatistic(queryStatisticsComponentType(), queryStatisticsComponentName(), SSTsubgraph, graphScope, StCostExecute, NULL, sgCost, 1, 0, StatsMergeReplace);
+
+            if (totalCost>costLimit)
+            {
+                LOG(MCwarning, thorJob, "ABORT job cost exceeds limit");
+                graph.fireException(MakeThorException(TE_CostExceeded, "Job cost exceeds limit (cost: %" I64F "d, limit %u)", totalCost, costLimit));
```
ghalliday: These need formatting correctly as costs using formatStatistic()

thorlcr/master/thdemonserver.cpp
ghalliday: I suspect you're right Jake, but probably not worth moving now.

thorlcr/master/thdemonserver.cpp
```diff
@@ -227,6 +252,20 @@ class DeMonServer : public CSimpleInterface, implements IDeMonServer
     void startGraph(CGraphBase *graph)
     {
         synchronized block(mutex);
+
+        IConstWorkUnit & wu =  graph->queryJob().queryWorkUnit();
+        workunitCost = aggregateCost(&wu);
+
+        const IPropertyTree *costs = queryCostsConfiguration();
+        unsigned softLimit = 0, hardLimit = 0;
```
ghalliday: These can't be unsigned - see comments above about units.

ghalliday: E.g. $100 = 11,000 yen with a 1E6 scaling factor is > unsigned.

ghalliday (commented):
@shamser some comments.  A few general areas, the units needs to be clearer, formatting required whenever a cost is presented to the user, and more though/care needed about how wokflow/graph/subgraph costs are aggregated.

ecl/eclagent/eclagent.cpp
```diff
@@ -3183,14 +3198,33 @@ void EclAgent::abortMonitor()
...
         }
+        if (abortmonitor->guillotineCost>0)
+        {
+            if (checkCostInterval<=waittime)
+            {
+                unsigned __int64 totalCost = aggregateCost(queryWorkUnit());
+                if (totalCost>abortmonitor->guillotineCost)
+                {
+                    DBGLOG("Guillotine triggered (cost exceeded)");
+                    errorText.appendf("Workunit cost limit (limit %d) exceeded", abortmonitor->guillotineCost);
```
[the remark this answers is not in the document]
shamser: I'll handle all formatting separately: https://track.hpccsystems.com/browse/HPCC-23805

thorlcr/master/thdemonserver.cpp
shamser: I'll handle all formatting separately: https://track.hpccsystems.com/browse/HPCC-23805

shamser:

thorlcr/master/thdemonserver.cpp
```diff
@@ -227,6 +251,21 @@ class DeMonServer : public CSimpleInterface, implements IDeMonServer
...
         synchronized block(mutex);
+
+        IConstWorkUnit & wu =  graph->queryJob().queryWorkUnit();
+        workunitCost = aggregateCost(&wu);
+
+        const IPropertyTree *costs = queryCostsConfiguration();
+        cost_type softLimit = 0, hardLimit = 0;
+        if (costs)
+        {
+            softLimit = costs->getPropInt64("@limit");
```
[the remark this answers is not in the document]
ghalliday: I see below you are scaling the result - which I think is correct - but in that case all of these variables should be doubles.

common/workunit/workunit.cpp
```diff
@@ -2487,8 +2487,73 @@ class CompoundAggregator : implements StatisticAggregator
...
+    bool haveWorkFlowCost = false;
+    StringBuffer workflowScope;
+    bool haveGraphCost = false;
+    StringBuffer graphScope;
+    StatsAggregation & summary;
+};
+
+cost_type aggregateCost(const IConstWorkUnit * wu, bool excludehThor)
+{
+    // depth 0: Agent/hThor global or compile
```
ghalliday, line 2544: compile is depth 1, since it has a scope of "compile"

common/workunit/workunit.cpp
ghalliday: Do you actually want to match global costs?  If so it will double count them.  I suspect the depth should be [1..3]

common/workunit/workunit.cpp
```diff
@@ -2487,8 +2487,73 @@ class CompoundAggregator : implements StatisticAggregator
     IArrayOf<StatisticAggregator> aggregators;
 };
 
+class CostAggregator : public StatisticAggregator
+{
+public:
+    CostAggregator(StatsAggregation & _summary, bool _excludehThor) : StatisticAggregator(StCostExecute), summary(_summary), excludehThor(_excludehThor) {}
+
+    virtual void noteStatistic(StatisticKind kind, unsigned __int64 value, IConstWUStatistic & extra) override
```
ghalliday, line 2495: This code is confusing for several reasons:
- haveGraphCost is calculated, but never used.
- It will not work for roxie (when roxie is run in per-workunit mode)
- Efficiency probably isn't a big issue, but in other situations using nextSibling() would be much more efficient.
- excludehThor looks like it has been misspelled.

ecl/eclagent/eclagent.cpp
```diff
@@ -2318,7 +2333,7 @@ void EclAgentWorkflowMachine::noteTiming(unsigned wfid, timestamp_type startTime
     updateWorkunitStat(wu, SSTworkflow, scope, StWhenStarted, nullptr, startTime, 0);
     updateWorkunitStat(wu, SSTworkflow, scope, StTimeElapsed, nullptr, elapsedNs, 0);
 
-    const __int64 cost = calcCost(agent.queryAgentMachineCost(), elapsedNs);
+    const cost_type cost = calcCost(agent.queryAgentMachineCost(), ns2Secs(elapsedNs));
```
ghalliday, line 2336: I think this should also be including the cost for all graphs for this workunit that are executed in thor, so that the cost for a workflow item is the total cost for that workflow item.

common/workunit/workunit.cpp
ghalliday: With the suggested change to workflow item costs, this logic should be removed.  There isn't a need for this class.  Something like the following instead:

```
  Owned<IConstWUScopeIterator> it = &wu->getScopeIterator(filter);
  for (it->first(); it->isValid(); )
    {
        if (it->getStat(CostExecute, value)) // I think this should always be true
       {  sum+= value; it->nextSibling(); }
        else
            it->next();
    }
    return sum;

```

ghalliday (commented):
@shamser a few more comments - more on which units (and data types) are you using, and comment on what workflow costs mean.

common/workunit/workunit.cpp
shamser: Sometimes the children of a scope is included/excluded based on the creator (hThor or thor) so we need to be able to distinguish the creator (to determine if it's an hthor stat or thor stat).  

The only way to do that is at present within the visitor class.

shamser:
@ghalliday Please can you review.

common/workunit/workunit.cpp
```diff
@@ -2487,8 +2487,74 @@ class CompoundAggregator : implements StatisticAggregator
...
+    virtual cost_type getTotalCost() const { return totalCost; }
+protected:
+    const bool excludeHThor;
+    bool haveWorkFlowCost = false;
+    StringBuffer workflowScope;
+    cost_type totalCost = 0;
+};
+
+// Aggregrate Thor or hThor costs
+cost_type aggregateCost(const IConstWorkUnit * wu, const char *scope, bool excludeHThor)
```
ghalliday, line 2532: I think you are using this for 3 different scenarios:

a) Give me the total time so far for the job
b) What is the total time within a workflow item that was not executed locally (EclAgentWorkflowMachine::noteTiming)
c) How much time has been spent in this process that is not included in the current time.  (EclAgent::doProcess)

I think you might need 3 different approaches
a) Iterate all scopes.  Whenever you find a stat in a scope call nextSibling(), otherwise you will double count workflow items and the graphs within them
b) In three minds!
  1) Use the code is currently implemented - but with a bit more work on the filter to ensure subgraph stats are not iterated.
  2) If the target cluster type is "Thor" then Total the costs within a wfid, and add them to the wfid cost.
  3) Keep track of the time spent executing graphs within a workflow item, always add the cost for graphs, but subtract off the time for local graphs.
It might help to consider how it is affected if other items within the workflow start recording costs (e.g., sprays?).  I suspect (1) is the best option but it should document how the extra cases are handled.
c)  This should probably be done by calculating it based on the local execute time for the eclagent - i.e. subtracting off the time spent processing workflow items, but adding on the costs for all child items.  But it raises the question of how jobs that use the scheduler are costed.  Does it mean there is no cost for the workunit until it is complete.  What if it never completes?  Where is the costs for waiting for a persist to be rebuilt stored?

The PR should really contain a comment describing the how the costs are meant to work.  E.g., the costs at each level include the costs from all lower levels.

common/workunit/workunit.cpp
```diff
@@ -2487,8 +2487,74 @@ class CompoundAggregator : implements StatisticAggregator
...
+{
+    // depth 1: workflow, depth 2: graph, depth 3: subgraph
+    WuScopeFilter filter;
+    if (scope && *scope) // All costs under specified scope (used to calculate all costs for a workflow)
+    {
+        filter.addScope(scope);
```
ghalliday, line 2538: I suspect as it stands this will also process all the subgraph stats.

common/workunit/workunit.cpp
```diff
@@ -2487,8 +2487,74 @@ class CompoundAggregator : implements StatisticAggregator
...
+    }
+    else
+        filter.addFilter("stat[CostExecute],depth[1..3],nested[0],where[CostExecute]");
+    filter.finishedFilter();
+    FilteredCostAggregator aggregator(excludeHThor);
+    Owned<IConstWUScopeIterator> it = &wu->getScopeIterator(filter);
+    for (it->first(); it->isValid(); )
+    {
+        it->playProperties(aggregator);
+        if ((it->getScopeType() == SSTgraph) && (it->queryStat(StCostExecute, MaxStatisticValue) != MaxStatisticValue))
```
ghalliday, line 2551: I'm not sure this should be special cased for a graph.  You can also use getStat to see if the stat exists.

common/workunit/workunit.cpp
```diff
@@ -2487,8 +2487,74 @@ class CompoundAggregator : implements StatisticAggregator
...
+    FilteredCostAggregator(bool _excludeHThor) : excludeHThor(_excludeHThor) {}
 
-//To calculate aggregates, create a scope iterator, and an instance of a StatisticAggregator, and play the attributes through the interface
+    virtual void noteStatistic(StatisticKind kind, unsigned __int64 value, IConstWUStatistic & extra) override
+    {
+        if (extra.getCreatorType() == SCThthor)
+        {
+            if (excludeHThor)
+                return;
+            if (haveWorkFlowCost)
```
ghalliday, line 2504: I'm not sure why this is hthor only.  It should also have been short-circuited with nextSibling() on the wofkflow item.

common/workunit/workunit.hpp
```diff
@@ -1688,8 +1690,10 @@ extern WORKUNIT_API bool isValidPriorityValue(const char * priority);
 extern WORKUNIT_API bool isValidMemoryValue(const char * memoryUnit);
 
 #define HourToSeconds(n) ((n)/3600)
-#define NanoSecondsToSeconds(n) ((double)(n)/1000000000)
-inline __int64 calcCost(double ratePerHour, __int64 timeNS) { return HourToSeconds(ratePerHour) * NanoSecondsToSeconds(timeNS) * 1e6; }
+#define ns2Secs(n) ((double)(n)/1000000000)
+#define ms2Secs(n) ((double)(n)/1000)
+inline cost_type calcCost(cost_type ratePerHour, double secs) { return HourToSeconds(ratePerHour) * secs; }
```
ghalliday, line 1695: HoursToSeconds is introducing a potentially significant rounding error.  E.g. if the cost was $0.001/hour - it would round to zero.

common/workunit/workunit.hpp
```diff
@@ -1688,8 +1690,10 @@ extern WORKUNIT_API bool isValidPriorityValue(const char * priority);
 extern WORKUNIT_API bool isValidMemoryValue(const char * memoryUnit);
 
 #define HourToSeconds(n) ((n)/3600)
```
ghalliday, line 1692: On reflection this should be an inline function, not a macro.  It is also confusing - why does converting from an hour to seconds divide by 3600?  I would expect the opposite.

common/workunit/workunit.hpp
```diff
@@ -1688,8 +1690,10 @@ extern WORKUNIT_API bool isValidPriorityValue(const char * priority);
...
-#define NanoSecondsToSeconds(n) ((double)(n)/1000000000)
-inline __int64 calcCost(double ratePerHour, __int64 timeNS) { return HourToSeconds(ratePerHour) * NanoSecondsToSeconds(timeNS) * 1e6; }
+#define ns2Secs(n) ((double)(n)/1000000000)
+#define ms2Secs(n) ((double)(n)/1000)
```
ghalliday, line 1694: again should be inline functions, not macros.

thorlcr/master/thdemonserver.cpp
```diff
@@ -227,6 +251,21 @@ class DeMonServer : public CSimpleInterface, implements IDeMonServer
     void startGraph(CGraphBase *graph)
     {
         synchronized block(mutex);
+
+        IConstWorkUnit & wu =  graph->queryJob().queryWorkUnit();
+        workunitCost = aggregateCost(&wu);
+
+        const IPropertyTree *costs = queryCostsConfiguration();
+        cost_type softLimit = 0, hardLimit = 0;
```
ghalliday, line 259: Should be type double

thorlcr/master/thdemonserver.cpp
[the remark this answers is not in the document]
ghalliday: Should be getPropReal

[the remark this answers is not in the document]
ghalliday: (to match changes in eclagent)

ghalliday (commented):
@shamser still a few issues.  The PR should have a comment somewhere describing how the costs are calculated, and how the complications are handled.

common/workunit/workunit.cpp
shamser: In reply to 1: I'm not sure that it is iterating subgraph unnecessary.  I've emailed you some log extracts.

common/workunit/workunit.cpp
shamser: I don't think it is iterating subgraph unnecessary.  I've emailed you some log extracts.

common/workunit/workunit.cpp
shamser: If this is not special-cased, then the workflow doesn't include the thor graph costs.

common/workunit/workunit.hpp
shamser: I've named this poorly.  It was designed to convert from hourly rate to per second rate.

shamser:
@ghalliday Please can you review.

ecl/eclagent/eclagent.cpp
```diff
@@ -565,7 +566,23 @@ EclAgent::EclAgent(IConstWorkUnit *wu, const char *_wuid, bool _checkVersion, bo
...
     IPropertyTree *costs = queryCostsConfiguration();
-    agentMachineCost = costs ? costs->getPropReal("@agent", 0.0): 0.0;
+    if (costs)
+    {
+        agentMachineCost = money2cost_type(costs->getPropReal("@agent"));
+        if (agentMachineCost)
+        {
+            double softCostLimit = costs->getPropReal("@limit");
+            double guillotineCost = wu->getDebugValueReal("maxCost", softCostLimit);
+            if (guillotineCost > 0.0)
```
ghalliday, line 576: This means if a user sets maxCost to 0 in a workunit they can avoid the hard cost limit.

thorlcr/master/thdemonserver.cpp
```diff
@@ -227,6 +251,22 @@ class DeMonServer : public CSimpleInterface, implements IDeMonServer
...
+
+        const IPropertyTree *costs = queryCostsConfiguration();
+        double softLimit = 0.0, hardLimit = 0.0;
+        if (costs)
+        {
+            softLimit = costs->getPropReal("@limit");
+            hardLimit = costs->getPropReal("@hardlimit");
+        }
+        double tmpcostLimit = wu.getDebugValueReal("maxCost", softLimit);
+        if (hardLimit && (tmpcostLimit > hardLimit))
```
ghalliday, line 266: Currently hard limits can be avoided by setting maxcost to 0..  Should be something like:

`hardLimit && ((tmpcostLimit == 0) || (tmpcostLimit > hardLimit))
`

ghalliday (commented):
@shamser looks good.  Only issue is the ability to avoid hard limits.

PR #13515, opened 2020-03-17
HPCC-23726 Reorganize thor.yaml configmaps and pass through --config

jakesmith:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jptree.cpp
+++ b/system/jlib/jptree.cpp
@@ -24,6 +24,7 @@
 #include "jlzw.hpp"
 #include "jregexp.hpp"
 #include "jstring.hpp"
+#include "jutil.hpp"
 #include "yaml.h"
 
 #ifdef __APPLE__
@@ -31,7 +32,7 @@
 #define environ (*_NSGetEnviron())
 #endif
 
-#include <algorithm>
+#include <initializer_list>
 
 #define MAKE_LSTRING(name,src,length) \
     const char *name = (const char *) alloca((length)+1); \
@@ -7754,16 +7755,8 @@ static const char * extractOption(const char * option, const char * cur)
     return nullptr;
 }
 
-static bool ignoreOption(const char * name)
-{
-    return streq(name, "config") || streq(name, "outputconfig");
-}
-
 static void applyCommandLineOption(IPropertyTree * config, const char * option, const char * value)
 {
-    if (ignoreOption(option))
-        return;
-
     if (islower(*option))
     {
         StringBuffer path;
@@ -7777,20 +7770,25 @@ static void applyCommandLineOption(IPropertyTree * config, const char * option, 
     }
 }
 
-static void applyCommandLineOption(IPropertyTree * config, const char * option)
+static void applyCommandLineOption(IPropertyTree * config, const char * option, std::initializer_list<const char *> ignoreOptions)
 {
     const char * eq = strchr(option, '=');
+    StringBuffer name;
+    const char *val = nullptr;
     if (eq)
     {
-        StringBuffer name;
         name.append(eq - option, option);
-        applyCommandLineOption(config, name, eq + 1);
+        option = name;
+        val = eq + 1;
     }
     else
     {
         //MORE: Support --x- and --x+?
-        applyCommandLineOption(config, option, "1");
+        val = "1";
     }
+    if (stdContains(ignoreOptions, option))
+        return;
+    applyCommandLineOption(config, option, val);
 }
 
 static Owned<IPropertyTree> componentConfiguration;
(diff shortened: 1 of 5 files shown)


jakesmith:

richardkchapman:
Looks ok. @ghalliday

system/jlib/jptree.cpp
```diff
@@ -7774,20 +7766,26 @@ static void applyCommandLineOption(IPropertyTree * config, const char * option,
...
+        option = name;
+        val = eq + 1;
     }
     else
     {
         //MORE: Support --x- and --x+?
-        applyCommandLineOption(config, option, "1");
+        val = "1";
     }
+    auto it = std::find(ignoreOptions.begin(), ignoreOptions.end(), option);
```
ghalliday: style: I know it is standard c++11, but whenever I see it looks ugly.  Can we add a template function into jlib
`bool contains(T1 x, T2 search) { return (std::find(x.begin(), x.end(), search) != x.end()); }`

system/jlib/jptree.cpp
```diff
@@ -7774,20 +7766,26 @@ static void applyCommandLineOption(IPropertyTree * config, const char * option,
     }
 }
 
-static void applyCommandLineOption(IPropertyTree * config, const char * option)
+static void applyCommandLineOption(IPropertyTree * config, const char * option, std::list<std::string> ignoreOptions)
```
ghalliday: efficiency: It doesn't really matter in this context, but if the parameters are going to be constant then std::initializer_list is lower overhead.  This could be std::initializer_list<const char *>.

ghalliday (commented):
@jakesmith two minor comments on the code.  I'm inclined to merge as is.

ghalliday:
@jakesmith looks like it has conflicts with something else that was merged.

system/jlib/jptree.cpp
jakesmith: yep, it is standard, but I agree also feels too long winded/verbose when I use it.
I'll add a helper.

system/jlib/jptree.cpp
jakesmith: ok, I may change it.

jakesmith:
>looks like it has conflicts with something else that was merged.

Richard's PR beat me to it :) -I'll rebase/resolve merges.

jakesmith:
@ghalliday - added a helper function (see stdIterContains in jutil.hpp), changed to initialize_list and rebased/resolved conflicts.

system/jlib/jptree.hpp
```diff
@@ -24,6 +24,9 @@
 #include "jiter.hpp"
 #include "jprop.hpp"
 
+#include <list>
```
ghalliday: Should now be 
#include <initializer_list>

system/jlib/jptree.hpp
```diff
@@ -293,7 +296,7 @@ inline static bool isValidXPathChr(char c)
     return ('\0' != c && (isalnum(c) || strchr(validChrs, c)));
 }
 
-jlib_decl IPropertyTree * loadArgsIntoConfiguration(IPropertyTree *config, const char * * argv);
+jlib_decl IPropertyTree * loadArgsIntoConfiguration(IPropertyTree *config, const char * * argv, std::initializer_list<std::string> ignoreOptions = {});
```
ghalliday: picky: initializer_list<const char *> would allow constant data to be passed.

system/jlib/jutil.hpp
```diff
@@ -556,5 +559,11 @@ class jlib_decl QuantileFilterIterator
     bool isQuantile;
 };
 
+template <typename Iterator, typename Value>
+inline bool stdIterContains(Iterator&& iter, Value &&v)
```
ghalliday: picky: I'm not sure why iter is in the name.   Iterator should really be container - begin and end get the iterators.

ghalliday (commented):
@jakesmith some minor comments.  I suspect the include may need to be fixed for the code to compile on some systems.

system/jlib/jutil.hpp
jakesmith: true, will change.

system/jlib/jptree.hpp
jakesmith: yes, will change.

system/jlib/jptree.hpp
jakesmith: changed.

PR #13545, opened 2020-03-24
HPCC-23762 Thor linger

jakesmith:
Allow thor to linger a configurable amount of time after it
has complete executing a graph.
The job/workflow detected that the workunit is associated with
a lingering Thor, it will communicate with it directly if it
can and avoid the latency of stopping/starting new master and
slave containers.

Also make a change to master/slave so that if containerPerAgent
is off in thor agent, then ensure master doesn't use a fix
port (akin to hthor process).

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/common/workunit/workunit.cpp
+++ b/common/workunit/workunit.cpp
@@ -13192,6 +13192,60 @@ bool isValidMemoryValue(const char *memoryUnit)
 }
 
 #ifdef _CONTAINERIZED
+bool executeGraphOnLingeringThor(IConstWorkUnit &workunit, const char *graphName)
+{
+    // check if lingering thor instance is up.
+
+    /* If code was dependent on reading a workunit in parallel, the whole area of
+     * workunit locking and reading will need revisiting, because at the moment the 
+     * workunit reading code does not lock at all.
+     * 
+     */
+
+    /* NB: forcing a reload here to ensure Debug values are recent.
+     * Could be improved by refreshing only the specific area of interest (i.e. Debug/ in this case).
+     * The area of persisting a non-locked connection to workunits should be resivisted..
+     */
+    workunit.forceReload();
+
+    Owned<IStringIterator> iter = &workunit.getDebugValues("thorinstance_*");
+    ForEach(*iter)
+    {
+        /* NB: Thor's set their running endpoint into Debug values of workunit
+         * Check to see if workunit has any Thor's available lingering instances.
+         */
+        SCMStringBuffer thorInstance;
+        iter->str(thorInstance);
+        SCMStringBuffer thorInstanceValue;
+        if (workunit.getDebugValueBool(thorInstance.s, false))
+        {
+            {
+                Owned<IWorkUnit> w = &workunit.lock();
+                w->setDebugValue(thorInstance.s, "0", true);
+            }
+            /* NB: there's a window where Thor could shutdown here.
+             * In that case, the sendRecv will fail and it will fall through to queueing.
+             */
+            const char *instanceName = strchr(thorInstance.str(), '_') + 1;
+
+            Owned<INode> masterNode = createINode(instanceName);
+            CMessageBuffer msg;
+            VStringBuffer jobStr("%s/%s", workunit.queryWuid(), graphName?graphName:"");
+            msg.append(jobStr);
+            if (queryWorldCommunicator().sendRecv(msg, masterNode, MPTAG_THOR, 10000))
+            {
+                bool ok;
+                msg.read(ok);
+                if (graphName) // if graphName==nullptr, then continue around to look at others
+                {
+                    if (ok)
+                        return true;
+                }
+            }
+        }
+    }
+    return false;
+}
 
 static void setResources(StringBuffer &jobYaml, const IConstWorkUnit *workunit, const char *process)
 {
--- a/ecl/eclagent/eclgraph.cpp
+++ b/ecl/eclagent/eclgraph.cpp
@@ -1519,13 +1519,17 @@ void EclAgent::executeThorGraph(const char * graphName)
     WUState state = WUStateUnknown;
     if (agentTopology->hasProp("@queue"))
     {
-        VStringBuffer queueName("%s.thor", agentTopology->queryProp("@queue"));
-        DBGLOG("Queueing wuid=%s, graph=%s, on queue=%s, timelimit=%u seconds", wuid.str(), graphName, queueName.str(), timelimit);
-        Owned<IJobQueue> queue = createJobQueue(queueName.str());
-        queue->connect(false);
-        VStringBuffer jobName("%s/%s", wuid.get(), graphName);
-        IJobQueueItem *item = createJobQueueItem(jobName);
-        queue->enqueue(item);
+        if (executeGraphOnLingeringThor(*queryWorkUnit(), graphName))
+            PROGLOG("Existing lingering Thor handled graph: %s", graphName);
+        else
+        {
+            VStringBuffer queueName("%s.thor", agentTopology->queryProp("@queue"));
+            DBGLOG("Queueing wuid=%s, graph=%s, on queue=%s, timelimit=%u seconds", wuid.str(), graphName, queueName.str(), timelimit);
+            Owned<IJobQueue> queue = createJobQueue(queueName.str());
+            VStringBuffer jobName("%s/%s", wuid.get(), graphName);
+            IJobQueueItem *item = createJobQueueItem(jobName);
+            queue->enqueue(item);
+        }
 
         // NB: overall max runtime if guillotine set handled by abortmonitor
         unsigned runningTimeLimit = queryWorkUnit()->getDebugValueInt("maxRunTime", 0);
@@ -1572,8 +1576,13 @@ void EclAgent::executeThorGraph(const char * graphName)
             }
         }
     }
-    else if (WUStateFailed == state)
-        throw makeStringException(0, "Workunit failed");
+    else
+    {
+        WorkunitUpdate w = updateWorkUnit();
+        WUState state = w->getState();
+        if (WUStateFailed == state)
+            throw makeStringException(0, "Workunit failed");
+    }
 
     setRunning();
     unlockWorkUnit();
--- a/roxie/ccd/ccdcontext.cpp
+++ b/roxie/ccd/ccdcontext.cpp
@@ -2280,16 +2280,19 @@ protected:
         // NB: If a single Eclagent were to want to launch >1 Thor, then the threading could be in the workflow above this call.
         setWUState(WUStateBlocked);
 
-        WUState state = WUStateUnknown;
         if (queryComponentConfig().hasProp("@queue"))
         {
-            VStringBuffer queueName("%s.thor", queryComponentConfig().queryProp("@queue"));
-            DBGLOG("Queueing wuid=%s, graph=%s, on queue=%s, timelimit=%u seconds", wuid.str(), graphName, queueName.str(), timelimit);
-            Owned<IJobQueue> queue = createJobQueue(queueName.str());
-            queue->connect(false);
-            VStringBuffer jobName("%s/%s", wuid.get(), graphName);
-            IJobQueueItem *item = createJobQueueItem(jobName);
-            queue->enqueue(item);
+            if (executeGraphOnLingeringThor(*workUnit, graphName))
+                PROGLOG("Existing lingering Thor handled graph: %s", graphName);
+            else
+            {
+                VStringBuffer queueName("%s.thor", queryComponentConfig().queryProp("@queue"));
+                DBGLOG("Queueing wuid=%s, graph=%s, on queue=%s, timelimit=%u seconds", wuid.str(), graphName, queueName.str(), timelimit);
+                Owned<IJobQueue> queue = createJobQueue(queueName.str());
+                VStringBuffer jobName("%s/%s", wuid.get(), graphName);
+                IJobQueueItem *item = createJobQueueItem(jobName);
+                queue->enqueue(item);
+            }
 
             unsigned runningTimeLimit = workUnit->getDebugValueInt("maxRunTime", 0);
             runningTimeLimit = runningTimeLimit ? runningTimeLimit : INFINITE;
@@ -2335,8 +2338,13 @@ protected:
                 }
             }
         }
-        else if (WUStateFailed == state)
-            throw makeStringException(0, "Workunit failed");
+        else
+        {
+            WorkunitUpdate w(&workUnit->lock());
+            WUState state = w->getState();
+            if (WUStateFailed == state)
+                throw makeStringException(0, "Workunit failed");
+        }
 
         setWUState(WUStateRunning);
 
@@ -3088,6 +3096,11 @@ public:
             debugContext->debugTerminate();
         if (workUnit)
         {
+#ifdef _CONTAINERIZED
+            // signal to any lingering Thor's that job is complete and they can quit before timeout.
+            executeGraphOnLingeringThor(*workUnit, nullptr);
+#endif
+
             if (options.failOnLeaks && !failed)
             {
                 cleanupGraphs();
--- a/thorlcr/master/thgraphmanager.cpp
+++ b/thorlcr/master/thgraphmanager.cpp
@@ -1088,11 +1088,13 @@ void thorMain(ILogMsgHandler *logHandler, const char *wuid, const char *graphNam
     {
         Owned<CDaliConnectionValidator> daliConnectValidator = new CDaliConnectionValidator(globals->getPropInt("@verifyDaliConnectionInterval", DEFAULT_VERIFYDALI_POLL));
         Owned<ILargeMemLimitNotify> notify;
-        if (multiThorMemoryThreshold) {
+        if (multiThorMemoryThreshold)
+        {
             StringBuffer ngname;
             if (!globals->getProp("@multiThorResourceGroup",ngname))
                 globals->getProp("@nodeGroup",ngname);
-            if (ngname.length()) {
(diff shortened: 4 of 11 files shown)


jakesmith:

thorlcr/thorutil/thormisc.hpp
```diff
@@ -321,6 +321,7 @@ class graph_decl CStreamFileOwner : public CSimpleInterfaceOf<IExtRowStream>
 #define DEFAULT_THORSLAVEPORT 20100
 #define DEFAULT_SLAVEPORTINC 20
 #define DEFAULT_QUERYSO_LIMIT 10
+#define DEFAULT_LINGER_SECS 60
```
richardkchapman: Feels a little long - we would expect next graph to come fairly immediately if it's coming, wouldn't we? I don't know what it costs to leave a cloud thor up for an extra minute, and I suppose in normal usage you'll tell it not to stay up unless there IS another graph coming, but...

richardkchapman:
@mckellyln Please review

thorlcr/thorutil/thormisc.hpp
jakesmith: It is signalled at end of job to stop regardless.

I don't know how long is normal really, every job is different, it's not unusual though for graph1 to run, then for it to spend sometime performing e.g. superfile operations, or blocked on some locked file for a short while (sometimes a long while), or e.g. perform a spray.

jakesmith:
@AttilaVamos - can you see what the smoketest failure is?

@mckellyln Please review

AttilaVamos:
@jakesmith yes it was a problem with environment/Smoketest clean-up. Fixed.

jakesmith:
retagging @mckellyln - please review

ecl/eclagent/eclgraph.cpp
```diff
@@ -1519,13 +1519,17 @@ void EclAgent::executeThorGraph(const char * graphName)
...
-        VStringBuffer jobName("%s/%s", wuid.get(), graphName);
-        IJobQueueItem *item = createJobQueueItem(jobName);
-        queue->enqueue(item);
+        if (executeGraphOnLingeringThor(*queryWorkUnit(), graphName))
+            PROGLOG("Existing lingering Thor handled graph: %s", graphName);
+        else
+        {
+            VStringBuffer queueName("%s.thor", agentTopology->queryProp("@queue"));
+            DBGLOG("Queueing wuid=%s, graph=%s, on queue=%s, timelimit=%u seconds", wuid.str(), graphName, queueName.str(), timelimit);
+            Owned<IJobQueue> queue = createJobQueue(queueName.str());
```
mckellyln: Do we need a queue->connect() here ?

jakesmith: no, it was there before but it was a mistake, and actually cause a problem on shutdown (that was hidden by container shutdown).
Only need it if wanting to listen to the queue, not post to it as it is in this context.

thorlcr/master/thmastermain.cpp
```diff
@@ -650,14 +646,22 @@ int main( int argc, const char *argv[]  )
...
-                unsigned port = getFixedPort(TPORT_mp);
-                LOG(MCdebugProgress, thorJob, "calling initClientProcess Port %d", port);
-                initClientProcess(serverGroup, DCR_ThorMaster, port);
+        for (;;)
+        {
+            try
+            {
+                LOG(MCdebugProgress, thorJob, "calling initClientProcess %d", thorEp.port);
+                initClientProcess(serverGroup, DCR_ThorMaster, thorEp.port);
+                if (0 == thorEp.port)
```
mckellyln: Not sure thorEp.port can be zero as that is handled above (line 615?)

thorlcr/slave/thslavemain.cpp
```diff
@@ -379,12 +383,14 @@ int main( int argc, const char *argv[]  )
         // TBD: use new config/init system for generic handling of init settings vs command line overrides
         if (0 == slfEp.port) // assume default from config if not on command line
             slfEp.port = globals->getPropInt("@slaveport", THOR_BASESLAVE_PORT);
+
+        startMPServer(DCR_ThorSlave, slfEp.port, false);
+        if (0 == slfEp.port)
```
mckellyln: same ques as above, can this be ?

thorlcr/master/thmastermain.cpp
jakesmith: It can be zero if specifically set to 0 (meaning dynamic).
Which is what the helm charts do if thormaster is running in containerPerAgent: false,
which means there could be multiple thormasters in the same container and hence dynamic ports (in same way eclagent does) is what is wanted.

thorlcr/slave/thslavemain.cpp
jakesmith: see answer above.

thorlcr/master/thgraphmanager.cpp
```diff
@@ -1109,31 +1111,89 @@ void thorMain(ILogMsgHandler *logHandler, const char *wuid, const char *graphNam
...
+                        if (!queryWorldCommunicator().recv(msg, NULL, MPTAG_THOR, nullptr, remaining))
+                            break;
+                        msg.read(next);
+
+                        // validate
+                        StringArray sArray;
+                        sArray.appendList(next, "/");
+                        if (2 == sArray.ordinality() && streq(sArray.item(0), wuid))
+                        {
+                            currentGraphName.set(sArray.item(1));
```
mckellyln: Do we break out here if currentGraphName is empty ("") ?

jakesmith: Yes see line 1193 below.

roxie/ccd/ccdcontext.cpp
```diff
@@ -2280,16 +2280,19 @@ class CRoxieContextBase : implements IRoxieSlaveContext, implements ICodeContext
...
         {
-            VStringBuffer queueName("%s.thor", queryComponentConfig().queryProp("@queue"));
-            DBGLOG("Queueing wuid=%s, graph=%s, on queue=%s, timelimit=%u seconds", wuid.str(), graphName, queueName.str(), timelimit);
-            Owned<IJobQueue> queue = createJobQueue(queueName.str());
-            queue->connect(false);
-            VStringBuffer jobName("%s/%s", wuid.get(), graphName);
-            IJobQueueItem *item = createJobQueueItem(jobName);
-            queue->enqueue(item);
+            if (executeGraphOnLingeringThor(*workUnit, graphName))
+                PROGLOG("Exiting lingering Thor handled graph: %s", graphName);
```
mckellyln: Exiting or Executing ?

jakesmith: should be "Existing"
Will fix.

thorlcr/master/thgraphmanager.cpp
mckellyln: Yes, sorry I read the break at line 1179 as part of the if on line 1177

mckellyln:
@jakesmith looks good.  I suppose the only way around the "if I waited one more second I would have prevented starting a whole new Thor" would be to change the model from wait and see to something were Thor is told more graphs are coming after this one, but that is a larger concept for another day.  I approve.

mckellyln (approved):

mckellyln:
But fix typo and squash

jakesmith:
>would be to change the model from wait and see to something were Thor is told more graphs are coming after this one

yes, although in general it (workflow) doesn't know. e.g. runs graph1, then does some conditional stuff and _may_ run graph2.

jakesmith:
@richardkchapman - rebased to candidate-7.8

ready to merge, but may want to leave until smoketest passes again, just in case!

thorlcr/master/thgraphmanager.cpp
```diff
@@ -1109,31 +1111,89 @@ void thorMain(ILogMsgHandler *logHandler, const char *wuid, const char *graphNam
...
+                        // reject/ignore duff message.
+                    }
+
+                    // remove lingering instance from workunit
+                    Owned<IWorkUnitFactory> factory;
+                    Owned<IConstWorkUnit> workunit;
+                    factory.setown(getWorkUnitFactory());
+                    workunit.setown(factory->openWorkUnit(wuid));
+                    Owned<IWorkUnit> w = &workunit->lock();
+                    w->setDebugValue(instance, "0", true);
```
ghalliday: Who's responsibility is it to set this value?  I think it should only do this if the workunit has timed out - otherwise the code in executeGraphOnLingeringThor() will have already cleared it, and this will be locking/unlocking the workunit for no benefit.

ghalliday: This will also potentially be blocked by the eclagent still holding a IWorkunit, and could also deadlock if there were multiple lingering thors...  Deadlock on IWorkUnit locks need more thought/careful review.

ghalliday: Even worse... if this is timing out because eclagent is in a long operation then I think eclagent will still have a read lock on the workunit (need to check carefully), which means this will probably deadlock.

ghalliday (commented):
@jakesmith a few comments.  Main concerns are over locking of workunits - which may require some carefull thought.

thorlcr/master/thgraphmanager.cpp
```diff
@@ -1109,31 +1111,89 @@ void thorMain(ILogMsgHandler *logHandler, const char *wuid, const char *graphNam
...
+                        }
+                        // reject/ignore duff message.
+                    }
+
+                    // remove lingering instance from workunit
+                    Owned<IWorkUnitFactory> factory;
+                    Owned<IConstWorkUnit> workunit;
+                    factory.setown(getWorkUnitFactory());
+                    workunit.setown(factory->openWorkUnit(wuid));
+                    Owned<IWorkUnit> w = &workunit->lock();
```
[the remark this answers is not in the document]
jakesmith: yes that's going to be a problem in several places I think.
We should be getting read than write, should be blocking trying to get exclusive lock without read lock - at least in some cases.

thorlcr/master/thgraphmanager.cpp
```diff
@@ -1109,31 +1111,89 @@ void thorMain(ILogMsgHandler *logHandler, const char *wuid, const char *graphNam
         Owned<CJobManager> jobManager = new CJobManager(logHandler);
         try
         {
-            if (wuid) // one-shot, quits after running
+#ifndef _CONTAINERIZED
+            jobManager->run();
+#else
+            unsigned lingerPeriod = globals->getPropInt("@lingerPeriod", DEFAULT_LINGER_SECS*1000);
```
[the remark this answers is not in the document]
jakesmith: That is true, but and end of job it does signal to the Thor(s) and terminates them.
So they (or 1 if not parallel workflow) linger whilst job is running, reuse the 1 if lingering, and shut it down when job is finished (when there are no more graphs coming).

if you knew it really was last graph (no conditional possible graphs to come), then yes a 'thisislast' hint would be useful.

common/workunit/workunit.cpp
```diff
@@ -13192,6 +13192,51 @@ bool isValidMemoryValue(const char *memoryUnit)
...
 
 #ifdef _CONTAINERIZED
+bool executeGraphOnLingeringThor(IConstWorkUnit &workunit, const char *graphName)
+{
+    // check if lingering thor instance is up.
+    Owned<IStringIterator> iter = &workunit.getDebugValues("thorinstance_*");
+    ForEach(*iter)
+    {
+        /* NB: Thor's set their running endpoint into Debug values of workunit
+            * Check to see if workunit has any Thor availble lingering instances.
```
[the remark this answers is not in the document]
jakesmith: will fix

common/workunit/workunit.cpp
```diff
@@ -13192,6 +13192,51 @@ bool isValidMemoryValue(const char *memoryUnit)
...
+        if (strToBool(thorInstanceValue.s))
+        {
+            {
+                Owned<IWorkUnit> w = &workunit.lock();
+                w->setDebugValue(thorInstance.s, "0", true);
+            }
+            /* NB: there's a window where Thor could shutdown here.
+             * In that case, the sendRecv will fail and it will fall through to queueing.
+             */
+            StringArray sArray;
```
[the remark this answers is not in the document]
jakesmith: yes agree, I'll change it.

common/workunit/workunit.cpp
```diff
@@ -13192,6 +13192,51 @@ bool isValidMemoryValue(const char *memoryUnit)
...
+    Owned<IStringIterator> iter = &workunit.getDebugValues("thorinstance_*");
+    ForEach(*iter)
+    {
+        /* NB: Thor's set their running endpoint into Debug values of workunit
+            * Check to see if workunit has any Thor availble lingering instances.
+            */
+        SCMStringBuffer thorInstance;
+        iter->str(thorInstance);
+        SCMStringBuffer thorInstanceValue;
+        workunit.getDebugValue(thorInstance.s, thorInstanceValue);
```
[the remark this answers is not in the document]
jakesmith: yes, I'll change.

thorlcr/master/thgraphmanager.cpp
jakesmith: > if this is timing out because eclagent is in a long operation then I think eclagent will still have a read lock on the workunit

@ghalliday - does eclagent (or anything) keep a longer term lock on the workunit?

jakesmith: If it does, that causes all sorts of problems as soon as it reaches out to anything else which also wants to update the workunit..

thorlcr/master/thgraphmanager.cpp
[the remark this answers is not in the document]
jakesmith: In fact, the workunit locking model is not read->write->read, it's "no lock"->write->"no lock"

thorlcr/master/thgraphmanager.cpp
jakesmith: Looking workunit (openWorkUnit) and the misleading named wuRead.. it is never read locked.
openWorkUnit() does not establish a lock, only a connection.

When eclagent or anything else needs to update it calls lock() and gets an exclusive (write) lock to the workunit.

So that does mean in many places (existing eclagent code and other places), a connection to a workunit could be being read, whilst another thread/process is updating the workunit..
Not great, but not new..

It also means that a reader (without a read lock) doesn't see new changes.. I think that's why there's forceReload() type calls in various places.

I think it is assumed in general, that only 1 thread or 1 process is reading or writing.

This whole area could do with revisiting, but beyond the scope of this PR.

ghalliday: So quite a few objections don't actually hold.  I'll look at the PR again to see if that introduces other complications.

thorlcr/master/thgraphmanager.cpp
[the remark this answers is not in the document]
ghalliday: Agreed.  The problem isn't if the job terminates, it is if it is still running.  It will block this thor for the linger period.    I suspect a much smaller linger period might be a better default.

thorlcr/master/thgraphmanager.cpp
ghalliday: @jakesmith I think the original comment still stands.  Either eclagent should set this to 0, or thor should (in the non-timeout case), but currently both try to.

common/workunit/workunit.cpp
```diff
@@ -13192,6 +13192,51 @@ bool isValidMemoryValue(const char *memoryUnit)
 }
 
 #ifdef _CONTAINERIZED
+bool executeGraphOnLingeringThor(IConstWorkUnit &workunit, const char *graphName)
+{
+    // check if lingering thor instance is up.
+    Owned<IStringIterator> iter = &workunit.getDebugValues("thorinstance_*");
```
ghalliday: Should this be reloading the workunit (especially the debug settings) before checking.  Otherwise I think it may still have the old values (e.g. after the graph completes), so it will always be 1 and wait for 10s before it realises it has gone.

I wonder if any data used for communicating state should *not* be in the workunit, but should be stored somewhere else.  Otherwise do you need to invalidate the whole workunit to update that state?  Or is the update magically incremental?

jakesmith: >Should this be reloading the workunit (especially the debug settings) before checking.

as we were talking offline - I was wondering same.. I'll look.

>I wonder if any data used for communicating state should not be in the workunit

Somewhere else in Dali for example?
It wants to be couple with the workunit somehow..


The workunit inferface has a forceReload() method, which does invalidate the whole connection cache.
The connection interface though (IRemoteConnection) can invalidate just parts of it, so could e.g. clear the Debug/ cache and would therefore force it to reload next time they were queried.

thorlcr/master/thgraphmanager.cpp
jakesmith: I had only intended to clear it on timeout, not when it had received next graph, where it had already been cleared.. so yeah I agree, will fix that.

common/workunit/workunit.cpp
richardkchapman: There is precendent for wu info being stored outside the workunit because it needs to be under a different locking regime - I think (from memory) we keep the state and/or abort requests outside

thorlcr/master/thgraphmanager.cpp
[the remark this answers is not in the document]
jakesmith: Reducing to 10s

jakesmith:
@ghalliday - please see new commit changes and replies above.

PR #13552, opened 2020-03-26
HPCC-22927 File service functions that can point to an external Esp should optionally take credentials

AttilaVamos:
Update functions with username and password:

- fSprayFixed
- SprayFixed
- fSprayVariable
- SprayVariable
- fSprayDelimited
- SprayDelimited
- fSprayXml
- SprayXml
- fDespray
- Despray
- fCopy
- Copy
- fReplicate
- Replicate
- fMonitorLogicalFileName
- MonitorLogicalFileName
- fMonitorFile
- MonitorFile
- WaitDfuWorkunit
- AbortDfuWorkunit

Signed-off-by: Attila Vamos attila.vamos@gmail.com

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

## Testing:

Tested manually

### DIFF
--- a/cmake_modules/go_rc.sh
+++ b/cmake_modules/go_rc.sh
@@ -30,6 +30,7 @@ if [ "$HPCC_MATURITY" = "closedown" ] ; then
     fi
   else
     NEW_POINT=$((HPCC_POINT+1))
+    NEW_MINOR=$HPCC_MINOR
   fi
   if [ "$GIT_BRANCH" != "candidate-$HPCC_MAJOR.$NEW_MINOR.x" ]; then
     echo "Current branch should be candidate-$HPCC_MAJOR.$NEW_MINOR.x"
@@ -56,6 +57,7 @@ else
     exit 2
   fi
   NEW_POINT=$HPCC_POINT
+  NEW_MINOR=$HPCC_MINOR
   NEW_SEQUENCE=$((HPCC_SEQUENCE+1))
 fi
 


AttilaVamos:
@jakesmith please review.

roxie/udplib/udptrr.cpp
```diff
@@ -466,54 +466,66 @@ class CReceiveManager : implements IReceiveManager, public CInterface
             adjustPriority(1);
         #endif
             UdpRequestToSendMsg msg;
-            unsigned timeout = 5000;  // Is this too long?
+            unsigned timeoutExpires = msTick() + 5000;
```
jakesmith: @AttilaVamos - looks like this PR includes some unintentional changes.

jakesmith (changes requested):

jakesmith:
Have added a few comments, but there's issue around storing password in the DFU workunit in general.

We should not be storing password in the workunit.
It looks like setUserAuth might do too already, but if so that needs fixing as well, it's a security leak.

And I don't think it is needed anyway.
The places where getPassword is called and a IUserDescriptor is formed/set, it is then passed to DFS,
but it ends up calling serializeWithoutPassword ...
Because Dali only does LDAP lookups based on username.

It's okay to pass the password to the esp service, but not to store it in the workunit.
So we should not add password to workunit here, and I think it will still work without it.

@AttilaVamos - can you check, remove the code that stores the password in the DFU workunit(s) and retreives it, and test if it continues to work as expected?

PR #13553, opened 2020-03-26
HPCC-23348 Roxie may lock up under load

richardkchapman:
Further fixes - if queue is entirely filled with aborted packets we would lock
up.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [x] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/roxie/udplib/udpsha.cpp
+++ b/roxie/udplib/udpsha.cpp
@@ -97,14 +97,6 @@ queue_t::~queue_t()
     delete [] elements; 
 }
 
-bool queue_t::empty() 
-{
-    c_region.enter();
-    bool res = (active_buffers == 0);
-    c_region.leave();
-    return res;
-}
-
 int queue_t::free_slots() 
 {
     int res=0;
@@ -148,9 +140,10 @@ void queue_t::pushOwn(DataBuffer *buf)
     data_avail.signal();
 }
 
-DataBuffer *queue_t::pop()
+DataBuffer *queue_t::pop(bool block)
 {
-    data_avail.wait();
+    if (!data_avail.wait(block ? INFINITE : 0))
+        return nullptr;
     DataBuffer *ret = NULL; 
     bool must_signal;
     {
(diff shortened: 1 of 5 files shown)


roxie/udplib/udptrs.cpp
```diff
@@ -266,49 +265,45 @@ class UdpReceiverEntry : public IUdpReceiverEntry
...
+                    packetsQueued--;
+                    return buffer;
                 }
             }
-            // If we get here, it suggests we were told to get a buffer but no queue has one
-            // Should never happen
-            MilliSleep(10);
-            DBGLOG("UdpSender: ------------- this code should never execute --------------- ");
         }
+        // If we get here, it suggests we were told to get a buffer but no queue has one. Should be rare but possible if data gets removed following an abort
```
ghalliday: It may be worth keeping a very short sleep here, if (packetsQueues != 0), otherwise this will potentially spin for a while.

roxie/udplib/udpsha.cpp
```diff
@@ -171,29 +164,48 @@ DataBuffer *queue_t::pop()
 }
 
 
-bool queue_t::removeData(const void *key, PKT_CMP_FUN pkCmpFn)
+unsigned queue_t::removeData(const void *key, PKT_CMP_FUN pkCmpFn)
 {
-    bool ret = false;
-    CriticalBlock b(c_region);
-    if (active_buffers) 
+    unsigned ret = 0;
```
ghalliday: picky: Not the most descriptive variable name.

ghalliday (commented):
@richardkchapman two minor comments

roxie/udplib/udptrs.cpp
richardkchapman: I think to get there (now) it would imply that packetsQueued was non-zero before deciding to make the call but (in the intervening small window) all the packets were removed. If packetsQueued was non-zero at the point we get here it would imply that someone had then added something else, and we shouldn't delay...

roxie/udplib/udpsha.cpp
richardkchapman: renamed and repushed

roxie/udplib/udptrs.cpp
ghalliday: I was concerned about the (tiny) window, where the elements have all been removed from the queues but the packet count hasn't been updated (and that thread has then been unscheduled).

I think not as unlikely as it sounds because this thread is a higher priority, so it make take precedence as soon as the critical section on the queue is released.

ghalliday:
@mckellyln I think this now looks good.  Please cast your eye over it as well.

PR #13574, opened 2020-03-31
HPCC-23773 Remove unwanted filter data

miguelvazq:
Fixes: HPCC-23772
FIxes: HPCC-23636

- Removed additional data that was being passed in by other calls from
toObject()
- Removed consecutive recent filters from appearing

Signed-off by: Miguel Vazquez <miguel.vazquez@lexisnexisrisk.com>

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/src/eclwatch/FilterDropDownWidget.js
+++ b/esp/src/eclwatch/FilterDropDownWidget.js
@@ -69,6 +69,8 @@ define([
         },
 
         _onFilterApply: function (event) {
+            var formData = this.toObject(event.currentTarget.form);
+            this.initRecentFilter(formData);
             this.filterDropDown.closeDropDown();
             this.emit("apply");
             this.refreshState();
@@ -105,13 +107,13 @@ define([
             return false;
         },
 
-        toObject: function () {
+        toObject: function (filter) {
             var context = this;
             if (this.filterDropDown.get("disabled")) {
                 return {};
             }
             var retVal = {};
-            arrayUtil.forEach(this.filterForm.getDescendants(), function (item, idx) {
+            arrayUtil.forEach(this.filterForm.getDescendants() || filter, function (item, idx) {
                 var name = item.get("name");
                 if (name) {
                     var value = item.get("value");
@@ -120,15 +122,6 @@ define([
                     }
                 }
             });
-            if (this.userName !== null) {
-                if (!Utility.isObjectEmpty(retVal)) {
-                    Recent.addToStack(this.params.ws_key, retVal, 5).then(function (val) {
-                        if (val) {
-                            context.loadRecentFilters(retVal);
-                        }
-                    });
-                }
-            }
             return retVal;
         },
 
@@ -150,11 +143,13 @@ define([
                 return;
 
             this.userName = dojoConfig.username;
+            var recentFilterLoaded = false;
             if (this.userName !== null) {
                 this.recentFilterNode = dom.byId(this.id + "RecentFilters");
 
-                if (params.widget) {
+                if (params.widget && !recentFilterLoaded) {
                     srcReact.render(srcReact.RecentFilters, { ws_key: params.ws_key, widget: params.widget, filter: {} }, this.recentFilterNode);
+                    recentFilterLoaded = true;
                 }
             }
         },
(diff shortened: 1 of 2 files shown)


miguelvazq:
@GordonSmith Please review.

miguelvazq:
@GordonSmith this error look familiar to you in regards to dojo-util?

esp/src/eclwatch/FilterDropDownWidget.js
```diff
@@ -105,13 +107,13 @@ define([
...
 
-        toObject: function () {
+        toObject: function (filter) {
             var context = this;
             if (this.filterDropDown.get("disabled")) {
                 return {};
             }
             var retVal = {};
-            arrayUtil.forEach(this.filterForm.getDescendants(), function (item, idx) {
+            arrayUtil.forEach(this.filterForm.getDescendants() || filter, function (item, idx) {
```
GordonSmith, line 116: When does `filter` actually come into play?  Doesn't `this.filterForm.getDescendants()` always return something?

miguelvazq: When a user physically click apply. Rather than different publishers, queries that call toObject()

esp/src/src/Utility.ts
```diff
@@ -273,6 +273,24 @@ export function isObjectEmpty(obj) {
...
+
+    for (let i = 0; i < obj1Props.length; ++i) {
+        const propName = obj1Props[i];
+
+        if (obj1[propName] !== obj2[propName]) {
+            return false;
+        }
+    }
+    return true;
+}
```
GordonSmith, line 292: This won't catch the following case:
```
obj1 ={
  a:1,
  b:2,
  c:3,
  e: undefined
}
obj2 ={
  a:1,
  b:2,
  c:3
  d:4
}
```

FYI @hpcc-js/util has a hashSum function which will create deep hashes of objects which you can then compare (which may not be what you want either).

GordonSmith (changes requested):
Two questions...

esp/src/src/Utility.ts
miguelvazq: Can I compare equality of two objects with that hashSum? I need to make sure that they do no match to allow the insertion into the store otherwise keep out.

GordonSmith: Yes - but its a deep compare, so you need to be a bit careful as to what you pass in.,,

GordonSmith: `hashSum(obj1) === hashSum(obj2)`

richardkchapman:
Is yhis supposed to target 7.10 ?

GordonSmith:
@miguelvazq FYI I rebased onto 7.8.x,

esp/src/src/Utility.ts
miguelvazq: Our filter forms usually are as follows:

```
{
    WUID: "xxxx-abc",
    PublishedBy: "mvazquez",
}
```

They usually do not go any deeper than that.

version.cmake
```diff
@@ -4,8 +4,8 @@
...
 set ( HPCC_PROJECT "community" )
 set ( HPCC_MAJOR 7 )
-set ( HPCC_MINOR 8 )
-set ( HPCC_POINT 1 )
-set ( HPCC_MATURITY "closedown" )
+set ( HPCC_MINOR 9 )
+set ( HPCC_POINT 0 )
+set ( HPCC_MATURITY "trunk" )
 set ( HPCC_SEQUENCE 0 )
 ###
```
GordonSmith, line 11: @miguelvazq can you revert this file?

cmake_modules/commonSetup.cmake
```diff
@@ -488,6 +488,11 @@ IF ("${COMMONSETUP_DONE}" STREQUAL "")
...
   endif ()
 
+  if (CMAKE_COMPILER_IS_GNUCXX)
+    #Ensure that missing symbols are reported as errors at link time (default for osx/windows)
+    SET (CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,-z,defs")
+    SET (CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,-z,defs")
+  endif()
   macro(HPCC_ADD_EXECUTABLE target)
     add_executable(${target} ${ARGN})
   endmacro(HPCC_ADD_EXECUTABLE target)
```
GordonSmith, line 498: @miguelvazq can you revert this file?

GordonSmith (approved):
@richardkchapman ready to pull.

PR #13580, opened 2020-04-02
HPCC-23370 Documentation on how to use minikube

richardkchapman:
Work in progress

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/.github/workflows/build-and-publish-debug.yml
+++ b/.github/workflows/build-and-publish-debug.yml
@@ -25,4 +25,5 @@ jobs:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
           buildtype: Debug
+          latest: 1   # This should only be set on the current minor branch
           
--- a/dockerfiles/buildall.sh
+++ b/dockerfiles/buildall.sh
@@ -60,6 +60,13 @@ set -e
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 pushd $DIR 2>&1 > /dev/null
 
+. ../cmake_modules/parse_cmake.sh
+parse_cmake
+
+if [[ "$HPCC_MATURITY" = "release" ]] && [[ "$INPUT_LATEST" = "1" ]] ; then
+  LATEST = 1
+fi
+
 build_image() {
   local name=$1
   local label=$2
@@ -74,8 +81,15 @@ build_image() {
        --build-arg BUILD_TYPE=${BUILD_TYPE} \
        --build-arg BUILD_THREADS=${BUILD_THREADS} \
        ${name}/ 
-    if [ "$PUSH" = "1" ] ; then
-      docker push hpccsystems/${name}:${label}
+    if [ "$LATEST" = "1" ] ; then
+      docker tag hpccsystems/${name}:${label} hpccsystems/${name}:latest
+      if [ "$PUSH" = "1" ] ; then
+        docker push hpccsystems/${name}:${label} hpccsystems/${name}:latest
+      fi
+    else
+      if [ "$PUSH" = "1" ] ; then
+        docker push hpccsystems/${name}:${label}
+      fi
     fi
   fi
 }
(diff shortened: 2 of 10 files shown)


dockerfiles/hpcc/values.schema.json
```diff
@@ -144,6 +144,10 @@
           "type": "array",
           "items": { "type": "string" }
         },
+        "useChildProcesses": {
+          "type": "boolean",
+          "description": "Launch each workunit compile in its own container, or as a child process"
```
jakesmith, line 149: sounds a little confusing? 
Perhaps "If true, compile in child process if true. If false (default) compile in separate container"

dockerfiles/buildall.sh
```diff
@@ -60,6 +60,13 @@ set -e
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 pushd $DIR 2>&1 > /dev/null
 
+. ../cmake_modules/parse_cmake.sh
+parse_cmake
```
jakesmith, line 64: is this right?

dockerfiles/buildall.sh
```diff
@@ -60,6 +60,13 @@ set -e
...
+
+if [[ "$HPCC_MATURITY" = "release" ]] && [[ "$HPCC_MAJOR" = "7" ]] && [[ "$HPCC_MINOR" = "8" ]] ; then
```
jakesmith, line 66: perhaps the major.minor latest should come from somewhere else, e.g. a ./latest file ?
so buildall.sh doesn't need to change when we want to up it.

dockerfiles/hpcc/templates/thor.yaml
```diff
@@ -5,15 +5,15 @@
 {{- $thorAgentName := printf "%s-thoragent" .name }}
 {{- $slaveName := printf "%s-slave" .name }}
 {{- $serviceName := printf "%s-svc" .name }}
-{{- $eclAgentDefaults := dict "name" $eclAgentName "replicas" 1 }}
+{{- $eclAgentDefaults := dict "name" $eclAgentName "useChildProcesses" true "replicas" 1 }}
```
jakesmith, line 8: if we are defaulting off in general, why not here too?

jakesmith:
@richardkchapman - added a few comments and questions

dockerfiles/buildall.sh
richardkchapman: yes I think so (the .sh file defines a function...)

dockerfiles/buildall.sh
richardkchapman: Need to change a file either way...

The concept is flawed though - tags on the old branch will still be compared against the old version, so will think they are still latest. Really we just want a boolean "is latest" that we set to false on the old branch at the same time we set to true on the new.

dockerfiles/hpcc/templates/thor.yaml
richardkchapman: Because the "workflow" agent running thor jobs is assumed to be lightweight

.github/workflows/build-and-publish-debug.yml
```diff
@@ -25,4 +25,5 @@ jobs:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
           buildtype: Debug
+          latest: 1
```
richardkchapman, line 28: We will need to ensure that this flag is set only in the branch that targets the current minor release. When a new minor release is published, we'll need to remove this from the old one (and set it in the new one)

dockerfiles/hpcc/templates/thor.yaml
```diff
@@ -101,13 +101,13 @@ data:
...
       type: thor
-{{- if not $thorAgentScope.containerPerAgent }}
+{{- if $thorAgentScope.useChildProceses }}
     thor:
 {{ toYaml $thorScope | indent 6 }}
 {{- end }}
     global:
 {{ include "hpcc.generateGlobalConfigMap" $ | indent 6 }}
-{{- if $thorAgentScope.containerPerAgent }}
+{{- if not $eclAgentScope.useChildProcesses }}
```
richardkchapman, line 110: Note - this was testing the wrong variable before

jakesmith (approved):
@richardkchapman - looks good to me

PR #13645, opened 2020-04-17
HPCC-23824 Containerized logging configuration changes

richardkchapman:
## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [x] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

## Testing:
Manual testing and inspection of generated helm charts

### DIFF
--- a/dockerfiles/hpcc/templates/_helpers.tpl
+++ b/dockerfiles/hpcc/templates/_helpers.tpl
@@ -52,6 +52,19 @@ singleNode: {{ .Values.global.singleNode | default false }}
 {{- end -}}
 
 {{/*
+Generate local logging info, merged with global
+Pass in dict with root and me
+*/}}
+{{- define "hpcc.generateLoggingConfig" -}}
+{{- $logging := deepCopy .me | mergeOverwrite .root.Values.global }}
+{{- if hasKey $logging "logging" }}
+logging:
+{{ toYaml $logging.logging | indent 2 }}
+{{- end -}}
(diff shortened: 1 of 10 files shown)


richardkchapman:
@rpastrana based on your PR, what do you think?

richardkchapman:
@ghalliday @jakesmith for comment

jakesmith:
Looks like a sensible change to me.
It's good to have a global default, 1 place to change it, and for components to individually be able to override.

One question I have, if a component does override (e.g. detail:) in it's defaultYaml - then it's default will be now be overridden by the global default.
Saying that, I don't think the defaultYaml's should be setting these properties, there are currently two places that do, but set them to the standard internal default anyway, so could be deleted.

rpastrana:
@richardkchapman this change seems to follow the standard.
jLog does seem to set defaults for detail in code, perhaps we do not need to set the global default in the values yaml. In fact some of the default yaml entries @jakesmith mentioned above seem to superfluously set details as well, I'll clean that up.

jakesmith:
@rpastrana - worth deleting the overrides in the component defaults that don't change the base default.
I think I'd leave the global setting in, you could comment it out and change comment to say override to change default, but it's probably clearer as it is.

rpastrana:
@richardkchapman back to you

richardkchapman:

PR #13669, opened 2020-04-24
HPCC-23906 Add an option (-ftimeRegex) to time regular expressions

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/ecl/hqlcpp/hqlhtcpp.cpp
+++ b/ecl/hqlcpp/hqlhtcpp.cpp
@@ -18337,6 +18337,25 @@ void HqlCppTranslator::doBuildNewRegexFindReplace(BuildCtx & ctx, const CHqlBoun
     // Because the search instance is created locally, the search parameter is always going to be valid
     // as long as the find instance.  Only exception could be if call created a temporary class instance.
     bool cloneSearch = false;
+    CHqlBoundExpr boundTimer, boundStart;
+    if (options.timeRegex)
+    {
+        StringBuffer regexName;
+        regexName.append(getOpString(expr->getOperator()));
+
+        if (pattern->queryId())
+            regexName.append("(").append(str(pattern->queryId())).append(")");
+        else if (pattern->isConstant())
+        {
+            StringBuffer patternText;
+            getUTF8Value(patternText, pattern);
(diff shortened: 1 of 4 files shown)


ghalliday:
@shamser please review.  Note this is disabled by default in 7.8, but I have added a jira to enable by default it in 7.10 (but some user interface issues could really do with improving before then).

ecl/hqlcpp/hqlhtcpp.cpp
```diff
@@ -18337,6 +18337,25 @@ void HqlCppTranslator::doBuildNewRegexFindReplace(BuildCtx & ctx, const CHqlBoun
     // Because the search instance is created locally, the search parameter is always going to be valid
     // as long as the find instance.  Only exception could be if call created a temporary class instance.
     bool cloneSearch = false;
+    CHqlBoundExpr boundTimer, boundStart;
+    if (options.timeRegex)
```
shamser, line 18341: The compilation of regex expression can take some time.  Not sure if that time is significant.  However, if this section is moved before doBuildRegexCompileInstance (line18322) then the timer accounts for the regex compile tome.

shamser (changes requested):
@ghalliday

ecl/hqlcpp/hqlhtcpp.cpp
ghalliday: That is true.  However the compile time would need to be recorded separately since it is generally done once in a different context.  If the regex is a constant and the search is done many times it is less significant.
It would potentially be useful to time the subgraph start/create time as well (possibly more so) since it would include the regex compile times.

ghalliday:
@shamser added HPCC-23964 to think about timing subgraph startup.  I would think about revisiting regex compile time if that proved to be significant.
(This was added because boost regexs show significant contention on a multi channel setup.)

shamser (approved):
@richardkchapman Reviewed. Looks good.

PR #13682, opened 2020-04-29
HPCC-23915 ESP can fail to restart after startup exception

mckellyln:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/platform/espp.cpp
+++ b/esp/platform/espp.cpp
@@ -446,6 +446,8 @@ int init_main(int argc, char* argv[])
         return -1;
     }
 
+    writeSentinelFile(sentinelFile);
+
     if (config && config->isValid())
     {
         PROGLOG("Configuring Esp Platform...");
@@ -475,8 +477,6 @@ int init_main(int argc, char* argv[])
             IERRLOG("ESP Unhandled General Exception.");
             return -1;
         }
-
-        writeSentinelFile(sentinelFile);
 
         result = work_main(*config, *server.get());
     }


mckellyln:
NOTE this code change helps the situation some but there are many code paths at startup that could thorw an exception and thus where to retry connections or throw and restart is open for discussion.
I think there seems to be multiple connect()s occuring near the same time concurrently, such as in CSessionCleaner::run() and initDali() that have xpaths "/Sessions" that might be worth looking into also.

mckellyln:
@jakesmith @Michael-Gardner please review.
@stuartort fyi

jakesmith:
I think we intentionally want to create the sentinel after a component is 'setup',
or rather we intentionally want to exit and not restart if it can't pass the initial stages.

I'm not sure the exception that caused this incident (the reason for the JIRA being opened) - should have happened in the early setup stage, i.e. I think should have been later and hence after the sentinel creation.

I'll add some more notes to JIRA.

mckellyln:
The multiple connect()s are probably fine as they do not have identical xpaths.
The code change in this PR moves the creation of the sentinel file earlier in the ESP startup, but there are still code paths that can throw before the sentinel file is written. Perhaps https://track.hpccsystems.com/browse/HPCC-23880 will provide a more complete solution.

mckellyln:
@jakesmith should we still do this PR or is this resolved at a different level now ?

jakesmith:
@mckellyln - I think this PR should be closed. The point at which the sentinel file is created at the moment seems about right, in that it is after initialization.
Now that the issue of the session related exceptions has moved beyond the initialization phase, I think it's okay as it is.

mckellyln:
ok, closing.

PR #13698, opened 2020-04-30
HPCC-23966 Document STD.Crypto Decrypt restriction

JamesDeFabia:
Signed-off-by: Jim DeFabia <jamesdefabia@lexisnexis.com>
@RussWhitehead Please review

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/docs/EN_US/ECLStandardLibraryReference/SLR-Mods/Crypto.xml
+++ b/docs/EN_US/ECLStandardLibraryReference/SLR-Mods/Crypto.xml
@@ -475,7 +475,8 @@ OUTPUT(decryptedStr);</programlisting>
 
     <para>The Decrypt function decrypts the given
     <emphasis>encryptedData</emphasis> using the options defined in the
-    Symmetric Encryption module.</para>
+    Symmetric Encryption module. You can only decrypt data that was encrypted
+    by the Standard Library's Encrypt method. </para>
 
     <para>Example:</para>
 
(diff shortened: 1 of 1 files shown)


RussWhitehead (approved):

RussWhitehead:
@JamesDeFabia Looks good

richardkchapman:
@RussWhitehead What is it we are storing in the encrypted output that makes this true (and why are we storing it there?)

RussWhitehead:
@richardkchapman The encrypted buffer that I return to the caller is an ECL DATA buffer in the format
    Initialization Vector Length
    Initialization Vector - Needed to decrypt
    Decrypted Data Length - Needed to preallocate return buffer
    Encrypted Data Length - Probably wasn't necessary but there nonetheless
    Encrypted Data

richardkchapman:
I'm a bit dubious that the extra info in there actually reduces the security of the encrypted information... Any cryptologists out there?

RussWhitehead:
@dcamper Has a lot of knowledge in this area.  However, without the passphrase, the extra information is of no real use

dcamper:
@richardkchapman The extra metadata does not introduce less security.  It is possible, in fact, to convert an encrypted blob from Std.Crypto's format to the more general format (and back again) without knowing any secrets.  You do need to know what encryption algorithm was used, but that is not a secret.  Because you can perform that translation, it follows that no additional information is needed, so there is no information leak.  Therefore, the security of the two formats is the same.

PR #13700, opened 2020-05-01
HPCC-23975 Convert act. stats to CThorStatsCollection

jakesmith:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jstats.cpp
+++ b/system/jlib/jstats.cpp
@@ -1189,36 +1189,6 @@ static int compareUnsigned(unsigned const * left, unsigned const * right)
     return (*left < *right) ? -1 : (*left > *right) ? +1 : 0;
 }
 
-StatisticsMapping::StatisticsMapping(const std::initializer_list<StatisticKind> &kinds)
-{
-    for (auto kind : kinds)
-    {
-        assert(kind != StKindNone);
-        assert(!indexToKind.contains(kind));
-        indexToKind.append(kind);
-    }
-    createMappings();
-}
-
-StatisticsMapping::StatisticsMapping(const StatisticsMapping * from, const std::initializer_list<StatisticKind> &kinds)
-{
-    ForEachItemIn(idx, from->indexToKind)
-        indexToKind.append(from->indexToKind.item(idx));
-    for (auto kind : kinds)
-    {
-        assert(kind != StKindNone);
-        assert(!indexToKind.contains(kind));
-        indexToKind.append(kind);
-    }
-    createMappings();
-}
-
-StatisticsMapping::StatisticsMapping()
-{
-    for (int i = StKindAll+1; i < StMax; i++)
-        indexToKind.append(i);
-    createMappings();
-}
 
 void StatisticsMapping::createMappings()
 {
@@ -1236,7 +1206,7 @@ void StatisticsMapping::createMappings()
     }
 }
 
-const StatisticsMapping allStatistics;
+const StatisticsMapping allStatistics(StKindAll);
 const StatisticsMapping heapStatistics({StNumAllocations, StNumAllocationScans});
 const StatisticsMapping diskLocalStatistics({StCycleDiskReadIOCycles, StSizeDiskRead, StNumDiskReads, StCycleDiskWriteIOCycles, StSizeDiskWrite, StNumDiskWrites, StNumDiskRetries});
 const StatisticsMapping diskRemoteStatistics({StTimeDiskReadIO, StSizeDiskRead, StNumDiskReads, StTimeDiskWriteIO, StSizeDiskWrite, StNumDiskWrites, StNumDiskRetries});
@@ -3690,3 +3660,5 @@ MODULE_INIT(INIT_PRIORITY_STANDARD)
     return true;
 }
 #endif
+
+
--- a/system/jlib/jstats.h
+++ b/system/jlib/jstats.h
@@ -400,13 +400,32 @@ protected:
 class jlib_decl StatisticsMapping
 {
 public:
-    //Takes a list of StatisticKind
-    StatisticsMapping(const std::initializer_list<StatisticKind> &kinds);
-    //Takes an existing Mapping, and extends it with a list of StatisticKind
-    StatisticsMapping(const StatisticsMapping * from, const std::initializer_list<StatisticKind> &kinds);
-    //Accepts all StatisticKind values
-    StatisticsMapping();
-
+    //Takes a list of StatisticKind and a variable number of existing mappings and combines
+    template <typename... Mappings>
+    StatisticsMapping(const std::initializer_list<StatisticKind> &kinds, const Mappings &... mappings) : StatisticsMapping(&mappings...)
+    {
+        for (auto kind : kinds)
+        {
+            assert((kind != StKindNone) && (kind != StKindAll));
+            assert(!indexToKind.contains(kind));
+            indexToKind.append(kind);
+        }
+        createMappings();
+    }
+    StatisticsMapping(StatisticKind kind)
+    {
+        if (StKindAll == kind)
+        {
+            for (int i = StKindAll+1; i < StMax; i++)
+                indexToKind.append(i);
+        }
+        else
+        {
+            assert(kind != StKindNone);
+            indexToKind.append(kind);
+        }
+        createMappings();
+    }
     inline unsigned getIndex(StatisticKind kind) const
     {
         dbgassertex(kind >= StKindNone && kind < StMax);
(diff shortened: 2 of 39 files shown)


system/jlib/jutil.hpp
```diff
@@ -565,5 +566,17 @@ inline bool stdContains(Container&& container, Value &&v)
...
+class jlib_decl COnScopeExit
+{
+    const std::function<void()> &exitFunc;
+public:
+    inline COnScopeExit(const std::function<void()> &_exitFunc) : exitFunc(_exitFunc) { }
+    inline ~COnScopeExit()
+    {
+        exitFunc();
+    }
+};
```
jakesmith, line 579: @ghalliday - I couldn't see an equivalent of this, it's a frequent pattern, recoded in various places.
Could be a separate PR ?

jakesmith:
@ghalliday - please review

system/jlib/jutil.hpp
ghalliday: Looks like a useful addition.

ghalliday (commented):
@jakesmith I'm just wondering about this.  In roxie the activities have a single set of stats defined in the base class with the mapping passed from the derived activity constructor.  I think that would probably make the code simpler (avoiding derived classes overriding deserialize etc.).

jakesmith:
>I'm just wondering about this. In roxie the activities have a single set of stats defined in the base class with the mapping passed from the derived activity constructor. I think that would probably make the code simpler (avoiding derived classes overriding deserialize etc.).

I also wondered about that.
I was a bigger change though and decided to go this route for now.

Perhaps assuming not other issues, will refactor to consolidate further in a separate PR?
@ghalliday

thorlcr/activities/indexread/thindexreadslave.cpp
```diff
@@ -80,6 +77,24 @@ class CIndexReadSlaveBase : public CSlaveActivity
...
+        unsigned __int64 startSeeks = 0, startScans = 0;
+    public:
+        inline CCaptureIndexStats(CRuntimeStatisticCollection &_stats, StatProvider &_statProvider) : stats(_stats), statProvider(_statProvider)
+        {
+            startSeeks = statProvider.querySeeks();
+            startScans = statProvider.queryScans();
+        }
+        inline ~CCaptureIndexStats()
+        {
+            stats.mergeStatistic(StNumIndexSeeks, statProvider.querySeeks() - startSeeks);
```
ghalliday, line 94: Slightly concerned about the extra overhead of this call.  Might be worth considering a fastpath for this particular case of summing a stat.

thorlcr/activities/keyedjoin/thkeyedjoin.cpp
```diff
@@ -37,7 +36,9 @@ class CKeyedJoinMaster : public CMasterActivity
     MemoryBuffer initMb;
     unsigned numTags = 0;
     std::vector<mptag_t> tags;
-    ProgressInfoArray progressInfoArr;
+    unsigned numStats = 0;
```
ghalliday, line 39: I don't think this is actually used.

ghalliday (commented):
@jakesmith I've scanned it.  It looks like it would be quite a bit simpler if you passed the mapping to the base, but I can review fully if you want to merge as is.

jakesmith:
>It looks like it would be quite a bit simpler if you passed the mapping to the base, but I can review fully if you want to merge as is.

I'll take a look first, perhaps I just had refactor fatigue.

ghalliday:
BTW Richard merged https://github.com/hpcc-systems/HPCC-Platform/pull/13712 which could be used in a couple of places in your code.

jakesmith:
@ghalliday - please review. NB: left 1st commit as it was. 2nd commit is further refactoring to amalgamate the stats into a single collection container in the base, with a single serialize/deserialize.

system/jlib/jstats.cpp
```diff
@@ -1236,7 +1206,18 @@ void StatisticsMapping::createMappings()
...
+{
+public:
+    AllStatisticsMapping()
+    {
+        for (int i = StKindAll+1; i < StMax; i++)
+            indexToKind.append(i);
+        createMappings();
+    }
+} allStatisticsMapping;
+const StatisticsMapping &allStatistics = allStatisticsMapping;
```
ghalliday, line 1220: What is the reason for this change?

system/jlib/jstats.cpp
```diff
@@ -1236,7 +1206,18 @@ void StatisticsMapping::createMappings()
     }
 }
 
-const StatisticsMapping allStatistics;
+
+class AllStatisticsMapping : public StatisticsMapping
```
ghalliday, line 1210: What is the advantage in having a different mapping class?

system/jlib/jstats.h
```diff
@@ -400,13 +400,18 @@ class jlib_decl StatisticsFilter : public CInterface
...
-    //Takes a list of StatisticKind
-    StatisticsMapping(const std::initializer_list<StatisticKind> &kinds);
-    //Takes an existing Mapping, and extends it with a list of StatisticKind
-    StatisticsMapping(const StatisticsMapping * from, const std::initializer_list<StatisticKind> &kinds);
-    //Accepts all StatisticKind values
-    StatisticsMapping();
-
+    //Takes a list of StatisticKind and a variable number of existing mappings and combines
+    template <typename... Mappings>
+    StatisticsMapping(const std::initializer_list<StatisticKind> &kinds, const Mappings &... mappings) : StatisticsMapping(&mappings...)
```
ghalliday, line 405: Changing to a reference makes sense, but why reverse the order, and why inline the function (I guess because it is now a template class?)

system/jlib/jstats.cpp
jakesmith: It's related to the changes in the constructor of StatisticsMapping.
Before the default constructor was the 'all' types constructor.
With the variadic list on the constructor, the default constructor had to be nop.

system/jlib/jstats.h
jakesmith: > why reverse the order

because the tail is a variadic list of existing StatisticMapping's.
e.g. for use case in thor like:
```
const StatisticsMapping joinActivityStatistics({StNumLeftRows, StNumRightRows}, basicActivityStatistics, spillStatistics);
```

>why inline the function (I guess because it is now a template class?)
yes has to be inline because it's a template.

ghalliday:
@jakesmith first impression is that it simplifies the code significantly.  I'm not sure why you have made the all stats change and related items.  (See questions.)  I'll take a look at the commits properly tomorrow.

system/jlib/jstats.cpp
ghalliday: ok, I understand from the weird variadic constructor.

  Another alternative is to use a different constructor.  E.g., a single stat, which if it is StKindAll adds them all.

system/jlib/jstats.h
ghalliday: Very clever, and painful to my brain.

system/jlib/jstats.cpp
jakesmith: yes considered an alternate constructor, but didn't think of a logical one to create.
Your suggestion would work though, if you prefer can change it to that.

jakesmith:
@ghalliday - will tag you back for now. Let me know if you think it would be better to have a separate constructor for 'all' , vs AllStatisticsMapping.

system/jlib/jstats.cpp
ghalliday: I think I would prefer only having a single StatisticsMapping class, so yes please if you are willing.  I don't mind what form the constructor takes (within reason!).

thorlcr/activities/join/thjoinslave.cpp
```diff
@@ -458,7 +457,10 @@ class JoinSlaveActivity : public CSlaveActivity, implements ILookAheadStopNotify
             leftStream.setown(iLoaderL->load(leftInputStream, abortSoon));
             isemptylhs = 0 == iLoaderL->numRows();
             stopLeftInput();
+
+            CRuntimeStatisticCollection spillStats(spillStatistics);
```
ghalliday, line 461: We should introduce
mergeStats(spillStats, iLoadedL, spillStatistics) to avoid the temporary StatsCollection.
Created HPCC-24004 to implement once this is merged

thorlcr/activities/keyedjoin/thkeyedjoinslave-legacy.cpp
```diff
@@ -2418,11 +2418,43 @@ class CKeyedJoinSlave : public CSlaveActivity, implements IJoinProcessor, implem
...
+    virtual void serializeStats(MemoryBuffer &mb) override
     {
-        CSlaveActivity::serializeStats(mb);
         ForEachItemIn(s, _statsArr)
-            mb.append(_statsArr.item(s));
+        {
+            unsigned __int64 v = _statsArr.item(s);
+            if (0 == v)
+                continue;
+            switch (s)
```
ghalliday, line 2428: simplify: better with a
constexpr StatisticKind mapping[] = { StNumIndexSeeks, StNumIndexScans, ... };
and index into that to provide the mapping.
stats.setStatistic(mapping[s]);

ghalliday (commented):
@jakesmith a few minor comments.  It looks like it has cleaned up the code.  I have created a jira HPCC-24004 to cover one further improvement.

jakesmith:
@ghalliday - please see changes in last commit.

PR #13727, opened 2020-05-08
HPCC-24003 Clean up some coverity issues

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dali/base/dasds.cpp
+++ b/dali/base/dasds.cpp
@@ -4832,9 +4832,11 @@ IPropertyTree *loadStore(const char *storeFilename, IPTreeMaker *iMaker, unsigne
 // Not really coalescing, blocking transations and saving store (which will delete pending transactions).
 class CLightCoalesceThread : implements ICoalesce, public CInterface
 {
-    bool stopped, within24;
+    bool stopped = false;
+    bool within24 = false;
     Semaphore sem;
-    unsigned writeTransactionsNow, lastSaveWriteTransactions, lastWarning;
+    unsigned lastSaveWriteTransactions = 0;
+    unsigned lastWarning = 0;
     unsigned idlePeriod, minimumTimeBetweenSaves, idleRate;
     Linked<IPropertyTree> config;
     Owned<IJlibDateTime> quietStartTime, quietEndTime;
@@ -4853,7 +4855,6 @@ public:
     IMPLEMENT_IINTERFACE;
     CLightCoalesceThread(IPropertyTree &_config, IStoreHelper *_iStoreHelper) : config(&_config), iStoreHelper(_iStoreHelper)
     {
-        stopped = false;
         idlePeriod = config->getPropInt("@lCIdlePeriod", DEFAULT_LCIDLE_PERIOD)*1000;
         minimumTimeBetweenSaves = config->getPropInt("@lCMinTime", DEFAULT_LCMIN_TIME)*1000;
         idleRate = config->getPropInt("@lCIdleRate", DEFAULT_LCIDLE_RATE);
@@ -8626,6 +8627,8 @@ bool CCovenSDSManager::fireException(IException *e)
     // so ignore unhandled exceptions for the moment! 
     IERRLOG(e, "Caught unhandled exception!");
     return true;
+
+#if 0
     { CHECKEDCRITICALBLOCK(unhandledCrit, fakeCritTimeout);
         if (processingUnhandled)
         {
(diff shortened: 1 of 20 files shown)


dali/base/dasds.cpp
```diff
@@ -4832,9 +4832,11 @@ IPropertyTree *loadStore(const char *storeFilename, IPTreeMaker *iMaker, unsigne
 // Not really coalescing, blocking transations and saving store (which will delete pending transactions).
 class CLightCoalesceThread : implements ICoalesce, public CInterface
 {
-    bool stopped, within24;
+    bool stopped = false;
+    bool within24 = false;
     Semaphore sem;
-    unsigned writeTransactionsNow, lastSaveWriteTransactions, lastWarning;
```
ghalliday, line 4837: @jakesmith writeTransactionsNow not used - there is a variable with the same name inside a function

ecl/hqlcpp/hqlinline.cpp
```diff
@@ -1820,9 +1820,6 @@ AliasKind ClassEvalContext::evaluateExpression(BuildCtx & ctx, IHqlExpression *
         }
     }
 
-    if (!evaluateLocally)
```
ghalliday, line 1823: can never be reached - see code above

esp/platform/cumulativetimer.hpp
```diff
@@ -41,11 +41,6 @@ class CumulativeTimer : public CInterface
     {
     }
 
-    CumulativeTimer(const CumulativeTimer& t)
```
ghalliday, line 44: @afishbeck this wasn't called, and is the same as the default copy constructor, except it left logLevel uninitialized.

esp/platform/espprotocol.hpp
```diff
@@ -163,8 +163,7 @@ class esp_http_decl CEspProtocol : public CInterface,
     CApplicationPortMap m_portmap;
     bool m_viewConfig;
     int m_MaxRequestEntityLength;
-    IEspContainer *m_container;
-    unsigned m_nextSeq;
```
ghalliday, line 167: @afishbeck was unused.

esp/platform/espsession.ipp
```diff
@@ -33,7 +33,7 @@ private:
     StringAttr m_commentURL;
     StringAttr m_domain;
     StringAttr m_ports;
-    int        m_maxage;
+    int        m_maxage = 0;
```
ghalliday, line 36: @afishbeck could this have caused any problems?

ghalliday (commented):
Please review

esp/platform/cumulativetimer.hpp
afishbeck: looks fine

esp/platform/espsession.ipp
afishbeck: If the uninitialized value happened to be negative then the cookie would be kept until browser closed.  A positive value would be time to live for the cookie.  Zero doesn't get set.  Odds are unless the value was a fairly small positive value we wouldn't have noticed.

dali/base/dasds.cpp
```diff
@@ -8681,6 +8684,7 @@ bool CCovenSDSManager::fireException(IException *e)
         }
     };
     unhandledThread.setown(new CHandleException(*this, e, restartOnError));
+#endif
     return true;
```
richardkchapman, line 8688: Doesn't this return true need to be inside the #if (or the one above removed)?

dali/server/daldap.cpp
```diff
@@ -148,8 +148,8 @@ class CDaliLdapConnection: implements IDaliLdapConnection, public CInterface
         Owned<ISecUser> user = ldapsecurity->createUser(username);
         user->setAuthenticateStatus(AS_AUTHENTICATED);
 
-        bool filescope = stricmp(key,"Scope")==0;
-        bool wuscope = stricmp(key,"workunit")==0;
+        bool filescope = key && stricmp(key,"Scope")==0;
```
richardkchapman, line 151: use strieq perhaps?

richardkchapman: (perhaps strieq should be made safe!)

richardkchapman:
@ghalliday One minor comment and one (probably incorrect) suggestion

dali/server/daldap.cpp
ghalliday: possibly, but I was aiming for minimal changes.  Not convinced about strieq.

dali/base/dasds.cpp
ghalliday: moved inside.

ghalliday:
@richardkchapman moved the return, but left the rest as it was

PR #13737, opened 2020-05-12
HPCC-23944 eclcc takes a long time to quit if k8s cluster stopped

richardkchapman:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

## Testing:
Manual.

### DIFF
--- a/common/workunit/workunit.cpp
+++ b/common/workunit/workunit.cpp
@@ -13904,7 +13904,7 @@ void deleteK8sJob(const char *componentName, const char *job)
     jobname.toLowerCase();
     VStringBuffer deleteJob("kubectl delete job/%s", jobname.str());
     StringBuffer output, error;
-    bool ret = runExternalCommand(output, error, deleteJob.str(), nullptr);
+    bool ret = runExternalCommand(componentName, output, error, deleteJob.str(), nullptr);
     DBGLOG("kubectl delete output: %s", output.str());
     if (error.length())
         DBGLOG("kubectl delete error: %s", error.str());
@@ -13912,23 +13912,44 @@ void deleteK8sJob(const char *componentName, const char *job)
         throw makeStringException(0, "Failed to run kubectl delete");
 }
 
-void waitK8sJob(const char *componentName, const char *job, const char *condition)
+void waitK8sJob(const char *componentName, const char *job)
 {
     VStringBuffer jobname("%s-%s", componentName, job);
     jobname.toLowerCase();
+    VStringBuffer waitJob("kubectl get jobs %s -o jsonpath={.status.active}", jobname.str());
+    VStringBuffer getScheduleStatus("kubectl get pods --selector=job-name=%s --output=jsonpath={.items[*].status.conditions[?(@.type=='PodScheduled')].status}", jobname.str());
 
-    if (isEmptyString(condition))
-        condition = "condition=complete";
-
-    // MORE - blocks indefinitely here if you request too many resources
-    VStringBuffer waitJob("kubectl wait --for=%s --timeout=10h job/%s", condition, jobname.str());  // MORE - make timeout configurable
-    StringBuffer output, error;
-    bool ret = runExternalCommand(output, error, waitJob.str(), nullptr);
-    DBGLOG("kubectl wait output: %s", output.str());
-    if (error.length())
-        DBGLOG("kubectl wait error: %s", error.str());
-    if (ret)
-        throw makeStringException(0, "Failed to run kubectl wait");
+    unsigned delay = 100;
+    unsigned start = msTick();
+    unsigned pendingTimeout = 10000;
+    for (;;)
+    {
+        StringBuffer output, error;
+        unsigned ret = runExternalCommand(componentName, output, error, waitJob.str(), nullptr);
+        if (ret || error.length())
+            throw makeStringExceptionV(0, "Failed to run %s: error %u: %s", waitJob.str(), ret, error.str());
+        if (!streq(output, "1"))  // status.active value
+        {
+            DBGLOG("kubectl jobs output: %s", output.str());
+            break;
+        }
+        ret = runExternalCommand(componentName, output.clear(), error.clear(), getScheduleStatus.str(), nullptr);
+        if (error.length())
+        {
+            DBGLOG("kubectl get pods error: %s", error.str());
+            break;
+        }
+        bool pending = streq(output, "False");
+        if (pending && msTick()-start > pendingTimeout)
+        {
+            VStringBuffer getReason("kubectl get pods --selector=job-name=%s \"--output=jsonpath={range .items[*].status.conditions[?(@.type=='PodScheduled')]}{.reason}{': '}{.message}{end}\"", jobname.str());
+            runExternalCommand(componentName, output.clear(), error.clear(), getReason.str(), nullptr);
+            throw makeStringExceptionV(0, "Failed to run %s - pod not scheduled after %d ms: %s ", jobname.str(), pendingTimeout, output.str());
+        }
+        MilliSleep(delay);
+        if (delay < 10000)
+            delay = delay * 2;
+    }
 }
 
 void launchK8sJob(const char *componentName, const char *wuid, const char *job, const std::list<std::pair<std::string, std::string>> &extraParams)
--- a/common/workunit/workunit.cpp
+++ b/common/workunit/workunit.cpp
@@ -13925,20 +13925,23 @@ void waitK8sJob(const char *componentName, const char *job)
     for (;;)
     {
         StringBuffer output, error;
-        unsigned ret = runExternalCommand(componentName, output, error, waitJob.str(), nullptr);
+        unsigned ret = runExternalCommand(nullptr, output, error, waitJob.str(), nullptr);
         if (ret || error.length())
             throw makeStringExceptionV(0, "Failed to run %s: error %u: %s", waitJob.str(), ret, error.str());
         if (!streq(output, "1"))  // status.active value
         {
+            // Job is no longer active - we can terminate
             DBGLOG("kubectl jobs output: %s", output.str());
             break;
         }
-        ret = runExternalCommand(componentName, output.clear(), error.clear(), getScheduleStatus.str(), nullptr);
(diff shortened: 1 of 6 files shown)


richardkchapman:
@jakesmith Please review

common/workunit/workunit.cpp
```diff
@@ -13540,31 +13540,53 @@ void deleteK8sJob(const char *componentName, const char *job)
...
-    StringBuffer output, error;
-    bool ret = runExternalCommand(output, error, waitJob.str(), nullptr);
-    DBGLOG("kubectl wait output: %s", output.str());
-    if (error.length())
-        DBGLOG("kubectl wait error: %s", error.str());
-    if (ret)
-        throw makeStringException(0, "Failed to run kubectl wait");
+    unsigned delay = 100;
+    unsigned start = msTick();
+    unsigned pendingTimeout = 10000;
```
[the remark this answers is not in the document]
jakesmith: I think if the image needs pulling, the pods status remains 'Pending' and can take some time i.e. could be more than 10s

common/workunit/workunit.cpp
```diff
@@ -13540,31 +13540,53 @@ void deleteK8sJob(const char *componentName, const char *job)
...
+        ret = runExternalCommand(componentName, output.clear(), error.clear(), getPhase.str(), nullptr);
+        if (error.length())
+        {
+            DBGLOG("kubectl get pods error: %s", error.str());
+            break;
+        }
+        bool pending = streq(output, "'Pending'");
+        if (pending && msTick()-start > pendingTimeout)
+            throw makeStringExceptionV(0, "Failed to run %s - pod still pending after %d ms", jobname.str(), pendingTimeout);
+        MilliSleep(delay);
```
jakesmith, line 13586: If it's status is 'Running' , should it revert to using "kubectl wait --for .. " here?

Not sure it matters, but could spot status 'Success' here, and exit before delay and looping around.

jakesmith:
@richardkchapman - added some comments and questions.

I think the 10s 'Pending' timeout may be an issue sometimes, when a new node needs to pull the image.

richardkchapman:
Yes, the 10 s limit is triggering too often at the moment. I'm working on another approach

richardkchapman:
@jakesmith Please re-review

ecl/eclccserver/eclccserver.cpp
```diff
@@ -464,7 +464,30 @@ class EclccCompileThread : implements IPooledThread, implements IErrorReporter,
 #ifdef _CONTAINERIZED
         if (!globals->getPropBool("@useChildProcesses", false) && !globals->hasProp("@workunit"))
         {
-            runK8sJob("eclccserver", wuid, wuid, globals->getPropBool("@deleteJobs", true));
+            try
+            {
+                runK8sJob("compile", wuid, wuid, globals->getPropBool("@deleteJobs", true));
+            }
+            catch (IException *E)
+            {
```
jakesmith, line 472: thinking it would better if the block of code below was handled outside of the catch? (e.g. set and and test after try block)
In case this code throws and causes unhandled exception issues.

ecl/eclccserver/eclccserver.cpp
```diff
@@ -464,7 +464,30 @@ class EclccCompileThread : implements IPooledThread, implements IErrorReporter,
...
+                    {
+                        StringBuffer msg;
+                        E->errorMessage(msg);
+                        addExceptionToWorkunit(workunit, SeverityError, "eclccserver", E->errorCode(), msg.str(), NULL, 0, 0, 0);
+                        workunit->setState(WUStateFailed);
+                    }
+                }
+                E->Release();
+                workunit->commit();
+                workunit.clear();
```
jakesmith, line 489: minor: don't really need the above 2 lines.

ecl/eclccserver/eclccserver.cpp
```diff
@@ -464,7 +464,30 @@ class EclccCompileThread : implements IPooledThread, implements IErrorReporter,
...
+                runK8sJob("compile", wuid, wuid, globals->getPropBool("@deleteJobs", true));
+            }
+            catch (IException *E)
+            {
+                Owned<IWorkUnitFactory> factory = getWorkUnitFactory();
+                workunit.setown(factory->updateWorkUnit(wuid.get()));
+                if (workunit)
+                {
+                    if (workunit->aborting() || workunit->getState()==WUStateAborted)
+                        workunit->setState(WUStateAborted);
```
jakesmith, line 478: not sure if causes a transaction or not, but no reason to call setState(WUStateAborted), if state is already WUStateAborted?

common/workunit/workunit.cpp
```diff
@@ -13540,31 +13540,57 @@ void deleteK8sJob(const char *componentName, const char *job)
...
+        {
+            DBGLOG("kubectl get pods error: %s", error.str());
+            break;
+        }
+        bool pending = streq(output, "False");
+        if (pending && msTick()-start > pendingTimeout)
+        {
+            VStringBuffer getReason("kubectl get pods --selector=job-name=%s \"--output=jsonpath={range .items[*].status.conditions[?(@.type=='PodScheduled')]}{.reason}{': '}{.message}{end}\"", jobname.str());
+            runExternalCommand(componentName, output.clear(), error.clear(), getReason.str(), nullptr);
+            throw makeStringExceptionV(0, "Failed to run %s - pod not scheduled after %d ms: %s ", jobname.str(), pendingTimeout, output.str());
```
jakesmith, line 13588: perhaps include the errorcode (if there is one), if this runExcternalCommand fails?

jakesmith:
@richardkchapman - some minor comments/questions.

common/workunit/workunit.cpp
richardkchapman: I don't think that would provide any useful information

ecl/eclccserver/eclccserver.cpp
richardkchapman: Just copying the style of code from elsewhere in this function - and it's not going to be speed critical. The code is clearer like this I think.

ecl/eclccserver/eclccserver.cpp
richardkchapman: Again, copying style from elsewhere - and I think the commit MAY be needed (I can't remember the semantics!)

richardkchapman:
Repushed following review

common/workunit/workunit.cpp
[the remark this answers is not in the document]
ghalliday: What is the conclusion about the 10s delay.  I have seen delays significantly larger than that on azure pulling the first image.

ecl/eclccserver/eclccserver.cpp
ghalliday: I agree with Jake, it doesn't make much sense to set the state to a value it already has.  I wonder if this code meant to say
   workunit->getState()==WUStateAborting
It looks like the code has been there since the original move from svn.

ecl/eclccserver/eclccserver.cpp
ghalliday: Again, code elsewhere > 9 years old.  I think it is implicit in the destuctor for the workunit.

ghalliday (commented):
A few minor comments.

ghalliday:
@richardkchapman

common/workunit/workunit.cpp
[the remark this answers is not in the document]
richardkchapman: I've recoded since Jake's initial comment so that it's not dependent on the "Pending" status but rather on whether the pod has been scheduled.

I need to check whether the "scheduled" happens before the pause for pulling an image...

I probably also need to renaming the "pending" variable

ecl/eclccserver/eclccserver.cpp
richardkchapman: I think the code is doing what it intended - the point of including the ==aborted test is to avoid taking the "else" arm.

ecl/eclccserver/eclccserver.cpp
richardkchapman: I may not need the commit, but I think I need the clear (it's a member of the threadpool object, not a local).

richardkchapman: Looking around the code, committing before releasing seems to be the common pattern.

There is a commit called for the destructor of a CLockedWorkunit - but the commit may fail and throw an exception which is not necessarily desirable inside a destructor.

I'd prefer to have this code match the paradigm used elsewhere in this file (and the rest of the codebase). Removing unwanted calls to commit() should be a separate issue (and a very low-priority one!)

richardkchapman:
@ghalliday Please re-review

common/workunit/workunit.cpp
```diff
@@ -13904,31 +13904,55 @@ void deleteK8sJob(const char *componentName, const char *job)
...
+        // Check whether pod has been scheduled yet - if resources are not available pods may block indefinitely waiting to be scheduled, and
+        // we would prefer them to fail instead.
+        bool pending = streq(output, "False");
+        if (pending && msTick()-start > pendingTimeout)
+        {
+            VStringBuffer getReason("kubectl get pods --selector=job-name=%s \"--output=jsonpath={range .items[*].status.conditions[?(@.type=='PodScheduled')]}{.reason}{': '}{.message}{end}\"", jobname.str());
+            runExternalCommand(componentName, output.clear(), error.clear(), getReason.str(), nullptr);
+            throw makeStringExceptionV(0, "Failed to run %s - pod not scheduled after %d ms: %s ", jobname.str(), pendingTimeout, output.str());
+        }
+        MilliSleep(delay);
```
ghalliday, line 13952: Should this be limited to (nowTime - pending [+delta])?  Otherwise it is likely to delay for maybe ~12.8s.

richardkchapman: I don't think it's critical to time out at EXACTLY the right time - the reason for the increasing delay is to try to be responsive when things are fast but not burn too much of a tight loop when things are not.

Limiting the delay COULD be done easily enough, I guess.

PR #13738, opened 2020-05-12
HPCC-24022 Implementation of multi-job linger

jakesmith:
Optionally allow lingering Thor instances to accept graphs from
any job.
This allows for significantly faster processing of lots of small
jobs by avoiding a cycle of master+slave pod creation per workunit.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/common/workunit/workunit.cpp
+++ b/common/workunit/workunit.cpp
@@ -13276,57 +13276,136 @@ bool isValidMemoryValue(const char *memoryUnit)
 }
 
 #ifdef _CONTAINERIZED
-bool executeGraphOnLingeringThor(IConstWorkUnit &workunit, const char *graphName)
+bool executeGraphOnLingeringThor(IConstWorkUnit &workunit, const char *graphName, const char *multiJobLingerQueueName)
 {
     // check if lingering thor instance is up.
+    // Returns true if successfully submitted graph to a lingering Thor.
 
     /* If code was dependent on reading a workunit in parallel, the whole area of
-     * workunit locking and reading will need revisiting, because at the moment the 
+     * workunit locking and reading will need revisiting, because at the moment the
      * workunit reading code does not lock at all.
-     * 
+     *
+     * Either,
+     * 1) multiJobThorLinger is set, meaning Thor instances will wait for any job's graph.
+     * OR
+     * 2) if not set, the Thor will wait for graphs from same job only, and will be
+     *    notified to quit by the agent, when the last graph is complete.
+     *
+     * This code is called from the client (agent) when it wants to execute a Thor graph.
+     * This function checks to see if there any registered lingering Thor instances available.
+     *
+     * Lingering Thor's register when they are idle either with the workunit itself in
+     * the case of [2], or with a Dali psuedo queue (a branch under /Status/ThorLinger),
+     * in the case of [1]
+     * If they have not been contacted within the 'lingerPeriod' they shutdown and remove
+     * their registered instance from the queue.
+     *
+     * The client will attempt to find a lingering Thor in either the workunit or a global
+     * named dali queue (depending on the configuration), it will then remove that entry,
+     * and attempt to post the graph to that lingering Thor. If the Thor acknowledges the
+     * request, this function exits with success. If it it fails, or there are no more
+     * lingering Thor instances, it returns with false to indicate that the graph was not
+     * submitted to a lingering Thor.
+     * The client should then queue the wuid/graph normally, which will invoke a new Thor
+     * instance.
      */
 
-    /* NB: forcing a reload here to ensure Debug values are recent.
-     * Could be improved by refreshing only the specific area of interest (i.e. Debug/ in this case).
-     * The area of persisting a non-locked connection to workunits should be resivisted..
-     */
-    workunit.forceReload();
-
-    Owned<IStringIterator> iter = &workunit.getDebugValues("thorinstance_*");
-    ForEach(*iter)
+    try
     {
-        /* NB: Thor's set their running endpoint into Debug values of workunit
-         * Check to see if workunit has any Thor's available lingering instances.
-         */
-        SCMStringBuffer thorInstance;
-        iter->str(thorInstance);
-        SCMStringBuffer thorInstanceValue;
-        if (workunit.getDebugValueBool(thorInstance.s, false))
+        if (multiJobLingerQueueName && graphName)
         {
-            {
-                Owned<IWorkUnit> w = &workunit.lock();
-                w->setDebugValue(thorInstance.s, "0", true);
-            }
-            /* NB: there's a window where Thor could shutdown here.
-             * In that case, the sendRecv will fail and it will fall through to queueing.
-             */
-            const char *instanceName = strchr(thorInstance.str(), '_') + 1;
+            // check a Dali queue for idle spare Thor's
+            DBGLOG("Checking multi job lingering queue: %s", multiJobLingerQueueName);
 
-            Owned<INode> masterNode = createINode(instanceName);
-            CMessageBuffer msg;
-            VStringBuffer jobStr("%s/%s", workunit.queryWuid(), graphName?graphName:"");
-            msg.append(jobStr);
-            if (queryWorldCommunicator().sendRecv(msg, masterNode, MPTAG_THOR, 10000))
+            VStringBuffer xpath("/Status/ThorLinger/%s", multiJobLingerQueueName);
+            while (true)
             {
-                bool ok;
-                msg.read(ok);
-                if (graphName) // if graphName==nullptr, then continue around to look at others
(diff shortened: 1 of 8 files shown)


jakesmith:
@ghalliday - please review

dockerfiles/hpcc/values.schema.json
```diff
@@ -299,6 +299,15 @@
           "description": "The number of slave pods",
           "minimum": 1
         },
+        "lingerPeriod": {
+          "type": "integer",
+          "description": "[Optional] Seconds to keep Thor instance running, waiting for more graphs to execute",
+          "minimum": 1
```
ghalliday, line 305: would it be better to allow 0 in order to disable it?

dockerfiles/hpcc/templates/thor.yaml
```diff
@@ -91,6 +91,9 @@ data:
 {{ toYaml (omit $thorScope "logging") | indent 6 }}
 {{- include "hpcc.generateLoggingConfig" (dict "root" $ "me" $thorScope) | indent 6 }}
     {{ $agentAppType }}:
+{{- if $thorScope.multiJobLinger }}
+      multiJobLinger: true
```
ghalliday, line 95: is this the normal way of doing an option like this?  or 
multiJobLinger: {{ $thorScope.multiJobLinger | default false }}
?
(I don't know what is expected more commonly.)

common/workunit/workunit.cpp
```diff
@@ -13276,58 +13276,120 @@ bool isValidMemoryValue(const char *memoryUnit)
...
+                    break;
+                IPropertyTree *root = conn->queryRoot();
+                if (0 == root->numChildren())
+                    break;
+                IPropertyTree *entry = root->queryPropTree("Thor*[1]");
+                const char *entryName = entry->queryName();
+
+                // get available instance and remove it
+                VStringBuffer entryXPath("%s/%s", xpath.str(), entryName);
+                Owned<IRemoteConnection> lingerInstanceConn = querySDS().connect(entryXPath, myProcessSession(), RTM_LOCK_WRITE | RTM_DELETE_ON_DISCONNECT | RTM_LOCK_WRITE, 5*1000);
```
ghalliday, line 13317: trivial: RTM_LOCK_WRITE repeated.

common/workunit/workunit.cpp
```diff
@@ -13276,58 +13276,120 @@ bool isValidMemoryValue(const char *memoryUnit)
...
-        SCMStringBuffer thorInstance;
-        iter->str(thorInstance);
-        SCMStringBuffer thorInstanceValue;
-        if (workunit.getDebugValueBool(thorInstance.s, false))
+        if (multiJobLingerQueueName && graphName)
         {
+            // check a Dali queue for idle spare Thor's
+            DBGLOG("Checking multi job lingering queue: %s", multiJobLingerQueueName);
+
+            VStringBuffer xpath("/Status/ThorLinger/%s", multiJobLingerQueueName);
```
ghalliday, line 13303: This code would benefit from a description of the process used to implement a lingering thor - e.g., who is listening to what, and how the next job is kicked off.

thorlcr/master/thgraphmanager.cpp
```diff
@@ -1170,41 +1205,45 @@ void thorMain(ILogMsgHandler *logHandler, const char *wuid, const char *graphNam
...
+                    if (multiJobLinger)
+                    {
+                        Owned<IRemoteConnection> conn = querySDS().connect(multiJobLingerXPath, myProcessSession(), RTM_CREATE_QUERY | RTM_LOCK_WRITE, 5*60*1000);
+
+                        StringBuffer instance;
+                        queryMyNode()->endpoint().getUrlStr(instance);
+                        VStringBuffer entryXPath("%s/Thor%" I64F "u", multiJobLingerXPath.str(), myProcessSession());
+                        multiJobLingerInstanceConn.setown(querySDS().connect(entryXPath, myProcessSession(), RTM_CREATE_QUERY | RTM_DELETE_ON_DISCONNECT | RTM_LOCK_WRITE, 5*1000));
+                        multiJobLingerInstanceConn->queryRoot()->setProp(nullptr, instance);
+                        multiJobLingerInstanceConn->changeMode(0); // unlock, because can be read and deleted by a reader
```
ghalliday, line 1219: minor: RT_NONE (or RT_LOCK_NONE) would be clearer - rather than 0

thorlcr/master/thgraphmanager.cpp
```diff
@@ -1095,6 +1095,38 @@ void closeThorServerStatus()
     }
 }
 
+
+#ifdef _CONTAINERIZED
+static int recvNextGraph(unsigned timeoutMs, const char *wuid, StringBuffer &retWuid, StringBuffer &retGraphName)
```
ghalliday, line 1100: Worth having a comment about the meaning of the return codes

thorlcr/master/thgraphmanager.cpp
```diff
@@ -1170,41 +1205,45 @@ void thorMain(ILogMsgHandler *logHandler, const char *wuid, const char *graphNam
...
+                    Owned<IRemoteConnection> multiJobLingerInstanceConn;
+                    VStringBuffer multiJobLingerXPath("/Status/ThorLinger/%s", globals->queryProp("@name"));
+                    if (multiJobLinger)
+                    {
+                        Owned<IRemoteConnection> conn = querySDS().connect(multiJobLingerXPath, myProcessSession(), RTM_CREATE_QUERY | RTM_LOCK_WRITE, 5*60*1000);
+
+                        StringBuffer instance;
+                        queryMyNode()->endpoint().getUrlStr(instance);
+                        VStringBuffer entryXPath("%s/Thor%" I64F "u", multiJobLingerXPath.str(), myProcessSession());
+                        multiJobLingerInstanceConn.setown(querySDS().connect(entryXPath, myProcessSession(), RTM_CREATE_QUERY | RTM_DELETE_ON_DISCONNECT | RTM_LOCK_WRITE, 5*1000));
```
ghalliday, line 1217: Some comments on the lifetime would help when reading/returning to this code.

ghalliday (commented):
@jakesmith the code looks good.  It could do with some comments though - especially documenting the process for finding lingering thors.

dockerfiles/hpcc/values.schema.json
jakesmith: it is disabled by default, if the property is not present.
So this is saying if used/present, assert that it is >=1

dockerfiles/hpcc/templates/thor.yaml
jakesmith: this is conditionally generating this property in another scope.
It's so that it appears in agent scope as well as (the original) thor scope, because it needs to be picked up at both ends.

It could unconditionally generate it as per your example, but I think it's ok to conditionally generate it only if present and true

common/workunit/workunit.cpp
jakesmith: will remove.

thorlcr/master/thgraphmanager.cpp
jakesmith: agree, will change.

thorlcr/master/thgraphmanager.cpp
jakesmith: will add

thorlcr/master/thgraphmanager.cpp
jakesmith: will add

common/workunit/workunit.cpp
jakesmith: There's some description at the top of this function, but I'll extend it further.

jakesmith:

ghalliday:
@jakesmith looks good, please squash.

I did wonder whether eclagent should preferentially connect to the thor that last serviced that query, w.g., by trying to connect to <Thor>+<wuid> before Thor*[1].  Possibly something to return to.
Also, is there a jira to document this feature?

jakesmith:
>I did wonder whether eclagent should preferentially connect to the thor that last serviced that query,

could do, but I think the standard linger option (with mulitJobLinger: false) caters for the case where you want to make sure there's no cross-job contamination.
If it multiJobLinger tried 1st to look for same instance, then it is likely that instance will have previously ran other jobs and quite likely graphs from other jobs would interweave.

jakesmith:
>Also, is there a jira to document this feature?

There's the values.scheme.json doc. :)
Apart from that, and the blogs, I don't think there's any JIRA on documenting it or other properties / options at the moment.
This may be an option that is worth highlighting, but at the moment I feel it is more use as a developer configuration option.

jakesmith:
@ghalliday - squashed.

PR #13741, opened 2020-05-12
HPCC-24026 Ensure k8s jobnames are unique

jakesmith:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/common/workunit/workunit.cpp
+++ b/common/workunit/workunit.cpp
@@ -13772,7 +13772,7 @@ void launchK8sJob(const char *componentName, const char *wuid, const char *job, 
     }
 
     StringBuffer output, error;
-    unsigned ret = runExternalCommand(output, error, "kubectl apply -f -", jobYaml.str());
+    unsigned ret = runExternalCommand(output, error, "kubectl replace --force -f -", jobYaml.str());
     DBGLOG("kubectl output: %s", output.str());
     if (error.length())
         DBGLOG("kubectl error: %s", error.str());


jakesmith:

Re. whitespaces (trimmed trailing), this is being caused by an option I have set in VSCode, to prevent trailing white spaces, which automatically strips any trailing whitespaces from end of lines when a file is saved.
It's convenient and avoids them being introduced and clears up the existing erroneous spaces on a per-file basis. Let me know if you okay with that in this/other commits, e.g. by reviewing in github by turning off white spaces from the diff.

richardkchapman:
I can't see the jira so there may be more commentary there, but it's not clear to me why a particular wu/graph would be executing more than once?

jakesmith:
The ECL construct "RECOVERY" can cause it.
It can re-run the same graph (and hence subgraph) again after failure.
I'll add the particular regression suite query name to the JIRA for reference that caused this issue to be hit.

jakesmith:
Well I would if JIRA wasn't down. The test query is : issue22196.ecl

roxie/ccd/ccdcontext.cpp
```diff
@@ -2317,11 +2317,13 @@ class CRoxieContextBase : implements IRoxieSlaveContext, implements ICodeContext
             }
         }
         else
-        {        
-            VStringBuffer job("%s-%s", wuid.str(), graphName);
+        {
+            CDateTime dt;
+            dt.setNow();
+            VStringBuffer job("%s-%s-%" I64F "u", wuid.str(), graphName, dt.getHash(0));
```
richardkchapman, line 2323: Feels like the logic should be in runK8sjob to me. I'm also wondering whether (since this is a very rare/unusual situation) we would be better adding a suffix only if job creation fails? Depends how much we care about job names being easy to type/decode...

jakesmith:
@richardkchapman - please re-review/merge.

PR #13750, opened 2020-05-14
HPCC-15886 Add function timing support to Thor

jakesmith:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/thorlcr/graph/thgraph.cpp
+++ b/thorlcr/graph/thgraph.cpp
@@ -28,6 +28,8 @@
 #include "rtlformat.hpp"
 #include "thorsoapcall.hpp"
 #include "thorport.hpp"
+#include "roxiehelper.hpp"
+
 
 
 PointerArray createFuncs;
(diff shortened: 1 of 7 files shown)


testing/regress/ecl-test.json
```diff
@@ -4,7 +4,7 @@
         "espIp" : "127.0.0.1",
         "espSocket": "8010",
         "dropzoneIp" : "127.0.0.1",
-        "dropzonePath" : "/var/lib/HPCCSystems/mydropzone/",
```
richardkchapman, line 7: I don't think you meant to include this

jakesmith: nope, just saw result and removed.

jakesmith:
@ghalliday - please review

thorlcr/graph/thgraph.cpp
```diff
@@ -320,7 +320,46 @@ CGraphElementBase *createGraphElement(IPropertyTree &node, CGraphBase &owner, CG
     return container;
 }
 
-CGraphElementBase::CGraphElementBase(CGraphBase &_owner, IPropertyTree &_xgmml) : owner(&_owner)
+///////////////////////////////////
+static const StatisticsMapping statsMap({});
+CActivityCodeContext::CActivityCodeContext() : stats(statsMap)
```
ghalliday, line 325: I expected this to be using the activity's stats.  Was there a reason why you created a new stats object instead?

ghalliday (commented):
@jakesmith one question/comment

thorlcr/graph/thgraph.cpp
jakesmith: @ghalliday - none of the activity stats will be recorded in this context.
This (ICodeContext) is only used by generated code when registering function timings, so all stats will be nested.

These nested stats are then merged into the activity stats during collection, inside CSlaveActivity::serializeStats

ghalliday: @jakesmith The alternative I was expecting (similar to roxie) is to pass the activity stats from the activity as a reference.  They wouldn't then need to be merged.

jakesmith: right agree that would be better, though it's a little awkward because of the existing framework.
In that the activity context is created in a generic container before the specific activity with the stats. is created.

But I've changed it so that it is passed through when the activity is created, which avoids the merge (see latest commit).

jakesmith:
@ghalliday - please review

jakesmith:
@ghalliday - appears my latest commit broke something, I'll circle back and find out why shortly.

AttilaVamos:
@jakesmith this latest result belongs to previous commit and generated by the sequential (backup) Smoketest.

jakesmith:
@ghalliday - please review

ghalliday:
@jakesmith do you think it is better or worse with the last commit?  It avoids the merge but means that master activities now contain a stats member.  It is possible some of them may actually want to record some stats, so that may actually be a useful addition.

jakesmith:
@ghalliday - I was hesitant, it's quite pervasive for the sake of a small improvement in efficiency at a code path which isn't critical.
But as you say, there could be stats in the future that we'd want to record/collate in the master (which it does not do at all at the moment).

I think on balance I'd keep it for that reason.

thorlcr/graph/thgraph.cpp
```diff
@@ -320,7 +322,47 @@ CGraphElementBase *createGraphElement(IPropertyTree &node, CGraphBase &owner, CG
...
+{
+    return containerGraph->queryLoopCounter();           // only called if value is valid
+}
+
+ISectionTimer * CActivityCodeContext::registerTimer(unsigned activityId, const char * name)
+{
+    if (!stats) // if master context and local CQ, there is no activity instance, and hence no setStats call
+        return queryNullSectionTimer();
+    CriticalBlock b(contextCrit);
+    auto it = functionTimers.find(name);
```
ghalliday, line 353: This isn't particularly time criticial, so it doesn't matter.  However do you know if this will cause a temporary std::string to be created?

ghalliday: I think the answer is yes.  Something to bear in mind for cases that are time critical.

ghalliday:
@jakesmith please squash.  I think ready to merge.

jakesmith:
@ghalliday - squashed

PR #13758, opened 2020-05-18
Azure example charts for creating persistent volume claims

ghalliday:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dockerfiles/examples/azure/azhpcc/templates/_helpers.tpl
+++ b/dockerfiles/examples/azure/azhpcc/templates/_helpers.tpl
@@ -1,0 +1,248 @@
+{{/*
+Expand the name of the chart.
+*/}}
+{{- define "hpcc.name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+
+{{/*
+Create a default fully qualified app name.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+If release name contains chart name it will be used as a full name.
+*/}}
+{{- define "hpcc.fullname" -}}
+{{- if .Values.fullnameOverride -}}
+{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" -}}
+{{- else -}}
+{{- $name := default .Chart.Name .Values.nameOverride -}}
+{{- if contains $name .Release.Name -}}
+{{- .Release.Name | trunc 63 | trimSuffix "-" -}}
+{{- else -}}
+{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Create chart name and version as used by the chart label.
+*/}}
+{{- define "hpcc.chart" -}}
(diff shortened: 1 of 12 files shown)


ghalliday:
@richardkchapman @jakesmith for discussion
- is this the correct solution for azure persistent files?
- what structure would be sensible for examples?
- What about _helpers.tpl ( I copied, but didn't cut it down - should it include the hpcc chart)
- Is it sensible to have a replacement values.yaml?  Should it be called values.yaml (in this pr it is currently azure.yaml).  I think it should probably be renamed.
- What is missing?  I should add extra documentation on how to use.  (I added a very poor README.md)

dockerfiles/examples/azure/azhpcc/templates/_helpers.tpl
```diff
@@ -0,0 +1,248 @@
+{{/*
+Expand the name of the chart.
+*/}}
```
richardkchapman, line 3: There looks to be much more in here than I would expect?

ghalliday: yes.  I'm not sure how this chart should relate to the global helm  charts.  Should it use a relative include to pick up the _helper from the main repo?

ghalliday: (see  comments above)

jakesmith: My feeling is that is should only be a setup for storage for Azure.
And similarly there'd be one to provide storage for GKE and AWS.

The HPCC helm values would only point to pre-existing PVC's, e.g. provisioned by one of the above, although they are just examples, and users may already have some they want to use.

This is the approach I have currently been using.

ghalliday: I generally agree - the heml chart only configures the storage, and currently clones _helpers.tpl, but I suspect it should include it instead.

As well as the storage there should probably be an example values.xml that ties together the standard helm charts and the pvcs defined here, and also defines some sensible resource limits - so that it is trivial to start a demonstration cluster on the appropriate engine.

dockerfiles/examples/azure/azure.yaml
```diff
@@ -0,0 +1,119 @@
...
+
+  # Set to true if using hostPath storageClass
+  # This will cause initContainers to be created to ensure that host mount points have correct uid/gid
+  hostStorage: false
+
+  # logging sets the default logging information for all components. Can be overridden locally
+  logging:
+    detail: 100
+
+  # Specify a defauleEsp to control which EclWatch service is returned from Std.File.GetEspURL, and other uses
```
richardkchapman, line 40: deafule

ghalliday: Cloned from values.yaml

richardkchapman: So fix it there too :)

ghalliday:
Superseded by https://github.com/hpcc-systems/HPCC-Platform/pull/13770

PR #13760, opened 2020-05-19
HPCC-24060 Add a DaemonSet to prepull HPCC image to k8s nodes

richardkchapman:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

## Testing:
Tested on Docker Desktop where it has no ill-effect but hard to say if it has any positive imact either.

### DIFF
--- a/helm/hpcc/templates/daemonset.yaml
+++ b/helm/hpcc/templates/daemonset.yaml
@@ -1,0 +1,23 @@
+{{- if .Values.global.prepull -}}
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: prepull
+spec:
+  selector:
+    matchLabels:
+      name: prepull 
+  template:
+    metadata:
+      labels:
+        name: prepull 
+    spec:
+      containers:
+      - name: prepull 
+{{ include "hpcc.addImageAttrs" (dict "root" $ "me" . "imagename" "platform-core") | indent 8 }}
+        command: ["sleep", "10000d"]
+        resources:
+          requests:
+            memory: 1Mi
+            cpu: 1m
+{{- end -}}
--- a/helm/hpcc/values.schema.json
+++ b/helm/hpcc/values.schema.json
@@ -123,6 +123,9 @@
     "global": {
       "type": "object",
       "properties": {
+        "prepull": {
+          "type": "boolean"
+        },
         "privileged": {
           "type": "boolean"
         },
--- a/helm/hpcc/values.yaml
+++ b/helm/hpcc/values.yaml
@@ -13,6 +13,10 @@ global:
   logging:
     detail: 100
 
+  #set to true to add a daemonset to every node in the cluster that will prepull the platform-core image,
+  #for faster pod startup
+  prepull: false
+  
   # Specify a defaultEsp to control which eclservices service is returned from Std.File.GetEspURL, and other uses
   # If not specified, the first esp component that exposes eclservices application is assumed.
   # Can also be overridden locally in individual components


richardkchapman:
@jakesmith What do you think?

jakesmith:
@richardkchapman - Interesting, I can see it does help the case for the 1st run of an installation where there are unused nodes in a node pool. I don't think it helps where auto-scaling though, because the auto-scaling would allocate new nodes due to pod resource pressure and the DaemonSet wouldn't get to pull the image any more quickly than the pod would have done.
And I guess you'd want to leave this option off, if you had a K8s cluster and node pool(s), that were being used for HPCC + other things.

Side question, how does docker handle 2 requests to pull an image? (which could happen anyway with 2 pods running on a clean node). I guess it must handle concurrency efficiently already, e.g. spot that there's an existing layer being pulled and block.

I see others are doing the same, for similar reasons.
Came across this : https://codefresh.io/kubernetes-tutorial/single-use-daemonset-pattern-pre-pulling-images-kubernetes/

Should we avoid starting the our container (and using a bit of OS resources), and run a minimal one instead and issues a docker pull <image> ?
Should we use an initContainer pattern, to avoid the DaemonSet restartPolicy potentially kicking in?

richardkchapman:
That article was the inspiration for this PR, and the comment at the bottom was why I did it this way rather than the way that the article described.

I think the ideas is that if nodes are held in reserve ready for pods, then this would help them be "more ready" - i.e it will move the bulk of the image pull time from the "start the first pod on a node" to the "start the node" time. Whether that makes a difference in practice, I don't know.

jakesmith:
right, ok, I missed that comment.
Makes sense not to use docker pull .. could still use an initContainer to avoid the DaemonSet restartPolicy ?
And the init container could finish, and hence not be running base (e.g. ubuntu20.04) image and any  resources it might use.
The running container (pause in the articles example) could be busybox or something trivial.

richardkchapman:
Probably not worth merging until we are able to test on a system where something like this would make a difference.

richardkchapman:
@mckellyln @jakesmith Are we in a position to do any more testing on this?

jakesmith:
There are a couple of further comments in the article now (https://codefresh.io/kubernetes-tutorial/single-use-daemonset-pattern-pre-pulling-images-kubernetes/). Last one mentions auto scaling.

The daemonset pod is only going to be deployed once the node pool scales, and isn't it only ever going to scale because resources can't be satisfied and therefore trigger scaling to run the pod that required it.
I'm not sure the daemonset helps in that case.

If there was a non-scaling pool of nodes in V8s that was always running, then it would help get the image to the nodes that weren't otherwise being used in advance.

@richardkchapman - I could test this on Azure with and without an auto scaling node pool.
I don't think I'll be able to tell if it helps in the auto-scaling case or not though..

helm/hpcc/templates/daemonset.yaml
```diff
@@ -0,0 +1,23 @@
...
+      name: prepull 
+  template:
+    metadata:
+      labels:
+        name: prepull 
+    spec:
+      containers:
+      - name: prepull 
+{{ include "hpcc.addImageAttrs" (dict "root" $ "me" . "imagename" "platform-core") | indent 8 }}
+        command: ["sleep", "10000d"]
```
ghalliday, line 18: you'll have retired before it times out?
Will this continue to poll to see if there are new images?

How does this approach compare with https://gist.github.com/itaysk/7bc3e56d69c4d72a549286d98fd557dd

richardkchapman:
Looking at the side @ghalliday linked to, it looks like they are taking a slightly different approach, at least in so far as they provide a helm chart that can be added to any installation to prepull any image. Perhaps that's a better approach than building prepull capability into HPCC.

I think I will close this on the basis of insufficient (proven) utility versus one more thing to maintain/document.

PR #13769, opened 2020-05-22
HPCC-24074 Avoid calling fileio->size() from compression code if overwriting

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jlzw.hpp
+++ b/system/jlib/jlzw.hpp
@@ -115,7 +115,7 @@ extern jlib_decl bool isCompressedFile(IFile *file);
 extern jlib_decl ICompressedFileIO *createCompressedFileReader(IFile *file,IExpander *expander=NULL, bool memorymapped=false, IFEflags extraFlags=IFEnone);
 extern jlib_decl ICompressedFileIO *createCompressedFileReader(IFileIO *fileio,IExpander *expander=NULL);
 extern jlib_decl ICompressedFileIO *createCompressedFileWriter(IFile *file,size32_t recordsize,bool append=false,bool setcrc=true,ICompressor *compressor=NULL, unsigned compMethod=COMPRESS_METHOD_LZ4, IFEflags extraFlags=IFEnone);
-extern jlib_decl ICompressedFileIO *createCompressedFileWriter(IFileIO *fileio,size32_t recordsize,bool setcrc=true,ICompressor *compressor=NULL, unsigned compMethod=COMPRESS_METHOD_LZ4);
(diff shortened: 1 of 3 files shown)


ghalliday:
@mckellyln please review.  I will also probably work around the issue another way.

mckellyln:
@ghalliday code looks good - but one small point, the variable "append" is perhaps better named "overwrite" or "replace" with the true/false reversed ? I only suggest that because "append" seems to suggest writes are always at the end and that code hasn't changed.

ghalliday:
@mckellyln see new (replacement commit).

mckellyln (approved):

mckellyln:
@ghalliday looks good.

system/jlib/jlzw.hpp
```diff
@@ -115,7 +115,7 @@ extern jlib_decl bool isCompressedFile(IFile *file);
 extern jlib_decl ICompressedFileIO *createCompressedFileReader(IFile *file,IExpander *expander=NULL, bool memorymapped=false, IFEflags extraFlags=IFEnone);
 extern jlib_decl ICompressedFileIO *createCompressedFileReader(IFileIO *fileio,IExpander *expander=NULL);
 extern jlib_decl ICompressedFileIO *createCompressedFileWriter(IFile *file,size32_t recordsize,bool append=false,bool setcrc=true,ICompressor *compressor=NULL, unsigned compMethod=COMPRESS_METHOD_LZ4, IFEflags extraFlags=IFEnone);
-extern jlib_decl ICompressedFileIO *createCompressedFileWriter(IFileIO *fileio,size32_t recordsize,bool setcrc=true,ICompressor *compressor=NULL, unsigned compMethod=COMPRESS_METHOD_LZ4);
+extern jlib_decl ICompressedFileIO *createCompressedFileWriter(IFileIO *fileio, bool overwrite, size32_t recordsize,bool setcrc=true,ICompressor *compressor=NULL, unsigned compMethod=COMPRESS_METHOD_LZ4);
```
richardkchapman, line 118: Now you have two versions of this overloaded function, with a bool parameter, which means the opposite in one than the other...

richardkchapman:
I don't necessarily disagree with Mark re the append vs overwrite, but I think needs to be consistent across all versions of the function.

mckellyln:
Yes, good point, I should have looked at the other protos.

ghalliday:
I'll fix the other function.

ghalliday:
@richardkchapman I reverted the change instead.  append is the correct parmaeter name for the first function, and not unreasonable for the second function.

PR #13772, opened 2020-05-24
HPCC-24071 Cryptohelper PKE to support binary passphrases

RussWhitehead:
Added interfaces to pass binary passphrase to PKI routines. Added a callback
function which will be invoked to set the passphrase and passphrase length
Also removed passphrase argument from public key methods, since it is not needed

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [x] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/security/cryptohelper/pke.cpp
+++ b/system/security/cryptohelper/pke.cpp
@@ -73,14 +73,27 @@ void CLoadedKey::finalize(RSA *_rsa, const char *_keyName)
     keyName.set(_keyName);
 }
 
+//called during PEM_read_bio_RSA_PUBKEY to set passphrase
+int passphraseCB(char *passPhraseBuf, int passPhraseBufSize, int rwflag, void *pPassPhraseMB)
+{
+    size32_t len = ((MemoryBuffer*)pPassPhraseMB)->length();
+    if (passPhraseBufSize >= (int)len)
+    {
+        memcpy(passPhraseBuf, ((MemoryBuffer*)pPassPhraseMB)->bufferBase(), len);
+        return len;
+    }
+    PROGLOG("Private Key Passphrase too long (%d bytes), max %d", len, passPhraseBufSize);
+    return 0;
+}
+
 class CLoadedPublicKeyFromFile : public CLoadedKey
 {
 public:
-    CLoadedPublicKeyFromFile(const char *keyFile, const char *passPhrase)
+    CLoadedPublicKeyFromFile(const char *keyFile)
     {
         if (!loadKeyFromFile(keyFile))
-            throwEVPExceptionV(0, "loadKeyFromFile: failed to open key: %s", keyFile);
-        RSA *rsaKey = PEM_read_bio_RSA_PUBKEY(keyBio, nullptr, nullptr, (void*)passPhrase);
+            throwEVPExceptionV(0, "CLoadedPublicKeyFromFile: failed to open key: %s", keyFile);
+        RSA *rsaKey = PEM_read_bio_RSA_PUBKEY(keyBio, nullptr, nullptr, nullptr);
         if (!rsaKey)
             throwEVPExceptionV(0, "Failed to create public key: %s", keyFile);
         finalize(rsaKey, keyFile);
@@ -90,10 +103,10 @@ public:
 class CLoadedPublicKeyFromMemory : public CLoadedKey
 {
 public:
-    CLoadedPublicKeyFromMemory(const char *key, const char *passPhrase)
+    CLoadedPublicKeyFromMemory(const char *key)
     {
         loadKeyFromMem(key);
-        RSA *rsaKey = PEM_read_bio_RSA_PUBKEY(keyBio, nullptr, nullptr, (void*)passPhrase);
+        RSA *rsaKey = PEM_read_bio_RSA_PUBKEY(keyBio, nullptr, nullptr, nullptr);
         if (!rsaKey)
             throwEVPException(0, "Failed to create public key");
         finalize(rsaKey, "<inline>");
@@ -103,11 +116,22 @@ public:
 class CLoadedPrivateKeyFromFile : public CLoadedKey
 {
 public:
+    CLoadedPrivateKeyFromFile(const char *keyFile, size32_t passPhraseLen, const void *passPhrase)
+    {
+        if (!loadKeyFromFile(keyFile))
+            throwEVPException(0, "CLoadedPrivateKeyFromFile: failed to open private key");
+        MemoryBuffer passPhraseMB;
+        passPhraseMB.setBuffer(passPhraseLen, (void *)passPhrase);
+        RSA *rsaKey = PEM_read_bio_RSAPrivateKey(keyBio, nullptr, passphraseCB, (void *)&passPhraseMB);//Binary passphrase
+        if (!rsaKey)
+            throwEVPException(0, "Failed to create private key");
+        finalize(rsaKey, keyFile);
+    }
     CLoadedPrivateKeyFromFile(const char *keyFile, const char *passPhrase)
     {
         if (!loadKeyFromFile(keyFile))
-            throwEVPException(0, "loadKeyFromFile: failed to open private key");
-        RSA *rsaKey = PEM_read_bio_RSAPrivateKey(keyBio, nullptr, nullptr, (void*)passPhrase);
+            throwEVPException(0, "CLoadedPrivateKeyFromFile: failed to open private key");
+        RSA *rsaKey = PEM_read_bio_RSAPrivateKey(keyBio, nullptr, nullptr, (void*)passPhrase);//plain text or no passphrase
         if (!rsaKey)
             throwEVPException(0, "Failed to create private key");
         finalize(rsaKey, keyFile);
@@ -117,24 +141,34 @@ public:
 class CLoadedPrivateKeyFromMemory : public CLoadedKey
 {
 public:
+    CLoadedPrivateKeyFromMemory(const char *key, size32_t passPhraseLen, const void *passPhrase)
+    {
+        MemoryBuffer passPhraseMB;
+        passPhraseMB.setBuffer(passPhraseLen, (void *)passPhrase);
+        loadKeyFromMem(key);
+        RSA *rsaKey = PEM_read_bio_RSAPrivateKey(keyBio, nullptr, passphraseCB, (void *)&passPhraseMB);//Binary passphrase
+        if (!rsaKey)
+            throwEVPException(0, "Failed to create private key");
+        finalize(rsaKey, "<inline>");
+    }
     CLoadedPrivateKeyFromMemory(const char *key, const char *passPhrase)
     {
         loadKeyFromMem(key);
-        RSA *rsaKey = PEM_read_bio_RSAPrivateKey(keyBio, nullptr, nullptr, (void*)passPhrase);
+        RSA *rsaKey = PEM_read_bio_RSAPrivateKey(keyBio, nullptr, nullptr, (void*)passPhrase);//plain text or no passphrase
         if (!rsaKey)
             throwEVPException(0, "Failed to create private key");
         finalize(rsaKey, "<inline>");
     }
 };
 
-CLoadedKey *loadPublicKeyFromFile(const char *keyFile, const char *passPhrase)
+CLoadedKey *loadPublicKeyFromFile(const char *keyFile)
 {
-    return new CLoadedPublicKeyFromFile(keyFile, passPhrase);
+    return new CLoadedPublicKeyFromFile(keyFile);
 }
 
-CLoadedKey *loadPublicKeyFromMemory(const char *key, const char *passPhrase)
+CLoadedKey *loadPublicKeyFromMemory(const char *key)
 {
-    return new CLoadedPublicKeyFromMemory(key, passPhrase);
+    return new CLoadedPublicKeyFromMemory(key);
 }
 
 CLoadedKey *loadPrivateKeyFromFile(const char *keyFile, const char *passPhrase)
@@ -145,6 +179,16 @@ CLoadedKey *loadPrivateKeyFromFile(const char *keyFile, const char *passPhrase)
 CLoadedKey *loadPrivateKeyFromMemory(const char *key, const char *passPhrase)
 {
     return new CLoadedPrivateKeyFromMemory(key, passPhrase);
+}
+
+CLoadedKey *loadPrivateKeyFromFile(const char *keyFile, size32_t passPhraseLen, const void *passPhrase)
+{
+    return new CLoadedPrivateKeyFromFile(keyFile, passPhraseLen, passPhrase);
+}
+
+CLoadedKey *loadPrivateKeyFromMemory(const char *key, size32_t passPhraseLen, const void *passPhrase)
+{
+    return new CLoadedPrivateKeyFromMemory(key, passPhraseLen, passPhrase);
 }
 
 size32_t _publicKeyEncrypt(OwnedEVPMemory &dstMem, size32_t dstMaxSz, size32_t inSz, const void *inBytes, const CLoadedKey &publicKey)
--- a/system/security/cryptohelper/pke.hpp
+++ b/system/security/cryptohelper/pke.hpp
@@ -50,10 +50,14 @@ public:
     const char *queryKeyName() const  { return keyName; }
 };
 
-jlib_decl CLoadedKey *loadPublicKeyFromFile(const char *keyFile, const char *passPhrase);
-jlib_decl CLoadedKey *loadPublicKeyFromMemory(const char *key, const char *passPhrase);
+jlib_decl CLoadedKey *loadPublicKeyFromFile(const char *keyFile);
+jlib_decl CLoadedKey *loadPublicKeyFromMemory(const char *key);
 jlib_decl CLoadedKey *loadPrivateKeyFromFile(const char *keyFile, const char *passPhrase);
 jlib_decl CLoadedKey *loadPrivateKeyFromMemory(const char *key, const char *passPhrase);
+
+//binary passphase, possibly containing embedded NULL characters
+jlib_decl CLoadedKey *loadPrivateKeyFromFile(const char *keyFile, size32_t passPhraseLen, const void *passPhrase);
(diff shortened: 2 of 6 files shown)


RussWhitehead:
@jakesmith Please review

system/security/cryptohelper/pke.cpp
```diff
@@ -102,39 +115,69 @@ class CLoadedPublicKeyFromMemory : public CLoadedKey
...
+        else
+            rsaKey= PEM_read_bio_RSAPrivateKey(keyBio, nullptr, nullptr, (void*)passPhrase);//plain text or no passphrase
         if (!rsaKey)
             throwEVPException(0, "Failed to create private key");
         finalize(rsaKey, "<inline>");
     }
+public:
+    CLoadedPrivateKeyFromMemory(const char *key, size32_t passPhraseLen, const void *passPhrase)
+    {
+        passPhraseMB.append(passPhraseLen, passPhrase);
```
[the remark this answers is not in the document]
jakesmith: I suspect it would be better to combine CLoadedPrivateKeyFromFile and CLoadedPrivateKeyFromMemory, as they duplicate code. The ctor share same prototypes but could differentiate with a bool.
Or could use a CLoadedPrivateKeyBase class.

system/security/cryptohelper/pke.cpp
```diff
@@ -73,14 +73,27 @@ void CLoadedKey::finalize(RSA *_rsa, const char *_keyName)
     keyName.set(_keyName);
 }
 
+//called during PEM_read_bio_RSA_PUBKEY to set passphrase
+int passphraseCB(char *passPhraseBuf, int passPhraseBufSize, int rwflag, void *pPassPhraseMB)
+{
+    size32_t len = ((MemoryBuffer*)pPassPhraseMB)->length();
+    if (passPhraseBufSize >= (int)len)
```
jakesmith, line 80: why would it not always be an exact match? If mismatched would indicate a problem wouldn't?

testing/unittests/cryptotests.cpp
```diff
@@ -684,7 +684,7 @@ class CryptoTestTiming : public CppUnit::TestFixture
         }
         printf("RSA %u cycles - reloading keys each iteration - %u ms\n", numCycles, timer.elapsedMs());
 
-        Owned<CLoadedKey> publicKey = loadPublicKeyFromMemory(pubKey, nullptr);
+        Owned<CLoadedKey> publicKey = loadPublicKeyFromMemory(pubKey);
         Owned<CLoadedKey> privateKey = loadPrivateKeyFromMemory(privKey, nullptr);
```
jakesmith, line 688: should add tests of the new methods that use CB.

jakesmith:
@RussWhitehead - added some comments/questions.

I'm not totally clear why this change is necessary?
I thought the pass phrase was by definition plain text?

RussWhitehead:
@jakesmith I thought the passphrase was plain text as well, but if that were the case then the callback would not have to return the actual length. This PR results from a customer project (Wells Fargo) that actually uses binary passphrases, and our code is choking at the first embedded NULL character

jakesmith:
I'm no expert.
@afishbeck @dcamper @mayx - could you comment?

system/security/cryptohelper/pke.cpp
```diff
@@ -73,14 +73,27 @@ void CLoadedKey::finalize(RSA *_rsa, const char *_keyName)
     keyName.set(_keyName);
 }
 
+//called during PEM_read_bio_RSA_PUBKEY to set passphrase
+int passphraseCB(char *passPhraseBuf, int passPhraseBufSize, int rwflag, void *pPassPhraseMB)
```
mayx, line 77: It seems openssl passes in "1024" for the passPhraseBufSize parameter. So a check for the limit should be done in constructor CLoadedPrivateKeyFromMemory(const char *key, size32_t passPhraseLen, const void *passPhrase), so that user would know if they pass in a passphrase longer than that. Also here an exception/error should be thrown or logged just in case...

system/security/cryptohelper/pke.cpp
```diff
@@ -73,14 +73,27 @@ void CLoadedKey::finalize(RSA *_rsa, const char *_keyName)
...
 class CLoadedPublicKeyFromFile : public CLoadedKey
 {
 public:
-    CLoadedPublicKeyFromFile(const char *keyFile, const char *passPhrase)
+    CLoadedPublicKeyFromFile(const char *keyFile)
     {
         if (!loadKeyFromFile(keyFile))
             throwEVPExceptionV(0, "loadKeyFromFile: failed to open key: %s", keyFile);
-        RSA *rsaKey = PEM_read_bio_RSA_PUBKEY(keyBio, nullptr, nullptr, (void*)passPhrase);
+        RSA *rsaKey= PEM_read_bio_RSA_PUBKEY(keyBio, nullptr, nullptr, nullptr);
```
mayx, line 96: Minor, space before =

system/security/cryptohelper/pke.cpp
```diff
@@ -90,10 +103,10 @@ class CLoadedPublicKeyFromFile : public CLoadedKey
 class CLoadedPublicKeyFromMemory : public CLoadedKey
 {
 public:
-    CLoadedPublicKeyFromMemory(const char *key, const char *passPhrase)
+    CLoadedPublicKeyFromMemory(const char *key)
     {
         loadKeyFromMem(key);
-        RSA *rsaKey = PEM_read_bio_RSA_PUBKEY(keyBio, nullptr, nullptr, (void*)passPhrase);
+        RSA *rsaKey= PEM_read_bio_RSA_PUBKEY(keyBio, nullptr, nullptr, nullptr);
```
mayx, line 109: space before =

system/security/cryptohelper/pke.cpp
afishbeck: From what I can tell passPhraseBufSize doesn't reflect the size of our mb, it's an internal buff which looks to be set to PEM_BUFSIZE  which is 1024 at least at the location I looked.  
We should probably make sure provided passphrases do not exceed this value early on, and give an error, rather than failing here?

afishbeck: FYI Openssl source code is pretty darn obfuscated at least at the top level.

system/security/cryptohelper/pke.hpp
```diff
@@ -50,11 +50,15 @@ class jlib_decl CLoadedKey : public CSimpleInterfaceOf<IInterface>
     const char *queryKeyName() const  { return keyName; }
 };
 
-jlib_decl CLoadedKey *loadPublicKeyFromFile(const char *keyFile, const char *passPhrase);
-jlib_decl CLoadedKey *loadPublicKeyFromMemory(const char *key, const char *passPhrase);
+jlib_decl CLoadedKey *loadPublicKeyFromFile(const char *keyFile);
+jlib_decl CLoadedKey *loadPublicKeyFromMemory(const char *key);
```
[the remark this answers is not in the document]
afishbeck: I was also wondering why we were removing this capability if it were already there.

RussWhitehead:
@afishbeck @dcamper @mayx I believe what Jake is asking is whether or not a passphrase can only be plain text. If so, this PR to support binary passphrasess is unnecessary.

afishbeck:
@russwhitehead the documentation isn’t very clear.  The OpenSSL source isn’t exactly easy easy to trace either.   But it looks like the pass phrase ends up treated as data even though the talk a bit about character sets in specific scenarios.  If you’ve tested, including with embedded nulls and it works and the customer already uses binary I would say supporting it is fine.

The callback mechanism is a little strange but I think it’s so they can support pre-registering a global callback for the entire process.

system/security/cryptohelper/pke.cpp
dcamper: I don't believe there is a hard limit of 1024 bytes here.  OpenSSL does use PEM_BUFFSIZE in most (all?) places where the callback is invoked, and PEM_BUFFSIZE does currently = 1024, but I don't think we should encode that limit on our end.  OpenSSL does have a habit of changing their minds....  That said, it may be a good idea to perform a sanity check on passPhraseBufSize and make sure it's not some unrealistic number.

system/security/cryptohelper/pke.cpp
```diff
@@ -102,39 +115,69 @@ class CLoadedPublicKeyFromMemory : public CLoadedKey
...
+private:
+    MemoryBuffer passPhraseMB;
+    void load(const char *keyFile, const char *passPhrase)
     {
         if (!loadKeyFromFile(keyFile))
             throwEVPException(0, "loadKeyFromFile: failed to open private key");
-        RSA *rsaKey = PEM_read_bio_RSAPrivateKey(keyBio, nullptr, nullptr, (void*)passPhrase);
+        RSA *rsaKey;
+        if (passPhraseMB.length())
+            rsaKey= PEM_read_bio_RSAPrivateKey(keyBio, nullptr, passphraseCB, (void *)&passPhraseMB);//Binary passphrase
```
dcamper, line 126: (void *)&passPhraseMB should be passPhraseMB.bufferBase() instead.

system/security/cryptohelper/pke.cpp
```diff
@@ -102,39 +115,69 @@ class CLoadedPublicKeyFromMemory : public CLoadedKey
 
 class CLoadedPrivateKeyFromFile : public CLoadedKey
 {
-public:
-    CLoadedPrivateKeyFromFile(const char *keyFile, const char *passPhrase)
+private:
+    MemoryBuffer passPhraseMB;
```
dcamper, line 119: I don't think this needs to be a member of the object.

Perhaps a better implementation would be to change load() to accept a MemoryBuffer& parameter, then change the calling constructors to build a MemoryBuffer as needed.  That would simplify load() as well.  You may also want to consider explicit zeroing the MemoryBuffer contents before destruction.

Same change could be applied to CLoadedPrivateKeyFromMemory.

system/security/cryptohelper/pke.cpp
```diff
@@ -102,39 +115,69 @@ class CLoadedPublicKeyFromMemory : public CLoadedKey
...
-    CLoadedPrivateKeyFromMemory(const char *key, const char *passPhrase)
+private:
+    MemoryBuffer passPhraseMB;
+    void load(const char *key, const char *passPhrase)
     {
         loadKeyFromMem(key);
-        RSA *rsaKey = PEM_read_bio_RSAPrivateKey(keyBio, nullptr, nullptr, (void*)passPhrase);
+        RSA *rsaKey;
+        if (passPhraseMB.length())
+            rsaKey= PEM_read_bio_RSAPrivateKey(keyBio, nullptr, passphraseCB, (void *)&passPhraseMB);//Binary passphrase
```
dcamper, line 154: (void *)&passPhraseMB should be passPhraseMB.bufferBase() instead.

mayx:
Now I'm kind of leaning toward thinking that nulls in passphrase is probably not a real concern. I did a quick test, in which the passphrase is "password", but in the call back, I added another char 'a' after the terminating 0, and returned strlen+2 as the length. Openssl still likes the password, which kind of suggests that it ignores what's after the null terminator. Not sure if there're other cases that I haven't thought of.

system/security/cryptohelper/pke.cpp
RussWhitehead: I need access to the MemoryBuffer object to query the length, so I need to pass the reference

mayx:
Actually disregard my last comment. I just found that I had an error in my test, and it actually didn't work after returning strlen+2. So nulls in binary is being used so dealing with binary is necessary @RussWhitehead @dcamper @afishbeck @jakesmith

system/security/cryptohelper/pke.cpp
dcamper: Got it.  I should have read the code better!

system/security/cryptohelper/pke.hpp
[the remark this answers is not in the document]
RussWhitehead: A passphrase is a word or phrase that protects private key files.  A public key file is not encrypted and is by definition public.

system/security/cryptohelper/pke.cpp
RussWhitehead: Yes @afishbeck, I cloned the OpenSSL repo a while back, and the only thing more unreadable than it is the OpenSSL API reference manual. This is the right place to check that the passphrase isn't too long, since we don't know what size OpenSSL passphrase supports until this point in the code

afishbeck:
Haha.  "I never thought anything would be harder to understand than their
source code... until I read their documentation."

system/security/cryptohelper/pke.cpp
RussWhitehead: @mayx @dcamper As I stated in the commentary for the callback, we should NOT impose a hardcoded limit, and should instead rely on the mazsize provided by OpenSSL as an input argument to the callback. FWIW, it is unlikely that a passphrase would be that long

mayx: It's true that in almost all circumstances, the user won't pass in anything longer than PEM_BUFFSIZE. But if they do, either purposely or inadvertently, should the user be told that the password is over openssl's limit and so didn't get set, or just silently not setting it?

mayx: To be clear, I'm not talking about limiting passPhraseBufsize that is passed in by openssl, I'm talking about the size of the password passed in by the user pPassPhraseMB, if its size is over passPhraseBufsize, should we silently discard the passphrase? And, would it be better to detect earlier on that the password is too long?

RussWhitehead: I agree we should add some message to alert the caller that the passphrase is too long once we detect that, but I won't know that until the callback.  I can throw from in there with a meaningful message

mayx: We already know openssl limits the buffer to PEM_BUFFSIZE, so we can use it to limit the passphrase. But we probably don't want to expose the openssl specific constant, which can be solved by adding a method getPasswordSizeLimit() maybe?

RussWhitehead: I think it a bad practice to impose a hard limit in our code based one examination of a compile time constant found in a 3rd parties source code.  The reason they return the buffersize in the call back is so we can do the checking there, as is mentioned in the documentation (return 0 if buffer too small). Again, it is unlikely this limit will be hit

mayx: Throwing or logging a meaningful message in the callback function is probably good enough for all practical purposes.

system/security/cryptohelper/pke.hpp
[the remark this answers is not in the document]
RussWhitehead: @jakesmith I tested with and without specifying a passphrase for loading the public key file, and as I suspected both worked fine.

RussWhitehead:
@jakesmith Please rereview. Let me know if you want it squashed

RussWhitehead:
@jakesmith Please review

testing/unittests/cryptotests.cpp
```diff
@@ -347,6 +385,65 @@ class CryptoUnitTest : public CppUnit::TestFixture
...
+
+            ASSERT(toEncryptMb.length() == decryptedMb.length());
+            ASSERT(0 == memcmp(toEncryptMb.bytes(), decryptedMb.bytes(), toEncryptMb.length()));
+
+            /////////////////////////////////////
+            //PKE tests using a passphrase buffer
+            /////////////////////////////////////
+
+            Owned<CLoadedKey> publicKeyPassphrase = loadPublicKeyFromMemory(pubKeyPassphrase);
+            Owned<CLoadedKey> privateKeyPassphrase = loadPrivateKeyFromMemory(privKeyPassphrase, 18, "ThisIsMyPassphrase");
```
jakesmith, line 417: to prove the point, that this was to support binary pass phrases, can you change this so that it's a binary pass phrase with nulls in middle of it?

system/security/cryptohelper/pke.cpp
jakesmith: Shouldn't this be '>' not '>=' ?

system/security/cryptohelper/pke.cpp
```diff
@@ -103,11 +116,22 @@ class CLoadedPublicKeyFromMemory : public CLoadedKey
 class CLoadedPrivateKeyFromFile : public CLoadedKey
 {
 public:
+    CLoadedPrivateKeyFromFile(const char *keyFile, size32_t passPhraseLen, const void *passPhrase)
+    {
+        if (!loadKeyFromFile(keyFile))
+            throwEVPException(0, "CLoadedPrivateKeyFromFile: failed to open private key");
+        MemoryBuffer passPhraseMB(passPhraseLen, passPhrase);
+        RSA *rsaKey = PEM_read_bio_RSAPrivateKey(keyBio, nullptr, passphraseCB, (void *)&passPhraseMB);//Binary passphrase
+        memset(passPhraseMB.bufferBase(), 0, passPhraseLen);//ensure local passphrase cleared from memory
```
jakesmith, line 125: not convinced this is worth it, since the passphrase will still be in memory in 'passPhrase'.

In fact, there's no need to copy it..
You need to pass in a single ptr, that captures memory + len, hence MemoryBuffer,
but you could do:
```
MemoryBuffer passPhraseMB;
passPhraseMB.setBuffer(passPhraseLen, passPhrase);
```
that will avoid the memcpy.

jakesmith:
@RussWhitehead - a couple of comments/questions.

system/security/cryptohelper/pke.cpp
jakesmith: @RussWhitehead - what is the verdict on this?

RussWhitehead: If it is equal then it is big enough for the passphrase

jakesmith:
@RussWhitehead - looks good.
@richardkchapman - ready to be merged.

RussWhitehead:
@richardkchapman Please merge

richardkchapman:
@ghalliday Can you sanity check too

ghalliday:
@richardkchapman looks ready.  It doesn't actually test nulls being in the passphrase, but no reason to think it will not work.

PR #13789, opened 2020-05-27
HPCC-23347 ESP won't generate sample XML for all methods in a service

lucasmauro:
## Type of change:
- [X] This change is a bug fix (non-breaking change which fixes an issue).

## Checklist:
  - [X] My code does not create any new warnings from compiler, build system, or lint.
- [X] The commit message is properly formatted and free of typos.
  - [X] The commit message title makes sense in a changelog, by itself.
  - [X] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization
  - [X] This change fixes the problem, not just the symptom
  - [X] The target branch of this pull request is appropriate for such a change.
- [X] There are no similar instances of the same problem that should be addressed

## Smoketest:
- [X] Send notifications about my Pull Request position in Smoketest queue.

### DIFF
--- a/esp/bindings/http/platform/httpbinding.cpp
+++ b/esp/bindings/http/platform/httpbinding.cpp
@@ -1935,30 +1935,41 @@ void EspHttpBinding::generateSampleXmlFromSchema(bool isRequest, IEspContext &co
     StringBuffer element, schemaXmlbuff(schemaxml);
     getXMLMessageTag(context, isRequest, methodQName.str(), element);
 
-    Owned<IXmlSchema> schema = createXmlSchema(schemaXmlbuff);
-    if (schema.get())
+    StringBuffer content("<?xml version=\"1.0\" encoding=\"UTF-8\"?>");
+    if (context.queryRequestParameters()->hasProp("display"))
+        content.append("<?xml-stylesheet type=\"text/xsl\" href=\"/esp/xslt/xmlformatter.xsl\"?>");
+
+    if (!methodQName || !*methodQName)
     {
-        IXmlType* type = schema->queryElementType(element);
-        if (type)
+        StringBuffer label(isRequest ? "Requests" : "Responses");
+        content.appendf("<Examples><%s>", label.str());
+        MethodInfoArray methods;
+        getQualifiedNames(context, methods);
+        ForEachItemIn(i, methods)
         {
-            StringBuffer content;
-            StringStack parent;
-            StringBuffer nsdecl("xmlns=\"");
-
-            content.append("<?xml version=\"1.0\" encoding=\"UTF-8\"?>");
-            if (context.queryRequestParameters()->hasProp("display"))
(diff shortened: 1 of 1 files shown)


richardkchapman:
@asselitx Please review

lucasmauro:
@afishbeck if you could also review, please.

esp/bindings/http/platform/httpbinding.cpp
```diff
@@ -1935,30 +1935,41 @@ void EspHttpBinding::generateSampleXmlFromSchema(bool isRequest, IEspContext &co
...
-    if (schema.get())
+    StringBuffer content("<?xml version=\"1.0\" encoding=\"UTF-8\"?>");
+    if (context.queryRequestParameters()->hasProp("display"))
+        content.append("<?xml-stylesheet type=\"text/xsl\" href=\"/esp/xslt/xmlformatter.xsl\"?>");
+
+    if (!methodQName || !*methodQName)
     {
-        IXmlType* type = schema->queryElementType(element);
-        if (type)
+        StringBuffer label(isRequest ? "Request" : "Response");
```
afishbeck, line 1944: In the native services these tags are plural "Requests", "Responses".

esp/bindings/http/platform/httpbinding.cpp
```diff
@@ -1935,30 +1935,41 @@ void EspHttpBinding::generateSampleXmlFromSchema(bool isRequest, IEspContext &co
...
+        content.appendf("<Examples><%s>", isRequest ? "Request" : "Response");
```
afishbeck, line 1945: Might as well use label.str() here.

afishbeck (changes requested):
@lucasmauro  minor changes, otherwise looks great.

esp/bindings/http/platform/httpbinding.cpp
lucasmauro: The JSON generation returns "Request" and "Response" (line 1912). Should we change them as well?

esp/bindings/http/platform/httpbinding.cpp
lucasmauro: Thanks!

esp/bindings/http/platform/httpbinding.cpp
afishbeck: Ah, I see that.  In some ways it's inconsistent, but yes, in native services it is plural in xml, and singular in JSON.  So I would make it plural here.

Btw, the reason is that in xml it is a root element that contains muliple children of different names.  So <requests><a/><b/><c/>  vs. In JSON it is an array. request[1], request[2].

lucasmauro: Understood, @afishbeck. Thanks for the explanation. I have just pushed your suggested changes.

afishbeck (approved):
@lucasmauro looks good.

PR #13797, opened 2020-05-29
HPCC-24117 Serial funnel deadlock fix

jakesmith:
(The JIRA issue for example)
This issue occurred with a group funnel which always pulls its
inputs serially. It had a global SORT on one arm and a global
GROUP on the other.
The non-parallel version of funnel also started its inputs
sequentially.

Because some activities stop their input early (in start() if for 
example it has exhausted it's local input, and because global SORT
depends on synchronizing stops across all slaves, the 1st input
can block on some slaves.
On others it can progress to the next funnel input, which can 
block if not all slaves have started (e.g. in global GROUP).

One solution may be to ensure no activity ever stops its input
early in start(), though quite a few do, and I'm not sure that
would cover all cases of similar deadlock.

Another fix, employed here, is to ensure all funnel input start's
are started in parallel. This prevents activities like global
GROUP from deadlocking, and thus ensures progress so that the 
global STOP will be stopped on all slaves.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/thorlcr/activities/funnel/thfunnelslave.cpp
+++ b/thorlcr/activities/funnel/thfunnelslave.cpp
@@ -15,6 +15,9 @@
     limitations under the License.
 ############################################################################## */
 
+#include <future>
+#include <vector>
+
 #include "jlib.hpp"
 #include "jset.hpp"
 #include "jqueue.tpp"
@@ -309,10 +312,10 @@ public:
                 unsigned i;
                 for (i=0; i<inputs.ordinality(); i++) eog[i] = false;
             }
-            current = NULL;
             currentMarker = 0;
             readThisInput = 0;
-            ForEachItemIn(i, inputs)
+
+            auto startInputNFunc = [&](unsigned i)
             {
                 try { startInput(i); }
                 catch (CATCHALL)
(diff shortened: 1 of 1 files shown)


jakesmith:
@ghalliday - please review

thorlcr/activities/funnel/thfunnelslave.cpp
```diff
@@ -309,20 +312,46 @@ class FunnelSlaveActivity : public CSlaveActivity
...
                     throw;
                 }
-                if (!current)
-                    current = queryInputStream(i);
+            };
+
+            std::vector<std::future<void>> results;
+            ForEachItemIn(i, inputs)
+            {
+                results.push_back(std::async(std::launch::async, startInputNFunc, i));
```
ghalliday, line 331: I assume you used this rather than CAsyncFor because it was simpler to code?

@richardkchapman  The code looks fine, but worth discussing what our policy is before situations like this?  Should we be using std::async more?

jakesmith: >I assume you used this rather than CAsyncFor because it was simpler to code?

A CAsyncFor may be more concise. 
But the use of a lambda function is a clearer pattern, than deriving a class to do this.

@ghalliday @richardkchapman , change it to use a derived CAsyncFor instance?

richardkchapman:
I'm fine with using std::async and lambdas where it makes the code clearer.

jakesmith:
@ghalliday - are you happy with way it is, I don't feel strongly about it, can rewrite using CAsyncFor if you think clearer.

ghalliday:
@jakesmith I think it is fine to use async (and probably clearer), but it will have different characteristics:
- an extra thread will be started since CAsyyncFor will execute the last one on the main thread
- it uses more synchronisation objects since there is one for each future.
In this context I doubt either matter too much - although you may want to change to call the last start directly.

jakesmith:
@ghalliday - have modified it to avoid a thread on 1 input.
I've std::launch::deferred to do this, so that rest of code doesn't have to deal with the exception handling differently.

ghalliday:
@jakesmith a neat solution.  Please squash and I'll merge.

jakesmith:
@ghalliday - squashed.

ghalliday:
@jakesmith What version do you think this should target?  I would have thought 7.8, but possibly 7.6

jakesmith:
@ghalliday - I targeted 7.10 on the basis that it is an ancient bug, that has only just surfaced, and it was a slightly non-trivial change. Also in the single case reported, there was a reasonable workaround.

ghalliday:
ok

PR #13801, opened 2020-06-01
HPCC-24039 Default values for maxActive are not realistic

richardkchapman:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [x] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/ecl/eclccserver/eclccserver.cpp
+++ b/ecl/eclccserver/eclccserver.cpp
@@ -890,7 +890,7 @@ int main(int argc, const char *argv[])
                 throw MakeStringException(0, "No queues found to listen on");
 #ifdef _CONTAINERIZED
             bool useChildProcesses = globals->getPropInt("@useChildProcesses", false);
-            unsigned maxThreads = globals->getPropInt("@maxActive", useChildProcesses ? 100 : 4);
+            unsigned maxThreads = globals->getPropInt("@maxActive", 4);
 #else
             // The option has been renamed to avoid confusion with the similarly-named eclcc option, but
             // still accept the old name if the new one is not present.
--- a/helm/hpcc/values.yaml
+++ b/helm/hpcc/values.yaml
@@ -63,7 +63,7 @@ eclagent:
   ## replicas indicates how many eclagent pods should be started
   replicas: 1
   ## maxActive controls how many workunits may be active at once (per replica)
-  maxActive: 100
+  maxActive: 4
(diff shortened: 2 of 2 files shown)


richardkchapman:
@ghalliday

ghalliday:
Changes look ok.

The default for maxActive in eclccserver.cpp looks a bit strange:

            unsigned maxThreads = globals->getPropInt("@maxActive", useChildProcesses ? 100 : 4);

I suspect 4 would be more sensible for both situations.

dockerfiles/hpcc/values.yaml
```diff
@@ -63,7 +63,7 @@ eclagent:
   ## replicas indicates how many eclagent pods should be started
   replicas: 1
   ## maxActive controls how many workunits may be active at once (per replica)
-  maxActive: 100
```
ghalliday, line 66: Also change default in agentexec.cpp?

ghalliday (commented):
@richardkchapman a couple of comments

dockerfiles/hpcc/values.yaml
richardkchapman: I was in two minds about this.

There's a difference between what we want the default to be if you don't specify in your helm chart, and what you want the default to be in values.yaml (representing a minimal system expected to be run on dodker desktop or similar).

If you are running on a full cloud and want to scale up according to load, 100 may be fine.

There's also an argument that we should be setting a similar limit for the child process case (in agentexec) as the k8s job case. Does setting childProcess imply you expect less parallelism, or that you expect smaller jobs?

richardkchapman:
@ghalliday This seems to be stalled

ghalliday:
I can merge this as is, because it is an improvement, but I think the defaults in the code also need fixing.  E.g. 100 threads in eclccserver when using child processes will bring any machine to its knees.  
100 in eclagent would possibly be ok if they didn't really do anything, but still seems too high.

richardkchapman:
In non-containerized system, agentexec always operates in child process mode and does not set any limit. I think we therefore picked a very high limit to avoid breaking usage that relied on similar patterns.

richardkchapman:
I don't mind setting eclccserver to 4 - but I am still unsure about correct default for agentexec

PR #13802, opened 2020-06-01
HPCC-24129 Don't prematurely flush blob nodes in noseek index builds

richardkchapman:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [x] Cloud-compatibility
  - [x] Premature optimization

## Smoketest:


system/jhtree/keybuild.cpp
```diff
@@ -310,18 +310,13 @@ class CKeyBuilder : public CInterfaceOf<IKeyBuilder>
...
-            {
-                CBlobWriteNode &pending = pendingBlobs.item(0);
+                CWriteNodeBase &pending = pendingNodes.item(0);
                 if (!prevLeafNode || pending.getFpos() > prevLeafNode->getFpos())
                     break;
                 writeNode(&pending, pending.getFpos());
-                pendingBlobs.remove(0);
+                pendingNodes.remove(0);
             }
         }
```
ghalliday, line 321: @richardkchapman the only problem I can foresee is:
first row has a blob
no other rows have blob

In which case all the leaf nodes will be buffered in memory until the end of the index.  I suspect adding some code here to add activeBlobNode to the pendingBlob list if (activeBlobNode->getFpos() - prevLeafNode->getFPos()) is > some threshold e.g. 1Mb would be the cleanest fix.

richardkchapman: I suspect there are a few other pathological patterns people can come up with that will break our indexes if they try...

richardkchapman: Not entirely sure how you write ECL in order to "not have a blob". Do we suppress blobs creation when a string marked with ,BLOB contains a zero-length string?

ghalliday: Good question, I don't know.

ghalliday: keybuild.cpp(512) - if size is 0 then no blob is created.

ghalliday: Which makes sense if you have images stored in data fields, no image means no blob overhead.

richardkchapman: I'm inclined to open a Jira to address this pathogenic case separately (and not especially urgently). It will mean that we can't assume the blob chain is continuous...

ghalliday: HPCC-24198

PR #13838, opened 2020-06-08
HPCC-24073 Ensure that index building output is buffered

ghalliday:
## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/ecl/hthor/hthor.cpp
+++ b/ecl/hthor/hthor.cpp
@@ -1157,7 +1157,10 @@ void CHThorIndexWriteActivity::execute()
         if (hasTrailingFileposition(helper.queryDiskRecordSize()->queryTypeInfo()))
             keyMaxSize -= sizeof(offset_t);
 
-        Owned<IFileIOStream> out = createIOStream(io, needsSeek);
+        Owned<IFileIOStream> out = createIOStream(io);
+        if (!needsSeek)
+            out.setown(createNoSeekIOStream(out));
+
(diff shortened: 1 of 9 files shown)


ecl/hthor/hthor.cpp
```diff
@@ -1157,7 +1157,10 @@ void CHThorIndexWriteActivity::execute()
         if (hasTrailingFileposition(helper.queryDiskRecordSize()->queryTypeInfo()))
             keyMaxSize -= sizeof(offset_t);
 
-        Owned<IFileIOStream> out = createIOStream(io, needsSeek);
+        Owned<IFileIOStream> out = createBufferedIOStream(io, 0x100000);
+        if (!needsSeek)
```
richardkchapman, line 1161: Feels like this makes it harder to implement a noseek stream that actually took advantage of the fact that it was not seeking (as opposed to merely checking)

ghalliday: I think it is likely to require a new interface to implement, and at that point we should switch to only supporting noseek indexes.
(E.g. have an interface where you request blocks of memory, can write directly into the output buffer, and then at a later date commit it, which will indicate that the data can be flushed.)  It would require some intelligence in the implementation, but could end up being much more efficient.

ghalliday:
@richardkchapman are you happy to merge this?  (#13837 is the PR for the 1st commit)

PR #13848, opened 2020-06-10
HPCC-24218 Maintain cross ESDL transform state and add new functions

afishbeck:
Use a scriptContext to support keeping script state for the entire transaction.

Allowing for the upcomming addition of multiple script entry points.

Add new script functions to add to scriptContext state:
    set-log-profile - signal which logging profile should be followed
    set-log-option - control logging options
    store-value - store values that can be accessed by other scripts

Add new script functions (single node operations for now):
   remove-node - remove a single node from the target tree
   rename-node - rename a single node on the target tree

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/esdllib/esdl_script.cpp
+++ b/esp/esdllib/esdl_script.cpp
@@ -22,7 +22,7 @@
 interface IEsdlTransformOperation : public IInterface
 {
     virtual const char *queryMergedTarget() = 0;
-    virtual bool process(IEspContext * context, IPropertyTree *content, IXpathContext * xpathContext) = 0;
+    virtual bool process(IEsdlScriptContext * context, IPropertyTree *content, IXpathContext * xpathContext) = 0;
     virtual void toDBGLog() = 0;
 };
 
@@ -71,12 +71,15 @@ class CEsdlTransformOperationBase : public CInterfaceOf<IEsdlTransformOperation>
 protected:
     StringAttr m_mergedTarget;
     StringAttr m_tagname;
+    StringAttr m_traceName;
     bool m_ignoreCodingErrors = false; //ideally used only for debugging
 
 public:
     CEsdlTransformOperationBase(IPropertyTree *tree, const StringBuffer &prefix)
     {
         m_tagname.set(tree->queryName());
+        if (tree->hasProp("@trace"))
+            m_traceName.set(tree->queryProp("@trace"));
         if (tree->hasProp("@_crtTarget"))
             m_mergedTarget.set(tree->queryProp("@_crtTarget"));
         m_ignoreCodingErrors = tree->getPropBool("@optional", false);
@@ -98,9 +101,11 @@ protected:
 public:
     CEsdlTransformOperationVariable(IPropertyTree *tree, const StringBuffer &prefix) : CEsdlTransformOperationBase(tree, prefix)
     {
+        if (m_traceName.isEmpty())
+            m_traceName.set(tree->queryProp("@name"));
         m_name.set(tree->queryProp("@name"));
         if (m_name.isEmpty())
-            esdlOperationError(ESDL_SCRIPT_MissingOperationAttr, m_tagname, "without name", m_name, !m_ignoreCodingErrors);
+            esdlOperationError(ESDL_SCRIPT_MissingOperationAttr, m_tagname, "without name", m_traceName, !m_ignoreCodingErrors);
         const char *select = tree->queryProp("@select");
         if (!isEmptyString(select))
             m_select.setown(compileXpath(select));
@@ -110,7 +115,7 @@ public:
     {
     }
 
-    virtual bool process(IEspContext * context, IPropertyTree *content, IXpathContext * xpathContext) override
+    virtual bool process(IEsdlScriptContext * context, IPropertyTree *content, IXpathContext * xpathContext) override
     {
         if (m_select)
             return xpathContext->addCompiledVariable(m_name, m_select);
@@ -136,12 +141,119 @@ public:
     {
     }
 
-    virtual bool process(IEspContext * context, IPropertyTree *content, IXpathContext * xpathContext) override
+    virtual bool process(IEsdlScriptContext * context, IPropertyTree *content, IXpathContext * xpathContext) override
     {
         if (m_select)
             return xpathContext->declareCompiledParameter(m_name, m_select);
         return xpathContext->declareParameter(m_name, "");
     }
+};
+
+class CEsdlTransformOperationSetSectionAttributeBase : public CEsdlTransformOperationBase
+{
+protected:
+    StringAttr m_name;
+    Owned<ICompiledXpath> m_xpath_name;
+    Owned<ICompiledXpath> m_select;
+
+public:
+    CEsdlTransformOperationSetSectionAttributeBase(IPropertyTree *tree, const StringBuffer &prefix, const char *attrName) : CEsdlTransformOperationBase(tree, prefix)
+    {
+        if (m_traceName.isEmpty())
+            m_traceName.set(tree->queryProp("@name"));
+        if (!isEmptyString(attrName))
+            m_name.set(attrName);
+        else
+        {
+            m_name.set(tree->queryProp("@name"));
+
+            const char *xpath_name = tree->queryProp("@xpath_name");
+            if (!isEmptyString(xpath_name))
+                m_xpath_name.setown(compileXpath(xpath_name));
+            else if (m_name.isEmpty())
+                esdlOperationError(ESDL_SCRIPT_MissingOperationAttr, m_tagname, "without name", m_traceName, !m_ignoreCodingErrors); //don't mention value, it's deprecated
+        }
+
+        const char *select = tree->queryProp("@select");
+        if (isEmptyString(select))
+            esdlOperationError(ESDL_SCRIPT_MissingOperationAttr, m_tagname, "without select", m_traceName, !m_ignoreCodingErrors); //don't mention value, it's deprecated
+        m_select.setown(compileXpath(select));
+    }
+
+    virtual void toDBGLog() override
+    {
+#if defined(_DEBUG)
+        DBGLOG(">%s> %s(name(%s), select('%s'))", m_traceName.str(), m_tagname.str(), (m_xpath_name) ? m_xpath_name->getXpath() : m_name.str(), m_select->getXpath());
+#endif
+    }
+
+    virtual ~CEsdlTransformOperationSetSectionAttributeBase(){}
+
+    virtual const char *getSectionName() = 0;
+
+    virtual bool process(IEsdlScriptContext * context, IPropertyTree *content, IXpathContext * xpathContext) override
+    {
+        if ((!m_name && !m_xpath_name) || !m_select)
+            return false; //only here if "optional" backward compatible support for now (optional syntax errors aren't actually helpful)
+        try
+        {
+            StringBuffer name;
+            if (m_xpath_name)
+                xpathContext->evaluateAsString(m_xpath_name, name);
+            else
+                name.set(m_name);
+
+            StringBuffer value;
+            xpathContext->evaluateAsString(m_select, value);
+            context->setAttribute(getSectionName(), name, value);
+        }
+        catch (IException* e)
+        {
+            int code = e->errorCode();
+            StringBuffer msg;
+            e->errorMessage(msg);
+            e->Release();
+            esdlOperationError(code, m_tagname, msg, m_traceName, !m_ignoreCodingErrors);
+        }
+        catch (...)
+        {
+            esdlOperationError(ESDL_SCRIPT_Error, m_tagname, "unknown exception processing", m_traceName, !m_ignoreCodingErrors);
+        }
+        return false;
+    }
+};
+
+class CEsdlTransformOperationStoreValue : public CEsdlTransformOperationSetSectionAttributeBase
+{
+public:
+    CEsdlTransformOperationStoreValue(IPropertyTree *tree, const StringBuffer &prefix) : CEsdlTransformOperationSetSectionAttributeBase(tree, prefix, nullptr)
+    {
+    }
+
+    virtual ~CEsdlTransformOperationStoreValue(){}
+    const char *getSectionName() override {return ESDLScriptCtxSection_Store;}
+};
+
+class CEsdlTransformOperationSetLogProfile : public CEsdlTransformOperationSetSectionAttributeBase
+{
+public:
+    CEsdlTransformOperationSetLogProfile(IPropertyTree *tree, const StringBuffer &prefix) : CEsdlTransformOperationSetSectionAttributeBase(tree, prefix, "profile")
+    {
+    }
+
+    virtual ~CEsdlTransformOperationSetLogProfile(){}
+    const char *getSectionName() override {return ESDLScriptCtxSection_Logging;}
+};
+
+class CEsdlTransformOperationSetLogOption : public CEsdlTransformOperationSetSectionAttributeBase
+{
+public:
+    CEsdlTransformOperationSetLogOption(IPropertyTree *tree, const StringBuffer &prefix) : CEsdlTransformOperationSetSectionAttributeBase(tree, prefix, nullptr)
+    {
+    }
+
+    virtual ~CEsdlTransformOperationSetLogOption(){}
+    const char *getSectionName() override {return ESDLScriptCtxSection_Logging;}
 };
 
 class CEsdlTransformOperationSetValue : public CEsdlTransformOperationBase
@@ -150,12 +262,12 @@ protected:
     Owned<ICompiledXpath> m_select;
     Owned<ICompiledXpath> m_xpath_target;
     StringAttr m_target;
-    StringAttr m_traceName;
 
 public:
     CEsdlTransformOperationSetValue(IPropertyTree *tree, const StringBuffer &prefix) : CEsdlTransformOperationBase(tree, prefix)
     {
-        m_traceName.set(tree->queryProp("@name"));
+        if (m_traceName.isEmpty() && tree->hasProp("@name"))
+            m_traceName.set(tree->queryProp("@name"));
 
         if (isEmptyString(tree->queryProp("@target")) && isEmptyString(tree->queryProp("@xpath_target")))
             esdlOperationError(ESDL_SCRIPT_MissingOperationAttr, m_tagname.str(), "without target", m_traceName.str(), !m_ignoreCodingErrors);
@@ -182,7 +294,7 @@ public:
 
     virtual ~CEsdlTransformOperationSetValue(){}
 
-    virtual bool process(IEspContext * context, IPropertyTree *content, IXpathContext * xpathContext) override
+    virtual bool process(IEsdlScriptContext * context, IPropertyTree *content, IXpathContext * xpathContext) override
     {
         if ((!m_xpath_target && m_target.isEmpty()) || !m_select)
             return false; //only here if "optional" backward compatible support for now (optional syntax errors aren't actually helpful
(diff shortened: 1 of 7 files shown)


system/xmllib/libxml_xpathprocessor.cpp
```diff
@@ -775,6 +789,352 @@ extern ICompiledXpath* compileXpath(const char * xpath)
...
+        xmlNodePtr sect = replaceSection(section);
+        if (tree==nullptr) //means delete content
+            return;
+        addChildFromPtree(sect, *tree, tree->queryName());
+    }
+
+    virtual void setAttribute(const char *section, const char *name, const char *value) override
+    {
+        xmlNodePtr sect = ensureSection(section);
+        xmlNewProp(sect, (const xmlChar *)name, (const xmlChar *)value);
```
timothyklemm, line 1068: Not sure this is sufficient. Consider a use case where an attribute could be set to a default, then a subsequent operation intends to override the default. My observation is that the first operation succeeds and the subsequent operation is ignored. xmlSetProp will update a node, but may not (documentation isn't clear) insert.

On a tangent, it might also make sense to have operations that can remove an attribute (using xmlUnsetProp). The ability to remove a previously applied attribute might be the preferred approach to overriding.

esp/esdllib/esdl_script.cpp
```diff
@@ -98,9 +101,11 @@ class CEsdlTransformOperationVariable : public CEsdlTransformOperationBase
 public:
     CEsdlTransformOperationVariable(IPropertyTree *tree, const StringBuffer &prefix) : CEsdlTransformOperationBase(tree, prefix)
     {
+        if (m_traceName.isEmpty())
+            m_traceName.set(tree->queryProp("@name"));
```
timothyklemm, line 105: This seems to be redundant, to the point that it could probably make sense moving in to the base class. If the base class handles m_traceName, it makes sense for it to handle fall-back options (could maybe add an optional fall-back name to the constructor parameter list to not assume).

I'm not sure I would go so far as to move storage for @name into the base, although I probably wouldn't complain if it happened.

esp/esdllib/esdl_script.cpp
```diff
@@ -769,155 +1046,145 @@ class CEsdlCustomTransform : public CInterfaceOf<IEsdlCustomTransform>
...
-        //in case transform wants to make use of these values:
-        //make them few well known values variables rather than inputs so they are automatically available
-        xpathContext->addVariable("query", tgtcfg->queryProp("@queryname"));
-        xpathContext->addVariable("method", method);
-        xpathContext->addVariable("service", service);
-        xpathContext->addVariable("request", reqtype);
+    //in case transform wants to make use of these values:
+    //make them few well known values variables rather than inputs so they are automatically available
+    StringBuffer temp;
+    xpathContext->addVariable("query", scriptCtx->getXPathString("target/*/@queryname", temp));
```
asselitx, line 1108: Why not also move this statement for the 'query' variable into the IXpathContext* createXpathContext() function in libxml_xpathprocessor.cpp where the other similar statements moved?

system/xmllib/libxml_xpathprocessor.cpp
```diff
@@ -775,6 +789,352 @@ extern ICompiledXpath* compileXpath(const char * xpath)
...
+        toXML(xml, nullptr, true);
+    }
+    IXpathContext* createXpathContext(const char *section, bool strictParameterDeclaration) override
+    {
+        xmlNodePtr sect = getSectionNode(section);
+        if (!sect)
+            throw MakeStringException(-1, "CEsdlScriptContext:createXpathContext: section not found %s", section);
+        CLibXpathContext *xpathContext = new CLibXpathContext(doc, sect, strictParameterDeclaration);
+        StringBuffer val;
+        xpathContext->addVariable("method", getAttribute("esdl", "method", val));
```
asselitx, line 1120: Why did you spit  up the initialization of the xpathContext between this createXpathContext() function and the processServiceAndMethodTransforms() method? I found it nice to see all the context setup in that one function- made it easy to tell which values were treated like parameters and which like implicit variables.

esp/esdllib/esdl_xpath_extensions_libxml.cpp
```diff
@@ -176,7 +189,163 @@ static void getFeatureSecAccessFlagsFunction (xmlXPathParserContextPtr ctxt, int
...
+static void getStoredStringValueFunction (xmlXPathParserContextPtr ctxt, int nargs)
+{
+    IEsdlScriptContext *scriptContext = getEsdlScriptContext(ctxt);
+    if (!scriptContext)
+    {
+        xmlXPathSetError((ctxt), XPATH_INVALID_CTXT);
+        return;
+    }
+
+    if (nargs != 1)
```
asselitx, line 207: Why must nargs==1? It is not used for anything?

esp/esdllib/esdl_script.cpp
afishbeck: I'll take a look but @name is only a fallback for trace on commands where it is already expected.  name is too valuable as a parameter to use it to mean trace.  So I want to avoid that in the future.  When dealing with name + value situations it's a shame if we can't make it intuitive because of how we did tracing.

system/xmllib/libxml_xpathprocessor.cpp
```diff
@@ -775,6 +789,352 @@ extern ICompiledXpath* compileXpath(const char * xpath)
...
+    ForEach(*children)
+        addChildFromPtree(node, children->query(), nullptr);
+}
+
+IPropertyTree *createPtreeFromXmlNode(xmlNodePtr node);
+
+void copyXmlNode(IPropertyTree *tree, xmlNodePtr node)
+{
+    for(xmlAttrPtr att=node->properties;att!=nullptr; att=att->next)
+        tree->setProp((const char *)att->name, (const char *)xmlGetProp(node, att->name));
```
asselitx, line 859: Do we need to manage the memory of the xmlChar* returned from xmlGetProp here? Or is it just returning a pointer to the xmlChar* it retains ownership of?

system/xmllib/libxml_xpathprocessor.cpp
```diff
@@ -775,6 +789,352 @@ extern ICompiledXpath* compileXpath(const char * xpath)
...
+static const char *queryAttrValue(xmlNodePtr node, const char *name)
+{
+    if (!node)
+        return nullptr;
+    xmlAttrPtr att = xmlHasProp(node, (const xmlChar *)name);
+    if (!att)
+        return nullptr;
+    if (att->type != XML_ATTRIBUTE_NODE || att->children==nullptr || (att->children->type != XML_TEXT_NODE && att->children->type != XML_CDATA_SECTION_NODE))
+        return nullptr;
+    return (const char *) att->children->content;
```
asselitx, line 883: So an attribute's value is its child node if that node is TEXT or CDATA?

system/xmllib/libxml_xpathprocessor.cpp
```diff
@@ -775,6 +789,352 @@ extern ICompiledXpath* compileXpath(const char * xpath)
...
+}
+
+class CEsdlScriptContext : public CInterfaceOf<IEsdlScriptContext>
+{
+private:
+    void *espCtx = nullptr;
```
asselitx, line 889: Why not store as an IEspContext *?

system/xmllib/libxml_xpathprocessor.cpp
```diff
@@ -775,6 +789,352 @@ extern ICompiledXpath* compileXpath(const char * xpath)
...
+    }
+    virtual void setContent(const char *section, const char *xml) override
+    {
+        xmlNodePtr sect = replaceSection(section);
+        if (xml==nullptr) //means delete content
+            return;
+
+        xmlParserCtxtPtr parserCtx = xmlCreateDocParserCtxt((const unsigned char *)xml);
+        if (!parserCtx)
+            throw MakeStringException(-1, "CEsdlScriptContext:setXmlSection: Unable to init parse of %s XML content", section);
```
asselitx, line 1047: Error message refers to incorrect function name.

system/xmllib/libxml_xpathprocessor.cpp
```diff
@@ -775,6 +789,352 @@ extern ICompiledXpath* compileXpath(const char * xpath)
...
+//        parserCtx->myDoc = doc;
+        parserCtx->node = sect;
+        xmlParseDocument(parserCtx);
+        int wellFormed = parserCtx->wellFormed;
+        xmlFreeDoc(parserCtx->myDoc); //dummy document
```
asselitx, line 1052: Should this still be in here? I see you commented out a line above setting parserCtx->myDoc.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -1447,7 +1447,21 @@ void EsdlServiceImpl::prepareFinalRequest(IEspContext &context,
...
         context.addTraceSummaryTimeStamp(LogNormal, "srt-custreqtrans");
-        processServiceAndMethodTransforms({serviceCrt, methodCrt}, &context, tgtcfg, srvdef, mthdef, reqProcessed, m_oEspBindingCfg.get());
+
+        //as we add more entry points the script context will move to wider scope
+        Owned<IEsdlScriptContext> scriptContext = createEsdlScriptContext(&context);
+        scriptContext->setContent("ESDLRequest", reqProcessed.str());
+        scriptContext->setAttribute("esdl", "service", srvdef.queryName());
+        scriptContext->setAttribute("esdl", "method", mthdef.queryMethodName());
+        scriptContext->setAttribute("esdl", "request_type", mthdef.queryRequestType());
+        scriptContext->setAttribute("esdl", "request", mthdef.queryRequestType());
```
asselitx, line 1457: Why do we have different attributes for request and request_type?

system/xmllib/libxml_xpathprocessor.cpp
```diff
@@ -775,6 +789,352 @@ extern ICompiledXpath* compileXpath(const char * xpath)
...
+    }
+    virtual void setContent(const char *section, IPropertyTree *tree) override
+    {
+        xmlNodePtr sect = replaceSection(section);
+        if (tree==nullptr) //means delete content
+            return;
+        addChildFromPtree(sect, *tree, tree->queryName());
+    }
+
+    virtual void setAttribute(const char *section, const char *name, const char *value) override
```
asselitx, line 1065: Maybe helpful looking toward the future, perhaps we should use some enumerations or constants for the section and name values used here. We might not want to *limit* it to an enumeration, but some standard constants would help with consistency.

system/xmllib/libxml_xpathprocessor.cpp
afishbeck: Agreed I'll change it to xmlSetProp.  I've looked at the code and that does create or update.

I'll also look at removal.

system/xmllib/libxml_xpathprocessor.cpp
afishbeck: Exactly I didn't want to expose the common library to ESP internals.  We could move things around or share some fairly isolated headers.  I think IEspContext may carry a lot of baggage.

system/xmllib/libxml_xpathprocessor.cpp
afishbeck: Setting myDoc above ends up serving no purpose but leaks this dummy doc.  So I had to do it this way not to leak.  If we ever find a problem I will have to switch gears, parse into a new doc, and then adopt the content into the main doc.  But I think this actually works if a bit confusing.

esp/services/esdl_svc_engine/esdl_binding.cpp
afishbeck: We use them as the same right now but I could see them sometimes changing.  We just happen to support the request tag that is named after the typename.  May never need to separate but I wanted to make it clear that they may not always be the same.

system/xmllib/libxml_xpathprocessor.cpp
afishbeck: I agree we could use constants but they would probably be in the esdl binding.  The esdl script can handle any names, the binding sets our model for what our entry points actually are.  Although some may be common like "store" so could consider some being in esdl_script and some being in the esdl service/binding.

esp/esdllib/esdl_script.cpp
```diff
@@ -136,26 +141,133 @@ class CEsdlTransformOperationParameter : public CEsdlTransformOperationVariable
...
+        }
+        catch (...)
+        {
+            esdlOperationError(ESDL_SCRIPT_Error, m_tagname, "unknown exception processing", m_traceName, !m_ignoreCodingErrors);
+        }
+        return false;
+    }
+};
+
+class CEsdlTransformOperationStoreValue : public CEsdlTransformOperationSetSectionAttributeBase
```
timothyklemm, line 226: Would there be value allowing more complex values to be stored? If we think we might want to store more structured data in the future, we might not want to rely on attributes now. I'm not saying we will. I'm just asking if it was considered.

esp/esdllib/esdl_xpath_extensions_libxml.cpp
```diff
@@ -198,9 +367,19 @@ void registerEsdlXPathExtensions(IXpathContext *xpathContext, IEspContext *conte
         xpathContext->registerFunction(nullptr, "validateFeaturesAccess", (void  *)validateFeaturesAccessFunction);
         xpathContext->registerFunction(nullptr, "secureAccessFlags", (void  *)secureAccessFlagsFunction);
         xpathContext->registerFunction(nullptr, "getFeatureSecAccessFlags", (void  *)getFeatureSecAccessFlagsFunction);
+        xpathContext->registerFunction(nullptr, "getStoredStringValue", (void  *)getStoredStringValueFunction);
+        xpathContext->registerFunction(nullptr, "storedValueExists", (void  *)storedValueExistsFunction);
+        xpathContext->registerFunction(nullptr, "getLogProfile", (void  *)getLogProfileFunction);
+        xpathContext->registerFunction(nullptr, "getLogOption", (void  *)getLogOptionFunction);
+        xpathContext->registerFunction(nullptr, "logOptionExists", (void  *)logOptionExistsFunction);
```
timothyklemm, line 374: Because I know the implementation details I could use this to test for a selected log profile, but should I have to? It seems like logProfileExists should be defined.

esp/esdllib/esdl_xpath_extensions_libxml.cpp
```diff
@@ -198,9 +367,19 @@ void registerEsdlXPathExtensions(IXpathContext *xpathContext, IEspContext *conte
...
     }
 
     xpathContext->registerFunction("urn:hpcc:esdl:script", "validateFeaturesAccess", (void  *)validateFeaturesAccessFunction);
     xpathContext->registerFunction("urn:hpcc:esdl:script", "secureAccessFlags", (void  *)secureAccessFlagsFunction);
     xpathContext->registerFunction("urn:hpcc:esdl:script", "getFeatureSecAccessFlags", (void  *)getFeatureSecAccessFlagsFunction);
+    xpathContext->registerFunction("urn:hpcc:esdl:script", "getStoredStringValue", (void  *)getStoredStringValueFunction);
+    xpathContext->registerFunction("urn:hpcc:esdl:script", "storedValueExists", (void  *)storedValueExistsFunction);
+    xpathContext->registerFunction("urn:hpcc:esdl:script", "getLogProfile", (void  *)getLogProfileFunction);
+    xpathContext->registerFunction("urn:hpcc:esdl:script", "getLogOption", (void  *)getLogOptionFunction);
+    xpathContext->registerFunction("urn:hpcc:esdl:script", "logOptionExists", (void  *)logOptionExistsFunction);
```
timothyklemm, line 384: This looks like a candidate for a helper function, passing the urn and registering all functions. As the function list grows it will be more helpful.

esp/esdllib/esdl_script.cpp
```diff
@@ -136,26 +141,133 @@ class CEsdlTransformOperationParameter : public CEsdlTransformOperationVariable
...
+{
+public:
+    CEsdlTransformOperationStoreValue(IPropertyTree *tree, const StringBuffer &prefix) : CEsdlTransformOperationSetSectionAttributeBase(tree, prefix, nullptr)
+    {
+    }
+
+    virtual ~CEsdlTransformOperationStoreValue(){}
+    const char *getSectionName() override {return "store";}
```
timothyklemm, line 234: The string literals should be defined names. Define the convention now, and hopefully avoid inconsistencies with future extensions.

esp/esdllib/esdl_script.cpp
```diff
@@ -226,6 +338,152 @@ class CEsdlTransformOperationSetValue : public CEsdlTransformOperationBase
...
+
+            StringBuffer name;
+            if (m_xpath_new_name)
+                xpathContext->evaluateAsString(m_xpath_new_name, name);
+            else
+                name.set(m_new_name);
+
+            //make sure we only remove one.  we can add a multi-node support in the future (when we are libxml2 based?)
+            //don't want users to depend on quirky behavior
+            path.append("[1]"); //don't need to check as multiple [] are ok
```
timothyklemm, line 400: With the libxml2 version of this, my intent was to pass an attribute specifying if all occurrences were to be affected. If not applied to all, a multiple node nodeset would be considered a failure. Just picking one of many seems like quirky behavior best to be avoided.

esp/esdllib/esdl_script.cpp
```diff
@@ -226,6 +338,152 @@ class CEsdlTransformOperationSetValue : public CEsdlTransformOperationBase
...
+            return false; //only here if "optional" backward compatible support for now (optional syntax errors aren't actually helpful
+        try
+        {
+            StringBuffer path;
+            if (m_xpath_target)
+                xpathContext->evaluateAsString(m_xpath_target, path);
+            else
+                path.set(m_target);
+
+            if (isWildString(path))
```
timothyklemm, line 463: It seems like we should just evaluate the path and delete a single node, if that is what getElements gives us. Unless this required because getElements can't evaluate the wildcard.

esp/esdllib/esdl_script.cpp
```diff
@@ -226,6 +338,152 @@ class CEsdlTransformOperationSetValue : public CEsdlTransformOperationBase
...
+                esdlOperationError(ESDL_SCRIPT_MissingOperationAttr, m_tagname.str(), "wildcard in target not yet supported", m_traceName.str(), !m_ignoreCodingErrors);
+
+            //make sure we only remove one.  we can add a multi-node support in the future (when we are libxml2 based?)
+            //don't want users to depend on quirky behavior
+            path.append("[1]"); //don't need to check as multiple [] are ok
```
timothyklemm, line 468: Same comment as with rename-node. Also, can this be used within a for-each to remove all matching nodes?

esp/esdllib/esdl_script.cpp
```diff
@@ -769,155 +1046,145 @@ class CEsdlCustomTransform : public CInterfaceOf<IEsdlCustomTransform>
...
     LogLevel level = LogMin;
+    if (!scriptCtx)
+        return;
     if (!transforms.size())
         return;
-    if (tgtcfg)
-        level = (unsigned) tgtcfg->getPropInt("@traceLevel", level);
+    if (isEmptyString(srcSection)||isEmptyString(tgtSection))
+        return;
+    level = scriptCtx->getXPathInt64("target/*/@traceLevel", level);
```
timothyklemm, line 1064: LogLevel is unsigned, not __int64. This should be generating a compiler warning for potential loss of data.

esp/esdllib/esdl_script.cpp
```diff
@@ -769,155 +1046,145 @@ class CEsdlCustomTransform : public CInterfaceOf<IEsdlCustomTransform>
...
+    }
 
-        toXML(theroot, content.clear());
+    scriptCtx->setContent(tgtSection, theroot);
 
-        if (level >= LogMax)
-            DBGLOG(1,"MODIFIED content: %s", content.str());
+    if (level >= LogMax)
+    {
+        scriptCtx->toXML(content.clear(), tgtSection);
```
timothyklemm, line 1183: This is helpful, but I found it to be more helpful changing this to the entire document instead of one section. With updates to sections outside of this one possible, seeing the whole document may be more useful.

system/xmllib/libxml_xpathprocessor.cpp
```diff
@@ -775,6 +789,352 @@ extern ICompiledXpath* compileXpath(const char * xpath)
     return new CLibCompiledXpath(xpath);
 }
 
+xmlNsPtr createNsFromXmlns(xmlNodePtr node, const char *attname, const char *uri, xmlNsPtr ns)
```
timothyklemm, line 792: Am I missing something, or is this unused?

system/xmllib/libxml_xpathprocessor.cpp
```diff
@@ -775,6 +789,352 @@ extern ICompiledXpath* compileXpath(const char * xpath)
...
+                xmlNewProp(node, (const xmlChar *)attname, (const xmlChar *)atts->queryValue());
+        }
+    }
+    if (prefix.isEmpty())
+        node->ns = nullptr;
+    else
+    {
+        xmlNsPtr ns = xmlSearchNs(node->doc, node, (const xmlChar *)prefix.str());
+        if (!ns) //invalid but let's repair it
+            ns = xmlNewNs(node, (const xmlChar *)"urn:hpcc:unknown", (const xmlChar *)prefix.str());
```
timothyklemm, line 846: What use case are we enabling that cannot be described as "sloppy"? There may be a case we need to handle this way, but I can't think of it.

system/xmllib/libxml_xpathprocessor.cpp
timothyklemm: We own the block returned from xmlGetProp, and are required to free it.

system/xmllib/xpathprocessor.hpp
```diff
@@ -81,4 +81,22 @@ class CXpathContextScope : CInterface
...
+    virtual IXpathContext* createXpathContext(const char *section, bool strictParameterDeclaration) = 0;
+    virtual void *queryEspContext() = 0;
+    virtual void setContent(const char *section, const char *xml) = 0;
+    virtual void setContent(const char *section, IPropertyTree *tree) = 0;
+    virtual void setAttribute(const char *section, const char *name, const char *value) = 0;
+    virtual const char *queryAttribute(const char *section, const char *name) = 0;
+    virtual const char *getAttribute(const char *section, const char *name, StringBuffer &s) = 0;
+    virtual const char *getXPathString(const char *xpath, StringBuffer &s) const = 0;
+    virtual __int64 getXPathInt64(const char *xpath, __int64 dft=0) const = 0;
+    virtual bool getXPathBool(const char *xpath, bool dft=false) const = 0;
```
timothyklemm, line 95: I know this naming fits existing code style (like PTree), but I'm not a fan. My preference would be to follow the design of IEspSecureContext. It would handle unsigned versus signed, or floating point versus integer, or short versus long.

A similar change could be made to getAttribute and setAttribute.

esp/esdllib/esdl_script.cpp
afishbeck: yes I did think about it.  I'm just leaving the actual content of the section for future consideration.  If we need something more than key value pairs I'd like to understand it before adding support... but will do so if we need it.

Also it will be more valuable when we have better ways of building structures.  Perhaps after we've moved away from IPtree on the right hand side.

esp/esdllib/esdl_script.cpp
afishbeck: Ok agreed.

esp/esdllib/esdl_script.cpp
afishbeck: I can throw an exception if multiple... and leave it to the script writer to add the [1] if desired.  There are cases where since this state can creep up externally that causing the whole transaction to fail could be undesirable.  This will mean script writers need to be diligent of this at all times.

esp/esdllib/esdl_script.cpp
afishbeck: We are not calling getElements. If I remember correclty removeProp did not like wildcards.

esp/esdllib/esdl_script.cpp
afishbeck: how do we get away with less than 9,223,372,036,854,775,807 log levels?

esp/esdllib/esdl_xpath_extensions_libxml.cpp
afishbeck: You could just check if getLogProfile returns content.  I'll think about it.

esp/esdllib/esdl_xpath_extensions_libxml.cpp
afishbeck: agreed the list is getting long enough.

system/xmllib/libxml_xpathprocessor.cpp
afishbeck: It's probably from a prior iteration, I'll check.

system/xmllib/libxml_xpathprocessor.cpp
afishbeck: I don't want to have to revisit just because we've got a ptree from somewhere that we want to be able to process and ptrees are notoriously bad about namespacing.  This is so I never have to look at this code again. ;)

esp/esdllib/esdl_script.cpp
timothyklemm: I guess I was thinking that more complex data would suggest child elements. We could use child elements with name and value for the current key value pairs, then for more complex data replace the value attribute with content. Just thought it might be more consistent when we grow into it. Either way works.

esp/esdllib/esdl_script.cpp
timothyklemm: I know it isn't using getElements. This was my way of suggesting it should use getElements. I just think taking '*' away, which is useful for more than looking for multiples, is undesirable.

esp/esdllib/esdl_script.cpp
timothyklemm: I didn't think implementing 128 bit integers was within the scope of this story, but if you think more options make sense...

richardkchapman:
@afishbeck I think this is with you at the moment?

esp/esdllib/esdl_script.cpp
afishbeck: So the getXPathInt64 is so we can support other values that might require large ints.  But I'll add a specific cast here so the compiler wont care.  I don't want to add mulitple getInt functions if I don't have to.

esp/esdllib/esdl_script.cpp
afishbeck: may move this in the future.  I'm just not sure where I want knowledge of the Target config section.  The other values are all from the ESDL section which comes from specific values in the ESDL definition.

esp/esdllib/esdl_script.cpp
afishbeck: I was emulating what was already logged here, but I can make it the whole blob for now.  That may end up being to verbose, especially as it grows, but we can change it in the future.

esp/esdllib/esdl_xpath_extensions_libxml.cpp
afishbeck: Running out of time for the initial 7.10 RC.  We can add a new xpath function for this later.  Just get if getLogProfile has content for now.

system/xmllib/libxml_xpathprocessor.cpp
afishbeck: It's of type XML_ATTRIBUTE_NODE content.  A CDATA node is a different type.  In this "queryAttrValue() function I am relying on the fact that these are simple attributes and not a compounding of multiple parts like some elements and perhaps more complex attributes might be.  If we try and use this function for uses that don't match this assumption we may need to make it less efficient and have to allocate and delete memory.  For now this one method errors on the side of efficiency.

system/xmllib/libxml_xpathprocessor.cpp
afishbeck: Changing to xmlSetProp.  And changing this function to remove a prop if value is nullptr.  But leaving adding that capability to scripts for future consideration.

system/xmllib/libxml_xpathprocessor.cpp
afishbeck: I thought these values could be understood by the script context, but wasn't sure about some others.  We can consider reorganizing further in the future.

system/xmllib/xpathprocessor.hpp
afishbeck: Can consider refactoring in the future.  At least this is following a precedent.  I don't think this is currently an interface that we can't easily change.

afishbeck:
@asselitx @timothyklemm updated, please review again.  Trying to get this in tomorrows release candidate.

esp/esdllib/esdl_xpath_extensions_libxml.cpp
```diff
@@ -176,7 +189,175 @@ static void getFeatureSecAccessFlagsFunction (xmlXPathParserContextPtr ctxt, int
...
+    {
+        xmlXPathSetArityError(ctxt);
+        return;
+    }
+
+    xmlChar *namestring = xmlXPathPopString(ctxt);
+    if (xmlXPathCheckError(ctxt)) //includes null check
+        return;
+
+    const char *value = scriptContext->queryAttribute("logging", (const char *)namestring);
```
timothyklemm, line 250: This comment applies to multiple functions. The literal, or literals elsewhere, should be using the defined names.

system/xmllib/libxml_xpathprocessor.cpp
```diff
@@ -833,6 +848,346 @@ extern ICompiledXpath* compileXpath(const char * xpath)
...
+        return getSectionNode(name, nullptr);
+    }
+    xmlNodePtr ensureSection(const char *name)
+    {
+        xmlNodePtr sect = getSection(name);
+        if (sect)
+            return sect;
+        return xmlNewChild(root, nullptr, (const xmlChar *) name, nullptr);
+    }
+    void removeSection(const char *name)
```
timothyklemm, line 1063: With this name I would I would expect it to call getSection and delete the node. The implementation looks like it removes any arbitrary nodes, which may include a section node.

system/xmllib/libxml_xpathprocessor.cpp
```diff
@@ -833,6 +848,346 @@ extern ICompiledXpath* compileXpath(const char * xpath)
...
+    {
+        xmlXPathObjectPtr eval = xmlXPathEval((const xmlChar *) name, xpathCtx);
+        if (!eval)
+            return;
+        if (XPATH_NODESET==eval->type && !xmlXPathNodeSetIsEmpty(eval->nodesetval))
+        {
+            xmlNodeSetPtr ns = eval->nodesetval;
+            int count = xmlXPathNodeSetGetLength(ns);
+            for(int i=0; i<count; i++)
```
timothyklemm, line 1072: Based on a comment in the example at http://www.xmlsoft.org/examples/xpath2.c, should the loop traverse in the opposite order? The evaluation could return a node and that node's descendant, and freeing the first node frees the descendant, invalidating the node reference.

system/xmllib/libxml_xpathprocessor.cpp
```diff
@@ -833,6 +848,346 @@ extern ICompiledXpath* compileXpath(const char * xpath)
...
+                    ns->nodeTab[i]=nullptr;
+                }
+            }
+        }
+        xmlXPathFreeObject(eval);
+    }
+    xmlNodePtr replaceSection(const char *name)
+    {
+        removeSection(name);
+        return xmlNewChild(root, nullptr, (const xmlChar *) name, nullptr);
```
timothyklemm, line 1088: Should there be verification that 'name' is an element name and not a path?

system/xmllib/libxml_xpathprocessor.cpp
afishbeck: This is fault tolerant if for some reason multiple nodes exist.  Should never happen but might as well clean them up.

system/xmllib/libxml_xpathprocessor.cpp
afishbeck: Agreed.

afishbeck:
@timothyklemm updated based on your comments.

esp/esdllib/esdl_xpath_extensions_libxml.cpp
```diff
@@ -176,7 +189,175 @@ static void getFeatureSecAccessFlagsFunction (xmlXPathParserContextPtr ctxt, int
...
+    {
+        xmlXPathSetArityError(ctxt);
+        return;
+    }
+
+    xmlChar *namestring = xmlXPathPopString(ctxt);
+    if (xmlXPathCheckError(ctxt)) //includes null check
+        return;
+
+    const char *value = scriptContext->queryAttribute("store", (const char *)namestring);
```
timothyklemm, line 342: "store" should be ESDLScriptCtxSection_Store.

system/xmllib/libxml_xpathprocessor.cpp
timothyklemm: If setContent is called with an XPath and not a section name, replaceSection will act on the XPath by allowing removeSection to remove all matching occurrences. The new child of 'root' might not be created, but we've already potentially destroyed the document.

Whether it's here or in setContent, it seems like we need to verify that we're manipulating a section and not an arbitrary location. Otherwise it seems like a loophole waiting to be abused.

asselitx:
@afishbeck Your changes look good and they pass the tests against existing transforms.

system/xmllib/libxml_xpathprocessor.cpp
afishbeck: I hope that people who knowingly try to pass in an xpath for the name get what they deserve when the document is destroyed.  But I do now check name is not a path via the call to sanityCheckSectionName().

afishbeck:
@richardkchapman this can be merged and included in RC1.

PR #13855, opened 2020-06-12
HPCC-24235 Simplified bundle support on cloud installs

richardkchapman:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [x] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/helm/hpcc/templates/dali.yaml
+++ b/helm/hpcc/templates/dali.yaml
@@ -14,7 +14,8 @@ spec:
       labels:
         run: {{ .name | quote }}
     spec:
-      {{- include "hpcc.checkDaliMount" (dict "root" $) | indent 6 }}
+      initContainers: 
+        {{- include "hpcc.checkDaliMount" (dict "root" $) | indent 6 }}
       containers:
       - name: {{ .name | quote }}
         workingDir: /var/lib/HPCCSystems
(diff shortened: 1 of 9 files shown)


richardkchapman:
@ghalliday  For comment/review

My main concern is that it involves a trip to github at eclccserver pod startup, which is not ideal for production systems. We could consider it as being useful for dev systems though - or think of another way.

Even if we don't do bundles this way, I think the changes to helm structure are worth including as I think there may be other situations where we will want an init container to seed data into the main one - roxie queries, for example.

helm/hpcc/templates/dali.yaml
```diff
@@ -14,7 +14,8 @@ spec:
       labels:
         run: {{ .name | quote }}
     spec:
-      {{- include "hpcc.checkDaliMount" (dict "root" $) | indent 6 }}
+      initContainers: 
+        {{- include "hpcc.checkDaliMount" (dict "root" $) | indent 6 }}
```
ghalliday, line 18: Should possibly now use indent 8, and reduce the indentation in the function.

ghalliday: ignore that the list syntax requires indent 6 still

PR #13860, opened 2020-06-15
HPCC-23282 Correct reading of a UTF-16 file

AttilaVamos:
This is an experimental feature. 

Add "bool keepSourceEncoding" parameter to SprayVariable(), SprayDelimited(), SprayXml() and SprayJson() functions to keep original source encoding or enable target file to be  UTF8 encoded. Defaults to TRUE.

Add same to DFUPlus spray

Handle encoding and improve error logging.

Add regression test case to test keeping original encoding and converting to UTF-8.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

## Testing:

Tested manually and added a new regression test case


AttilaVamos:
@jakesmith please review.

dali/dfuplus/dfuplus.cpp
```diff
@@ -525,7 +532,14 @@ bool CDfuPlusHelper::variableSpray(const char* srcxml,const char* srcip,const ch
     else
         encoding = format; // may need extra later
 
-    req->setSourceFormat(CDFUfileformat::decode(encoding));
+    DFUfileformat formatCode = CDFUfileformat::decode(encoding);
+    if (formatCode != DFUff_unknown)
+        req->setSourceFormat(formatCode);
+    else
+        throw MakeStringException(0, "Unknown source file format: '%s'", encoding);
```
jakesmith, line 539: trivial: makeStringExceptionV prefered

dali/ft/filecopy.cpp
```diff
@@ -1326,22 +1326,23 @@ void FileSprayer::checkFormats()
         {
         case FFTfixed:
             if ((tgtType != FFTvariable)&&(tgtType != FFTvariablebigendian))
-                throwError(DFTERR_BadSrcTgtCombination);
+                throwError2(DFTERR_BadSrcTgtCombination, FileFormatTypeStr[srcType], FileFormatTypeStr[tgtType]);
```
jakesmith, line 1329: not sure if possible, but could check srcType and tgtType are in range before switch statement, to prevent crash here if not. i.e. check that they are less than < FileFormatType::FFTlast

dali/ft/filecopy.cpp
```diff
@@ -1326,22 +1326,23 @@ void FileSprayer::checkFormats()
...
             break;
         case FFTblocked:
             if ((tgtType != FFTvariable)&&(tgtType != FFTvariablebigendian))
-                throwError(DFTERR_BadSrcTgtCombination);
+                throwError2(DFTERR_BadSrcTgtCombination, FileFormatTypeStr[srcType], FileFormatTypeStr[tgtType]);
             break;
         case FFTcsv:
-            throwError(DFTERR_BadSrcTgtCombination);
+                throwError2(DFTERR_BadSrcTgtCombination, FileFormatTypeStr[srcType], FileFormatTypeStr[tgtType]);
```
jakesmith, line 1344: formatting: indentation

dali/ft/filecopy.cpp
```diff
@@ -2685,6 +2686,9 @@ void FileSprayer::setTarget(IDistributedFile * target)
             srcFormat.separate.set("\\,");
 
         tgtFormat.set(srcFormat);
+        if ((operation == dfu_import) && ( !options->getPropBool("@keepSourceEncoding")))
```
jakesmith, line 2689: trivial: extra space between ( and !

jakesmith: >( !options->getPropBool("@keepSourceEncoding"))

should this be :  (!options->getPropBool("@keepSourceEncoding", true)) , as in other places?

ecllibrary/std/File.ecl
```diff
@@ -472,19 +474,20 @@ EXPORT SprayVariable(varstring sourceIP, varstring sourcePath, integer4 sourceMa
  * @param expireDays    Number of days to auto-remove file. Default is -1, not expire.
  * @param dfuServerQueue Name of target DFU Server queue. Default is '' (empty) for the first DFU queue in the environment.
  * @param noSplit      Don't split a file part to multiple target parts. Default is FALSE.
+ * @param keepSourceEncoding Prevent target file UTF8 encoding. Defaults to TRUE.
```
jakesmith, line 477: >Prevent target file UTF8 encoding

I think better wording would be:

"If disabled, variable imports will default to UTF8. Defaults to TRUE."

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -2189,6 +2189,12 @@ bool CFileSprayEx::onSprayVariable(IEspContext &context, IEspSprayVariable &req,
         if (!req.getExpireDays_isNull())
             options->setExpireDays(req.getExpireDays());
 
+        if (req.getKeepSourceEncoding_isNull())
+            // Set default value
+            options->setKeepSourceEncoding(true);
```
jakesmith, line 2194: Should remove this afaics. It shouldn't be necessary to set this
It will not set for older eclwatch clients who create DFU workunits (that don't have this code),
but that's fine, because in DFU where this option is read, it defaults to true if not set.

, it should assume (default) keepSourceFormatting, so

plugins/fileservices/fileservices.cpp
```diff
@@ -1031,6 +1055,9 @@ FILESERVICES_API char * FILESERVICES_CALL implementSprayXml(ICodeContext *ctx, c
     if (noSplit)
         req->setNosplit(true);
 
+    if (keepSourceEncoding)
+        req->setKeepSourceEncoding(true);
```
jakesmith, line 1059: I think you need an else statement here (as with code in implementSprayVariable) to call setKeepSourceEncoding(false), because otherwise servers will assume keep source formatting.

plugins/fileservices/fileservices.cpp
```diff
@@ -1141,6 +1179,9 @@ FILESERVICES_API char * FILESERVICES_CALL implementSprayJson(ICodeContext *ctx,
...
```
jakesmith, line 1183: as per last comment (and code in implementSprayVariable) think you need else statement here.

jakesmith (changes requested):

dali/dfu/dfuwu.cpp
```diff
@@ -2095,6 +2095,16 @@ class CDFUoptions: public CLinkedDFUWUchild, implements IDFUoptions
     {
         queryRoot()->setPropBool("@noCommon",val);
     }
+
+    bool getKeepSourceEncoding() const
+	{
+		return queryRoot()->getPropBool("@keepSourceEncoding", true);
+	}
```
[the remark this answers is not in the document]
ghalliday: still an issue

ghalliday:
no comments beyond jake's

dali/dfu/dfuwu.cpp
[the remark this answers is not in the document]
AttilaVamos: Changed

dali/dfuplus/dfuplus.cpp
AttilaVamos: Changed

dali/ft/filecopy.cpp
AttilaVamos: Updated

dali/ft/filecopy.cpp
AttilaVamos: Changed

ecllibrary/std/File.ecl
AttilaVamos: Changed

plugins/fileservices/fileservices.cpp
AttilaVamos: Else added.

plugins/fileservices/fileservices.cpp
AttilaVamos: Else added

richardkchapman:
@jakesmith Is this ready to merge?

esp/services/ws_fs/ws_fsService.cpp
jakesmith: @AttilaVamos - discussed offline, but I not sure this is right.

IDFUWorkunit::getKeepSourceEncoding() return true by default (if DFU workunit option is not set).
That makes sense - that is the default wanted, for new and legacy workunits.

If the request is isNull(), then this code should not set the DFU workunit option.. i.e. it will be the same as legacy workunits, without the setting set, and default to true via IDFUWorkunit::getKeepSourceEncoding()

I suspect the code here should look like:
```
if (!req.getKeepSourceEncoding_isNull() && !req.getKeepSourceEncoding())
    options->setKeepSourceEncoding(false);
```

Meaning if request is non null and it is false, set it into the DFU workunit option.
Otherwise leaving it unset, means DFUWorknit::getKeepSourceEncoding() will return true by default as desired.

dali/dfu/dfuwu.cpp
jakesmith, line 2102: @AttilaVamos - there still appears to be something wrong with the formatting/indentation.

jakesmith:
@AttilaVamos - I think you have some unpushed changed (e.g. formatting fixes) ?
And then probably just the issue re. unwanted call to IDFUWorkUnit::setKeepSourceEncoding() 

The PR should also be targeted at master.

AttilaVamos:
I close this PR because there is a re-targeted and updated vesrion: https://github.com/hpcc-systems/HPCC-Platform/pull/14123.

PR #13874, opened 2020-06-17
HPCC-24263 Issues with HTTP keepalive in Roxie

richardkchapman:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

## Testing:
using curl - it would lock up after a "503 Too many active queries" result prior to this fix.

### DIFF
--- a/common/thorhelper/roxiehelper.hpp
+++ b/common/thorhelper/roxiehelper.hpp
@@ -351,8 +351,8 @@ interface SafeSocket : extends IInterface
 {
     virtual ISocket *querySocket()  = 0;
     virtual size32_t write(const void *buf, size32_t size, bool takeOwnership=false) = 0;
-    virtual bool readBlock(MemoryBuffer &ret, unsigned maxBlockSize, unsigned timeout = (unsigned) WAIT_FOREVER) = 0;
-    virtual bool readBlock(StringBuffer &ret, unsigned timeout, HttpHelper *pHttpHelper, bool &, bool &, unsigned maxBlockSize) = 0;
+    virtual bool readBlocktms(MemoryBuffer &ret, unsigned timeoutms, unsigned maxBlockSize) = 0;
+    virtual bool readBlocktms(StringBuffer &ret, unsigned timeoutms, HttpHelper *pHttpHelper, bool &, bool &, unsigned maxBlockSize) = 0;
     virtual void checkSendHttpException(HttpHelper &httphelper, IException *E, const char *queryName) = 0;
     virtual void sendSoapException(IException *E, const char *queryName) = 0;
     virtual void sendJsonException(IException *E, const char *queryName) = 0;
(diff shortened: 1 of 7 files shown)


richardkchapman:
@afishbeck Please review

roxie/ccd/ccdprotocol.cpp
```diff
@@ -57,7 +57,7 @@ class CHpccProtocolPlugin : implements IHpccProtocolPlugin, public CInterface
         trapTooManyActiveQueries = ctx.ctxGetPropBool("@trapTooManyActiveQueries", true);
         numRequestArrayThreads = ctx.ctxGetPropInt("@requestArrayThreads", 5);
         maxHttpConnectionRequests = ctx.ctxGetPropInt("@maxHttpConnectionRequests", 10);
-        maxHttpKeepAliveWait = ctx.ctxGetPropInt("@maxHttpKeepAliveWait", 5000);
+        maxHttpKeepAliveWait = ctx.ctxGetPropInt("@maxHttpKeepAliveWait", 5000); // MORE - is this supposed to be 5000 seconds? Because that's what it seems to be interpreted as
```
afishbeck, line 60: Right, so this is the amount of time we will wait to see if another request arrives on the same connection.  It can certainly be made shorter than 83 1/3 minutes.

afishbeck (approved):
Change looks good.  Comment correctly points out that the maxHttpKeepAliveWait can be changed to a more reasonable value.

roxie/ccd/ccdprotocol.cpp
richardkchapman: What would be a reasonable value? 5 seconds?

afishbeck: Yes that would be good.  Probably the original intent of the 5000.

common/thorhelper/roxiehelper.hpp
```diff
@@ -399,8 +399,8 @@ class THORHELPER_API CSafeSocket : implements SafeSocket, public CInterface
     ISocket *querySocket() { return sock; }
     virtual CriticalSection &queryCrit() { return crit; };
     size32_t write(const void *buf, size32_t size, bool takeOwnership=false);
-    bool readBlock(MemoryBuffer &ret, unsigned maxBlockSize, unsigned timeout = (unsigned) WAIT_FOREVER);
```
richardkchapman, line 402: Note that parameter order in implementation did not match!

richardkchapman:

roxie/ccd/ccdcontext.cpp
```diff
@@ -1054,7 +1054,7 @@ class StreamedRawDataReader : public RawDataReader
 
             // Note: I am the only thread reading (we only support a single input dataset in roxiepipe mode)
             MemoryBuffer reply;
-            client.readBlock(reply, readTimeout);
+            client.readBlocktms(reply, readTimeout, maxBlockSize);
```
ghalliday, line 1057: This is now a time out of 300ms instead of 5 minutes.  Is that correct?   I wonder if it is meant to be long for this call.

ghalliday:
@richardkchapman one question.  Commits should also be squashed I think.

roxie/ccd/ccdcontext.cpp
richardkchapman: It was actually infinite before, I think, though that was not intentional.

300ms may be a bit short though.

richardkchapman: Actually the parameter order bug confused me I think.

Fixed.

richardkchapman:
Fixed and squashed

PR #13882, opened 2020-06-19
HPCC-24279 Allow platform code to detect whether it is called from embedded C++

richardkchapman:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/ecl/hqlcpp/hqlcpp.cpp
+++ b/ecl/hqlcpp/hqlcpp.cpp
@@ -12016,6 +12016,10 @@ void HqlCppTranslator::buildCppFunctionDefinition(BuildCtx & ctx, IHqlExpression
     }
 
     MemberFunction userFunc(*this, ctx, proto, MFdynamicproto);
+    userFunc.ctx.addQuoted("#undef EmbedCodeBlock");
+    VStringBuffer eblock("EmbedCodeBlock __b%d;", rand());
+    userFunc.ctx.addQuoted(eblock.str());
+
     if (location)
         userFunc.ctx.addLine(locationFilename, startLine);
     userFunc.ctx.addQuoted(body);
--- a/system/jlib/jscm.hpp
+++ b/system/jlib/jscm.hpp
@@ -211,4 +211,10 @@ interface jlib_thrown_decl IException : public IInterface
     virtual MessageAudience errorAudience() const = 0;
 };
 
+class jlib_decl EmbedCodeBlock
+{
+public:
+    EmbedCodeBlock();
+    ~EmbedCodeBlock();
+};
 #endif
--- a/system/jlib/jthread.cpp
+++ b/system/jlib/jthread.cpp
@@ -37,6 +37,21 @@
 
 #define LINUX_STACKSIZE_CAP (0x200000)
 
+static thread_local unsigned _inEmbed = 1;
+extern jlib_decl bool inEmbed()
+{
+    return _inEmbed!=0;
+}
+
+EmbedCodeBlock::EmbedCodeBlock()
+{
+    _inEmbed++;
+}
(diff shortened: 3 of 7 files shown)


ecl/hqlcpp/hqlcpp.cpp
```diff
@@ -12016,6 +12016,10 @@ void HqlCppTranslator::buildCppFunctionDefinition(BuildCtx & ctx, IHqlExpression
     }
 
     MemberFunction userFunc(*this, ctx, proto, MFdynamicproto);
+    userFunc.ctx.addQuoted("#undef EmbedCodeBlock");
+    VStringBuffer eblock("EmbedCodeBlock __b%d;", rand());
```
ghalliday, line 12020: You can't use rand() - you need to use some consistent numbering scheme.

system/jlib/jscm.hpp
```diff
@@ -211,4 +211,10 @@ interface jlib_thrown_decl IException : public IInterface
     virtual MessageAudience errorAudience() const = 0;
 };
 
+class jlib_decl EmbedCodeBlock
```
ghalliday, line 214: Definitely hesitant about a dependency of generated code on a class defined in jlib.  I think it is probably ok, but it is the first instance.

system/jlib/jthread.cpp
```diff
@@ -37,6 +37,21 @@
 
 #define LINUX_STACKSIZE_CAP (0x200000)
 
+static thread_local unsigned _inEmbed = 1;
```
ghalliday, line 40: embedDepth would probably be a better name

ecl/hqlcpp/hqlcpp.cpp
ghalliday: I suspect you have done it so that the generated code cannot know the name, but it would make it impossible to check the regression suite output.

ghalliday (commented):
@richardkchapman the main problem I can think of is when you have a streaming input.  When the embedded code calls
`input->nextRow(),
`
nextRow() will be assumed to be within embedded code, unless that adjusts the count.

It is still possible to call the functionality - even if it involves some illegal c++:

```
x := BEGINC++
EmbedCodeBlock local;
local.~EmbedCodeBlock();
local.~EmbedCodeBlock();
... I'm unsafe
new (&local) EmbedCodeBlock();
new (&local) EmbedCodeBlock();

```
or something similar.

system/jlib/jscm.hpp
richardkchapman: It is the first instance (other than people using SERVICE, LIBRARY(jlib)...) but it doesn't really create any new dependency (in that eclrtl is dependent on jlib).

ecl/hqlcpp/hqlcpp.cpp
richardkchapman: That did occur to me, but I'm sure there's a way of making it consistent when running regression suite without the name being known. I didn't bother fixing that issue for this POC though. As you surmise, it's to make it harder for embedded code to explicitly call destructor.

richardkchapman:
Assuming that (somehow) I fix the nextrow issue, I'm not sure there is any situation where embedded code ends up calling embedded code - in which case the explicit destructor loophole can be detected by detecting and aborting any time count reaches 2.

richardkchapman:
Issues are caused by code using standard threads.

Not sure how/if this could be fixed, or whether should abandon the idea. Closing for now.

PR #13896, opened 2020-06-23
HPCC-24287 Analyzer: Report excess rejected rows from index join

shamser:
## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:


shamser:

common/wuanalysis/anarule.cpp
```diff
@@ -165,11 +165,40 @@ class IoSkewRule : public AActivityRule
...
+                stat_type cost = timeAvgLocalExecute * preFilteredPer / 100;
+                result.set(ANA_KEYED_JOIN_EXCESSREJECTS_ID, cost, "Large number of rows from left dataset rejected in keyed join");
+                updateInformation(result, activity);
+                return true;
+            }
+        }
+        return false;
+    }
+private:
+    static const stat_type preFilteredThreshold = 20; //percent
```
ghalliday, line 194: Is this a sensible default?  I think this means if >20% rows are rejected then the warning is reported.  For keyed join that may be fine, for index read it would too high.

This should probably be a member of WuAnalyseOptions.

common/wuanalysis/anarule.cpp
```diff
@@ -165,11 +165,40 @@ class IoSkewRule : public AActivityRule
...
+        stat_type preFiltered = activity.getStatRaw(StNumPreFiltered);
+        if (preFiltered)
+        {
+            IWuEdge * inputEdge = activity.queryInput(0);
+            stat_type rowscnt = inputEdge->getStatRaw(StNumRowsProcessed);
+            stat_type preFilteredPer = (preFiltered*100/ rowscnt); // prefiltered %
+            if (  preFilteredPer > preFilteredThreshold)
+            {
+                IWuActivity * inputActivity = inputEdge->querySource();
+                stat_type timeAvgLocalExecute = inputActivity->getStatRaw(StTimeLocalExecute, StAvgX);
```
ghalliday, line 184: why the time of the input activity?  That may be true some of the time, but hard to tell.  A comment should explain why.

ghalliday, line 184: This rule would potentially work on hthor/roxie as well, but they do not have an average for execute time.

ghalliday (commented):
@shamser looks good, a few minor comments

shamser:
@ghalliday

ghalliday:
@shamser please squash

common/wuanalysis/anarule.cpp
```diff
@@ -165,11 +165,39 @@ class IoSkewRule : public AActivityRule
...
+    KeyedJoinExcessRejectedRowsRule() : ActivityKindRule(TAKkeyedjoin) {}
+
+    virtual bool check(PerformanceIssue & result, IWuActivity & activity, const WuAnalyseOptions & options) override
+    {
+        stat_type preFiltered = activity.getStatRaw(StNumPreFiltered);
+        if (preFiltered)
+        {
+            IWuEdge * inputEdge = activity.queryInput(0);
+            stat_type rowscnt = inputEdge->getStatRaw(StNumRowsProcessed);
+            stat_type preFilteredPer = statPercent( preFiltered*100 ) / rowscnt;
```
shamser, line 180: @ghalliday Any possibility for inputEdge->getStatRaw(StNumRowsProcessed to return 0 here?

shamser:
@ghalliday Squashed.  Please can you review.

richardkchapman:
This should probably be targeting 7.10.x (unless it's vital to go in between rc2 and gold, which should really only mena regressions...)

shamser:
@ghalliday

common/wuanalysis/anarule.cpp
ghalliday: @shamser I would say very unlikely, but probably best to protect against division by zero just in case.  (E.g. a bug in the engine meaning that a row count wasn't added to an edge.)

ghalliday:
@shamser very close to ready.

common/wuanalysis/anarule.cpp
shamser: I'll be updating the rules to work with hThor and roxie in a separate Jira.

shamser:
@ghalliday

common/wuanalysis/anarule.cpp
```diff
@@ -165,11 +165,42 @@ class IoSkewRule : public AActivityRule
...
+    virtual bool check(PerformanceIssue & result, IWuActivity & activity, const WuAnalyseOptions & options) override
+    {
+        stat_type preFiltered = activity.getStatRaw(StNumPreFiltered);
+        if (preFiltered)
+        {
+            IWuEdge * inputEdge = activity.queryInput(0);
+            stat_type rowscnt = inputEdge->getStatRaw(StNumRowsProcessed);
+            if (rowscnt)
+            {
+                stat_type preFilteredPer = statPercent( (long double) preFiltered * 100.0 / rowscnt );
```
ghalliday, line 182: why long double?  double is fine.

ghalliday: hmm I see the function takes a long double.  That is unusual.

shamser: @ghalliday Shall I change it to double?

shamser:
@ghalliday

common/wuanalysis/anarule.cpp
ghalliday: @shamser I debated this for a while.  I was inclined to merge as-is, but since it is an overloaded function I think it would be better to change it to double (otherwise everything that wants to call it needs an unusual cast).  As far as I can see it was not previously called.

shamser:
@ghalliday

PR #13900, opened 2020-06-24
HPCC-24312 Security manager and related interface updates

timothyklemm:
- Change ISecManager

Add supported feature interrogation mechanism. The platform can make
decisions about what the current security manager supports based on the
manager's claims instead of relying on RTTI or build settings.

Add new interface method to retrieve user data for unauthenticated users.
Username is the only required input. The expected output includes user
properties normally set after user authentication but in no way suggests
successful authentication.

Add optional IEspSecureContext to all methods not already receiving it.

- Change IAuthMap

Add addToBackend, allowing it to push resources to any manager, not
just the manager specified during construction.

- Change ISecUser

Add supported feature interrogation mechanism.

Add interface methods to access new user data as a PTree node. This is
new information provided by certain managers; it is not meant to be a
new form of data already available in the user interface.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/bindings/http/platform/httpservice.cpp
+++ b/esp/bindings/http/platform/httpservice.cpp
@@ -2019,8 +2019,8 @@ void CEspHttpServer::logoutSession(EspAuthRequest& authReq, unsigned sessionID, 
                 if (user)
                 {
                     //inform security manager that user is logged out
-                    Owned<ISecUser> secUser = secmgr->createUser(user);
-                    secmgr->logoutUser(*secUser);
+                    Owned<ISecUser> secUser = secmgr->createUser(user, ctx->querySecureContext());
+                    secmgr->logoutUser(*secUser, ctx->querySecureContext());
                 }
             }
 
--- a/system/security/LdapSecurity/ldapsecurity.cpp
+++ b/system/security/LdapSecurity/ldapsecurity.cpp
@@ -66,6 +66,21 @@ MemoryBuffer& CLdapSecUser::getUserSid()
     return m_usersid;
 }
 //interface ISecUser
+SecFeatureSet CLdapSecUser::queryFeatures(SecFeatureSupportLevel level) const
+{
+    switch (level)
+    {
+    case SFSL_Safe:
+        return s_safeFeatures;
+    case SFSL_Implemented:
+        return s_implementedFeatures;
+    case SFSL_Unsafe:
+        return SUF_ALL_FEATURES & ~s_safeFeatures;
+    default:
+        return SUF_NO_FEATURES;
+    }
+}
+
 const char * CLdapSecUser::getName()
 {
     return m_name.get();
@@ -631,23 +646,38 @@ CLdapSecManager::~CLdapSecManager()
 }
 
 //interface ISecManager : extends IInterface
-ISecUser * CLdapSecManager::createUser(const char * user_name)
+SecFeatureSet CLdapSecManager::queryFeatures(SecFeatureSupportLevel level) const
+{
+    switch (level)
+    {
+    case SFSL_Implemented:
+        return s_implementedFeatures;
+    case SFSL_Safe:
+        return s_safeFeatures;
+    case SFSL_Unsafe:
+        return SMF_ALL_FEATURES & ~s_safeFeatures;
+    default:
+        return SMF_NO_FEATURES;
+    }
+}
+
+ISecUser * CLdapSecManager::createUser(const char * user_name, IEspSecureContext* secureContext)
 {
     return (new CLdapSecUser(user_name, NULL));
 }
 
-ISecResourceList * CLdapSecManager::createResourceList(const char * rlname)
+ISecResourceList * CLdapSecManager::createResourceList(const char * rlname, IEspSecureContext* secureContext)
 {
     return (new CLdapSecResourceList(rlname));
 }
 
-bool CLdapSecManager::subscribe(ISecAuthenticEvents & events)
+bool CLdapSecManager::subscribe(ISecAuthenticEvents & events, IEspSecureContext* secureContext)
 {
     m_subscriber.set(&events);
     return true;
 }
 
-bool CLdapSecManager::unsubscribe(ISecAuthenticEvents & events)
+bool CLdapSecManager::unsubscribe(ISecAuthenticEvents & events, IEspSecureContext* secureContext)
 {
     if (&events == m_subscriber.get())
     {
@@ -824,7 +854,7 @@ SecAccessFlags CLdapSecManager::authorizeEx(SecResourceType rtype, ISecUser & us
         return SecAccess_Full;
 
     Owned<ISecResourceList> rlist;
-    rlist.setown(createResourceList("resources"));
+    rlist.setown(createResourceList("resources", secureContext));
     rlist->addResource(resourcename);
     
     bool ok = authorizeEx(rtype, user, rlist.get(), secureContext);
@@ -909,13 +939,13 @@ SecAccessFlags CLdapSecManager::authorizeEx(SecResourceType rtype, ISecUser & us
         return SecAccess_Unavailable;
 }
 
-SecAccessFlags CLdapSecManager::getAccessFlagsEx(SecResourceType rtype, ISecUser & user, const char * resourcename)
+SecAccessFlags CLdapSecManager::getAccessFlagsEx(SecResourceType rtype, ISecUser & user, const char * resourcename, IEspSecureContext* secureContext)
 {
     if(!resourcename || !*resourcename)
         return SecAccess_Unavailable;
 
     Owned<ISecResourceList> rlist0;
-    rlist0.setown(createResourceList("resources"));
+    rlist0.setown(createResourceList("resources", secureContext));
     rlist0->addResource(resourcename);
     
     CLdapSecResourceList * reslist = (CLdapSecResourceList*)rlist0.get();
@@ -981,7 +1011,7 @@ bool CLdapSecManager::authorize(ISecUser& sec_user, ISecResourceList * Resources
 }
 
 
-SecAccessFlags CLdapSecManager::authorizeFileScope(ISecUser & user, const char * filescope)
+SecAccessFlags CLdapSecManager::authorizeFileScope(ISecUser & user, const char * filescope, IEspSecureContext* secureContext)
 {
     if(filescope == 0 || filescope[0] == '\0')
         return SecAccess_Full;
@@ -1000,19 +1030,19 @@ SecAccessFlags CLdapSecManager::authorizeFileScope(ISecUser & user, const char *
     }
 
     Owned<ISecResourceList> rlist;
-    rlist.setown(createResourceList("FileScope"));
+    rlist.setown(createResourceList("FileScope", secureContext));
     rlist->addResource(managedFilescope.length() ? managedFilescope.str() : filescope );
     
-    bool ok = authorizeFileScope(user, rlist.get());
+    bool ok = authorizeFileScope(user, rlist.get(), secureContext);
     if(ok)
         return rlist->queryResource(0)->getAccessFlags();
     else
         return SecAccess_Unavailable;
 }
 
-bool CLdapSecManager::authorizeFileScope(ISecUser & user, ISecResourceList * resources)
+bool CLdapSecManager::authorizeFileScope(ISecUser & user, ISecResourceList * resources, IEspSecureContext* secureContext)
 {
-    return authorizeEx(RT_FILE_SCOPE, user, resources);
+    return authorizeEx(RT_FILE_SCOPE, user, resources, secureContext);
 }
 
 bool CLdapSecManager::authorizeViewScope(ISecUser & user, StringArray & filenames, StringArray & columnnames)
@@ -1061,29 +1091,29 @@ bool CLdapSecManager::authorizeViewScope(ISecUser & user, StringArray & filename
     return true;
 }
 
-SecAccessFlags CLdapSecManager::authorizeWorkunitScope(ISecUser & user, const char * wuscope)
+SecAccessFlags CLdapSecManager::authorizeWorkunitScope(ISecUser & user, const char * wuscope, IEspSecureContext* secureContext)
 {
     if(wuscope == 0 || wuscope[0] == '\0')
         return SecAccess_Full;
 
     Owned<ISecResourceList> rlist;
-    rlist.setown(createResourceList("WorkunitScope"));
+    rlist.setown(createResourceList("WorkunitScope", secureContext));
     rlist->addResource(wuscope);
     
-    bool ok = authorizeWorkunitScope(user, rlist.get());
+    bool ok = authorizeWorkunitScope(user, rlist.get(), secureContext);
     if(ok)
         return rlist->queryResource(0)->getAccessFlags();
     else
         return SecAccess_Unavailable;
 }
     
-bool CLdapSecManager::authorizeWorkunitScope(ISecUser & user, ISecResourceList * resources)
+bool CLdapSecManager::authorizeWorkunitScope(ISecUser & user, ISecResourceList * resources, IEspSecureContext* secureContext)
 {
-    return authorizeEx(RT_WORKUNIT_SCOPE, user, resources);
+    return authorizeEx(RT_WORKUNIT_SCOPE, user, resources, secureContext);
 }
 
 
-bool CLdapSecManager::addResourcesEx(SecResourceType rtype, ISecUser& sec_user, ISecResourceList * resources, SecPermissionType ptype, const char* basedn)
+bool CLdapSecManager::addResourcesEx(SecResourceType rtype, ISecUser& sec_user, ISecResourceList * resources, SecPermissionType ptype, const char* basedn, IEspSecureContext* secureContext)
 {
     CLdapSecResourceList * reslist = (CLdapSecResourceList*)resources;
     if(!reslist)
@@ -1095,22 +1125,22 @@ bool CLdapSecManager::addResourcesEx(SecResourceType rtype, ISecUser& sec_user, 
     return m_ldap_client->addResources(rtype, sec_user, rlist, ptype, basedn);
 }
 
-bool CLdapSecManager::addResourceEx(SecResourceType rtype, ISecUser& user, const char* resourcename, SecPermissionType ptype, const char* basedn)
+bool CLdapSecManager::addResourceEx(SecResourceType rtype, ISecUser& user, const char* resourcename, SecPermissionType ptype, const char* basedn, IEspSecureContext* secureContext)
 {
     Owned<ISecResourceList> rlist;
-    rlist.setown(createResourceList("resources"));
+    rlist.setown(createResourceList("resources", secureContext));
     rlist->addResource(resourcename);
     
-    return addResourcesEx(rtype, user, rlist.get(), ptype, basedn);
+    return addResourcesEx(rtype, user, rlist.get(), ptype, basedn, secureContext);
 }
 
 
-bool CLdapSecManager::addResources(ISecUser& sec_user, ISecResourceList * resources)
+bool CLdapSecManager::addResources(ISecUser& sec_user, ISecResourceList * resources, IEspSecureContext* secureContext)
 {
-    return addResourcesEx(RT_DEFAULT, sec_user, resources);
+    return addResourcesEx(RT_DEFAULT, sec_user, resources, PT_DEFAULT, nullptr, secureContext);
 }
 
-bool CLdapSecManager::addUser(ISecUser & user)
+bool CLdapSecManager::addUser(ISecUser & user, IEspSecureContext* secureContext)
 {
     bool ok = m_ldap_client->addUser(user);
     if(!ok)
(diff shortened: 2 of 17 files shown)


timothyklemm:
@afishbeck, @RussWhitehead please review.

system/security/shared/seclib.hpp
```diff
@@ -33,6 +34,17 @@
 #define SECLIB "seclib"
 #define LDAPSECLIB "LdapSecurity"
 
+template <typename index_t>
+constexpr uint64_t makeBitMask(index_t bitIndex)
```
afishbeck, line 38: Can we name these something less likely to conflict like makeSecFeatureMask?

system/security/LdapSecurity/ldapsecurity.cpp
```diff
@@ -27,6 +27,8 @@ using namespace cryptohelper;
 
 #include "workunit.hpp"
 
+const uint64_t CLdapSecManager::s_implementedFeatures = (makeBitMask(SMF_NUMBER_OF_INDICES) - 1) & ~makeBitMask(SMF_RetrieveUserData);
+
```
afishbeck, line 31: Some of these use cases for the mask routines aren't very intuitive for secmanger developers.  When I think what you really seem to be doing is makeSecFeatureMaskAllExcept(SMF_NUMBER_OF_INDICES, SMF_RetrieveUserData)?  Maybe add an explict function that makes this more intuitive?

system/security/LdapSecurity/ldapsecurity.cpp
```diff
@@ -66,6 +68,19 @@ MemoryBuffer& CLdapSecUser::getUserSid()
     return m_usersid;
 }
 //interface ISecUser
+uint64_t CLdapSecUser::queryFeatures(SecurityFeatureSupportLevel level) const
+{
+    switch (level)
+    {
+    case SFSL_Safe:
+        return (s_featureMask | makeBitMask(SUF_GetDataElement, SUF_GetDataElements));
```
afishbeck, line 76: Should this be another static variable?

system/security/LdapSecurity/ldapsecurity.cpp
```diff
@@ -631,23 +646,37 @@ CLdapSecManager::~CLdapSecManager()
...
 //interface ISecManager : extends IInterface
-ISecUser * CLdapSecManager::createUser(const char * user_name)
+uint64_t CLdapSecManager::queryFeatures(SecurityFeatureSupportLevel level, IEspSecureContext* secureContext) const
+{
+    switch (level)
+    {
+    case SFSL_Implemented:
+        return s_implementedFeatures;
+    case SFSL_Safe:
+        return (s_implementedFeatures | makeBitMask(SMF_RetrieveUserData));
```
afishbeck, line 656: Another static?

system/security/shared/seclib.hpp
```diff
@@ -325,6 +396,7 @@ interface IAuthMap : extends IInterface
     virtual bool shouldAuth(const char * path) = 0;
     virtual ISecResourceList * queryResourceList(const char * path) = 0;
     virtual ISecResourceList * getResourceList(const char * path) = 0;
+    virtual bool addToBackend(interface ISecManager* backend, interface IEspSecureContext* secureContext = nullptr) = 0;
```
afishbeck, line 399: The fact that the secmgr being passed in is considered the backend is kind of beside the point.  Can we just call this method something like copyResourceList() or something similar?

afishbeck (changes requested):
@timothyklemm some comments.

system/security/shared/seclib.hpp
timothyklemm: They can be renamed, but they are generic functions that just happened to be introduced for use with security features. Would it be better to rename them for this use case or move them to something where a generic name is more appropriate?

system/security/shared/seclib.hpp
timothyklemm: The name can be changed. I was originally trying to limit changes related to this - existing usage of the auth map didn't change. If it's going to change, does it make sense to go "all the way" with the change? The constructor accepts a manager instance, and holds it solely for use by this method. In all existing cases, the map construction and the call of this method occur in the same method. I see no reason for the map to accept (or retain) a reference to the manager except when calling this method.

system/security/shared/seclib.hpp
afishbeck: To be honest I don't want to promote this style of bitmask preparation.  It not as efficient as having the bits pre specified.  It's OK if the bitmasks are established once globally and cached or stored in static variables.  I would usually prefer having the bits set in the enumeration and used directly.

```
enum SecManagerFeatureIndex : unsigned
{
    SMF_CreateUser = 0x000001,
    SMF_CreateResourceList = 0x000002,
    SMF_Subscribe = 0x000004,

```
etc.

That would be more efficient for both the plugin and the user.  Functions for setting AllExcept, etc could still be used.

system/security/shared/seclib.hpp
```diff
@@ -198,10 +217,60 @@ enum authStatus : int
...
+    SUF_NUMBER_OF_INDICES
+};
+
 class CDateTime;
 interface IPropertyIterator;
 interface ISecUser : extends IInterface
 {
+    virtual uint64_t queryFeatures(SecurityFeatureSupportLevel level) const = 0;
+    inline bool checkFeatures(SecurityFeatureSupportLevel level, uint64_t mask) const { return ((queryFeatures(level) & mask) == mask); }
+    inline bool checkUnsafeFeatures(uint64_t mask) const { return checkFeatures(SFSL_Unsafe, mask); }
```
RussWhitehead, line 270: Can we improve the variable name "mask" to some thing like "SUFmask" or something more meaningful to a plugin developer looking at this for the first time?

system/security/shared/seclib.hpp
```diff
@@ -19,6 +19,7 @@
 #define _SECLIB_HPP__
 
 #include "jlib.hpp"
+#include "jptree.hpp"
```
RussWhitehead, line 22: Minor. Why did we need to add this? IPropertyTree was already part of this file so I presume it is already included

system/security/shared/seclib.hpp
```diff
@@ -337,53 +409,110 @@ enum secManagerType : int
...
-    virtual unsigned getPasswordExpirationWarningDays() = 0;
-    virtual bool createUserScopes() = 0;
-    virtual aindex_t getManagedScopeTree(SecResourceType rtype, const char * basedn, IArrayOf<ISecResource>& scopes) = 0;
-    virtual SecAccessFlags queryDefaultPermission(ISecUser& user) = 0;
-    virtual bool clearPermissionsCache(ISecUser & user) = 0;
-    virtual bool authenticateUser(ISecUser & user, bool * superUser) = 0;
-    virtual secManagerType querySecMgrType() = 0;
-    virtual const char* querySecMgrTypeName() = 0;
-    virtual bool logoutUser(ISecUser & user) = 0;
+    virtual uint64_t queryFeatures(SecurityFeatureSupportLevel level, IEspSecureContext* secureContext = nullptr) const = 0;
```
RussWhitehead, line 465: Minor, but could you add comment that this returns a mask of SUF features?  Developers should not have to look at the implementation to figure that out

system/security/shared/seclib.hpp
```diff
@@ -198,10 +217,60 @@ enum authStatus : int
...
+    SUF_GetProperty,
+    SUF_SetProperty,
+    SUF_GetPropertyInt,
+    SUF_SetPropertyInt,
+    SUF_GetPropertyIterator,
+    SUF_Clone,
+    SUF_GetDataElement,
+    SUF_GetDataElements,
+
+    SUF_NUMBER_OF_INDICES
```
RussWhitehead, line 261: How safe is this for an externally developed security manager plugin` that is not built with the platform?  How would you want the plugin to behave if the platform later adds new SUF features and does not rebuild that plugin?

system/security/plugins/singleuserSecurity/singleUserSecurity.cpp
```diff
@@ -44,16 +44,34 @@ class CSingleUserSecurityManager : public CBaseSecurityManager
 
     ~CSingleUserSecurityManager() {}
 
-    secManagerType querySecMgrType() override
+    uint64_t queryFeatures(SecurityFeatureSupportLevel level, IEspSecureContext* secureContext = nullptr) const override
+    {
+        uint64_t mask = CBaseSecurityManager::queryFeatures(level, secureContext);
```
RussWhitehead, line 49: Improve variable name, SUFFeatures or something along those lines...

system/security/plugins/htpasswdSecurity/htpasswdSecurity.cpp
```diff
@@ -60,14 +60,32 @@ class CHtpasswdSecurityManager : public CBaseSecurityManager
...
+        return mask;
+    }
+
+    secManagerType querySecMgrType(IEspSecureContext* secureContext = nullptr)
 	{
 		return SMT_HTPasswd;
 	}
 
-	inline virtual const char* querySecMgrTypeName() { return "htpasswd"; }
+    inline virtual const char* querySecMgrTypeName(IEspSecureContext* secureContext = nullptr) { return "htpasswd"; }
```
RussWhitehead, line 86: Formatting got all messed up in this source file. Not just here but in many places

RussWhitehead (commented):
A few comments listed separately.  I want to point out that ESP must be changed no longer cast the ISecManager to a LDAPSecManager, but instead call your new interface to determine the level of support. Shall we open a separate Jira for that or does on exist

timothyklemm:
@RussWhitehead the scope of this change is to enable the move away from assuming specific implementations or using dynamic_cast. The actual move will be achieved with one or more stories that I have not attempted to create. I know enough to know that I don't know enough right now to make that decision.

system/security/plugins/htpasswdSecurity/htpasswdSecurity.cpp
timothyklemm: I probably copied and pasted changes from elsewhere before I realized how many tabs were in this file to start.

system/security/shared/seclib.hpp
timothyklemm: Probably because I wanted to make my IDE stop complaining about a definition it couldn't seem to find through jlib.hpp, and I just forgot to remove it.

system/security/shared/seclib.hpp
timothyklemm: Based on comments from @afishbeck, the enumerations are being redefined to define the bit values, and not indices (with number of indices replaced by a mask for all). As long as the bit values aren't redefined from build to build, a third party plugin should work with the bits it knows about.

system/security/shared/seclib.hpp
timothyklemm: In other code that is on hold at the moment, I included using statements to define singular and plural types (in this case, feature versus features). Then the signatures were very explicit in denoting whether a singular value or multiple values were expected as input or returned. That approach was not as well received as I would have hoped, but I think it would address your concern by making the code somewhat self documenting. I can add a comment too, but I think returning a type name other than uint64_t would be even better.

system/security/shared/seclib.hpp
```diff
@@ -409,64 +399,66 @@ enum secManagerType : int
...
+    // Returns the combination of SecManagerFeature values representing features with the indicated support level.
+    virtual SecManagerFeatureSet queryFeatures(SecurityFeatureSupportLevel level, IEspSecureContext* secureContext = nullptr) const = 0;
+    // Returns true if the given combinsation of SecManagerFeature values is a subset of the complete feature set for the indicated support level.
+    inline bool checkFeatures(SecurityFeatureSupportLevel level, SecManagerFeatureSet festureSet, IEspSecureContext* secureContext = nullptr) const { return ((queryFeatures(level, secureContext) & festureSet) == festureSet); }
+    inline bool checkUnsafeFeatures(SecManagerFeatureSet festureSet, IEspSecureContext* secureContext = nullptr) const { return checkFeatures(SFSL_Unsafe, festureSet, secureContext); }
+    inline bool checkSafeFeatures(SecManagerFeatureSet festureSet, IEspSecureContext* secureContext = nullptr) const { return checkFeatures(SFSL_Safe, festureSet, secureContext); }
+    inline bool checkImplementedFeatures(SecManagerFeatureSet festureSet, IEspSecureContext* secureContext = nullptr) const { return checkFeatures(SFSL_Implemented, festureSet, secureContext); }
```
afishbeck, line 461: If you commoned up the types for SecManagerFeatureSet and SecUserFeatureSet could you put these functions in a common base interface?  ISecInterface or similar?

afishbeck: i.e. if you just used uint64_t users would know it was just a common bitmask variable.

timothyklemm: It was uint64_t, and Russ asked for comments and name changes to clarify that the values are masks.

The types could be unified, but the user currently has no knowledge or use for the secure context. Is it worth complicating the user interface with unused parameters to eliminate four one line inline methods?

afishbeck: how would using less typedefs and moving these up into a common base interface add unused parameters?

afishbeck: oh I see what you mean about the unused parameter.  How is the context used in these get supported features functions?

afishbeck: Moving these up may be a good excuse not to pass the secureContext if it's only being done in the name of consistency.

timothyklemm: When the manager wrapper (for future managers) is introduced, and the wrapper can reload the configuration, any manager call can trigger a reload. The context is being passed so that a potentially time consuming load can be tracked. I don't want people complaining about what should have been a trivial request (like returning a pre-calculated bit mask) taking too long without being able to record that the delay was caused by a manager re-configuration.

timothyklemm: It's also a consideration that a reload could actually load a different manager class, so we wouldn't be able to assume the bit masks are constants.

afishbeck: You mean the security manager may reload itself because of a call to queryFeatures?  I could see tracking the time of a reload for the various calls that may cause a reload to happen.. although we're trying to move away from complexity.

But why would a reload be triggered by calling queryFeatures?  I think we should avoid such side effects.  And why would not passing a context imply that we were assuming static feature bits?

system/security/shared/seclib.hpp
```diff
@@ -205,62 +228,52 @@ enum authStatus : int
...
+static const SecFeatureBit SUF_Clone                         = 0x0000000800000000;
+static const SecFeatureBit SUF_GetDataElement              = 0x0000001000000000;
+static const SecFeatureBit SUF_GetDataElements             = 0x0000002000000000;
+static const SecFeatureSet SUF_ALL_FEATURES                = 0x0000003FFFFFFFFF; // update to include all added feature bits
 
 class CDateTime;
 interface IPropertyIterator;
 interface IPropertyTreeIterator;
-interface ISecUser : extends IInterface
+interface ISecUser : extends IInterface, implements ISecFeatureChecks
```
afishbeck, line 275: @timothyklemm Not a huge fan of multiple inheritance unless you really are implementing interfaces that are independent.  Should ISecFeatureChecks derive from IInterface and these others derive only from ISecFeatureChecks?

timothyklemm: My intent was to define the interface in a way that it could be mixed in anywhere, without worrying about whether IInterface was already inherited through a different path. I wanted to avoid needing to use virtual inheritance. I don't have a specific use case in mind where this is an issue; it was more of a preventative measure.

I wonder if, when used as "the" base for different objects, it should be renamed to something more abstract (along the lines of ISecEntity). The current name is meant to convey the addition of one specific behavior, instead of being an umbrella of (potentially more) common behaviors.

afishbeck: The general rule is that all interface should be directly or indirectly derived from IInterface.  If you want it to be "contained" then I would have a function return it rather than use multiple inheritance.
To keep things clean I would usually say an interface class can inherit one interface, and object class can inherit one object and multiple interfaces.
I agree that when thinking of this as the common base interface the name could be better.
Could be as simple as ISecInterface or ISecObject.  Depends a bit on what else we can envision ever going in there.

afishbeck (commented):
@timothyklemm minor comments, otherwise looks fine.  Has LDAP been tested with this change?

system/security/shared/seclib.hpp
```diff
@@ -198,9 +228,51 @@ enum authStatus : int
...
+static const SecFeatureBit SUF_GetPropertyIterator           = 0x0000000400000000;
+static const SecFeatureBit SUF_Clone                         = 0x0000000800000000;
+static const SecFeatureBit SUF_GetDataElement              = 0x0000001000000000;
+static const SecFeatureBit SUF_GetDataElements             = 0x0000002000000000;
+static const SecFeatureSet SUF_ALL_FEATURES                = 0x0000003FFFFFFFFF; // update to include all added feature bits
+
 class CDateTime;
 interface IPropertyIterator;
-interface ISecUser : extends IInterface
+interface IPropertyTreeIterator;
```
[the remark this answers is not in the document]
timothyklemm: The build fails without either a forward declaration or the include of jptree.hpp (which I replaced based upon Russ' comment). The forward declaration is more consistent, given the existing forward declaration of IPropertyIterator on the line above. The declaration may be visible in many places this file is used, but that is probably just a coincidence that the header was included (either directly or indirectly) before this one.

system/security/shared/seclib.hpp
```diff
@@ -198,9 +228,51 @@ enum authStatus : int
...
+static const SecFeatureBit SUF_GetPasswordDaysRemaining    = 0x0000000020000000;
+static const SecFeatureBit SUF_GetProperty                 = 0x0000000040000000;
+static const SecFeatureBit SUF_SetProperty                 = 0x0000000080000000;
+static const SecFeatureBit SUF_GetPropertyInt                = 0x0000000100000000;
+static const SecFeatureBit SUF_SetPropertyInt                = 0x0000000200000000;
+static const SecFeatureBit SUF_GetPropertyIterator           = 0x0000000400000000;
+static const SecFeatureBit SUF_Clone                         = 0x0000000800000000;
+static const SecFeatureBit SUF_GetDataElement              = 0x0000001000000000;
+static const SecFeatureBit SUF_GetDataElements             = 0x0000002000000000;
+static const SecFeatureSet SUF_ALL_FEATURES                = 0x0000003FFFFFFFFF; // update to include all added feature bits
```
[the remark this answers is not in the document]
timothyklemm: The alignment was done intentionally - every four rows, corresponding to every hex digit - because it was easier for me to confirm I was setting correct bits.

system/security/LdapSecurity/ldapsecurity.ipp
```diff
@@ -370,50 +376,51 @@ public:
...
+    ISecUser * createUser(const char * user_name, IEspSecureContext* secureContext = nullptr);
+    ISecResourceList * createResourceList(const char * rlname, IEspSecureContext* secureContext = nullptr);
+    bool subscribe(ISecAuthenticEvents & events, IEspSecureContext* secureContext = nullptr);
+    bool unsubscribe(ISecAuthenticEvents & events, IEspSecureContext* secureContext = nullptr);
+    bool authorize(ISecUser& sec_user, ISecResourceList * Resources, IEspSecureContext* secureContext = nullptr);
+    bool authorizeEx(SecResourceType rtype, ISecUser& sec_user, ISecResourceList * Resources, IEspSecureContext* secureContext = nullptr);
+    SecAccessFlags authorizeEx(SecResourceType rtype, ISecUser& sec_user, const char* resourcename, IEspSecureContext* secureContext = nullptr);
+    virtual SecAccessFlags authorizeFileScope(ISecUser & user, const char * filescope, IEspSecureContext* secureContext = nullptr);
+    virtual bool authorizeFileScope(ISecUser & user, ISecResourceList * resources, IEspSecureContext* secureContext = nullptr);
+    virtual SecAccessFlags authorizeWorkunitScope(ISecUser & user, const char * wuscope, IEspSecureContext* secureContext = nullptr);
```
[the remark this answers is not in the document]
timothyklemm: I agree, which is why the methods I added use override instead of virtual. I wasn't sure that making what is essentially a logically transparent change to existing signatures was in the scope of this change, or if it would be viewed like replacing tabs (valid but unnecessary changes that the make it harder to recognize the intent of the PR).

esp/bindings/http/platform/httpservice.cpp
```diff
@@ -2019,8 +2019,8 @@ void CEspHttpServer::logoutSession(EspAuthRequest& authReq, unsigned sessionID,
                 if (user)
                 {
                     //inform security manager that user is logged out
-                    Owned<ISecUser> secUser = secmgr->createUser(user);
-                    secmgr->logoutUser(*secUser);
+                    Owned<ISecUser> secUser = secmgr->createUser(user, ctx->querySecureContext());
```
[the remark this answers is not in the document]
timothyklemm: Will change.

afishbeck (approved):
@timothyklemm looks good.

afishbeck:
@timothyklemm can you squash the commits?

timothyklemm:

richardkchapman:
@RussWhitehead Have you approved this PR?

RussWhitehead:
@richardkchapman I approve, please merge

PR #13901, opened 2020-06-24
HPCC-23063 Fix security manager plugin loading

timothyklemm:
Accept configuration attribute "libName" as an alternative to "LibName".
Existing plugins create configurations with the former and fail to load.

Accept configuration attribute "instanceFactoryName" as an alternative to
"InstanceFactoryName". Existing plugins create configurations with the former
and expect a hard-coded default value.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/security/shared/secloader.hpp
+++ b/system/security/shared/secloader.hpp
@@ -47,12 +47,22 @@ public:
         StringBuffer libName, instFactory;
         secMgrCfg->getProp("@LibName", libName);
         if (libName.isEmpty())
-            throw MakeStringException(-1, "%s library name not specified for %s", lsm, bindingName);
+        {
+            // @libName is commonly used
+            secMgrCfg->getProp("@libName", libName);
+            if (libName.isEmpty())
+                throw MakeStringException(-1, "%s library name not specified for %s", lsm, bindingName);
+        }
         //TODO Search for LibName in plugins folder, or in specified location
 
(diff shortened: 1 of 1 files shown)


timothyklemm:
@RussWhitehead this is a small fix to get security manager plugins loading. I tried using the single user plugin to test HPCC-24312, and it wouldn't load because the configurations created for it use the wrong case.

RussWhitehead (commented):
Although this looks fine, why is it better than just fixing the offending plugins to do it correctly?

timothyklemm:
I ran into this with my plugin when moving to 7.0. The lower case names worked in 6.x, and stopped working in 7.0. I just updated my plugin. That works for configurations made after the fix, but not for configurations created before. By changing the loader to accept either form, we fix everybody.

RussWhitehead:
Who generates these keynames in the configuration?  Should we open a Jira against the  plugin config files to make them consistent, even if you handle it in this code?  Otherwise, this looks good for merge

timothyklemm:
The plugin defines the names it uses. There is no opportunity for a transform to change the plugin's names to match the names expected by code. We could fix the plugins so new configurations match the first attempt in code, but retain the fallback for older configurations.

PR #13902, opened 2020-06-24
HPCC-24155 Update Kafka plugin to support UTF-8 encoded messages

dcamper:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [x] Performance
  - [ ] Security
  - [x] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

## Testing:

Added UTF8-specific regression tests.  All existing and new tests pass on all engines.

### DIFF
--- a/plugins/kafka/kafka.cpp
+++ b/plugins/kafka/kafka.cpp
@@ -921,6 +921,17 @@ namespace KafkaPlugin
     // Advertised Entry Point Functions
     //--------------------------------------------------------------------------
 
+    ECL_KAFKA_API bool ECL_KAFKA_CALL publishMessage(ICodeContext* ctx, const char* brokers, const char* topic, const char* message, const char* key)
+    {
+        std::call_once(pubCacheInitFlag, setupPublisherCache, ctx->queryContextLogger().queryTraceLevel());
+
+        Publisher* pubObjPtr = publisherCache->getPublisher(brokers, topic, POLL_TIMEOUT);
+
+        pubObjPtr->sendMessage(message, key);
+
+        return true;
+    }
+
     ECL_KAFKA_API bool ECL_KAFKA_CALL publishMessage(ICodeContext* ctx, const char* brokers, const char* topic, size32_t lenMessage, const char* message, size32_t lenKey, const char* key)
     {
         std::call_once(pubCacheInitFlag, setupPublisherCache, ctx->queryContextLogger().queryTraceLevel());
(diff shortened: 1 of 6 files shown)


plugins/kafka/kafka.cpp
```diff
@@ -1031,9 +1033,10 @@ namespace KafkaPlugin
 // Plugin Initialization and Teardown
 //==============================================================================
 
-#define CURRENT_KAFKA_VERSION "kafka plugin 1.0.0"
+#define CURRENT_KAFKA_VERSION "kafka plugin 1.1.0"
 
 static const char* kafkaCompatibleVersions[] = {
+    "kafka plugin 1.0.0",
```
dcamper, line 1039: I am unsure if changes made in this PR resulted in an incompatible module or not.  If it is incompatible, this line should be removed.  Feedback welcome.

dcamper:
@richardkchapman Please review or assign for review.  Thanks!

plugins/kafka/kafka.cpp
richardkchapman: To be honest the compatible versions mechanism is pretty redundant these days, as we ship plugins in lock-step with platform versions and expect them to match.

richardkchapman:
I think in the interests of existing ECL code continuing to compile, I would expect to add new functions taking/returning UTF8 rather than replacing the exiting ones.

@ghalliday Any comments?

dcamper:
New compiles should Just Work, as evidenced by the original (unchanged) regression tests.  I am curious how the change would affect something like deployed ROXIE queries, though.

richardkchapman:
They would fail to load due to unresolved symbol, if they called publishMessage with the old parameter mangling.

You can avoid this by leaving the old function in place (calling on to the new, if that's simplest)

Something like:

ECL_KAFKA_API bool ECL_KAFKA_CALL publishMessage(const char* brokers, const char* topic, const char* message, const char* key)
{
return publishMessage(nullptr, brokers, topic, strlen(message), message, strlen(key, key);
}

You'd need to ensure that the function you call on to would accept ctx being null .

Incidentally, should brokers and topic also be utf8 ?

dcamper:
Excellent point, Richard.  I should have done that to begin with.  I will make that change now.

dcamper:
FYI, the previous code version actually had an incorrect header method signature for publishMessage().  The .ecllib file cites a context, the header did not, but the actual method did include a context parameter.  So the previous code advertised a method that did not actually exist while including a concrete method that just happened to match what callers were expecting.  TL;DR, a null contextPtr isn't a problem.

richardkchapman:
A null context will crash:

        std::call_once(pubCacheInitFlag, setupPublisherCache, ctx->queryContextLogger().queryTraceLevel());

dcamper:
Yes, but it doesn't happen; that's what I'm saying.  The advertised signature in the header (the one without ctx) is never referenced.

PR #13913, opened 2020-06-26
HPCC-24323 Assign jlog detail thresholds

rpastrana:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jlog.hpp
+++ b/system/jlib/jlog.hpp
@@ -236,6 +236,21 @@ typedef unsigned LogMsgDetail;
 #define DefaultDetail   100
 #define TopDetail (LogMsgDetail)-1
 
+/*
+ * Log message thresholds, assigned to log message category types.
+ * It represents the lowest logging level (detail) required to output
+ * messages of the given category.
+ */
+constexpr int CriticalMsgThreshold    = 1;  //Use to declare categories reporting critical events (log level => 1)
+constexpr int FatalMsgThreshold       = 1;  //Use to declare categories reporting Fatal events (log level => 1)
+constexpr int ErrMsgThreshold         = 10; //Use to declare categories reporting Err messages (log level => 10)
+constexpr int WarnMsgThreshold        = 20; //Use to declare categories reporting Warn messages (log level => 20)
+constexpr int AudMsgThreshold         = 30; //Use to declare categories reporting Aud messages (log level => 30)
+constexpr int ProgressMsgThreshold    = 50; //Use to declare categories reporting Progress messages (log level => 50)
+constexpr int InfoMsgThreshold        = 60; //Use to declare categories reporting Info messages (log level => 60)
+constexpr int DebugMsgThreshold       = 80; //Use to declare categories reporting Debug messages (log level => 80)
+constexpr int ExtraneousMsgThreshold  = 90; //Use to declare categories reporting Extraneous messages (log level => 90)
+
 // Typedef for LogMsgSysInfo
 
 typedef unsigned LogMsgId;
@@ -796,23 +811,23 @@ extern jlib_decl ILogMsgHandler * attachLogMsgMonitorFromPTree(IPropertyTree * t
 extern jlib_decl void attachManyLogMsgMonitorsFromPTree(IPropertyTree * tree);            // Takes tree containing many <monitor> elements
 
 // Standard categories and unknown jobInfo
-constexpr LogMsgCategory MCdisaster(MSGAUD_all, MSGCLS_disaster);
-constexpr LogMsgCategory MCuserError(MSGAUD_user, MSGCLS_error);
-constexpr LogMsgCategory MCoperatorError(MSGAUD_operator, MSGCLS_error);
-constexpr LogMsgCategory MCinternalError(MSGAUD_programmer, MSGCLS_error, 1);
-constexpr LogMsgCategory MCauditError(MSGAUD_audit, MSGCLS_error);
-constexpr LogMsgCategory MCuserWarning(MSGAUD_user, MSGCLS_warning);
-constexpr LogMsgCategory MCoperatorWarning(MSGAUD_operator, MSGCLS_warning);
-constexpr LogMsgCategory MCinternalWarning(MSGAUD_programmer, MSGCLS_warning, 1);
-constexpr LogMsgCategory MCauditWarning(MSGAUD_audit, MSGCLS_warning);
-constexpr LogMsgCategory MCuserProgress(MSGAUD_user, MSGCLS_progress);
-constexpr LogMsgCategory MCoperatorProgress(MSGAUD_operator, MSGCLS_progress);
-constexpr LogMsgCategory MCdebugProgress(MSGAUD_programmer, MSGCLS_progress);
-constexpr LogMsgCategory MCuserInfo(MSGAUD_user, MSGCLS_information);
-constexpr LogMsgCategory MCdebugInfo(MSGAUD_programmer, MSGCLS_information);
-constexpr LogMsgCategory MCauditInfo(MSGAUD_audit, MSGCLS_information);
-constexpr LogMsgCategory MCstats(MSGAUD_operator, MSGCLS_progress);
-constexpr LogMsgCategory MCoperatorInfo(MSGAUD_operator, MSGCLS_information);
+constexpr LogMsgCategory MCdisaster(MSGAUD_all, MSGCLS_disaster, FatalMsgThreshold);
+constexpr LogMsgCategory MCuserError(MSGAUD_user, MSGCLS_error, ErrMsgThreshold);
+constexpr LogMsgCategory MCoperatorError(MSGAUD_operator, MSGCLS_error, ErrMsgThreshold);
+constexpr LogMsgCategory MCinternalError(MSGAUD_programmer, MSGCLS_error, ErrMsgThreshold);
+constexpr LogMsgCategory MCauditError(MSGAUD_audit, MSGCLS_error, ErrMsgThreshold);
+constexpr LogMsgCategory MCuserWarning(MSGAUD_user, MSGCLS_warning, WarnMsgThreshold);
+constexpr LogMsgCategory MCoperatorWarning(MSGAUD_operator, MSGCLS_warning, WarnMsgThreshold);
+constexpr LogMsgCategory MCinternalWarning(MSGAUD_programmer, MSGCLS_warning, WarnMsgThreshold);
+constexpr LogMsgCategory MCauditWarning(MSGAUD_audit, MSGCLS_warning, WarnMsgThreshold);
+constexpr LogMsgCategory MCuserProgress(MSGAUD_user, MSGCLS_progress, ProgressMsgThreshold);
+constexpr LogMsgCategory MCoperatorProgress(MSGAUD_operator, MSGCLS_progress, ProgressMsgThreshold);
+constexpr LogMsgCategory MCdebugProgress(MSGAUD_programmer, MSGCLS_progress, DebugMsgThreshold);
+constexpr LogMsgCategory MCuserInfo(MSGAUD_user, MSGCLS_information, InfoMsgThreshold);
+constexpr LogMsgCategory MCdebugInfo(MSGAUD_programmer, MSGCLS_information, DebugMsgThreshold);
+constexpr LogMsgCategory MCauditInfo(MSGAUD_audit, MSGCLS_information, AudMsgThreshold);
+constexpr LogMsgCategory MCstats(MSGAUD_operator, MSGCLS_progress, ProgressMsgThreshold);
+constexpr LogMsgCategory MCoperatorInfo(MSGAUD_operator, MSGCLS_information, InfoMsgThreshold);
 
 inline LogMsgCategory MCexception(IException * e, LogMsgClass cls = MSGCLS_error) { return LogMsgCategory((e)->errorAudience(),cls); }
 


rpastrana:
@jakesmith please review.
Note: the thresholds chosen are somewhat arbitrary, and can certainly be adjusted. 
The most important aspect to note is that although the mechanism to filter out log messages by desired log level (detail), most message categories were assigned the max, and very few were assigned the lowest level, even though the range was established to be 1-100.

system/jlib/jlog.hpp
```diff
@@ -796,23 +806,23 @@ extern jlib_decl ILogMsgHandler * attachLogMsgMonitorFromPTree(IPropertyTree * t
...
-constexpr LogMsgCategory MCdebugProgress(MSGAUD_programmer, MSGCLS_progress);
-constexpr LogMsgCategory MCuserInfo(MSGAUD_user, MSGCLS_information);
-constexpr LogMsgCategory MCdebugInfo(MSGAUD_programmer, MSGCLS_information);
-constexpr LogMsgCategory MCauditInfo(MSGAUD_audit, MSGCLS_information);
-constexpr LogMsgCategory MCstats(MSGAUD_operator, MSGCLS_progress);
-constexpr LogMsgCategory MCoperatorInfo(MSGAUD_operator, MSGCLS_information);
+constexpr LogMsgCategory MCdisaster(MSGAUD_all, MSGCLS_disaster, FatalMsgThreshold);
+constexpr LogMsgCategory MCuserError(MSGAUD_user, MSGCLS_error, ErrMsgThreshold);
+constexpr LogMsgCategory MCoperatorError(MSGAUD_operator, MSGCLS_error, ErrMsgThreshold);
+constexpr LogMsgCategory MCinternalError(MSGAUD_programmer, MSGCLS_error, AllMsgThreshold);
```
jakesmith, line 812: Internal logging this should have a high default threshold, could perhaps share default threshold of another one, e.g. DebugMsgThreshold

system/jlib/jlog.hpp
```diff
@@ -236,6 +236,16 @@ typedef unsigned LogMsgDetail;
 #define DefaultDetail   100
 #define TopDetail (LogMsgDetail)-1
 
+constexpr int AllMsgThreshold         = 1;
+constexpr int FatalMsgThreshold       = 1;
+constexpr int ErrMsgThreshold         = 20-1;
+constexpr int WarnMsgThreshold        = 30-1;
```
jakesmith, line 242: does this suggest 2-19 are categorized as error messages?
20-29 are categorized as warning messages?
For clarity would be good to add a comment saying these are ranges, x->y

system/jlib/jlog.hpp
```diff
@@ -796,23 +806,23 @@ extern jlib_decl ILogMsgHandler * attachLogMsgMonitorFromPTree(IPropertyTree * t
...
-constexpr LogMsgCategory MCuserInfo(MSGAUD_user, MSGCLS_information);
-constexpr LogMsgCategory MCdebugInfo(MSGAUD_programmer, MSGCLS_information);
-constexpr LogMsgCategory MCauditInfo(MSGAUD_audit, MSGCLS_information);
-constexpr LogMsgCategory MCstats(MSGAUD_operator, MSGCLS_progress);
-constexpr LogMsgCategory MCoperatorInfo(MSGAUD_operator, MSGCLS_information);
+constexpr LogMsgCategory MCdisaster(MSGAUD_all, MSGCLS_disaster, FatalMsgThreshold);
+constexpr LogMsgCategory MCuserError(MSGAUD_user, MSGCLS_error, ErrMsgThreshold);
+constexpr LogMsgCategory MCoperatorError(MSGAUD_operator, MSGCLS_error, ErrMsgThreshold);
+constexpr LogMsgCategory MCinternalError(MSGAUD_programmer, MSGCLS_error, AllMsgThreshold);
+constexpr LogMsgCategory MCauditError(MSGAUD_audit, MSGCLS_error, ErrMsgThreshold-10);
```
jakesmith, line 813: A bit confusing. To this sets it to 9, so a lower threshold than MCuserError. 
I don't you gain any clarity by separating out the level of audit vs user levels, which can already be separated out by audience.

system/jlib/jlog.hpp
```diff
@@ -796,23 +806,23 @@ extern jlib_decl ILogMsgHandler * attachLogMsgMonitorFromPTree(IPropertyTree * t
...
+constexpr LogMsgCategory MCuserWarning(MSGAUD_user, MSGCLS_warning, WarnMsgThreshold+5);
```
jakesmith, line 814: as above, it's a warning, so think clearer if has same threshold, and can already filter if wanted by audience.

jakesmith:
@rpastrana - looks sensible change, a couple of minor comments.

system/jlib/jlog.hpp
rpastrana: Agreed, I think we're in synch but using contrary terms. Internal errs are assigned the lowest threshold. In affect, internal errors will always be streamed.

High threshold == only output if the admin has requested logging at a high level

system/jlib/jlog.hpp
rpastrana: I agree here as well. I think the category itself is rather confusing as it is - I think it means it is a high priority audit level event....  maybe this belongs to the AllMsgThreshold.  or maybe just above that threshold AllMsgThreshold+10

system/jlib/jlog.hpp
rpastrana: well... we have thresholds, and windows in between those thresholds. Some times the category doesn't belong at the base level. You might be right about this particular category, but in general designating a category to be slightly above a particular threshold is fine. Otherwise, the range should be 1-10 (which it prob should be)

system/jlib/jlog.hpp
jakesmith: yes, internal should only be output if admin has requested logging at a high level.

system/jlib/jlog.hpp
jakesmith: not sure, audit may or may not be wanted, vs diagnostic logging for example.
Audit logging historically has gone to Dali, and written to a separate audit log trail.
You could argue that the difference audiences, should be able to have different logging levels, so that you could turn off user, or operator logging for example.

system/jlib/jlog.hpp
jakesmith: there's perhaps no harm differentiating, but I'm not sure it's clear to someone who might configure the system why these differ, where as audience, e.g. audit , internal, user do carry more meaning.
I'd make them the same, and if we ever wanted to differentiate, then allow e.g. "user" audience messages to have a different logging level.

system/jlib/jlog.hpp
rpastrana: You might be 100% correct, It might not be wanted but this change isn't intended to edit the predefined logmessage categories. I'm only assigning levels in order to make the logging.details operable. 

With that said, I did take a peek and that particular category isn't used much, I only found one instance of it:
in jthread: AERRLOG("Unauthorized pipe program(%s)",head.str());

That seems to imply a different type of audit. 
For this issue I'm assigning that log message category to  AllMsgThreshold.

system/jlib/jlog.hpp
jakesmith: I am not sure that's true.
I thought internal, is akin to programmer, i.e. only of interest to devs, and so only if logging level turned up.
Most of the IERRLOG with this category, do not seem useful to most, so should not always be output afaics.

system/jlib/jlog.hpp
```diff
@@ -796,23 +806,23 @@ extern jlib_decl ILogMsgHandler * attachLogMsgMonitorFromPTree(IPropertyTree * t
...
+constexpr LogMsgCategory MCinternalWarning(MSGAUD_programmer, MSGCLS_warning, WarnMsgThreshold);
+constexpr LogMsgCategory MCauditWarning(MSGAUD_audit, MSGCLS_warning, WarnMsgThreshold);
+constexpr LogMsgCategory MCuserProgress(MSGAUD_user, MSGCLS_progress, ProgressMsgThreshold);
+constexpr LogMsgCategory MCoperatorProgress(MSGAUD_operator, MSGCLS_progress, ProgressMsgThreshold);
+constexpr LogMsgCategory MCdebugProgress(MSGAUD_programmer, MSGCLS_progress, DebugMsgThreshold);
+constexpr LogMsgCategory MCuserInfo(MSGAUD_user, MSGCLS_information, InfoMsgThreshold);
+constexpr LogMsgCategory MCdebugInfo(MSGAUD_programmer, MSGCLS_information, DebugMsgThreshold);
+constexpr LogMsgCategory MCauditInfo(MSGAUD_audit, MSGCLS_information, AudMsgThreshold);
+constexpr LogMsgCategory MCstats(MSGAUD_operator, MSGCLS_progress, ProgressMsgThreshold);
+constexpr LogMsgCategory MCoperatorInfo(MSGAUD_operator, MSGCLS_information, InfoMsgThreshold+5);
```
jakesmith, line 825: I think as with other ones (already changed), I think should have same threshold.

system/jlib/jlog.hpp
rpastrana: Well, Internal Error means just that, and those should always be reported. 
I do see a lot instances of IERRLOG that seem to contain debug-like messages... I consider that to be a misuse of the IERRLOG function

system/jlib/jlog.hpp
jakesmith: It seems our definition of "internal" means "an internal programming warning/error" and not "serious internal error"
The vast majority of messages in this category should be in a lower priority than FMCatal* 

I'm not sure if this breaks from convention others use in general, or if it simply depends on different definition of "internal" and there is no concise formal meaning.

To keep things consistent, we should decide in this PR, whether to change all our uses of MCInternal, if they are misleading/against best practices, to a lower priority category.
Or, change the MCInternal threshold from being same as FatalMsgThreshold as it is as the moment, to a threshold level that better fits the majority of it's use cases, e.g. ProgressMsgThreshold.

The latter is the simplest change right now, and since the default maxDetail means it will not suppress any logging anyway, it will only make any difference is logging detail is turned down.
We can then think about revisiting and changing the meaning of how MCInternal is used if we want later in a separate JIRA/PR.

system/jlib/jlog.hpp
```diff
@@ -236,6 +236,16 @@ typedef unsigned LogMsgDetail;
 #define DefaultDetail   100
 #define TopDetail (LogMsgDetail)-1
 
+constexpr int AllMsgThreshold         = 1;
+constexpr int FatalMsgThreshold       = 1;
+constexpr int ErrMsgThreshold         = 20-1;
```
jakesmith, line 241: from offline conv. and your amended version:
```
constexpr int ErrMsgThreshold         = 10-1; //Used to declare categories which report Err messages (log level => 10)
constexpr int WarnMsgThreshold        = 20-1; //Used to declare categories which report Warn messages (log level => 20)
```

can these be ?:
```
constexpr int ErrMsgThreshold         = 10; //Used to declare categories which report Err messages (log level => 10 < 20)
constexpr int WarnMsgThreshold        = 20; //Used to declare categories which report Warn messages (log level => 20 < 30)
```

i.e. boundary start at 10,20,30 .. 

I think there was a suspicions that if starts at 10, and the maxDetails were [e.g.] set to 10, the jlog impl. would exclude them.

rpastrana:
@jakesmith please review. 
Also as FYI: https://track.hpccsystems.com/browse/HPCC-24366

system/jlib/jlog.hpp
jakesmith: As per discussion offline, in JIRA and with @shamser.
The default override level fo the MC\*internal* categories, appears to have always been wrong.
i.e. setting it to the absolute highest priority and equal to MCdisaster can't be right.

I think we should get rid of AllMsgThreshold, and change MCinternalError to use same default threshold as other error categories, i.e.:
```
constexpr LogMsgCategory MCinternalError(MSGAUD_programmer, MSGCLS_error, ErrMsgThreshold);
```

@shamser - does that make sense?

system/jlib/jlog.hpp
jakesmith: @rpastrana - still don't think these +5 , -10 etc. are helpful.

I think should just be InfoMsgThreshold here, and similar in other places where -10, -5.
If ever wanted to differentiate, then they are differentiated by audience and could be filtered accordingly. 

@shamser - what do you think?

system/jlib/jlog.hpp
shamser: I agree with @jakesmith  on using the set threshold levels (i.e. using WarnMsgThreshold)

shamser (changes requested):
@rpastrana @jakesmith These changes make a lot of sense.  I've made couple of suggestions relating to making the thresholds less complicated but overall it look good.

system/jlib/jlog.hpp
shamser: I agree with @jakesmith on this one.  I don't feel there is any need have an offset simply to distinguish the audience.

By the way, is there a configuration option to set which audience types are logged?

shamser:
@rpastrana @jakesmith

jakesmith:
@rpastrana

rpastrana:
@jakesmith per offline convo I think we're now completely in synch, please review.

jakesmith (approved):
Looks good.

@ghalliday - please merge

ghalliday:
@rpastrana there some other cases that haven't been covered by this change:  In particular I would be suspicious of any use of DefaultDetail.  Only example I have found so far is MCexception in jlog.hpp(833).

(I removed the default detail level from the constructor of LogMsgCategory to find it.)

ghalliday:
@rpastrana do you wnat me to merge as-is?

PR #13914, opened 2020-06-29
HPCC-24460 Explore the definition of storage planes

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:


ghalliday:
@jakesmith @richardkchapman @afishbeck 
This is my "straw-man" pull request for a more general structure for defining how the data is stored in a system. Please criticise it and pull it to pieces.

It is based around the idea of multiple storage planes.  There are 3 broad categories
i) accessed via a local file mount
ii) accessed via a particular file api (e.g., azure/s3)  (protocol, no mount)
iii) physical files using dafilesrv (hosts defined).  (Arguably a variant of ii.)

I propose that each component instance has a default storage plane for storing data, but that can be overridden for a cluster, workunit, workflow item or output using the current cluster syntax.  There may also be a desire to map the default storage plane for a given top-level scope.

I suspect something similar (but different) should be used as a mechanism for mapping azure blobs and named drop zones.  Some (soft-reserved) scope names which map to a storage plane, but imply file:: access syntax/rules.   This could be a special kind of data plane.

Note: If no planes are defined the system will continue to use the existing mechanism.

Some issues/questions
- Does this cover all the cases that we want to support?
- Is it overkill?
- Do we define spill file location in the same way (as a new category)?
- Long term an esp meta-data server may need to be the only component that requires the configuration of the storage planes.  That would allow changes to be made, and esp replaced to allow them to be active.
- the components need to have access to the mounts.  Does this mean that if a new mount is required that the component needs to be restarted?  (I suspect it may...)
- Should the components only mount a subset of the planes.  I don't see any big advantage, and it could cause problems.

This PR does not include
- Schema for the new section.
- Copying the hostname configuration
- Overriding the default storage plane for a component

RussWhitehead:
@ghalliday @afishbeck Is this the place for discussing protecting data file access, possibly similar to how we currently use LDAP and file scope rules?

ghalliday:
@RussWhitehead probably not directly.  I had been expecting that wouldn't change - our mechanisms for protecting data file access remain the same.  i.e. We continue to use LDAP with the file scope rules, and restricted access flag.

This is more to do with where the data for a logical file is actually stored.

afishbeck:
Still processing, but some basic questions:

1. Are our replication schemes primarily for data availability?  If so do we still need to do it ourselves or could we start to rely on cloud platform replication?

2. Where will protocol specific configuration be stored?  For example given the object:

- name: azureBlobPlane
    prefix: azure://ghallidaystorage      # Not sure if this should be different from the mount.
    speed: 500
Assuming we need to build a connection string something like:
DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;
AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;
BlobEndpoint=http://127.0.0.1:10000/devstoreaccount1;

Does ghallidaystorage map to?:
a. the secret name to retrieve the account key
b. the account name
What about the BlobEndpoint, and underlying protocol etc.
Same questions would apply for any protocol we decide to support so a generic, "the protocol config goes here" would be good.

afishbeck:
If the storage plane secrets were stored as application level secrets (azure keyvault or hashicorp vault for example) then it would be easier at least in that aspect to have dynamic storage plains that don't require a restart.

afishbeck:
Just noticed this line "Long term an esp meta-data server may need to be the only component that requires the configuration of the storage planes. That would allow changes to be made, and esp replaced to allow them to be active."  Which when implemented would answer my questions about where the config may go and how it could be more dynamic.

ghalliday:
@afishbeck 

> Are our replication schemes primarily for data availability? If so do we still need to do it ourselves or could we start to rely on cloud platform replication?

On cloud platforms they would almost certainly be done by the cloud provider.  Replication is still useful on bare-metal systems, and also in-house private clouds without the big providers replication support.

> Where will protocol specific configuration be stored? For example given the object:

At the moment I had assumed any key information would be part of the prefix (or a secret).  E.g. in the azure case, the prefix defines the storageaccount (and would probably also define the blob container).  The access key is defined by a secret whose name is derived from the storageaccount.  (azure-<account>).

We could have an extra field (or structure) to contain extra protocol specific information, but if it cannot be encoded in the filename then I'm not sure how we would pass the information through without refactoring the low level interfaces.  (Do you have any examples that can't be done that way - if so we will need to think about doing that work earlier rather than later.)

I'm not sure what the blobendpoint is - I never needed it.  The protocol I think I defaulted to https.  I can see why you might want to configure that.

afishbeck:
For azure resources perhaps the secret should be the entire connection string?  Do you ever need the account key outside of the connection string?
There are actually different flavors of connection strings.  You can use a shared access signature rather than the full account name and key I think.  There are other ways you get a token from Azure ActiveDirectory and need less secure access strings (maybe).
The BlobEndpoint and other endpoints are defaulted I think but can be made explicit:
`"One scenario where you might wish to specify an explicit endpoint is when you've mapped your Blob storage endpoint to a custom domain. In that case, you can specify your custom endpoint for Blob storage in your connection string. You can optionally specify the default endpoints for the other services if your application uses them."`
My main point in showing that was that we may need to be flexible and leave that up to the user.. perhaps store the entire connection string as a secret?

afishbeck:
In Azure we should probably (at least optionaly) support pod managed identities for accessing resources and services.
https://docs.microsoft.com/en-us/azure/aks/developer-best-practices-pod-security

image

image

helm/examples/storageplanes/baremetal.yaml
```diff
@@ -0,0 +1,46 @@
+#An example of a complex storeage definition for a bare metal system
```
richardkchapman, line 1: typo - storeage

helm/hpcc/values.yaml
```diff
@@ -36,6 +36,18 @@ global:
...
 
 storage:
+  planes:
+  # name: default settings
+  #   name: <required>
+  #   prefix: <path>                        # Root directory for accessing the plane (if pvc defined), or url to access plane.
+  #   numDevices: 1                         # number of devices that are part of the plane
+  #   speed: 1000                           # idea is that if the file is on multiple planes, this can help select between them, would (latency, throughput be better?). Higher=better
+  #   replication: nullptr                  # a list or single item indicating which planes the data should be replicated onto
+  #   includeDeviceInPath: false            # Is the device number appended to the mount for the physical path for a file?
```
richardkchapman, line 46: This feels a bit dodgy - if it's something you have to set one way for some types and another way for other types, then it's not something that the helm chart user should be setting up...

Perhaps a philosophical question.

ghalliday: I think this is only ever going to be set in weird bare-metal configurations that we don't currently support (to allow local files to be spread over multiple disks.)  Otherwise it is equal to (numDevices != 1)

ghalliday:
@jakesmith please can you have an initial look at this.  I think what is here is ready, it may not be complete.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -101,19 +125,73 @@ Add ConfigMap volume for a component
 
 {{/*
 Add data volume mount
+If Any storage planes are defined, then they specify the pvcs that will be mounted
```
jakesmith, line 128: if any storage planes are defined that name pvc's ?

helm/hpcc/values.schema.json
```diff
@@ -146,14 +156,85 @@
...
+          "type": "boolean"
+        },
+        "speed": {
+          "description": "the speed of reading the data direct from a local mount/connection",
+          "type": "integer"
+        },
+        "remoteSpeed": {
+          "description": "the speed of reading indirectly",
+          "type": "integer"
+        },
```
jakesmith, line 230: I think agree, as you were saying, that these type of options are best left/re-introduced later, if/when we start to have heuristics to use them.

helm/hpcc/values.schema.json
```diff
@@ -146,14 +156,85 @@
...
+        "pvc": {
+          "description": "optional name of the persistent volume claim for this plane",
+          "type": "string"
+        },
+        "hosts": {
+          "description": "optional name of the host group (for bare metal storage)",
+          "type": "string"
+        },
+        "numDevices": {
+          "description": "optional name of devices in the storage plane",
```
jakesmith, line 216: optional number of devices

jakesmith:
@ghalliday - a few minor comments.

I haven't tried the PR yet, I think I should be able to define a storage plane that becomes is used for data, i.e. replaces the legacy dataStorage PVC definition, if (for now) I ensure that it has the same mount path ?

Should dllStorage and daliStorage also work that way (in _helpers.tpl)?

helm/examples/storageplanes/baremetal.yaml
```diff
@@ -0,0 +1,46 @@
...
+    count: 100
+    offset: 300
+
+storage:
+  planes:
+  #Bare metal system with attached storage
+  - name: thor400
+    prefix: /var/lib/hpccsystems/hpcc-data       # only used if the local host matches the host for the device
+    hosts: thor400
+    replication: [ attachedThor400MirrorPlane ]
```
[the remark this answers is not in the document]
ghalliday: should be thor400mirror

ghalliday:
@jakesmith please can you look at this again.  I think all of your comments have been addressed.  The Helm scripts have changed significantly, so probably worth reviewing as if it is new.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -149,6 +293,10 @@ Add the secret volume mounts for a component
 {{- end -}}
 {{- end -}}
 
+{{- define "hpcc.echo" -}}
```
richardkchapman, line 296: Is this used? A debugging aid?

If it's used it should have a comment describing what it is. If it's a debugging aid, perhaps it can now be removed?

helm/hpcc/templates/_helpers.tpl
```diff
@@ -101,36 +118,163 @@ Add ConfigMap volume for a component
...
+{{- /*Create local variables which always exist to avoid having to check if intermediate key values exist*/ -}}
+{{- $storage := (.root.Values.storage | default dict) -}}
+{{- $planes := ($storage.planes | default list) -}}
+{{- $dataStorage := ($storage.dataStorage | default dict) -}}
+{{- $spillStorage := ($storage.spillStorage | default dict) -}}
+{{- $daliStorage := ($storage.daliStorage | default dict) -}}
+{{- $dllStorage := ($storage.dllStorage | default dict) -}}
+{{- $daliStoragePlane := ($daliStorage.plane | default "") -}}
+{{- $dllStoragePlane := ($dllStorage.plane | default "") -}}
+{{- range $plane := .root.Values.storage.planes -}}
```
richardkchapman, line 168: Should this be using $planes? If not,  I don't think $planes is used

helm/hpcc/templates/_helpers.tpl
```diff
@@ -101,36 +118,163 @@ Add ConfigMap volume for a component
...
 
 {{/*
 Add data volume
 */}}
 {{- define "hpcc.addDataVolume" -}}
+{{- /*Create local variables which always exist to avoid having to check if intermediate key values exist*/ -}}
+{{- $storage := (.root.Values.storage | default dict) -}}
+{{- $planes := ($storage.planes | default list) -}}
+{{- $dataStorage := ($storage.dataStorage | default dict) -}}
+{{- $spillStorage := ($storage.spillStorage | default dict) -}}
```
richardkchapman, line 163: Not sure all these are used?

helm/hpcc/templates/_helpers.tpl
```diff
@@ -101,36 +118,163 @@ Add ConfigMap volume for a component
...
 {{/*
 Add data volume mount
+If any storage planes are defined that name pvcs they will be mounted
 */}}
 {{- define "hpcc.addDataVolumeMount" -}}
+{{- /*Create local variables which always exist to avoid having to check if intermediate key values exist*/ -}}
+{{- $storage := (.root.Values.storage | default dict) -}}
+{{- $planes := ($storage.planes | default list) -}}
+{{- $dataStorage := ($storage.dataStorage | default dict) -}}
+{{- $spillStorage := ($storage.spillStorage | default dict) -}}
```
richardkchapman, line 128: Is this used?

helm/hpcc/templates/_helpers.tpl
ghalliday: yes it should

helm/hpcc/templates/_helpers.tpl
ghalliday: should be deleted.  It was an experiment

helm/hpcc/templates/_helpers.tpl
ghalliday: I'll check and clean up.

ghalliday:
@AttilaVamos any idea what these smoke test errors mean?

ghalliday:
@richardkchapman fixed the helm issues.

helm/storage.rst
```diff
@@ -0,0 +1,97 @@
...
+
+  name: hpcc-<class>-plane
+  prefix: /var/lib/HPCCSystems/hpcc<class>
+  pvc: storage.<class>Storage.pvc
+
+If no pvc was defined then an implicit pvc will be used.
+
+Notes
+=====
+I suggest replication is only support if explicitly defined using the storage planes, or if it is deduced from the group information on a bare metal system.
```
jakesmith, line 92: typo: supported

helm/storage.rst
```diff
@@ -0,0 +1,97 @@
...
+
+In the future we may want to restrict which storage planes
+I suggest we deprecate the following options:
+global.defaultDataPath
+global.defaultMirrorPath
```
jakesmith, line 97: I wonder if we can issue a warning that deprecated in the schema, suggesting a plane should be defined if non-defaults are used.

jakesmith:
@ghalliday - looks good, a few minor comments.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -101,36 +118,161 @@ Add ConfigMap volume for a component
...
 {{- define "hpcc.addDataVolumeMount" -}}
+{{- /*Create local variables which always exist to avoid having to check if intermediate key values exist*/ -}}
+{{- $storage := (.root.Values.storage | default dict) -}}
+{{- $planes := ($storage.planes | default list) -}}
+{{- $dataStorage := ($storage.dataStorage | default dict) -}}
+{{- $daliStorage := ($storage.daliStorage | default dict) -}}
+{{- $dllStorage := ($storage.dllStorage | default dict) -}}
+{{- $daliStoragePlane := ($daliStorage.plane | default "") -}}
+{{- $dllStoragePlane := ($dllStorage.plane | default "") -}}
+{{- range $plane := $planes -}}
```
jakesmith, line 132: @ghalliday - was there the idea that a Thor could have a list of accessible storage planes, by default all,
such that queries running on that Thor were only allowed/mounted those SP's, so that if a configuration wanted to, they could isolate SP's ?

ghalliday: The current implementation is that thor/hthor/roxie has access to all storage planes (except dali/dlls).  Each engine has a default plane, but it can be overwritten by using the CLUSTER() attribute on an output.  (That side is currently a work in progress.)
We may want to restrict the planes somehow later, but that can be a future refinement.

helm/storage.rst
ghalliday: I could add a warning entry in the notes.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -101,36 +118,161 @@ Add ConfigMap volume for a component
...
+ {{- if $plane.pvc -}}
+  {{- if and (ne $plane.name $daliStoragePlane) (ne $plane.name $dllStoragePlane) -}}
+   {{- $num := int ( $plane.numDevices | default 1 ) -}}
+   {{- if le $num 1 }}
+- name: {{ $plane.name }}-pv
+  mountPath: {{ $plane.prefix | quote }}
+   {{- else }}
+    {{- range $elem := until $num }}
```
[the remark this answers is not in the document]
ghalliday: Thanks, I didn't know about that.  (I think I actually need untilStep 1 (add $num 1) 1
Actually seq $num is even better... and more intuitive.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -57,15 +66,23 @@ esp:
...
+{{- /* Add implicit planes if data or spill storage plane not specified*/ -}}
+{{- if not $dataStorage.plane }}
+  - name: hpcc-data-plane
+    mount: {{ .Values.global.defaultDataPath | default "/var/lib/HPCCSystems/hpcc-data" | quote }}
 {{- end }}
-{{- if .Values.global.defaultMirrorPath }}
-    mirror: {{ .Values.global.defaultMirrorPath }}
+{{- if not $spillStorage.plane }}
+  - name: hpcc-spill-plane
+    mount: {{ .Values.global.defaultSpillPath | default "/var/lib/HPCCSystems/hpcc-spill" | quote }}
```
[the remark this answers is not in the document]
ghalliday: I think that might be something worth doing once the code is all using the storage planes.

dali/base/dadfs.cpp
```diff
@@ -10006,13 +10005,88 @@ class CInitGroups
...
+
+                //Two main type of storage plane - with a host group (bare metal) and without.
+                IPropertyTree *existingGroup = queryExistingGroup(gname);
+                const char * hosts = plane.queryProp("@hosts");
+                const char * prefix = plane.queryProp("@prefix");
+                if (hosts)
+                {
+                    IPropertyTree *existingClusterGroup = queryExistingGroup(gname);
+                    if (!existingClusterGroup)
+                        UNIMPLEMENTED_X("Bare metal storage planes not yet supported");
```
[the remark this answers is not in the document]
ghalliday: yes.  I think that will be in a second phase once nas-style planes are working.

dali/base/dadfs.cpp
```diff
@@ -10006,13 +10005,88 @@ class CInitGroups
             }
         }
     }
+
+    IPropertyTree * createStorageGroup(const char * name, size32_t size, const char * path)
+    {
+        std::vector<std::string> hosts;
+        for (unsigned n=0; n<size; n++)
+            hosts.push_back("localhost");
```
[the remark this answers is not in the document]
ghalliday: neat, I didn't know that.

helm/hpcc/templates/_helpers.tpl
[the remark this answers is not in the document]
ghalliday: Hmm.  apparently seq isn't defined.  And you need to add a cast on the computed argument.

jakesmith:
@ghalliday - looks good, good to squash/merge.

jakesmith (approved):

PR #13917, opened 2020-06-30
HPCC-23925 Enable regex timings by default

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/ecl/hqlcpp/hqlcpp.cpp
+++ b/ecl/hqlcpp/hqlcpp.cpp
@@ -1860,7 +1860,6 @@ void HqlCppTranslator::cacheOptions()
         DebugOption(options.generateDiskFormats, "generateDiskFormats", false),
         DebugOption(options.maxOptimizeSize, "maxOptimizeSize", 5),             // Remove the overhead from very small functions e.g. function prolog
         DebugOption(options.minNoOptimizeSize, "minNoOptimizeSize", 10000),     // functions larger than this will take a long time to optimize, better to not try
-        DebugOption(options.timeRegex, "timeRegex", false),
     };
 
     //get options values from workunit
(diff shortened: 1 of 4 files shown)


ghalliday:
Full regression testing not yet complete - a couple of examples grow by a surprising amount.

ghalliday:
@richardkchapman please review.

Note, I ended up implementing this is a different way because the previous implementation had problems
- It was generating a unique entry for each regex pattern, but some functions use > 1,000 so it would have overloaded the stats
- It was previously timing the regexfind and evaluating the arguments passed to regex find.
- it wasn't timing regexreplace.

Instead it is implemented by adding a new optional parameter to the time attribute on an external function which allows you to override the text associated with it in the stats.

richardkchapman:
What's the overhead of these TIME attributes (I can't remember how it works)?

richardkchapman:
@ghalliday One question

ghalliday:
Overhead is
once:
i) time to register the timer.  Fairly low, but not indsignificant

Then for each timing 2 virtual calls, two calls to get_cycles_now(), an atomic increment or two.
I'm guessing <100 cycles

ghalliday:
I could add an option to disable function timings.

richardkchapman:
Should we also be timing the regex compile calls?

ghalliday:
We could (it would be simple).  I didn't because I didn't want to add more overhead.  I was also wondering about moving static patterns to global constants - which would make them hard to time.  It would potentially be significant if they were used in roxie queries.

richardkchapman:
The regex compiles SHOULD be less frequent than the regex calls - so the timing overhead would be less. And if they are not, that's information we would want to know?

Global constants for static patterns sounds like a reasonable idea but I don't think that should stop us timing them when they are not (and ideally not even when they are).

PR #13921, opened 2020-07-01
HPCC-24187 Parallel load roxie queries

richardkchapman:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [ ] Security
  - [x] Thread-safety
  - [ ] Cloud-compatibility
  - [x] Premature optimization

## Smoketest:

### DIFF
--- a/roxie/ccd/ccdquery.cpp
+++ b/roxie/ccd/ccdquery.cpp
@@ -27,6 +27,9 @@
 
 #include "thorplugin.hpp"
 
+#include <thread>
+#include <mutex>
+
 void ActivityArray::append(IActivityFactory &cur)
 {
     hash.setValue(cur.queryId(), activities.ordinality());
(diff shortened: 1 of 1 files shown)


richardkchapman:
@ghalliday Please review

roxie/ccd/ccdquery.cpp
```diff
@@ -1101,6 +1127,34 @@ class CQueryFactory : implements IQueryFactory, implements IResourceContext, pub
...
+                }
+                addToMap();  // Publishes for slaves to see
+            }
+            catch (IException *E)
+            {
+                EXCLOG(E);
+                e.setown(E);
+            }
+        });
+        if (e)
```
ghalliday, line 1154: should possibly move e so it is a member called initException, or similar.

roxie/ccd/ccdquery.cpp
```diff
@@ -1696,33 +1789,35 @@ static void checkWorkunitVersionConsistency(const IQueryDll *dll)
...
+        else
+        {
+            if (dll && !selfTestMode)
+            {
+                checkWorkunitVersionConsistency(dll);
+                Owned<ISharedOnceContext> sharedOnceContext;
+                IPropertyTree *workflow = dll->queryWorkUnit()->queryWorkflowTree();
+                if (workflow && workflow->hasProp("Item[@mode='once']"))
+                    sharedOnceContext.setown(new CSharedOnceContext);
+                ret.setown(new CRoxieServerQueryFactory(id, dll, dynamic_cast<const IRoxiePackage&>(package), hashValue, sharedOnceContext, isDynamic));
```
ghalliday, line 1813: Not directly connected to this PR, but this dynamic_cast is unnecessary (also line 1816).  If it isn't optimized away it is a waste...

roxie/ccd/ccdquery.cpp
```diff
@@ -106,16 +128,18 @@ class CQueryDll : implements IQueryDll, public CInterface
     }
     static const CQueryDll *getQueryDll(const char *dllName, bool isExe)
     {
-        CriticalBlock b(dllCacheLock);
-        CQueryDll *dll = dllCache.getValue(dllName);
-        if (dll && dll->isAliveAndLink())
-            return dll;
-        else
+        Owned<CQueryDll> dll;
```
ghalliday, line 131: Using an Owned here isn't correct:  I think you need to go back to using a pointer.

Two threads get to this point with an object from the cache.
Third thread releases the last reference to the object and blocks on the crtical section in beforeDispose
1st thread gets here, sees the object is dead.  That will cause dll to be released  - since it is in an owned
2nd thread gets here, adn object now appears to be alive since link count has wrapped below the dead marker
3rd thread disposes of the object

PR #13955, opened 2020-07-08
HPCC-24018 Add IncludeServiceNames option into WUInfo

wangkx:
Default to false.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/services/ws_workunits/ws_workunitsHelpers.cpp
+++ b/esp/services/ws_workunits/ws_workunitsHelpers.cpp
@@ -898,6 +898,27 @@ void WsWuInfo::getGraphTimingData(IArrayOf<IConstECLTimingData> &timingData)
         it->playProperties(visitor);
 }
 
+void WsWuInfo::getServiceNames(IEspECLWorkunit &info, unsigned long flags)
+{
+    if (!(flags & WUINFO_IncludeServiceNames))
+        return;
+
+    StringArray serviceNames;
+    WuScopeFilter filter;
+    filter.addScopeType("activity");
+    filter.addOutputAttribute(WaServiceName);
+    filter.addRequiredAttr(WaServiceName);
+    filter.finishedFilter();
+    Owned<IConstWUScopeIterator> it = &cw->getScopeIterator(filter);
+    ForEach(*it)
+    {
+        StringBuffer serviceName;
+        const char *value = it->queryAttribute(WaServiceName, serviceName);
+        if (!isEmptyString(value))
+            serviceNames.append(value);
+    }
+    info.setServiceNames(serviceNames);
+}
 
 void WsWuInfo::getEventScheduleFlag(IEspECLWorkunit &info)
 {
@@ -1056,6 +1077,7 @@ void WsWuInfo::getInfo(IEspECLWorkunit &info, unsigned long flags)
     getDebugValues(info, flags);
     getApplicationValues(info, flags);
     getWorkflow(info, flags);
+    getServiceNames(info, flags);
 }
 
 unsigned WsWuInfo::getWorkunitThorLogInfo(IArrayOf<IEspECLHelpFile>& helpers, IEspECLWorkunit &info, unsigned long flags, unsigned& helpersCount)
--- a/esp/services/ws_workunits/ws_workunitsHelpers.hpp
+++ b/esp/services/ws_workunits/ws_workunitsHelpers.hpp
@@ -134,6 +134,7 @@ private:
 #define WUINFO_IncludeHelpers           0x8000
 #define WUINFO_IncludeAllowedClusters   0x10000
 #define WUINFO_IncludeTotalClusterTime  0x20000
+#define WUINFO_IncludeServiceNames      0x40000
 #define WUINFO_All                      0xFFFFFFFF
 
 class WsWuInfo
@@ -199,6 +200,7 @@ public:
     bool getResultEclSchemas(IConstWUResult &r, IArrayOf<IEspECLSchemaItem>& schemas);
     void getResult(IConstWUResult &r, IArrayOf<IEspECLResult>& results, unsigned long flags);
     void getStats(const WuScopeFilter & filter, const StatisticsFilter& statsFilter, bool createDescriptions, IArrayOf<IEspWUStatisticItem>& statistics);
+    void getServiceNames(IEspECLWorkunit &info, unsigned long flags);
 
     void getWorkunitEclAgentLog(const char* eclAgentInstance, const char* agentPid, MemoryBuffer& buf, const char* outFile);
     void getWorkunitThorLog(const char *processName, MemoryBuffer& buf, const char* outFile);
--- a/esp/scm/ws_workunits_req_resp.ecm
+++ b/esp/scm/ws_workunits_req_resp.ecm
@@ -352,6 +352,7 @@ ESPrequest WUInfoRequest
     [min_ver("1.66")] bool IncludeHelpers(true);
     [min_ver("1.66")] bool IncludeAllowedClusters(true);
     [min_ver("1.73")] bool IncludeTotalClusterTime(true);
+    [min_ver("1.78")] bool IncludeServiceNames(false);
     [min_ver("1.16")] bool SuppressResultSchemas(false);
     [min_ver("1.25")] string ThorSlaveIP;
 };
--- a/esp/scm/ws_workunits_struct.ecm
+++ b/esp/scm/ws_workunits_struct.ecm
@@ -438,6 +438,7 @@ ESPStruct [nil_remove] ECLWorkunit
     [min_ver("1.50")] int DebugValueCount;
     [min_ver("1.50")] int WorkflowCount;
     [min_ver("1.67")] int HelpersCount;
+    [min_ver("1.78")] ESParray<string> ServiceNames;
 };
 
 ESPStruct [nil_remove] WUECLAttribute
--- a/esp/services/ws_workunits/ws_workunitsService.cpp
+++ b/esp/services/ws_workunits/ws_workunitsService.cpp
@@ -1622,6 +1622,8 @@ bool CWsWorkunitsEx::onWUInfo(IEspContext &context, IEspWUInfoRequest &req, IEsp
                     flags|=WUINFO_IncludeAllowedClusters;
                 if (req.getIncludeTotalClusterTime())
                     flags|=WUINFO_IncludeTotalClusterTime;
+                if (req.getIncludeServiceNames())
+                    flags|=WUINFO_IncludeServiceNames;
 
                 PROGLOG("WUInfo: %s %lx", wuid.str(), flags);
 


wangkx:
@afishbeck @GordonSmith please review.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -898,6 +898,38 @@ void WsWuInfo::getGraphTimingData(IArrayOf<IConstECLTimingData> &timingData)
...
+{
+    if (!(flags & WUINFO_IncludeSOAPDatasetNode))
+        return;
+
+    IArrayOf<IConstSOAPDatasetNode> datasetNodes;
+    Owned<IConstWUGraphIterator> graphs = &cw->getGraphs(GraphTypeActivities);
+    ForEach(*graphs)
+    {
+        Owned <IPropertyTree> xgmml = graphs->query().getXGMMLTree(false);
+        Owned<IPropertyTreeIterator> iter = xgmml->getElements("//node[@label='SOAP dataset']");
```
afishbeck, line 911: @wangkx 
Technically the label is a description and could change, the real identifier is the ` <att name="_kind" value="55"/>` 
Where value is a few different possibilities.
But I realize that filtering by an att named value may be very slow.
If you can't filter by kind, the label should be stable at least for now.

Currently the label can be several values which also complicates your filtering.
"SOAP dataset", "SOAP action", "HTTP dataset"

wangkx: Could we assume that the WUInfo user knows the label and add the label to WUInfoRequest?

jakesmith: I would have thought it should be by activity kind ..
But then should this service be agnostic about what it's been asked to include, i.e. why just a particular set of kinds?
Should the service cope with a list of activity types to find? So e.g. could request CSV Write activities instead.

I'm a bit concerned about the efficiency and requirements here.
The result will be a flattened (i.e. you will no longer be able to tell which graph/subgraph they came from) list of activities found, with all their attributes.
Ultimately how is the info. consumed by the client requesting it?
@GordonSmith @wangkx

afishbeck: @jakesmith it's unfortunate that kind is an attribute of node.  Fits in with a discussion of how inefficient xgmml can be.
Is there an efficient way of filtering nodes that fall into a set of activity kinds?  SOAP and HTTP call alone have a total of 5 kinds.
I don't think this use case is about understanding the graph of the query.  It's more about knowing what services etc the query calls out to.  If the list of soap and http calls were cached somewhere more accessible that would work to.

jakesmith:
>It's more about knowing what services etc the query calls out to. If the list of soap and http calls were cached somewhere more accessible that would work too.

ok that makes sense. I guess SOAP type activities are a good proxy for knowing what services are related to a query (PIPE is another possible route).

>Is there an efficient way of filtering nodes that fall into a set of activity kinds? SOAP and HTTP call alone have a total of 5 kinds.

There isn't an efficient IPT/xpath way. Multiple IPT/xpath calls to get each one would be very inefficient.
But //node is going to end up pulling everything (and getGraphs has already done so anyway).
Walking the nodes and filtering by a set of _kind's wouldn't add much overhead.
i.e. something like:
```
Owned<IPropertyTreeIterator> iter = xgmml->getElements("//node");
ForEach(*iter)
{
    unsigned kind = iter.query().getPropInt("att[@name=\"_kind\"]/@value);
    if (kind)
    {
        const char *kindStr = getActivityText(kind);
        if (seachSet.find(kindStr) != searchSet.end())
        {
             // add
        }
    }
}


```

jakesmith:
The above would be a bit more efficient if the search set contained the kind's as unsigned's.

jakesmith:
>It's more about knowing what services etc the query calls out to.
>SOAP and HTTP call alone have a total of 5 kinds.

@afishbeck - should this request members/methods be called getServiceActivities or similar vs "SOAPDataset"* ?

afishbeck:
Yes that name would make more sense to me.

wangkx:
@afishbeck you may check my changes.

esp/services/ws_workunits/ws_workunitsHelpers.hpp
```diff
@@ -134,11 +134,14 @@ struct WsWUExceptions
 #define WUINFO_IncludeHelpers           0x8000
 #define WUINFO_IncludeAllowedClusters   0x10000
 #define WUINFO_IncludeTotalClusterTime  0x20000
-#define WUINFO_IncludeSOAPDatasetNode   0x40000
+#define WUINFO_IncludeServiceActivities 0x40000
 #define WUINFO_All                      0xFFFFFFFF
 
 class WsWuInfo
 {
+    std::set<unsigned> serviceActivities;
```
afishbeck, line 142: Very picky, but the name "serviceActivities" could be kinds or nodes.. maybe call this "serviceActivityKinds"?

afishbeck (commented):
@wangkx one fairly trivial comment, otherwise looks good.

wangkx:
@afishbeck Renamed.

esp/services/ws_workunits/ws_workunitsHelpers.hpp
```diff
@@ -134,10 +134,14 @@ struct WsWUExceptions
 #define WUINFO_IncludeHelpers           0x8000
 #define WUINFO_IncludeAllowedClusters   0x10000
 #define WUINFO_IncludeTotalClusterTime  0x20000
+#define WUINFO_IncludeServiceActivities 0x40000
 #define WUINFO_All                      0xFFFFFFFF
 
 class WsWuInfo
 {
+    std::set<unsigned> serviceActivityKinds;
```
jakesmith, line 142: @wangkx - an unordered_set is more efficient.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -898,6 +899,54 @@ void WsWuInfo::getGraphTimingData(IArrayOf<IConstECLTimingData> &timingData)
...
+void WsWuInfo::getServiceActivities(IEspECLWorkunit &info, unsigned long flags)
+{
+    if (!(flags & WUINFO_IncludeServiceActivities))
+        return;
+
+    IArrayOf<IConstServiceActivity> ServiceActivityNodes;
+    Owned<IConstWUGraphIterator> graphs = &cw->getGraphs(GraphTypeActivities);
+    ForEach(*graphs)
+    {
+        Owned <IPropertyTree> xgmml = graphs->query().getXGMMLTree(false);
```
jakesmith, line 920: trivial: extra space after Owned

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -898,6 +899,54 @@ void WsWuInfo::getGraphTimingData(IArrayOf<IConstECLTimingData> &timingData)
...
+            IPropertyTree &treeNode = iter->query();
+            unsigned kind = treeNode.getPropInt("att[@name=\"_kind\"]/@value");
+            if (kind == 0)
+                continue;
+
+            if (serviceActivityKinds.find(kind) == serviceActivityKinds.end())
+                continue;
+
+            IArrayOf<IEspNamedValue> attrs;
+            IPropertyTreeIterator* attIterator = treeNode.getElements("att");
```
jakesmith, line 933: leaks

jakesmith (changes requested):
would be worth fixing the leak.

wangkx:
@jakesmith thanks for the review. Could you please review my fixes?

jakesmith (approved):
@wangkx - looks good.

jakesmith:
Please squash.

wangkx:
@jakesmith squashed

afishbeck (approved):
@wangkx looks good

wangkx:

ghalliday:
@wangkx how does this fit in with https://track.hpccsystems.com/browse/HPCC-23970?
It looks like it is trying to solve the same problem, and the information should now be available to Gordon through an existing call.

There are a couple of issues with this PR - it doesn't work with child queries, and it is likely the graph format is going to change in the near to medium term, so it would be good to not add new code that relies on the current format.

ghalliday:
(If the information is added to this service, it should internally use the same code as wu details to extract the list.)

wangkx:
> how does this fit in with https://track.hpccsystems.com/browse/HPCC-23970? It looks like it is trying to solve the same problem, and the information should now be available to Gordon through an existing call.
@GordonSmith does the https://track.hpccsystems.com/browse/HPCC-24018 duplicate https://track.hpccsystems.com/browse/HPCC-23970?

GordonSmith:
@wangkx It looks like @ghalliday PR only attaches the information to the WU - your PR is still needed as its exposes the ESPmethod - but I suspect you should possible just read the new information rather than gathering it your self?

ghalliday:
@GordonSmith You should be able to extract it using the appropriate wudetails call.  However, a modified version of this PR is needed if you want it returned by Wuinfo.  Which do you want?

GordonSmith:
@ghalliday Maybe both - It may be a misconception, but I consider the WUDetails calls to be "heavy", WUInfo calls to be "medium" and WUQuery calls to be "light".  If you happy I can use WUDetails (with appropriate conditions via ESP) for things like this, I should probably start using it more?

ghalliday:
Anything that reads the entire graph is not really light, or medium, weight.  I think for the moment @wangkx should implement this using the getScopeIterator() functionality, but we should consider extracting the information so it can be accessed more efficiently (either some kind of cache, or cloning this information).

wangkx:
> wangkx should implement this using the getScopeIterator() functionality.

@ghalliday I see that the 'serviceName' att has been added into some graph node. How can I use the to getScopeIterator() to get the 'serviceName' att from graph node? Which WuScopeFilter should I use?

ghalliday:
@wangkx it looks like you can't currently filter the scope iterator to only those items with a particular attribute.  (You can filter by statistic.)  

I have opened a jira (HPCC-24423) and will fix it.  For the moment I would use
addOutputAttribute(WaServiceName);
and then call iter->queryAttribute(WaServiceName) - which may currently come back null until I have implemented the filter.

ghalliday:
PR released

wangkx:
@ghalliday I just created a WU with a SOAPCALL WsWorkunits.WUAction using the code from your PR. I can see the <att name="serviceName" value="WUAction"/>inside graph node, There is no 'serviceName' inside the Statistic section. Is that correct behavior?
Then, I added the following code into my WUInfo. The queryAttribute(WaServiceName) returns null. Is there anything wrong in my code?
WuScopeFilter filter("source[global],depth[1,]");
    filter.sourceFlags = SSFsearchGraphStats;
    filter.addOutputAttribute(WaServiceName);
    Owned<IConstWUScopeIterator> it = &cw->getScopeIterator(filter);
    ForEach(*it)
    {
        StringBuffer s;
        DBGLOG("####aaaaa(%s)(%s)(%s)", it->queryScope(), queryScopeTypeName(it->getScopeType()), it->queryAttribute(WaServiceName, s));
    }

wangkx:
Looks like the filter.sourceFlags = SSFsearchAll helps.

wangkx:
@ghalliday this PR has been changed to use the getScopeIterator() functionality. Please review

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -898,6 +898,28 @@ void WsWuInfo::getGraphTimingData(IArrayOf<IConstECLTimingData> &timingData)
         it->playProperties(visitor);
 }
 
+void WsWuInfo::getServiceNames(IEspECLWorkunit &info, unsigned long flags)
+{
+    if (!(flags & WUINFO_IncludeServiceNames))
+        return;
+
+    StringArray serviceNames;
+    WuScopeFilter filter("depth[1,]");
```
ghalliday, line 907: You don't really need the depth filter - it will only filter a tiny fraction of the items, and will cost more to check.

For the filter you should be able to either use the string, or specify particular attributes.  You also shouldn't need to specify the source if you have provided enough information.

So you should either use
   WuScopeFilter filter("attr[servicename]");
or
   WuScopeFilter filter;
   filter.addOutputAttribute(WaServiceName);
   filter.finishedFilter();

Once the other PR has been merged this should be:
   WuScopeFilter filter("where[servicename],attr[servicename]");
or
   WuScopeFilter filter;
   filter.addOutputAttribute(WaServiceName);
   filter.addRequiredAttribute(WaServiceName);
   filter.finishedFilter();

There should also be a call to filter.finishedFilter();

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -898,6 +898,28 @@ void WsWuInfo::getGraphTimingData(IArrayOf<IConstECLTimingData> &timingData)
...
+    filter.sourceFlags = SSFsearchGraph;
```
ghalliday, line 908: Only the public member functions should be used.  I will investigate why those member variables are not private

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -898,6 +898,28 @@ void WsWuInfo::getGraphTimingData(IArrayOf<IConstECLTimingData> &timingData)
...
+    filter.addOutputAttribute(WaServiceName);
+    Owned<IConstWUScopeIterator> it = &cw->getScopeIterator(filter);
+    ForEach(*it)
+    {
+        if (it->getScopeType() != SSTactivity)
+            continue;
+
+        StringBuffer serviceName;
+        it->queryAttribute(WaServiceName, serviceName);
```
ghalliday, line 917: This needs to be

    const char * value = it->queryAttribute(WaServiceName, serviceName);
    if (value)

There is no guarantee that the parameter passed as a scratchpad will be modfied.

common/workunit/wuattr.cpp
```diff
@@ -247,6 +247,9 @@ extern WORKUNIT_API const char * queryAttributeValue(IPropertyTree & src, WuAttr
     case WaIdRemoteSubGraph:
         value = scratchpad.clear().append(ChildGraphScopePrefix).append(value).str();
         break;
+    case WaServiceName: //Should be default:?
```
ghalliday, line 250: This should be deleted - you cannot rely on scratchpad being modified

ghalliday (commented):
@w

ghalliday:
@wangkx

esp/services/ws_workunits/ws_workunitsHelpers.cpp
wangkx: I thought that specifying a source will limit the search area. I will change.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
wangkx: Ok

esp/services/ws_workunits/ws_workunitsHelpers.cpp
wangkx: Will change

common/workunit/wuattr.cpp
wangkx: Will delete

wangkx:
@ghalliday please review my changes.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -898,6 +898,28 @@ void WsWuInfo::getGraphTimingData(IArrayOf<IConstECLTimingData> &timingData)
...
+        return;
+
+    StringArray serviceNames;
+    WuScopeFilter filter;
+    filter.addOutputAttribute(WaServiceName);
+    filter.finishedFilter();
+    Owned<IConstWUScopeIterator> it = &cw->getScopeIterator(filter);
+    ForEach(*it)
+    {
+        if (it->getScopeType() != SSTactivity)
```
ghalliday, line 913: This filter should be added to the WuScopeFilter, rather than being checked in the iterator loop:

    filter.addScopeType("activity");

I have also merged the PR that allows a restriction on the attribute, so you can also add

    filter.addRequiredAttr(WaServiceName);

ghalliday:
@wangkx sorry for the delay - I was sure I had replied.  It looks good - one further comment.

wangkx:
@ghalliday I just made the changes and rebased the commits to the latest 7.10.x. Please review.

ghalliday:
@wangkx looks good.  Please squash and I will merge.

wangkx:
@ghalliday Squashed. Please merge.

PR #13956, opened 2020-07-08
HPCC-24382 Ftslave error INTERNAL: Command send to wrong computer

mckellyln:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

## Testing:

We need to manually test this specific PR with 7.6.xx and prior 7.8 versions to see if it solves the issue correctly.

### DIFF
--- a/common/remote/rmtspawn.cpp
+++ b/common/remote/rmtspawn.cpp
@@ -447,6 +447,30 @@ void CRemoteSlave::run(int argc, char * argv[])
                         byte action;
                         msg.read(action);
 
+                        // <= 7.6.xx clients also send:
+                        //     num passwords followed by those pwds
+                        // >= 7.8.0 clients until now do not send password data
+
+                        // so next unsigned could be either num passwords or netaddr
+                        // if unsigned value <= 10 assumme it is num passwords
+                        // as IP address will surely have higher bits set and be > 10
+
+                        unsigned numPasswds = 0;
+                        size32_t origPos = msg.getPos();
+                        msg.read(numPasswds);
+                        if (numPasswds <= 10)
+                        {
+                            for (int i=0; i<numPasswds; i++)
+                            {
+                                IpAddress tip;
+                                tip.ipdeserialize(msg);
+                                StringAttr password, username;
+                                msg.read(password).read(username);
+                            }
+                        }
+                        else
+                            msg.reset(origPos);
+
                         ok = processCommand(action, masterSocket, msg, results);
                     }
                     catch (IException * e)
--- a/dali/ft/filecopy.cpp
+++ b/dali/ft/filecopy.cpp
@@ -336,6 +336,11 @@ bool FileTransferThread::performTransfer()
 
         //Send message and wait for response...
         msg.append(action);
+
+        // send 0 for password info that was in <= 7.6 versions
+        unsigned zero = 0;
+        msg.append(zero);
+
         ep.serialize(msg);
         sprayer.srcFormat.serialize(msg);
         sprayer.tgtFormat.serialize(msg);


mckellyln:
This is an initial attempt at trying to make things more compatible.
If this version dfu server launches an older version ftslave then it should be able to handle both 7.6 and 7.8 formats.
But the other way around is still an issue - ie if a <= 7.6 version dfu server launches a >= 7.8 version ftslave then the older dfu server will be expecting the additional password info that is not sent.
We could add/send an additional 0 value field but that breaks compatibility between this 7.8.x version and prior 7.8 versions.

jakesmith:
>We could add/send an additional 0 value field but that breaks compatibility between this 7.8.x version and all prior 7.8 versions.

I think that's how it should originally have been changed, and still should, to keep cross compatibility (in both directions)

Yes, it won't help previous 7.8.x builds prior to the fix, but they should be encouraged to upgrade to resolve this issue, which they'll need to do anyway.
And then they will have a 7.8 that is can spray to a 7.6 too.

mckellyln:
@jakesmith @ghalliday after discussion added 0 for password info.
But couldn't we know the version and do the correct thing in all cases ?

jakesmith:
>But couldn't we know the version and do the correct thing in all cases ?

we could if the version of the dfuserver was (always) sent to ftslave, I doubt it is.

mckellyln:
There is in rmtspawn.cpp:
```
                        buffer.read(connectVersion);
...
                        if (version == connectVersion)
```
which I need to research to see if this provides any benefit.

mckellyln:
@jakesmith please review.
@AttilaVamos can you test ?

AttilaVamos:
@mckellyln The fix seems OK. Tested locally in a mixed 
```
hpccsystems-platform 7.6.52-1Debug
Start HPCC system
Starting mydafilesrv ...       [   OK    ]   
Starting mydali ...            [   OK    ]   
Starting mydfuserver ...       [   OK    ]   
Starting myeclagent ...        [   OK    ]   
Starting myeclccserver ...     [   OK    ]   
Starting myeclscheduler ...    [   OK    ]   
Starting myesp ...             [   OK    ]   
Starting myroxie ...           [   OK    ]   
Starting mysasha ...           [   OK    ]   
Starting mytoposerver ...      [   OK    ]   
```
and 
```
hpccsystems-platform 7.8.29-closedown0Debug
Start HPCC system
Starting mydafilesrv ...       [   OK    ]   
Starting mythor ...            [   OK    ]   
```
environment with spray a file to Thor. The operation finished without any problem.

common/remote/rmtspawn.cpp
```diff
@@ -447,6 +447,21 @@ void CRemoteSlave::run(int argc, char * argv[])
...
+
+                        // so next unsigned could be either num passwords or netaddr
+                        // password info has been set to 0 for many versions
+                        // if next unsigned is not 0 then assumme it is an IP address
+                        // as IP address will surely have higher bits set and be > 0
+
+                        unsigned numPasswds = 0;
+                        size32_t origPos = msg.getPos();
+                        msg.read(numPasswds);
+                        if (numPasswds != 0)
```
mckellyln, line 462: Should I add an additional check here on the value of numPasswds and possibly assert() ?

common/remote/rmtspawn.cpp
```diff
@@ -447,6 +447,21 @@ void CRemoteSlave::run(int argc, char * argv[])
...
+                            msg.reset(origPos);
```
jakesmith, line 463: If it's non zero - implying it is actually a 7.6 that has sent passwords, we'd have to cope with deserializing them too. It would hit deserialization problems as it is I think if non-zero.

But I think we can assert that it should always be 0 - @ghalliday ?

mckellyln: I just meant we could check for values outside what is reasonable for an IP address.  But if password info sent has been 0 for a long time then I think no need here, just leave it as 0 or not 0 to know.

ghalliday:
@mckellyln I'm getting paranoid, and checked the dataland environment.  That has:

    <Domain name="localdomain" password="" username=""/>
 
So although there is no username/password I think that will mean that it could be serializing a single blank password, rather than 0 passwords.  So your original idea/code is probably better - to skip them if it looks like < 10.
One minor comment - the original code used StringAttr instead of StringBuffer as the destination for the password/usename.  I'm not sure if it makes a difference, but may as well match it.

mckellyln:
@ghalliday updated and squashed.

PR #13958, opened 2020-07-08
HPCC-24408 Allow extern "C" OPTIMIZE to be generated

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/ecl/hqlcpp/hqlcatom.cpp
+++ b/ecl/hqlcpp/hqlcatom.cpp
@@ -52,6 +52,8 @@ IAtom * dictionaryAtom;
 IAtom * eclAtom;
 IAtom * endAtom;
 IAtom * ensureCapacityAtom;
+IAtom * externAtom;
+IAtom * externCAtom;
 IAtom * fileAtom;
 IAtom * funnelAtom;
 IAtom * forceAllCheckAtom;
@@ -1469,6 +1471,8 @@ MODULE_INIT(INIT_PRIORITY_HQLATOM-1)
     MAKEATOM(ecl);
     MAKEATOM(end);
     MAKEATOM(ensureCapacity);
+    MAKEATOM(extern);
+    MAKEATOM(externC);
     MAKEATOM(file);
     MAKEATOM(forceAllCheck);
     MAKEATOM(funnel);
--- a/ecl/hqlcpp/hqlcatom.hpp
+++ b/ecl/hqlcpp/hqlcatom.hpp
@@ -50,6 +50,8 @@ extern IAtom * dictionaryAtom;
 extern IAtom * eclAtom;
 extern IAtom * endAtom;
 extern IAtom * ensureCapacityAtom;
+extern IAtom * externAtom;
+extern IAtom * externCAtom;
 extern IAtom * fileAtom;
 extern IAtom * forceAllCheckAtom;
 extern IAtom * funnelAtom;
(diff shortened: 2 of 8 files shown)


ghalliday:
@mckellyln please review.  I haven't yet found a definitive answer to which version supports the attributes on member functions.

mckellyln:
@ghalliday what is the symptom or error ?
I tried compiling a small test case with gcc/g++ 4.8.5 and it seems to honor the \_\_attribute\_\_ correctly.

mckellyln:
@ghalliday I compiled a small program with the attribute on class member functions with g++ 4.8.5 and it did not complain.
What / where are the errors from this ?

ghalliday:
There is a zip file associated with https://track.hpccsystems.com/browse/HPCC-24385  That has the c++ that failed within it.

mckellyln:
I think we can just move these macros after any extern foo declarations and it should be ok.
Is that possible ?  So instead of:
orig:
```
NOOPTIMIZE extern "C" ECL_API size32_t tfB9UB1(ARowBuilder & crSelf, const byte * src) {
```
proposed:
```
extern "C" NOOPTIMIZE ECL_API size32_t tfB9UB1(ARowBuilder & crSelf, const byte * src) {
```

mckellyln:
If that is not possible or reasonable then I am fine with this PR as it is.

ghalliday:
I will see if I can make that change.  Since there is a work around I will not merge immediately, so fix will be in 7.8.32

ghalliday:
@mckellyln, thanks for investigating the issue some more and coming up with a better fix.
I've taken a look at what would be required to do what you suggest, and it would be slightly tricky with the current code.  I suggest we merge this if you are happy with it, and I will improve the fix for 7.10.x

mckellyln:
I tried g++ 7.3.1 and see the same error:
```
g++ (GCC) 7.3.1 20180303 (Red Hat 7.3.1-5)
W20200708-114754.cpp:54659:19: error: expected unqualified-id before string constant
 NOOPTIMIZE extern "C" ECL_API size32_t tfB9UB1(ARowBuilder & crSelf, const byte * src) {
                   ^~~
W20200708-114754.cpp:82567:19: error: expected unqualified-id before string constant
 NOOPTIMIZE extern "C" ECL_API size32_t tf7G2F1(ARowBuilder & crSelf, const byte * src) {
                   ^~~
```
Did you use clang ?  I thik we need to change the version test.

mckellyln:
@ghalliday but yes, I'm good with turning this off as you have - for the appropriate versions of g++.

ghalliday:
It might be a problem for all versions of gcc - in which case I need to fix it properly asap.

ghalliday:
Looks like it is a problem on 9.2.x as well.  I will go back and implement a better fix as you suggested.

ghalliday:
@mckellyln I've reimplemented the code so it generates as you suggested.  Examples ecl/regress/(alien1/memleak4).ecl illustrate the problem.
Thanks for being thorough.

mckellyln (approved):

mckellyln:
@ghalliday I am ok with this.  I dont really know this part of the code-base, but it seems ok.
Perhaps we can put alien1.ecl or memleak14.ecl into the smoke-test regression suite ?

ghalliday:
@shamser please can you double check.

shamser (changes requested):
@ghalliday reminder: title of PR needs updating to reflect change.

shamser (approved):
@richardkchapman Reviewed.

PR #13964, opened 2020-07-10
HPCC-24398 Assign MCexception a log level based on its class

rpastrana:
- Adds helper methods to map log message class to default log level
- Utilizes helper method to assign MCexception reasonable log levels

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jlog.hpp
+++ b/system/jlib/jlog.hpp
@@ -241,15 +241,15 @@ typedef unsigned LogMsgDetail;
  * It represents the lowest logging level (detail) required to output
  * messages of the given category.
  */
-constexpr int CriticalMsgThreshold    = 1;  //Use to declare categories reporting critical events (log level => 1)
-constexpr int FatalMsgThreshold       = 1;  //Use to declare categories reporting Fatal events (log level => 1)
-constexpr int ErrMsgThreshold         = 10; //Use to declare categories reporting Err messages (log level => 10)
-constexpr int WarnMsgThreshold        = 20; //Use to declare categories reporting Warn messages (log level => 20)
-constexpr int AudMsgThreshold         = 30; //Use to declare categories reporting Aud messages (log level => 30)
-constexpr int ProgressMsgThreshold    = 50; //Use to declare categories reporting Progress messages (log level => 50)
-constexpr int InfoMsgThreshold        = 60; //Use to declare categories reporting Info messages (log level => 60)
-constexpr int DebugMsgThreshold       = 80; //Use to declare categories reporting Debug messages (log level => 80)
-constexpr int ExtraneousMsgThreshold  = 90; //Use to declare categories reporting Extraneous messages (log level => 90)
+constexpr LogMsgDetail CriticalMsgThreshold    = 1;  //Use to declare categories reporting critical events (log level => 1)
+constexpr LogMsgDetail FatalMsgThreshold       = 1;  //Use to declare categories reporting Fatal events (log level => 1)
+constexpr LogMsgDetail ErrMsgThreshold         = 10; //Use to declare categories reporting Err messages (log level => 10)
+constexpr LogMsgDetail WarnMsgThreshold        = 20; //Use to declare categories reporting Warn messages (log level => 20)
+constexpr LogMsgDetail AudMsgThreshold         = 30; //Use to declare categories reporting Aud messages (log level => 30)
+constexpr LogMsgDetail ProgressMsgThreshold    = 50; //Use to declare categories reporting Progress messages (log level => 50)
+constexpr LogMsgDetail InfoMsgThreshold        = 60; //Use to declare categories reporting Info messages (log level => 60)
+constexpr LogMsgDetail DebugMsgThreshold       = 80; //Use to declare categories reporting Debug messages (log level => 80)
+constexpr LogMsgDetail ExtraneousMsgThreshold  = 90; //Use to declare categories reporting Extraneous messages (log level => 90)
 
 // Typedef for LogMsgSysInfo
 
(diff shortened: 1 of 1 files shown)


rpastrana:
@ghalliday please review

system/jlib/jlog.hpp
```diff
@@ -829,7 +829,51 @@ constexpr LogMsgCategory MCauditInfo(MSGAUD_audit, MSGCLS_information, AudMsgThr
 constexpr LogMsgCategory MCstats(MSGAUD_operator, MSGCLS_progress, ProgressMsgThreshold);
 constexpr LogMsgCategory MCoperatorInfo(MSGAUD_operator, MSGCLS_information, InfoMsgThreshold);
 
-inline LogMsgCategory MCexception(IException * e, LogMsgClass cls = MSGCLS_error) { return LogMsgCategory((e)->errorAudience(),cls); }
+/*
+ * Function to determine log level (detail) for exceptions, based on log message class
+ */
+inline int mapClassToDefaultDetailLevel(LogMsgClass cls)
```
ghalliday, line 835: picky: return type of LogMsgDetail would be better.  Same for other function.

system/jlib/jlog.hpp
```diff
@@ -829,7 +829,51 @@ constexpr LogMsgCategory MCauditInfo(MSGAUD_audit, MSGCLS_information, AudMsgThr
...
+{
+	switch (cls)
```
ghalliday, line 837: indent: should use (4) spaces to indent rather than tabs

system/jlib/jlog.hpp
```diff
@@ -829,7 +829,51 @@ constexpr LogMsgCategory MCauditInfo(MSGAUD_audit, MSGCLS_information, AudMsgThr
...
+		case MSGCLS_information:
+			return InfoMsgThreshold;
+		case MSGCLS_progress:
+			return ProgressMsgThreshold;
+		default:
+			return DefaultDetail;
+	}
+}
+
+/*
```
ghalliday, line 855: Curious why this needs a different function.  If an exception is reported with class information (I don't think any are), why would it be given a warning level of detail?

system/jlib/jlog.hpp
```diff
@@ -829,7 +829,51 @@ constexpr LogMsgCategory MCauditInfo(MSGAUD_audit, MSGCLS_information, AudMsgThr
...
 
-inline LogMsgCategory MCexception(IException * e, LogMsgClass cls = MSGCLS_error) { return LogMsgCategory((e)->errorAudience(),cls); }
+/*
+ * Function to determine log level (detail) for exceptions, based on log message class
+ */
+inline int mapClassToDefaultDetailLevel(LogMsgClass cls)
+{
+	switch (cls)
+	{
+		case MSGCLS_disaster:
```
ghalliday, line 839: style: we don't generally indent case labels within switch statements.

ghalliday (commented):
@rpastrana a few minor comments

system/jlib/jlog.hpp
rpastrana: Agreed

system/jlib/jlog.hpp
rpastrana: There are a few instances of MCexception created w/ a WarningClass, and since callers can specify any class (via constructor or other functions such as LogMsgPrepender::report JLog.cpp L:1384) this was meant as a policy to ensure appropriate levels assigned to MCexcpetion. It prob makes more sense to default to ERRthreshold...

rpastrana:
@ghalliday please review. Note second commit might be outside of scope

system/jlib/jlog.hpp
```diff
@@ -829,7 +829,54 @@ constexpr LogMsgCategory MCauditInfo(MSGAUD_audit, MSGCLS_information, AudMsgThr
...
+        return InfoMsgThreshold;
+    case MSGCLS_progress:
+        return ProgressMsgThreshold;
+    default:
+        return DefaultDetail;
+    }
+}
+
+/*
+ * Function to determine log level (detail) for excpetion based log messages,
```
rpastrana, line 856: Spotted this spelling error and will fix once @ghalliday determines if the changes in L:244-252 are within scope or not

system/jlib/jlog.hpp
ghalliday: @rpastrana 
I'm still not sure why an exception must have a minimum level of a warning.  If the class did happen to be information (I don't see any examples) then the detail level should match.  My objection is that it seems like unnecessary code, and I'm not convinced it is correct to force a minimum detail level.

There are two unusual uses of MCexception in daftdir.cpp e.g. 
MCexception(e)(1000)
which will set the detail level to 1000.  The value of the detail level should probably be changed- they might be candidates for an information class, although a warning is probably more appropriate.

There are also several calls to that same function:
    constexpr LogMsgCategory  operator ()(unsigned newDetail) const { return LogMsgCategory(audience, msgClass, newDetail); }
which allow you to override the detail level.  You probably want another PR investigating those and making sure that they fit in with the detail levels you expect.

rpastrana: @ghalliday default log level higher than warning doesn't make sense for an exception, but if you insist, we can map the class type to log level directly. 

Although I agree with you, it seems odd to overwrite the log level to 1000, that's a judgement call by the original developer I'm not sure what criteria we could use to change those values.

rpastrana:
@ghalliday please review and merge

PR #13971, opened 2020-07-13
HPCC-24342 Sasha/others components should send their credentials when send…

RussWhitehead:
…ing requests to Dali

Add SashaUser and SashaUserPassword to Sasha configuration, send these creds
in Dali requests for authentication

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [x] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dali/sasha/saverify.cpp
+++ b/dali/sasha/saverify.cpp
@@ -199,6 +199,7 @@ public:
 class CFileCrcList
 {
     CIpTable dafilesrvips;
+    Owned<IUserDescriptor> udesc;
 public:
     bool &stopped;
     CIArrayOf<CFileCrcItem> list;
@@ -206,6 +207,10 @@ public:
     CFileCrcList(bool &_stopped)
         : stopped(_stopped)
     {
+        StringBuffer userName;
+        serverConfig->getProp("@sashaUser", userName);
+        udesc.setown(createUserDescriptor());
+        udesc->set(userName.str(), nullptr);
     }
 
     void add(RemoteFilename &filename,unsigned partno,unsigned copy,unsigned crc)
@@ -221,7 +226,7 @@ public:
 
     void verifyFile(const char *name,CDateTime *cutoff)
     {
-        Owned<IDistributedFile> file=queryDistributedFileDirectory().lookup(name,UNKNOWN_USER,false,false,false,nullptr,defaultPrivilegedUser);
+        Owned<IDistributedFile> file=queryDistributedFileDirectory().lookup(name,udesc,false,false,false,nullptr,defaultPrivilegedUser);
         if (!file)
             return;
         IPropertyTree &fileprops = file->queryAttributes();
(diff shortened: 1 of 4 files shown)


RussWhitehead:
Added docs Jira https://track.hpccsystems.com/browse/HPCC-24421

RussWhitehead:
@jakesmith Please review

dali/sasha/saverify.cpp
```diff
@@ -199,13 +200,22 @@ class CFileCrcItem: public CInterface
...
+    Owned<IUserDescriptor> udesc;
 public:
     bool &stopped;
     CIArrayOf<CFileCrcItem> list;
 
     CFileCrcList(bool &_stopped)
         : stopped(_stopped)
     {
+        StringBuffer userName;
+        serverConfig->getProp("@SashaUser", userName);
```
ghalliday, line 212: minor: we tend to use lower case letters for the start of attributes.

dali/sasha/saverify.cpp
```diff
@@ -199,13 +200,22 @@ class CFileCrcItem: public CInterface
...
+        StringBuffer password;
+        serverConfig->getProp("@SashaUserPassword", password);
```
ghalliday, line 214: I don't think we want to store passwords in the configuration files.  I think they can probably be read from the eclwatch page, and the decryption function is freely available.  Should discuss with @afishbeck on the best approach.

ghalliday (commented):
@afishbeck what are your thoughts?

dali/sasha/saverify.cpp
afishbeck: We are definitely trying to move away from this and would rather not add new cases.

In the cloud it's fairly clear, we would:
1. only let internal components connect to DALI.
2. When we can't automatically trust where the component is coming from use certificates or keys to establish trust.
3. Where we do need passwords or keys for things it would be stored as kubernetes secrets, or in a vault.

For bare metal it's not as clear.

Do DALI restrictions limit access to known endpoints?  If so then perhaps just the component/instance name without pw is enough?

If we need pw we should start treating them like secrets and storing them outside the config file where only hpcc user can access them.  

Longer term when we add vault support we can support it in bare metal as well.

ghalliday: Yes, I was thinking about it some more, and I don't think the password is needed.  Nothing else passes the password to lookup any more (since it isn't even stored), so there is no need (or benefit) for sasha to.

RussWhitehead:
@ghalliday @afishbeck Confirmed password is not required, only a valid LDAP username to be used for permission lookup. Removed password from configuration and now only send username to Dali.
Please rereview

richardkchapman:
Commit message title doesn't really make sense

RussWhitehead:
@richardkchapman I thought I fixed the commit message in this commit https://github.com/hpcc-systems/HPCC-Platform/pull/13971/commits/dcb1b17394c2c8e4cf9636441974c2324bfccb5a
Please advise what I am doing wrong

ghalliday:
@RussWhitehead it all looks good to me.  Please can you squash, also double checking target branch is correct (it probably is).

PR #13977, opened 2020-07-14
HPCC-24425 DESDL: Allow script to control in process logging agents

afishbeck:
We will be using the scriptContext for multiple new features and need it to exist for the entire transaction.

Including that the scriptContext will exist for transaction logging at the end of the transaction.

Provide a low level mechanism where the script can skip a particular log thread queue entry.
Example:
 <xsdl:set-log-option name="thread-queue-myserv-agent1" select="'skip'"/>

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/logging/loggingmanager/loggingmanager.cpp
+++ b/esp/logging/loggingmanager/loggingmanager.cpp
@@ -106,7 +106,7 @@ bool CLoggingManager::updateLog(IEspLogEntry* entry, StringBuffer& status)
     if (entry->getLogInfoTree())
         return updateLog(entry->getEspContext(), entry->getOption(), entry->getLogInfoTree(), entry->getExtraLog(), status);
 
-    return updateLog(entry->getEspContext(), entry->getOption(), entry->getUserContextTree(), entry->getUserRequestTree(),
+    return updateLog(entry->getEspContext(), entry->getOption(), entry->getUserContextTree(), entry->getUserRequestTree(), entry->getScriptValuesTree(),
         entry->getBackEndReq(), entry->getBackEndResp(), entry->getUserResp(), entry->getLogDatasets(), status);
 }
 
@@ -156,7 +156,7 @@ bool CLoggingManager::updateLog(IEspContext* espContext, const char* option, IPr
     return bRet;
 }
 
-bool CLoggingManager::updateLog(IEspContext* espContext, const char* option, IPropertyTree* userContext, IPropertyTree* userRequest,
+bool CLoggingManager::updateLog(IEspContext* espContext, const char* option, IPropertyTree* userContext, IPropertyTree* userRequest, IPropertyTree *scriptValues,
     const char* backEndReq, const char* backEndResp, const char* userResp, const char* logDatasets, StringBuffer& status)
 {
     if (!initialized)
@@ -189,6 +189,8 @@ bool CLoggingManager::updateLog(IEspContext* espContext, const char* option, IPr
         }
         Owned<IEspUpdateLogRequestWrap> req =  new CUpdateLogRequestWrap(nullptr, option, espContextTree.getClear(), LINK(userContext), LINK(userRequest),
             backEndReq, backEndResp, userResp, logDatasets);
+        if (scriptValues)
+            req->setScriptValuesTree(scriptValues);
         Owned<IEspUpdateLogResponse> resp =  createUpdateLogResponse();
         bRet = updateLog(espContext, *req, *resp, status);
     }
@@ -218,6 +220,50 @@ bool CLoggingManager::updateLog(IEspContext* espContext, IEspUpdateLogRequestWra
     return bRet;
 }
 
+static bool checkEnabledLogVariant(IPropertyTree *scriptValues, const char *profile, const char *tracename, const char *group, const char *logtype)
+{
+    bool checkProfile = !isEmptyString(profile);
+    bool checkType = !isEmptyString(logtype);
+
+    if (checkProfile && (isEmptyString(group) || !strieq(profile, group)))
+    {
+        ESPLOG(LogNormal, "'%s' log entry disabled - log profile '%s' disabled", tracename, profile);
+        return false;
+    }
+    else if (checkType)
+    {
+        VStringBuffer xpath("@disable-log-type-%s", logtype);
+        if (scriptValues->getPropBool(xpath, false))
+        {
+            ESPLOG(LogNormal, "'%s' log entry disabled - log type '%s' disabled", tracename, logtype);
+            return false;
+        }
+     }
+    return true;
+}
+
+bool checkSkipThreadQueue(IPropertyTree *scriptValues, IUpdateLogThread &logthread)
+{
+    if (!scriptValues)
+        return false;
+    Linked<IEspLogAgent> agent = logthread.getLogAgent(); //badly named function get functions should link
+    if (!agent)
+        return false;
+
+    const char *profile = scriptValues->queryProp("@profile");
+    Owned<IEspLogAgentVariantIterator> variants = agent->getVariants();
+    if (isEmptyString(profile) && !variants->first())
+        return false;
+
+    ForEach(*variants)
+    {
+        const IEspLogAgentVariant& variant = variants->query();
+        if (checkEnabledLogVariant(scriptValues, profile, variant.getName(), variant.getGroup(), variant.getType()))
+            return false;
+    }
+    return true;
+}
+
 bool CLoggingManager::updateLog(IEspContext* espContext, IEspUpdateLogRequestWrap& req, IEspUpdateLogResponse& resp)
 {
     if (!initialized)
@@ -228,6 +274,7 @@ bool CLoggingManager::updateLog(IEspContext* espContext, IEspUpdateLogRequestWra
         if (espContext)
             espContext->addTraceSummaryTimeStamp(LogMin, "LMgr:startQLog");
 
+        Linked<IPropertyTree> scriptValues = req.getScriptValuesTree();
         if (oneTankFile || decoupledLogging)
         {
             Owned<CLogRequestInFile> reqInFile = new CLogRequestInFile();
@@ -253,6 +300,8 @@ bool CLoggingManager::updateLog(IEspContext* espContext, IEspUpdateLogRequestWra
                     IUpdateLogThread* loggingThread = loggingAgentThreads[x];
                     if (loggingThread->hasService(LGSTUpdateLOG))
                     {
+                        if (checkSkipThreadQueue(scriptValues, *loggingThread))
+                            continue;
                         loggingThread->queueLog(logRequest);
                     }
                 }
@@ -265,6 +314,11 @@ bool CLoggingManager::updateLog(IEspContext* espContext, IEspUpdateLogRequestWra
                 IUpdateLogThread* loggingThread = loggingAgentThreads[x];
                 if (loggingThread->hasService(LGSTUpdateLOG))
                 {
+                    //leave the fact that a script can control the thread queue as an option undocumented, naming scheme could change,
+                    //  controlling the queue is a very low level mechanism and should be frowned upon
+                    //  once scripts can communicate with the agent that should be the mechanism to skip a particular type of logging
+                    if (checkSkipThreadQueue(scriptValues, *loggingThread))
+                        continue;
                     loggingThread->queueLog(&req);
                 }
             }
--- a/esp/services/esdl_svc_engine/esdl_binding.cpp
+++ b/esp/services/esdl_svc_engine/esdl_binding.cpp
@@ -650,6 +650,26 @@ static inline bool isPublishedQuery(EsdlMethodImplType implType)
     return (implType==EsdlMethodImplRoxie || implType==EsdlMethodImplWsEcl);
 }
 
+IEsdlScriptContext* EsdlServiceImpl::checkCreateEsdlServiceScriptContext(IEspContext &context, IEsdlDefService &srvdef, IEsdlDefMethod &mthdef, IPropertyTree *tgtcfg)
+{
+    IEsdlCustomTransform *methodCrt = m_customRequestTransformMap.getValue(mthdef.queryMethodName());
+    if (!m_serviceLevelRequestTransform && !methodCrt)
+        return nullptr;
+
+    Owned<IEsdlScriptContext> scriptContext = createEsdlScriptContext(&context);
+
+    scriptContext->setAttribute(ESDLScriptCtxSection_ESDLInfo, "service", srvdef.queryName());
+    scriptContext->setAttribute(ESDLScriptCtxSection_ESDLInfo, "method", mthdef.queryMethodName());
+    scriptContext->setAttribute(ESDLScriptCtxSection_ESDLInfo, "request_type", mthdef.queryRequestType());
+    scriptContext->setAttribute(ESDLScriptCtxSection_ESDLInfo, "request", mthdef.queryRequestType());  //this could diverge from request_type in the future
+
+    if (tgtcfg)
+        scriptContext->setContent(ESDLScriptCtxSection_TargetConfig, tgtcfg);
+    if (m_oEspBindingCfg)
+        scriptContext->setContent(ESDLScriptCtxSection_BindingConfig, m_oEspBindingCfg.get());
+    return scriptContext.getClear();
+}
+
 void EsdlServiceImpl::handleServiceRequest(IEspContext &context,
                                            IEsdlDefService &srvdef,
                                            IEsdlDefMethod &mthdef,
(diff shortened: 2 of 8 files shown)


afishbeck:
@asselitx @timothyklemm @wangkx this is just a draft PR, I have not fully tested and have not tested the logging changes at all yet.
Please review.
Tim the control of log threads is low level and not meant to be the final way we control logging.  Focus on the functionality of that particular feature not whether we want to actually use it long term.  Once your log agent control is in place we can use that.
Terrence if you could try this out when you are set up I would appreciate it.
@wangkx , can you especially look at the logentry and logthread changes?

esp/logging/loggingmanager/loggingmanager.cpp
```diff
@@ -256,11 +267,17 @@ bool CLoggingManager::updateLog(IEspContext* espContext, IEspUpdateLogRequestWra
...
             for (unsigned int x = 0; x < loggingAgentThreads.size(); x++)
             {
                 IUpdateLogThread* loggingThread = loggingAgentThreads[x];
                 if (loggingThread->hasService(LGSTUpdateLOG))
                 {
+                    //leave the fact that a script can control the thread queue as an option undocumented, naming scheme could change,
+                    //  controlling the queue is a very low level mechanism and should be frowned upon
+                    //  once scripts can communicate with the agent that should be the mechanism to skip a particular type of logging
+                    if (checkSkipThreadQueue(scriptValues, loggingThread->queryControlName()))
+                        continue;
```
wangkx, line 280: Why not do the same before the line 264? We may also need a way to do the same when the Failover is enabled.

afishbeck: @wangkx Updated based on your suggestion.  I'm not that familiar with that code so perhaps when Terrence is done he can give you the config and you can modify it to test that use case?

wangkx: I like that because I will have a better config for testing.

You update looks fine. I only reviewed the logentry and logthread changes. @asselitx and @timothyklemm may review more.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -1427,39 +1452,22 @@ void EsdlServiceImpl::prepareFinalRequest(IEspContext &context,
...
-        scriptContext->setAttribute(ESDLScriptCtxSection_ESDLInfo, "service", srvdef.queryName());
-        scriptContext->setAttribute(ESDLScriptCtxSection_ESDLInfo, "method", mthdef.queryMethodName());
-        scriptContext->setAttribute(ESDLScriptCtxSection_ESDLInfo, "request_type", mthdef.queryRequestType());
-        scriptContext->setAttribute(ESDLScriptCtxSection_ESDLInfo, "request", mthdef.queryRequestType());  //this could diverge from request_type in the future
-
-        scriptContext->setContent(ESDLScriptCtxSection_TargetConfig, tgtcfg);
-        scriptContext->setContent(ESDLScriptCtxSection_BindingConfig, m_oEspBindingCfg.get());
-
-        processServiceAndMethodTransforms(scriptContext, {serviceCrt, methodCrt}, ESDLScriptCtxSection_ESDLRequest, ESDLScriptCtxSection_FinalRequest);
+        processServiceAndMethodTransforms(scriptContext, {m_serviceLevelRequestTransform, methodCrt}, ESDLScriptCtxSection_ESDLRequest, ESDLScriptCtxSection_FinalRequest);
```
timothyklemm, line 1470: This should not be called if both transforms are NULL (it wasn't - see old line 1447), or this method should be updated to recognize when all initializer list values are NULL. There's a lot of overhead involved preparing to process no transforms.

esp/logging/loggingmanager/loggingmanager.cpp
```diff
@@ -250,6 +262,11 @@ bool CLoggingManager::updateLog(IEspContext* espContext, IEspUpdateLogRequestWra
                 IUpdateLogThread* loggingThread = loggingAgentThreads[x];
                 if (loggingThread->hasService(LGSTUpdateLOG))
                 {
+                    //leave the fact that a script can control the thread queue as an option undocumented, naming scheme could change,
+                    //  controlling the queue is a very low level mechanism and should be frowned upon
+                    //  once scripts can communicate with the agent that should be the mechanism to skip a particular type of logging
```
timothyklemm, line 267: Maybe it's just semantics, but this comment rubs me the wrong way. As long as the manager is driving the individual agents (i.e., not decoupled), it seems like controlling the queue is a mechanism that should be embraced as an optimization rather than be frowned upon. I would replace this mechanism, which is very low level, with one that looks at the agent behind the thread and compares the agent's types and groups with selected profiles and options to decide if any agent type is needed. In other words, replace explicit queue control with implicit control based on the same information the agent will use to control its own behavior.

esp/logging/loggingmanager/loggingmanager.cpp
```diff
@@ -217,6 +219,15 @@ bool CLoggingManager::updateLog(IEspContext* espContext, IEspUpdateLogRequestWra
     return bRet;
 }
 
+bool checkSkipThreadQueue(IPropertyTree *scriptValues, const char *controlName)
+{
+    if (!scriptValues || !controlName || !*controlName)
+        return false;
+    VStringBuffer controlAttr("@%s", controlName);
+    const char *controlValue = scriptValues->queryProp(controlAttr);
+    return (controlValue && strieq(controlValue, "skip"));
```
asselitx, line 228: So a script would direct the manager to skip logging for a particular agent by adding a scriptValues attribute like thread-queue-<service>-<agent>="skip"

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -1427,39 +1452,22 @@ void EsdlServiceImpl::prepareFinalRequest(IEspContext &context,
     // access to them, so pull them from ourself. When they are passed in the
     // error checking has already been completed.
```
asselitx, line 1453: This comment is no longer accurate since the CRTs are no longer passed in. 
N.B. I also intend to verify that this refactoring continues to do the error checking required in all cases.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -662,15 +682,15 @@ void EsdlServiceImpl::handleServiceRequest(IEspContext &context,
...
                                            unsigned int flags)
 {
+    Owned<IEsdlScriptContext> scriptContext;
+
     const char *mthName = mthdef.queryName();
     context.addTraceSummaryValue(LogMin, "method", mthName);
     const char* srvName = srvdef.queryName();
 
     if (m_serviceLevelCrtFail)
         throw MakeStringException(-1, "%s::%s disabled due to Custom Transform errors. Review transform template in configuration.", srvdef.queryName(), mthName);
```
asselitx, line 692: You should be able to remove this error checking here. It looks like it is being performed unconditionally inside prepareFinalRequest().

system/xmllib/libxml_xpathprocessor.cpp
```diff
@@ -908,8 +908,9 @@ void copyXmlNode(IPropertyTree *tree, xmlNodePtr node)
 {
     for(xmlAttrPtr att=node->properties;att!=nullptr; att=att->next)
     {
+        VStringBuffer name("@%s", (const char *)att->name);
```
asselitx, line 911: I don't have a problem with this as-is, but wondering if it was changed to facilitate debugging?

asselitx:
Ok I've finished this round Tony.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -1427,39 +1452,22 @@ void EsdlServiceImpl::prepareFinalRequest(IEspContext &context,
...
-    }
+    if (m_serviceLevelCrtFail)
+        throw MakeStringException(-1, "%s::%s disabled due to service-level Custom Transform errors. Review transform template in configuration.", srvdef.queryName(), mthName);
 
-    if (!methodCrt)
-    {
-        StringAttr *crtErrorMessage = m_methodCRTransformErrors.getValue(mthName);
-        if (crtErrorMessage && !crtErrorMessage->isEmpty())
-            throw MakeStringException(-1, "%s::%s disabled due to method-level Custom Transform errors: %s. Review transform template in configuration.", srvdef.queryName(), mthName, crtErrorMessage->str());
+    m_serviceLevelRequestTransform;
```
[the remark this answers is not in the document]
asselitx: Move the instruction pointer ahead? :)

esp/logging/loggingmanager/loggingmanager.cpp
afishbeck: @asselitx I added set-log-option in my previous PR.. it's already in 7.10.x

esp/logging/loggingmanager/loggingmanager.cpp
afishbeck: I thought you already had a JIRA for adding that capability.  Once we have that we shouldn't need to use this queue centric approach.  And once we support decoupled logging the queue wont directly correspond to an agent anyway.  The sooner your changes are available the better.  Do we need to create an additional JIRA?

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -910,7 +936,8 @@ bool EsdlServiceImpl::handleResultLogging(IEspContext &espcontext, IPropertyTree
         entry->setBackEndReq(rawreq);
         entry->setBackEndResp(rawresp);
         entry->setLogDatasets(logdata);
-
+        if (scriptContext)
```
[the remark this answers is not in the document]
afishbeck: agreed I'll move it up.  We may get to the point where we always create the context, but I currently only do if there are service or method level request transforms.

esp/services/esdl_svc_engine/esdl_binding.cpp
afishbeck: will do.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -910,7 +936,8 @@ bool EsdlServiceImpl::handleResultLogging(IEspContext &espcontext, IPropertyTree
...
+            entry->setOwnScriptValuesTree(scriptContext->createPTreeFromSection(ESDLScriptCtxSection_Logging));
```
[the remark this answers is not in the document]
afishbeck: The scriptValues above was part of my left in debug code, it will be going away.

esp/services/esdl_svc_engine/esdl_binding.cpp
afishbeck: If there are no transforms there will currently be no scriptContext.  But I may readd the check just so the overhead doesn't reappear if in the future we create scriptContext for some other reason and there are no CRTs.

system/xmllib/libxml_xpathprocessor.cpp
afishbeck: This is code that I added in a prior PR but hadn't used yet.  It ended up turning attributes into elements.  This fixes that and copies attributes as attributes.  I wish it wasn't necessary but ptree requires the @ sign.

esp/services/esdl_svc_engine/esdl_binding.cpp
timothyklemm: My expectation has been that service, when it takes ownership of the log content blob, should create the same content structure with or without defined transforms (for any or all entry points) so the next entry point (including the log update entry point in the agent) don't have to figure out what happened before it. I would err on the side of caution and avoid the overhead.

esp/logging/loggingmanager/loggingmanager.cpp
timothyklemm: There is a JIRA to make the agent pay attention to the information. The point I've unsuccessfully tried to make is that a faster path to a solution exists. The log thread owns a log agent. The log agents that we need to control now can specify types and groups, and all agents support the interface to access this data. The long term solution is to use profile (groups) and options (types) in the agent. A short term solution can use profile and options in the exact same place you're changing to accomplish the same outcome.

The queue control is a sufficient solution until we have requirements to either disable part of an agent or use an agent in multiple profiles. The one use case I know of for one agent supporting multiple types does not apply, and I don't know of any requirement to use an agent with multiple profiles.

afishbeck: How does the log manager ask for the agents types and groups?  If it's already exposed I can look into checking it.. I just don't want to make any log agent changes or even extensive log manager changes in this PR.  If the info is already exposed let me know how to access it and I'll see what I can do.

wangkx: The agent group (not agent type) is introduced in Decoupled Logging service. It has not been implemented for the logging manager in D-ESDL side. We may implement it after this PR.

timothyklemm: virtual IEspLogAgentVariantIterator* getVariants() const = 0;
This agent interface method return an iterator of variants, which are triplets of agent name, type, and group. All agents should be responding with at least one variant (name, "", ""). The scapps migration agent can be configured with a list of type and group. Each list item maps to a variant in the iterator.

Other agents should be able to add support by updating their configuration schemas. The code to process the configuration is in the agent base class.

afishbeck: I am making the change to check group in the log manager.  I expect you will do whatever it takes to package up the scriptValues so that the various logging agents have access to them locally and remotely later.  But my change should let us control by supported group names (aka profile name).

afishbeck:
@timothyklemm as we discussed offline, I've made the changes we've discussed but can't test them.  Please test and debug and if you need to make changes share them with me.

I've now also rebased and targeted master.

esp/logging/loggingmanager/loggingmanager.cpp
```diff
@@ -217,6 +219,62 @@ bool CLoggingManager::updateLog(IEspContext* espContext, IEspUpdateLogRequestWra
...
+    //if an entry has an enabled group but no type that entry is enabled
+    if (!logtype || !*logtype)
+        return true;
+    typeEntryCount++;
+    //if our type isn't disabled then this entry is enabled
+    VStringBuffer xpath("@disable-log-type-%s", logtype);
+    return (!scriptValues->getPropBool(xpath, false));
+}
+
+bool checkSkipThreadQueue(IPropertyTree *scriptValues, IUpdateLogThread &logthread)
```
asselitx, line 242: It may be very helpful in this function (and checkEnabledLogEntry) to print a message (when at high log levels) when a log agent is skipped and for what reason.

afishbeck:
@asselitx updated please retest when you return
@timothyklemm updated based on our discussion

esp/logging/loggingmanager/loggingmanager.cpp
```diff
@@ -219,59 +219,44 @@ bool CLoggingManager::updateLog(IEspContext* espContext, IEspUpdateLogRequestWra
...
 
-static bool checkEnabledLogEntry(IPropertyTree *scriptValues, const char *profile, const char *group, unsigned &groupEntryCount, const char *logtype, unsigned &typeEntryCount)
+static bool checkEnabledLogVariant(IPropertyTree *scriptValues, const char *profile, const char *tracename, const char *group, const char *logtype)
 {
-    if ((!group || *group) && (!logtype || !*logtype)) //get this special case out of the way
+    bool checkProfile = !isEmptyString(profile);
+    bool checkType = !isEmptyString(logtype);
+
+    if (checkProfile && (isEmptyString(group) || !strieq(profile, group)))
         return false;
```
timothyklemm, line 228: With a log message for entry disabled by type, it seems incomplete to not log an entry disabled by profile. The log won't be reliable for debugging if only some disabled entries are included.

afishbeck: true, I'll add one.

esp/services/esdl_svc_engine/esdl_binding.cpp
afishbeck: (added a comment about why it's also here)

afishbeck:
@wangkx @asselitx @timothyklemm I've converted this to a real PR.  Please do a final review / signoff.
@timothyklemm Before Terrence's test I made that one last change regarding logging when a set-log-profile causes a variation to be disabled.

asselitx:
Changes look good to me.

wangkx:
Looks fine for me.

PR #13980, opened 2020-07-15
HPCC-21218 Add support for capping size of esp log files

shamser:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jlog.cpp
+++ b/system/jlib/jlog.cpp
@@ -1002,14 +1002,24 @@ void FileLogMsgHandlerXML::addToPTree(IPropertyTree * tree) const
 }
 
 // RollingFileLogMsgHandler
+#define MIN_LOGFILE_SIZE_LIMIT 10000
+#define LOG_LINE_SIZE_ESTIMATE 80
 
-RollingFileLogMsgHandler::RollingFileLogMsgHandler(const char * _filebase, const char * _fileextn, unsigned _fields, bool _append, bool _flushes, const char *initialName, const char *_alias, bool daily)
-  : handle(0), messageFields(_fields), alias(_alias), filebase(_filebase), fileextn(_fileextn), append(_append), flushes(_flushes)
+RollingFileLogMsgHandler::RollingFileLogMsgHandler(const char * _filebase, const char * _fileextn, unsigned _fields, bool _append, bool _flushes, const char *initialName, const char *_alias, bool daily, long _maxLogFileSize)
+  : handle(0), messageFields(_fields), alias(_alias), filebase(_filebase), fileextn(_fileextn), append(_append), flushes(_flushes), maxLogFileSize(_maxLogFileSize)
 {
+    if (_maxLogFileSize)
+    {
+        if (_maxLogFileSize < MIN_LOGFILE_SIZE_LIMIT)                 // Setting the cap too low, doesn't work well
+            maxLogFileSize = MIN_LOGFILE_SIZE_LIMIT;
+        maxLogFileSize = _maxLogFileSize - (LOG_LINE_SIZE_ESTIMATE*2); // Trying to keep log file size below capped
+    };
+
     time_t tNow;
     time(&tNow);
     localtime_r(&tNow, &startTime);
     doRollover(daily, initialName);
+    checkRollover();
 }
 
 RollingFileLogMsgHandler::~RollingFileLogMsgHandler()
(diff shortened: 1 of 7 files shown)


shamser:
@afishbeck Please can you review.

system/jlib/jlog.cpp
```diff
@@ -2851,6 +2873,7 @@ class jlib_decl CComponentLogFileCreator : implements IComponentLogFileCreator,
     StringBuffer logDirSubdir;
 
     bool         rolling;
+    long         maxLogFileSize;
```
afishbeck, line 2876: Could initialize the value here.  I've been doing that for all new member variables.

system/jlib/jlog.cpp
```diff
@@ -1047,19 +1055,33 @@ void RollingFileLogMsgHandler::addToPTree(IPropertyTree * tree) const
...
+        if (linesLoggedSinceSizeChecked > sizeCheckInterval)
+        {
+            long fsize = ftell(handle);
+            if ((fsize==-1 && errno==EOVERFLOW) || (fsize >= maxLogFileSize))
+            {
+                localtime_r(&tNow, &startTime);
+                doRollover(false);
+            }
+            else
+                sizeCheckInterval = (maxLogFileSize - fsize) / 160 + 1;
```
afishbeck, line 1081: I can see that toward the end of the maxLogFileSize you would want to start checking more frequently, but I'm not sure the interval has to shrink each time as we go along.  Depends a bit on the size of each log line I guess.

afishbeck (commented):
@shamser a couple minor comments otherwise looks good.

system/jlib/jlog.cpp
shamser: I could do a fixed interval initially and then shrink when it get near the limit.  However, the recalculation is quick and I didn't think it was worth complicating the code.  Do you think it's worth changing this?

system/jlib/jlog.cpp
shamser: I've done this.  Note that I have had to move the variable to make it the last member variable, otherwise the compiler complains.

shamser:
@afishbeck I've made one change following the review. I'd prefer to leave the recalculating the interval as it is -please see my comment on this.  

Note: there is an additional change to address the race-condition where  multiple threads check the file size at the same time.

afishbeck:
@shamser yes it was just a comment that you didn't need to increase the rate constantly, but the current way works fine and I don't mind it staying how it is.

shamser:
@ghalliday I believe this is ready to merge.

system/jlib/jlog.cpp
```diff
@@ -1003,13 +1003,21 @@ void FileLogMsgHandlerXML::addToPTree(IPropertyTree * tree) const
...
 // RollingFileLogMsgHandler
 
-RollingFileLogMsgHandler::RollingFileLogMsgHandler(const char * _filebase, const char * _fileextn, unsigned _fields, bool _append, bool _flushes, const char *initialName, const char *_alias, bool daily)
-  : handle(0), messageFields(_fields), alias(_alias), filebase(_filebase), fileextn(_fileextn), append(_append), flushes(_flushes)
+RollingFileLogMsgHandler::RollingFileLogMsgHandler(const char * _filebase, const char * _fileextn, unsigned _fields, bool _append, bool _flushes, const char *initialName, const char *_alias, bool daily, long _maxLogFileSize)
+  : handle(0), messageFields(_fields), alias(_alias), filebase(_filebase), fileextn(_fileextn), append(_append), flushes(_flushes), maxLogFileSize(_maxLogFileSize)
 {
+    if (_maxLogFileSize)
+    {
+        if (_maxLogFileSize < 10000)                 // Setting the cap too low, doesn't work well
```
ghalliday, line 1011: It would be better if the constants 10000, 150, 160 had symbolic names e.g. minLogFileLimit, logLineEstimate

system/jlib/jlog.ipp
```diff
@@ -633,6 +634,9 @@ protected:
     mutable CriticalSection   crit;
     mutable struct tm         startTime;
     bool printHeader = true;
+    long                      maxLogFileSize = 0;
+    long                      linesLoggedSinceSizeChecked = 0; // How lines logged since last size check for rollover
+    long                      sizeCheckInterval = 0;
```
ghalliday, line 639: picky: use of long variables is a bit unusual, and sizeCheckInterval isn't very descriptive.

ghalliday (commented):
@shamser looks ok.  A couple of suggestions.  Please squash so it can be merged.
Also, what is the correct target version, and is there a jira to add documentation?

ghalliday:
@shamser cherry-picked back to 7.10.x

PR #13987, opened 2020-07-20
HPCC-24445 Validate sort connections

jakesmith:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/thorlcr/msort/tsortl.cpp
+++ b/thorlcr/msort/tsortl.cpp
@@ -65,14 +65,22 @@ struct TransferStreamHeader
 {
     rowcount_t numrecs;
     rowcount_t pos;
-    size32_t recsize;
     unsigned id;
-    TransferStreamHeader(rowcount_t _pos, rowcount_t _numrecs, unsigned _recsize, unsigned _id)
-        : pos(_pos), numrecs(_numrecs), recsize(_recsize), id(_id)
+    unsigned crc = 0;
+    TransferStreamHeader(rowcount_t _pos, rowcount_t _numrecs, unsigned _id)
+        : pos(_pos), numrecs(_numrecs), id(_id)
     {
+        crc = getCrc();
     }
     TransferStreamHeader() {}
-    void winrev() { _WINREV(pos); _WINREV(numrecs); _WINREV(recsize);  _WINREV(id); }
+    void winrev() { _WINREV(pos); _WINREV(numrecs); _WINREV(id); _WINREV(crc); }
+    unsigned getCrc() const
+    {
+        unsigned retCrc = crc32((const char *)&numrecs, sizeof(numrecs), 0);
+        retCrc = crc32((const char *)&pos, sizeof(pos), retCrc);
+        retCrc = crc32((const char *)&id, sizeof(id), retCrc);
+        return retCrc;
+    }
 };
 
 
@@ -281,7 +289,7 @@ public:
 IRowStream *ConnectMergeRead(unsigned id, IThorRowInterfaces *rowif,SocketEndpoint &nodeaddr,rowcount_t startrec,rowcount_t numrecs)
 {
     Owned<ISocket> socket = DoConnect(nodeaddr);
-    TransferStreamHeader hdr(startrec,numrecs,0,id);
+    TransferStreamHeader hdr(startrec, numrecs, id);
 #ifdef _FULL_TRACE
     StringBuffer s;
     nodeaddr.getUrlStr(s);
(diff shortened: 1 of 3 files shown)


jakesmith:
@mckellyln - please review

thorlcr/msort/tsorts1.cpp
```diff
@@ -336,23 +336,33 @@ protected: friend class CSortMerge;
...
                 }
-                catch (IJSOCK_Exception *e) { // retry if failed
-                    PrintExceptionLog(e,"WARNING: Exception(ConnectMergeWrite)");
+                catch (IJSOCK_Exception *e) // retry if failed
+                {
+                    PrintExceptionLog(e, "WARNING: Exception(ConnectMergeWrite)");
                     if (--numretries==0)
-                        throw e;
+                        throw;
+                    e->Release();
```
mckellyln, line 358: Should the socket be closed here before continuing ?

jakesmith: yes I think the old exception handling and the new, both leak the ISocket.
They should close and release it before continuing.

mckellyln (approved):

mckellyln:
@jakesmith looks good.
Perhasps could apply to 7.8.x as well ?

jakesmith:
yes I wondered what would be best target, perhaps even 7.6.x

thorlcr/msort/tsortl.cpp
```diff
@@ -298,6 +306,12 @@ ISocketRowWriter *ConnectMergeWrite(IThorRowInterfaces *rowif,ISocket *socket,si
     TransferStreamHeader hdr;
     socket->read(&hdr,sizeof(hdr));
```
richardkchapman, line 307: What happens if less than sizeof(hdr) bytes are available?

jakesmith: not new, but it would block indefinitely until/if ever, it got sizeof(hdr)
It could be changed to readtms, and specify size(hdr) as max, and timeout if gets less over a short period.

mckellyln: You would want min == max == sizeof(hdr) with a timeout.
Even if sizeof(hdr) is small you cannot always guarantee all the msg arrives at once.

mckellyln: You could use min == size of the crc if it were the first item of hdr and then continue on with an addl read if ok and don't have it all.

jakesmith: To guard against a 'rogue' client connecting and sending less than sizeof(hdr),
you want min to be 1, and loop until read sizeof(hdr), but timeout once started reading.
If it's a genuine client, it should send all swiftly once there is anything on the wire.

If it's a foreign client sending a single byte, then it should timeout after a short time, and reject it.

mckellyln: Yes, a loop until done is the ideal way.

jakesmith:
@richardkchapman @mckellyln - added guard against rogue client connecting/blocking/sending 0 or a few bytes.

jakesmith:
As this is not new/an ancient issue, after discussion with Richard, retargeted to master.

@richardkchapman @mckellyln - please review

mckellyln:
very minor:
If less than sizeof(hdr) bytes are received the exception would probably be either timeout or close which is fine, but it is different from the new TE_InvalidSortConnect msg.
read variable should perhaps be something like amtRead.
But all looks good.

jakesmith:
> be either timeout or close which is fine, but it is different from the new TE_InvalidSortConnect msg.

yes, but the pre-existing code already handled those exceptions (i.e. IJSOCK_Exception's).
This scenario has been tested btw.

thorlcr/msort/tsortl.cpp
```diff
@@ -296,8 +304,30 @@ IRowStream *ConnectMergeRead(unsigned id, IThorRowInterfaces *rowif,SocketEndpoi
 ISocketRowWriter *ConnectMergeWrite(IThorRowInterfaces *rowif,ISocket *socket,size32_t bufsize,rowcount_t &startrec,rowcount_t &numrecs)
 {
     TransferStreamHeader hdr;
-    socket->read(&hdr,sizeof(hdr));
+    unsigned remaining = sizeof(hdr);
+    byte *dst = (byte *)&hdr;
+
+    /*
+     * A client has connected at this stage, the hdr should be sent switfly.
+     * A generous 1 minute timeout between reads, if longer, timeout exception,
```
richardkchapman, line 312: Not sure what circumstances data is sent here - but might that 1 minute cause issues ever?

jakesmith: There is a 5 minute connect timeout on the connection that writes the header,
so if there were a series of these rogue clients blocking for several minutes, getting in quickly before the Thor header write connnect, it could cause problems.
So this won't solve some DoS type scenarios.

richardkchapman: Is it just the header read when first connecting, that we have that timeout on, or every read of data?

I'm more concerned about breaking existing valid scenarios than not catching all hypothetical DoS attacks here...

jakesmith: This is just for the header 1st send when sort connects.

richardkchapman:

ghalliday (commented):
@richardkchapman looks good.  One trivial typo.

jakesmith:
@richardkchapman - corrected typo' and squashed.

PR #14026, opened 2020-07-27
HPCC-17585 SOAPCALL support for persistent outgoing connections

mayx:
- Add the persistent connection source files to the thorhelper project
- Add logic to SOAPCALL/HTTPCALL to Keep and reuse a connection when
  conditions are met
- Add a map in persistent connection implementation to keep track
  of available connections for endpoints to make it faster to locate an
  available connection for an endpoint
- Read from configuration entries to turn on and off this feature, and to
  control the max number of reuses of a connection.
- Add protocol parameter to differentiate different types of connections,
  especially TLS and regular to avoid misuses and exploits.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Cloud-compatibility
  - [x] Premature optimization

## Smoketest:

### DIFF
--- a/common/thorhelper/CMakeLists.txt
+++ b/common/thorhelper/CMakeLists.txt
@@ -68,6 +68,8 @@ set (    SRCS
          thorrparse.ipp
          thorsort.cpp
          thorsort.hpp
+         persistent.cpp
+         persistent.hpp
          
          roxiedebug.hpp
          roxiedebug.ipp
@@ -80,24 +82,24 @@ set (    SRCS
 include_directories ( 
          ${CMAKE_BINARY_DIR}
          ${CMAKE_BINARY_DIR}/oss
-         ./../../system/security/securesocket 
-         ./../../common/remote 
-         ./../../system/jhtree 
-         ./../../rtl/eclrtl 
-         ./../../rtl/nbcd 
-         ./../../system/include 
-         ./../../system/mp
-         ./../../system/jlib
-         ./../../system/security/zcrypt
-         ./../../dali/base
-         ./../deftype 
-         ./../workunit
-         ./../../rtl/include 
-         ./../../ecl/hql
-         ./../../roxie/roxiemem
-         ./../../testing/unittests
-         ./../../system/tbb_sm/tbb/include
-         ./../../system/security/shared
+         ${HPCC_SOURCE_DIR}/system/security/securesocket
+         ${HPCC_SOURCE_DIR}/common/remote
+         ${HPCC_SOURCE_DIR}/system/jhtree
+         ${HPCC_SOURCE_DIR}/rtl/eclrtl
+         ${HPCC_SOURCE_DIR}/rtl/nbcd
+         ${HPCC_SOURCE_DIR}/system/include
+         ${HPCC_SOURCE_DIR}/system/mp
+         ${HPCC_SOURCE_DIR}/system/jlib
+         ${HPCC_SOURCE_DIR}/system/security/zcrypt
+         ${HPCC_SOURCE_DIR}/dali/base
+         ${HPCC_SOURCE_DIR}/common/deftype
+         ${HPCC_SOURCE_DIR}/common/workunit
+         ${HPCC_SOURCE_DIR}/rtl/include
+         ${HPCC_SOURCE_DIR}/ecl/hql
+         ${HPCC_SOURCE_DIR}/roxie/roxiemem
+         ${HPCC_SOURCE_DIR}/testing/unittests
+         ${HPCC_SOURCE_DIR}/system/tbb_sm/tbb/include
+         ${HPCC_SOURCE_DIR}/system/security/shared
          ${HPCC_SOURCE_DIR}/fs/dafsclient
     )
 
--- a/common/thorhelper/persistent.hpp
+++ b/common/thorhelper/persistent.hpp
@@ -18,20 +18,30 @@
 #ifndef __PERSISTENT_HPP__
 #define __PERSISTENT_HPP__
 
+#ifdef THORHELPER_EXPORTS
+ #define THORHELPER_API DECL_EXPORT
+#else
+ #define THORHELPER_API DECL_IMPORT
+#endif
+
 #include "jlib.hpp"
 #include "jsocket.hpp"
 
 #define DEFAULT_MAX_PERSISTENT_IDLE_TIME 60
 #define DEFAULT_MAX_PERSISTENT_REQUESTS  100
 
-enum class PersistentLogLevel { PLogNone=0, PLogMin=1, PLogNormal=5, PLogMax=10};
-
+enum class PersistentLogLevel { PLogNone=0, PLogMin=1, PLogNormal=5, PLogMax=10 };
+enum class PersistentProtocol { ProtoTCP=0, ProtoTLS=1 };
 interface IPersistentHandler : implements IInterface
 {
-    virtual void add(ISocket* sock, SocketEndpoint* ep = nullptr) = 0;
+    // Add a new socket to the pool for reuse
+    virtual void add(ISocket* sock, SocketEndpoint* ep = nullptr, PersistentProtocol proto = PersistentProtocol::ProtoTCP) = 0;
+    // Remove a socket from the pool
     virtual void remove(ISocket* sock) = 0;
+    // Put a socket back to the pool for further reuse, or remove its record from the pool when "keep" is false
     virtual void doneUsing(ISocket* sock, bool keep, unsigned usesOverOne = 0) = 0;
-    virtual Linked<ISocket> getAvailable(SocketEndpoint* ep = nullptr, bool* pShouldClose = nullptr) = 0;
+    // Get one available socket from the pool
+    virtual ISocket* getAvailable(SocketEndpoint* ep = nullptr, bool* pShouldClose = nullptr, PersistentProtocol proto = PersistentProtocol::ProtoTCP) = 0;
     virtual void stop(bool wait) = 0;
     virtual bool inDoNotReuseList(SocketEndpoint* ep) = 0;
 };
@@ -41,6 +51,8 @@ interface IPersistentSelectNotify
     virtual bool notifySelected(ISocket *sock,unsigned selected, IPersistentHandler* handler, bool shouldClose) = 0;
 };
 
-IPersistentHandler* createPersistentHandler(IPersistentSelectNotify* notify, int maxIdleTime = DEFAULT_MAX_PERSISTENT_IDLE_TIME, int maxReqs = DEFAULT_MAX_PERSISTENT_REQUESTS, PersistentLogLevel loglevel=PersistentLogLevel::PLogMin, bool enableDoNotReuseList=false);
+inline size32_t httpVerOffset() { return 5; /*strlen("HTTP ")*/ }
+THORHELPER_API bool isHttpPersistable(const char* httpVer, const char* conHeader);
+THORHELPER_API IPersistentHandler* createPersistentHandler(IPersistentSelectNotify* notify, int maxIdleTime = DEFAULT_MAX_PERSISTENT_IDLE_TIME, int maxReqs = DEFAULT_MAX_PERSISTENT_REQUESTS, PersistentLogLevel loglevel=PersistentLogLevel::PLogMin, bool enableDoNotReuseList=false);
 
 #endif //__PERSISTENT_HPP__
--- a/common/thorhelper/thorsoapcall.cpp
+++ b/common/thorhelper/thorsoapcall.cpp
@@ -30,6 +30,7 @@
 #include "eclrtl.hpp"
 #include "roxiemem.hpp"
 #include "zcrypt.hpp"
+#include "persistent.hpp"
 
 using roxiemem::OwnedRoxieString;
 
@@ -42,6 +43,7 @@ using roxiemem::OwnedRoxieString;
 #define CONTENT_LENGTH "Content-Length: "
 #define CONTENT_ENCODING "Content-Encoding"
 #define ACCEPT_ENCODING "Accept-Encoding"
+#define CONNECTION "Connection"
 
 unsigned soapTraceLevel = 1;
 
@@ -507,6 +509,28 @@ public:
     }
 } *blacklist;
 
+static IPersistentHandler* persistentHandler = nullptr;
+static CriticalSection persistentCrit;
+static std::atomic<bool> persistentInitDone{false};
+
+void initPersistentHandler()
+{
+    CriticalBlock block(persistentCrit);
+    if (!persistentInitDone)
+    {
+#ifndef _CONTAINERIZED
+        const IProperties &conf = queryEnvironmentConf();
+        int maxPersistentRequests = conf.getPropInt("maxPersistentRequests", DEFAULT_MAX_PERSISTENT_REQUESTS);
+#else
+        const IPropertyTree& conf = queryComponentConfig();
+        int maxPersistentRequests = conf.getPropInt("@maxPersistentRequests", DEFAULT_MAX_PERSISTENT_REQUESTS);
+#endif
+        if (maxPersistentRequests != 0)
+            persistentHandler = createPersistentHandler(nullptr, DEFAULT_MAX_PERSISTENT_IDLE_TIME, maxPersistentRequests, PersistentLogLevel::PLogMin, true);
+        persistentInitDone = true;
+    }
+}
+
 MODULE_INIT(INIT_PRIORITY_STANDARD)
 {
     blacklist = new BlackLister;
@@ -518,6 +542,12 @@ MODULE_EXIT()
 {
     blacklist->stop();
     delete blacklist;
+
+    if (persistentHandler)
+    {
+        persistentHandler->stop(true);
+        ::Release(persistentHandler);
+    }
 }
 
 //=================================================================================================
@@ -644,7 +674,7 @@ interface IWSCAsyncFor: public IInterface
     virtual void checkTimeLimitExceeded(unsigned * _remainingMS) = 0;
 
     virtual void createHttpRequest(Url &url, StringBuffer &request) = 0;
-    virtual int readHttpResponse(StringBuffer &response, ISocket *socket) = 0;
+    virtual int readHttpResponse(StringBuffer &response, ISocket *socket, bool &keepAlive) = 0;
     virtual void processResponse(Url &url, StringBuffer &response, ColumnProvider * meta) = 0;
 
     virtual const char *getResponsePath() = 0;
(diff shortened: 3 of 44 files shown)


mayx:
@afishbeck please review...

esp/bindings/http/client/httpclient.cpp
```diff
@@ -189,7 +189,7 @@ CHttpClient::~CHttpClient()
         }
         else if (m_persistentHandler && !m_disableKeepAlive && m_persistable)
         {
-            m_persistentHandler->add(m_socket, &m_ep);
+            m_persistentHandler->add(m_socket, &m_ep, strieq(m_protocol.get(), "HTTPS")?PersistentProtocol::ProtoTls:PersistentProtocol::ProtoDefault);
```
afishbeck, line 192: Unless ProtoDefault is always HTTP, then is there no way of explicitly creating an HTTP entry?  I.e. if ProtoTls and ProtoDefault are both HTTPS.

common/thorhelper/thorsoapcall.cpp
```diff
@@ -2065,32 +2113,49 @@ class CWSCAsyncFor : implements IWSCAsyncFor, public CInterface, public CAsyncFo
                 {
                     checkTimeLimitExceeded(&remainingMS);
                     Url &connUrl = master->proxyUrlArray.empty() ? url : master->proxyUrlArray.item(0);
-                    socket.setown(blacklist->connect(connUrl.port, connUrl.host, master->logctx, (unsigned)master->maxRetries, master->timeoutMS, master->roxieAbortMonitor));
-                    if (stricmp(url.method, "https") == 0)
+                    ep.setown(new SocketEndpoint(connUrl.host.get(), connUrl.port));
+                    if (strieq(url.method, "https"))
+                        proto = PersistentProtocol::ProtoTls;
+                    bool shouldClose = false;
+                    Linked<ISocket> psock = persistentHandler?persistentHandler->getAvailable(ep, &shouldClose, proto):nullptr;
```
afishbeck, line 2120: I'm planning on adding support for the user specifying the client cert to use.  Can we somehow use the client cert info a as part of the look up for an available socket?  The cert info is added as part of the context inside createSecureSocket().. the change will mean there may be more than one.

common/thorhelper/thorsoapcall.cpp
```diff
@@ -507,6 +509,27 @@ class BlackLister : public CInterface, implements IThreadFactory
     }
 } *blacklist;
 
+static IPersistentHandler* persistentHandler = nullptr;
+static Mutex persistentMutex;
+static bool initDone = false;
```
afishbeck, line 514: Perhaps a more specific variable name?

esp/platform/persistent.hpp
```diff
@@ -41,6 +41,9 @@ interface IPersistentSelectNotify
     virtual bool notifySelected(ISocket *sock,unsigned selected, IPersistentHandler* handler, bool shouldClose) = 0;
 };
 
+#define HTTP_VER_SKIP 5
```
afishbeck, line 44: HTTP_VER_OFFSET?

afishbeck (commented):
@mayx some initial questions/comments.  Can review in more detail after.

esp/bindings/http/client/httpclient.cpp
mayx: For https ProtoTls should always be used. But I can see how "Default" could be misleading. Let me change it to "Regular".

common/thorhelper/thorsoapcall.cpp
mayx: OK I can change it to a more specific name like persistentInitDone

esp/platform/persistent.hpp
mayx: Makes sense.

mayx:
@afishbeck changes made per suggestions...

esp/bindings/http/client/httpclient.cpp
afishbeck: Ok.  Or TCP.

common/thorhelper/thorsoapcall.cpp
afishbeck: Well roxie is shared by many ECL developers and we have no control over how they manage their client certificates.  It may never happen but could happen.

mayx: OK, yes we should take certificate info into consideration. I guess one of us can merge first, then whatever change is needed can go with the latter merge...

esp/bindings/http/client/httpclient.cpp
mayx: TCP does sound better, let me change to that.

common/thorhelper/thorsoapcall.cpp
```diff
@@ -507,6 +509,27 @@ class BlackLister : public CInterface, implements IThreadFactory
...
+static IPersistentHandler* persistentHandler = nullptr;
+static Mutex persistentMutex;
+static bool persistentInitDone = false;
+inline void initPersistentHandler()
+{
+    synchronized block(persistentMutex);
+    if (!persistentInitDone)
+    {
+#ifndef _CONTAINERIZED
+        const IProperties &conf = queryEnvironmentConf();
```
afishbeck, line 521: These are identical except for the property name.  You could just have:
```
#ifndef _CONTAINERIZED
    const char *propname = "maxPersistentRequests";
#else
    const char *propname = "@maxPersistentRequests";
#endif
```

common/thorhelper/thorsoapcall.cpp
afishbeck: Right, let's wait on this.  I may remove custom client certificates from my first pass at SOAPCALL secrets.

afishbeck (approved):
@mayx looks fine.

afishbeck:
@richardkchapman perhaps @jakesmith should take a look when he get's back.

richardkchapman:
@jakesmith

jakesmith:
@mayx - added some comments/questions.

fairly complex code before and more so before, it could do with some comment blocks, describing the general approach, and some comments on the areas where it adds/retrieves from the tables etc.

It is really necessary/warranted to have a table of tables, top level for ep+proto to CSocketSet, then ISocket set.
i.e. could it be just a single ep+proto -> ISocket set ? A doubt a large single HT vs 2 HT lookups is going to be slower, probably faster, and code would be quite a bit simpler.
I'd consider using a std::unordered_set, for efficiency and it would reduce the code complexity.

I wondered if there is always a CPersistentInfo object per socket(?), could the availKeep map not point to it? {ep+proto}->CPersistentInfo.
So that getAavilable performed 1 lookup instead of 2 to get the info and socket.

mayx:
@jakesmith the second level of table here is to speed up deletion of a socket, otherwise have to iterate to find the socket to delete. I will change it to use std::unordered_set. I will add the CPersisInfo to the availKeeper table to avoid 2nd lookup. Also I'll add some comments and make other suggested changes. Thanks.

mayx:
@jakesmith Changes made per suggestion, please review again...

AttilaVamos:
@mayx This PR generated some core files and crashed because it consumed the whole disk volume (20 GB) assigned to this instance (Host: ip-10-20-0-35.ca-central-1.compute.internal). 
I was able  to recover a trace file:
```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  Shared<IPersistentHandler>::operator IPersistentHandler* (this=0x0) at /home/centos/smoketest/PR-14026/HPCC-Platform/common/thorhelper/./../../system/jlib/jscm.hpp:66
66          inline operator CLASS *() const             { return ptr; }

Thread 1 (Thread 0x7f19d97dc140 (LWP 57761)):
#0  Shared<IPersistentHandler>::operator IPersistentHandler* (this=0x0) at /home/centos/smoketest/PR-14026/HPCC-Platform/common/thorhelper/./../../system/jlib/jscm.hpp:66
#1  _modExit545 () at /home/centos/smoketest/PR-14026/HPCC-Platform/common/thorhelper/thorsoapcall.cpp:551
#2  0x00007f19d22e9864 in InitTable::exit (this=this@entry=0x1315060, ctx=ctx@entry=0x0) at /home/centos/smoketest/PR-14026/HPCC-Platform/system/jlib/jlib.cpp:110
#3  0x00007f19d22e9945 in ExitModuleObjects (ctx=<optimized out>) at /home/centos/smoketest/PR-14026/HPCC-Platform/system/jlib/jlib.cpp:40
#4  0x0000000000412867 in main ()
```

mayx:
@AttilaVamos thanks, I just uploaded the fix...

common/thorhelper/thorsoapcall.cpp
```diff
@@ -507,6 +509,32 @@ class BlackLister : public CInterface, implements IThreadFactory
...
+            const IProperties &conf = queryEnvironmentConf();
+            int maxPersistentRequests = conf.getPropInt("maxPersistentRequests", DEFAULT_MAX_PERSISTENT_REQUESTS);
+#else
+            IPropertyTree& conf = queryComponentConfig();
+            int maxPersistentRequests = conf.getPropInt("@maxPersistentRequests", DEFAULT_MAX_PERSISTENT_REQUESTS);
+#endif
+            if (maxPersistentRequests == 0)
+                return new PersistentHandlerWrapper(nullptr);
+            else
+                return new PersistentHandlerWrapper(createPersistentHandler(nullptr, DEFAULT_MAX_PERSISTENT_IDLE_TIME, maxPersistentRequests, PersistentLogLevel::PLogMin, true));
```
jakesmith, line 534: I don't think the Singleton is a good fit here actually, I may not have realized this code only conditionally required the object, so I can see why you've added the wrapper to workaround the requirements of the Singleton.

As queryPersistentHandler() is not called at a high frequency (only once per CWSCAsyncFor ctor), then a crit + an initialized boolean is prob. best. Can't see the previous commit now, but probably how you had it.

jakesmith:
@mayx - I think clearer now, added some mostly minor comments.
I think the CPersistentInfo should be consistently link counted.

Please keep any new changes in a separate commit for now, it helps to see prev. whilst reviewing.

mayx:
@jakesmith changes made per last review, please review again...

common/thorhelper/thorsoapcall.cpp
```diff
@@ -2243,5 +2322,7 @@ class CWSCAsyncFor : implements IWSCAsyncFor, public CInterface, public CAsyncFo
 
 IWSCAsyncFor * createWSCAsyncFor(CWSCHelper * _master, IXmlWriterExt &_xmlWriter, ConstPointerArray &_inputRows, PTreeReaderOptions _options)
 {
+    if (!persistentInitDone)
+        initPersistentHandler();
```
jakesmith, line 2326: this double check pattern isn't 100% safe as it stands, persisentInitDone needs to be an atomic.
See : https://preshing.com/20130930/double-checked-locking-is-fixed-in-cpp11/

jakesmith:
@mayx - a few final comments

esp/platform/persistent.cpp
```diff
@@ -86,9 +88,9 @@ class CAvailKeeper : implements IInterface, public CInterface
             auto iter = sset->begin();
             if (iter != sset->end())
```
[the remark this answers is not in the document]
mayx: There're 2 problems with removing empty SocketSet: (1) it has to be re-allocated again if the same ep shows up after deletion (2) to delete it, will have to iterate through the map to find it, which can potentially be expensive. Considering the unlikeliness of having huge number of used-once ep's, I tend to think it's better not to clean up. What do you think @jakesmith ?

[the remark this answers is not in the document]
mayx: 2nd issue can be solved by getting hold of the ep from the sockinfo before erasing, so the major concern (#2 above) is not valid. I'll do the clean up. @jakesmith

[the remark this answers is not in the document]
mayx: I did some performance test and found that removing empty socket sets seems to have a non-negligible impact on the throughput. I think that's because when the sockets are busy and get picked up right after they're put back, so the socket set gets empty all the time, so has to be constantly re-created. So for now I'm thinking not to do the removal. What do you think @jakesmith ?

esp/platform/persistent.cpp
```diff
@@ -47,33 +47,35 @@ class CPersistentInfo : implements IInterface, public CInterface
...
+struct LinkedPersistentInfoHash
+{
+    size_t operator()(const Linked<CPersistentInfo>& linkedinfo) const
+    {
+        return std::hash<CPersistentInfo*>()(linkedinfo);
+    }
+};
 
-class CAvailKeeper : implements IInterface, public CInterface
+using SocketSet = std::unordered_set<Linked<CPersistentInfo>, LinkedPersistentInfoHash>;
```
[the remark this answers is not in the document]
mayx: Changing it to Owned is causing a ref count issue. I think it may have something to do with how unordered_set copies or returns elements, and how Owned deals with it. I'll investigate more and see if I can find the issue.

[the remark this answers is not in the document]
jakesmith: Did you LINK the CPersistentInfo pointer when inserting?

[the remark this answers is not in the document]
mayx: Yes I did.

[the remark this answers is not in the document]
mayx: Found the issue. Turned out, socketset->erase(sockinfo) has to be socketset->erase(LINK(sockinfo)), which makes sense as the pointer will be turned into Owned.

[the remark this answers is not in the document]
jakesmith: hm ok, yes makes some sense, but it's ugly/not nice to need link for reasons like this.
In that case, it would be cleaner to have the elements as Linked<> items.
Performance difference is going to be v. small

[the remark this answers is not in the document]
mayx: Agree, Linked does fit better in this case.

esp/platform/persistent.cpp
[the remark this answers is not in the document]
jakesmith: right. If there aren't expected to be an ever growing number, then prob. ok.
You could alternatively, look at the size on add, and when hits some high threshold walk and remove empty ones.
@mayx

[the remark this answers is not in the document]
mayx: OK, let me do that just in case.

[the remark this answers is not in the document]
mayx: What do you think would be a good threshold, 1000?

mayx:
@jakesmith changes pushed per latest review (as a separate commit), please review again...

jakesmith:
@mayx - looks good, could be merged as is (after squashing), 2 new minor comments.

mayx:
@jakesmith can you have a quick review of the last commit again before I squash the commits?

esp/platform/persistent.cpp
```diff
@@ -140,6 +151,18 @@ class CAvailKeeper
...
     }
+    void cleanup()
+    {
+        HashIterator iter(m_avail4ep);
+        std::vector<EpSocketSetMapping*> elems;
+        for (iter.first(); iter.isValid(); iter.next())
+        {
+            if ((*m_avail4ep.mapToValue(&iter.query()))->empty())
+                elems.push_back(static_cast<EpSocketSetMapping*>(&iter.query()));
+        }
```
jakesmith, line 162: the Map should be iterator range compatible, so I think you express the above loop without the HashIterator, more simply as:
```
    for (auto &e: m_avail4ep)
    {
        if (e.getValue()->empty())
            elems.push_back(&e);
    }
```

mayx:
@jakesmith change on iterating made per suggestion, also commits are squashed. Please have a final review...

jakesmith (approved):
@mayx - looks good to merge

jakesmith:

common/thorhelper/CMakeLists.txt
```diff
@@ -68,6 +68,8 @@ set (    SRCS
          thorrparse.ipp
          thorsort.cpp
          thorsort.hpp
+         ${HPCC_SOURCE_DIR}/esp/platform/persistent.cpp
```
richardkchapman, line 71: Feels wrong for thorhelper to be including files from esp/platform.

Is there a better structure that could be considered? For example, should persistent.cpp/hpp be in thorhelper dll (and exported from there rather than copied into other places)?

mayx: thorhelper does seem to be a more common component, let me give it a try...

mayx: Would it make sense to put them under jlib instead? The function seems to be generic enough, also would help avoid adding thorhelper to the include path of quite some projects. @richardkchapman

richardkchapman: @mayx Depends what they do, but quite possibly...

mayx: I tried this option and it does make the change much easier. As it only depends on jlib so it doesn't bring other changes to jlib other than the 2 files themselves. Before I check in my changes, I just want to make sure that @mckellyln and @jakesmith are OK with it?

jakesmith: Not sure about moving it to jlib, it doesn't quite feel core enough or likely to be generically used elsewhere to me.
thorhelper (despite its misleading name) is a 2nd tier support library used by quite a few components, including esp, and this feels like it would be more at home there.
@mayx

mayx: OK I'll move it to thorhelper.

mayx: @jakesmith I moved the files to thorhelper and added thorhelper to the include path of the projects that need it. There's no change to the code itself. Please have a quick review when you have a chance...

common/thorhelper/persistent.hpp
```diff
@@ -41,6 +51,8 @@ interface IPersistentSelectNotify
     virtual bool notifySelected(ISocket *sock,unsigned selected, IPersistentHandler* handler, bool shouldClose) = 0;
 };
 
-IPersistentHandler* createPersistentHandler(IPersistentSelectNotify* notify, int maxIdleTime = DEFAULT_MAX_PERSISTENT_IDLE_TIME, int maxReqs = DEFAULT_MAX_PERSISTENT_REQUESTS, PersistentLogLevel loglevel=PersistentLogLevel::PLogMin, bool enableDoNotReuseList=false);
+#define HTTP_VER_OFFSET 5
```
jakesmith, line 54: minor: could this be moved out of the header, and a helper function that uses it be exposed instead?

jakesmith:
@mayx - 1 comment/question

common/thorhelper/persistent.hpp
mayx: @jakesmith removed HTTP_VER_OFFSET and added helper function httpVerOffset, please review again...

jakesmith:

PR #14043, opened 2020-07-30
HPCC-22103 Remove sleep from flush loop in receiveDaFsBuffer

mckellyln:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

## Testing:

QA suite / smoke-test

### DIFF
--- a/fs/dafsclient/rmtclient.cpp
+++ b/fs/dafsclient/rmtclient.cpp
@@ -388,16 +388,16 @@ void flushDaFsSocket(ISocket *socket)
     try
     {
         sendDaFsBuffer(socket, sendbuf);
-        char buf[1024];
+        char buf[16*1024];
         for (;;)
         {
-            Sleep(1000);    // breathe
             size32_t szread;
-            SOCKREADTMS(socket)(buf, 1, sizeof(buf), szread, 1000*60);
+            SOCKREADTMS(socket)(buf, 1, sizeof(buf), szread, 1000*30);
             totread += szread;
         }
     }
-    catch (IJSOCK_Exception *e) {
+    catch (IJSOCK_Exception *e)
+    {
         if (totread)
             PROGLOG("%d bytes discarded",totread);
         if (e->errorCode()!=JSOCKERR_timeout_expired)
@@ -411,17 +411,21 @@ void receiveDaFsBuffer(ISocket * socket, MemoryBuffer & tgt, unsigned numtries, 
 {
     sRFTM tm(dafsMaxReceiveTimeMs);
     size32_t gotLength = receiveDaFsBufferSize(socket, numtries,tm.timemon);
-    if (gotLength) {
+    Owned<IException> exc;
+    if (gotLength)
+    {
         size32_t origlen = tgt.length();
-        try {
-            if (gotLength>maxsz) {
+        try
+        {
+            if (gotLength>maxsz)
+            {
                 StringBuffer msg;
                 msg.appendf("receiveBuffer maximum block size exceeded %d/%d",gotLength,maxsz);
-                PrintStackReport();
                 throw createDafsException(DAFSERR_protocol_failure,msg.str());
             }
             unsigned timeout = SERVER_TIMEOUT*(numtries?numtries:1);
-            if (tm.timemon) {
+            if (tm.timemon)
+            {
                 unsigned remaining;
                 if (tm.timemon->timedout(&remaining)||(remaining<10))
                     remaining = 10;
@@ -431,29 +435,27 @@ void receiveDaFsBuffer(ISocket * socket, MemoryBuffer & tgt, unsigned numtries, 
             size32_t szread;
             SOCKREADTMS(socket)((gotLength<4000)?tgt.reserve(gotLength):tgt.reserveTruncate(gotLength), gotLength, gotLength, szread, timeout);
         }
-        catch (IJSOCK_Exception *e) {
-            if (e->errorCode()!=JSOCKERR_timeout_expired) {
-                EXCLOG(e,"receiveDaFsBuffer(1)");
-                PrintStackReport();
+        catch (IException *e)
+        {
+            exc.setown(e);
+        }
+
+        if (exc.get())
+        {
+            tgt.setLength(origlen);
+            EXCLOG(exc, "receiveDaFsBuffer");
+            PrintStackReport();
+            if (JSOCKERR_timeout_expired != exc->errorCode())
+            {
                 if (!tm.timemon||!tm.timemon->timedout())
                     flushDaFsSocket(socket);
             }
-            else {
-                EXCLOG(e,"receiveDaFsBuffer");
-                PrintStackReport();
-            }
-            tgt.setLength(origlen);
-            throw;
+            IJSOCK_Exception *JSexc = dynamic_cast<IJSOCK_Exception *>(exc.get());
+            if (JSexc != nullptr)
+                throw LINK(JSexc);
+            else
+                throw exc.getClear();
         }
-        catch (IException *e) {
-            EXCLOG(e,"receiveDaFsBuffer(2)");
-            PrintStackReport();
-            if (!tm.timemon||!tm.timemon->timedout())
-                flushDaFsSocket(socket);
-            tgt.setLength(origlen);
-            throw;
-        }
-
     }
     tgt.setEndian(__BIG_ENDIAN);
 }


mckellyln:
@jakesmith please review.

jakesmith:
@mckellyln - sounds like part of the problem described in the JIRA, I think should be a separate issue or a sub-task.
Also the description of this PR "Handle exception / **flush** socket in receiveDaFsBuffer" doesn't seem to fit the change (to sleep).

fs/dafsclient/rmtclient.cpp
```diff
@@ -391,7 +391,8 @@ void flushDaFsSocket(ISocket *socket)
         char buf[1024];
         for (;;)
         {
-            Sleep(1000);    // breathe
+            if (!totread)
+                Sleep(1000);    // breathe
```
jakesmith, line 395: this means it sleeps for 1 sec. every call to flushDaFsSocket, because sleep unconditionally called the 1st time,
Shouldn't it only sleep if nothing read the 2nd time onwards?

mckellyln: ok, sure.  I thought it was in the flushDaFsSocket loop sleeping each iteration whilst reading in a large buffer.
This would prevent that, sleeping only once.

mckellyln: I'll move the sleep out of the loop - perhaps even eliminate it

jakesmith: yeah, I don't know why the sleep's there all. We should try to avoid default delays like this if at all possible.
If it could ever go into a tight loop, it makes some sense, but readtms with its 1 minute timeout should mean it doesn't go into a tight loop.

mckellyln:
@jakesmith moved sleep out of loop instead of checking totread==0, and made it less time.
Changed title.
I didn't think there were 10000s of stack traces and log msgs of receiveDaFsBuffer(1) or (2) from this, suggesting it was in the loop in receiveDaFsBuffer ?  But I'm not 100% sure.

jakesmith:
@mckellyln - I think you're going to make further changes, so tagging you for now.

mckellyln:
@jakesmith please re-review.

fs/dafsclient/rmtclient.cpp
```diff
@@ -431,29 +436,28 @@ void receiveDaFsBuffer(ISocket * socket, MemoryBuffer & tgt, unsigned numtries,
...
                 PrintStackReport();
-                if (!tm.timemon||!tm.timemon->timedout())
-                    flushDaFsSocket(socket);
             }
-            else {
-                EXCLOG(e,"receiveDaFsBuffer");
+            else
+            {
+                EXCLOG(exception,"receiveDaFsBuffer(1)");
                 PrintStackReport();
```
jakesmith, line 455: I don't think there's any use/distinction between the if and else as far as the exception and stack logging is concerned, the (1) isn't useful, the exception text will already show that's it's a timeout expired exception.
So I'd move/common up the EXCLOG and PrintStackReport lines to above the if.

fs/dafsclient/rmtclient.cpp
```diff
@@ -431,29 +436,28 @@ void receiveDaFsBuffer(ISocket * socket, MemoryBuffer & tgt, unsigned numtries,
...
+                if (!tm.timemon||!tm.timemon->timedout())
+                    flushDaFsSocket(socket);
             }
-            tgt.setLength(origlen);
-            throw;
+            throw exception;
```
jakesmith, line 459: this needs to be:
```
    throw exception.getClear();
```

Otherwise, without either get clearing it, or linking it.The Owned<IException> will destroy it, and mean anything handling it will see a duff ptr.

fs/dafsclient/rmtclient.cpp
```diff
@@ -411,17 +411,22 @@ void receiveDaFsBuffer(ISocket * socket, MemoryBuffer & tgt, unsigned numtries,
...
         size32_t origlen = tgt.length();
-        try {
-            if (gotLength>maxsz) {
+        try
+        {
+            if (gotLength>maxsz)
+            {
                 StringBuffer msg;
                 msg.appendf("receiveBuffer maximum block size exceeded %d/%d",gotLength,maxsz);
                 PrintStackReport();
```
jakesmith, line 424: not new, but think should remove this PrintStackReport() line, as the exception handling is going to log it again, and will be able to tell it was because thrown here.

jakesmith:
@mckellyln - a couple of comments.

fs/dafsclient/rmtclient.cpp
mckellyln: Thx, yes.

fs/dafsclient/rmtclient.cpp
mckellyln: Agreed

fs/dafsclient/rmtclient.cpp
mckellyln: ok, sure.

mckellyln:
@jakesmith updated to address your comments, thank you.  Please re-review and I will squash.

jakesmith:
@mckellyln - looks good.

mckellyln:
ok, thx, I will squash

mckellyln:
@richardkchapman / @ghalliday ready for merge.

fs/dafsclient/rmtclient.cpp
```diff
@@ -431,29 +435,23 @@ void receiveDaFsBuffer(ISocket * socket, MemoryBuffer & tgt, unsigned numtries,
...
-            if (!tm.timemon||!tm.timemon->timedout())
-                flushDaFsSocket(socket);
-            tgt.setLength(origlen);
-            throw;
+            if (JSOCKERR_timeout_expired != exception->errorCode())
+            {
+                if (!tm.timemon||!tm.timemon->timedout())
+                    flushDaFsSocket(socket);
+            }
+            throw exception.getClear();
```
ghalliday, line 453: @mckellyln Is there any potential issue if this was caught as a IJSOCK_Exception, but rethrown as an IException?  There are quite a few places that catch an IJSOCK_Exception then an IException and process them differently.

mckellyln: Yes, there are two places where a catch from receiveDaFsBuffer() differentiates the type of exception.
I originally was planning to use two different exceptions and throw a matching type.  I'll update it now, thanks.

mckellyln: @ghalliday updated to catch each exception type.  Is this what you meant.
Thanks for ffinding this.

jakesmith: Right, with only a catch (IException *) and later throw, other catches will no longer see it as a IJSock_Exception (if that's what it was originally).
It's annoying and a problem elsewhere, where we generalize on catch (IException *).

fs/dafsclient/rmtclient.cpp
```diff
@@ -431,29 +435,28 @@ void receiveDaFsBuffer(ISocket * socket, MemoryBuffer & tgt, unsigned numtries,
...
         }
-        catch (IException *e) {
-            EXCLOG(e,"receiveDaFsBuffer(2)");
-            PrintStackReport();
-            if (!tm.timemon||!tm.timemon->timedout())
-                flushDaFsSocket(socket);
+        catch (IException *e)
+        {
+            exception.setown(e);
             tgt.setLength(origlen);
```
jakesmith, line 446: prob. worth moving this line (and one on line 441) into the if (exception) block.

fs/dafsclient/rmtclient.cpp
ghalliday: @mckellyln The problem isn't with the catch, the IException * will match both.

The problem is when the exception is thrown.  If throw; is used it is thrown with the type of the original exception.  If throw e is used it is thrown with the type that is thrown.  In this case it will be thrown with a type of IException *.

If there is some calling code that catches on (IJSOCK_Exception *) it will not match.  E.g. sendRemoteCommand() only has a 

catch (IJSOCK_Exception *e)

You could either have two owned variables, and throw the first that matches, or do a dynamic cast, and throw the IJSOCK_Exception * if the cast succeeds.

jakesmith: I needed reminding about that problem too, thanks.

mckellyln:
@ghalliday I've updated the code to throw the correct type.
How does it look ?
thx.

ghalliday:
@mckellyln I think this now looks ok, please can you squash

mckellyln:
Squashed, ready for merge.
Thank you.

ghalliday:
@richardkchapman please sanity check/merge

PR #14049, opened 2020-08-03
HPCC-24536 Suppress some verbose Thor logging by default

jakesmith:
Allow it to be turned back on per workunit if necessary,
via #option('maxLogDetail', <value>);

Where low priority logging has been changed to logging level 200.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/thorlcr/master/thgraphmanager.cpp
+++ b/thorlcr/master/thgraphmanager.cpp
@@ -711,6 +711,12 @@ void CJobManager::run()
         {
             factory.setown(getWorkUnitFactory());
             workunit.setown(factory->openWorkUnit(wuid));
+
+            unsigned maxLogDetail = workunit->getDebugValueInt("maxlogdetail", DefaultDetail); 
+            ILogMsgFilter *existingLogHandler = queryLogMsgManager()->queryMonitorFilter(logHandler);
+            dbgassertex(existingLogHandler);
+            verifyex(queryLogMsgManager()->changeMonitorFilterOwn(logHandler, getCategoryLogMsgFilter(existingLogHandler->queryAudienceMask(), existingLogHandler->queryClassMask(), maxLogDetail)));
+
             allDone = execute(workunit, wuid, graphName, agentep);
             daliLock.clear();
             reply(workunit, wuid, NULL, agentep, allDone);
--- a/thorlcr/msort/tsortm.cpp
+++ b/thorlcr/msort/tsortm.cpp
@@ -47,10 +47,6 @@
 #include "tsortm.hpp"
 
 
-#ifdef _TESTING
-#define _TRACE
-#endif
-
 #ifdef _DEBUG
 //#define _MEMTRACE
 #endif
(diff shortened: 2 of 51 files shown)


jakesmith:
@mckellyln - please review

thorlcr/master/thgraphmanager.cpp
```diff
@@ -711,6 +711,12 @@ void CJobManager::run()
         {
             factory.setown(getWorkUnitFactory());
             workunit.setown(factory->openWorkUnit(wuid));
+
+            unsigned maxLogDetail = workunit->getDebugValueInt("maxlogdetail", DefaultDetail);
```
mckellyln, line 715: Does lowercase matter here ?  Typically wouldn't it be maxLogDetail ?

thorlcr/msort/tsortm.cpp
```diff
@@ -124,7 +120,7 @@ class CSortNode: public SortSlaveMP
         try {
 
             StringBuffer epstr;
-            ActPrintLog(activity, "Connect to %s:%d",endpoint.getIpText(epstr).str(),(unsigned)mpport);
+            ActPrintLog(activity, thorDetailedLogLevel, "Connect to %s:%d",endpoint.getIpText(epstr).str(),(unsigned)mpport);
```
mckellyln, line 123: ::ActPrintLog() ?

thorlcr/msort/tsortm.cpp
```diff
@@ -688,7 +642,6 @@ class CSortMaster : public IThorSorterMaster, public CSimpleInterface
             }
         } afor3(slaves, splitMap, numnodes, numsplits);
         afor3.For(numnodes, 20, true);
-#ifdef _TRACE
 #ifdef TRACE_PARTITION
```
mckellyln, line 645: if (!REJECTLOG(MCthorDetailedDebugInfo)) before the #ifdef here ?

thorlcr/msort/tsortm.cpp
jakesmith: only needed if ambiguous with class method, there is none in this context.

thorlcr/master/thgraphmanager.cpp
jakesmith: no, workunit debug values and the call, are case insensitive

thorlcr/msort/tsortm.cpp
jakesmith: I didn't change every existing compile time directive that controlled debug tracing like this.
There is a bit of overhead even with REJECTLOG.
Could change, but don't think it's warranted for this and most other compile time section of tracnig.

mckellyln (approved):

mckellyln:
@jakesmith looks good.
Many changes but almost all are mechanical.

jakesmith:
@ghalliday - please check and merge.

thorlcr/thorutil/thormisc.hpp
```diff
@@ -409,6 +412,14 @@ inline void ActPrintLog(const CGraphElementBase *container, const char *format,
     ActPrintLogArgs(container, thorlog_ecl, MCdebugProgress, format, args);
     va_end(args);
 }
+inline void ActPrintLog(const CGraphElementBase *container, unsigned traceLevel, const char *format, ...) __attribute__((format(printf, 3, 4)));
```
ghalliday, line 415: Is this function, and the GraphPrintLog function actually used?

jakesmith: yes, both are used.

PR #14053, opened 2020-08-04
HPCC-24541 Add CLocalWorkUnit::getWUProcessLogInfo()

wangkx:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/common/workunit/workunit.cpp
+++ b/common/workunit/workunit.cpp
@@ -62,6 +62,8 @@
 #include <string>
 #include <algorithm>
 
+#define ECLAGENTLOGSEARCHSTR        "eclagent."
+
 using namespace cryptohelper;
 
 static int workUnitTraceLevel = 1;
@@ -4216,6 +4218,8 @@ public:
             { return c->getAbortBy(str); }
     virtual unsigned __int64 getAbortTimeStamp() const
             { return c->getAbortTimeStamp(); }
+    virtual void getWUProcessLogInfo(const char *processType, const char *processName, IArrayOf<IConstWUProcessLogInfo> &processLogs) const
+            { c->getWUProcessLogInfo(processType, processName, processLogs); }
     virtual void import(IPropertyTree *wuTree, IPropertyTree *graphProgressTree)
             { return c->import(wuTree, graphProgressTree); }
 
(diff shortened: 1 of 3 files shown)


wangkx:
@jakesmith please review this PR. This is a part of #13258. It may be used by https://track.hpccsystems.com/browse/HPCC-24156 and https://track.hpccsystems.com/browse/HPCC-21753. The next step is to move more process log related code from WsWorkuits.so into workunit.so (like getSlaveLogFileName() and usingDedicatedLogFiles()).

common/workunit/workunit.cpp
```diff
@@ -8142,6 +8195,106 @@ bool CLocalWorkUnit::getDebugValueBool(const char * propname, bool defVal) const
...
+
+    Owned<IPropertyTreeIterator> processInstances = getProcesses(processType, isEmptyString(processName) ? nullptr : processName);
+    ForEach (*processInstances)
+    {
+        StringBuffer logSpec;
+        IPropertyTree &proc = processInstances->query();
+        proc.getProp("@log", logSpec);
+
+        const char *processName = proc.queryName();
+        if (isEmptyString(processName) || logSpec.isEmpty())
```
jakesmith, line 8250: minor: any reason to use a const char * for processName, but a StringBuffer for logSpec? If you don't need a StringBuffer, better to use proc.queryProp

common/workunit/workunit.cpp
```diff
@@ -8142,6 +8195,106 @@ bool CLocalWorkUnit::getDebugValueBool(const char * propname, bool defVal) const
...
+
+        unsigned numberOfThorSlaves = 0, processID = 0;
+        StringBuffer groupName;
+        if (strieq(processType, "Thor"))
+        {
+            //The groupName here may not be used if the WU is imported from a ZAP report because
+            //the getClusterThorGroupName() may not work for the WU imported from a ZAP report.
+            getClusterThorGroupName(groupName, processName);
+
+            numberOfThorSlaves = proc.getPropInt("@max", 0);
```
jakesmith, line 8264: minor: don't need the default 0

common/workunit/workunit.cpp
```diff
@@ -8142,6 +8195,106 @@ bool CLocalWorkUnit::getDebugValueBool(const char * propname, bool defVal) const
...
+            if (numberOfThorSlaves == 0)
+            { //Legacy WU
+                StringAttr clusterName(queryClusterName());
+                if (clusterName.isEmpty())
+                { //Cluster name may not be set yet
+                    IWARNLOG("Cluster name may not be set for workunit %s", queryWuid());
```
jakesmith, line 8270: in which situations is the cluster name not set yet?

common/workunit/workunit.cpp
```diff
@@ -8142,6 +8195,106 @@ bool CLocalWorkUnit::getDebugValueBool(const char * propname, bool defVal) const
...
+                }
+                else
+                {
+                    //This may not work for the WU imported from a ZAP report.
+                    Owned<IConstWUClusterInfo> clusterInfo = getTargetClusterInfo(clusterName);
```
jakesmith, line 8275: should not expect / rely on the cluster being present in the environment, for an imported wuid..

common/workunit/workunit.cpp
```diff
@@ -8142,6 +8195,106 @@ bool CLocalWorkUnit::getDebugValueBool(const char * propname, bool defVal) const
...
+StringBuffer & getProcessLogDateFromLogFileName(const char * processType, const char * logFileNameWithPath, StringBuffer & logDate)
+{
+    //WU xml has a 'log' attribute for each log file used by a process. The code here
+    //tries to parse log date from the 'log' attribute if the attribute contains the
+    //log date information. Take a thor master log as an example (ECLAgent log is
+    //similar). The 'log' attribute looks like '...thormaster.yyyy_mm_dd.log' if the
+    //'log' attribute does contain the log date information. If thor master stores all
+    //of its log lines for a WU into one log file, there is no need for the log file
+    //name to have the 'yyyy_mm_dd' part (10 byte long). The log attribute may look
+    //like '...thormaster.log'.
```
jakesmith, line 8207: >If thor master stores all of its log lines for a WU into one log file there is no need for the log file name to have the 'yyyy_mm_dd' part 

It's just the ZAP file creation that causes a filename without a date stamp in it isn't it?
If we ever changed bare-metal to write to a file per wuid, the file format would change (e.g. to contain the wuid name).

common/workunit/workunit.hpp
```diff
@@ -1190,6 +1190,18 @@ interface IConstWorkUnitInfo : extends IInterface
...
 };
 
+interface IConstWUProcessLogInfo : extends IInterface
+{
+    virtual const char *getLogSpec() const = 0;
+    virtual const char *getProcessName() const = 0;
+    virtual unsigned getProcessID() const = 0;
+    virtual const char *getGroupName() const = 0;
+    virtual const char *getLogDate() const = 0;
+    virtual const char *getPattern() const = 0;
```
jakesmith, line 1200: difficult to say without seeing how this new interface would be used, but feels like some of these should internal implementation details, e.g. how does the caller use groupname (especially if there is no group name because the WU was imported).

Feels like the caller should be asking for the log, and the implementtion should be using these details to get it, so the caller doesn't have to concern itself with the internal details of how to access it.

jakesmith:
@wangkx - back to you for now.

PR #14056, opened 2020-08-04
HPCC-24546 ECL HTTPCALL and SOAPCALL support for secrets

afishbeck:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:


afishbeck:
@ghalliday please take a look.  The certificates part is still a bit of a work in progress.

afishbeck:
The bare metal case will mostly be a copying of the secrets to the right location.

system/jlib/jutil.cpp
```diff
@@ -3056,21 +3056,37 @@ extern jlib_decl void setSecretMount(const char * path)
         secretDirectory.set(path);
 }
 
-extern jlib_decl StringBuffer & getSecret(StringBuffer & result, const char * name, const char * key)
+extern jlib_decl StringBuffer & getSecret(StringBuffer & result, const char * name, const char * key, bool required)
```
ghalliday, line 3059: Maybe we should also export a

    extern jlib_decl IFile * getSecretFile(const char * name, const char * key);

that these functions can use, and can be used externally.

ghalliday: although that would tie us to using a mount as an implementation, so I don't think it should be public.  We may want to extend the implementation to use other vault implementations.,  In which case there will not be an underlying file.  At that point we should probably introduce an ISecret though to avoid checks for existance followed by retrieving the value.

common/thorhelper/thorsoapcall.cpp
```diff
@@ -941,46 +938,102 @@ class CWSCHelper : implements IWSCHelper, public CInterface
...
+        if (0==strncmp(hosts, "secret:", 7))
         {
-            if (wscMode == SCrow)
+            const char *finger = hosts.get()+7;
+            if (isEmptyString(finger))
+                throw MakeStringException(0, "%sCALL HTTP-CONNECT SECRET specified with no name", wscType == STsoap ? "SOAP" : "HTTP");
+
+            StringBuffer secretName("http-connect-");
+            secretName.append(finger);
+            if (!isEmptyString(helper->getProxyAddress()))
```
ghalliday, line 953: This will potentially leak - see the previous code which assigned it to a variable first.

common/thorhelper/thorsoapcall.cpp
```diff
@@ -941,46 +938,102 @@ class CWSCHelper : implements IWSCHelper, public CInterface
...
             }
         }
         else
+        {
+            UrlListParser urlListParser(hosts);
+            urlListParser.getUrls(urlArray);
+            proxyAddress.set(s.setown(helper->getProxyAddress()));
+        }
+
+        if ((numUrls = urlArray.ordinality()) == 0)
```
ghalliday, line 994: trivial:  The code existed already, but (arguably) clearer to assign to numUrls first.

helm/examples/httpcall/README.md
```diff
@@ -0,0 +1,39 @@
+# Containerized HPCC Systems HTTPCALL and SOAPCALL Secrets
+
+## HPCC Systems HTTPCALL Secrets
+
+## About HTTP-CONNECT Secrets:
+
+HTTP-CONNECT secrets consist of a url string and optional additional secrets associated with that URL.  If we supported providing credentials
+without an associated url then those credentials could be sent anywhere and wouldn't be secret very long.
+
+Besides the URL values can currently be set for proxy (tusted for keeping these secrets), username, password, tls.key
```
ghalliday, line 10: typo: tusted->trusted

ghalliday (commented):
@afishbeck a couple of comments.  I wonder whether we should think about how the secrets will be extended to support hashicorp etc.  sooner rather than later.  E.g., should they be dependent on the user, how do we avoid querying them more than once.  Implement a mechanism for adding secret hooks.  Does the user have the notion of different named secret stores, that they can select between???

system/jlib/jutil.cpp
afishbeck: I just directly violated the idea that they aren't mounted with my tls change.  I hadn't read your comment yet.  I'll keep looking at the tls stuff and try to get rid of needing to use a file rather then memory.  I've gotten further but still haven't quite put a good test together.

afishbeck:
@ghalliday I'll look at refactoring away the change I just made to get the secret path for tls keys.

Regarding using vaults.  I was planning on looking at least 2 things:
1. making it extensible with specific prefixes for hashicorp vault and azure vault.  Storing the vault connect info in a secret.  What child values the vault-connect secret has could be type dependent.. and mysecret name could evolve into a path (depending on the particular vault type.
    I.e. something like 
                'secret:localsecret'
                'vault.{vaultAccessLocalSecret}:mysecretname'
2. using something like helm-secrets... to map hashicorp, azure, and other secrets into kubernetes secrets.  This is really more of an operational management technique.

afishbeck:
I think hashicorp secrets can contain multiple key/value pairs but an azure secret might be just one value so need to figure out how the mutliple parts used for something like my http-connect-name secrets will be mapped.  For example stored connection string may work for some things, but not storing the cert or key.

HTTPCALL('vault.hashi.name:mysecret','GET', 'text/xml', ServiceOutRecord, LOG);

Might be good enough, but  

HTTPCALL('vault.azure.name:mysecret','GET', 'text/xml', ServiceOutRecord, LOG);

Might need some figuring out.

afishbeck:
Actually when it comes to mirroring hashicorp vault secrets into kubernetes secrets there is the vault sidecar:

https://learn.hashicorp.com/tutorials/vault/kubernetes-sidecar

ghalliday:
Interesting about the sidecar.  I'm not sure that will work for the secrets that need keys supplied by two people (e.g., the system and the user).
Also if you use a prefix or suffix to indicate where a secret is coming from, that will not work with my current scheme for implicitly deriving the secret name from the storage account for accessing an azure storage account.  (I could change that syntax so the secret name could also be specified - it worked a bit like that in the first version).

I think it would be valuable to spend some time thinking through and designing exactly what support for secrets we want - no that we understand some of the issues.

richardkchapman:
Sidecar injectors seem to have some issues (in that you need to know to terminate the sidecar pod - especially on jobs - or your pod may never terminate).

afishbeck:
I'm prototyping direct access from HASHICORP vault.  The sidecar is
interesting but also complicated, besides the logistics of running it, in
that the user can transform what gets injected.

afishbeck:
Replace by https://github.com/hpcc-systems/HPCC-Platform/pull/14122.

PR #14070, opened 2020-08-07
HPCC-24582 Add DESDL support for numeric boolean output

afishbeck:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/esdllib/esdl_transformer2.cpp
+++ b/esp/esdllib/esdl_transformer2.cpp
@@ -311,7 +311,10 @@ void Esdl2Base::output_content(Esdl2TransformerContext &ctx, const char * conten
                 switch (type_id)
                 {
                     case ESDLT_BOOL:
-                        ctx.writer->outputBool(strToBool(content), tagname);
+                        if (ctx.flags & ESDL_TRANS_NUMERIC_BOOLEAN)
+                            ctx.writer->outputNumericString(strToBool(content) ? "1" : "0", tagname);
+                        else
+                            ctx.writer->outputBool(strToBool(content), tagname);
                         break;
                     case ESDLT_INT8:
                     case ESDLT_INT16:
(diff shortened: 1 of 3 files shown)


afishbeck:
@timothyklemm @asselitx please review.

esp/esdllib/esdl_transformer2.cpp
```diff
@@ -1558,13 +1561,39 @@ Esdl2Transformer::~Esdl2Transformer()
 
 }
 
+static const char *queryMethodMetaData(IEsdlDefMethod *mthdef, IEsdlMethodInfo *mi, const char *tag)
```
asselitx, line 1564: Is this correct: this function takes either mthdef or mi because it's called from different locations that will already have a pointer to either one or the other. Furthermore, the mi object actually contains a pointer to the IEsdlDefMethod, so either way we are getting the information from essentially the same place.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -758,6 +758,11 @@ void EsdlServiceImpl::handleServiceRequest(IEspContext &context,
 
         implType = getEsdlMethodImplType(tgtcfg->queryProp("@querytype"));
 
+        bool use_numeric_bool = srvdef.getPropBool("numeric_bool");
+        unsigned txResultFlags = ESDL_TRANS_OUTPUT_ROOT;
+        if (use_numeric_bool)
+            txResultFlags |= ESDL_TRANS_NUMERIC_BOOLEAN;
+
```
asselitx, line 765: So the service level setting is applied here, but is possible to override in the method when calling process or processHPCCResult.

asselitx:
Code looks good and passes testing. Ready for merge.

esp/esdllib/esdl_transformer2.cpp
afishbeck: I'll see if I can simplify it then.

afishbeck: Yes that is correct.  I just wanted a single consolidated easy to call function that took either interface type.  They do ultimately get the info from the same place.

esp/services/esdl_svc_engine/esdl_binding.cpp
afishbeck: yes, exactly.  The method override exists int the common transformer code.. but setting the default from the service level info had to happen here.

PR #14071, opened 2020-08-07
HPCC-24463 Start using storageplane information in hthor and thor

ghalliday:
## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/common/thorhelper/roxiehelper.cpp
+++ b/common/thorhelper/roxiehelper.cpp
@@ -2393,8 +2393,13 @@ void FlushingJsonBuffer::setScalarUInt(const char *resultName, unsigned sequence
 ClusterWriteHandler::ClusterWriteHandler(char const * _logicalName, char const * _activityType)
     : logicalName(_logicalName), activityType(_activityType)
 {
-    makePhysicalPartName(logicalName.get(), 1, 1, physicalName, false);
-    splitFilename(physicalName, &physicalDir, &physicalDir, &physicalBase, &physicalBase);
+}
+
+void ClusterWriteHandler::getPhysicalName(StringBuffer & name, const char * cluster) const
+{
+    Owned<IStoragePlane> plane = getStoragePlane(cluster);
+    const char * prefix = plane ? plane->queryPrefix() : nullptr;
+    makePhysicalPartName(logicalName.get(), 1, 1, name, 0, DFD_OSdefault, prefix);
 }
 
 void ClusterWriteHandler::addCluster(char const * cluster)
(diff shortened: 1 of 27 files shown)


ghalliday:
@richardkchapman @jakesmith please take a look at this.  I will add information to the jira about how I tested it.

common/thorhelper/roxiehelper.cpp
```diff
@@ -2393,15 +2393,31 @@ void FlushingJsonBuffer::setScalarUInt(const char *resultName, unsigned sequence
 ClusterWriteHandler::ClusterWriteHandler(char const * _logicalName, char const * _activityType)
     : logicalName(_logicalName), activityType(_activityType)
 {
-    makePhysicalPartName(logicalName.get(), 1, 1, physicalName, false);
```
richardkchapman, line 2396: Looks like you have moved this code from construction time to being done every time the information is asked for. Could that be a problem if caller asks a lot? Does it need caching?

Probably not, but just checking...

dali/base/dafdesc.cpp
```diff
@@ -3139,3 +3144,136 @@ void setupContainerizedStorageLocations()
...
+        IPropertyTree & cur = iter->query();
+        //Check if this has already been done - so the function is safe to call more than once
+        const char * oldgroup = cur.queryProp("@group");
+        if (oldgroup)
+            continue;
+
+        const char * name = cur.queryProp("@name");
+        group.clear().append(name).toLowerCase();
+
+#ifdef _DEBUG
```
richardkchapman, line 3219: Why only check in debug mode?

richardkchapman: Is this more of a dbgassertex, or is it validating user-supplied info?

dockerfiles/startall.sh
```diff
@@ -21,6 +21,7 @@
 # Usage startup.sh [<image.version>] [<helm-install-options>]
 
 DOCKER_REPO=hpccsystems
+scriptdir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
```
richardkchapman, line 24: Looks like a good change, but not really related to the rest of the PR...

ghalliday: It is because my test scripts are in a different directory, and they were calling this one.  Do you want it in a separate PR?

dali/base/dafdesc.cpp
ghalliday: I guess it is validating user input - checking there aren't storage planes that only different in their case.  I suspect the helm chart would fail to deploy in that case, but I don't think there is any harm in making this unconditional.  (Potential O(n^2) operation, but I can't see it ever being an issue, and Jake's hash change would limit it.)

common/thorhelper/roxiehelper.cpp
ghalliday: I think it would mean that the function could potentially be called 3 times instead of once.
getLocalPhysicalFilename()
splitPhysicalFilename()
copyPhysical()
are each only called one time.

The function itself is not too expensive, and the change is required to support multiple storage planes properly.  So I think it is fine to no longer cache the results.

richardkchapman: ok

dockerfiles/startall.sh
richardkchapman: You can leave it in here

PR #14075, opened 2020-08-10
HPCC-23295 Add the time a graph is blocked waiting for a thor

shamser:
## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/common/workunit/workunit.cpp
+++ b/common/workunit/workunit.cpp
@@ -13606,6 +13606,10 @@ void executeThorGraph(const char * graphName, IConstWorkUnit &workunit, const IP
 {
     unsigned timelimit = workunit.getDebugValueInt("thorConnectTimeout", config.getPropInt("@thorConnectTimeout", 60));
     StringAttr wuid(workunit.queryWuid());
+    IConstWUGraph *graph = workunit.getGraph(graphName);
+    if (!graph)
+        throw makeStringExceptionV(0, "getGraph() returns nullptr for %s", graphName);
+    unsigned wfid = graph->getWfid();
 
 #ifdef _CONTAINERIZED
     // NB: If a single agent were to want to launch >1 Thor, then the threading could be in the workflow above this call.
@@ -13618,6 +13622,8 @@ void executeThorGraph(const char * graphName, IConstWorkUnit &workunit, const IP
     WUState state = WUStateUnknown;
     if (config.hasProp("@queue"))
     {
+        CCycleTimer elapsedTimer;
+
         bool multiJobLinger = config.getPropBool("@multiJobLinger");
         if (executeGraphOnLingeringThor(workunit, graphName, multiJobLinger ? config.queryProp("@queue") : nullptr))
             PROGLOG("Existing lingering Thor handled graph: %s", graphName);
@@ -13636,12 +13642,19 @@ void executeThorGraph(const char * graphName, IConstWorkUnit &workunit, const IP
         runningTimeLimit = runningTimeLimit ? runningTimeLimit : INFINITE;
 
         std::list<WUState> expectedStates = { WUStateRunning, WUStateWait };
+        unsigned __int64 blockedTime = 0;
         for (unsigned i=0; i<2; i++)
         {
             WUState state = waitForWorkUnitToComplete(wuid, timelimit*1000, expectedStates);
             DBGLOG("Got state: %s", getWorkunitStateStr(state));
             if (WUStateWait == state) // already finished
+            {
+                // workunit may have spent time in blocked state, but then transitioned to
+                // wait state quickly such that this code did not see its running state.
+                if (!blockedTime)
+                    blockedTime = elapsedTimer.elapsedNs();
                 break;
+            }
             else if ((INFINITE != timelimit) && (WUStateUnknown == state))
                 throw makeStringExceptionV(0, "Query %s failed to start within specified timelimit (%u) seconds", wuid.str(), timelimit);
             else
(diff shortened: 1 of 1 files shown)


shamser:
@ghalliday Please can you review?

common/workunit/workunit.cpp
```diff
@@ -13648,7 +13650,11 @@ void executeThorGraph(const char * graphName, IConstWorkUnit &workunit, const IP
         VStringBuffer job("%s-%s", wuid.str(), graphName);
         runK8sJob("thormaster", wuid, job, queryComponentConfig().getPropBool("@deleteJobs", true), { { "graphName", graphName} });
     }
-
+    unsigned __int64 blockedTime = elapsedTimer.elapsedNs();
```
ghalliday, line 13653: @shamser I think this timing the execution of the thor job, rather than the blocked time.
Also, Jake is better placed to review the code.

shamser: @jakesmith Please can you review.

shamser: I've removed this code.  I'm looking at how blocked time may be tracked in container environment.

shamser:

common/workunit/workunit.cpp
```diff
@@ -13595,6 +13595,8 @@ void executeThorGraph(const char * graphName, IConstWorkUnit &workunit, const IP
 {
     unsigned timelimit = workunit.getDebugValueInt("thorConnectTimeout", config.getPropInt("@thorConnectTimeout", 60));
     StringAttr wuid(workunit.queryWuid());
+    IConstWUGraph *graph = workunit.getGraph(graphName);
+    unsigned wfid = graph ? graph->getWfid() : 0;
```
jakesmith, line 13599: is graph ever expected to be null? if so how?
If it's not expected, then it's going to go wrong soon after this, would be better to catch here and throw a meaningful error.

common/workunit/workunit.cpp
```diff
@@ -13641,6 +13645,11 @@ void executeThorGraph(const char * graphName, IConstWorkUnit &workunit, const IP
             }
             timelimit = runningTimeLimit;
             expectedStates = { WUStateWait };
+            unsigned __int64 blockedTime = elapsedTimer.elapsedNs();
+            {
+                Owned<IWorkUnit> w = &workunit.lock();
+                updateWorkunitStat(w, SSTgraph, graphName, StTimeBlocked, nullptr, blockedTime, wfid);
```
jakesmith, line 13651: I think you need to also cater for the : if (WUStateWait == state) // already finished
case above too.

Because it could in theory spend some time blocked, but then v. quickly transition from running to finished, so this code would see the state change to WUStateWait only, exit, and not record any blocked time.

jakesmith:
@shamser - a couple of questions/issues.

common/workunit/workunit.cpp
shamser: @jakesmith If it has already finished (i.e. it's WUStateWait==state) then we can't really say that it has blocked.  As it hasn't really blocked,  would we be misleadingly recording this as blocked time?

jakesmith: @shamser - It is initially in WUStateBlocked state before it enters the for loop (line 13632).
It then expects after some blocked time, to transition to WUStateRunning.
When Thor finishes executing the graph, it will set the state to WUStateWait.

If the job is blocked for some time, but Thor executes the graph quickly, waitForWorkUnitToComplete will miss the transition from Running->Wait. It will only see Blocked->Wait.
The graph may have been blocked for an hour, but as it stands, it won't record any blocked time, because this code sees WUStateWait 1st and breaks out before it gets to the new code that records blocked time.

shamser: Thanks for the explanation.  I'll make the change.

shamser:
@jakesmith Made some changes. Please can you review agan.

common/workunit/workunit.cpp
```diff
@@ -13636,12 +13642,17 @@ void executeThorGraph(const char * graphName, IConstWorkUnit &workunit, const IP
...
 
         std::list<WUState> expectedStates = { WUStateRunning, WUStateWait };
+        unsigned __int64 blockedTime = 0;
         for (unsigned i=0; i<2; i++)
         {
             WUState state = waitForWorkUnitToComplete(wuid, timelimit*1000, expectedStates);
             DBGLOG("Got state: %s", getWorkunitStateStr(state));
             if (WUStateWait == state) // already finished
+            {
+                if (!blockedTime)
```
jakesmith, line 13652: worth a comment here I think, along the lines of:
// wuid may have spent time in blocked state, but then transitioned to wait state quickly such that this code did not see it's running state.

common/workunit/workunit.cpp
```diff
@@ -13650,9 +13661,13 @@ void executeThorGraph(const char * graphName, IConstWorkUnit &workunit, const IP
                 if (it == expectedStates.end())
                     throw makeStringExceptionV(0, "Query %s failed, state: %s", wuid.str(), getWorkunitStateStr(state));
             }
+            if (!blockedTime)
```
jakesmith, line 13664: If it reaches here, blockedTime cannot already be set here, so no point in this test.
Either it was wait state and broke out, or it running (1st transition/1st time through loop), or an unexpected state.

jakesmith:
@shamser - a couple of comments. Please squash after addressing.

shamser:
@jakesmith I've addressed the comments and squashed

jakesmith:
@shamser - looks good.

jakesmith:
@richardkchapman - ready to merge

PR #14080, opened 2020-08-11
HPCC-24635 Add NetworkPolicy to helm charts for ingress control

richardkchapman:
## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/common/workunit/workunit.cpp
+++ b/common/workunit/workunit.cpp
@@ -14044,13 +14044,13 @@ static void setResources(StringBuffer &jobYaml, const IConstWorkUnit *workunit, 
 }
 
 
-void deleteK8sJob(const char *componentName, const char *job)
+void deleteK8sResource(const char *componentName, const char *job, const char *resource)
 {
     VStringBuffer jobname("%s-%s", componentName, job);
     jobname.toLowerCase();
-    VStringBuffer deleteJob("kubectl delete job/%s", jobname.str());
+    VStringBuffer deleteResource("kubectl delete %s/%s", resource, jobname.str());
     StringBuffer output, error;
-    bool ret = runExternalCommand(componentName, output, error, deleteJob.str(), nullptr);
+    bool ret = runExternalCommand(componentName, output, error, deleteResource.str(), nullptr);
     DBGLOG("kubectl delete output: %s", output.str());
     if (error.length())
         DBGLOG("kubectl delete error: %s", error.str());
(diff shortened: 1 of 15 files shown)


richardkchapman:
@jakesmith Please review

helm/hpcc/templates/network.yaml
```diff
@@ -0,0 +1,71 @@
...
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: default-deny-all
+spec:
+  podSelector: {}
+  policyTypes:
+  - Ingress
+  - Egress
+  egress:
```
richardkchapman, line 11: You need to allow egress to kube-system for dns lookup to work, However this method is not very reliable, as it relies on the label having been set on the kube-system namespace which it is not by default.

Alternatives are:

1. Don't add egress controls
2. Allow egress to any machine on port 53 (dns)

helm/hpcc/templates/network.yaml
```diff
@@ -0,0 +1,71 @@
...
+---
+kind: NetworkPolicy
+apiVersion: networking.k8s.io/v1 
+metadata:
+  name: allow-apiserver
+spec:
+  policyTypes:
+  - Egress
+  podSelector: {}
+  egress:
```
richardkchapman, line 26: Need to allow egress to the server used by kubectl, for components that use it. However the cidr to specify will vary from system to system.

Alternatives:
1. Don't block egress by default
2. Allow helm user to specify kubectl server to unblock, if they want to block egress

jakesmith:
```
[ERROR] templates/: template: hpcc/templates/thor.yaml:40:19: executing "hpcc/templates/thor.yaml" at <include "hpcc.daliArg" $>: error calling include: template: hpcc/templates/_helpers.tpl:364:19: executing "hpcc.daliArg" at <index .Values.dali 0>: error calling index: reflect: slice index out of range
```

@richardkchapman - is that new?

common/workunit/workunit.cpp
```diff
@@ -14100,13 +14100,23 @@ void waitK8sJob(const char *componentName, const char *job)
...
 }
 
-void launchK8sJob(const char *componentName, const char *wuid, const char *job, const std::list<std::pair<std::string, std::string>> &extraParams)
+bool applyK8sYaml(const char *componentName, const char *wuid, const char *job, const char *suffix, const std::list<std::pair<std::string, std::string>> &extraParams, bool optional)
 {
-    VStringBuffer jobname("%s-%s", componentName, job);
+    StringBuffer jobname(job);
     jobname.toLowerCase();
-    VStringBuffer jobSpecFilename("/etc/config/%s-jobspec.yaml", componentName);
+    VStringBuffer jobSpecFilename("/etc/config/%s-%s.yaml", componentName, suffix);
```
jakesmith, line 14107: pedantic: perhaps should remove 'job' from here and jobYaml below, to become k8sSpecFilename / k8sYaml

richardkchapman:
I think that's from one of the tests that's supposed to fail

helm/hpcc/templates/network.yaml
```diff
@@ -0,0 +1,71 @@
...
+  podSelector: {}
+  policyTypes:
+  - Ingress
+  - Egress
+  egress:
+  - to:
+#  This only works if we know that the kube-system namespace has been labelled, which we cannot control?
+    - namespaceSelector:
+        matchLabels:
+          name: kube-system
```
jakesmith, line 16: so if the name space of the kube-system is not "kube-system" this NP will not be in force?
In other words, if name space doesn't match - this deny-all NP is not used?

jakesmith:
@richardkchapman - all looks sensible, a couple of questions.

helm/hpcc/templates/network.yaml
richardkchapman: This isn't a deny - it's an allow... meaning if it is not applied no-one can call kubectl

And it's not the name of the namespace that matters but its label, which is not present by default.

richardkchapman:
@ghalliday

ghalliday:
@richardkchapman needs squashing and rebasing, otherwise looks ok.

PR #14092, opened 2020-08-19
HPCC-24633 Add readiness/startup probes to helm charts

richardkchapman:

## Type of change:

## Checklist:
  - [x] The commit is signed.
- [x]  I have read the CONTRIBUTORS document.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [x] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

## Testing:

Manual testing under minikube

### DIFF
--- a/helm/hpcc/templates/_helpers.tpl
+++ b/helm/hpcc/templates/_helpers.tpl
@@ -314,6 +314,28 @@ Add Secret volume for a component
 {{- end -}}
 
 {{/*
+Add sentinel-based probes for a component
+*/}}
+{{- define "hpcc.addSentinelProbes" -}}
+env:
+- name: "SENTINEL"
+  value: "{{ .name }}.sentinel"
+startupProbe:
+  exec:
+    command:
+    - cat
+    - "{{ .name }}.sentinel"
+  failureThreshold: 30
+  periodSeconds: 10
+readinessProbe:
+  exec:
+    command:
+    - cat
+    - "{{ .name }}.sentinel"
+  periodSeconds: 10
+{{ end -}}
+
+{{/*
 Return a value indicating whether a storage plane is defined or not.
 */}}
 {{- define "hpcc.isValidStoragePlane" -}}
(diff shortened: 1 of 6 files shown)


richardkchapman:
@jakesmith Please review

jakesmith:
@richardkchapman - looks good.

We might want to improve later, to add custom probe's for each component.
Am mainly thinking about Dali - which potentially takes minutes to be ready.
We could just move the sentinel, which is currently created very early, but it might be better to probe using a dali client, to see if it's responding normally.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -313,6 +313,30 @@ Add Secret volume for a component
...
+env:
+- name: "SENTINEL"
+  value: "{{ .name }}.sentinel"
+startupProbe:
+  exec:
+    command:
+    - cat
+    - "{{ .name }}.sentinel"
+  initialDelaySeconds: 5
+  failureThreshold: 30
```
jakesmith, line 329: this could be too low for Dali startup with a big store.

When does it start probing? Is it only once the pod has reported as started?

richardkchapman: The startup probe will (I assume) start when the pod is marked as "running" 30 here means 30 retries, so timeout is actually 30*5

We may as you say want to make it possible for us - or helm user - to override for different components

jakesmith (approved):

richardkchapman:
@ghalliday

helm/hpcc/templates/_helpers.tpl
```diff
@@ -313,6 +313,30 @@ Add Secret volume for a component
...
+{{- define "hpcc.addSentinelProbes" -}}
+env:
+- name: "SENTINEL"
+  value: "{{ .name }}.sentinel"
+startupProbe:
+  exec:
+    command:
+    - cat
+    - "{{ .name }}.sentinel"
+  initialDelaySeconds: 5
```
ghalliday, line 328: @richardkchapman this might be too high for some components - e.g. delaying how quickly the system can scale.  Is there a jira to improve the configuration?

richardkchapman:
@ghalliday I have removed the initialDelay - it's not needed.

Re making the startupProbe configurable - I'll create a Jira to do so but it's not clear yet whether we are likely to need it or which components might do so...

PR #14101, opened 2020-08-21
HPCC-24648 Record translation not spotting some mismatches

richardkchapman:
Fields that name a nested record (or ifblock) in src but some other
type of field in destination (or vice versa) should give a "cannot
translate" error, but are currently treated as if the record was
removed / added.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/rtl/eclrtl/rtldynfield.cpp
+++ b/rtl/eclrtl/rtldynfield.cpp
@@ -1584,12 +1584,15 @@ private:
     void createMatchInfo()
     {
         unsigned defaulted = 0;
+        bool destHasNested = destRecInfo.hasNested();
+        bool sourceHasNested = sourceRecInfo.hasNested();
         for (unsigned idx = 0; idx < destRecInfo.getNumFields(); idx++)
         {
             const RtlFieldInfo *field = destRecInfo.queryField(idx);
             const RtlTypeInfo *type = field->type;
             MatchInfo &info = matchInfo[idx];
-            info.matchIdx = sourceRecInfo.getFieldNum(destRecInfo.queryName(idx));
+            const char *name = destRecInfo.queryName(idx);
+            info.matchIdx = sourceRecInfo.getFieldNum(name);
             if (info.matchIdx == (unsigned) -1)
             {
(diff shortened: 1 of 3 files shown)


jakesmith (approved):

richardkchapman:
@ghalliday Please sanity-check

rtl/eclrtl/rtldynfield.cpp
```diff
@@ -1602,6 +1605,24 @@ class GeneralRecordTranslator : public CInterfaceOf<IDynamicTransform>
                 if ((field->flags & RFTMispayloadfield) == 0)
                     matchFlags |= match_keychange;
                 defaulted++;
+                // If dest field is in a nested record, we need to check that there's no "non-record" field in source matching current nested record name
+                if (destHasNested && name)
```
ghalliday, line 1609: minor: I don't think name can ever be null

ghalliday (commented):
@richardkchapman looks good.  One minor comment.

rtl/eclrtl/rtldynfield.cpp
richardkchapman: Probably not in the expanded case. But is there any harm in leaving the check in?

ghalliday: only that it was slightly confusing, and coverity might start being picky.
Also, should it target 7.8.x?

richardkchapman:
@ghalliday Rebased to 7.8.x

rtl/eclrtl/rtldynfield.cpp
ghalliday: @richardkchapman coverity may complain that queryOriginalField(name) will crash if name is null.  Should also protect that or remove this test.

richardkchapman: I have moved the test. I'm wary of removing it...

PR #14108, opened 2020-08-25
HPCC-24656 Add support for resource limits in helm

shamser:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/common/workunit/workunit.cpp
+++ b/common/workunit/workunit.cpp
@@ -14038,25 +14038,26 @@ bool executeGraphOnLingeringThor(IConstWorkUnit &workunit, const char *graphName
     return false;
 }
 
-static void setResources(StringBuffer &jobYaml, const IConstWorkUnit *workunit, const char *process)
+static void setResources(IPropertyTree *workerConfig, const IConstWorkUnit *workunit, const char *process)
 {
+    auto setResourcesItem = [&workerConfig](const char *category, const char *resourceName,  unsigned value, const char *units)
+    {
+        if (!value) return;
+        VStringBuffer xpath("spec/template/spec/containers/resources/%s@%s", category, resourceName);
+        ensurePTree(workerConfig, xpath.str());
+
+        VStringBuffer v("%u%s", value, units);
+        workerConfig->setProp(xpath.str(), v.str());
+    };
+
     StringBuffer s;
-    unsigned memRequest = workunit->getDebugValueInt(s.clear().appendf("%s-memRequest", process), 0);
-    unsigned memLimit = workunit->getDebugValueInt(s.clear().appendf("%s-memLimit", process), 0);
-    if (memLimit && memLimit < memRequest)
-        memLimit = memRequest;
-    if (memRequest)
-        jobYaml.replaceString("#request-memory", s.clear().appendf("memory: \"%uMi\"", memRequest));
-    if (memLimit)
-        jobYaml.replaceString("#limit-memory", s.clear().appendf("memory: \"%uMi\"", memLimit));
-    unsigned cpuRequest = workunit->getDebugValueInt(s.clear().appendf("%s-cpuRequest", process), 0);
-    unsigned cpuLimit = workunit->getDebugValueInt(s.clear().appendf("%s-cpuLimit", process), 0);
-    if (cpuLimit && cpuLimit < cpuRequest)
-        cpuLimit = cpuRequest;
-    if (cpuRequest)
-        jobYaml.replaceString("#request-cpu", s.clear().appendf("cpu: \"%um\"", cpuRequest));
-    if (cpuLimit)
-        jobYaml.replaceString("#limit-cpu", s.clear().appendf("cpu: \"%um\"", cpuLimit));
+    unsigned memRequest = workunit->getDebugValueInt(s.clear().appendf("resource-%s-memory", process), 0);
+    setResourcesItem("requests", "memory", memRequest, "Mi");
+    setResourcesItem("limits", "memory", memRequest, "Mi");
+
+    unsigned cpuRequest = workunit->getDebugValueInt(s.clear().appendf("resource-%s-cpu", process), 0);
+    setResourcesItem("requests", "cpu", cpuRequest, "m");
+    setResourcesItem("limits", "cpu", cpuRequest, "m");
 }
 
 
@@ -14152,13 +14153,23 @@ bool applyK8sYaml(const char *componentName, const char *wuid, const char *job, 
     }
     jobYaml.replaceString("%args", args.str());
 
+// Disable ability change resources from within workunit
+// - all values are unquoted by toYAML.  This caused problems when prevously string values are
+// outputted unquoted and then treated as a non string type -e.g. labels in metadata.
+// - Also, ability to control if and how much users may change resources should be provided.
+#if 0
     Owned<IWorkUnitFactory> factory = getWorkUnitFactory();
     if (factory)
     {
         Owned<IConstWorkUnit> workunit = factory->openWorkUnit(wuid);
         if (workunit)
-            setResources(jobYaml, workunit, componentName);
+        {
+            Owned<IPropertyTree> workerConfig = createPTreeFromYAMLString(jobYaml.length(), jobYaml.str(), 0, ptr_none, nullptr);
+            setResources(workerConfig, workunit, componentName);
+            toYAML(workerConfig, jobYaml.clear(), 2, 0);
+        }
     }
+#endif
 
     StringBuffer output, error;
     unsigned ret = runExternalCommand(componentName, output, error, "kubectl replace --force -f -", jobYaml.str());
--- a/helm/hpcc/templates/_helpers.tpl
+++ b/helm/hpcc/templates/_helpers.tpl
@@ -593,3 +593,17 @@ resources:
 {{ toYaml .me | indent 4 }}
 {{- end }}
 {{- end -}}
+
+{{/*
+Add resources object for stub pods
+Pass in a dictionary with me defined
+*/}}
+{{- define "hpcc.addStubResources" }}
+resources:
+  requests:
+    cpu: "50m"
+    memory: "50M"
+  limits:
+    cpu: "50m"
+    memory: "50M"
+{{- end -}}
--- a/helm/hpcc/templates/_helpers.tpl
+++ b/helm/hpcc/templates/_helpers.tpl
@@ -579,3 +579,17 @@ Generate list of available services
   public: {{ $esp.public }}
 {{ end -}}
 {{- end -}}
+
+{{/*
+Add resource object
+Pass in a dictionary with me defined
+*/}}
+{{- define "hpcc.addResources" }}
+{{- if .me  }}
+resources:
+  requests:
+{{ toYaml .me | indent 4 }}
+  limits:
+{{ toYaml .me | indent 4 }}
+{{- end }}
+{{- end -}}
--- a/helm/hpcc/templates/dali.yaml
+++ b/helm/hpcc/templates/dali.yaml
@@ -28,6 +28,7 @@ spec:
               ]
 {{ include "hpcc.addSentinelProbes" . | indent 8 }}
 {{- include "hpcc.addSecurityContext" (dict "root" $ "me" .) | indent 8 }}
+{{- include "hpcc.addResources" (dict "me" .resources) | indent 8 }}
 {{ include "hpcc.addImageAttrs" (dict "root" $ "me" .) | indent 8 }}
         volumeMounts:
         - name: {{ .name }}-configmap-volume
--- a/helm/hpcc/templates/eclagent.yaml
+++ b/helm/hpcc/templates/eclagent.yaml
@@ -30,6 +30,17 @@ spec:
                 {{ include "hpcc.daliArg" $ }}
               ]
 {{- include "hpcc.addSecurityContext" (dict "root" $ "me" .) | indent 8 }}
+{{- if .useChildProcesses }}
+{{- include "hpcc.addResources" (dict "me" .resources) | indent 8 }}
+{{- else }}
+        resources:
+          requests:
+            cpu: "50m"
+            memory: "50M"
+          limits:
+            cpu: "50m"
+            memory: "50M"
+{{- end }}
 {{ include "hpcc.addImageAttrs" (dict "root" $ "me" .) | indent 8 }}
         volumeMounts:
 {{ include "hpcc.addConfigMapVolumeMount" . | indent 8 }}
@@ -77,13 +88,7 @@ data:
           - name: {{ $apptype }}-%jobname
 {{- include "hpcc.addSecurityContext" (dict "root" $ "me" .) | indent 12 }}
 {{ include "hpcc.addImageAttrs" (dict "root" $ "me" .) | indent 12 }}
-            resources:
-              requests:
-                #request-memory
-                #request-cpu
-              limits:
-                #limit-memory
-                #limit-cpu
+{{- include "hpcc.addResources" (dict "me" .resources) | indent 12 }}
             workingDir: /var/lib/HPCCSystems
             command: [ {{ $apptype | quote }} ]
             args: [
--- a/helm/hpcc/templates/eclagent.yaml
+++ b/helm/hpcc/templates/eclagent.yaml
@@ -33,13 +33,7 @@ spec:
 {{- if .useChildProcesses }}
 {{- include "hpcc.addResources" (dict "me" .resources) | indent 8 }}
 {{- else }}
-        resources:
-          requests:
-            cpu: "50m"
-            memory: "50M"
-          limits:
-            cpu: "50m"
-            memory: "50M"
+{{- include "hpcc.addStubResources" . | indent 8 }}
 {{- end }}
 {{ include "hpcc.addImageAttrs" (dict "root" $ "me" .) | indent 8 }}
         volumeMounts:
--- a/helm/hpcc/templates/eclccserver.yaml
+++ b/helm/hpcc/templates/eclccserver.yaml
@@ -30,6 +30,17 @@ spec:
                 {{ include "hpcc.daliArg" $ }}
               ]
 {{- include "hpcc.addSecurityContext" (dict "root" $ "me" .) | indent 8 }}
+{{- if .useChildProcesses }}
+{{- include "hpcc.addResources" (dict "me" .resources) | indent 8 }}
+{{- else }}
+        resources:
+          requests:
+            cpu: "50m"
+            memory: "50M"
+          limits:
+            cpu: "50m"
+            memory: "50M"
+{{- end }}
 {{ include "hpcc.addImageAttrs" (dict "root" $ "me" .) | indent 8 }}
         volumeMounts:
 {{ include "hpcc.addConfigMapVolumeMount" . | indent 8 }}
@@ -78,14 +89,8 @@ data:
           containers:
           - name: compile-%jobname
 {{- include "hpcc.addSecurityContext" (dict "root" $ "me" .) | indent 12 }}
+{{- include "hpcc.addResources" (dict "me" .resources) | indent 12 }}
 {{ include "hpcc.addImageAttrs" (dict "root" $ "me" .) | indent 12 }}
-            resources:
-              requests:
-                #request-memory
-                #request-cpu
-              limits:
-                #limit-memory
-                #limit-cpu
             workingDir: /var/lib/HPCCSystems
             command: [ eclccserver ]
             args: [
--- a/helm/hpcc/templates/eclccserver.yaml
+++ b/helm/hpcc/templates/eclccserver.yaml
@@ -33,13 +33,7 @@ spec:
 {{- if .useChildProcesses }}
 {{- include "hpcc.addResources" (dict "me" .resources) | indent 8 }}
 {{- else }}
-        resources:
-          requests:
-            cpu: "50m"
-            memory: "50M"
-          limits:
-            cpu: "50m"
-            memory: "50M"
+{{- include "hpcc.addStubResources" . | indent 8 }}
 {{- end }}
 {{ include "hpcc.addImageAttrs" (dict "root" $ "me" .) | indent 8 }}
         volumeMounts:
--- a/helm/hpcc/templates/roxie.yaml
+++ b/helm/hpcc/templates/roxie.yaml
@@ -27,6 +27,7 @@ spec:
       - name: {{ $toponame | quote }}
 {{ include "hpcc.addSentinelProbes" $roxie | indent 8 }}
 {{- include "hpcc.addSecurityContext" (dict "root" $ "me" .) | indent 8 }}
+{{- include "hpcc.addResources" (dict "me" $roxie.resources) | indent 8 }}
 {{ include "hpcc.addImageAttrs" (dict "root" $ "me" $roxie) | indent 8 }}
         workingDir: /var/lib/HPCCSystems
         command: [ toposerver ] 
@@ -160,6 +161,7 @@ spec:
 {{- end }}
 {{ include "hpcc.addSentinelProbes" $roxie | indent 8 }}
 {{ include "hpcc.addSecurityContext" (dict "root" $ "me" .) | indent 8 }}
+{{- include "hpcc.addResources" (dict "me" $roxie.resources) | indent 8 }}
 {{ include "hpcc.addImageAttrs" (dict "root" $ "me" $roxie) | indent 8 }}
         volumeMounts:
 {{ include "hpcc.addConfigMapVolumeMount" $roxie | indent 8 }}
@@ -226,6 +228,7 @@ spec:
 {{ include "hpcc.addSentinelProbes" $roxie | indent 8 }}
 {{- end }}
 {{ include "hpcc.addSecurityContext" (dict "root" $ "me" .) | indent 8 }}
+{{- include "hpcc.addResources" (dict "me" $roxie.resources) | indent 8 }}
 {{ include "hpcc.addImageAttrs" (dict "root" $ "me" $roxie) | indent 8 }}
         volumeMounts:
 {{ include "hpcc.addConfigMapVolumeMount" $roxie | indent 8 }}
--- a/helm/hpcc/templates/thor.yaml
+++ b/helm/hpcc/templates/thor.yaml
@@ -42,13 +42,11 @@ spec:
                 {{ printf "\"--name=%s\"" .name }}
               ]
 {{- include "hpcc.addSecurityContext" (dict "root" $ "me" .) | indent 8 }}
-        resources:
-          requests:
-            cpu: "50m"
-            memory: "50M"
-          limits:
-            cpu: "50m"
-            memory: "50M"
+{{- if .useChildProcesses }}
+{{- include "hpcc.addResources" (dict "me" $thorScope.eclAgentResources) | indent 8 }}
+{{- else }}
+{{- include "hpcc.addStubResources" . | indent 8 }}
+{{- end }}
 {{ include "hpcc.addImageAttrs" (dict "root" $ "me" .) | indent 8 }}
         volumeMounts:
 {{ include "hpcc.addConfigMapVolumeMount" . | indent 8 }}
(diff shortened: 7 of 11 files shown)


helm/hpcc/templates/_helpers.tpl
```diff
@@ -501,3 +501,19 @@ Generate instance queue names
  {{- end }}
 {{- end -}}
 {{- end -}}
+
+{{/*
+Add cpu resource request
+Pass in a dictionary with me defined
```
richardkchapman, line 507: With me and root defined...

shamser:
@richardkchapman Please can you review

common/workunit/workunit.cpp
```diff
@@ -14024,23 +14024,52 @@ bool executeGraphOnLingeringThor(IConstWorkUnit &workunit, const char *graphName
 
 static void setResources(StringBuffer &jobYaml, const IConstWorkUnit *workunit, const char *process)
 {
+    class GenerateResources
```
richardkchapman, line 14027: Feels like you are hand-rolling a lambda here - perhaps a lambda would be clearer?

richardkchapman: Actually I'm not sure what this code is doing. Took me a couple of reads to work out how it compiled!

common/workunit/workunit.cpp
```diff
@@ -14024,23 +14024,52 @@ bool executeGraphOnLingeringThor(IConstWorkUnit &workunit, const char *graphName
...
+    {
+    public:
+        GenerateResources(StringBuffer &_jobYaml, IPropertyTree *_cconfig): jobYaml(_jobYaml), cconfig(_cconfig) {};
+
+        void setResourceValue(const char *category, const char *resourceName,  unsigned wuvalue, const char *units)
+        {
+            if (!wuvalue && !cconfig)
+                return;
+            StringBuffer tag2replace;
```
richardkchapman, line 14036: VStringBuffer is often a clearer way to code this sort of thing.

common/workunit/workunit.cpp
```diff
@@ -14024,23 +14024,52 @@ bool executeGraphOnLingeringThor(IConstWorkUnit &workunit, const char *graphName
...
+            tag2replace.setf("#%s-%s", category, resourceName);
+
+            StringBuffer s;
```
richardkchapman, line 14039: clear() redundant - again looks like a VStringBuffer in disguise

helm/hpcc/templates/_helpers.tpl
```diff
@@ -556,3 +556,20 @@ Generate list of available services
...
+
+
+{{/*
+Add resource object
+Pass in a dictionary with root and me defined
+*/}}
+{{- define "hpcc.addResources" }}
+{{- if .me.resources }}
+resources:
+{{- toYaml .me.resources | nindent 2 -}}
```
richardkchapman, line 568: I think you want indent here rather than nindent (and change the {{- to {{

common/workunit/workunit.cpp
```diff
@@ -14024,23 +14024,52 @@ bool executeGraphOnLingeringThor(IConstWorkUnit &workunit, const char *graphName
...
+                p.appendf("./%s/@%s", category, resourceName);
+                const char *v = cconfig->queryProp(p.str());
+                if (v)
+                    jobYaml.replaceString(tag2replace, s.appendf("\"%s\"", v));
+            }
+        }
+    private:
+        StringBuffer &jobYaml;
+        Owned<IPropertyTree> cconfig;
+    } generateResources (jobYaml, queryComponentConfig().getPropTree("childContainersResources"));
```
richardkchapman, line 14057: I find the way this is coded very confusing. The fact that the calls to setResourceValue also use the config as an input parameter and modify jobYaml is hidden from all the calls...

richardkchapman:
@shamser Mostly looks good but some questions about coding style.

richardkchapman:
@ghalliday Can you also review

common/workunit/workunit.cpp
```diff
@@ -14024,23 +14024,52 @@ bool executeGraphOnLingeringThor(IConstWorkUnit &workunit, const char *graphName
...
     unsigned cpuRequest = workunit->getDebugValueInt(s.clear().appendf("%s-cpuRequest", process), 0);
     unsigned cpuLimit = workunit->getDebugValueInt(s.clear().appendf("%s-cpuLimit", process), 0);
     if (cpuLimit && cpuLimit < cpuRequest)
         cpuLimit = cpuRequest;
-    if (cpuRequest)
-        jobYaml.replaceString("#request-cpu", s.clear().appendf("cpu: \"%um\"", cpuRequest));
-    if (cpuLimit)
-        jobYaml.replaceString("#limit-cpu", s.clear().appendf("cpu: \"%um\"", cpuLimit));
+    generateResources.setResourceValue("requests", "cpu", cpuRequest, "m");
+    generateResources.setResourceValue("limits", "cpu", cpuLimit, "m");
```
ghalliday, line 14072: This feels like it is on the border of being worth generalizing.  E.g. if we support ephemeral-storage, and/or extended resources.

ghalliday: Not a new issues, but what about different contains in the job requiring different resources.  E.g. thormaster v slave v the agent?

helm/hpcc/templates/_helpers.tpl
```diff
@@ -556,3 +556,20 @@ Generate list of available services
...
+{{ else }}
+{{- if .root.Values.global.resources }}
```
ghalliday, line 570: Does it make sense to have global resources?  If so it would make more sense for them to be defined per component type.  Also, they should be merged with the local overrides, otherwise defining one limit locally may disable multiple global limits.

helm/hpcc/templates/eclccserver.yaml
```diff
@@ -81,11 +82,11 @@ data:
 {{ include "hpcc.addImageAttrs" (dict "root" $ "me" .) | indent 12 }}
             resources:
               requests:
```
ghalliday, line 84: Should the whole section be replaced?  E.g. to allow for user-defined resources etc.

helm/hpcc/templates/thor.yaml
```diff
@@ -42,6 +42,7 @@ spec:
                 {{ printf "\"--name=%s\"" .name }}
               ]
 {{- include "hpcc.addSecurityContext" (dict "root" $ "me" .) | indent 8 }}
+{{- include "hpcc.addResources" (dict "root" $ "me" .) | indent 8 }}
```
ghalliday, line 45: It isn't very clear that resources defined for thor will actually apply to the eclagent, and that thor resources need to be configured with childContainerResources

helm/hpcc/templates/eclccserver.yaml
```diff
@@ -30,6 +30,7 @@ spec:
                 {{ include "hpcc.daliArg" $ }}
               ]
 {{- include "hpcc.addSecurityContext" (dict "root" $ "me" .) | indent 8 }}
+{{- include "hpcc.addResources" (dict "root" $ "me" .) | indent 8 }}
```
ghalliday, line 33: This is a bit confusing.  If useChildProcesses is defined the memory for eclccserver is defined one way, if it is false it is defined a different way.  It would be easier to configure and understand if a single setting to apply to either.

ghalliday (commented):
@shamser a few comments.
The main one is that I'm not sure the way of defining them is quite right.  The resources should be defined for the logical components, and then those get mapped to the underlying pods.  E.g. I want to say that eclccserver requires 16GiB of memory, and define it in the same way whether or not child processes are used.  But all agents should only need say 500MiB.
Something to discuss more widely.

helm/hpcc/values.schema.json
```diff
@@ -246,6 +249,30 @@
...
+        "cpu": {
+          "oneOf": [
+            { "type": "string" },
+            { "type": "number" }
+          ]
+        },
+        "memory": {
+          "oneOf": [
+            { "type": "string" },
+            { "type": "integer" }
```
jakesmith, line 264: what is the reason for onef here being "integer" vs "number" for cpu?

helm/hpcc/values.schema.json
```diff
@@ -555,6 +603,12 @@
         "spillPlane": {
           "description": "The storage plane to write spill files to",
           "type": "string"
+        },
+        "resources": {
+          "$ref": "#/definitions/resources"
+        },
+        "childContainersResources": {
```
jakesmith, line 610: I think this may need to become masterContainerResources and slaveContainerResources

helm/hpcc/values.yaml
```diff
@@ -122,6 +135,13 @@ eclagent:
   useChildProcesses: false
   ## type may be 'hthor' (the default) or 'roxie', to specify that the roxie engine rather than the hthor engine should be used for eclagent workunit processing
   type: hthor
+  #childContainersResources:
```
jakesmith, line 138: would "childContainerResources" be clearer ?

helm/hpcc/values.schema.json
richardkchapman: You can request milliCPUs ...

helm/hpcc/values.schema.json
richardkchapman: worker, not slave...

helm/hpcc/values.schema.json
shamser: yes, number allows decimals and you can specify milliCPUs as decimals.  Memory can't be decimal, only integers.

helm/hpcc/templates/eclagent.yaml
```diff
@@ -78,11 +79,11 @@ data:
 {{ include "hpcc.addImageAttrs" (dict "root" $ "me" .) | indent 12 }}
             resources:
               requests:
-                #request-memory
-                #request-cpu
+                #requests-memory
+                #requests-cpu
```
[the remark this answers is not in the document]
shamser: It maps directly to the resource structure in K8s. 
#resources:
#  requests:
#    memory: "1G"
..

It makes my code simpler.  I can pass the category name + limit name and the code can (1) lookup the value for \ChildContainerResources\[category-name]\[limit-name]" and (2)
search  "#[category-name]-[limit-name]" and replace with value.

helm/hpcc/values.schema.json
jakesmith: yes ok, that makes sense.

helm/hpcc/templates/eclccserver.yaml
shamser: It's not straight-forward to support user-defined resources & copying over whole section:
1) Each of these values (requests/memory, requests/cpu, limit/memory & limit/cpu) are generated by merging the values from workunit debug value and the values from ChildContainerResources/slaveContainerResources.  It won't know what to do with any user-defined values.
2) The indentation will need be handled.

To support user-defined resources, is not trivial.  I suggest that we add support for user-defined resources in the next release. @ghalliday what do you think?

helm/hpcc/templates/thor.yaml
shamser: The resources define the limits for pods created directly by helm.  The values are copied over automatically by helm and used to set the limits for pods that it starts.

ChildContainerResources are for any pods created by one of our processes.  Any pods created using *-jobspec.yaml use the ChildContainerResources.

helm/hpcc/templates/eclccserver.yaml
shamser: Eclccserver's memory is always defined by "resources".  The "compile" pod's resources are defined by ChildContainerResources.  How would you suggest it works?

common/workunit/workunit.cpp
```diff
@@ -14024,23 +14024,52 @@ bool executeGraphOnLingeringThor(IConstWorkUnit &workunit, const char *graphName
...
     unsigned memRequest = workunit->getDebugValueInt(s.clear().appendf("%s-memRequest", process), 0);
     unsigned memLimit = workunit->getDebugValueInt(s.clear().appendf("%s-memLimit", process), 0);
     if (memLimit && memLimit < memRequest)
         memLimit = memRequest;
-    if (memRequest)
-        jobYaml.replaceString("#request-memory", s.clear().appendf("memory: \"%uMi\"", memRequest));
-    if (memLimit)
-        jobYaml.replaceString("#limit-memory", s.clear().appendf("memory: \"%uMi\"", memLimit));
+    generateResources.setResourceValue("requests", "memory", memRequest, "Mi");
+    generateResources.setResourceValue("limits", "memory", memLimit, "Mi");
```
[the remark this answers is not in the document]
shamser: We are using the same generic function to set the lines for cpu and memory.  The memory limits defined here are using the "Mi" unit and  the cpu limits are using the "m" units (see below).  Please can you clarify.

helm/hpcc/values.schema.json
shamser: I'm happy to rename to workerContainerResources.  

However, we shouldn't be using "masterContainerResources". "resources" is the standard name for the limits for resources started with k8s.  The "resources" object is copied into the deployment automatically and nothing more needs to be done by the template coding.   If we were to rename "resources" to masterContainerResources, we'd need to rename masterContainerResources to "resources" when generating the deployment.  Furthermore,  we'd be asking users to configure the hpcc pods  in a non-standard way.

helm/hpcc/values.yaml
shamser: How about workerContainerResources?

helm/hpcc/templates/eclagent.yaml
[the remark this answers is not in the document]
jakesmith: >It makes my code simpler. I can pass the category name 

I don't think "requests-memory" vs "request-memory" is a big deal here, 
but should aim for being as clear as possible to the end-user of these settings, rather than for making code slightly simpler.

helm/hpcc/templates/eclccserver.yaml
jakesmith: It may be same as Gavin is saying, but I think the generation of the whole resources: section should be generated here, if needed, and in other similar jobspec places. 

Any per-job overrides can still be picked up from properties and merged in, including any that a user has decided to provide.
That probably means changing the current search and replace code, to load the yaml into a property tree and setting the overrides into the right place in the property tree, as opposed to the search and replace technique currently used.

helm/hpcc/templates/thor.yaml
jakesmith: perhaps it should be flattened, so that user can supply an optional

- jobAgentContainerResources: 
- thorAgentContainerResources:
- thorWorkerContainerResources:
- thorManagerContinerResources:

I suspect the 1st 2, which are v. lightweight would generally not be used.
Each of these can be copied (if present) into the appropriate generates "resources:" place in the generated yaml, along with any workunit overrides if present.

common/workunit/workunit.cpp
```diff
@@ -14018,23 +14018,55 @@ bool executeGraphOnLingeringThor(IConstWorkUnit &workunit, const char *graphName
...
+    
+    if (strcmp(process,"thormaster")==0)
+    {
+        dbglogXML(&queryComponentConfig());
+        cconfig.setown(queryComponentConfig().getPropTree("thorMasterResources"));
+    }
+    else if (strcmp(process,"thorslave")==0) 
+        cconfig.setown(queryComponentConfig().getPropTree("thorSlaveResources"));
+    else
+        cconfig.setown(queryComponentConfig().getPropTree("childContainersResources"));
```
jakesmith, line 14031: Feels like the selection of a section in values etc. and generation of resources should mainly be done by the helm chart templates/helper.
Substitiouon of any overrides can still be achieved in a generic way I think, when processing a jobspec, by matching the locating matching process type properties, parsing them and setting them into the generated and loaded yaml resource section.

helm/hpcc/templates/_helpers.tpl
jakesmith: I agree that would be a good scheme. 
A per component type set of defaults, used as a base, overridden by any per instance definitions, and/or any workunit overrides in the case of jobs (merged in both cases).

helm/hpcc/templates/eclagent.yaml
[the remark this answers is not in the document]
jakesmith: see other comments - issue become moot, if generating whole section.
The workunit properties that match the process and conform to the pattern of our choosing would all be picked up and used to override the generated 'resources' section.
It might be better to use a more generic pattern than the current one, so user can add their own limit/requests, e.g.:
```
#option('resourceLimit-hthor-memory', ...);
#option('resourceRequest-hthor-cpu', ...);
```
i.e. so that the postfix is used as the resources sub-property.

helm/hpcc/templates/thor.yaml
```diff
@@ -119,6 +121,9 @@ data:
 {{ toYaml (omit $thorAgentScope "logging") | indent 6 }}
 {{- include "hpcc.generateLoggingConfig" (dict "root" $ "me" $thorAgentScope) | indent 6 }}
       type: thor
+{{- if not $thor.thorMasterResources -}}
+{{ deepCopy $thor.thorMasterResources | indent 6 }}
```
jakesmith, line 125: this isn't the right place for the thor manager/master resources setting afaics.
This is the thorAgent resource settings.

The thor manager settings need to be generated into the thormaster-jobspec.yaml

helm/hpcc/templates/thor.yaml
shamser: @jakesmith Would these objects be under the Global section?  If that is the case, would overrides be supported under individual components section?

jakesmith: I had only been thinking of them as nodes under the thor instance name (e.g. under thor:).
But yes, I think as with defaults per component type (as per Gavin's point: https://github.com/hpcc-systems/HPCC-Platform/pull/14108#discussion_r487836236), you'd want a way to define global defaults per these component subtype too.

ghalliday:
I think the key thing is to work out the logical way of describing the options in the values.yaml to provide the kind of configuration the user (and the defaults) will need, and then work from that.
E.g. 
thor master/worker/agent all need different values for the different pods.
eclccserver wants a default that will work whether or not child processes are used, same for thor
you want to be able to default the typical memory requirements for a thor, but also allow them to be overridden for a specific thor.
eclagent needs different resources if it is running as hthor, than if it is there to submit graphs to thor.

helm/hpcc/templates/eclccserver.yaml
shamser: I have looked into replacing the current search & replace with a PropertyTree.  I have found more complications.  runK8sJob/LaunchK8sJob are passed a list of string pairs that are search and replaced in the yaml.  For example, a valid call is runK8sJob( ... , { { "graphName", graphName} });

"%graphName" is replaced by the actual graph name.  The problem is that we do not know the path to the searched-for string - do we iterate over every element of the IPropertyTree examining the node value for the tag? Is there a more efficient way? @jakesmith

helm/hpcc/templates/thor.yaml
shamser: That's not a problem.  I have implemented something similar already, although the names are a little different.

helm/hpcc/templates/thor.yaml
shamser: thor-agent reads the values from this section and writes it into thormaster-jobspec.  So, it does eventually end up in thormaster-jobspec.yaml as the resources object. It could be written directly into thormaster-jobspec.yaml as the resources object, if we didn't need to override the values from the workunit.

common/workunit/workunit.cpp
```diff
@@ -14016,25 +14016,32 @@ bool executeGraphOnLingeringThor(IConstWorkUnit &workunit, const char *graphName
     return false;
 }
 
-static void setResources(StringBuffer &jobYaml, const IConstWorkUnit *workunit, const char *process)
+static void setResources(IPropertyTree *workerConfig, const IConstWorkUnit *workunit, const char *process)
 {
+    auto setResourcesItem = [&workerConfig](const char *category, const char *resourceName,  unsigned value, const char *units)
+    {
+        if (!value) return;
+        VStringBuffer xpath("spec/template/spec/containers/resources/%s@%s", category, resourceName);
```
shamser, line 14024: If end-users ever extend the deployment to have multiple containers, I am not sure if this will work.

helm/hpcc/values.schema.json
```diff
@@ -156,6 +171,32 @@
...
+              "type": "string",
+              "pattern" : "^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$"
+            },
+            { "type": "integer" }
+          ]
+        }
+      }
+    },
+    "resources": {
+      "type": "object"
```
shamser, line 198: resources object are no longer validated, in order to allow users to provide arbitary values for limits and requests.

shamser:
@jakesmith Hopefully, this is closer to what you've suggested. Please can you review.

jakesmith:
@shamser - tagging you back for now.

shamser:
@jakesmith @richardkchapman @ghalliday Please can you review.  By the way, I've disabled the code implementing user-specified resource (ability to specify resources with #option).   The reason is that toYAML (which is used by this section) is causing problems.  I've opened a jira to implement user-specified resources: https://track.hpccsystems.com/browse/HPCC-24777

helm/hpcc/templates/_helpers.tpl
```diff
@@ -579,3 +579,17 @@ Generate list of available services
   public: {{ $esp.public }}
 {{ end -}}
 {{- end -}}
+
+{{/*
+Add resource object
+Pass in a dictionary with me defined
+*/}}
+{{- define "hpcc.addResources" }}
+{{- if .me  }}
```
jakesmith, line 588: strictly speaking, should this be checking that the .me dictionary has >0 entries? ( as opposed to e.g. being set to a scalar)

helm/hpcc/templates/dali.yaml
```diff
@@ -28,6 +28,7 @@ spec:
               ]
 {{ include "hpcc.addSentinelProbes" . | indent 8 }}
 {{- include "hpcc.addSecurityContext" (dict "root" $ "me" .) | indent 8 }}
+{{ include "hpcc.addResources" (dict "me" .resources) | indent 8 }}
```
jakesmith: should be {{- I think. Otherwise it's always going to generate a newline here, i.e. even if no resources.

helm/hpcc/templates/thor.yaml
```diff
@@ -156,13 +164,7 @@ data:
...
 {{- include "hpcc.addSecurityContext" (dict "root" $ "me" .) | indent 12 }}
 {{ include "hpcc.addImageAttrs" (dict "root" $ "me" .) | indent 12 }}
-            resources:
-              requests:
-                #request-memory
-                #request-cpu
-              limits:
-                #limit-memory
-                #limit-cpu
+{{- include "hpcc.addResources" (dict "me" $thorScope.agentResources) | indent 12 }}
```
jakesmith, line 167: Didn't we settle on eclagentResources, managerResources and workerResources?
it may be fine, but 'agent' is a bit overloaded

helm/hpcc/templates/thor.yaml
```diff
@@ -81,6 +88,7 @@ spec:
                 {{ printf "\"--name=%s\"" .name }}
               ]
 {{- include "hpcc.addSecurityContext" (dict "root" $ "me" .) | indent 8 }}
+{{- include "hpcc.addResources" (dict "me" $thorScope.agentResources) | indent 8 }}
```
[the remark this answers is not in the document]
jakesmith: Actually, if useChildProcesses is in this context, it would spawn master and slaves as child processes.
Turning on useChildProcesses here is a developer only option really, there are some known issues with it.

Should generate stub resources when off, and when on... all bets are off! (it would need to be sum total of master + N * workers to make any sense)

helm/hpcc/values.yaml
```diff
@@ -144,14 +147,21 @@ eclagent:
...
 - name: roxie-workunit
   replicas: 1
   prefix: roxie_workunit
   maxActive: 20
   useChildProcesses: true
   type: roxie
-    
+  #resources:
+  #  cpu: "1"
+  #  memory: "1014M"
```
jakesmith, line 163: odd amount?

helm/hpcc/values.yaml
```diff
@@ -183,6 +200,9 @@ esp:
   replicas: 1
   servicePort: 8010
   public: false
+  #resources:
+  #  cpu: "2"
```
jakesmith, line 204: these are all a guess/example, but not sure really, but services having more than 1 cpu vs others, may be misleading.

jakesmith:
@shamser - looks good generally, added some minor comments, only 1 significant issue I think, in thor.yaml re.  eclagent resources (https://github.com/hpcc-systems/HPCC-Platform/pull/14108/files#r494910309)

helm/hpcc/templates/dali.yaml
```diff
@@ -28,6 +28,7 @@ spec:
               ]
 {{ include "hpcc.addSentinelProbes" . | indent 8 }}
 {{- include "hpcc.addSecurityContext" (dict "root" $ "me" .) | indent 8 }}
+{{- include "hpcc.addResources" (dict "me" .resources) | indent 8 }}
```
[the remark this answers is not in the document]
shamser: I confirm that not error occurs when resources are not provided.

helm/hpcc/templates/_helpers.tpl
shamser: It wouldn't cause a problem if it did have 0 items.  Empty resources will just be ignored.

helm/hpcc/values.yaml
shamser: This left over from testing.  I wanted to make sure the values were coming from the right place.

shamser:
@jakesmith I've modified as suggested.

helm/hpcc/values.yaml
```diff
@@ -207,6 +233,9 @@ esp:
   replicas: 1
   public: true
   servicePort: 8510
+  #resources:
+  #  cpu: "2"
```
jakesmith, line 237: trivial: not sure any of these esp services warrant 2 cpus vs e.g. myeclccserver having just 1.

helm/hpcc/templates/thor.yaml
```diff
@@ -81,6 +86,9 @@ spec:
                 {{ printf "\"--name=%s\"" .name }}
               ]
 {{- include "hpcc.addSecurityContext" (dict "root" $ "me" .) | indent 8 }}
+{{- if not .useChildProcesses }}
+{{- include "hpcc.addStubResources" . | indent 8 }}
```
jakesmith, line 90: perhaps should put a comment in here for the useChildProcess: true case, to say something like:
userChildProcesses=true in Thor is an experimental option only. It will cause manager and worker processes will be launched in this pod, and the pod would require enough resource for them all.

jakesmith:
@shamser - added a few trivial comments.

helm/hpcc/templates/roxie.yaml
```diff
@@ -27,6 +27,7 @@ spec:
       - name: {{ $toponame | quote }}
 {{ include "hpcc.addSentinelProbes" $roxie | indent 8 }}
 {{- include "hpcc.addSecurityContext" (dict "root" $ "me" .) | indent 8 }}
+{{- include "hpcc.addResources" (dict "me" $roxie.resources) | indent 8 }}
```
jakesmith, line 30: This is resources for the toposerver, I think this should pickup a different resource section.
@richardkchapman should there be resource sections like Thor, for topo, server replicas, and channel replicas?

richardkchapman: toposerver needs minimal resources, like an agentexec stub, and does not need to be configurable

jakesmith: ok, so need to generate the stub resources here for topo.
Should there be separate resource sections for server and channel replicas?

helm/hpcc/templates/_helpers.tpl
```diff
@@ -579,3 +579,31 @@ Generate list of available services
...
+Pass in a dictionary with me defined
+*/}}
+{{- define "hpcc.addStubResources" }}
+resources:
+  requests:
+    cpu: "50m"
+    memory: "50M"
+  limits:
+    cpu: "50m"
+    memory: "50M"
```
jakesmith, line 608: I was testing this yesterday, and at least for the stubs that fork and call kubectl, 50M is not enough.
kubectl ends up getting killed by cgroup oom policy.
In my tests it looks like kubectl can consume about 28M.

I think changing this to 100M should be enough.

jakesmith: We don't need to generate 'requests' if they are always going to be same as limits as they are here.
If limits, but no requests are specified, the requests are implicitly the same as the limits.

PR #14122, opened 2020-08-28
HPCC-24546 Support for Hashicorp vaults and kubernetes secrets

afishbeck:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/common/thorhelper/thorsoapcall.cpp
+++ b/common/thorhelper/thorsoapcall.cpp
@@ -18,6 +18,7 @@
 #include "jliball.hpp"
 #include "jqueue.tpp"
 #include "jisem.hpp"
+#include "jsecrets.hpp"
 
 #include "rtlformat.hpp"
 
@@ -279,7 +280,7 @@ public:
         free(fullText);
     }
 
-    unsigned getUrls(UrlArray &array)
+    unsigned getUrls(UrlArray &array, const char *basic_credentials=nullptr)
     {
         if (fullText)
         {
@@ -289,7 +290,10 @@ public:
             char *url = strtok_r(copyFullText, "|", &saveptr);
             while (url != NULL)
             {
-                array.append(*new Url(url));
+                Owned<Url> item = new Url(url);
+                if (basic_credentials)
+                    item->userPasswordPair.set(basic_credentials);
+                array.append(*item.getClear());
                 url = strtok_r(NULL, "|", &saveptr);
             }
 
@@ -805,8 +809,11 @@ private:
     static CriticalSection secureContextCrit;
     static Owned<ISecureSocketContext> secureContext;
 
+    Owned<ISecureSocketContext> customSecureContext;
+
     CTimeMon timeLimitMon;
     bool complete, timeLimitExceeded;
+    bool customClientCert = false;
     IRoxieAbortMonitor * roxieAbortMonitor;
 
 protected:
@@ -863,15 +870,6 @@ public:
         {
             s.setown(helper->getXpathHintsXml());
             xpathHints.setown(createPTreeFromXMLString(s.get()));
-        }
-
-        StringAttr proxyAddress;
-        proxyAddress.set(s.setown(helper->getProxyAddress()));
-        if (!proxyAddress.isEmpty())
-        {
-            UrlListParser proxyUrlListParser(proxyAddress);
-            if (0 == proxyUrlListParser.getUrls(proxyUrlArray))
-                throw MakeStringException(0, "SOAPCALL PROXYADDRESS specified no URLs");
         }
 
         if (wscType == STsoap)
@@ -941,45 +939,99 @@ public:
         else
             rowTransformer = NULL;
 
+        StringBuffer proxyAddress;
+        proxyAddress.set(s.setown(helper->getProxyAddress()));
+
         OwnedRoxieString hosts(helper->getHosts());
-        UrlListParser urlListParser(hosts);
-        if ((numUrls = urlListParser.getUrls(urlArray)) > 0)
+        if (isEmptyString(hosts))
+            throw MakeStringException(0, "%sCALL specified no URLs",wscType == STsoap ? "SOAP" : "HTTP");
+        if (0==strncmp(hosts, "secret:", 7))
         {
-            if (wscMode == SCrow)
+            const char *finger = hosts.get()+7;
+            if (isEmptyString(finger))
+                throw MakeStringException(0, "%sCALL HTTP-CONNECT SECRET specified with no name", wscType == STsoap ? "SOAP" : "HTTP");
+            if (!proxyAddress.isEmpty())
+                throw MakeStringException(0, "%sCALL PROXYADDRESS can't be used with HTTP-CONNECT secrets", wscType == STsoap ? "SOAP" : "HTTP");
+            StringAttr vaultId;
+            const char *thumb = strchr(finger, ':');
+            if (thumb)
             {
-                numRowThreads = 1;
-
-                numUrlThreads = helper->numParallelThreads();
-                if (numUrlThreads == 0)
-                    numUrlThreads = 1;
-                else if (numUrlThreads > MAXWSCTHREADS)
-                    numUrlThreads = MAXWSCTHREADS;
-
-                numRecordsPerBatch = 1;
+                vaultId.set(finger, thumb-finger);
+                finger = thumb + 1;
             }
-            else
+            StringBuffer secretName("http-connect-");
+            secretName.append(finger);
+            Owned<IPropertyTree> secret = (vaultId.isEmpty()) ? getSecret("ecl", secretName) : getVaultSecret("ecl", vaultId, secretName, nullptr);
+            if (!secret)
+                throw MakeStringException(0, "%sCALL %s SECRET not found", wscType == STsoap ? "SOAP" : "HTTP", secretName.str());
+
+            StringBuffer url;
+            getSecretKeyValue(url, secret, "url");
+            if (url.isEmpty())
+                throw MakeStringException(0, "%sCALL %s HTTP SECRET must contain url", wscType == STsoap ? "SOAP" : "HTTP", secretName.str());
+            UrlListParser urlListParser(url);
+            StringBuffer auth;
+            getSecretKeyValue(auth, secret, "username");
+            if (auth.length())
             {
-                unsigned totThreads = helper->numParallelThreads();
-                if (totThreads < 1)
-                    totThreads = 2; // default to 2 threads
-                else if (totThreads > MAXWSCTHREADS)
-                    totThreads = MAXWSCTHREADS;
-
-                numUrlThreads = (numUrls < totThreads)? numUrls: totThreads;
-
-                numRowThreads = totThreads / numUrlThreads;
-                if (numRowThreads < 1)
-                    numRowThreads = 1;
-                else if (numRowThreads > MAXWSCTHREADS)
-                    numRowThreads = MAXWSCTHREADS;
-
-                numRecordsPerBatch = helper->numRecordsPerBatch();
-                if (numRecordsPerBatch < 1)
-                    numRecordsPerBatch = 1;
+                if (strchr(auth, ':'))
+                    throw MakeStringException(0, "%sCALL HTTP-CONNECT SECRET username contains illegal colon", wscType == STsoap ? "SOAP" : "HTTP");
+                auth.append(':');
+                getSecretKeyValue(auth, secret, "password");
             }
+            urlListParser.getUrls(urlArray, auth);
+            proxyAddress.set(secret->queryProp("proxy"));
+            getSecretKeyValue(proxyAddress.clear(), secret, "proxy");
         }
         else
+        {
+            UrlListParser urlListParser(hosts);
+            urlListParser.getUrls(urlArray);
+        }
+
+        numUrls = urlArray.ordinality();
+        if (numUrls == 0)
             throw MakeStringException(0, "%sCALL specified no URLs",wscType == STsoap ? "SOAP" : "HTTP");
+
+        if (!proxyAddress.isEmpty())
+        {
+            UrlListParser proxyUrlListParser(proxyAddress);
+            if (0 == proxyUrlListParser.getUrls(proxyUrlArray))
+                throw MakeStringException(0, "%sCALL proxy address specified no URLs",wscType == STsoap ? "SOAP" : "HTTP");
+        }
+
+        if (wscMode == SCrow)
+        {
+            numRowThreads = 1;
+
+            numUrlThreads = helper->numParallelThreads();
+            if (numUrlThreads == 0)
+                numUrlThreads = 1;
+            else if (numUrlThreads > MAXWSCTHREADS)
+                numUrlThreads = MAXWSCTHREADS;
+
+            numRecordsPerBatch = 1;
+        }
+        else
+        {
+            unsigned totThreads = helper->numParallelThreads();
+            if (totThreads < 1)
+                totThreads = 2; // default to 2 threads
+            else if (totThreads > MAXWSCTHREADS)
+                totThreads = MAXWSCTHREADS;
+
+            numUrlThreads = (numUrls < totThreads)? numUrls: totThreads;
+
+            numRowThreads = totThreads / numUrlThreads;
+            if (numRowThreads < 1)
+                numRowThreads = 1;
+            else if (numRowThreads > MAXWSCTHREADS)
+                numRowThreads = MAXWSCTHREADS;
+
+            numRecordsPerBatch = helper->numRecordsPerBatch();
+            if (numRecordsPerBatch < 1)
+                numRecordsPerBatch = 1;
+        }
 
         for (unsigned i=0; i<numRowThreads; i++)
             threads.append(*new CWSCHelperThread(this));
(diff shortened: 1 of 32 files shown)


afishbeck:
@richardkchapman @ghalliday please review.

The helm/examples/secrets/README.md file is meant to be a very easy to use walk through of setting up and using vault and kubernetes secrets.. please include usability comments in your review.

I've kept the secret and vault functions in jutil.cpp for now, but can move them once we've agreed on the approach.

I've added open source cpp-http code for my REST calls.  The nice thing is it's all in one header file.  But when I move the code I could use our CHttpClient code.  I didn't want to link esphttp to jlib.  And this library is a bit more low level.

afishbeck:
Another option for REST calls could be boost.. I didn't want to deal with large dependencies though.

afishbeck:
Removed tests that were only meant for my environment, and updated what seems to be arbitrary cpp-httplib openssl version deprecation.

AttilaVamos:
I am sure the schedule2 failure isn't related to this PR.

afishbeck:
@richardkchapman @ghalliday

afishbeck:
Updated caching logic to check all viable caching locations before going disk or vaults.

system/jlib/jutil.cpp
```diff
@@ -3030,19 +3051,117 @@ int getEnum(const char *v, const EnumMapping *map, int defval)
...
+{
+    if(!url || strlen(url) <= 7)
+        throw MakeStringException(-1, "Invalid URL %s", url);
+    if (0 == strncasecmp(url, "HTTPS://", 8))
+        https = true;
+    else if (0 == strncasecmp(url, "HTTP://", 7))
+        https = false;
+    else
+        throw MakeStringException(-1, "Invalid URL, no protocol specified %s", url);
+    if (https)
```
ghalliday, line 3105: trivial: curious why this code is separated from the code above.

system/jlib/jutil.cpp
```diff
@@ -3030,19 +3051,117 @@ int getEnum(const char *v, const EnumMapping *map, int defval)
...
+    }
+}
+
+static inline void extractUrlProtocol(const char *&url, bool &https, StringBuffer *scheme)
+{
+    if(!url || strlen(url) <= 7)
+        throw MakeStringException(-1, "Invalid URL %s", url);
+    if (0 == strncasecmp(url, "HTTPS://", 8))
+        https = true;
+    else if (0 == strncasecmp(url, "HTTP://", 7))
```
ghalliday, line 3101: question: are there likely to be other protocols?

system/jlib/jutil.cpp
```diff
@@ -3056,19 +3175,481 @@ extern jlib_decl void setSecretMount(const char * path)
...
+    {
+        CriticalBlock block(secretCacheCS);
+        secret.setown(secretCache->getPropTree(name));
+        if (secret)
+        {
+            unsigned created = (unsigned) secret->getPropInt("@created");
+            if (created && (created < timeoutThreshold))
+            {
+                secretCache->removeProp(name);
+                puts("\nremoved from local cache\n");
```
ghalliday, line 3486: minor: Debug tracing that should probably be removed/cleaned up.

system/jlib/jutil.cpp
```diff
@@ -3056,19 +3175,481 @@ extern jlib_decl void setSecretMount(const char * path)
...
+        return nullptr;
+    unsigned timeoutThreshold = msTick() - getSecretTimeout();
+    Owned<IPropertyTree> secret;
+    {
+        CriticalBlock block(secretCacheCS);
+        secret.setown(secretCache->getPropTree(name));
+        if (secret)
+        {
+            unsigned created = (unsigned) secret->getPropInt("@created");
+            if (created && (created < timeoutThreshold))
```
ghalliday, line 3483: nit picking: This could cause wrapping issues if created is a very large number - it could never time out.  You need to check if now - created > getSecretTimeout().

system/jlib/jutil.cpp
```diff
@@ -3056,19 +3175,481 @@ extern jlib_decl void setSecretMount(const char * path)
...
+
+static void addCachedLocalSecret(const char *name, IPropertyTree *secret)
+{
+    if (!secret || isEmptyString(name))
+        return;
+    secret->setPropInt("@created", (int)msTick());
+    {
+        CriticalBlock block(secretCacheCS);
+        secretCache->setPropTree(name, LINK(secret));
+        puts("\nadded to local cache\n");
```
ghalliday, line 3504: more debug tracing.  (Should be moved out of the cs if kept).

system/jlib/jutil.cpp
```diff
@@ -3056,19 +3175,481 @@ extern jlib_decl void setSecretMount(const char * path)
...
+IVaultManager *ensureVaultManager()
+{
+    CriticalBlock block(secretCS);
+    if (!vaultManager)
+        vaultManager.setown(new CVaultManager());
+    return vaultManager;
+}
+
+CriticalSection secretCacheCS;
+static Owned<IPropertyTree> secretCache = createPTree();
```
ghalliday, line 3469: need to be careful about construction ordering.  I'm not sure how much of a problem it is these days, but safer to use a MODULE_INIT to initialize because jptree has some interesting initialization.

system/jlib/jutil.cpp
```diff
@@ -3056,19 +3175,481 @@ extern jlib_decl void setSecretMount(const char * path)
...
+static Owned<CVaultManager> vaultManager;
+IVaultManager *ensureVaultManager()
+{
+    CriticalBlock block(secretCS);
+    if (!vaultManager)
+        vaultManager.setown(new CVaultManager());
+    return vaultManager;
+}
+
+CriticalSection secretCacheCS;
```
ghalliday, line 3468: minor: add static

system/jlib/jutil.cpp
```diff
@@ -3056,19 +3175,481 @@ extern jlib_decl void setSecretMount(const char * path)
...
+    tree->setPropInt("@created", (int) msTick());
+    ForEach(*dir)
+    {
+        StringBuffer name;
+        dir->getName(name);
+        if (!validateXMLTag(name))
+            continue;
+        StringBuffer value;
+        value.loadFile(&dir->query());
+        tree->setProp(name, value.str());
```
ghalliday, line 3533: What implications does this have for binary secrets?  Is it ok to assume they will never be used?

system/jlib/jutil.cpp
```diff
@@ -3056,19 +3175,481 @@ extern jlib_decl void setSecretMount(const char * path)
...
+    {
+        if (!vaultmgr->getCachedSecretFromVault(category, vaultId, kind, json, name, version))
+            vaultmgr->requestSecretFromVault(category, vaultId, kind, json, name, version);
+    }
+    return createPTreeFromVaultSecret(json.str(), kind);
+}
+
+extern jlib_decl IPropertyTree *getSecret(const char *category, const char * name)
+{
+    //check for any chached first
```
ghalliday, line 3623: typo: cached

system/jlib/jutil.cpp
```diff
@@ -3056,19 +3175,481 @@ extern jlib_decl void setSecretMount(const char * path)
...
+    StringBuffer schemeHostPort;
+    StringBuffer path;
+    StringBuffer username;
+    StringBuffer password;
+    StringAttr name;
+    StringAttr role;
+    StringAttr token;
+
+public:
+    CVault(const char *name_, IPropertyTree *vault) : name(name_)
```
ghalliday, line 3213: trivial: code elsewhere uses a prefix of _ for clashing parameter names.

system/jlib/jutil.cpp
```diff
@@ -3056,19 +3175,481 @@ extern jlib_decl void setSecretMount(const char * path)
...
+        StringBuffer login_token;
+        login_token.loadFile("/var/run/secrets/kubernetes.io/serviceaccount/token");
+        if (login_token.length())
+        {
+            std::string json;
+            json.append("{\"jwt\": \"").append(login_token.str()).append("\", \"role\": \"").append(role.str()).append("\"}");
+            httplib::Client cli(schemeHostPort.str());
+            if (username.length() && password.length())
+                cli.set_basic_auth(username, password);
+            if (httplib::Result res = cli.Post("/v1/auth/kubernetes/login", json, "application/json"))
```
ghalliday, line 3248: minor: code generally assigns first and then checks.

system/jlib/jutil.cpp
```diff
@@ -3056,19 +3175,481 @@ extern jlib_decl void setSecretMount(const char * path)
...
+        CriticalBlock block(vaultCS);
+        IPropertyTree *tree = cache->queryPropTree(secret);
+        if (tree)
+        {
+            VStringBuffer vername("v.%s", isEmptyString(version) ? "latest" : version);
+            IPropertyTree *envelope = tree->queryPropTree(vername);
+            if (!envelope)
+                return false;
+            unsigned created = (unsigned) envelope->getPropInt("@created");
+            if (created && (created < timeoutThreshold))
```
ghalliday, line 3281: minor: similar problem with wrapping as the local cache

system/jlib/jutil.cpp
```diff
@@ -3056,19 +3175,481 @@ extern jlib_decl void setSecretMount(const char * path)
...
+            {
+                tree->removeTree(envelope);
+                puts("\nremoved from vault cache\n");
```
ghalliday, line 3284: similar comments about tracing with puts

system/jlib/jutil.cpp
```diff
@@ -3056,19 +3175,481 @@ extern jlib_decl void setSecretMount(const char * path)
...
+            }
+            else
+            {
+                DBGLOG("Vault %s error accessing secret %s.%s [%d](%d) - vault: %s - response: %s", name.str(), secret, version ? version : "", res->status, res.error(), name.str(), res->body.c_str());
+            }
+        }
+        return false;
+    }
+};
+class CVaultSet
```
ghalliday, line 3347: minor: newline between classes.

system/jlib/jutil.cpp
```diff
@@ -3056,19 +3175,481 @@ extern jlib_decl void setSecretMount(const char * path)
...
+        {
+            IPropertyTree &vault = iter->query();
+            const char *name = vault.queryName();
+            vaults.emplace(name, std::unique_ptr<CVault>(new CVault(name, &vault)));
+        }
     }
-    //MORE: cache the secret for up to secretTimeoutMs
+    bool getCachedSecret(CVaultKind &kind, StringBuffer &content, const char *secret, const char *version)
+    {
+        std::map<std::string, std::unique_ptr<CVault>>::iterator it = vaults.begin();
```
ghalliday, line 3364: possibly simpler to use "auto" instead of the explicit type (and elsewhere)

system/jlib/jutil.cpp
```diff
@@ -3056,19 +3175,481 @@ extern jlib_decl void setSecretMount(const char * path)
         secretDirectory.set(path);
 }
 
-extern jlib_decl StringBuffer & getSecret(StringBuffer & result, const char * name, const char * key)
+enum class CVaultKind { kv_v1, kv_v2 };
```
ghalliday, line 3178: is it worth splitting this code into a jsecret.hpp/cpp?  (I don't have a strong opinion either way.)

system/jlib/jutil.cpp
```diff
@@ -3030,19 +3051,117 @@ int getEnum(const char *v, const EnumMapping *map, int defval)
...
+    splitUrlAuthority(authority, authorityLen, user, password, schemeHostPort, nullptr);
+}
+
+//---------------------------------------------------------------------------------------------------------------------
+
+
 static StringBuffer secretDirectory;
 static CriticalSection secretCS;
-static unsigned secretTimeoutMs = UINT_MAX;
+static unsigned secretTimeoutMs = 60 * 60 * 1000;
```
ghalliday, line 3155: Is this typical for k8s?  (I never researched what the standard default was.)

ghalliday (commented):
@afishbeck the changes look good to me.  A few comments, but none of them very serious.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -313,6 +313,22 @@ Add Secret volume for a component
...
+Generate vault info
+*/}}
+{{- define "hpcc.generateVaultConfig" -}}
+{{- $categories := .categories -}}
+vaults:
+{{- range  $category, $vault := .root.Values.vaults }}
+  {{ $category }}:
+  {{- range $vaultid, $vault := . }}
+    {{ $vaultid }}:
+      type: {{ $vault.type }}
```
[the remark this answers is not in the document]
afishbeck: Yes, good catch I changed from type to kind but missed it here.  kv-v2 happens to be the default.

system/jlib/jutil.cpp
afishbeck: In terms of vaults I would think most are HTTPS/HTTP accessible.  In terms of general future thinking I could see us supporting protocol buffers / gRPC in the future.  I could remove the https flag for now, since I don't use it.  Or change is to an ssl flag.  I was trying to make functions that might be useful elsewhere.

system/jlib/jutil.cpp
afishbeck: Side effect of how the code evolved, it would be better moved up.

system/jlib/jutil.cpp
afishbeck: Yes I'll split it out.  Wanted an initial sanity check on the approach first.

system/jlib/jutil.cpp
afishbeck: I'll see if there's some consensus somewhere.

system/jlib/jutil.cpp
afishbeck: Forgot to scrub these personal debug statements out.

system/jlib/jutil.cpp
afishbeck: I'll do a little research.

gfortil (changes requested):

helm/examples/secrets/README.md
```diff
@@ -0,0 +1,137 @@
...
+```bash
+kubectl port-forward vault-0 8200:8200
+```
+
+## Configure vault kubernetes auth
+
+Enabling kubernetes auth will allow k8s nodes to access the vault via their kubernetes.io access tokens.
+
+```bash
+vault auth enable kubernetes
```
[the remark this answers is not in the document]
afishbeck: I've moved the login up higher and now enable kubernetes should work from outside the vault-0 pod.

afishbeck:
Updated to support binary secrets, moved secret code to jsecrets.cpp, cleaned up based on review comments.
@gfortil updated README.md based on your comments.  Among other things I now assume user is running from HPCC-Platform/helm.
@stuartort if you want to test when you get to the hpcc helm install line add " --set global.image.root=afishbeck" as follows:

helm install myhpcc hpcc/  --set global.image.root=afishbeck --set global.image.version=latest -f examples/secrets/values-secrets.yaml

afishbeck:
@AttilaVamos not sure what "Conflicting files, should skip build and test" means.  I'll try rebasing.

afishbeck:
rebased

afishbeck:
Also changed the vault yaml format a bit.
@ghalliday please review

afishbeck:
rebased again to resolve conflict with my other PR which was just merged.

afishbeck:
@ghalliday

afishbeck:
@AttilaVamos I doubt those roxie timeouts had much to do with this change, can we re-run the test?

AttilaVamos:
@afishbeck The schedule2 is definitely not related to this PR. The workflow test is new, I should check it. However I think it isn't related, as well based on that test has been successful on in the another  Smoketest environment (Host: ip-10-20-0-221.ca-central-1.compute.internal).

ghalliday:
Added HPCC-24707 to cover the last issue - not likely to be related to tony's change, but worth investigating.

ghalliday:
@ghalliday - to make sure I review it next

system/jlib/jsecrets.cpp
```diff
@@ -0,0 +1,702 @@
...
+        }
+    }
+    CVaultKind getVaultKind() const { return kind; }
+    void kubernetesLogin()
+    {
+        CriticalBlock block(vaultCS);
+        if (token.length())
+            return;
+        StringBuffer login_token;
+        login_token.loadFile("/var/run/secrets/kubernetes.io/serviceaccount/token");
```
ghalliday, line 262: Does this require the pods to have any access rights?  Some recent PRs of Richard's have tied down both the access rights and the network access.  Does that introduce problems here?

ghalliday (commented):
@afishbeck this looks like it is very close to being ready.  A few questions/comments.

helm/hpcc/values.schema.json
```diff
@@ -55,6 +55,31 @@
...
+        "storage": {
+          "$ref": "#/definitions/vaultCategory"
+        },
+        "esp": {
+          "$ref": "#/definitions/vaultCategory"
+        },
+        "ecl": {
+          "$ref": "#/definitions/vaultCategory"
+        },
+        "ecl-user": {
```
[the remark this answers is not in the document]
ghalliday: We should consider publishing the k8s secrets in subdirectories corresponding to the category - I think that would avoid the issue.

common/thorhelper/thorsoapcall.cpp
```diff
@@ -801,10 +805,13 @@ class CWSCHelper : implements IWSCHelper, public CInterface
     unsigned done;
     Owned<IPropertyTree> xpathHints;
     Linked<ClientCertificate> clientCert;
+    bool customClientCert = false;
```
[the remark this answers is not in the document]
afishbeck: Yes, if I need to make any other changes I will include this change as well.

helm/hpcc/values.schema.json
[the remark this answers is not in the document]
afishbeck: The categories allow the various types of secrets to be organized and secured differently. To prevent leaking and usage for unintended purposes.  I haven't made use of "ecl-user" yet, but that is going to be the least secure category, ecl code will have direct access to the contents of the secret.  For example, if I need a key for use with some plugin or to pass in some non standard way to a 3rd party.  Because the ecl can see the contents it could return it, write it, or send it anywhere.  I don't want gateway or storage plane credentials to be accessible in that way.

When it comes to the organization of the actual vault it could be a simple as:
http://${env.VAULT_SERVICE_HOST}:${env.VAULT_SERVICE_PORT}/v1/secret/data/storage/${secret}
http://${env.VAULT_SERVICE_HOST}:${env.VAULT_SERVICE_PORT}/v1/secret/data/ecl/${secret}
http://${env.VAULT_SERVICE_HOST}:${env.VAULT_SERVICE_PORT}/v1/secret/data/ecl-user/${secret}

But those different sections can have different access rights, and allow the secrets to be organized based on how and where they are used.

Over time we may want to segment the vault access by components.  So certain roxies can call some gateways but hthor can't for example.  But I'd rather get a feel for how these things are used in the wild a bit first.  Our users can give us some feedback on how they need to further restrict.

I definitely agree on the idea of the k8s secret subdirectories.  That might mean I no longer have to mangle the names of the HTTP-CONNECT secrets.  That was for example to prevent them being used in the way the ecl-user category will be in the future.

system/jlib/jsecrets.cpp
afishbeck: Based on a quick comment from @richardkchapman I think the line in the README where I do:

vault write auth/kubernetes/role/hpcc-vault-access \
        bound_service_account_names=hpcc-default \
        bound_service_account_namespaces=default \
        policies=hpcc-kv-ro \
        ttl=24h
may be enough to grant access.
It's more of an operational thing on the vault side granting access to our default account, or whatever the user may change the account to.  
I also support passing in the authentication token as a k8s secret.  Hopefully that will cover some other authentication schemes the user may set up.  The user can take whatever steps necessary to generate that.
Vault supports many authentication schemes and we can add more over time based on what scenarios users need, or we want to support in general.

afishbeck:
@ghalliday updated

1. Secret categories are based on usage, not just where they should go
2. Added new system category for HPCC system level secrets.
3. k8s secret path includes category.
4. Fixed vault configMap entries for THOR
5. README now demonstrates enabling all account names.
6. Minor improvements.

ghalliday:
@afishbeck I like the changes - please can you squash and I will scan it one more time before merging.

afishbeck:
@ghalliday squashed changes

PR #14133, opened 2020-09-03
HPCC-24689 Add an option to use dafilersv storage in cloud

jakesmith:
For testing purposes.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/thorlcr/master/thmastermain.cpp
+++ b/thorlcr/master/thmastermain.cpp
@@ -967,6 +967,34 @@ int main( int argc, const char *argv[]  )
                 throwStringExceptionV(0, "Failed to connect to all nodes");
             PROGLOG("verified mp connection to rest of cluster");
 
+#ifdef _CONTAINERIZED
+            if (globals->getPropBool("@_dafsStorage"))
+            {
+/* NB: This option is a developer option only.
+
+ * It is intended to be used to bring up a temporary Thor instance that uses local node storage,
+ * as the data plane.
+ * 
+ * It is likely to be deprecated or need reworking, when DFS is refactored to use SP's properly.
+ * 
+ * The mechanism works by:
+ * a) Creating a pseudo StoragePlane (publishes group to Dali).
+ * b) Spins up a dafilesrv thread in each slave container.
+ * c) Changes the default StoragePlane used to publish files, to point to the SP/group created in step (a).
+ * 
+ * In this way, a Thor instance, whilst up, will act similarly to a bare-metal system, using local disks as storage.
+ * This allows quick cloud based allocation/simulation of bare-metal type clusters for testing purposes.
+ * 
+ * NB: This isn't a real StoragePlane, and it will not be accessible by any other component.
+ *
+ */
+                StringBuffer uniqueGrpName;
+                queryNamedGroupStore().addUnique(&queryProcessGroup(), uniqueGrpName);
+                // change default plane
+                queryComponentConfig().setProp("storagePlane", uniqueGrpName);
+                PROGLOG("Persistent Thor group created with group name: %s", uniqueGrpName.str());
+            }
+#endif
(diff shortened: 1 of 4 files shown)


jakesmith:
@ghalliday - please review

thorlcr/master/thmastermain.cpp
```diff
@@ -967,6 +967,23 @@ int main( int argc, const char *argv[]  )
                 throwStringExceptionV(0, "Failed to connect to all nodes");
             PROGLOG("verified mp connection to rest of cluster");
 
+#ifdef _CONTAINERIZED
+            if (globals->getPropBool("@_dafsStorage"))
```
ghalliday, line 971: @jakesmith 
This will not work if there are multiple replicas for the same component.  Should probably document that, and make it clear this is a developer only debugging option.

thorlcr/master/thmastermain.cpp
```diff
@@ -967,6 +967,23 @@ int main( int argc, const char *argv[]  )
...
+                for (unsigned n=0; n<queryProcessGroup().ordinality(); n++)
+                {
+                    INode &node = queryProcessGroup().queryNode(n);
+                    StringBuffer host;
+                    node.endpoint().getIpText(host);
+                    hosts.push_back(host.str());
+                }
+                queryNamedGroupStore().add(thorName, hosts);
+                // change default plane
+                queryComponentConfig().setProp("storagePlane", thorName);
```
ghalliday, line 983: This should also be creating a storageplane entry - otherwise it will cause problems in the next stages of the disk refactoring.

thorlcr/master/thmastermain.cpp
jakesmith: yes, for that to work, when there are multiple instances of Thor, the SP name/group name, that each is using, needs to be unique.
I hadn't considered that scenario, but it's an easy change and makes it a bit more flexible, so I'll make that change.

thorlcr/master/thmastermain.cpp
jakesmith: As discussed, it can't create a real SP from a running instance dynamically.
This will either require re-thinking once DFS is refactored.

thorlcr/master/thmastermain.cpp
jakesmith: I will add some notes in the code, describing what, how and why this option is there.

jakesmith:
@ghalliday - please see 2nd commit

thorlcr/mfilemanager/thmfilemanager.cpp
```diff
@@ -518,9 +518,12 @@ class CFileManager : public CSimpleInterface, implements IThorFileManager
...
                 {
 #ifdef _CONTAINERIZED
-                    Owned<IStoragePlane> plane = getStoragePlane(groupNames.item(gn));
-                    assertex(plane);
-                    curDir.append(plane->queryPrefix());
+                    if (!globals->getPropBool("@_dafsStorage"))
+                    {
+                        Owned<IStoragePlane> plane = getStoragePlane(groupNames.item(gn));
+                        assertex(plane);
+                        curDir.append(plane->queryPrefix());
```
ghalliday, line 525: @jakesmith  Doesn't this mean curDir won't be set?

jakesmith: @ghalliday - yes if _dafsStorage is set, then curDir is blank, so it's same as if storage plane has no prefix.
It means that 'dir' will be left blank, and makePhysicalPartName will use the default (no diroverride).

jakesmith:
@ghalliday - squashed.

PR #14135, opened 2020-09-04
advice

jakesmith:
## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/common/thorhelper/thorcommon.cpp
+++ b/common/thorhelper/thorcommon.cpp
@@ -1377,10 +1377,10 @@ IExtRowStream *createRowStreamEx(IFile *file, IRowInterfaces *rowIf, offset_t of
         {
             // JCSMORE should pass in a flag for rw_compressblkcrc I think, doesn't look like it (or anywhere else)
             // checks the block crc's at the moment.
-            fileio.setown(createCompressedFileReader(file, eexp, UseMemoryMappedRead));
+            fileio.setown(createCompressedFileReader(file, eexp, UseMemoryMappedRead, UseMemoryMappedRead ? IFEnone : IFEsequential));
         }
         else
-            fileio.setown(file->open(IFOread));
+            fileio.setown(file->open(IFOread, IFEsequential));
         if (!fileio)
             return NULL;
         if (maxrows == (unsigned __int64)-1)
--- a/system/jlib/jfile.cpp
+++ b/system/jlib/jfile.cpp
@@ -1720,6 +1720,12 @@ public:
 
 };
 
+static bool useSeqAdvice = false; // for now
+void setUseSeqAdvice(bool tf)
+{
+    useSeqAdvice = tf;
+}
+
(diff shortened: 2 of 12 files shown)


jakesmith:
We could release this and experiment with it further, but from local experimentations on HDD and SDD is actually performs worse.
My local test is with a 4 slave Thor that is reading a 10GB file.
Between each run I flush the disk cache (echo 3 > /proc/sys/vm/drop_caches).
10 runs of each test.

The HDD tests are consistently ~15% *slower* with the POSIX_FADV_SEQUENTIAL advice.
The SDD test speed different with or without the flag is too small to be significant.

jakesmith:
@richardkchapman @mckellyln - any thoughts?

richardkchapman:
I did a quick web search to see if it shed any light on when use of sequential might make things slower, and found this:

https://stackoverflow.com/questions/3755765/what-posix-fadvise-args-for-sequential-file-write/3755850

which I thought was interesting though not directly related. In particular the bit about good write performance!

jakesmith:
I looked at whether we already perhaps using POSIX_FADV_NOREUSE (we're not), thinking it would be a sensible thing to use for large Thor writes.
However, the manual says:
```
In kernels before 2.6.18, POSIX_FADV_NOREUSE had the same semantics as POSIX_FADV_WILLNEED. This was probably a bug;
**since kernel 2.6.18, this flag is a no-op.**
```

sync_file_range sounds interesting. I wonder though if the reason POSIX_FADV_SEQUENTIAL is worse, is because the OS does a better job of anticipating readahead with overlapping requests that the test involved.
i.e. might be best to let the smarts be done by the kernel.

It would be good to avoid filling the cache unnecessarily though, and/or prioritizing what stays in the disk cache.
O_DIRECT looks to be one way potentially.
Or we could use POSIX_FADV_DONTNEED more aggressively (e.g. after every block read).

mckellyln:
I think we do use sync_file_range() now, after writes() if the IFEnocache is set.
I've mentioned O_DIRECT before, sometimes with that comes having to use block boundaries and sometimes it might mean rolling our own async read-ahead where it helps.

mckellyln:
io_uring is a new method to look into.  Not sure how it works with NAS though.

mckellyln:
On a single / spindle disk extra read-ahead might hurt; SSD or multi-disk raid might show a small gain.
Read-ahead algo can be tricky and should take into consideration hardware etc.
I doubt the POSIX_FADV_SEQUENTIAL code does that.

mckellyln:
@jakesmith PR looks good - just not sure if it should be enabled.
Perhaps also we should by default set IFEnocache flag on Thor reads ?

jakesmith:
Closing for now.

PR #14137, opened 2020-09-04
HPCC-24666 Improve filterLogContent in ESDL logging

wangkx:
1. In the existing filterLogContent(), an extra tree layer is
added to the log request even if the log request is already
in a tree structure. That is fixed now. The related code is
moved in the front of the filterLogContent().

2. For the log request not in one tree structure (the log
request with ESPContext, UserContext, etc), only change in
this fix is to move the ScriptValues from the LogContent
branch to the UpdateLogRequest branch.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/logging/logginglib/loggingagentbase.cpp
+++ b/esp/logging/logginglib/loggingagentbase.cpp
@@ -154,174 +154,180 @@ void CLogContentFilter::filterLogContentTree(StringArray& filters, IPropertyTree
 
 IEspUpdateLogRequestWrap* CLogContentFilter::filterLogContent(IEspUpdateLogRequestWrap* req)
 {
-    Owned<IPropertyTree> scriptValues = req->getScriptValuesTree();
-    const char* logContent = req->getUpdateLogRequest();
     Owned<IPropertyTree> logRequestTree = req->getLogRequestTree();
-    Owned<IPropertyTree> updateLogRequestTree = createPTree("UpdateLogRequest");
+    const char* logContent = req->getUpdateLogRequest();
+    if (logRequestTree || !isEmptyString(logContent))
+    {
+        if (groupFilters.length() < 1)
+        {//No filter
+            if (!logRequestTree)
+                return LINK(req);
 
-    StringBuffer source;
-    if (logBackEndReq && logBackEndResp && groupFilters.length() < 1)
-    {//No filter
+            //The first version of the filterLogContent() always returns the UpdateLogRequest string.
+            StringBuffer updateLogRequestXML;
+            toXML(logRequestTree, updateLogRequestXML);
+            req->setUpdateLogRequest(updateLogRequestXML);
+            return LINK(req);
+        }
+
+        if (!logRequestTree)
+            logRequestTree.setown(createPTreeFromXMLString(logContent));
+
+        Owned<IPropertyTree> filteredLogRequestTree = createPTree(logRequestTree->queryName());
+        bool logContentEmpty = true;
+        filterLogContentTree(groupFilters.item(0).getFilters(), logRequestTree, filteredLogRequestTree, logContentEmpty);
+        if (logContentEmpty)
+            throw MakeStringException(EspLoggingErrors::UpdateLogFailed, "The filtered content is empty.");
+
+        StringBuffer updateLogRequestXML;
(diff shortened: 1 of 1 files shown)


wangkx:
@timothyklemm please review

esp/logging/logginglib/loggingagentbase.cpp
```diff
@@ -154,174 +154,172 @@ void CLogContentFilter::filterLogContentTree(StringArray& filters, IPropertyTree
...
+        if (logContentEmpty)
+            throw MakeStringException(EspLoggingErrors::UpdateLogFailed, "The filtered content is empty.");
 
+        if (getEspLogLevel()>=LogMax)
+        {
+            StringBuffer updateLogRequestXML;
+            toXML(filteredLogRequestTree, updateLogRequestXML);
+            ESPLOG(LogMax, "filtered content and option: <%s>", updateLogRequestXML.str());
+        }
+        req->setLogRequestTree(filteredLogRequestTree.getClear());
```
timothyklemm, line 180: This could lead to confusion. If the input contained non-empty logContent, that text remains in the input and the input now contains a tree that should be assumed to not match (we expect the filters to filter). Either the two values should be equivalent or only one value should be be present on exit.

Previously, the value returned was always the string in a new instance. If changed to use a tree (when a tree was given on input), must ensure that the consumers of this data are prepared for a tree and not assuming the string will be populated.

esp/logging/logginglib/loggingagentbase.cpp
```diff
@@ -154,174 +154,172 @@ void CLogContentFilter::filterLogContentTree(StringArray& filters, IPropertyTree
...
+    {
+        if (noFilter)
+            return LINK(req);
+
+        if (!logRequestTree)
+            logRequestTree.setown(createPTreeFromXMLString(logContent));
+
+        Owned<IPropertyTree> filteredLogRequestTree = createPTree(logRequestTree->queryName());
+        bool logContentEmpty = true;
+        filterLogContentTree(groupFilters.item(0).getFilters(), logRequestTree, filteredLogRequestTree, logContentEmpty);
```
timothyklemm, line 170: We know that at least one filter should be applied, but we don't know that it is defined in groupFilters - which may be empty. Need to properly handle groupFilters AND logBackEndReq AND logBackEndResp.

esp/logging/logginglib/loggingagentbase.cpp
```diff
@@ -154,174 +154,172 @@ void CLogContentFilter::filterLogContentTree(StringArray& filters, IPropertyTree
...
-    Owned<IPropertyTree> scriptValues = req->getScriptValuesTree();
-    const char* logContent = req->getUpdateLogRequest();
+    bool noFilter = logBackEndReq && logBackEndResp && (groupFilters.length() < 1);
     Owned<IPropertyTree> logRequestTree = req->getLogRequestTree();
-    Owned<IPropertyTree> updateLogRequestTree = createPTree("UpdateLogRequest");
+    const char* logContent = req->getUpdateLogRequest();
+    if (logRequestTree || !isEmptyString(logContent))
+    {
+        if (noFilter)
+            return LINK(req);
```
timothyklemm, line 163: This has the potential to break any deployment that has already discovered the extra element level and adapted to it. The change is probably worthwhile, but may warrant a red book mention in case anyone's logging stops working.

timothyklemm (changes requested):

esp/logging/logginglib/loggingagentbase.cpp
timothyklemm: On second thought, it looks like the change that introduced this behavior has only been included in nightly builds. If this is true, this comment can be disregarded.

timothyklemm (approved):
There remains an issue with applying filters multiple times to a request. This occurs when the agent defines the only filters. The manager now assembles log components into a single document, and agent filters defined to apply to components will now be applied to a single document.

This behavior is not caused by this PR and https://track.hpccsystems.com/browse/HPCC-24710 has been opened to address it.

wangkx:

PR #14150, opened 2020-09-08
HPCC-24632 Use archive before shared mysql client lib, if found

mckellyln:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/cmake_modules/FindMYSQL.cmake
+++ b/cmake_modules/FindMYSQL.cmake
@@ -26,8 +26,8 @@ IF (NOT MYSQL_FOUND)
   IF (WIN32)
     SET (mysql_lib "libmysql")
   ELSE()
-    SET (mysql_lib "mysqlclient_r")    # Use the re-entrant version if it exists
-    SET (mysql_lib_alt "mysqlclient")  # (but newer versions do not make the distinction and supply just the one)
+    SET (mysql_lib "libmysqlclient_r.a" "mysqlclient_r")  # Use the re-entrant version if it exists
+    SET (mysql_lib_alt "libmysqlclient.a" "mysqlclient")  # (but newer versions do not make the distinction and supply just the one)
   ENDIF()
   IF (NOT "${EXTERNALS_DIRECTORY}" STREQUAL "")
     IF (UNIX)


mckellyln:
Note: see https://github.com/xwang2713/cloud-image-build/pull/5
This PR should be merged before that one.

mckellyln:
Some distros would link in the archive version of libmysqlclient.a and some linked in the shared lib.
In order to use a newer version of this lib and not have to include the package - as it may not be available from default reposiitories at HPCC installation time - we specify to look for the archive verison of the lib before the shared version.
Should we only look for the archive version ?

mckellyln:
The reason for wanting a newer client lib version is to be able to use some newer options that are in the embed library, such as MYSQL_OPT_MAX_ALLOWED_PACKET.
@xwang2713 @richardkchapman for review/comment.

xwang2713:
I will update cloud-image-build repo not only  CentOS 7 but also  Amazon 2 which is similar to CentOS 7 when this PR is merged. CentOS 8 should be OK since it has mysql-devel package which is MySQL 8.0

mckellyln:
Thanks.
@richardkchapman your thoughts ?

richardkchapman:
Is this change appropriate for a point release?

mckellyln:
Good point, but then it would mean only in 7.12 and not available for current users.
But MySQL 8 is released, and this previous 5.7.30 is not the latest.
I will re-target.

mckellyln:
We can delay the other repo (build) PR so that the mysql client version does not change until that is merged.
This PR ensures when/if the mysql client lib version changes (and no longer has a _r archive) the archive version is still used (if present) so that a newer mysql client package is not an install dependency.
But I do feel like the build repo should me more closely tied to platform releases, if that is possible.

mckellyln:
I suppose we should not change these things in a point release, so I will also re-target.

mckellyln:
rebased to master

mckellyln:
Since the build PR (https://github.com/xwang2713/cloud-image-build/pull/5) is not hpcc version based I'm not sure any of this matters as we cannot use one version of the mysql client for 7.6 and a different one for 7.12 on the same el7 build platform.

PR #14153, opened 2020-09-09
HPCC-24567 JWT-based security manager plugin

dcamper:
Feature, file scope, and workunit scope permissions are supported in legacy configuration.  Custom external service (to provide the JWT token) is required.  Container support will be coming in a later update.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

## Testing:

Manual tests only.


dcamper:
@richardkchapman Please review or assign for review.  Thanks!

richardkchapman:
@RussWhitehead @afishbeck Please review

afishbeck:
@timothyklemm can you take a look?  You along with Russ are most familiar with implementing security plugins.  I'll take a closer look after.

RussWhitehead (changes requested):
Some comments. Also did you get a chance to memleak test it (Valgrind)?
I focused on the platform interface, since I don't know about the JWT and CURL stuff. Hopefully Tony can help with that

system/security/shared/secloader.hpp
```diff
@@ -69,20 +71,30 @@ class SecLoader
...
             throw MakeStringException(-1, "%s Security Manager %s failed to instantiate in call to %s", lsm, libName.str(), instFactory.str());
         return pPSM;
     }
 
+    static ISecManager* loadPluggableSecManager(const char * bindingName, IPropertyTree* bindingCfg, IPropertyTree* secMgrCfg)
+    {
+        return _loadPluggableSecManager<ISecManager>(bindingName, bindingCfg, secMgrCfg);
+    }
+
+    static IDaliLdapConnection* loadLDAPPluggableSecManager(const char * bindingName, IPropertyTree* bindingCfg, IPropertyTree* secMgrCfg)
```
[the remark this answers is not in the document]
dcamper: The name I chose was supposed to reflect the caller's expectations, which is a pointer to a "Dali LDAP connection" instance.  It is called only from within Dali.  What is happening here is that a generic security manager plugin is loaded but then casts to IDaliLdapConnection*.  That works, provided the plugin implements IDaliLdapConnection (which the JWT plugin does).  The code would abort if something like htpasswd was requested.

dali/server/daldap.cpp
```diff
@@ -272,6 +273,22 @@ class CDaliLdapConnection: implements IDaliLdapConnection, public CInterface
 };
 
 
+IDaliLdapConnection *createDaliSecMgrPluginConnection(IPropertyTree *propTree)
+{
+    if (propTree && propTree->hasProp("@type"))
+    {
+        Owned<IPropertyTree>    secMgrCfg;
+
```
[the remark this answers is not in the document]
dcamper: Done.

dali/server/daserver.cpp
```diff
@@ -139,6 +139,58 @@ void usage(void)
...
+            {
+                const char* authPluginType = proptree->queryProp("@authPluginType");
+
+                if (authPluginType)
+                {
+                    VStringBuffer xpath("SecurityManagers/SecurityManager[@name='%s']", authPluginType);
+
+                    foundTree.set(configTree->getPropTree(xpath));
+                }
+            }
```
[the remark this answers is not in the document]
dcamper: Missing tree nodes could arise when upgrading a cluster without touching the legacy configuration, but @authMethod is has been defined for a very long time.  @authPluginType is new, but if it is missing that means that no attempt has been made to add a JWT plugin; it is not an error.  That said, it _is_ worthwhile to log a missing security manager tree node a few lines down.  I've add that.

dali/server/daserver.cpp
```diff
@@ -139,6 +139,58 @@ void usage(void)
...
+#ifdef _CONTAINERIZED
+    IPropertyTree *jwtConfigPTree = configTree->queryPropTree("jwt");
+
+    if (jwtConfigPTree)
+    {
+        foundTree.set(createPTreeFromXMLString("<SecurityManagers/>"));
+
+        StringAttr configName(jwtConfigPTree->queryProp("@objname"));
+
+        VStringBuffer secMgrXML("<SecurityManager name='%s' instanceFactoryName='createInstance' libName='libjwtSecurity.so' type='JwtSecurityManager'/>", configName.str());
```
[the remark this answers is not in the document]
dcamper: This is admittedly terrible, and I blame Tony.  This code is inspired by the changes to esp/platform/application_config.cpp where, as a way to bridge the container-aware and legacy configurations, we basically replicate the the legacy XML configuration structure when running in a container.  The alternative to coding names like this would be to include them in YAML.

@afishbeck Would you recommend moving the library name here (and the names in application_config.cpp) into the YAML configurations or leave them coded?

esp/applications/eclqueries/jwt.yaml
```diff
@@ -0,0 +1,13 @@
+jwt:
+   objname: jwtsecmgr
+
+   ## Default properties
+   allowSelfSignedCert: false
+   keyFilePath: /opt/HPCCSystems/secrets/jwt-security/key
```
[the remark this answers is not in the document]
dcamper: It is a default value that does make sense, but it would perhaps make more sense to support the newer secrets vault as an option.

But note, this PR does not fully support containers just yet.  I'm target legacy configurations at the moment and hope to address container support Real Soon Now.

esp/platform/application_config.cpp
```diff
@@ -78,14 +78,8 @@ static void copyDirectories(IPropertyTree *target, IPropertyTree *src)
     }
 }
 
-//returns true if ldap is enabled.
-//  only support ldap or "none" for now.
-//  "none" must be explicit, don't want to accidentally turn off security
-bool addSecurity(IPropertyTree *legacyEsp, IPropertyTree *appEsp, StringBuffer &bindAuth)
```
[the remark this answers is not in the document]
dcamper: The comments were moved to lines 144-145.  Though, to be honest, I don't think the code matches the comments.  I've added an explicit check for "none".

esp/platform/application_config.cpp
```diff
@@ -111,7 +105,60 @@ bool addSecurity(IPropertyTree *legacyEsp, IPropertyTree *appEsp, StringBuffer &
...
+    VStringBuffer secMgrXML("<SecurityManager name='%s' instanceFactoryName='createInstance' libName='libjwtSecurity.so' type='JwtSecurityManager'/>", configName.str());
+    IPropertyTree *secMgrTree = legacySecMgrs->addPropTree("SecurityManager", createPTreeFromXMLString(secMgrXML));
+    VStringBuffer jwtSecMgrXML("<JwtSecurityManager build='_' buildSet='%s' instanceFactoryName='createInstance'/>", configName.str());
+    IPropertyTree *jwtSecMgrTree = secMgrTree->addPropTree("JwtSecurityManager", createPTreeFromXMLString(jwtSecMgrXML));
+
+    copyAttributes(jwtSecMgrTree, appJWT);
+    jwtSecMgrTree->setProp("@loginEndpoint", loginEndpoint);
+    jwtSecMgrTree->setProp("@refreshEndpoint", refreshEndpoint);
+    jwtSecMgrTree->setProp("@clientID", clientID);
+    jwtSecMgrTree->setProp("@libName", "libjwtSecurity.so");
```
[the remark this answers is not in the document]
dcamper: That line should actually have been merged with the jwtSecMgrXML string literal, which I've now done.  However, I will wait for Tony's feedback on perhaps moving these values into YAML before (potentially) turning them into string constants.

esp/platform/application_config.cpp
```diff
@@ -111,7 +105,60 @@ bool addSecurity(IPropertyTree *legacyEsp, IPropertyTree *appEsp, StringBuffer &
...
+
+// returns true if an authorization-aware security manager is enabled.
+//  "none" must be explicit, don't want to accidentally turn off security
+bool addSecurity(IPropertyTree *legacyEsp, IPropertyTree *appEsp, StringBuffer &bindAuth)
+{
+    const char *auth = appEsp->queryProp("@auth");
+
+    if (auth)
+    {
+        if (streq(auth, "ldap"))
```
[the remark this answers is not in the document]
dcamper: Good question, and I'm not sure if YAML is case sensitive or not.  I've went ahead and changed them to strieq() though.

esp/platform/application_config.cpp
```diff
@@ -129,6 +176,37 @@ void bindAuthResources(IPropertyTree *legacyAuthenticate, IPropertyTree *app, co
         legacyAuthenticate->addPropTree("Feature", LINK(&features->query()));
 }
 
+void bindJWTAuthResources(IPropertyTree *legacyAuthenticate, IPropertyTree *app, const char *service)
+{
+    IPropertyTree *appJWT = app->queryPropTree("jwt");
+    if (!appJWT)
+        throw MakeStringException(-1, "can't find application JWT settings.  To run without security set auth=none");
```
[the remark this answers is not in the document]
dcamper: Copied and pasted from elsewhere in that code unit.  I've changed all occurrences to "Cannot".

initfiles/componentfiles/configschema/xsd/dali.xsd
```diff
@@ -104,9 +104,15 @@
                             <xs:restriction base="xs:string">
                                 <xs:enumeration value="kerberos" hpcc:description=""/>
                                 <xs:enumeration value="simple" hpcc:description=""/>
+                                <xs:enumeration value="secmgrPlugin" hpcc:description=""/>
                             </xs:restriction>
                         </xs:simpleType>
                     </xs:attribute>
+                    <xs:attribute name="authPluginType" type="xs:string"
+                                  hpcc:displayName="Security Manager Plugin Name"
+                                  hpcc:visibleIf="/Environment/Software/LDAPServerProcess"
```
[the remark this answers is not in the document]
dcamper: That is actually correct.  I did not change the panels visible in configmgr, so labels such as that one are still in force.

initfiles/componentfiles/configschema/xsd/dali.xsd
```diff
@@ -104,9 +104,15 @@
...
+                                  hpcc:requiredIf=".[@authMethod=('secmgrPlugin')]"
+                                  hpcc:tooltip="Security manager plugin name (when authentication method is secmgrPlugin)"/>
```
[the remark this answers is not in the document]
dcamper: Fixed in this PR, but there are a number of other instances in other code as well.

system/security/plugins/jwtSecurity/jwtCache.hpp
```diff
@@ -0,0 +1,459 @@
...
+{
+    private:
+
+        typedef std::map<std::string, SecAccessFlags> PermissionMap;
+
+    public:
+
+        JWTUserInfo();
+
+        ~JWTUserInfo();
```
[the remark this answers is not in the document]
dcamper: This class has no parent or child classes; virtual not necessary.

system/security/plugins/jwtSecurity/jwtCache.hpp
```diff
@@ -0,0 +1,459 @@
...
+/**
+ * Instances of JWTUserInfo represent all of the cached values for a particular
+ * user.  The intention is that instances would be accessed via another
+ * container, such as a map keyed off the user's username.  See JWTUserCache class.
+ */
+class JWTUserInfo
+{
+    private:
+
+        typedef std::map<std::string, SecAccessFlags> PermissionMap;
```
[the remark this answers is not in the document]
dcamper: Destructor will clean it up.

system/security/plugins/jwtSecurity/jwtCache.hpp
```diff
@@ -0,0 +1,459 @@
...
+};
+
+/**
+ * This is basically a thread-safe map of username->JWTUserInfo objects.
+ */
+class JWTUserCache
+{
+    private:
+
+        typedef std::map<std::string, JWTUserInfo> UserPermissionMap;
```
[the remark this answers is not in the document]
dcamper: When the plugin is unloaded and the destructor called.

dali/server/daserver.cpp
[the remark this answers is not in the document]
afishbeck: As part of another feature I'm adding a generic way of adding SecurityManager plugins.  If you want to wait for that for containerized mode that would be fine.  The new stuff does have the library as part of the yaml.

system/security/plugins/jwtSecurity/jwtCache.hpp
```diff
@@ -0,0 +1,459 @@
...
+#include <string>
+
+#include "basesecurity.hpp"
+
+/**
+ * Instances of JWTUserInfo represent all of the cached values for a particular
+ * user.  The intention is that instances would be accessed via another
+ * container, such as a map keyed off the user's username.  See JWTUserCache class.
+ */
+class JWTUserInfo
```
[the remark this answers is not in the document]
dcamper: Does that apply when a copy of the retrieved value is returned?  I'm using copies, rather than references, to avoid issues with multi-threaded management.

system/security/plugins/jwtSecurity/jwtCache.cpp
```diff
@@ -0,0 +1,396 @@
...
+{
+}
+
+JWTUserCache::~JWTUserCache()
+{
+}
+
+bool JWTUserCache::has(const std::string& userName) const
+{
+    return userPermMap.find(userName) != userPermMap.end();
```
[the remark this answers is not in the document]
dcamper: Good catch.  Fixed here and in get() as well.

system/security/plugins/jwtSecurity/jwtCache.cpp
```diff
@@ -0,0 +1,396 @@
...
+
+        userPermMap[userName] = userInfo;
+    }
+
+    return *this;
+}
+
+bool JWTUserCache::get(const std::string& userName, JWTUserInfo& userInfo)
+{
+    UserPermissionMap::iterator     foundIter = userPermMap.find(userName);
```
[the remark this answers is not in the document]
dcamper: Fixed.

system/security/plugins/jwtSecurity/jwtEndpoint.cpp
```diff
@@ -0,0 +1,257 @@
...
+}
+
+void FreeMemoryStruct(MemoryStruct& mem)
+{
+    if (mem.memory)
+    {
+        free(mem.memory);
+        mem.memory = nullptr;
+    }
+    mem.size = 0;
```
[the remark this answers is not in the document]
dcamper: They should be separate.  The if() block removes the pointer to allocated memory only if it exists, but the result of the entire function call should be a "zeroed" memory struct.

system/security/plugins/jwtSecurity/jwtEndpoint.cpp
```diff
@@ -0,0 +1,257 @@
...
+
+size_t CaptureIncomingCURLReply(void* contents, size_t size, size_t nmemb, void* userp)
+{
+    size_t          incomingDataSize = size * nmemb;
+    MemoryStruct*   mem = static_cast<struct MemoryStruct*>(userp);
+    size_t          MAX_BUFFER_SIZE = 4194304; // 2^22
+
+    if (mem->size + incomingDataSize < MAX_BUFFER_SIZE)
+    {
+        if (mem->size + incomingDataSize > mem->capacity)
```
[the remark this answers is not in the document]
dcamper: Done.

system/security/plugins/jwtSecurity/configxml/jwtsecmgr.xsd
```diff
@@ -0,0 +1,131 @@
...
+          <xs:appinfo>
+            <tooltip>URL or unique name of this HPCC Systems cluster</tooltip>
+          </xs:appinfo>
+        </xs:annotation>
+      </xs:attribute>
+
+      <xs:attribute name="wuDefaultAccess" use="required" default="Full">
+        <xs:annotation>
+          <xs:appinfo>
+            <tooltip>Default access level for workunit scopes</tooltip>
```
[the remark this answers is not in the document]
dcamper: Good catch; done.

system/security/plugins/jwtSecurity/jwtSecurity.cpp
```diff
@@ -0,0 +1,1112 @@
...
+            else if (included_algo == "ES512")
+                jwtVerifier = jwtVerifier.allow_algorithm(jwt::algorithm::es512(keyContents, "", "", ""));
+            else if (included_algo == "PS256")
+                jwtVerifier = jwtVerifier.allow_algorithm(jwt::algorithm::ps256(keyContents, "", "", ""));
+            else if (included_algo == "PS384")
+                jwtVerifier = jwtVerifier.allow_algorithm(jwt::algorithm::ps384(keyContents, "", "", ""));
+            else if (included_algo == "PS512")
+                jwtVerifier = jwtVerifier.allow_algorithm(jwt::algorithm::ps512(keyContents, "", "", ""));
+            else
+                PROGLOG("CJwtSecurityManager error: Unknown token algorithm: %s", decodedToken.get_algorithm().c_str());
```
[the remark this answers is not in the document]
dcamper: Excellent catch!  If that had stayed in place, not algorithm check would have been added to verifier and so outrageously bad tokens will have succeeded.  I've converted that to a string exception.

system/security/plugins/jwtSecurity/jwtSecurity.cpp
```diff
@@ -0,0 +1,1112 @@
...
+                case RT_DEFAULT:
+                    {
+                        accessFlag = userInfo.getFeaturePerm(resourceName, SecAccess_Full);
+                    }
+                    break;
+
+                case RT_FILE_SCOPE:
+                    {
+                        // Scope hpccinternal::<username> always has full access
+                        if (resourceName && strlen(resourceName) > 14 && strncmp(resourceName, "hpccinternal::", 14) == 0 && stricmp(&resourceName[14], user.getName()) == 0)
```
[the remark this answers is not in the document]
dcamper: It is specifically 'hpccinternal::<username>' and it actually does always have full access.  The scope is used for spill and some temp files, I believe.  Including this here prevents us from having to define it for every user.

HOWEVER:  Reviewing this logic, I see that it is incomplete.  Users trying to access someone else's hpccinternal scope should be explicitly denied.  I've made that change.

dali/server/daserver.cpp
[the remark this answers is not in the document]
dcamper: Excellent, thank you.  I'll leave this as-is for now, pending your changes.

dcamper:
@RussWhitehead I've finished the round of modifications based on your feedback.  Can you check it again when you get a chance.  Thanks!

system/security/plugins/jwtSecurity/jwtEndpoint.cpp
```diff
@@ -0,0 +1,257 @@
...
+            headers = nullptr;
+        }
+        curl_easy_cleanup(curlHandle);
+        FreeMemoryStruct(captureBuffer);
+    }
+
+    return apiResponse;
+}
+
+std::string tokenFromRefresh(const std::string& jwtEndPoint, bool allowSelfSignedCert, const std::string& clientID, const std::string& refreshToken)
```
timothyklemm, line 174: With the exception of the json parameters, and a couple logging statements, this is identical to tokenFromLogin. Could they be refactored to share code instead of replicating?

system/security/plugins/jwtSecurity/jwtCache.hpp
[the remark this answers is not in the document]
timothyklemm: The "get" prefix usually suggests that the caller is responsible for releasing the returned value, while "query" usually means you are just looking at somebody else's reference and are not expected to release it. A copy that the caller is responsible for disposing should be "get".

system/security/plugins/jwtSecurity/jwtCache.cpp
```diff
@@ -0,0 +1,410 @@
...
+
+        userPermMap.clear();
+    }
+
+    return *this;
+}
+
+unsigned int JWTUserCache::count() const
+{
+    return userPermMap.size();
```
timothyklemm, line 409: Should this map read also be guarded?

system/security/plugins/jwtSecurity/jwtCache.cpp
```diff
@@ -0,0 +1,410 @@
...
+        CriticalBlock block(crit);
+
+        foundIter = userPermMap.find(userName);
+    }
+
+    if (foundIter != userPermMap.end())
+    {
+        userInfo = foundIter->second;
+        return true;
+    }
```
timothyklemm, line 380: This looks to be unsafe. Once the guard around find is released, the contents of what was found can be changing during the copy. For example, before foundIter->second is copied foundIter could be erased.

dali/server/daserver.cpp
```diff
@@ -139,6 +139,58 @@ void usage(void)
...
+                    VStringBuffer xpath("SecurityManagers/SecurityManager[@name='%s']", authPluginType);
+
+                    foundTree.set(configTree->getPropTree(xpath));
+                }
+            }
+        }
+    }
+#endif
+
+    return foundTree;
```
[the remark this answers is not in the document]
dcamper: Good catch.  Support for the containerized version is not included in this PR -- there is some support, but incomplete -- but what is there should match.  I will make sure the node is "SecurityManager" for both.

system/security/plugins/jwtSecurity/README.md
```diff
@@ -0,0 +1,164 @@
+## JWT Authorization Security Manager Plugin
+
+The purpose of this plugin is to provide authentication and authorization capabilities for HPCC Systems users, with the
+credentials passed via valid JWT tokens.
+
+The intention is to adhere as closely as possibly to the OpenID Connect (OIC) specification, which is a simple identity
+layer on top of the OAuth 2.0 protocol, while maintaining compatibility with the way HPCC Systems performs
+authentication and authorization today.  More information about the OpenID Connect specification can be found found at
```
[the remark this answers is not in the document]
dcamper: Fixed.

system/security/plugins/jwtSecurity/README.md
```diff
@@ -0,0 +1,164 @@
...
+
+If the session token is not present, the plugin will call a ```JWT login service``` (also known as a JWT login endpoint)
+with the username and password, plus a nonce value for additional security.
+
+That service authenticates the username/password credentials.  If everything is good, the service constructs an
+OIC-compatible token that includes authorization information for that user and returns it to the plugin.  The
+token is validated according to the OIC specification, including signature verification.
+
+Note that token signature verification requires an additional piece of information.  Tokens can be signed with a
+hash-based algorithm or with a pubic key-based algorithm (the actual algorithm used is determined by the JWT service).
```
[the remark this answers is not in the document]
dcamper: I _could_ be a new class of crypto here....

Fixed.

RussWhitehead:
@dcamper Please squash when @afishbeck @timothyklemm are satisfied. I would like to review it again, this time from a more functional perspective

system/security/plugins/jwtSecurity/jwtSecurity.hpp
```diff
@@ -0,0 +1,38 @@
...
+    #define JWTSECURITY_API DECL_EXPORT
+#endif //JWTSECURITY_EXPORTS
+
+#endif
+
+extern "C"
+{
+    JWTSECURITY_API ISecManager* createInstance(const char* serviceName, IPropertyTree& secMgrCfg, IPropertyTree& bndCfg);
+
+    JWTSECURITY_API IDaliLdapConnection* createLDAPJWTInstance(const char* serviceName, IPropertyTree& secMgrCfg, IPropertyTree& bndCfg);
```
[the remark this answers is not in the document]
dcamper: It's the casting that blows this up.  Dali is expecting a pointer to an IDaliLdapConnection instance.  I could have removed this function and forced a cast, but using a separate function (along with the templated function that is called) enforces the proper class hierarchy in the secmgr plugin:  JWT implements IDaliLdapConnection while others do not.  If someone tries to make one of the other plugins work, it will fail at runtime.

system/security/plugins/jwtSecurity/jwtSecurity.cpp
```diff
@@ -0,0 +1,1121 @@
...
+            }
+
+            gUserCache.set(subject, userInfo);
+
+            verificationResult = true;
+        }
+        catch (const std::runtime_error& e)
+        {
+            // Most jwt-cpp exceptions are subclasses of std::runtime_error
+            PROGLOG("CJwtSecurityManager::verifyToken error: %s", e.what());
```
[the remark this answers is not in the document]
dcamper: I don't know if it would be the best, as I have not kept up on the logging changes.  Would DBGLOG be more appropriate for these?  Or a different function?

system/security/plugins/jwtSecurity/jwtSecurity.hpp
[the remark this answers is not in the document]
dcamper: I have removed this function.  It was a holdover from when I was trying to directly instantiate the object within Dali rather than go through the loader process.

system/security/plugins/jwtSecurity/jwtSecurity.cpp
```diff
@@ -0,0 +1,1121 @@
...
+    }
+
+    /**
+     * @return  New string composed of random characters.  Size of string
+     * governed by the NONCE_LEN value defined within the function.
+     */
+    static std::string generateNonce()
+    {
+        const char      nonceLetters[] = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
+        size_t          NONCE_LEN = 24;
```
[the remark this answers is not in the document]
dcamper: Done.

system/security/plugins/jwtSecurity/jwtSecurity.cpp
```diff
@@ -0,0 +1,1121 @@
...
+        }
+
+        return true;
+    }
+
+    virtual SecAccessFlags authorizeEx(SecResourceType rtype, ISecUser& user, const char* resourceName, IEspSecureContext* secureContext) override
+    {
+        SecAccessFlags  resultFlag = SecAccess_Unavailable;
+
+        try
```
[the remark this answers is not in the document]
dcamper: The oversight here is that the other form of authorizeEx() needs the try...catch.  Fixed.

system/security/plugins/jwtSecurity/jwtCache.cpp
dcamper: std::map::size() is retrieved from a member variable.  While it is possible to get an incorrect value due to threaded access, a mutex does not add a guarantee that the value would be correct anyway.

system/security/plugins/jwtSecurity/jwtCache.cpp
dcamper: Good catch.  Fixed.

system/security/shared/secloader.hpp
```diff
@@ -18,6 +18,7 @@
 #ifndef _SECLOADER_HPP__
 #define _SECLOADER_HPP__
 #include "seclib.hpp"
+#include "../../../dali/server/daldap.hpp"
```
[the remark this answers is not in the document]
dcamper: Good point.  Done.

dcamper:
@timothyklemm I have pushed changes for your review.  Please take a look.  Thanks!

dcamper:
Squashed and rebased.  Please re-review.  Thanks!

esp/applications/eclqueries/jwt.yaml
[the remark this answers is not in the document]
afishbeck: The secrets code has been merged.
And you can use the secrets access code on baremetal as well.  The file just needs to be put in the right place.

Secrets are categorized you can use the system category for now.  
Either always use "jwt-security" as the secret name or configure it something like:
  secret: jwt-security


Then in bare metal put your key at /opt/HPCCSystems/secrets/system/jwt-security/key.

For both baremetal and k8s, in your code access the contents using:

If key is stored as a string:
StringBuffer s;
getSecretValue(s, "system", "jwt-security", "key", true);

esp/applications/eclservices/jwt_authorization_map.yaml
```diff
@@ -0,0 +1,153 @@
...
+   resource_map:
+      WsSMC:
+         Feature:
+         -  path: SmcAccess
+            resource: SmcAccess
+            description: Base access to ESP services
+         -  path: ThorQueueAccess
+            resource: ThorQueueAccess
+            description: Access to thor queues
+         -  path: ConfigAccess"
```
afishbeck, line 15: Extra quote.

system/security/plugins/jwtSecurity/README.md
```diff
@@ -0,0 +1,164 @@
...
+### Theory of Operations
+
+The plugin is called by the HPCC Systems ```esp``` process when a user needs to be authenticated.  That call will
+contain the user's username and either a reference to a session token or a password.  The session token is present only
+for already-authenticated users.
+
+If the session token is not present, the plugin will call a ```JWT login service``` (also known as a JWT login endpoint)
+with the username and password, plus a nonce value for additional security.
+
+That service authenticates the username/password credentials.  If everything is good, the service constructs an
```
afishbeck, line 43: Flavio may prefer it if we sent a hash of the password rather than the value.

system/security/plugins/jwtSecurity/README.md
```diff
@@ -0,0 +1,164 @@
...
+modify a user's authorization while the user is logged in.  This plugin fully supports refresh tokens by validating
+token lifetime at every authorization check and calling a ```JWT refresh service``` (also known as a JWT refresh
+endpoint) when needed.  This largely follows the OIC specification.
+
+### Deviations From OIC Specification
+
+* Initial authentication:  As stated above, the ```esp``` process will gather the username/password credentials instead
+of a third party, then send those credentials off to another service.  In a true OIC configuration, the client process
+(the ```esp``` process) never sees user credentials and relies on an external service to gather them from the user.
+* The request made to the ```JWT login service``` is a POST HTTP call containing four items in JSON format; example:
```
afishbeck, line 71: HTTPS?

system/security/plugins/jwtSecurity/README.md
```diff
@@ -0,0 +1,164 @@
...
+allows you to use third-party JWT libraries to construct and validate those tokens.
+
+### HPCC Systems Configuration Notes
+
+Several items must be defined in the platform's configuration.  Within configmgr, the ```jwtsecmgr``` Security Manager
+plugin must be added as a component and then modified according to your environment:
+
+* The URL or unique name of this HPCC Systems cluster, used as the ```client_id``` in token requests
+* Full URL to the ```JWT Login Endpoint``` (should be https, but not required)
+* Full URL to the ```JWT Refresh Endpoint``` (should be https, but not required)
```
afishbeck, line 106: Should we require HTTPS for these?

system/security/plugins/jwtSecurity/README.md
```diff
@@ -0,0 +1,164 @@
...
+* Boolean indicating whether to accept self-signed certificates for those endpoints
+* Full pathname to the key file
```
afishbeck, line 108: As mentioned elsewhere we could make this a secret... even in baremetal if we put the file in the right place the secret code will find it.

system/security/plugins/jwtSecurity/README.md
```diff
@@ -0,0 +1,164 @@
...
+true.
+
+### HPCC Systems Authorization and JWT Claims
+
+This plugin supports all authorizations documented in the HPCC Systems® Administrator's Guide with the exception of
+"View Permissions".  Loosely speaking, the permissions are divided into three groups:  Feature, Workunit Scope, and File
+Scope.
+
+Feature permissions are supported exactly as documented.  A specific permission would exist as a JWT claim, by name,
+with the associated value being the name of the permission.  For example, to grant read-only access to ECL Watch, use
```
afishbeck, line 131: In the future if we support other sources of JWTs and have less control over say the naming of claims or structure of the token we should implement mapping that says "for permission xyz the path to look at in the token is a.b.c and the value required is x".  That is the purpose of our mapping files now, but if we support multiple sources of jwts then we may need one per source.

afishbeck: My point being that when we can dictate the JWT format / claims as we can now, that's great.. we may not always be able to in the future.

system/security/plugins/jwtSecurity/jwtEndpoint.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+            headers = curl_slist_append(headers, "Content-Type: application/json;charset=UTF-8");
+            curl_easy_setopt(curlHandle, CURLOPT_HTTPHEADER, headers);
+
+            curl_easy_setopt(curlHandle, CURLOPT_URL, jwtEndPoint.c_str());
+            curl_easy_setopt(curlHandle, CURLOPT_POST, 1);
+            curl_easy_setopt(curlHandle, CURLOPT_POSTFIELDS, credentialsStr.c_str());
+            if (allowSelfSignedCert)
+                curl_easy_setopt(curlHandle, CURLOPT_SSL_VERIFYPEER, 0);
+            curl_easy_setopt(curlHandle, CURLOPT_FOLLOWLOCATION, 1);
```
afishbeck, line 117: Do you find these settings are necessary for practical reasons?  Seems like not passing credentials somewhere else would be more secure.

system/security/plugins/jwtSecurity/jwtSecurity.cpp
```diff
@@ -0,0 +1,1129 @@
...
+
+    /**
+     * Returns a copy of the string argument with particular whitespace
+     * characters removed from the end.
+     *
+     * @param   s           The string to process
+     *
+     * @return  New string with trailing whitespace characters removed.
+     */
+    static std::string TrimString(const std::string& s)
```
afishbeck, line 276: Any reason these methods start with uppercase and others don't?  Methods are usually lowercase in most of our code.

system/security/plugins/jwtSecurity/jwtSecurity.cpp
```diff
@@ -0,0 +1,1129 @@
...
+                jwtVerifier = jwtVerifier.with_issuer(endpointURL);
+
+            // Add a nonce claim only if a nonce value was provided
+            if (!nonce.empty())
+                jwtVerifier = jwtVerifier.with_claim("nonce", jwt::claim(nonce));
+
+            // Append signature algorithm verifier
+            std::string included_algo = str_toupper(decodedToken.get_algorithm());
+            if (included_algo == "HS256")
+                jwtVerifier = jwtVerifier.allow_algorithm(jwt::algorithm::hs256(keyContents));
```
afishbeck, line 354: I've heard of attacks where you are configured with a public key but the token says it uses a symmetric algorithm... which it used the public key to encrypt.. i.e. the mismatch allowed it to fool the receiver that it was a valid token.  Do we know if that will work here?  If not perhaps we can recognize if we are using a symmetric or public key and restrict the algorithms?

afishbeck:
@dcamper  some questions / comments.  I wonder a bit about globally caching a single token per username.  Especially if we ever support multiple sources of tokens, but can't think of too compelling a reason to change it now.

dali/server/daserver.cpp
```diff
@@ -139,6 +139,61 @@ void usage(void)
...
+
+#ifdef _CONTAINERIZED
+    IPropertyTree *jwtConfigPTree = configTree->queryPropTree("jwt");
+
+    if (jwtConfigPTree)
+    {
+        StringAttr configName(jwtConfigPTree->queryProp("@objname"));
+
+        VStringBuffer secMgrXML("<SecurityManager name='%s' instanceFactoryName='createInstance' libName='libjwtSecurity.so' type='JwtSecurityManager'/>", configName.str());
+        foundTree.set(createPTreeFromXMLString(secMgrXML));
```
afishbeck, line 154: You are taking double ownership here.  In a successful scenario it's fine because you return ownership and the caller releases, but if there is an exception this will leak.  I think it's better to use setown() and then when you return do "return foundTree.getClear()".  That will pass the ownership to the caller, but if there is an exception it will release the object.

dali/server/daserver.cpp
```diff
@@ -139,6 +139,61 @@ void usage(void)
...
+
+            if (authMethod && stricmp(authMethod, "secmgrPlugin") == 0)
+            {
+                const char* authPluginType = proptree->queryProp("@authPluginType");
+
+                if (authPluginType)
+                {
+                    VStringBuffer xpath("SecurityManagers/SecurityManager[@name='%s']", authPluginType);
+
+                    foundTree.set(configTree->getPropTree(xpath));
```
afishbeck, line 182: And here.

esp/applications/eclservices/jwt_authorization_map.yaml
dcamper: Fixed.  Also fixed in esp/applications/eclwatch/jwt_authorization_map.yaml.

system/security/plugins/jwtSecurity/README.md
dcamper: Good point.  I will work with the endpoint developer to get that implemented.

system/security/plugins/jwtSecurity/README.md
dcamper: Fixed.

system/security/plugins/jwtSecurity/README.md
dcamper: I don't think we should require https, but I could be convinced otherwise.  A sealed network environment, or a dev environment, may not want the additional complexity.  Let me know if you want this requirement added.

system/security/plugins/jwtSecurity/README.md
dcamper: I think this is something we should revisit only if it comes up.  What you say makes sense, but I'm not convinced the need for a map like that will arise in the real world.  I'd like to see a concrete example before embarking on a mapping implementation.

system/security/plugins/jwtSecurity/jwtEndpoint.cpp
dcamper: CURLOPT_UNRESTRICTED_AUTH is actually not needed; it is used only for basic auth, which we are not using.  That's what I get for copying and pasting from another project.  I've removed this one.

CURLOPT_FOLLOWLOCATION is needed to handle 3xx redirects used in some load balancer situations.

Are there other settings of concern?

system/security/plugins/jwtSecurity/jwtSecurity.cpp
dcamper: More copying and pasting madness (I confess to using that style in my own code).  Fixed.

system/security/plugins/jwtSecurity/README.md
afishbeck: If we send password hashes it may be less critical.

system/security/plugins/jwtSecurity/README.md
afishbeck: Yes, we can revisit if and when needed.

system/security/plugins/jwtSecurity/jwtEndpoint.cpp
afishbeck: I don't see any others that concern me.

system/security/plugins/jwtSecurity/jwtSecurity.cpp
dcamper: That is clever.  I've added a check for public key format, then limiting the algorithms checked appropriately.  Great idea!

system/security/plugins/jwtSecurity/jwtEndpoint.cpp
```diff
@@ -0,0 +1,192 @@
...
+        // Signals an error to libcurl
+        incomingDataSize = 0;
+
+        PROGLOG("CJwtSecurityManager::CaptureIncomingCURLReply() exceeded buffer size %ld", MAX_BUFFER_SIZE);
+    }
+
+    return incomingDataSize;
+}
+
+std::string _tokenFromEndpoint(const std::string& jwtEndPoint, bool allowSelfSignedCert, const std::string& credentialsStr)
```
afishbeck, line 87: Some inconsitent naming schemes here.  Lower case for extern?  Underscore for shared internal?  And uppercase for some other internals?
Should the internal functions be static?

dcamper: More copying and pasting from another project.  Thanks for catching these.  I've made the internal functions static as well, but I'm not sure if that makes much of a difference here.

system/security/plugins/jwtSecurity/README.md
dcamper: I will definitely change to sending a hashed password, as noted elsewhere (I think).

system/security/plugins/jwtSecurity/jwtEndpoint.cpp
afishbeck: Thanks.  Different compilers have different default export semantics and trying to get in the habit of using static to avoid name collisions.

system/security/plugins/jwtSecurity/jwtEndpoint.cpp
```diff
@@ -0,0 +1,192 @@
...
+{
+    std::string     apiResponse;
+
+    // Call JWT endpoint
+    CURL*   curlHandle = curl_easy_init();
```
afishbeck, line 92: Have you thought through the thread safety if this is the place the global curl init gets done?  What if we start using libcurl more places?  It looks like our current use of libcurl all comes from 3rd party libraries... but once you start using it other components might follow suit.

afishbeck: "If you did not already call curl_global_init, curl_easy_init does it automatically. This may be lethal in multi-threaded cases, since curl_global_init is not thread-safe, and it may result in resource problems because there is no corresponding cleanup.

You are strongly advised to not allow this automatic behaviour, by calling curl_global_init yourself properly. See the description in libcurl(3) of global environment requirements for details of how to use this function."

afishbeck: FYI I recently added https://github.com/hpcc-systems/HPCC-Platform/tree/master/system/httplib for REST calls related to my secrets.  I didn't do a plus and minus comparison with libcurl.  I wasn't aware it was already a build dependency.  httplib is a single header implementation which simplified the dependencies (and perhaps other issues).

dali/server/daserver.cpp
dcamper: Fixed.

system/security/plugins/jwtSecurity/jwtEndpoint.cpp
dcamper: httplib looks interesting.  It is definitely a subset implementation, though.  I wonder how it handles redirects and self-signed certs, for instance.  (Those two things come to mind just from this JWT work.)

In looking around for an example of a module-based construction/destruction I stumbled across an azure library in our code base that seems to also initialize libcurl incorrectly:  azure/azure-storage-cpplite/include/http/libcurl_http_client.h.

Would it be appropriate to put the global init management into MODULE_INIT/MODULE_EXIT functions to at least mitigate problems with this library?

esp/applications/eclqueries/jwt.yaml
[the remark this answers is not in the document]
dcamper: Tony, this implies that key location is not user-configurable, right?  Or are you expecting the code to parse a path (ominous grumblings at this point)?

system/security/plugins/jwtSecurity/jwtEndpoint.cpp
afishbeck: A proper call to the global init should then fix any use of curl_easy_init() later since it only calls global_init if it hasn't already.  Since it can apply to other 3rd party usage as well we may want a real fix.

@ghalliday any thoughts on where or how we should make a truly threadsafe global call curl_global_init()?  This could also benefit the AWS library use of libcurl.

From AWS library:

        bool initAndCleanupCurl;
        /**
        * libCurl infects everything with its global state. If it is being used then we automatically initialize and clean it up.
        * If this is a problem for you, set this to false. If you manually initialize libcurl please add the option CURL_GLOBAL_ALL to your init call.
        */

esp/applications/eclqueries/jwt.yaml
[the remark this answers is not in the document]
afishbeck: Well for my stuff going forward I hope to start treating all keys, certs, credentials, etc. as managed secrets.  The bare metal case is a little tricky.. not in how you find the secret, that just uses the jsecrets API, but in how the secret gets to the secret location in the first place.
Vaults can be used in k8s or baremetal, but again how it gets set up for baremetal is a question.
For k8s the secret would either be configured through k8s or stored in a vault and the configuration is pretty straight forward (I think).
Now the truth is if using secrets it would probably be best to combine everything to do with talking to and verifying tokens for a given jwt server under one secret with multiple child named values.. including the URL.  If you wanted to go that route you could look at my example for creating secrets for soapcall to use.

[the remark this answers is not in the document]
dcamper: That helps, thanks.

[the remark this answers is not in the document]
dcamper: This is done.  I took your advice and swapped 'secretsPath' in for 'keyFilePath'.  I know in the vault that doesn't make as much sense, but 'secrets' sound too much like actually defining the secret value rather than its location.

system/security/plugins/jwtSecurity/jwtEndpoint.cpp
dcamper: It would be cool to have the engines perform global libcurl init/cleanup at binary execution/termination time.  From what I see in the docs, that would be safe and ensure safety for any function calling curl_easy_init().

dcamper:
Tony, I've pushed the changes from your review, except for the libcurl issue.  Please take a look.  Thanks!

system/security/plugins/jwtSecurity/README.md
dcamper: Implemented SHA256.

afishbeck: Should we consider PBKDF2, bcrypt or scrypt to make attacking the hash more difficult?  Or perhaps at least HMAC-SHA256?

afishbeck: We only hash once per session and then use the token we receive so the overhead of doing this once may be worth it?

dcamper: HMAC requires a secret key.  Do you envision this as a configuration item?

bcrypt uses Blowfish as a hash, which is not one of Blowfish's original design goals.  OpenSSL doesn't support that use for that reason.

afishbeck: I was just listing schemes to consider.  Don't think adding an extra key is desirable, unless one was already provided. Perhaps scrypt then, since it is included in openssl and I don't think has key requirements?

dcamper: I am not in favor of adding another secret, either.  If the goal is prevent rainbow attacks, what about SHA256(client_id + password)?  In other words, seeding the hash with the JWT client_id (which is defined as a configuration parameter) before hashing in the password?

dcamper: Actually, do it the other way around (pw + client_id).  That way the endpoint can store only sha256(pw) and then mix the client_id in to verify.

system/security/plugins/jwtSecurity/jwtEndpoint.cpp
ghalliday: @afishbeck I think it would be safe to call it in one of the initialisation blocks.  E.g.

```
MODULE_INIT(INIT_PRIORITY_STANDARD)
{
    cur_global_init();
    return true;
}
MODULE_EXIT()
{
    curl_global_cleanup();
}

```
Then it should be called after the dll has been loaded - even if it is loaded dynamically - which should avoid the potential loaded locking the documentation mentions.

That does mean that any options set by other uses will be ignored.  I'm not sure if that will cause issues.

system/security/plugins/jwtSecurity/README.md
dcamper: Settled this a bit differently:  SHA256(SHA256(pw) + nonce) will be sent to the endpoint.  The endpoint saves SHA256(pw) in a database, so on that end they compare the sent hash with SHA256(retrieved_hash + nonce) to validate.

dcamper:
@afishbeck @RussWhitehead Squashed and rebased.  All container support has been removed to avoid muddying the waters.  Please take a look.  Thanks!

afishbeck (approved):
@dcamper looks fine to me now.  Not sure if we need @jakesmith to take a look since it touches DALI code.

richardkchapman:
Also needs rebasing onto latest master - and preferably changing to target 7.12.0 branch

richardkchapman:
@jakesmith can you review

dcamper:
Rebased and retargeted.

system/security/plugins/jwtSecurity/jwtSecurity.cpp
```diff
@@ -0,0 +1,1128 @@
...
+                getSecretValue(rawKeyContents, "system", secretsPath, "key", true);
+
+                if (rawKeyContents.isEmpty())
+                    throw MakeStringException(-1, "CJwtSecurityManager: Key at path '%s' is empty", secretsPath.str());
+
+                keyContents = trailingStringTrim(trailingStringTrim(rawKeyContents.str()));
+                keyIsPublicKey = isPrefixString("-----BEGIN PUBLIC KEY-----", keyContents);
+
+                // Save last check time
+                keyFilePathLastChecked = time(nullptr);
```
jakesmith, line 584: I think better to have a CCycleTimer member and reset here.

system/security/plugins/jwtSecurity/jwtSecurity.cpp
```diff
@@ -0,0 +1,1128 @@
...
+                    const char*     resourceName = resource->getName();
+                    SecAccessFlags  accessFlag = _authorizeEx(rtype, user, resourceName, secureContext, userInfo);
+
+                    resource->setAccessFlags(accessFlag);
+                }
+            }
+
+            result = true;
+        }
+        catch (...)
```
jakesmith, line 917: in general catch IException as well or only.

system/security/plugins/jwtSecurity/jwtSecurity.cpp
```diff
@@ -0,0 +1,1128 @@
...
+                    }
+                }
+                else
+                {
+                    // No token in the K/V store;
+                    return SecAccess_Unavailable;
+                }
+            }
+
+            if (key && stricmp(key,"Scope") == 0)
```
jakesmith, line 1047: trivial: could use strisame

system/security/plugins/jwtSecurity/jwtSecurity.cpp
```diff
@@ -0,0 +1,1128 @@
...
+    StringBuffer                loginEndpoint;              //!< Full URL to login endpoint; from configuration
+    StringBuffer                refreshEndpoint;            //!< Full URL to refresh endpoint; from configuration
+    bool                        allowSelfSignedCert;        //!< true = allow self-signed certificate; from configuration
+    StringBuffer                secretsPath;                //!< Secrets vault space or subdirectory name containing key; from configuration; @see ensureKeyLoaded()
+    time_t                      keyFilePathLastChecked;     //!< Timestamp of last time keyFilePath was checked for modification; @see ensureKeyLoaded()
+    SecAccessFlags              defaultPermAccess;          //!< Default permission value for permissions; from configuration
+    SecAccessFlags              defaultWUAccess;            //!< Default permission value for workunit scopes; from configuration
+    SecAccessFlags              defaultFileAccess;          //!< Default permission value for file scopes; from configuration
+    std::string                 keyContents;                //!< Contents of secret key; @see ensureKeyLoaded()
```
jakesmith, line 1093: not sure about mixed use of StringBuffer and std::string in same new class, would be clearer if consistent.

jakesmith:
@dcamper - various minor comments.

dali/server/daldap.cpp
```diff
@@ -272,6 +273,20 @@ class CDaliLdapConnection: implements IDaliLdapConnection, public CInterface
 };
 
 
+IDaliLdapConnection *createDaliSecMgrPluginConnection(IPropertyTree *propTree)
```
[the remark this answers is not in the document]
dcamper: The SecLoader::loadPluggableSecManager() call within that function will not accept a const.  Unless you want me to change it there as well?  I haven't checked to see if that is possible.

system/security/plugins/jwtSecurity/jwtSecurity.cpp
```diff
@@ -0,0 +1,1128 @@
...
+                    userInfo.addFileScopePerm(x->scopePattern, SecAccess_Access, defaultFileAccess, true);
+                else if (x->claim == "DenyFileScopeRead")
+                    userInfo.addFileScopePerm(x->scopePattern, SecAccess_Read, defaultFileAccess, true);
+                else if (x->claim == "DenyFileScopeWrite")
+                    userInfo.addFileScopePerm(x->scopePattern, SecAccess_Write, defaultFileAccess, true);
+                else if (x->claim == "DenyFileScopeFull")
+                    userInfo.addFileScopePerm(x->scopePattern, SecAccess_Full, defaultFileAccess, true);
+            }
+
+            gUserCache.set(subject, userInfo);
```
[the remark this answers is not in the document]
dcamper: Agreed that it is a bit expensive, but I deliberately chose this method for thread safety (once the object is copied we don't care what happens to the global JWTCache contents).  It is extremely easy to reason about the safety with this style.  Is higher performance necessary here?  Or necessary enough to warrant a rewrite?

dali/server/daldap.cpp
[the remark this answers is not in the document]
jakesmith: right ok, not unless it's trivial.

dcamper:
@jakesmith Changes made, temporarily kept in a separate commit for re-review.  Please take another look.  Thanks!

system/security/plugins/jwtSecurity/jwtSecurity.cpp
```diff
@@ -1044,9 +1053,9 @@ class CJwtSecurityManager : implements IDaliLdapConnection, public CBaseSecurity
                 }
             }
 
-            if (key && stricmp(key,"Scope") == 0)
+            if (key && strisame(key,"Scope"))
```
jakesmith, line 1056: minor: don't need to check for key=null with strisame
(and on line 1053)

jakesmith:
@dcamper - a couple of minor comments, and question re. verifyToken cost and use of allocated object

system/security/plugins/jwtSecurity/jwtSecurity.cpp
dcamper: verifyToken should be called only under three conditions on a per-user basis:

1. During initial login
2. During token refresh
3. Resurrecting a token from Dali, which happens only when copying a token between processes (e.g. esp stores the token and Dali retrieves it)

The count of the second two items is related to token lifetime, and repeated during a user's session.  A token with a 15-minute lifetime means that those steps will be repeated roughly every 15 minutes, per user.  Actual trigger is based on usage rather than a timer, though:  Permission checks include a check for token timeout, which could trigger a refresh.  Actual token lifetime is defined by the JWT endpoint.

Once verified, a token is stored in a cache and not verified again.

Even with hundreds of concurrent users, I don't think this would be burdensome.

So yeah, I'm pushing back slightly on changing this but if you feel that I should make the change to an owned pointer, let me know.

dcamper:
@jakesmith Additional changes made in a new commit.  Thanks!

system/security/plugins/jwtSecurity/jwtSecurity.cpp
jakesmith: Fair enough, it doesn't sound like it will be called enough to impact performance.
Personally, I'd still change it, it should be a minor change that doesn't affect thread safety, and avoids unnecessary copying of the various hash maps that each JWTUserInfo contains.

jakesmith:
@dcamper - see reply re. verifyToken, but looks good. Please squash.

dcamper:
@jakesmith Squashed!

dcamper:
@richardkchapman Please merge.  Thanks!

system/security/plugins/jwtSecurity/jwtCache.cpp
```diff
@@ -0,0 +1,392 @@
...
+    return !fileScopePermMap.empty();
+}
+
+//-----------------------------------------------------------------------------
+// Private methods
+//-----------------------------------------------------------------------------
+
+std::string JWTUserInfo::_globToExample(const std::string& pattern) const
+{
+    std::string     s1 = std::regex_replace(pattern, std::regex("\\[![0-9].+?\\]\\*?"), "a");
```
ghalliday, line 230: Likely to be more efficient if std::regex were static constants outside the function.  I don't know how time critical this function is though.

system/security/plugins/jwtSecurity/jwtCache.cpp
```diff
@@ -0,0 +1,392 @@
...
+
+bool JWTUserCache::get(const std::string& userName, JWTUserInfo& userInfo)
+{
+    {
+        CriticalBlock               block(crit);
+        UserPermissionMap::iterator foundIter = userPermMap.find(userName);
+
+        if (foundIter != userPermMap.end())
+        {
+            userInfo = foundIter->second;
```
ghalliday, line 359: Potentially expensive

system/security/plugins/jwtSecurity/jwtCache.hpp
```diff
@@ -0,0 +1,455 @@
...
+};
+
+/**
+ * This is basically a thread-safe map of username->JWTUserInfo objects.
+ */
+class JWTUserCache
+{
+    private:
+
+        typedef std::map<std::string, JWTUserInfo> UserPermissionMap;
```
ghalliday, line 433: This is potentially expensive to add elements and resize because the child object has reasonably complicated members.  It will also be expensive to return.

It might be better to have std::map<std::string, std::unique_ptr<JWTUserInfo>>, or a link counted pointer.

system/security/plugins/jwtSecurity/jwtEndpoint.cpp
```diff
@@ -0,0 +1,230 @@
...
+#include <openssl/sha.h>
+#include <string.h>
+
+#include "jlog.hpp"
+
+#include "nlohmann/json.hpp"
+
+#include "jwtEndpoint.hpp"
+
+struct MemoryStruct
```
ghalliday, line 29: It looks like a MemoryBuffer would provide the functionality you need and simplify the code.

system/security/plugins/jwtSecurity/jwtEndpoint.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+    return secondHash;
+}
+
+static std::string tokenFromEndpoint(const std::string& jwtEndPoint, bool allowSelfSignedCert, const std::string& credentialsStr)
+{
+    std::string     apiResponse;
+
+    // Call JWT endpoint
+    CURL*   curlHandle = curl_easy_init();
```
ghalliday, line 120: How expensive is this call?  Should it be cached in a thread local variable?

system/security/plugins/jwtSecurity/jwtCache.cpp
```diff
@@ -0,0 +1,392 @@
...
+    // Ideally, this function should be traversing a tree of scope permissions, with each
+    // scope level a tree node; that would be much faster and would also avoid requiring
+    // ordering multiple patterns prior to calling _addScopePerm()
+    std::string         scopeAsExample(_convertPathname(_globToExample(scope)));
+    SecAccessFlags      accessValue = defaultFlag;
+    int                 longestScopeLen = 0;
+
+    for (PermissionMap::const_iterator x = scopePermMap.begin(); x != scopePermMap.end(); x++)
+    {
+        if (::fnmatch(x->first.c_str(), scopeAsExample.c_str(), (FNM_CASEFOLD | FNM_LEADING_DIR)) == 0)
```
ghalliday, line 297: Is this available on all platforms, or do we disable on windows?  Is there a better, more portable way to do the same, e,g. using std::regex, which could be precalulated in the scope permissions.

system/security/plugins/jwtSecurity/jwtSecurity.cpp
```diff
@@ -0,0 +1,1134 @@
...
+
+            // Stuff claims into a new cache object for this user
+            JWTUserInfo userInfo;
+
+            userInfo.setExpirationTime(std::chrono::system_clock::to_time_t(decodedToken.get_expires_at()));
+            userInfo.setRefreshToken(refreshToken);
+            userInfo.setJWTToken(token);
+            for (auto& e : decodedToken.get_payload_claims())
+            {
+                std::string     key(e.first);
```
ghalliday, line 422: could this be const std::string &?

system/security/plugins/jwtSecurity/jwtCache.cpp
dcamper: I've moved the std::regex parameters to static const std::regex file variables.

system/security/plugins/jwtSecurity/jwtEndpoint.cpp
dcamper: Initializing libcurl's handle is not expensive, though some memory (for state) is set aside and structs initialized.  This is called very infrequently, only during initial user login and during token refreshes.

system/security/plugins/jwtSecurity/jwtCache.cpp
dcamper: The original design was to leverage file globbing rather than regex for patterns because they are more intuitive (primarily because globbing is not greedy).  Windows supposedly does have a similar function -- PathMatchSpecA() -- but I am not at all sure it _really_ functions the same as fnmatch() and I lack a Windows development environment to test.

My suspicion is that if someone is running a cluster on Windows then they are more likely to use Active Directory than JWT as an authentication/authorization scheme.  Therefore, I would argue that this plugin should be disabled for Windows.

That said, I can be convinced otherwise.  Do we have a hand-rolled pattern matcher in jlib I can use instead?  If so, that may  be an even better route.

system/security/plugins/jwtSecurity/jwtEndpoint.cpp
dcamper: Swapped in MemoryBuffer.  Good call.

system/security/plugins/jwtSecurity/jwtCache.hpp
dcamper: Switched to using std::unique_ptr for the map value.

dcamper:
@ghalliday I've made the changes you requested, except for ::fnmatch, and I've kept them in a separate commit for now.  Please take another look.  Thanks!

dcamper:
Closing this PR because of difficulties with loading it.  Will open a new PR.

dcamper:
New PR:  https://github.com/hpcc-systems/HPCC-Platform/pull/14210

PR #14195, opened 2020-09-21
HPCC-24761 Add Benford Law test code to Std.DataPatterns

dcamper:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Any new interfaces or exported functions are appropriately commented.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

## Testing:

New self tests added.

### DIFF
--- a/ecllibrary/std/DataPatterns/Benford.ecl
+++ b/ecllibrary/std/DataPatterns/Benford.ecl
@@ -1,0 +1,320 @@
+/**
+ * Benford's law, also called the Newcomb–Benford law, or the law of anomalous
+ * numbers, is an observation about the frequency distribution of leading
+ * digits in many real-life sets of numerical data.
+ *
+ * Benford's law doesn't apply to every set of numbers, but it usually applies
+ * to large sets of naturally occurring numbers with some connection like:
+ *
+ *      Companies' stock market values
+ *      Data found in texts — like the Reader's Digest, or a copy of Newsweek
+ *      Demographic data, including state and city populations
+ *      Income tax data
+ *      Mathematical tables, like logarithms
+ *      River drainage rates
+ *      Scientific data
+ *
+ * The law usually doesn't apply to data sets that have a stated minimum and
+ * maximum, like interest rates or hourly wages. If numbers are assigned,
+ * rather than naturally occurring, they will also not follow the law. Examples
+ * of assigned numbers include: zip codes, telephone numbers and Social
+ * Security numbers.
+ *
+ * For more information: https://en.wikipedia.org/wiki/Benford%27s_law
+ *
+ * This function computes the distribution of digits within one or more
+ * attributes in a dataset and displays the result, one attribute per row,
+ * with an "expected" row showing the expected distributions.  Included
+ * in each data row is a chi-squared computation for that row indicating how
+ * well the computed result matches the expected result (if the chi-squared
+ * value exceeds the one shown in the --EXPECTED-- row then the data row
+ * DOES NOT follow Benford's Law).
+ *
+ * Note that when computing the distribution of the most significant digit,
+ * the digit zero is ignored.  So for instance, the values 0100, 100, 1.0,
+ * 0.10, and 0.00001 all have a most-significant digit of '1'.  The digit
+ * zero is considered for all other positions.
+ *
+ * @param   inFile          The dataset to process; REQUIRED
+ * @param   fieldListStr    A string containing a comma-delimited list of
+ *                          attribute names to process; note that attributes
+ *                          listed here must be top-level attributes (not child
+ *                          records or child datasets); use an empty string to
+ *                          process all top-level attributes in inFile;
+ *                          OPTIONAL, defaults to an empty string
+ * @param   digit           The 1-based digit within the number to examine; the
+ *                          first significant digit is '1' and it only increases;
+ *                          OPTIONAL, defaults to 1, meaning the most-significant
+ *                          non-zero digit
+ * @param   sampleSize      A positive integer representing a percentage of
+ *                          inFile to examine, which is useful when analyzing a
+ *                          very large dataset and only an estimated data
+ *                          analysis is sufficient; valid range for this
+ *                          argument is 1-100; values outside of this range
+ *                          will be clamped; OPTIONAL, defaults to 100 (which
+ *                          indicates that all rows in the dataset will be used)
+ *
+ * @return  A new dataset with the following record structure:
+ *
+ *      RECORD
+ *          STRING      attribute;   // Name of data attribute examined
+ *          DECIMAL4_1  zero;        // Percentage of rows with digit of '0'
+ *          DECIMAL4_1  one;         // Percentage of rows with digit of '1'
+ *          DECIMAL4_1  two;         // Percentage of rows with digit of '2'
+ *          DECIMAL4_1  three;       // Percentage of rows with digit of '3'
+ *          DECIMAL4_1  four;        // Percentage of rows with digit of '4'
+ *          DECIMAL4_1  five;        // Percentage of rows with digit of '5'
+ *          DECIMAL4_1  six;         // Percentage of rows with digit of '6'
+ *          DECIMAL4_1  seven;       // Percentage of rows with digit of '7'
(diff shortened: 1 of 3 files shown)


dcamper:
@richardkchapman Please assign for review.  Thanks!

ecllibrary/std/DataPatterns/Benford.ecl
```diff
@@ -0,0 +1,314 @@
...
+        const char*     ch = s;
+
+        while (*ch)
+        {
+            if (isdigit(*ch) && (pos > 1 || *ch != '0'))
+            {
+                ++digitsFound;
+                if (digitsFound >= pos)
+                {
+                    foundDigit = *ch - 48;
```
richardkchapman, line 185: More normal (and more self-documenting) to say *ch - '0'

ecllibrary/std/DataPatterns/Benford.ecl
```diff
@@ -0,0 +1,314 @@
...
+ * digits, with "-- EXPECTED DIGIT n --" showing as the attribute name.'n' will
+ * be replaced with the value of 'digit' which indicates which digit position
+ * was examined.
+ *
+ * Note that when viewing the results for the mosts significant digit (digit = 1),
+ * the 'zero' field will show a -1 value, indicating that it was ignored.
+ */
+EXPORT Benford(inFile, fieldListStr = '\'\'', digit = 1, sampleSize = 100) := FUNCTIONMACRO
+
+    #UNIQUENAME(minDigit);
```
richardkchapman, line 91: You seem to be going through a lot of effort to make sure that code cannot be shared between uses here.

I suspect Benford should be a MODULE rather than a FUNCTIONMACRO

richardkchapman:
Feels like a case where the introspection that was nearly added to 7.0 would have been useful... It may be that enough of it is accessible (if not documented) to be used...

@ghalliday Can you take a look

ecllibrary/std/DataPatterns/Benford.ecl
```diff
@@ -0,0 +1,314 @@
...
+    LOCAL %workingInFile% :=
+        #IF(%trimmedFieldList% = '')
+            %sampledData%
+        #ELSE
+            TABLE(%sampledData%, {#EXPAND(%trimmedFieldList%)})
+        #END;
+
+    // Helper function that returns the first non-zero digit in a string
+    #UNIQUENAME(FirstDigit);
+    LOCAL UNSIGNED1 %FirstDigit%(VARSTRING s, UNSIGNED1 pos) := EMBED(C++)
```
ghalliday, line 173: Is the parameter a string or a varstring?   The casts in the calls are to varstrings, but the parameter is a string.  I suspect better to have a string parameter (len, ptr)

ecllibrary/std/DataPatterns/Benford.ecl
```diff
@@ -0,0 +1,314 @@
...
+        unsigned char   foundDigit = 10; // impossible value
```
ghalliday, line 174: Could do with 

#option pure

otherwise it will be re-evaluated for each call.  (I suspect that is currently 10 times.)

ecllibrary/std/DataPatterns/Benford.ecl
```diff
@@ -0,0 +1,314 @@
...
+                    #SET(recLevel, %recLevel% - 1)
+                #ELSEIF(%recLevel% = 0)
+                    #SET(fieldNum, %fieldNum% + 1)
+                    + TABLE
+                        (
+                            %workingInFile%(LENGTH(REGEXREPLACE('[^\\d]', (STRING)%@name%, '')) >= %minDigit%),
+                            {
+                                UNSIGNED2   %idField% := %fieldNum%,
+                                DECIMAL4_1  zero := IF(%clampedDigit% = 1, -1.0, COUNT(GROUP, %FirstDigit%((STRING)%@name%, %clampedDigit%) = 0) / COUNT(GROUP) * 100),
+                                DECIMAL4_1  one := COUNT(GROUP, %FirstDigit%((STRING)%@name%, %clampedDigit%) = 1) / COUNT(GROUP) * 100,
```
ghalliday, line 250: Seems a shame to call function %FirstDigit% multiple times

ecllibrary/teststd/DataPatterns/TestBenford.ecl
```diff
@@ -0,0 +1,168 @@
...
+        ASSERT(BenfordResults1[2].seven = (DECIMAL4_1)0),
+        ASSERT(BenfordResults1[2].eight = (DECIMAL4_1)0),
+        ASSERT(BenfordResults1[2].nine = (DECIMAL4_1)0),
+        ASSERT(BenfordResults1[2].chi_squared = (DECIMAL7_3)166.496),
+        ASSERT(BenfordResults1[2].num_values = 100)
+    ];
+
+    SHARED BenfordResults2 := DataPatterns.Benford(TestData, digit := 2);
+    SHARED TestBenford2 := [
+        ASSERT(BenfordResults2[1].zero = (DECIMAL4_1)12),
```
ghalliday, line 50: FYI You can use D as a suffix on a number to make a decimal constant.  E.g. 11.4D.

ghalliday:
@dcamper I'm not sure the introspection would really solve the problem, but I think the code could be restructured to avoid everything being in a macro.

ecllibrary/std/DataPatterns/Benford.ecl
dcamper: Indeed, I do this (#UNIQUENAME et al) on purpose.  I've run into many cases where calling the same function macro multiple times from the same scope creates problems due to symbol reuse.  This particular function is a good example of a function that could be called multiple times with slightly different arguments in the same scope.  Using #UNIQUENAME with LOCAL for an ECL attribute may be overkill, though, I admit.

I'm curious about using a MODULE in place of a FUNCTIONMACRO:  How would the code perform the macro-level introspection?

ecllibrary/std/DataPatterns/Benford.ecl
dcamper: Fixed.

ecllibrary/std/DataPatterns/Benford.ecl
dcamper: Converted parameter to STRING.

ecllibrary/std/DataPatterns/Benford.ecl
dcamper: Done.  I mistakenly thought that was the default.

ecllibrary/std/DataPatterns/Benford.ecl
dcamper: Agreed.  Is there a syntax where I can avoid that, within a TABLE()?

ecllibrary/teststd/DataPatterns/TestBenford.ecl
dcamper: Much cleaner; changed.

dcamper:
Changes from code review have been pushed.  Please take another look.

@ghalliday Hints regarding changes to avoid a FUNCTIONMACRO would be appreciated.  The introspection bit -- specifically, not knowing the structure of the input dataset and handling an optional subset of fields -- is the reason this is a macro.

dcamper:
@richardkchapman @ghalliday All changes have been made (including a bug fix or two).  Unless you would like a rewrite to avoid using a function macro, this should be ready for another look.  Thanks!

ecllibrary/std/DataPatterns/Benford.ecl
ghalliday: Unfortunately I don't think there is a simple way

ghalliday:
@richardkchapman looks better to me.

I'm not sure the benefit would out weight the cost of restructuring, unless this is likely to be called many times from the same query.  Even then I suspect most things will be commoned up by the code gen.

richardkchapman:
I agree - for this particular use case it's unlikely to cause any issues as it is, and it matches the style used by the rest of the DataPatterns code.

PR #14204, opened 2020-09-25
HPCC-24778 Fix potential include path issues for persistent.hpp

mayx:
- Specify the correct include path in espprotocol.hpp

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Cloud-compatibility
  - [x] Premature optimization

## Smoketest:

### DIFF
--- a/esp/platform/espprotocol.hpp
+++ b/esp/platform/espprotocol.hpp
@@ -26,7 +26,7 @@
 //SCM Interfaces
 #include "esp.hpp"
 #include "xslprocessor.hpp"
-#include "persistent.hpp"
+#include "../../common/thorhelper/persistent.hpp"
 
 //STL
 #include <algorithm>


mayx:
@jakesmith please review...

esp/platform/espprotocol.hpp
```diff
@@ -26,7 +26,7 @@
 //SCM Interfaces
 #include "esp.hpp"
 #include "xslprocessor.hpp"
-#include "persistent.hpp"
+#include "../../common/thorhelper/persistent.hpp"
```
jakesmith, line 29: is that always the relative path (if espprotocol.hpp is included from different projects) ?
 
@mayx - Why is the include path (common/thorhelper) not in the cmake file?

mayx: I tested, even if the file is included in another layer, for example esp/services/ws_account, this still works. I think the compiler looks for the relative path from the location of the hpp (espprotocol.hpp), not where it's included

mayx: I only updated the include path of projects under HPCC-Platform, but some projects outside of it also uses espprotocol.hpp.

jakesmith: do you mean LN (non-community) projects?
Still not sure why fix isn't to update the CMakeLists.txt files to include the correct path?

mayx:
Will update the include paths instead...

esp/platform/espprotocol.hpp
richardkchapman: Better still would be to structure the includes so that we did not include this file from a header file, but rather from those source files that actually need it.

mayx: The interface defined in the header file is needed by the class defined in the other header file. We're good for now, but I can look into whether it's possible to improve on this...

PR #14219, opened 2020-10-02
HPCC-24802 Improve WsStore pre-configured store process

rpastrana:
- Eliminates misleading warning
- Base class does not attempt to create pre-existing stores

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/services/ws_store/espstorelib/daliKVStore.cpp
+++ b/esp/services/ws_store/espstorelib/daliKVStore.cpp
@@ -35,7 +35,7 @@ bool CDALIKVStore::createStore(const char * apptype, const char * storename, con
     VStringBuffer xpath("Store[%s='%s'][1]", DALI_KVSTORE_NAME_ATT,  storename);
     if (root->hasProp(xpath.str()))
     {
-        IWARNLOG("DALI KV Store: Cannot create app '%s' entry, it already exists", storename);
+        LOG(MCuserInfo,"DALI Keystore createStore(): '%s' entry already exists", storename);
         return false;
     }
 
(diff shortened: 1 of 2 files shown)


rpastrana:
@jakesmith please review

esp/services/ws_store/espstorelib/daliKVStore.cpp
```diff
@@ -35,7 +35,7 @@ bool CDALIKVStore::createStore(const char * apptype, const char * storename, con
     VStringBuffer xpath("Store[%s='%s'][1]", DALI_KVSTORE_NAME_ATT,  storename);
     if (root->hasProp(xpath.str()))
     {
-        IWARNLOG("DALI KV Store: Cannot create app '%s' entry, it already exists", storename);
+        LOG(MCuserInfo,"DALI Keystore createStore(): '%s' entry already exists", storename);
```
jakesmith, line 38: Is it going to be common to reach here / is it normal for it to already exist?
If so, I'd have thought better not to log each time, only when it is actually created, rather than every time already exists.
But if this code path is only expected to be hit on startup or similar (low-frequency) it makes sense as it.

jakesmith:
@rpastrana - 1 minor question, but looks fine.

esp/services/ws_store/espstorelib/daliKVStore.cpp
rpastrana: We were reaching this code because the base class blindly attempted to create pre-configured stores on startup. That logic has been updated. This function is exposed as a WebService method and can be accessed by end users, but it should be very infrequent. If it ends up being called many times, this log entry will help expose a problem in its usage.

rpastrana:
@richardkchapman please merge

PR #14224, opened 2020-10-12
HPCC-24792 Add validateFileType to #OPTION docs

JamesDeFabia:
Signed-off-by: Jim DeFabia <jamesdefabia@lexisnexis.com>
@jakesmith please review

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/docs/EN_US/ECLLanguageReference/ECLR_mods/Templ-OPTION.xml
+++ b/docs/EN_US/ECLLanguageReference/ECLR_mods/Templ-OPTION.xml
@@ -531,6 +531,15 @@
             </row>
 
             <row>
+              <entry><emphasis>validateFileType</emphasis></entry>
+
+              <entry>Default: true</entry>
+
+              <entry>If false, the compiler uses the definition in the ECL
+              workunit and ignores the file type in the DFU record.</entry>
+            </row>
+
+            <row>
               <entry><emphasis>compressInternalSpills</emphasis></entry>
 
               <entry>Default true</entry>


docs/EN_US/ECLLanguageReference/ECLR_mods/Templ-OPTION.xml
```diff
@@ -530,6 +530,15 @@
               <entry>Upper limit of a CSV line read in MB.</entry>
             </row>
 
+            <row>
+              <entry><emphasis>validateFileType</emphasis></entry>
+
+              <entry>Default: true</entry>
+
+              <entry>If false, the compiler uses the definition in the ECL
+              workunit and ignores the file type in the DFU record.</entry>
```
jakesmith, line 539: This isn't a compile time option.
It's checked at query run time by the engines.
e.g. when a query is run on hthor and thor and looks up a file, if this option is on (default),
if the published file has a file type, it validates that the ECL used to read it, matches.

NB: the JIRA description should be updated too.

jakesmith: Also, there are exceptions to when this check is applied:

1. if the ECL is reading CSV, the check is ignored, since CSV is commonly used to read files of an unknown format.
2. if the ECL is reading a fixed record width flat file, the check is ignored. Similar to (1), there are exceptions when reading a blocks of fixed width records.
3. if the ECL is reading XML or JSON, and there's a mismatch - it will only issue a warning not an error.
I can't remember why, but sometimes folks read XML or JSON back as e.g. CSV.

jakesmith (changes requested):

jakesmith:
@JamesDeFabia - back to you.

PR #14226, opened 2020-10-13
HPCC-24788 ESP services need to know if the request is canceled

mayx:
- Add an interface and callback function for services to check and
  be notified if the peer has closed the connection
- Use jsocket's wait_read and avail_read to detect socket connection

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Cloud-compatibility
  - [x] Premature optimization

## Smoketest:

### DIFF
--- a/esp/platform/espprotocol.cpp
+++ b/esp/platform/espprotocol.cpp
@@ -755,3 +755,24 @@ int CEspProtocol::countBindings(int port)
     else
         return apport->getBindingCount();
 }
+
+bool checkEspConnection(IEspContext& ctx)
+{
+    CHttpRequest* req = dynamic_cast<CHttpRequest*>(ctx.queryRequest());
+    if (!req)
+        return false;
+    ISocket* sock = req->getSocket();
+    if (!sock)
+        return false;
+    if (sock->OShandle() == INVALID_SOCKET)
+        return false;
+    int ret = sock->wait_read(0);
+    if (ret < 0)
+        return false;
+    else if (ret > 0)
+    {
+        if (sock->avail_read() == 0)
+            return false;
+    }
+    return true;
+}
--- a/esp/platform/espprotocol.hpp
+++ b/esp/platform/espprotocol.hpp
@@ -215,4 +215,6 @@ public:
     virtual int countBindings(int port);
 };
 
+esp_http_decl bool checkEspConnection(IEspContext& ctx);
+
 #endif


mayx:
@afishbeck @mckellyln please review. @wangkx please check if this would fit your need. 

There're 2 ways to utilize the close detector, one is to synchronously call the "isPeerClosed()" function, the other is to set a callback function that will be called back asynchronously as soon as the close is detected. You can choose to use either way, or both ways.

An example to is as follows:
```
//will be called asynchronously as soon peer closing is detected
void mycallback(void* data)
{
    //Now you can use the info you put in data to stop the running method
    //The type and format of data is totally up to you to define
    DBGLOG("Received notice that peer closed connection!");
}

bool Cws_accountEx::onVerifyUser(IEspContext &context, IEspVerifyUserRequest &req, IEspVerifyUserResponse &resp)
{
    Owned<IConnectionCloseDetector> detector = createConnectionCloseDetector(&context);
    detector->setCallback(mycallback, nullptr); //Setting up callback is optional. callbackdata is set to nullptr here, but should  be set
                                                //to an object you define that contains enough info to stop your running method and do proper clean up if necessary.
    Sleep(2000); //do something
    if (detector->isPeerClosed())
    {
        DBGLOG("Peer closed!");
        return false;
    }
    Sleep(2000); //do something else
    if (detector->isPeerClosed())
    {
        DBGLOG("Peer closed!");
        return false;
    }
    return true;
}
```

wangkx:
Thanks Yanrui. This does fit my need.

esp/platform/espprotocol.cpp
```diff
@@ -755,3 +755,92 @@ int CEspProtocol::countBindings(int port)
...
+                        m_detector->notifyClose();
+                        break;
+                    }
+                    else
+                        Sleep(10); // avoid tight loop in case there're still outstanding data on the wire.
+                }
+            }
+            return 0;
+        }
+    };
```
mckellyln, line 801: The idea that poll() returns ok and then the input queue is empty *suggests strongly* the socket was closed at the other end.  Probably works great in all our cases.
I'm not sure it would behave any differently, but do also wonder about using 
recv(MSG_DONTWAIT|MSG_PEEK)
instead ?
How long would we wait ?  Should keepalive be enabled so that socket remains ok until user explicitly calls close() ?

mayx: Hi Mark, the use case of the function is for some long-running service methods (for example the WsWorkunit::WuResult method), where the user sometimes would stop the method from the browser, thus making it pointless for the service method to keep running. So the detector is simply a non-intrusive way to check if the connection is closed by the other end, and it's not responsible for keeping the connection alive. The detector waits till the connection is closed, or when the service method is completed (when the detector object is released to be exact), whichever comes first.

The jlib avail_read() function calls ioctl(sock, FIONREAD, &avail) underneath, and it works pretty well from tests. recv(MSG_DONTWAIT|MSG_PEEK) will likely work as good, but I think ioctl is slightly better here because it's less intrusive, in that we don't need to allocate a buffer to actually receive any data.

mayx: Of course for tls avail_read() does use SSL_peek and SSL_pending, a solution that you and I worked out sometime ago...

mckellyln: I was thinking of:
char data[2];
recv(fd, data, 1, MSG_DONTWAIT|MSG_PEEK)
But I agree its probably no better than ioctl(FIONREAD), I just wanted to mention this as an alternative that is maybe a little more portable :-)

esp/platform/espprotocol.cpp
```diff
@@ -755,3 +755,92 @@ int CEspProtocol::countBindings(int port)
...
+    CloseNotifyCallback m_callbackfunc = nullptr;
+    void* m_callbackdata = nullptr;
+    bool m_isPeerClosed = false;
```
mckellyln, line 804: If one thread is reading m_isPeerClosed and another thread is writing m_isPerrClosed then maybe it should be an atomic ?  I mean if it gets updated slightly later perhaps its fine.

mayx: I don't think atomic will buy us anything here, the reader will get either true or false, with or without atomic, depending on the timing of the call...

mayx: True that the reader could've get a true with atomic but may get a false without, but that's a very, very rare occasion (writer calls first but finishes later), and even that happens it's no big deal, the reader will get the correct result next time it checks, and the callback function still sends back the notice instantaneously. Basically it's not worth the overhead to handle a rare harmless situation...

mckellyln: Thats what I meant - as long as you are ok with how it could be then its fine.

mayx: Yes this is ok.

mckellyln: Same for stop ?  Is it ok or even possible to call m_detector->notifyClosed() from its dtor ?

mayx: Yes it's the same for stop. I think it's better to call notifyClosed() right away instead of waiting for the destructor to avoid delay or complications.

esp/platform/espprotocol.cpp
mckellyln: Does the parent (or any other) thread call wait_read() + read() on the same socket at the same time ?
If another thread is calling wait_read() + read() often, is there ever a chance of a race where the other thread reads all the data so this threads avail_read() returns 0 ?

mayx: No this function will be called in the service method, after all the request data has all been read in.

esp/platform/espprotocol.cpp
mckellyln: But its in a thread running in a loop ?

esp/platform/espprotocol.cpp
mckellyln: But its in a thread running in a loop ?

mayx: No it's not, read is done and there's no more loop there...

mckellyln: Isn't this a loop that happens in a thread ?
```
while(!stop)
    wait_read()
    if avail_read() == 0
       // socket must have been closed ...
```

mayx: Yes this thread runs in a loop, but it's the only thread that does wait_read(), no other thread does wait_read() or read() at the same time...

esp/platform/espprotocol.cpp
```diff
@@ -755,3 +755,92 @@ int CEspProtocol::countBindings(int port)
...
+    }
+    virtual ~ConnectionCloseDetector()
+    {
+        if (m_closeDetectionThread)
+            m_closeDetectionThread->stop();
+    }
+    virtual bool isPeerClosed()
+    {
+        return m_isPeerClosed;
+    }
```
mckellyln, line 835: If m_isPeerClosed is true should m_closeDetectionThread->stop() be called ?

mckellyln:
I think this is ok.  The callback feature is nice, but do wonder about making it more than it needs to be ?  For example, if the use case is the parent calling isPeerClosed() periodically, then could this be:
```
bool IsClientConnectionOK(IEspContext* ctx)
{
    if (!ctx || !ctx->queryRequest())
        return false;
    CHttpRequest* req = dynamic_cast<CHttpRequest*>(ctx->queryRequest());
    if (!req)
        return false;
    ISocket* sock = req->getSocket();
    if (!sock)
        return false;
    int ret = sock->wait_read(0);
    if (ret == 0)
        return true;
    else if (ret < 0)
        return false;
    else
        if (sock->avail_read() <= 0)
            return false;
    return true;
}
```
It should be noted somehwere the use case as designed is that the parent is not expected to do any reads while this is running.  (altho the window for any race would be extremely tiny)

mayx:
@mckellyln you brought up a good point. If asynchronous call back is not needed, then we don't need to create a separate thread to do this. @wangkx do you need the asynchronous call back feature, or do you only need to synchronously/actively check if the connection is closed from time to time in your code flow?

wangkx:
@mayx For now, I just need the connection check. Can we make the callback feature an option just in case we need it in the future?

mayx:
Yes we can do that.

mayx:
@mckellyln change made in a new commit per latest discussions, please review again...

esp/platform/espprotocol.cpp
```diff
@@ -809,20 +811,36 @@ class ConnectionCloseDetector : implements IConnectionCloseDetector, public CInt
...
     {
         if (!ctx || !ctx->queryRequest())
             return;
         CHttpRequest* req = dynamic_cast<CHttpRequest*>(ctx->queryRequest());
         if (!req)
             return;
-        ISocket* sock = req->getSocket();
-        if (!sock)
+        m_sock = req->getSocket();
+        if (!m_sock)
```
mckellyln, line 836: Is it ok for ctor to return without creating m_sock ?  How will parent know it failed ?

esp/platform/espprotocol.cpp
```diff
@@ -809,20 +811,36 @@ class ConnectionCloseDetector : implements IConnectionCloseDetector, public CInt
         if (m_callbackfunc)
             m_callbackfunc(m_callbackdata);
     }
+    bool checkClose()
+    {
+        int ret = m_sock->wait_read(0);
```
mckellyln, line 816: Is m_sock always ok ?

esp/platform/espprotocol.cpp
```diff
@@ -831,16 +849,13 @@ class ConnectionCloseDetector : implements IConnectionCloseDetector, public CInt
     }
     virtual bool isPeerClosed()
     {
+        if (!m_isAsync && !m_isPeerClosed)
```
mckellyln, line 852: Perhaps isPeerClosed() could be called both when non-async and async ?

mayx: isPeerClosed() can be called by both sync and async. Here the check is to control whether to call checkClose(), which is not necessary for async because it's done in the async thread...

esp/platform/espprotocol.cpp
mayx: I can throw an exception.

esp/platform/espprotocol.cpp
mckellyln: ok, sure.

esp/platform/espprotocol.cpp
mckellyln: ok.

mayx:
@mckellyln added exception in new commit, please review again...

mckellyln:
ok, cool.
Is there any chance m_sock goes away ?  Does it need checking before referencing ?

mayx:
We're just holding a pointer don't think there's a way to check. For the use case the socket won't go away...

mayx:
If we change to a pointer reference, we can check if the pointer was set to null in its original location. Slightly safer I guess...

mckellyln:
ok.  I just don't like the idea of a thread checking m_sock->func() in a loop if m_sock could ever go away due an exception or whatever.

mayx:
@mckellyln I've changed the socket to "Owned" to make sure it won't go away. Please review it in the new commit...

mayx:
@mckellyln I just added code to check if the socket is still valid in the same commit, please review again...

mayx:
After further thinking and discussion, I think it's better to make the solution simple. The original implementation was done with asynchronous callback in mind. Since that is not needed now (we can always add it when it's needed), we can simplify the implementation to a function call checkEspConnection, which returns true if the connection is good. I also squashed the commits. @mckellyln please review again...

esp/platform/espprotocol.cpp
```diff
@@ -755,3 +755,24 @@ int CEspProtocol::countBindings(int port)
     else
         return apport->getBindingCount();
 }
+
+bool checkEspConnection(IEspContext& ctx)
+{
```
mckellyln, line 760: Should you check if ctx is ok before this cast ?

mckellyln:
@mayx just one question.

esp/platform/espprotocol.cpp
mayx: No it's guaranteed to be not null where it's going to be called (reason it's passed in as reference both here and the service method)

mckellyln: ok

mckellyln:
Looks good to merge.

mckellyln (approved):

richardkchapman:
Is this needed in 7.12.x or is master ok?

mckellyln:
@wangkx will use this for some upcoming code.  We need to ask him or perhaps @afishbeck if its needed in 7.12.

afishbeck:
Since this is a new function that doesn't pose any risk on it's own I would say add it to 7.12.x.  That way we can decide on a case by case basis which branch any changes Kevin might make should go in.

richardkchapman:
Merged manually to 7.12.x

PR #14252, opened 2020-10-27
HPCC-24910 Records may still be pulled by PullerThread after stopped

richardkchapman:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [x] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/roxie/ccd/ccdserver.cpp
+++ b/roxie/ccd/ccdserver.cpp
@@ -2184,7 +2184,8 @@ protected:
     IEngineRowStream *inputStream;
     IRecordPullerCallback *helper;
     Semaphore started;                      // MORE: GH->RKC I'm pretty sure this can be deleted, since handled by RestartableThread
-    bool groupAtOnce, eof, eog;
+    bool groupAtOnce, eog;
+    std::atomic<bool> eof;
     CriticalSection crit;
 
 public:
@@ -2277,6 +2278,7 @@ public:
             CriticalBlock c(crit); // stop is called on our consumer's thread. We need to take care calling stop for our input to make sure it is not in mid-nextRow etc etc.
             if (inputStream)
                 inputStream->stop();
+            eof = true;
         }
         RestartableThread::join();
     }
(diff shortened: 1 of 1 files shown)


richardkchapman:
@ghalliday Please review

roxie/ccd/ccdserver.cpp
```diff
@@ -2297,12 +2298,13 @@ class RecordPullerThread : public RestartableThread
...
             return false;
-        while (preload)
+        while (preload && !eof)
         {
-            const void * row;
+            const void * row = nullptr;
             {
                 CriticalBlock c(crit); // See comments in stop for why this is needed
-                row = inputStream->nextRow();
+                if (!eof)
```
ghalliday, line 2306: It probably isn't very speed critical, but 
```
    if (eof)
       break

``` might generate better code (and would avoid the need for the check on the while).  With the check on the while eof should probably be an atomic.

ghalliday: Actually return false; rather than break;

ghalliday:
@richardkchapman what version should this target.  I would have thought 7.12.x, but possibly earlier.  A comment on the code, and checkboxes need filling in.

richardkchapman:
@ghalliday Edited as suggested, retargeted, and checkboxes filled in

roxie/ccd/ccdserver.cpp
richardkchapman: I changed it as suggested, then changed my mind and changed it back. I don't think you want to skip the eog processing?

You might argue that we should use a separate variable rather than eof to signify "stopped asynchronously"

ghalliday: Fair enough.  eof, should probably be atomic then, otherwise compilers could avoid the retest within the critical section.

ghalliday: @richardkchapman

PR #14254, opened 2020-10-27
HPCC-24896 Base64 encode hashed user password in JWT SecMgr plugin

dcamper:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/security/plugins/jwtSecurity/jwtEndpoint.cpp
+++ b/system/security/plugins/jwtSecurity/jwtEndpoint.cpp
@@ -50,28 +50,34 @@ static size_t captureIncomingCURLReply(void* contents, size_t size, size_t nmemb
 static std::string hashString(const std::string& s)
 {
     SHA256_CTX  context;
-    char        hashedValue[SHA256_DIGEST_LENGTH + 1];
+    char        hashedValue[SHA256_DIGEST_LENGTH];
 
     memset(hashedValue, 0, sizeof(hashedValue));
 
     if (!SHA256_Init(&context))
         throw makeStringException(-1, "CJwtSecurityManager: OpenSSL ERROR calling SHA256_Init while hashing user password");
 
-    if (!SHA256_Update(&context, (unsigned char*)s.c_str(), s.size()))
+    if (!SHA256_Update(&context, (unsigned char*)s.data(), s.size()))
         throw makeStringException(-1, "CJwtSecurityManager: OpenSSL ERROR calling SHA256_Update while hashing user password");
 
     if (!SHA256_Final((unsigned char*)hashedValue, &context))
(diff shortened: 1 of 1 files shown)


dcamper:
@ghalliday Please review.  Thank you!

richardkchapman:
This should probably target 7.12.x ?

dcamper:
Argh.  You're right, Richard.  Rebased and retargeted to 7.12.x.

system/security/plugins/jwtSecurity/jwtEndpoint.cpp
```diff
@@ -50,28 +50,29 @@ static size_t captureIncomingCURLReply(void* contents, size_t size, size_t nmemb
...
 
-    return hashedValue;
+    return std::string(hashedValue, SHA256_DIGEST_LENGTH);
 }
 
-static std::string hashUserPW(const std::string& pw, const std::string& nonce)
+static void hashUserPW(const std::string& pw, const std::string& nonce, StringBuffer& encodedHash)
 {
-    std::string     firstHash(hashString(pw));
+    std::string     saltedPW(pw.substr(0, 2) + pw);
```
ghalliday, line 71: This doesn't seem to be directly related, and it doesn't seem to be a conventional password SALT.  Is it correct?

ghalliday (commented):
@dcamper the change looks sensible.  One question.

system/security/plugins/jwtSecurity/jwtEndpoint.cpp
dcamper: I should have noted that change in the Jira.  The reason behind it is, of course, to mitigate rainbow table attacks, both for transmitted and stored passwords.  Random values are preferred for salts but cannot be used in this unusual configuration, where the client is capturing credentials but authenticating happens remotely (well, perhaps you could, but it would involve an extra message exchange for every auth to trade random values).  In a full OpenID Connect implementation the user would present credentials only at the auth server, where random salts could be used.

This type of salt does have a security flaw in that an attacker with access to the saved passwords will be able to see if multiple users have the same password, even if they cannot use (common) rainbow tables to look up the plaintext version.

TL;DR: A two-character salt based on the user's pw value is not the best but it is better than an unsalted value and the compromise fits into our modified OpenID Connect auth scheme.

If you have a better idea for the above, let me know!

ghalliday: It sounds like a decent compromise.  It might be worth adding a comment to explain.  I can merge with or without the comment.

dcamper: Comment added and squashed.

PR #14273, opened 2020-11-02
HPCC-24897 Fix esdl tool xsl parameter handling

asselitx:
Changes introduced in HPCC-22357 treat xsl parameters to the
EsdlDefinitionHelper as expressions, requiring strings to be quoted.
Institute correct string quoting in the esdl tool where it uses
the EsdlDefinitionHelper.

Add regression test cases to exercise the esdl tool's xsd, wsdl and cpp
commands.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:


asselitx:
@timothyklemm and @afishbeck could you please review?

esp/esdllib/esdl_def_helper.cpp
```diff
@@ -249,9 +249,15 @@ void EsdlDefinitionHelper::toXSD( IEsdlDefObjectIterator &objs, StringBuffer &xs
...
+        // Since the xsl string params are quoted with ' we must remove
+        // it here before including it as the value of an attribute in
+        // the esxdl element
+        tns.stripChar('\'');
+        if( tns.isEmpty() )
+            tns.set("urn:unknown");
+
+        xml.appendf("<esxdl name=\"custom\" EsdlXslTypeId=\"%d\" xmlns:tns=\"%s\" ns_uri=\"%s\">", xslId, tns.str(), tns.str());
         this->toXML( objs, xml, version, opts, flags );
         xml.append("</esxdl>");
```
timothyklemm, line 262: This block was pretty simple and might not have warranted a stand-alone function, but the increased complexity may now justify a bit of refactoring to avoid deviations between the two occurrences.

tools/esdlcmd/esdlcmd_core.cpp
```diff
@@ -328,17 +328,20 @@ class Esdl2XSDCmd : public EsdlHelperConvertCmd
 
         generateNamespace(tns);
 
-        params->setProp( "tnsParam", tns );
-        params->setProp( "optional", optOptional );
+        // All params are treated as expressions, so any strings must be quoted
+        // 1/0 are not equivalent to true/false, as 0 evaluates to true
+
+        params->setProp( "tnsParam", StringBuffer('\'').append(tns).append('\'') );
```
timothyklemm, line 334: I would use VStringBuffer("'%s'", tns.str()) instead of StringBuffer('\'').append(tns).append('\'').

timothyklemm:
I can appreciate the appeal of using a real world example, but I think a minimalist example that clearly exercises the points of concern might be better. Does this example cover everything we care about today? Will it cover what we care about after the next ESDL revision and, if not, would we inject deviations from the real world into the existing test case or would we be expected to mimic a different real world example that may or may not be comprehensive?

esp/esdllib/esdl_def_helper.cpp
asselitx: Created a class method to encapsulate the logic of determining which namespace to use and unquoting it if required.

tools/esdlcmd/esdlcmd_core.cpp
asselitx: Done, wrapped in macro.

testing/esp/esdlcmd/README.rst
```diff
@@ -0,0 +1,26 @@
...
+For example, the Test Case named 'wsdl1' to generate the wsdl for WsDfu, has a file named
+
+    'wsdl1.wsdl'
+
+in the baselines directory. Likewise, the Test Case named
+
+    'cpp-dfu1'
+
+that generates c++ files for the WsDfu service has a folder named 'cpp-dfu1' in the baselines
+directory.
```
timothyklemm, line 26: WsDfu is no longer the example.

testing/esp/esdlcmd/baselines/wstest-cpp-installdir/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+cmake_minimum_required(VERSION 3.0)
```
timothyklemm, line 1: Copyright and license info? Comment applies to all new files, with the possible exception of the readme.

testing/esp/esdlcmd/baselines/wstest-cpp-installdir/WsTestServiceBase.cpp
```diff
@@ -0,0 +1,527 @@
+
+#include "WsTestServiceBase.hpp"
+#include "jliball.hpp"
+#include "jlog.hpp"
+#include "jptree.hpp"
+
+extern "C" int onWsTestAllVersionReport(const char* CtxStr, const char* ReqStr, StringBuffer& RespStr)
+{
+    Owned<WsTestServiceBase> service = createWsTestServiceObj();
+    return service->AllVersionReport(CtxStr, ReqStr, RespStr);
```
timothyklemm, line 10: It seems odd to return the 64 bit address of an allocated buffer as a 32 bit integer.

testing/esp/esdlcmd/baselines/wstest-cpp-installdir/WsTestService.hpp
```diff
@@ -0,0 +1,35 @@
...
+    }
+    virtual int MinVersion(const char* CtxStr, const char* ReqStr, StringBuffer& RespStr)
+    {
+        return WsTestServiceBase::MinVersion(CtxStr, ReqStr, RespStr);
+    }
+    virtual int Ping(const char* CtxStr, const char* ReqStr, StringBuffer& RespStr)
+    {
+        return WsTestServiceBase::Ping(CtxStr, ReqStr, RespStr);
+    }
+    virtual int VersionRangeReport(const char* CtxStr, const char* ReqStr, StringBuffer& RespStr)
```
timothyklemm, line 29: Can 'virtual' be replaced with 'override' where appropriate?

testing/esp/esdlcmd/baselines/wstest-cpp-installdir/WsTestServiceBase.cpp
```diff
@@ -0,0 +1,527 @@
...
+    Owned<IPropertyTree> ptree;
+    ptree.setown(createPTreeFromXMLString(ReqStr));
+    UnSe->unserialize(req.get(), ptree.get());
+    Owned<AllVersionReportResponse> resp = service->AllVersionReport(ctx.get(), req.get());
+    RespStr = "<Response><Results><Result><Dataset name=\"AllVersionReportResponse\"><Row>";
+    UnSe->serialize(resp.get(), RespStr);
+    RespStr.append("</Row></Dataset></Result></Results></Response>");
+
+    return 0;
+}
```
timothyklemm, line 464: If the service instance (line 453) overloaded a single method name (line 458), this method could be generalized as a template method to eliminate (currently) three additional nearly identical copies.

testing/esp/esdlcmd/baselines/wstest-cpp-installdir/WsTestServiceBase.hpp
```diff
@@ -0,0 +1,167 @@
...
+
+#include "jlib.hpp"
+#include "jptree.hpp"
+#include "jarray.hpp"
+#include "primitivetypes.hpp"
+
+using namespace std;
+using namespace cppplugin;
+
+class EsdlContext : public CInterface, implements IInterface
```
timothyklemm, line 12: Implementing IInterface is only necessary if you intend to treat the instance as an IInterface (which is not necessary for use with Owned). Is there a reason for inheriting both CInterface and IInterface in so many places?

testing/esp/esdlcmd/esdlcmd-test.py
```diff
@@ -0,0 +1,368 @@
...
+
+        if (not baseline.exists()):
+            logging.error('Missing baseline file %s', str(baseline))
+            return False
+
+        if (not outName.exists()):
+            logging.error('Missing output for test %s', self.name)
+            return False
+
+        if (not filecmp.cmp(str(baseline), str(outName))):
```
timothyklemm, line 137: I don't know python, so can't comment too much. One question, though, is whether equality is necessary or equivalence might be sufficient? In other words, is there an expectation that one input will yield the exact same ordering and whitespace every time?

testing/esp/esdlcmd/baselines/wstest-cpp-installdir/CMakeLists.txt
asselitx: These are auto-generated by the esdl tool, and are added here not as code but as baseline output for the test cases. If I add copyright comments then I'd need to take that into account when making the comparison against the test case output. I'd like to avoid that if I can but if the comments are required even this situation I'll do that.

testing/esp/esdlcmd/baselines/wstest-cpp-installdir/WsTestService.hpp
asselitx: This is auto-generated code by the esdl tool. I didn't comb through the code in detail, but included the case mostly to verify the correct handling of the INSTALL_DIR parameter.

testing/esp/esdlcmd/baselines/wstest-cpp-installdir/WsTestServiceBase.hpp
asselitx: There may be some opportunities to improve the auto-generated code, but I'm not going to address it as part of this ticket.

testing/esp/esdlcmd/esdlcmd-test.py
asselitx: I think that we should check for equality. Tony has mentioned that the wsdls and xsds generated should preserve order from the interface definitions.

testing/esp/esdlcmd/README.rst
asselitx: Ok I'll fix those comments.

asselitx: Ok I'll fix those comments.

asselitx:
@timothyklemm and @rpastrana could you please review this latest change?

rpastrana (approved):
@asselitx looks good. Skimmed over the tests and it ** seems ** fine to me. I left a couple of minor comments.
Also, are some of the classes in wstest-cpp-installdir generated? If so, does it make sense to generate at build time? Otherwise, do they need to be updated manually everytime the ECM files are updated?

timothyklemm (approved):
The readme reintroduced a reference to WsDfu, which is no longer the basis for the test case. But that seems trivial.

PR #14294, opened 2020-11-12
HPCC-24938 Remove "known positives" from helm chart test

richardkchapman:

## Type of change:

## Checklist:
  - [x] The commit is signed.
- x ] I have given due consideration to all of the following potential concerns:
  - [ ] Scalability
  - [ ] Performance
  - [x] Security
  - [ ] Thread-safety
  - [x] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

## Testing:
In progress

### DIFF
--- a/dockerfiles/platform-build/Dockerfile
+++ b/dockerfiles/platform-build/Dockerfile
@@ -20,8 +20,8 @@
 ARG BASE_VER=7.12
 FROM hpccsystems/platform-build-base:${BASE_VER}
 
-RUN groupadd -g 1000 hpcc
-RUN useradd -s /bin/bash -r -m -N -c "hpcc runtime User" -u 999 -g hpcc hpcc
+RUN groupadd -g 10001 hpcc
+RUN useradd -s /bin/bash -r -m -N -c "hpcc runtime User" -u 10000 -g hpcc hpcc
 RUN passwd -l hpcc 
 
 RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.17.0/bin/linux/amd64/kubectl && chmod +x ./kubectl && mv ./kubectl /usr/local/bin
--- a/dockerfiles/stopall.sh
+++ b/dockerfiles/stopall.sh
@@ -18,8 +18,29 @@
 ##############################################################################
 
 # Utility script for stopping a local cluster started by startall.sh
+wait=0
+
+while [ "$#" -gt 0 ]; do
+  arg=$1
+  case "${arg}" in
+      -w) wait=1
+         ;;
+      *) echo "Usage: stoptall.sh [options]"
+         echo "    -w  Wait for all pods to terminate"
+         exit
+         ;;
+    esac
+  shift
+done
 
 helm uninstall mycluster
 helm uninstall localfile
 kubectl delete jobs --all 
 kubectl delete networkpolicy --all 
+if [[ $wait == 1 ]] ; then
+  sleep 2
+  while (kubectl get pods | grep -q ^NAME) ; do
+    echo Waiting...
+    sleep 2
+  done
+fi
--- a/helm/hpcc/templates/_helpers.tpl
+++ b/helm/hpcc/templates/_helpers.tpl
@@ -109,6 +109,10 @@ logging:
 Add ConfigMap volume mount for a component
 */}}
 {{- define "hpcc.addConfigMapVolumeMount" -}}
+- name: {{ .name }}-temp-volume
+  mountPath: /tmp
+- name: {{ .name }}-hpcctmp-volume
+  mountPath: /var/lib/HPCCSystems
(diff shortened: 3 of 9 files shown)


richardkchapman:
Note - there doesn't seem to be a way to suppress the "readonly file system" or "low ID" issues in kube-score, so I decided to try to fix them instead.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -109,6 +109,8 @@ logging:
 Add ConfigMap volume mount for a component
 */}}
 {{- define "hpcc.addConfigMapVolumeMount" -}}
+- name: {{ .name }}-temp-volume
```
richardkchapman, line 112: Arguably this should be a separate helper function OR we should rename it

helm/hpcc/values.yaml
```diff
@@ -167,6 +167,7 @@ eclccserver:
   replicas: 1
   ## Set to false if you want to launch each workunit compile in its own container, true to run as child processes in eclccserver pod.
   useChildProcesses: false
+  keepJobs: all
```
richardkchapman, line 170: Would be nice if there was a way to override this on all debug systems...

I suspect will get to the point where I want to maintain a separate "debugging" values.yaml and use that from startall.sh - the debugging values.yaml may or may not be something we will want to check in.

testing/helm/run.sh
```diff
@@ -57,7 +57,12 @@ fi
 if type kube-score >/dev/null 2> /dev/null; then
    echo Running kube-score...
    # Note we force all replicas to be > 1 as some checks are not done on replicas=1 cases e.g. antiaffinity
-   helm template $hpccchart ${options} | sed "s/replicas: 1/replicas: 2/" | kube-score score --output-format ci - >results.txt 2>errors.txt
+   helm template $hpccchart ${options} | sed "s/replicas: 1/replicas: 2/" | \
+     kube-score score --output-format ci \
+        --ignore-container-cpu-limit \
+        --ignore-container-memory-limit \
+        --ignore-test deployment-has-poddisruptionbudget \
```
richardkchapman, line 64: Should create a JIRA to come back and look at what it would take to re-enable this one.

testing/helm/tests/secrets.yaml
```diff
@@ -1,5 +0,0 @@
-secrets:
-  ecl:
-    azure-magic: mymagic-secret
-  all:
```
richardkchapman, line 4: Or change to system rather than all ?

testing/helm/run.sh
richardkchapman: I have now done so

richardkchapman:
@ghalliday Please review

ghalliday:
@jakesmith please can you sanity check

helm/hpcc/values.yaml
richardkchapman: I removed this line ad created a (private, unreleased) version of values for debugging purposes instead.

dockerfiles/platform-build/Dockerfile
```diff
@@ -20,8 +20,8 @@
 ARG BASE_VER=7.12
 FROM hpccsystems/platform-build-base:${BASE_VER}
 
-RUN groupadd -g 1000 hpcc
-RUN useradd -s /bin/bash -r -m -N -c "hpcc runtime User" -u 999 -g hpcc hpcc
+RUN groupadd -g 10001 hpcc
+RUN useradd -s /bin/bash -r -m -N -c "hpcc runtime User" -u 10000 -g hpcc hpcc
```
jakesmith, line 24: 0-999 typically reserved for system accounts.
Any reason this is 10K(+1) vs 1K(+1) ?

dockerfiles/stopall.sh
```diff
@@ -18,8 +18,29 @@
...
+done
 
 helm uninstall mycluster
 helm uninstall localfile
 kubectl delete jobs --all 
 kubectl delete networkpolicy --all 
+if [[ $wait == 1 ]] ; then
+  sleep 2
+  while (kubectl get pods | grep -q ^NAME) ; do
+    echo Waiting...
```
jakesmith, line 43: trivial: as it's our internal utility script may not matter, but there could be other pods in same namespace that this picks up, that are not to with hpcc.
Perhaps log matches it's still waiting for?

jakesmith (approved):
a couple of trivial comments/questions, but looks fine afaics.

dockerfiles/stopall.sh
richardkchapman: Or don't use -w in cases where it's not useful...

dockerfiles/platform-build/Dockerfile
richardkchapman: Because that's what kube-score requires to consider it acceptable...

PR #14308, opened 2020-11-17
HPCC-24895 Return session token in VerifyUser response

wangkx:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/scm/ws_account.ecm
+++ b/esp/scm/ws_account.ecm
@@ -71,11 +71,12 @@ ESPrequest VerifyUserRequest
 
 ESPresponse [exceptions_inline] VerifyUserResponse
 {
-	int retcode;
+    [min_ver("1.06")] unsigned SessionToken;
+    int retcode;
 };
 
 //Kevin/russ does this method need feature level check?
-ESPservice [auth_feature("NONE"), version("1.05"), default_client_version("1.05"), exceptions_inline("./smc_xslt/exceptions.xslt")] ws_account
+ESPservice [auth_feature("NONE"), version("1.06"), default_client_version("1.06"), exceptions_inline("./smc_xslt/exceptions.xslt")] ws_account
 {
     ESPmethod [client_xslt("/esp/xslt/account_myaccount.xslt")] MyAccount(MyAccountRequest, MyAccountResponse);
     ESPmethod [client_xslt("/esp/xslt/account_input.xslt")] UpdateUserInput(UpdateUserInputRequest, UpdateUserInputResponse);
--- a/esp/services/ws_account/ws_accountService.cpp
+++ b/esp/services/ws_account/ws_accountService.cpp
@@ -240,6 +240,7 @@ bool Cws_accountEx::onVerifyUser(IEspContext &context, IEspVerifyUserRequest &re
             throw MakeStringException(ECLWATCH_OLD_CLIENT_VERSION, "Client version not found");
         }
 
+        double version = context.getClientVersion();
         int minor = 0;
         int major = 0;
         const char* dot1 = strrchr(ver, '.');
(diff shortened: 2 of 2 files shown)


wangkx:
@afishbeck please review

afishbeck (approved):
@wangkx looks fine.

richardkchapman:
@ghalliday Can you comment?

This FEELS like it needs some extra security scrutiny to me...

ghalliday:
I don't completely understand how the sessions work.  Is the problem that the session id is returned in a cookie, but Gordon cannot get at the cookie locally?  Because httpOnly is set?  Is this new functionality more secure than disabling that flag?

My impression is that it isn't generally recommended to provide a session id as a part of the url.  It might be logged, or accessible by someone else.  What is the plan for using this session id once he has it?  Create a cookie to be sent for the child pages?

Definitely not this commit, but I noticed the recommended session id is a 128bits string, rather than 32bits.  The current method of generating session ids could also be improved.  (And there is no check for duplicates.)

wangkx:
The session id is returned in a cookie. But, Gordon cannot get the cookie because he is using a non browser based client application. So, in this fix, I return the session ID in the VerifyUser response. @GordonSmith could you please comment?

I just created https://track.hpccsystems.com/browse/HPCC-25010 for improving the session ID.

GordonSmith:
In this instance I will be setting the session token into a cookie in an embedded web page (which I hope will work).

Sequence of events:
1.  Client app calls verify user and gets a valid token
2.  Client app launched embedded webpage to display ecl watch with session token embedded in cookie

This is so the ECL IDE can display an ECL Watch page without having to login again (having logged into the IDE).

GordonSmith:
(I am not 100% sure this will work, and I may be adding a Jira to allow the session to passed via POST message as well.

afishbeck:
With this current design I don't think it's too far from the IDE just doing what the browser does anyway, although the session token is returned to the client in a different way (in the body rather than a cookie).

If we start sending back the token in an HTTP post we could consider changing the token into a new kind of signed token explicitly for this purpose.  Kind of like an oath token.  Gordon calls ESP to get the auth token, and then passes it through the browser to log the user in.  And that's when the session cookie gets created.

Or, we could do that anyway, even if Gordon doesn't need to use POST.

afishbeck:
I'm thinking we restrict this as much as we can.. and make it a more formal REST/SOAP login mechanism.  I don't think the current function even quite works for what Gordon wanted anyway since it returns an existing session which is the opposite of what we want for the IDE and for security.

Perhaps:
1. Instead of VerifyUser, create a new explicit REST/SOAP login function.. something like ClientLogin.
2. Can only be used from REST clients to do a fresh login.  Cannot be used to get my "current session id" especially from an ESP browser application.
3. As a REST client credentials will be there for the request... not an existing cookie.  Kevin can enforce that.
4. The token should be verifiable that it was created by ESP and has an entry in the DALI session store.
5. Ideally we would only support this over HTTPS.  Is that at all feasible at this point?
--- stage 2 --
6. When the token gets used it should have to come from the same origin IP as the login request.
7. We should never accept the token as a URL parameter.  Ideally the token would be passed in as credentials, not as part of the message.  Like as bearer token or as password.. and should only be accepted by one method...
We could add a another new method like StartClientSession that takes the login token as its credential and basically turns it into a session cookie.  With the same restrictions about IP, etc.  But comes from a browser that isn't logged in and logs it in.

After one time use, the token then is no longer useful for anything else. (session in dali is now in active state)

From Gordon's perspective he would
1. Make a REST/SOAP call to the login function and get a token.
2. Direct the browser to the StartClientSession method with the token as the credential.

GordonSmith:
@afishbeck this sounds plausible - it also sounds similar to some of the bigger oauth type stuff we are looking at for the cloud?  
Given this isn't high priority should we be waiting for the "full" solution, or do you see this as being different enough?

afishbeck:
@GordonSmith if this can wait for our OAUTH solution, and doesn't need to work with non-oauth environments then that would be ideal.  We need to solve OAUTH for command line and other clients as well.

GordonSmith:
@afishbeck I am ok with that.

wangkx:
Based on the latest comments, I am closing this PR.

PR #14311, opened 2020-11-18
HPCC-23841 Roxie activity reset without stop

richardkchapman:
SEQUENTIAL action would not stop upstream dependencies if one threw an
exception.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Cloud-compatibility
  - [x] Premature optimization

## Smoketest:

## Testing:
Ran agg1ds test with no dali attached. Reproduced the error prior to this fix, and behaves correctly after.

### DIFF
--- a/roxie/ccd/ccdserver.cpp
+++ b/roxie/ccd/ccdserver.cpp
@@ -21114,8 +21114,17 @@ public:
     virtual void doExecuteAction(unsigned parentExtractSize, const byte * parentExtract) 
     {
         unsigned numBranches = helper.numBranches();
-        for (unsigned branch=1; branch <= numBranches; branch++)
-            executeDependencies(parentExtractSize, parentExtract, branch);
+        try
+        {
+            for (unsigned branch=1; branch <= numBranches; branch++)
+                executeDependencies(parentExtractSize, parentExtract, branch);
+        }
+        catch (...)
+        {
+            for (unsigned branch=1; branch <= numBranches; branch++)
+                stopDependencies(parentExtractSize, parentExtract, branch);
+            throw;
+        }
     }
 
 };


richardkchapman:
@ghalliday Please review

roxie/ccd/ccdserver.cpp
```diff
@@ -21114,8 +21114,17 @@ class CRoxieServerSequentialActionActivity : public CRoxieServerActionBaseActivi
...
-            executeDependencies(parentExtractSize, parentExtract, branch);
+        try
+        {
+            for (unsigned branch=1; branch <= numBranches; branch++)
+                executeDependencies(parentExtractSize, parentExtract, branch);
+        }
+        catch (...)
+        {
+            for (unsigned branch=1; branch <= numBranches; branch++)
+                stopDependencies(parentExtractSize, parentExtract, branch);
```
ghalliday, line 21125: @richardkchapman 

Not directly connected... but why does stopDependencies call 

        dependencies.item(idx).stop();

rather than 

        dependencies.item(idx).stopSink(dependencyIndexes.item(idx));

I am wondering if this change might cause activities to be stopped, and not executed, that might otherwise be executed.  I'm not sure it could happen now, but might if there was a catchaction activity.  However stopDependendencies is called in other situations...

You may also only want to call stop on the branch that failed and any subsequent branches.

richardkchapman: You can't call stopSink on something that is not considered a sink. Which most dependencies are not (seems to be correlated to what the dependencyControlId is)

richardkchapman: Seems to me that if nobody above the sequential has caught the exception, then it's correct for the exception to stop anything that it would have called.

Anything below the execption that catches, may still go on to call other things, including other arms of splitters...

ghalliday: If branch 3 fails, then it will have already called stop on branches 1 and 2.  I  think that should be ok since the code allow stop() to be called too many times.

richardkchapman: Yes, I was replying to the first part of the comment, not the second. As you say, it's ok to call stop again on the branches that have already executed.

PR #14317, opened 2020-11-18
HPCC-24949 Add cc: and bcc: support to Std.System.Email.SendEmail* ECL functions

dcamper:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

## Testing:

Manual testing with fake SMTP server and reviewing SMTP transcripts.

### DIFF
--- a/common/remote/rmtsmtp.cpp
+++ b/common/remote/rmtsmtp.cpp
@@ -465,6 +465,8 @@ class CMailInfo
     StringArray *warnings;
     StringArray recipients;
     StringBuffer to;
+    StringBuffer cc;
+    StringBuffer bcc;
     StringAttr subject;
     StringAttr mailServer;
     unsigned port;
@@ -475,10 +477,12 @@ class CMailInfo
     unsigned inlen;
 
     static char const * toHeader;
+    static char const * ccHeader;
+    static char const * bccHeader;
     static char const * subjectHeader;
     static char const * senderHeader;
 public:
-    CMailInfo(char const * _to, char const * _subject, char const * _mailServer, unsigned _port, char const * _sender, StringArray *_warnings) 
+    CMailInfo(char const * _to, char const * _cc, char const * _bcc, char const * _subject, char const * _mailServer, unsigned _port, char const * _sender, StringArray *_warnings)
         : subject(_subject), mailServer(_mailServer), port(_port), sender(_sender), lastAction("process initialization"), inlen(0)
     {
         warnings = _warnings;
@@ -487,9 +491,23 @@ public:
             throw MakeStringException(0, "email sender address too long: %" I64F "u characters",  static_cast<__uint64>(sender.length()));
         validator.validateAddress(sender.get(), "email sender address");
 
-        getRecipients(validator, _to);
+        getRecipients(validator, _to, to);
         if(strlen(toHeader) + to.length() > 998)
             throw MakeStringException(0, "Email recipient address list too long: %u characters", to.length());
+
+        if (_cc && _cc[0])
+        {
+            getRecipients(validator, _cc, cc);
+            if(strlen(ccHeader) + cc.length() > 998)
+                throw MakeStringException(0, "Email cc: recipient address list too long: %u characters", cc.length());
+        }
+
+        if (_bcc && _bcc[0])
+        {
+            getRecipients(validator, _bcc, bcc);
+            if(strlen(bccHeader) + bcc.length() > 998)
+                throw MakeStringException(0, "Email bcc: recipient address list too long: %u characters", cc.length());
+        }
 
         if(strlen(subjectHeader) + subject.length() > 998)
             throw MakeStringException(0, "Email subject too long: %" I64F "u characters",  static_cast<__uint64>(subject.length()));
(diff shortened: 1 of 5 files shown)


common/remote/rmtsmtp.cpp
```diff
@@ -574,6 +586,8 @@ class CMailInfo
     {
         header.append(senderHeader).append(sender.get()).append("\r\n");
         header.append(toHeader).append(to.str()).append("\r\n");
+        if (!cc.isEmpty())
+            header.append(ccHeader).append(cc.str()).append("\r\n");
```
richardkchapman, line 590: Should bcc be here too?

richardkchapman:
May want to leave old prototypes in place for the sake of older workunits so they don't have to recompile.

richardkchapman:
@dcamper A couple of comments

common/remote/rmtsmtp.cpp
dcamper: No.  This is within the DATA section, which is the visible portion of the email the recipients see.  It has nothing to do with delivering the email.  We omit the bcc recipients to satisfy the "blind" part of the bcc definition.

dcamper:
@richardkchapman I've reinstated the original fileservices interfaces.  Please review this latest commit.  Thanks!

dcamper:
Don't merge; I just found an issue.

dcamper:
@richardkchapman Additional check for cc and bcc recipients add.  OK for review.  Thanks, and sorry for the confusion.

common/remote/rmtsmtp.cpp
```diff
@@ -487,10 +491,24 @@ class CMailInfo
...
+        {
+            getRecipients(validator, _cc, cc);
+            if(strlen(ccHeader) + cc.length() > 998)
+                throw MakeStringException(0, "Email cc: recipient address list too long: %u characters", cc.length());
+        }
+
+        if (_bcc && _bcc[0])
+        {
+            getRecipients(validator, _bcc, bcc);
+            if(strlen(bccHeader) + bcc.length() > 998)
```
richardkchapman, line 508: This 998 char limit seems a bit arbitrary (especially for bcc where people sometimes want 1000's of entries). Is that something from our code, of from the protocol?

common/remote/rmtsmtp.cpp
richardkchapman: Fair enough

common/remote/rmtsmtp.cpp
dcamper: I copied that check to match what was already there, thinking there was a Good Reason for it.

I think someone read the RFC wrong and imposed that limit where it didn't need to be.  RFC 821 (SMTP) says that there is a 1000 character limit for a "text line" but that really refers to SMTP commands, not the body (which is where these recipient strings wind up).  I think we could probably remove this check; do you agree?

RFC 821 says that a maximum of 100 recipients (to + cc + bcc) is allowed, but that is a per-server config setting and, in the wild, the actual limit is all over the place.  This limit is not checked in our code, but I'm not sure if it is worthwhile to add it because the _real_ limit is imposed by the mail server.

richardkchapman: Perhaps leave as is for now but open another Jira to remove the limit should anyone have the time/need

dcamper: Sounds good.  I've squashed the commits.

richardkchapman:
@ghalliday Can you sanity-check please

common/remote/rmtsmtp.cpp
```diff
@@ -574,6 +592,8 @@ class CMailInfo
     {
         header.append(senderHeader).append(sender.get()).append("\r\n");
         header.append(toHeader).append(to.str()).append("\r\n");
+        if (!cc.isEmpty())
+            header.append(ccHeader).append(cc.str()).append("\r\n");
         header.append(subjectHeader).append(subject.get()).append("\r\n");
```
ghalliday, line 597: Forgot to add the bcc entries.

ghalliday: I see richard's comment... and your answer!

dcamper: I wonder if this is an indication that I should add a comment describing why bcc is omitted, so a future maintainer doesn't "fix" it.

ghalliday (commented):
@dcamper I think the rest is ok - the order of parameters gets a bit confusing..

common/remote/rmtsmtp.cpp
ghalliday: probably worthwhile, although it was because two items were added at the same time that it stood out.

common/remote/rmtsmtp.cpp
```diff
@@ -790,28 +812,43 @@ static void doSendEmail(CMailInfo & info, CMailPart const & part)
...
+    CMailInfo info(to, cc, bcc, subject, mailServer, port, sender, warnings);
     CTextMailPart bodyPart(body, "text/plain; charset=ISO-8859-1", NULL);
     doSendEmail(info, bodyPart);
 }
 
-void sendEmailAttachText(const char * to, const char * subject, const char * body, const char * attachment, const char * mimeType, const char * attachmentName, const char * mailServer, unsigned int port, const char * sender, StringArray *warnings)
+void sendEmail(const char * to, const char * subject, const char * body, const char * mailServer, unsigned port, const char * sender, StringArray *warnings)
 {
-    CMailInfo info(to, subject, mailServer, port, sender, warnings);
+    sendEmail(to, subject, nullptr, nullptr, body, mailServer, port, sender, warnings);
```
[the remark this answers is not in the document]
dcamper: Good catch.  I just pushed a commit to fix.

dcamper:
@ghalliday Squash for merge?

ghalliday:
yes please - it is ready to merge.

dcamper:
Squashed.  Thank you!

PR #14322, opened 2020-11-19
HPCC-24575 Add ability to use jlog's JOB field

richardkchapman:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Cloud-compatibility
  - [x] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jlog.cpp
+++ b/system/jlib/jlog.cpp
@@ -187,6 +187,115 @@ unsigned getPositionOfField(unsigned logfields, unsigned positionoffield)
 
 // LogMsg
 
+LogMsgJobInfo::~LogMsgJobInfo()
+{
+    if (isDeserialized)
+        free((void *) jobIDStr);
+}
+
+const char * LogMsgJobInfo::queryJobIDStr() const
+{
+    if (isDeserialized)
+        return jobIDStr;
+    else if (jobID == UnknownJob)
+        return "Unknown";
+    else
+        return theManager->queryJobId(jobID);
+}
+
+LogMsgJobId LogMsgJobInfo::queryJobID() const
+{
+    if (isDeserialized)
+        return UnknownJob;  // Or assert?
+    else
+        return jobID;
+}
+
+void LogMsgJobInfo::setJobID(LogMsgUserId id)
+{
+    if (isDeserialized)
+        free((void *) jobIDStr);
+    jobID = id;
+    isDeserialized = false;
+}
+
+void LogMsgJobInfo::serialize(MemoryBuffer & out) const
+{
+    if (isDeserialized)
+        out.append(jobIDStr);
+    else
+        out.append(theManager->queryJobId(jobID));
+    out.append(userID);
+}
+
+void LogMsgJobInfo::deserialize(MemoryBuffer & in)
+{
+    StringBuffer idStr;
+    in.read(idStr);
+    jobIDStr = idStr.detach();
+    in.read(userID);
+    isDeserialized = true;
+}
+
+static LogMsgJobInfo globalDefaultJobInfo(UnknownJob, UnknownUser);
+static thread_local LogMsgJobInfo defaultJobInfo = globalDefaultJobInfo;
+
+const LogMsgJobInfo unknownJob(UnknownJob, UnknownUser);
+
+void resetThreadLogging()
+{
+    // Note - as implemented the thread default job info is determined by what the global one was when the thread was created.
+    // There is an alternative interpretation, that an unset thread-local one should default to whatever the global one is at the time the thread one is used.
+    // In practice I doubt there's a lot of difference as global one is likely to be set once at program startup
+    defaultJobInfo = globalDefaultJobInfo;
+}
+
+const LogMsgJobInfo & checkDefaultJobInfo(const LogMsgJobInfo & _jobInfo)
+{
+    if (&_jobInfo == &unknownJob)
+    {
+        return defaultJobInfo;
+    }
+    return _jobInfo;
+}
+
+void setDefaultJobId(const char *id, bool threaded)
+{
+    setDefaultJobId(theManager->addJobId(id), threaded);
+}
+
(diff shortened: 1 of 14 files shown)


system/jlib/jlog.cpp
```diff
@@ -1737,9 +1828,19 @@ void CLogMsgManager::doReport(const LogMsg & msg) const
 {
     try
     {
-        ReadLockBlock block(monitorLock);
-        ForEachItemIn(i, monitors)
-            monitors.item(i).processMessage(msg);
+        switch (msg.queryCategory().queryClass())
```
ghalliday, line 1831: I wonder if you should use a new audience of internal as well, so you can quickly isolate these new messages

system/jlib/jlog.cpp
```diff
@@ -1737,9 +1828,19 @@ void CLogMsgManager::doReport(const LogMsg & msg) const
...
+        case MSGCLS_addid:
+            doAddJobId(msg.queryJobInfo().queryJobID(), msg.queryText());
+            break;
+        case MSGCLS_removeid:
+            doRemoveJobId(msg.queryJobInfo().queryJobID());
+            break;
+        default:
+            ReadLockBlock block(monitorLock);
+            ForEachItemIn(i, monitors)
+                monitors.item(i).processMessage(msg);
```
ghalliday, line 1842: worth adding a break, even though not needed

system/jlib/jlog.cpp
```diff
@@ -187,6 +187,60 @@ unsigned getPositionOfField(unsigned logfields, unsigned positionoffield)
...
+{
+    // Note - as implemented the thread default job info is determined by what the global one was when the thread was created.
+    // There is an alternative interpretation, that an unset thread-local one should default to whatever the global one is at the time the thread one is used.
+    // In practice I doubt there's a lot of difference as global one is likely to be set once at program startup
+    defaultJobInfo = globalDefaultJobInfo;
+}
+
+const LogMsgJobInfo & checkDefaultJobInfo(const LogMsgJobInfo & _jobInfo)
+{
+    if (&_jobInfo == &unknownJob)
```
ghalliday, line 206: I think, better test is if _jobInfo.queryJobID() == UnknownJob

system/mp/mplog.cpp
```diff
@@ -610,12 +610,13 @@ void stopLogMsgReceivers()
 
 // LISTENER HELPER FUNCTIONS
 
-ILogMsgListener * startLogMsgListener()
+/*ILogMsgListener * startLogMsgListener()
```
ghalliday, line 613: yes, delete if unused

system/jlib/jlog.cpp
```diff
@@ -1538,6 +1592,43 @@ CLogMsgManager::~CLogMsgManager()
...
+    pushMsg(new LogMsg(id, nullptr));
+}
+
+const char * CLogMsgManager::queryJobId(LogMsgJobId id) const
+{
+    // NOTE - thread safety is important here. We have to consider two things:
+    // 1. Whether an id (and therefore an entry in this table) can be invalidated between the return statement and someone using the result
+    //    It is up to the calling application to ensure that it does not call removeJobId() on an ID that may still be being used for logging by another thread.
+    // 2. Whether the table lookup may coincide with a table add, and crash in getValue/setValue
+    //    This is a non-issue in queueing mode as all gets/sets happen on a single thread
```
ghalliday, line 1613: There is potentially a problem though if queryJobId() is called from any thread other than the logging thread.  I wonder if it should always lock..

ghalliday (commented):
@richardkchapman a few minor comments

system/jlib/jlog.cpp
richardkchapman: Depends whether the expense is a concern. I did wonder about a variant of CLeavableCriticalBlock that will assert that all accesses are from the same thread - but it's not easy to code without assuming a singleton

system/jlib/jlog.cpp
richardkchapman: Why?

ghalliday: Because it will work on a copy of unknownJob (e.g., uninitialised global variables)

system/mp/mplog.cpp
richardkchapman: Seems to be. I will remove

system/jlib/jlog.cpp
richardkchapman: It will also give wrong results if you tried to set just the user ID (though I don't know what the "right" result would be, or if userId is ever in use or likely to be (should we do similar things with userId as I just did with JobID ?

richardkchapman:
@ghalliday Please re-review

system/mp/mplog.hpp
```diff
@@ -57,13 +57,9 @@ extern mp_decl void stopLogMsgReceivers();
...
 
-// LISTENER HELPER FUNCTIONS
-
-extern mp_decl ILogMsgListener * startLogMsgListener();
-extern mp_decl void stopLogMsgListener();
-extern mp_decl bool connectLogMsgListenerToChild(INode * childNode);
-extern mp_decl bool connectLogMsgListenerToChildOwn(INode * childNode);
-extern mp_decl bool disconnectLogMsgListenerFromChild(INode * childNode);
-extern mp_decl bool disconnectLogMsgListenerFromChildOwn(INode * childNode);
+//extern mp_decl bool connectLogMsgListenerToChild(INode * childNode);
```
ghalliday, line 60: Should these all be deleted?

system/jlib/jlog.cpp
ghalliday: I still don't like the idea of comparing an address of an object, but it's probably ok.

ghalliday (commented):
@richardkchapman I think it is probably ready

richardkchapman:
@ghalliday Repushed to delete a few commented-out lines

PR #14324, opened 2020-11-19
HPCC-23805 Allow format of costs to be configured

shamser:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jstats.cpp
+++ b/system/jlib/jstats.cpp
@@ -25,6 +25,9 @@
 #include "jfile.hpp"
 #include "jerror.hpp"
 #include <math.h>
+#include <sstream>
+#include <iomanip>
+#include <ctype.h>
 
 #ifdef _WIN32
 #include <sys/timeb.h>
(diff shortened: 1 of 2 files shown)


shamser:
Documentation jira:https://track.hpccsystems.com/browse/HPCC-24900

shamser:
@ghalliday

system/jlib/jstats.cpp
```diff
@@ -396,6 +400,43 @@ static StringBuffer & formatIPV4(StringBuffer & out, unsigned __int64 value)
     return out.appendf("%d.%d.%d.%d", ip1, ip2, ip3, ip4);
 }
 
+static const char *getMoneyLocale()
+{
+    static bool haveLocale = false;
```
ghalliday, line 405: this isn't thread safe.  Better would be to use querySingleton() in jmutex.hpp.

system/jlib/jstats.cpp
```diff
@@ -396,6 +400,43 @@ static StringBuffer & formatIPV4(StringBuffer & out, unsigned __int64 value)
...
+        moneyLocale.set(localestr.str());
+    }
+
+    return moneyLocale.get();
+}
+
+static StringBuffer & formatMoney(StringBuffer &out, unsigned __int64 value)
+{
+    std::stringstream ss;
+    ss.imbue(std::locale(getMoneyLocale()));
```
ghalliday, line 435: better to have a static std::locale which is the value set up by getMoneyLocale to avoid the cost of recreating the locale object.

system/jlib/jstats.cpp
```diff
@@ -396,6 +400,43 @@ static StringBuffer & formatIPV4(StringBuffer & out, unsigned __int64 value)
...
 
+static const char *getMoneyLocale()
+{
+    static bool haveLocale = false;
+    static StringAttr moneyLocale;
+
+    if (!haveLocale)
+    {
+        StringBuffer localestr;
+        if (isGlobalConfigReady())
```
ghalliday, line 411: Why is this check necessary?  The global configuration should be set up by all components early in their lifetime.

system/jlib/jstats.cpp
```diff
@@ -396,6 +400,43 @@ static StringBuffer & formatIPV4(StringBuffer & out, unsigned __int64 value)
...
+    }
+
+    return moneyLocale.get();
+}
+
+static StringBuffer & formatMoney(StringBuffer &out, unsigned __int64 value)
+{
+    std::stringstream ss;
+    ss.imbue(std::locale(getMoneyLocale()));
+    ss << std::showbase << std::put_money(value/1E4);
```
ghalliday, line 436: Shouldn't this be cost_type2money(value)?

ghalliday (commented):
@shamser a few comments

system/jlib/jstats.cpp
shamser: No. put_money requires the money amount multiplied by 100.   Should I do  cost_type2money(value)*100 to make the code clearer?

ghalliday: That sounds weird!
Looking at the examples on https://en.cppreference.com/w/cpp/io/manip/put_money I think it is implying that the money amount for the us is actually cents.  (The japanese yen example is not scaled.)
More details here:  https://www.stroustrup.com/3rd_loc.pdf
So you could either document that it is in cents, or multiply by 10^ locale...?..frac_digits

shamser:

system/jlib/jstats.cpp
```diff
@@ -400,40 +399,53 @@ static StringBuffer & formatIPV4(StringBuffer & out, unsigned __int64 value)
...
-    static bool haveLocale = false;
-    static StringAttr moneyLocale;
-
-    if (!haveLocale)
+public:
+    ~MoneyLocale()
+    {
+        std::locale * tmp = nullptr;
+        locale.exchange(tmp, std::memory_order_relaxed);
+        delete tmp;
```
ghalliday, line 409: minor: "delete locale.load()" is sufficient - this object is being destroyed, so it can't be accessed from any other thread

system/jlib/jstats.cpp
```diff
@@ -396,6 +399,56 @@ static StringBuffer & formatIPV4(StringBuffer & out, unsigned __int64 value)
...
+    }
+    std::locale & queryMoneyLocale()
+    {
+        return *querySingleton(locale, cslock, [this]{ return this->createMoneyLocale(); });
+    }
+
+private:
+    static CriticalSection cslock;
+    std::atomic<std::locale *> locale {nullptr};
+} moneyLocale;
```
ghalliday, line 437: trivial: this could be a module-static variable, but barely worth changing.

ghalliday (commented):
@shamser a couple of comments, looks ready to merge withj a minor tweak

shamser:
@ghalliday

PR #14330, opened 2020-11-20
HPCC-24969 Fix sasha archive age issue

jakesmith:
The archiving mechanism checks to see if a workunit is old and
ready to be archived, based on the timestamps of recent activity,
record in the workunit.
This used to be recorded as TimeStamps/TimeStamp, but was changed
when the stats. was refactored (to Statistis/Statistic).
The sasha archiving code wasn't updated at that time.

The symptom of this bug, was that an old workunit, that was
recently active (e.g. a scheduled workunit that runs regularly),
could be archived when it should not have been.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dali/sasha/saarch.cpp
+++ b/dali/sasha/saarch.cpp
@@ -1007,11 +1007,12 @@ class CWorkUnitArchiver: public CBranchArchiver
             }
             getWorkUnitCreateTime(wuid,time);
             // get latest time stamp
-            Owned<IPropertyTreeIterator> iter = e.getElements("TimeStamps/TimeStamp/*");
+            Owned<IPropertyTreeIterator> iter = e.getElements("Statistics/Statistic");
             if (iter) {
                 ForEach(*iter) {
                     CDateTime cmp;
-                    cmp.setString(iter->query().queryProp(NULL));
+                    timestamp_type ts = iter->query().getPropInt64("@ts");
+                    cmp.setTimeStamp(ts);
                     if (time.isNull()||(!cmp.isNull()&&(cmp.compare(time)>0)))
                         time.set(cmp);
                 }
--- a/common/workunit/workunit.cpp
+++ b/common/workunit/workunit.cpp
@@ -8149,6 +8149,9 @@ void CLocalWorkUnit::setStatistic(StatisticCreatorType creatorType, const char *
 
     if (!statTree)
     {
+        /* NB: Sasha archive uses this structure directly
+         * if it changes, the code in saarch.cpp needs updating
+         */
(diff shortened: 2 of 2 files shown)


jakesmith:
@ghalliday - please review

dali/sasha/saarch.cpp
```diff
@@ -1007,11 +1007,12 @@ class CWorkUnitArchiver: public CBranchArchiver
             }
             getWorkUnitCreateTime(wuid,time);
             // get latest time stamp
-            Owned<IPropertyTreeIterator> iter = e.getElements("TimeStamps/TimeStamp/*");
+            Owned<IPropertyTreeIterator> iter = e.getElements("Statistics/Statistic");
             if (iter) {
                 ForEach(*iter) {
                     CDateTime cmp;
-                    cmp.setString(iter->query().queryProp(NULL));
+                    timestamp_type ts = iter->query().getPropInt64("@ts");
```
ghalliday, line 1014: @jakesmith This is slightly different to before.  I think before it would have been the most recent timestamp.  Now it will be the most recent stats - both timestamps and other.  Could this cause any problems?  (I can't think of any at the moment.)

dali/sasha/saarch.cpp
```diff
@@ -1433,7 +1434,7 @@ class CSashaArchiverServer: public ISashaServer, public Thread
 
     void runArchiver(CBranchArchiver &archiver,Owned<IRemoteConnection> &conn)
     {
-        if (!stopped&&archiver.ready()) {
+        if (!stopped) {//if (!stopped&&archiver.ready()) {
```
ghalliday, line 1437: restore or remove the commented out code.

ghalliday (commented):
@jakesmith a couple of comments

dali/sasha/saarch.cpp
jakesmith: I think it's ok, arguably an improvement, if a stat. outside of a TimeStamp updates, it would be better to consider it 'touched'.
That may changed conceivably, if we added stats for things like how many times the workunit had been looked at or something I suppose.

dali/sasha/saarch.cpp
jakesmith: whoops, I'll restore

jakesmith:
@ghalliday - fixed accidental release of commented out line

PR #14331, opened 2020-11-21
HPCC-25002 ESDL Transform Script enhancements

afishbeck:
1. Refactor away from ptree result handling.. allowing access to
any part of the script context.  Preserving order.  And allowing
more powerful xpath processing.
2. Refactor away from PTree based script parsing / loading.  Preserving
order.
3. Support changing the target and source locations on the fly.
4. Support builing of structured output.
5. Support calling other web services from within scripts.
6. Support dynamically tokenizing strings.  Iterating over order
accessing the result.
7. Support building structured variables.
8. Support dynamically adding namespaces to target content.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/bindings/SOAP/xpp/xpp/XmlPullParser.h
+++ b/esp/bindings/SOAP/xpp/xpp/XmlPullParser.h
@@ -174,7 +174,8 @@ namespace xpp {
     }
 
 
-    int skipSubTree()  {
+    bool skipSubTreeEx()  {
+      bool hasChildren = false;
       int level = 1;
       StartTag stag;
       int type = XmlPullParser::END_TAG;
@@ -184,13 +185,20 @@ namespace xpp {
         case XmlPullParser::START_TAG:
           readStartTag(stag);
           ++level;
+          hasChildren = true;
           break;
         case XmlPullParser::END_TAG:
           --level;
           break;
         }
       }
-      return type;
+      return hasChildren;
+    }
+
+    //backward compatability
+    int skipSubTree() {
+      skipSubTreeEx();
+      return XmlPullParser::END_TAG;
     }
 
     const SXT_STRING getPosDesc() const {
--- a/esp/esdllib/esdl_script.cpp
+++ b/esp/esdllib/esdl_script.cpp
@@ -18,15 +18,31 @@
 #include "espcontext.hpp"
 #include "esdl_script.hpp"
 #include "wsexcept.hpp"
+#include "httpclient.hpp"
+
+#include <xpp/XmlPullParser.h>
+using namespace xpp;
 
 interface IEsdlTransformOperation : public IInterface
 {
-    virtual const char *queryMergedTarget() = 0;
-    virtual bool process(IEsdlScriptContext * context, IPropertyTree *content, IXpathContext * xpathContext) = 0;
+    virtual bool process(IEsdlScriptContext * scriptContext, IXpathContext * targetContext, IXpathContext * sourceContext) = 0;
     virtual void toDBGLog() = 0;
 };
 
-IEsdlTransformOperation *createEsdlTransformOperation(IPropertyTree *element, const StringBuffer &prefix);
+IEsdlTransformOperation *createEsdlTransformOperation(XmlPullParser &xpp, const StringBuffer &prefix, bool withVariables, bool ignoreCodingErrors);
+void createEsdlTransformOperations(IArrayOf<IEsdlTransformOperation> &operations, XmlPullParser &xpp, const StringBuffer &prefix, bool withVariables, bool ignoreCodingErrors);
+void createEsdlTransformChooseOperations(IArrayOf<IEsdlTransformOperation> &operations, XmlPullParser &xpp, const StringBuffer &prefix, bool withVariables, bool ignoreCodingErrors);
+typedef void (*esdlOperationsFactory_t)(IArrayOf<IEsdlTransformOperation> &operations, XmlPullParser &xpp, const StringBuffer &prefix, bool withVariables, bool ignoreCodingErrors);
+
+bool getStartTagValueBool(StartTag &stag, const char *name, bool defaultValue)
+{
+    if (isEmptyString(name))
+        return defaultValue;
+    const char *value = stag.getValue(name);
+    if (isEmptyString(value))
+        return defaultValue;
+    return strToBool(value);
+}
 
 inline void esdlOperationError(int code, const char *op, const char *msg, const char *traceName, bool exception)
 {
@@ -69,44 +85,89 @@ static inline StringBuffer &makeOperationTagName(StringBuffer &s, const StringBu
 class CEsdlTransformOperationBase : public CInterfaceOf<IEsdlTransformOperation>
 {
 protected:
-    StringAttr m_mergedTarget;
     StringAttr m_tagname;
     StringAttr m_traceName;
     bool m_ignoreCodingErrors = false; //ideally used only for debugging
 
 public:
-    CEsdlTransformOperationBase(IPropertyTree *tree, const StringBuffer &prefix)
+    CEsdlTransformOperationBase(XmlPullParser &xpp, StartTag &stag, const StringBuffer &prefix)
     {
-        m_tagname.set(tree->queryName());
-        if (tree->hasProp("@trace"))
-            m_traceName.set(tree->queryProp("@trace"));
-        if (tree->hasProp("@_crtTarget"))
-            m_mergedTarget.set(tree->queryProp("@_crtTarget"));
-        m_ignoreCodingErrors = tree->getPropBool("@optional", false);
-    }
-
-    virtual const char *queryMergedTarget() override
-    {
-        return m_mergedTarget;
+        m_tagname.set(stag.getLocalName());
+        m_traceName.set(stag.getValue("trace"));
+        m_ignoreCodingErrors = getStartTagValueBool(stag, "optional", false);
     }
 };
 
+class CEsdlTransformOperationWithChildren : public CEsdlTransformOperationBase
+{
+protected:
+    IArrayOf<IEsdlTransformOperation> m_children;
+    bool m_withVariables = false;
 
-class CEsdlTransformOperationVariable : public CEsdlTransformOperationBase
+public:
+    CEsdlTransformOperationWithChildren(XmlPullParser &xpp, StartTag &stag, const StringBuffer &prefix, bool withVariables, esdlOperationsFactory_t factory) : CEsdlTransformOperationBase(xpp, stag, prefix), m_withVariables(withVariables)
+    {
+        //load children
+        if (factory)
+            factory(m_children, xpp, prefix, withVariables, m_ignoreCodingErrors);
+        else
+            createEsdlTransformOperations(m_children, xpp, prefix, withVariables, m_ignoreCodingErrors);
+    }
+
+    virtual ~CEsdlTransformOperationWithChildren(){}
+
+    virtual bool processChildren(IEsdlScriptContext * scriptContext, IXpathContext * targetContext, IXpathContext * sourceContext)
+    {
+        if (!m_children.length())
+            return false;
+
+        Owned<CXpathContextScope> scope = m_withVariables ? new CXpathContextScope(sourceContext, m_tagname) : nullptr;
+        bool ret = false;
+        ForEachItemIn(i, m_children)
+        {
+            if (m_children.item(i).process(scriptContext, targetContext, sourceContext))
+                ret = true;
+        }
+        return ret;
+    }
+
+    virtual void toDBGLog () override
+    {
+    #if defined(_DEBUG)
+        ForEachItemIn(i, m_children)
+            m_children.item(i).toDBGLog();
+    #endif
+    }
+};
+
+class CEsdlTransformOperationWithoutChildren : public CEsdlTransformOperationBase
+{
+public:
+    CEsdlTransformOperationWithoutChildren(XmlPullParser &xpp, StartTag &stag, const StringBuffer &prefix) : CEsdlTransformOperationBase(xpp, stag, prefix)
+    {
+        if (xpp.skipSubTreeEx())
+            esdlOperationError(ESDL_SCRIPT_Error, m_tagname, "should not have child tags", m_traceName, !m_ignoreCodingErrors);
+    }
+
+    virtual ~CEsdlTransformOperationWithoutChildren(){}
+
+};
+
+class CEsdlTransformOperationVariable : public CEsdlTransformOperationWithChildren
 {
 protected:
     StringAttr m_name;
     Owned<ICompiledXpath> m_select;
 
 public:
-    CEsdlTransformOperationVariable(IPropertyTree *tree, const StringBuffer &prefix) : CEsdlTransformOperationBase(tree, prefix)
+    CEsdlTransformOperationVariable(XmlPullParser &xpp, StartTag &stag, const StringBuffer &prefix) : CEsdlTransformOperationWithChildren(xpp, stag, prefix, true, nullptr)
     {
         if (m_traceName.isEmpty())
-            m_traceName.set(tree->queryProp("@name"));
-        m_name.set(tree->queryProp("@name"));
+            m_traceName.set(stag.getValue("name"));
+        m_name.set(stag.getValue("name"));
         if (m_name.isEmpty())
             esdlOperationError(ESDL_SCRIPT_MissingOperationAttr, m_tagname, "without name", m_traceName, !m_ignoreCodingErrors);
-        const char *select = tree->queryProp("@select");
+        const char *select = stag.getValue("select");
         if (!isEmptyString(select))
             m_select.setown(compileXpath(select));
     }
@@ -115,11 +176,22 @@ public:
     {
     }
 
-    virtual bool process(IEsdlScriptContext * context, IPropertyTree *content, IXpathContext * xpathContext) override
+    virtual bool process(IEsdlScriptContext * scriptContext, IXpathContext * targetContext, IXpathContext * sourceContext) override
     {
         if (m_select)
-            return xpathContext->addCompiledVariable(m_name, m_select);
-        return xpathContext->addVariable(m_name, "");
+            return sourceContext->addCompiledVariable(m_name, m_select);
+        else if (m_children.length())
+        {
+            VStringBuffer xpath("/esdl_script_context/temporaries/%s", m_name.str());
+            CXpathContextLocation location(targetContext);
+
+            targetContext->ensureLocation(xpath, true);
+            processChildren(scriptContext, targetContext, sourceContext); //bulid nodeset
+            sourceContext->addXpathVariable(m_name, xpath);
+            return false;
+        }
+        sourceContext->addVariable(m_name, "");
+        return false;
     }
 
     virtual void toDBGLog() override
@@ -130,10 +202,290 @@ public:
     }
 };
 
+class CEsdlTransformOperationHttpContentXml : public CEsdlTransformOperationWithChildren
+{
+
+public:
+    CEsdlTransformOperationHttpContentXml(XmlPullParser &xpp, StartTag &stag, const StringBuffer &prefix) : CEsdlTransformOperationWithChildren(xpp, stag, prefix, true, nullptr)
+    {
+    }
+
+    virtual ~CEsdlTransformOperationHttpContentXml(){}
+
+    bool process(IEsdlScriptContext * scriptContext, IXpathContext * targetContext, IXpathContext * sourceContext) override
+    {
+        CXpathContextLocation location(targetContext);
+        targetContext->addElementToLocation("content");
+        return processChildren(scriptContext, targetContext, sourceContext);
+    }
+
+    virtual void toDBGLog () override
+    {
+    #if defined(_DEBUG)
+        DBGLOG (">>>>>>>>>>> %s >>>>>>>>>>", m_tagname.str());
+        CEsdlTransformOperationWithChildren::toDBGLog();
+        DBGLOG (">>>>>>>>>>> %s >>>>>>>>>>", m_tagname.str());
+    #endif
+    }
+};
+
+interface IEsdlTransformOperationHttpHeader : public IInterface
+{
+    virtual bool processHeader(IEsdlScriptContext * scriptContext, IXpathContext * targetContext, IXpathContext * sourceContext, IProperties *headers) = 0;
+};
+
+
+class CEsdlTransformOperationHttpHeader : public CEsdlTransformOperationWithoutChildren, implements IEsdlTransformOperationHttpHeader
+{
+protected:
+    StringAttr m_name;
+    Owned<ICompiledXpath> m_xpath_name;
+    Owned<ICompiledXpath> m_value;
+
+public:
+    IMPLEMENT_IINTERFACE_USING(CEsdlTransformOperationWithoutChildren)
+
+    CEsdlTransformOperationHttpHeader(XmlPullParser &xpp, StartTag &stag, const StringBuffer &prefix) : CEsdlTransformOperationWithoutChildren(xpp, stag, prefix)
+    {
+        m_name.set(stag.getValue("name"));
+        const char *xpath_name = stag.getValue("xpath_name");
+        if (m_name.isEmpty() && isEmptyString(xpath_name))
+            esdlOperationError(ESDL_SCRIPT_MissingOperationAttr, m_tagname, "without name or xpath_name", m_traceName, !m_ignoreCodingErrors);
+        if (!isEmptyString(xpath_name))
+            m_xpath_name.setown(compileXpath(xpath_name));
+
+        const char *value = stag.getValue("value");
+        if (isEmptyString(value))
+            esdlOperationError(ESDL_SCRIPT_MissingOperationAttr, m_tagname, "without value", m_traceName, !m_ignoreCodingErrors);
+        m_value.setown(compileXpath(value));
+    }
+
+    virtual ~CEsdlTransformOperationHttpHeader(){}
+
+    bool process(IEsdlScriptContext * scriptContext, IXpathContext * targetContext, IXpathContext * sourceContext) override
+    {
+        return processHeader(scriptContext, targetContext, sourceContext, nullptr);
+    }
+
+    bool processHeader(IEsdlScriptContext * scriptContext, IXpathContext * targetContext, IXpathContext * sourceContext, IProperties *headers) override
+    {
+        CXpathContextLocation location(targetContext);
+        targetContext->addElementToLocation("header");
+        StringBuffer name;
+        if (m_xpath_name)
+            sourceContext->evaluateAsString(m_xpath_name, name);
+        else
+            name.set(m_name);
+
+        StringBuffer value;
+        if (m_value)
+            sourceContext->evaluateAsString(m_value, value);
+        if (name.length() && value.length())
+        {
+            if (headers)
+                headers->setProp(name, value);
+            targetContext->ensureSetValue("@name", name, true);
+            targetContext->ensureSetValue("@value", value, true);
+        }
+        return false;
+    }
+
+    virtual void toDBGLog () override
+    {
+    #if defined(_DEBUG)
+        DBGLOG ("> %s (%s, value(%s)) >>>>>>>>>>", m_tagname.str(), m_xpath_name ? m_xpath_name->getXpath() : m_name.str(), m_value ? m_value->getXpath() : "");
+    #endif
+    }
+};
+
+
+
+class CEsdlTransformOperationHttpPostXml : public CEsdlTransformOperationBase
+{
+protected:
+    StringAttr m_name;
+    StringAttr m_section;
+    Owned<ICompiledXpath> m_url;
+    IArrayOf<IEsdlTransformOperationHttpHeader> m_headers;
+    Owned<IEsdlTransformOperation> m_content;
+
+public:
+    CEsdlTransformOperationHttpPostXml(XmlPullParser &xpp, StartTag &stag, const StringBuffer &prefix) : CEsdlTransformOperationBase(xpp, stag, prefix)
+    {
+        m_name.set(stag.getValue("name"));
+        if (m_traceName.isEmpty())
+            m_traceName.set(m_name.str());
+        m_section.set(stag.getValue("section"));
+        if (m_section.isEmpty())
+            m_section.set("temporaries");
+        if (m_name.isEmpty())
+            esdlOperationError(ESDL_SCRIPT_MissingOperationAttr, m_tagname, "without name", m_traceName, !m_ignoreCodingErrors);
+        const char *url = stag.getValue("url");
+        if (isEmptyString(url))
+            esdlOperationError(ESDL_SCRIPT_MissingOperationAttr, m_tagname, "without url", m_traceName, !m_ignoreCodingErrors);
+        m_url.setown(compileXpath(url));
+
+        int type = 0;
+        while((type = xpp.next()) != XmlPullParser::END_DOCUMENT)
+        {
+            switch(type)
+            {
+                case XmlPullParser::START_TAG:
+                {
+                    StartTag stag;
+                    xpp.readStartTag(stag);
+                    const char *op = stag.getLocalName();
+                    if (isEmptyString(op))
+                        esdlOperationError(ESDL_SCRIPT_Error, m_tagname, "unknown error", m_traceName, !m_ignoreCodingErrors);
+                    if (streq(op, "http-header"))
+                        m_headers.append(*new CEsdlTransformOperationHttpHeader(xpp, stag, prefix));
+                    else if (streq(op, "content"))
+                        m_content.setown(new CEsdlTransformOperationHttpContentXml(xpp, stag, prefix));
+                    else
+                        xpp.skipSubTreeEx();
+                    break;
+                }
+                case XmlPullParser::END_TAG:
+                case XmlPullParser::END_DOCUMENT:
+                    return;
+            }
+        }
+    }
+
+    virtual ~CEsdlTransformOperationHttpPostXml()
+    {
+    }
+
+    void buildHeaders(IEsdlScriptContext * scriptContext, IXpathContext * targetContext, IXpathContext * sourceContext, IProperties *headers)
+    {
+        if (!m_headers.length())
+            return;
+        CXpathContextLocation location(targetContext);
+        targetContext->addElementToLocation("headers");
+        ForEachItemIn(i, m_headers)
+            m_headers.item(i).processHeader(scriptContext, targetContext, sourceContext, headers);
+    }
+
+    void buildRequest(IEsdlScriptContext * scriptContext, IXpathContext * targetContext, IXpathContext * sourceContext, const char *url, IProperties *headers)
+    {
+        CXpathContextLocation location(targetContext);
+        targetContext->addElementToLocation("request");
+        targetContext->ensureSetValue("@url", url, true);
+        buildHeaders(scriptContext, targetContext, sourceContext, headers);
+        if (m_content)
+            m_content->process(scriptContext, targetContext, sourceContext);
+    }
+
+    void postRequest(IEsdlScriptContext * scriptContext, IXpathContext * targetContext, IXpathContext * sourceContext, const char *url, IProperties *headers)
+    {
+        VStringBuffer xpath("/esdl_script_context/%s/%s/request/content/*[1]", m_section.str(), m_name.str());
+
+        StringBuffer content;
+        sourceContext->toXml(xpath, content);
+        if (!content)
+            return;
+        CXpathContextLocation location(targetContext);
+        targetContext->addElementToLocation("response");
+
+        Owned<IHttpClientContext> httpCtx = getHttpClientContext();
+        Owned<IHttpClient> httpclient = httpCtx->createHttpClient(NULL, url);
+        if (!httpclient)
+            return;
+
+        try
+        {
+            StringBuffer status;
+            StringBuffer response;
+            StringBuffer errmsg;
+
+            if (headers && !headers->hasProp("Accept"))
+                headers->setProp("Accept", "text/html, application/xml");
+
+            HttpClientErrCode code = HttpClientErrCode::OK;
+            Owned<IHttpMessage> resp = httpclient->sendRequestEx("POST", "text/xml", content, code, errmsg, headers);
+            targetContext->ensureSetValue("@status", status.str(), true);
+
+            StringBuffer err;
+            err.append((int) code);
+            targetContext->ensureSetValue("@error-code", err.str(), true);
+            if (code != HttpClientErrCode::OK)
+                throw MakeStringException(ESDL_SCRIPT_Error, "ESDL Script error sending request in http-post-xml %s url(%s)", m_traceName.str(), url);
+
+            resp->getStatus(status);
+            targetContext->ensureSetValue("@status", status.str(), true);
+
+            resp->getContent(response);
+            if (!response.trim().length())
+                throw MakeStringException(ESDL_SCRIPT_Error, "ESDL Script empty result calling http-post-xml %s url(%s)", m_traceName.str(), url);
+
+            StringBuffer contentType;
+            resp->getContentType(contentType);
+            targetContext->ensureSetValue("@content-type", contentType.str(), true);
+            if (strnicmp("text/xml", contentType.str(), 8)==0 || strnicmp("application/xml", contentType.str(), 15) ==0)
+            {
+                CXpathContextLocation content_location(targetContext);
+                targetContext->addElementToLocation("content");
+                targetContext->addXmlContent(response.str());
+            }
+            else
+            {
+                targetContext->ensureSetValue("@status", "error", true);
+                targetContext->ensureSetValue("text", response.str(), true);
+            }
+        }
+        catch(IMultiException *me)
+        {
+            StringBuffer xml;
+            me->serialize(xml);
+            CXpathContextLocation content_location(targetContext);
+            targetContext->ensureSetValue("@status", "error", true);
+            targetContext->addElementToLocation("content");
+            targetContext->addXmlContent(xml.str());
+            me->Release();
+        }
+        catch(IException *E)
+        {
+            StringBuffer xml;
+            Owned<IMultiException> me = makeMultiException("ESDLScript");
+            me->append(*LINK(E));
+            me->serialize(xml);
+            CXpathContextLocation content_location(targetContext);
+            targetContext->ensureSetValue("@status", "error", true);
+            targetContext->addElementToLocation("content");
+            targetContext->addXmlContent(xml.str());
+            E->Release();
+        }
+    }
+
+    virtual bool process(IEsdlScriptContext * scriptContext, IXpathContext * targetContext, IXpathContext * sourceContext) override
+    {
+        VStringBuffer xpath("/esdl_script_context/%s/%s", m_section.str(), m_name.str());
+        CXpathContextLocation location(targetContext);
+        targetContext->ensureLocation(xpath, true);
+        StringBuffer url;
+        if (m_url)
+            sourceContext->evaluateAsString(m_url, url);
+
+        Owned<IProperties> headers = createProperties();
+        buildRequest(scriptContext, targetContext, sourceContext, url, headers);
+        postRequest(scriptContext, targetContext, sourceContext, url, headers);
+
+        sourceContext->addXpathVariable(m_name, xpath);
+        return false;
+    }
+
+    virtual void toDBGLog() override
+    {
+#if defined(_DEBUG)
+        DBGLOG(">%s> %s with name(%s) url(%s)", m_name.str(), m_tagname.str(), m_name.str(), m_url ? m_url->getXpath() : "url error");
+#endif
+    }
+};
+
 class CEsdlTransformOperationParameter : public CEsdlTransformOperationVariable
 {
 public:
-    CEsdlTransformOperationParameter(IPropertyTree *tree, const StringBuffer &prefix) : CEsdlTransformOperationVariable(tree, prefix)
+    CEsdlTransformOperationParameter(XmlPullParser &xpp, StartTag &stag, const StringBuffer &prefix) : CEsdlTransformOperationVariable(xpp, stag, prefix)
     {
     }
 
@@ -141,15 +493,15 @@ public:
     {
     }
 
-    virtual bool process(IEsdlScriptContext * context, IPropertyTree *content, IXpathContext * xpathContext) override
+    virtual bool process(IEsdlScriptContext * scriptContext, IXpathContext * targetContext, IXpathContext * sourceContext) override
     {
         if (m_select)
-            return xpathContext->declareCompiledParameter(m_name, m_select);
-        return xpathContext->declareParameter(m_name, "");
+            return sourceContext->declareCompiledParameter(m_name, m_select);
+        return sourceContext->declareParameter(m_name, "");
     }
 };
 
-class CEsdlTransformOperationSetSectionAttributeBase : public CEsdlTransformOperationBase
+class CEsdlTransformOperationSetSectionAttributeBase : public CEsdlTransformOperationWithoutChildren
 {
 protected:
     StringAttr m_name;
@@ -157,24 +509,24 @@ protected:
     Owned<ICompiledXpath> m_select;
 
 public:
-    CEsdlTransformOperationSetSectionAttributeBase(IPropertyTree *tree, const StringBuffer &prefix, const char *attrName) : CEsdlTransformOperationBase(tree, prefix)
+    CEsdlTransformOperationSetSectionAttributeBase(XmlPullParser &xpp, StartTag &stag, const StringBuffer &prefix, const char *attrName) : CEsdlTransformOperationWithoutChildren(xpp, stag, prefix)
     {
         if (m_traceName.isEmpty())
-            m_traceName.set(tree->queryProp("@name"));
+            m_traceName.set(stag.getValue("name"));
         if (!isEmptyString(attrName))
             m_name.set(attrName);
         else
         {
-            m_name.set(tree->queryProp("@name"));
+            m_name.set(stag.getValue("name"));
 
-            const char *xpath_name = tree->queryProp("@xpath_name");
+            const char *xpath_name = stag.getValue("xpath_name");
             if (!isEmptyString(xpath_name))
                 m_xpath_name.setown(compileXpath(xpath_name));
             else if (m_name.isEmpty())
                 esdlOperationError(ESDL_SCRIPT_MissingOperationAttr, m_tagname, "without name", m_traceName, !m_ignoreCodingErrors); //don't mention value, it's deprecated
         }
 
-        const char *select = tree->queryProp("@select");
+        const char *select = stag.getValue("select");
         if (isEmptyString(select))
             esdlOperationError(ESDL_SCRIPT_MissingOperationAttr, m_tagname, "without select", m_traceName, !m_ignoreCodingErrors); //don't mention value, it's deprecated
         m_select.setown(compileXpath(select));
@@ -191,7 +543,7 @@ public:
 
     virtual const char *getSectionName() = 0;
 
-    virtual bool process(IEsdlScriptContext * context, IPropertyTree *content, IXpathContext * xpathContext) override
+    virtual bool process(IEsdlScriptContext * scriptContext, IXpathContext * targetContext, IXpathContext * sourceContext) override
     {
         if ((!m_name && !m_xpath_name) || !m_select)
             return false; //only here if "optional" backward compatible support for now (optional syntax errors aren't actually helpful)
@@ -199,13 +551,13 @@ public:
         {
             StringBuffer name;
             if (m_xpath_name)
-                xpathContext->evaluateAsString(m_xpath_name, name);
+                sourceContext->evaluateAsString(m_xpath_name, name);
             else
                 name.set(m_name);
 
             StringBuffer value;
-            xpathContext->evaluateAsString(m_select, value);
-            context->setAttribute(getSectionName(), name, value);
+            sourceContext->evaluateAsString(m_select, value);
+            scriptContext->setAttribute(getSectionName(), name, value);
         }
         catch (IException* e)
         {
@@ -226,7 +578,7 @@ public:
 class CEsdlTransformOperationStoreValue : public CEsdlTransformOperationSetSectionAttributeBase
 {
 public:
-    CEsdlTransformOperationStoreValue(IPropertyTree *tree, const StringBuffer &prefix) : CEsdlTransformOperationSetSectionAttributeBase(tree, prefix, nullptr)
+    CEsdlTransformOperationStoreValue(XmlPullParser &xpp, StartTag &stag, const StringBuffer &prefix) : CEsdlTransformOperationSetSectionAttributeBase(xpp, stag, prefix, nullptr)
     {
     }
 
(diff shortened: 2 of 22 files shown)


afishbeck:
@asselitx @timothyklemm please test and review.  Leaving as a draft for now, until it's functionality has been tested by someone besides me.

afishbeck:
Examples of usage can be found in "testing/unittests/esdltests.cpp" and in "initfiles/examples/EsdlExample/esdl_binding_entrypoints.xml".

Notice that besides examples of functionality esdl_binding_entrypoints.xml shows how order is now preserved if the script content is added as CDATA.

testing/unittests/esdltests.cpp
```diff
@@ -1527,9 +1551,540 @@ constexpr const char * result = R"!!(<soap:Envelope xmlns:soap="http://schemas.x
...
+                  <ID>GI101</ID>
+                  <ID>GI102</ID>
+               </FullName>
+            </Person>
+          </root>
+        )!!";
+
+        static constexpr const char * script = R"!!(<es:CustomRequestTransform xmlns:es="urn:hpcc:esdl:script" target="Person">
+            <es:variable name='value' select="'abc'"/>
+            <es:http-post-xml url="'http://127.0.0.1:9876'" section="logging" name="roxiestuff">
```
timothyklemm, line 1724: Is it valid to assume this service is published on the localhost every time the test is run?

testing/unittests/esdltests.cpp
```diff
@@ -1527,9 +1551,540 @@ constexpr const char * result = R"!!(<soap:Envelope xmlns:soap="http://schemas.x
...
+      Owned<IEspContext> ctx = createEspContext(nullptr);
+      Owned<IEsdlScriptContext> scriptContext = createTestScriptContext(ctx, input, config1);
+
+      bool legacy = false;
+      Owned<IEsdlTransformMethodMap> map = createEsdlTransformMethodMap();
+      map->addMethodTransforms("", serverScripts, legacy);
+      map->addMethodTransforms("mymethod", methodScripts, legacy);
+
+      IEsdlTransformSet *serviceSet = map->queryMethodEntryPoint("", "BackendResponse");
+      IEsdlTransformSet *methodSet = map->queryMethodEntryPoint("mymethod", "BackendResponse");
```
timothyklemm, line 2074: Just wondering if it would make sense to combine the service and method collections internally (with service transforms shared instead of duplicated) between methods? It could streamline lookup and use of transform sets by decoupling that code from knowledge about how they're defined.

esp/esdllib/esdl_script.cpp
```diff
@@ -947,74 +1490,44 @@ class CEsdlCustomTransform : public CInterfaceOf<IEsdlCustomTransform>
...
-                    m_prefix.append(':');
-                }
-            }
-        }
+        const char *tag = stag.getLocalName();
 
-        m_name.set(tree.queryProp("@name"));
-        m_target.set(tree.queryProp("@target"));
+        m_name.set(stag.getValue("name"));
+        m_target.set(stag.getValue("target"));
```
timothyklemm, line 1499: Since we can use @target as a shortcut in place of a target element, and we have to for legacy, does it make sense to provide a corresponding shortcut for the source document?

asselitx:
While running tests on the code, I see some possibly undesirable behavior that might be new. @timothyklemm I seem to recall you may have made a change to prevent this from happening. For context I'm testing a simple example with one method-level transform.

When sending the request to the back end (Roxie) the transform script is included at the location Context/Row/Common/ESP/Config/Method/Scripts

esp/esdllib/esdl_script.cpp
```diff
@@ -18,15 +18,32 @@
...
+#include "httpclient.hpp"
+
+#include <xpp/XmlPullParser.h>
+using namespace xpp;
 
 interface IEsdlTransformOperation : public IInterface
 {
     virtual const char *queryMergedTarget() = 0;
-    virtual bool process(IEsdlScriptContext * context, IPropertyTree *content, IXpathContext * xpathContext) = 0;
+    virtual bool process(IEsdlScriptContext * scriptContext, IXpathContext * targetContext, IXpathContext * xpathContext) = 0;
```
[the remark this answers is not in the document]
afishbeck: Definitely agree.  I'll see if I can get it changed this week.  If not I don't want to hold this PR up and I'll open a JIRA as something to clean up later.

esp/esdllib/esdl_script.cpp
```diff
@@ -75,14 +92,12 @@ class CEsdlTransformOperationBase : public CInterfaceOf<IEsdlTransformOperation>
...
     {
-        m_tagname.set(tree->queryName());
-        if (tree->hasProp("@trace"))
-            m_traceName.set(tree->queryProp("@trace"));
-        if (tree->hasProp("@_crtTarget"))
-            m_mergedTarget.set(tree->queryProp("@_crtTarget"));
-        m_ignoreCodingErrors = tree->getPropBool("@optional", false);
+        m_tagname.set(stag.getLocalName());
+        m_traceName.set(stag.getValue("trace"));
+        m_mergedTarget.set(stag.getValue("_crtTarget"));
```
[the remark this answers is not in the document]
afishbeck: True, I'll remove it.

esp/esdllib/esdl_script.cpp
```diff
@@ -91,22 +106,75 @@ class CEsdlTransformOperationBase : public CInterfaceOf<IEsdlTransformOperation>
...
+    #endif
+    }
+};
+
+class CEsdlTransformOperationWithoutChildren : public CEsdlTransformOperationBase
+{
+public:
+    CEsdlTransformOperationWithoutChildren(XmlPullParser &xpp, StartTag &stag, const StringBuffer &prefix) : CEsdlTransformOperationBase(xpp, stag, prefix)
+    {
+        xpp.skipSubTree();
```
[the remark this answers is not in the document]
afishbeck: I'll consider errors or exceptions.  Agree it's better to prevent them for now.

system/xmllib/libxml_xpathprocessor.cpp
```diff
@@ -372,6 +392,374 @@ class CLibXpathContext : public CInterfaceOf<IXpathContext>
...
+        xmlXPathObjectPtr obj = xmlXPathEval((const xmlChar*) xpath, m_xpathContext);
+        if (!obj || obj->type!=xmlXPathObjectType::XPATH_NODESET || !obj->nodesetval || obj->nodesetval->nodeNr==0)
+            return;
+        if (obj->nodesetval->nodeNr>1 && !all)
+        {
+            xmlXPathFreeObject(obj);
+            throw MakeStringException(XPATHERR_InvalidState,"XpathContext:remove: ambiguous xpath xpath '%s' to remove all set 'all'", xpath);
+        }
+        xmlNodePtr *nodes = obj->nodesetval->nodeTab;
+        for (int i = 0; i < obj->nodesetval->nodeNr; i++)
```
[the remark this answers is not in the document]
afishbeck: Good catch.  I think your right, but I'll verify to make sure.

testing/unittests/esdltests.cpp
afishbeck: I don't currently include this test in the run.  It's there for Terrence and I, and to be run explicitly for test when set up correctly.  I'd like to get to the point where we have a way to run this type of test.

testing/unittests/esdltests.cpp
afishbeck: Yes it probably would but I'd have to think about the design.  I doubt I'll get that into this PR.  I may open a JIRA.

esp/esdllib/esdl_script.cpp
afishbeck: Yes. I had thought about that, and it's a good idea.  Not sure if it will make it in this PR or not, but if not I'll open a JIRA.

afishbeck:
@asselitx I'm not a huge fan of how that information gets added to the Roxie request.  I think it should be more explicit.. like add the gateway info and maybe something else.  Then if there is more needed we can do it in script.  Fixing it completely is beyond the scope of this PR.

BUT I thought I removed the script stuff explicitly and may have just missed a scenario.  Can you provide an example of the xpath into that content all the way to the start of the transforms.  I probably just need the tag names and the xpath will provide that.  I fixed it for the xpath I was using but perhaps I missed something or there is an xpath difference.

timothyklemm:
@afishbeck, @asselitx I think the inclusion of the Method element in the request predates the inclusion of transforms within it. I did find documentation referring to this element as a location where the roxie could look for information like gateways. It's not where our internal roxies look, but there is a precedent for inclusion. Adding **contextRemove="."** to each Method element is the solution to exclude, with other xpath options available for more granular control.

It might be worth defining a service-level value that eliminates the need to add this to every method. I'm not sure of a good way to reverse the logic and make people that actually want to see the content request it.

asselitx:
Yes @timothyklemm you are right. I tested with the contextRemove="." attribute and it eliminates that portion of the binding config from the context sent. For our purposes there is no need to make any immediate changes.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -919,15 +964,29 @@ void EsdlServiceImpl::handleServiceRequest(IEspContext &context,
...
-bool EsdlServiceImpl::handleResultLogging(IEspContext &espcontext, IEsdlScriptContext *scriptContext, IPropertyTree * reqcontext, IPropertyTree * request,const char *rawreq, const char * rawresp, const char * finalresp, const char * logdata)
+bool EsdlServiceImpl::handleResultLogging(IEspContext &espcontext, IEsdlScriptContext *scriptContext, IEsdlDefService &srvdef, IEsdlDefMethod &mthdef, IPropertyTree * reqcontext, IPropertyTree * request,const char *rawreq, const char * rawresp, const char * finalresp, const char * logdata)
 {
+    StringBuffer temp;
+    if (scriptContext)
+    {
+        IEsdlTransformSet *servicePLTs = m_transforms->queryMethodEntryPoint("", ESDLScriptEntryPoint_PreLogging);
+        IEsdlTransformSet *methodPLTs = m_transforms->queryMethodEntryPoint(mthdef.queryName(), ESDLScriptEntryPoint_PreLogging);
+
+        scriptContext->appendContent(ESDLScriptCtxSection_LogData, "LogDataset", logdata);
```
asselitx, line 978: I think the second parameter may be misspelled. Current scripts expect the element to be named "LogDatasets". Although it is redundant the current structure is UpdateLogRequest/LogDatasets/LogDatasets/Dataset[@name="..."]

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -919,15 +964,29 @@ void EsdlServiceImpl::handleServiceRequest(IEspContext &context,
...
+
+        if (servicePLTs || methodPLTs)
+        {
+            processServiceAndMethodTransforms(scriptContext, {servicePLTs, methodPLTs}, ESDLScriptCtxSection_LogData, nullptr);
+            scriptContext->toXML(temp, ESDLScriptCtxSection_PreESDLResponse);
+            logdata = temp.str();
```
asselitx, line 984: @afishbeck I don't understand what this should accomplish. PreESDLResponse does not appear to be populated after processing the transforms. It results in logdata=nullptr, and then the LogDatasets element is not available to the LogAgents because it is set on line 1000 below.

Did you mean to serialize `ESDLScriptCtxSection_LogData` here? If so then the LogDatasets will be passed to the LogAgents correctly if we also make the spelling change on line 978.

If we intend to pass the contents of PreESDLResponse to the agents as the LogDatasets, then we'll need to concatenate the LogData section also, or we'll need to see why the PreESDLResponse is not populated in this case.

esp/esdllib/esdl_script.cpp
```diff
@@ -115,10 +183,19 @@ class CEsdlTransformOperationVariable : public CEsdlTransformOperationBase
...
             return xpathContext->addCompiledVariable(m_name, m_select);
+        else if (m_children.length())
+        {
+            VStringBuffer xpath("/esdl_script_context/temporaries/%s", m_name.str());
+            CXpathContextLocation location(targetContext);
+
+            targetContext->ensureLocation(xpath, true);
+            processChildren(scriptContext, targetContext, xpathContext); //bulid nodeset
+            xpathContext->addXpathVariable(m_name, xpath);
+        }
```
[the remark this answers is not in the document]
afishbeck: Oops forgot the return above.   Not meant to fall through.

esp/services/esdl_svc_engine/esdl_binding.cpp
afishbeck: Agreed.  Will change.

esp/services/esdl_svc_engine/esdl_binding.cpp
afishbeck: Yes, typo should be ESDLScriptCtxSection_LogData section being serialized.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -1454,18 +1529,22 @@ void EsdlServiceImpl::prepareFinalRequest(IEspContext &context,
...
+        IEsdlTransformSet *serviceRTs = m_transforms->queryMethodEntryPoint("", ESDLScriptEntryPoint_BackendRequest);
+        IEsdlTransformSet *methodRTs = m_transforms->queryMethodEntryPoint(mthName, ESDLScriptEntryPoint_BackendRequest);
 
         context.addTraceSummaryTimeStamp(LogNormal, "srt-custreqtrans");
         scriptContext->setContent(ESDLScriptCtxSection_ESDLRequest, reqProcessed.str());
-        if (m_serviceLevelRequestTransform || methodCrt) //slightly redundant, but less fragile to future changes to check again here
-            processServiceAndMethodTransforms(scriptContext, {m_serviceLevelRequestTransform, methodCrt}, ESDLScriptCtxSection_ESDLRequest, ESDLScriptCtxSection_FinalRequest);
+        if (serviceCRTs || serviceRTs || methodCRTs || methodRTs)
+            processServiceAndMethodTransforms(scriptContext, {serviceCRTs, serviceRTs, methodCRTs, methodRTs}, ESDLScriptCtxSection_ESDLRequest, ESDLScriptCtxSection_FinalRequest);
         scriptContext->toXML(reqProcessed.clear(), ESDLScriptCtxSection_FinalRequest);
```
asselitx, line 1548: I think this should be part of the `if (serviceCRTs || serviceRTs || methodCRTs || methodRTs)` clause. Otherwise when there are no transforms (Legacy or BackendRequest) we will clear out the reqProcessed which is just about to be sent to the back end.

testing/unittests/esdltests.cpp
```diff
@@ -1008,7 +1032,7 @@ constexpr const char * result = R"!!(<soap:Envelope xmlns:soap="http://schemas.x
     <Param name='testcase' value="for each friend"/>
     <Param name='section' select="'Friends'"/>
     <Param name='garbage' select="''"/>
-    <Param name='ForBuildListPath' select='/esdl_script_context/ESDLRequest/root/extra/Friends/Name'/><!--absolute path may change, highly frowned upon-->
+    <Param name='ForBuildListPath' select='/esdl_script_context/esdl_request/root/extra/Friends/Name'/>
```
asselitx, line 1035: testEsdlTransformAbsoluteSoapPath above will also need this change ESDLRequest -> esdl_request

esp/esdllib/esdl_script.cpp
```diff
@@ -130,51 +207,312 @@ class CEsdlTransformOperationVariable : public CEsdlTransformOperationBase
...
+    Owned<IEsdlTransformOperation> m_content;
+
+public:
+    CEsdlTransformOperationHttpPostXml(XmlPullParser &xpp, StartTag &stag, const StringBuffer &prefix) : CEsdlTransformOperationBase(xpp, stag, prefix)
+    {
+        if (m_traceName.isEmpty())
+            m_traceName.set(stag.getValue("name"));
+        m_name.set(stag.getValue("name"));
+        m_section.set(stag.getValue("section"));
+        if (m_section.isEmpty())
```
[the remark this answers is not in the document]
afishbeck: it might have just been set on the line above.  this is basically saying the default is temporaries if stag "section" is not set.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -919,15 +964,29 @@ void EsdlServiceImpl::handleServiceRequest(IEspContext &context,
...
+
+        scriptContext->appendContent(ESDLScriptCtxSection_LogData, "LogDataset", logdata);
+
+        if (servicePLTs || methodPLTs)
+        {
+            processServiceAndMethodTransforms(scriptContext, {servicePLTs, methodPLTs}, ESDLScriptCtxSection_LogData, nullptr);
+            scriptContext->toXML(temp, ESDLScriptCtxSection_PreESDLResponse);
+            logdata = temp.str();
+        }
+    }
```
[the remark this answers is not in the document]
afishbeck: JIRA is broken right now.  Let's try and remember to create one for this.  Either of us can do it when we are adding more entry points (or before).

esp/services/esdl_svc_engine/esdl_binding.cpp
afishbeck: agreed

initfiles/examples/EsdlExample/esdl_binding_entrypoints.xml
```diff
@@ -0,0 +1,147 @@
...
+                      </es:element>
+                    </es:element>
+                  </es:element>
+                </es:element>
+              </es:content>
+            </es:http-post-xml>
+            <es:target xpath="soap:Body">
+            <es:target xpath="resp:RoxieEchoPersonInfoResponse">
+            <es:target xpath="resp:Results/resp:Result">
+            <es:target xpath="ds1:Dataset[@name='RoxieEchoPersonInfoResponse']">
```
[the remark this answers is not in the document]
afishbeck: ok will fix.

system/xmllib/libxml_xpathprocessor.cpp
```diff
@@ -372,6 +392,374 @@ class CLibXpathContext : public CInterfaceOf<IXpathContext>
...
+        if (!wellFormed)
+            throw MakeStringException(-1, "XpathContext:addXmlContent: Unable to parse %s XML content", xml);
+
+    }
+    virtual bool ensureLocation(const char *xpath, bool required) override
+    {
+        if (isEmptyString(xpath))
+            return true;
+        if (strstr(xpath, "//")) //maybe in the future, scripting error rather than not found, required==false has no effect
+            throw MakeStringException(XPATHERR_InvalidState,"XpathContext:ensureLocation: '//' not currently allowed '%s'", xpath);
```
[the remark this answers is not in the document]
afishbeck: Agreed, but still not this PR.

system/xmllib/libxml_xpathprocessor.cpp
```diff
@@ -341,15 +348,28 @@ class CLibXpathContext : public CInterfaceOf<IXpathContext>
...
+
     xmlXPathObjectPtr findVariable(const char *name, const char *ns_uri, CLibXpathScope *scope)
     {
+        xmlXPathObjectPtr obj = nullptr;
+        if (parent)
+        {
+            obj = parent->findVariable(name, ns_uri, scope);
+            if (obj)
+                return obj;
+        }
```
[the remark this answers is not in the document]
afishbeck: Parent was a bad name.  In our case this is source context with variables and target context without.  I've changed the variable name to primary which may be less confusing.  Don't want to dictate future uses by calling it source, but it's not parent either.

esp/esdllib/esdl_script.cpp
```diff
@@ -130,51 +207,312 @@ class CEsdlTransformOperationVariable : public CEsdlTransformOperationBase
     }
 };
 
+class CEsdlTransformOperationHttpContentXml : public CEsdlTransformOperationWithChildren
```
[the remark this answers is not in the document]
afishbeck: After I've added a few more operations we can clean this stuff up.  I still want to see how things work out.

esp/esdllib/esdl_script.cpp
```diff
@@ -130,51 +207,312 @@ class CEsdlTransformOperationVariable : public CEsdlTransformOperationBase
...
+    {
+    #if defined(_DEBUG)
+        DBGLOG (">>>>>>>>>>> %s >>>>>>>>>>", m_tagname.str());
+        CEsdlTransformOperationWithChildren::toDBGLog();
+        DBGLOG (">>>>>>>>>>> %s >>>>>>>>>>", m_tagname.str());
+    #endif
+    }
+};
+
+interface IEsdlTransformOperationHttpHeader : public IInterface
```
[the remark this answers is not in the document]
afishbeck: If this doesn't get reused in the next few operations I may simplify.

esp/esdllib/esdl_script.cpp
```diff
@@ -727,37 +1094,34 @@ class CEsdlTransformOperationConditional : public CEsdlTransformOperationWithChi
...
+        if (isEmptyString(op))
+            esdlOperationError(ESDL_SCRIPT_UnknownOperation, m_tagname, "unrecognized conditional", !m_ignoreCodingErrors);
+        //m_ignoreCodingErrors means op may still be null
+        if (!op || streq(op, "if"))
+            m_op = 'i';
+        else if (streq(op, "when"))
+            m_op = 'w';
+        else if (streq(op, "otherwise"))
+            m_op = 'o';
+        if (m_op!='o')
```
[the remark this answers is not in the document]
afishbeck: Should never get here without a recognized name, but I will add that for future proofing.

esp/esdllib/esdl_script.cpp
```diff
@@ -797,31 +1161,61 @@ class CEsdlTransformOperationConditional : public CEsdlTransformOperationWithChi
...
-            loadOtherwise(tree, prefix);
+            case XmlPullParser::START_TAG:
+            {
+                StartTag opTag;
+                xpp.readStartTag(opTag);
+                const char *op = opTag.getLocalName();
+                if (streq(op, "when"))
+                    operations.append(*new CEsdlTransformOperationConditional(xpp, opTag, prefix));
+                else if (streq(op, "otherwise"))
+                    otherwise.setown(new CEsdlTransformOperationConditional(xpp, opTag, prefix));
```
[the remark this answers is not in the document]
afishbeck: Adding checking.

esp/esdllib/esdl_script.cpp
```diff
@@ -834,62 +1228,243 @@ class CEsdlTransformOperationChoose : public CEsdlTransformOperationWithChildren
...
+public:
+    CEsdlTransformOperationEnsureTarget(XmlPullParser &xpp, StartTag &stag, const StringBuffer &prefix) : CEsdlTransformOperationTarget(xpp, stag, prefix)
+    {
+        m_ensure = true;
+    }
+
+    virtual ~CEsdlTransformOperationEnsureTarget(){}
+};
+
+class CEsdlTransformOperationSource : public CEsdlTransformOperationWithChildren
```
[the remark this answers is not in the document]
afishbeck: Can refactor later.  I'd like to wait a bit in case these diverge.

afishbeck:
@timothyklemm @asselitx updated, please review.  I'll have to resolve some new merge conflicts soon.

esp/bindings/SOAP/xpp/xpp/XmlPullParser.h
```diff
@@ -174,7 +174,8 @@ namespace xpp {
     }
 
 
-    int skipSubTree()  {
+    bool skipSubTree()  {
```
timothyklemm, line 177: The platform isn't currently using the return value, so this change looks reasonable. But should we assume that nothing else uses the existing return value?

timothyklemm (commented):
@afishbeck Just a few questions/comments.

esp/esdllib/esdl_script.cpp
```diff
@@ -1407,7 +1409,7 @@ void createEsdlTransformOperations(IArrayOf<IEsdlTransformOperation> &operations
     }
 }
 
-IEsdlTransformOperation *createEsdlTransformOperation(XmlPullParser &xpp, const StringBuffer &prefix, bool withVariables)
+IEsdlTransformOperation *createEsdlTransformOperation(XmlPullParser &xpp, const StringBuffer &prefix, bool withVariables, bool ignoreCodingErrors)
```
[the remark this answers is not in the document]
afishbeck: The parameter has to be there because the function has to match the signature "esdlOperationsFactory_t" which allows functions to make use of the parameter, but doesn't require it.  It would generally not be passed to a given operation since they have their own ignore option.  Rather if the function itself is being called as a child and has an error in it's own code, it needs to know what to do.

esp/bindings/SOAP/xpp/xpp/XmlPullParser.h
afishbeck: The code couldn't make any real use of it since it always returned XmlPullParser::END_TAG whether it cycled through some children or not, but I'll make a new function anyway.

system/xmllib/libxml_xpathprocessor.cpp
```diff
@@ -718,10 +741,10 @@ class CLibXpathContext : public CInterfaceOf<IXpathContext>, implements IVariabl
         if (obj->nodesetval->nodeNr>1 && !all)
         {
             xmlXPathFreeObject(obj);
-            throw MakeStringException(XPATHERR_InvalidState,"XpathContext:remove: ambiguous xpath xpath '%s' to remove all set 'all'", xpath);
+            throw MakeStringException(XPATHERR_InvalidState,"XpathContext:remove: ambiguous xpath '%s' to remove all set 'all'", xpath);
         }
         xmlNodePtr *nodes = obj->nodesetval->nodeTab;
-        for (int i = 0; i < obj->nodesetval->nodeNr; i++)
+        for (int i = obj->nodesetval->nodeNr; i >= 0; i--)
```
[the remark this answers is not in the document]
afishbeck: Yes, thanks for catching that.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -1526,10 +1519,10 @@ void EsdlServiceImpl::prepareFinalRequest(IEspContext &context,
...
 
-    if (m_serviceLevelCrtFail)
-        throw MakeStringException(-1, "%s::%s disabled due to service-level Custom Transform errors. Review transform template in configuration.", srvdef.queryName(), mthName);
+    if (m_serviceScriptError.length())
+        throw MakeStringException(-1, "%s::%s disabled due to service-level Custom Transform errors. [%s]. Review transform template in configuration.", srvdef.queryName(), mthName, m_serviceScriptError.str());
 
-    StringAttr *crtErrorMessage = m_requestTransformErrors.getValue(mthName);
+    StringAttr *crtErrorMessage = m_methodScriptErrors.getValue(mthName);
     if (crtErrorMessage && !crtErrorMessage->isEmpty())
         throw MakeStringException(-1, "%s::%s disabled due to method-level Custom Transform errors: %s. Review transform template in configuration.", srvdef.queryName(), mthName, crtErrorMessage->str());
```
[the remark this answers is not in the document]
afishbeck: I'll try and make consistent.

afishbeck:
@timothyklemm @asselitx I've updated based on comments, but also rebased to latest master.

esp/esdllib/esdl_script.cpp
```diff
@@ -69,44 +85,89 @@ static inline StringBuffer &makeOperationTagName(StringBuffer &s, const StringBu
...
     Owned<ICompiledXpath> m_select;
 
 public:
-    CEsdlTransformOperationVariable(IPropertyTree *tree, const StringBuffer &prefix) : CEsdlTransformOperationBase(tree, prefix)
+    CEsdlTransformOperationVariable(XmlPullParser &xpp, StartTag &stag, const StringBuffer &prefix) : CEsdlTransformOperationWithChildren(xpp, stag, prefix, true, nullptr)
     {
         if (m_traceName.isEmpty())
-            m_traceName.set(tree->queryProp("@name"));
-        m_name.set(tree->queryProp("@name"));
+            m_traceName.set(stag.getValue("name"));
```
asselitx, line 166: Might be beneficial to have this logic in the base class to ensure that all operations default to the name as traceName when its empty.

esp/esdllib/esdl_script.cpp
```diff
@@ -115,11 +176,22 @@ class CEsdlTransformOperationVariable : public CEsdlTransformOperationBase
...
     }
 
-    virtual bool process(IEsdlScriptContext * context, IPropertyTree *content, IXpathContext * xpathContext) override
+    virtual bool process(IEsdlScriptContext * scriptContext, IXpathContext * targetContext, IXpathContext * sourceContext) override
     {
         if (m_select)
-            return xpathContext->addCompiledVariable(m_name, m_select);
-        return xpathContext->addVariable(m_name, "");
+            return sourceContext->addCompiledVariable(m_name, m_select);
+        else if (m_children.length())
```
asselitx, line 183: This case assigns the value of the variable to the result of processing the children?

esp/esdllib/esdl_script.cpp
```diff
@@ -130,51 +202,312 @@ class CEsdlTransformOperationVariable : public CEsdlTransformOperationBase
...
+            StringBuffer status;
+            StringBuffer response;
+            httpclient->sendRequest(headers, "POST", "text/xml", content, response, status);
+            targetContext->ensureSetValue("@status", status.str(), true);
+            size32_t len = response.trim().length();
+            if (!len)
+            {
+                throw MakeStringException(ESDL_SCRIPT_Error, "ESDL Script empty result calling http-post-xml %s url(%s)", m_traceName.str(), url);
+            }
+            else if (*response.str()=='<')
```
asselitx, line 406: I don't know if we really can or need to do anything differently here, but does the HTTP spec say anything about guarantees of response type given a certain Content Type? I'm guessing not really, but we could have a situation here where some text is returned and just happens to start with '<'. Then we'd just return whatever XML parse error resulting from addXmlContent().

esp/esdllib/esdl_script.cpp
```diff
@@ -328,42 +666,86 @@ class CEsdlTransformOperationSetValue : public CEsdlTransformOperationBase
...
+protected:
+    StringAttr m_prefix;
+    StringAttr m_uri;
+    bool m_current = false;
+
+public:
+    CEsdlTransformOperationNamespace(XmlPullParser &xpp, StartTag &stag, const StringBuffer &prefix) : CEsdlTransformOperationWithoutChildren(xpp, stag, prefix)
+    {
+        if (m_traceName.isEmpty())
+            m_traceName.set(stag.getValue("name"));
```
asselitx, line 689: Does the namespace operation really use @name? If not perhaps @prefix or @uri would be better fallbacks for the traceName.

esp/esdllib/esdl_script.cpp
```diff
@@ -328,42 +666,86 @@ class CEsdlTransformOperationSetValue : public CEsdlTransformOperationBase
...
+
+        const char *pfx = stag.getValue("prefix");
+        const char *uri = stag.getValue("uri");
+        if (!pfx && isEmptyString(uri))
```
asselitx, line 693: Not a big deal at all but why not use isEmptyString() to check both?

esp/esdllib/esdl_script.cpp
```diff
@@ -328,42 +666,86 @@ class CEsdlTransformOperationSetValue : public CEsdlTransformOperationBase
...
+    bool all = false;
 
 public:
-    CEsdlTransformOperationRenameNode(IPropertyTree *tree, const StringBuffer &prefix) : CEsdlTransformOperationBase(tree, prefix)
+    CEsdlTransformOperationRenameNode(XmlPullParser &xpp, StartTag &stag, const StringBuffer &prefix) : CEsdlTransformOperationWithoutChildren(xpp, stag, prefix)
     {
-        if (isEmptyString(tree->queryProp("@new_name")) && isEmptyString(tree->queryProp("@xpath_new_name")))
+        const char *new_name = stag.getValue("new_name");
+        const char *xpath_new_name = stag.getValue("xpath_new_name");
+        if (isEmptyString(new_name) && isEmptyString(xpath_new_name))
```
asselitx, line 730: Perhaps set m_traceName to new_name or xpath_new_name if empty?

esp/esdllib/esdl_script.cpp
```diff
@@ -377,29 +759,75 @@ class CEsdlTransformOperationRenameNode : public CEsdlTransformOperationBase
...
                 path.set(m_target);
 
             StringBuffer name;
             if (m_xpath_new_name)
-                xpathContext->evaluateAsString(m_xpath_new_name, name);
+                sourceContext->evaluateAsString(m_xpath_new_name, name);
             else
                 name.set(m_new_name);
 
             //make sure we only rename one.  we can add a multi-node support in the future (when we are libxml2 based?)
```
asselitx, line 780: Is this comment now obsolete?

esp/esdllib/esdl_script.cpp
```diff
@@ -834,62 +1230,243 @@ class CEsdlTransformOperationChoose : public CEsdlTransformOperationWithChildren
...
+    bool process(IEsdlScriptContext * scriptContext, IXpathContext * targetContext, IXpathContext * sourceContext) override
     {
-        StringBuffer xpath;
-        IPropertyTree * otherwise = tree->queryPropTree(makeOperationTagName(xpath, prefix, "otherwise"));
-        if (!otherwise)
-            return;
-        m_children.append(*new CEsdlTransformOperationConditional(otherwise, prefix));
+        CXpathContextLocation location(targetContext);
+        bool success = false;
+        if (m_ensure)
```
asselitx, line 1259: Will we be able to set this to true in the future via an `@ensure` attribute?

esp/esdllib/esdl_script.cpp
```diff
@@ -834,62 +1230,243 @@ class CEsdlTransformOperationChoose : public CEsdlTransformOperationWithChildren
...
+
+class CEsdlTransformOperationElement : public CEsdlTransformOperationWithChildren
+{
+protected:
+    StringBuffer m_name;
+    StringBuffer m_nsuri;
+
+public:
+    CEsdlTransformOperationElement(XmlPullParser &xpp, StartTag &stag, const StringBuffer &prefix) : CEsdlTransformOperationWithChildren(xpp, stag, prefix, true, nullptr)
+    {
```
asselitx, line 1358: if `@trace` is empty could we fallback to using `@name`?

esp/esdllib/esdl_script.cpp
```diff
@@ -1024,7 +1541,7 @@ class CEsdlCustomTransform : public CInterfaceOf<IEsdlCustomTransform>
         else
             prefixes.appendUniq("");
     }
-    virtual void toDBGLog() override
+    virtual void toDBGLogx() override
```
asselitx, line 1544: Is the trailing x in the function name a typo?

afishbeck: oops that was me tracking down usage.  I'll change the name back.

system/xmllib/libxml_xpathprocessor.cpp
```diff
@@ -372,6 +393,396 @@ class CLibXpathContext : public CInterfaceOf<IXpathContext>
...
+    }
+
+    virtual bool setLocation(ICompiledXpath * compiledXpath, bool required) override
+    {
+        if (!compiledXpath)
+            return false;
+
+        CLibCompiledXpath * ccXpath = static_cast<CLibCompiledXpath *>(compiledXpath);
+        if (!ccXpath)
+            throw MakeStringException(XPATHERR_InvalidState,"XpathProcessor:evaluateAsNumber: Error: BAD compiled XPATH");
```
asselitx, line 440: Shouldn't the error read `XpathProcessor:setLocation`?

system/xmllib/libxml_xpathprocessor.cpp
```diff
@@ -372,6 +393,396 @@ class CLibXpathContext : public CInterfaceOf<IXpathContext>
...
+
+        xmlXPathObjectPtr obj = evaluate(ccXpath->getCompiledXPathExpression(), compiledXpath->getXpath());
+        return setLocation(obj, required, compiledXpath->getXpath());
+    }
+
+    virtual bool setLocation(const char *s, bool required) override
+    {
+        StringBuffer xpath;
+        replaceVariables(xpath, s, false, this, "{$", "}");
```
asselitx, line 449: So this change allows us to use the ${...} variable substitution as strings in xpaths wherever this function is used?

system/xmllib/libxml_xpathprocessor.cpp
```diff
@@ -372,6 +393,396 @@ class CLibXpathContext : public CInterfaceOf<IXpathContext>
...
+
+            xmlXPathObjectPtr obj = xmlXPathNodeEval(node, (const xmlChar*) finger, m_xpathContext);
+            xmlXPathSetContextNode(current, m_xpathContext); //restore
+
+            if (!obj)
+                throw MakeStringException(XPATHERR_InvalidInput,"XpathContext:ensureLocation: invalid xpath '%s' at node '%s'", xpath, finger);
+
+            if (obj->type!=xmlXPathObjectType::XPATH_NODESET || !obj->nodesetval || obj->nodesetval->nodeNr==0)
+            {
+                if (*finger=='.' || strpbrk(finger, "[()]*") || strstr(finger, "::")) //complex nodes must already exist
```
asselitx, line 519: Will be good to ensure that we document this constraint.

system/xmllib/libxml_xpathprocessor.cpp
```diff
@@ -372,6 +393,396 @@ class CLibXpathContext : public CInterfaceOf<IXpathContext>
...
+    virtual void setLocationNamespace(const char *prefix, const char * uri, bool current) override
+    {
+        xmlNodePtr node = m_xpathContext->node;
+        if (!node)
+            return;
+        xmlNsPtr ns = nullptr;
+        if (isEmptyString(uri))
+            ns = xmlSearchNs(node->doc, node, (const xmlChar *) prefix);
+        else
+            ns = xmlNewNs(node, (const xmlChar *) uri, (const xmlChar *) prefix);
```
asselitx, line 559: What if the prefix/uri is a redefinition? Will xmlNewNs throw an error? Or what if the prefix is reused with a different uri? Asking mostly because some of the libxml errors aren't very friendly in that they don't  show up in the esp.log file.

system/xmllib/libxml_xpathprocessor.cpp
```diff
@@ -195,14 +195,16 @@ class XpathContextState
...
 typedef std::vector<XpathContextState> XPathContextStateVector;
 
-class CLibXpathContext : public CInterfaceOf<IXpathContext>
+class CLibXpathContext : public CInterfaceOf<IXpathContext>, implements IVariableSubstitutionHelper
 {
 public:
     XPathInputMap provided;
     xmlDocPtr m_xmlDoc = nullptr;
     xmlXPathContextPtr m_xpathContext = nullptr;
     ReadWriteLock m_rwlock;
```
asselitx, line 204: Which operations does this need to protect? Only some of the methods in this class use it. Some do a fair amount of work before calling `setLocation()` and others like `rename()` and `remove()` don't appear to use it at all, which seems like it might be an oversight.

asselitx:
I've completed my review @afishbeck . With the possible exception of the use of m_rwlock in libxml_xpathprocessor.cpp none of my comments are critical - at worst they may be future cleanup.

esp/bindings/SOAP/xpp/xpp/XmlPullParser.h
timothyklemm: Your change is fine to address the original comment, but I am concerned about the original implementation.

If skipSubTree is used on a document not known to be well-formed, an unbalanced start tag can put it into an infinite loop. It only leaves the loop when the end tags balance. I remember writing code before where I checked to make sure I wasn't at the end of the document after skipping. I didn't realize until now that it's an unhandled error, unless I've misread next() and it doesn't return END_DOCUMENT infinitely.

It's also not clear that the well-formed behavior is consistent given different starting positions. Skipping from the start tag skips the entire element. Skipping from content skips the remainder of the current element. Skipping from the end tag of the current element skips the remainder of the parent element (and foolishly skipping from the end of the root element will never return). The first two make sense to me, but I could easily see the third being a NO-OP instead.

esp/esdllib/esdl_script.cpp
afishbeck: Yes, exactly.  It creates a temporary location for that xml in a special section of the script context tree and then changes the target for processing of the children to that location.  The end result is structured xml content stored in the variable.

esp/esdllib/esdl_script.cpp
afishbeck: Yeah even the original use of name was a bit weird.  I'll remove it or use one of the other values for now.

esp/esdllib/esdl_script.cpp
afishbeck: "" is a valid prefix, but if you are relying on the uri it can't be empty.  That is , pfx=nullptr, uri="" has nothing to match on, pfx="", uri=nullprt can search for an existing prefix definition for "".

esp/esdllib/esdl_script.cpp
afishbeck: will do.

esp/esdllib/esdl_script.cpp
afishbeck: yes, I think so.

esp/esdllib/esdl_script.cpp
afishbeck: done.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -410,74 +410,155 @@ void EsdlServiceImpl::configureUrlMethod(const char *method, IPropertyTree &entr
...
+//Need to fix up serialized PTREE order for backward compatability of scripts
+
+StringBuffer &toScriptXML(IPropertyTree *tree, StringBuffer &xml, int indent)
+{
+    StringArray raise;
+    raise.append("xsdl:param");
+    raise.append("xsdl:variable");
+
+    StringArray lower;
+    lower.append("xsdl:otherwise");
```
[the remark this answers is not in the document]
afishbeck: just here for completeness, and to make the modified xml read more accurately.

system/xmllib/libxml_xpathprocessor.cpp
afishbeck: I may just remove this in the future.  The scripting is not multithreaded and would have to be majorly refactored to make it so.  It's not a declarative language.

system/xmllib/libxml_xpathprocessor.cpp
afishbeck: Yes let me know when you'd like to discuss it.  I'll have to refresh my memory on exactly where this type of substitution is supported.

system/xmllib/libxml_xpathprocessor.cpp
afishbeck: Yes.  Once explained in the docs it would be pretty clear why this is necessary.  For example I can't add a node named a*b, but I could fine a node that matched that pattern.

system/xmllib/libxml_xpathprocessor.cpp
afishbeck: Right, I don't think you'd get an error redefining a prefix.  And routines relying on it would find the nearest location where the prefix was defined.  libxml errors and implementation can be a bit quirky though.

system/xmllib/xpathprocessor.hpp
```diff
@@ -90,23 +105,45 @@ class CXpathContextScope : CInterface
     }
 };
 
+class CXpathContextLocation : CInterface
+{
+private:
+    Linked<IXpathContext> context;
+public:
+    IMPLEMENT_IINTERFACE;
+    CXpathContextLocation(IXpathContext *ctx) : context(ctx)
```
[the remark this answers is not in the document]
afishbeck: I now check for null.  I want this to be easy to use and not have anyone worry about the current value, casting to reference, etc.

system/xmllib/libxml_xpathprocessor.cpp
```diff
@@ -372,6 +393,396 @@ class CLibXpathContext : public CInterfaceOf<IXpathContext>
...
+        }
+        xmlNodePtr node = current;
+
+        StringArray xpathnodes;
+        xpathnodes.appendList(xpath, "/");
+        ForEachItemIn(i, xpathnodes)
+        {
+            const char *finger = xpathnodes.item(i);
+            if (isEmptyString(finger)) //shouldn't happen, scripting error rather than not found, required==false has no effect
+                throw MakeStringException(XPATHERR_InvalidState,"XpathContext:ensureValue: '//' not allowed '%s'", xpath);
```
timothyklemm, line 587: Possibly misleading message. You've already rejected any occurrence of "//", so this is caused by trimmed whitespace when splitting the path, or (possibly) a leading or trailing '/'. It would be better to distinguish this exception message from the one above.

system/xmllib/libxml_xpathprocessor.cpp
```diff
@@ -1064,6 +1499,36 @@ class CEsdlScriptContext : public CInterfaceOf<IEsdlScriptContext>
     {
         return getSectionNode(name, nullptr);
     }
+
+    xmlNodePtr ensureCopiedSection(const char *target, const char *source, bool replace)
+    {
+        sanityCheckSectionName(target);
+        sanityCheckSectionName(source);
+        if (isEmptyString(source)||isEmptyString(target))
+            return nullptr;
```
timothyklemm, line 1508: These checks may be backwards. If strpbrk, used by sanityCheckSectionName, is anything like other standard library functions that behave badly when passed null, null should be checked first (or as part of the sanity check if it makes sense).

timothyklemm (dismissed):

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -1431,22 +1505,31 @@ void EsdlServiceImpl::prepareFinalRequest(IEspContext &context,
...
+    }
 
     if (scriptContext)
     {
-        IEsdlCustomTransform *methodCrt = m_customRequestTransformMap.getValue(mthName);
+        IEsdlTransformSet *serviceCRTs = m_transforms->queryMethodEntryPoint("", ESDLScriptEntryPoint_Legacy);
+        IEsdlTransformSet *methodCRTs = m_transforms->queryMethodEntryPoint(mthName, ESDLScriptEntryPoint_Legacy);
+
+        IEsdlTransformSet *serviceRTs = m_transforms->queryMethodEntryPoint("", ESDLScriptEntryPoint_BackendRequest);
+        IEsdlTransformSet *methodRTs = m_transforms->queryMethodEntryPoint(mthName, ESDLScriptEntryPoint_BackendRequest);
```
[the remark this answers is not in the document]
afishbeck: These aren't unordered and ordered.  These are entry point names.  They are equavent to each other and if I add a "BackEndRequest" entry point to a service that already has Legacy entry points it will work.  Whether things are ordered or now is based on how the xml is added as CDATA or inline.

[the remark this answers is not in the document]
timothyklemm: I get the technicalities, but in reality there is no existing legacy markup that uses CDATA (thus all such markup is unordered) and new markup will likely use CDATA (and be ordered). I question whether it is necessary and whether an arbitrary decision that legacy transforms will be processed first will be sufficient. Maybe I (as a potential end user) want new markup to set things up for the legacy, so should we support an option to switch the order? Maybe I have an unrealistic expectation that I should somehow be able to intermingle old and new.

My thought (as a reviewer) is that we should support both, but not together. It's just my opinion, and I'm probably still influenced by my abandoned refactoring where I wasn't even looking for legacy if new markup was found. I also wasn't allowing legacy syntax to include new operations. It was pick one or the other, but never both.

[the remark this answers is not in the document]
afishbeck: We will eventually deprecate the old name, but I don't want to dictate that people have to do that right away just to use a new tag.  I'm going to fix it so that the order will be preserved across both types.

afishbeck:
@asselitx @timothyklemm updated based on comments.  Please check httpclient changes (required to check content-type) for any possible regressions as its a shared library.

esp/esdllib/esdl_xpath_extensions_libxml.cpp
```diff
@@ -345,6 +345,67 @@ static void storedValueExistsFunction (xmlXPathParserContextPtr ctxt, int nargs)
...
+    {
+        xmlFree(delimiters);
+        return;
+    }
+
+    StringBuffer resultPath;
+    if (!scriptContext->tokenize((const char *)str, (const char *)delimiters, resultPath))
+    {
+        xmlXPathSetError((ctxt), XPATH_EXPR_ERROR);
+        return;
```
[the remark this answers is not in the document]
asselitx: Also what about str? Could it be leaked here or above when `if(xmlXPathCheckError(ctxt)...)` is true and str!=NULL?

esp/esdllib/esdl_script.cpp
```diff
@@ -396,22 +396,37 @@ class CEsdlTransformOperationHttpPostXml : public CEsdlTransformOperationBase
...
+            StringBuffer contentType;
+            resp->getContentType(contentType);
+            targetContext->ensureSetValue("@content-type", contentType.str(), true);
+            if (strnicmp("text/xml", contentType.str(), 8)==0 || strnicmp("application/xml", contentType.str(), 15) ==0)
             {
                 CXpathContextLocation content_location(targetContext);
                 targetContext->addElementToLocation("content");
                 targetContext->addXmlContent(response.str());
             }
             else
```
timothyklemm, line 426: Curious if, with the availability of content-type, it might be beneficial to recognize JSON (or anything we can load into a PTree) then convert to XML. Add the text so somebody expecting the actual response can get it, but do the conversion automatically so it's ready to use.

On the other hand, if any non-XML content type is considered an error, should we make sure the request asks for XML results?

esp/esdllib/esdl_script.cpp
```diff
@@ -417,17 +845,18 @@ class CEsdlTransformOperationRenameNode : public CEsdlTransformOperationBase
     }
 };
 
-class CEsdlTransformOperationRemoveNode : public CEsdlTransformOperationBase
+class CEsdlTransformOperationRemoveNode : public CEsdlTransformOperationWithoutChildren
 {
 protected:
     StringAttr m_target;
     Owned<ICompiledXpath> m_xpath_target;
+    bool all = false;
```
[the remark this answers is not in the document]
timothyklemm: Only one of two operations with this comment has been changed.

asselitx:
@afishbeck httpclient changes look good

richardkchapman:
Some build failures of the form:

D:\a\HPCC-Platform\HPCC-Platform\esp\bindings\http/platform/httptransport.ipp(165,27): error C3668: 'CHttpMessage::getHeader': method with override specifier 'override' did not override any base class methods [D:\a\HPCC-Platform\HPCC-Platform\build\esp\protocols\http\esphttp.vcxproj]
1333

esp/esdllib/esdl_script.cpp
afishbeck: I'll ask for xml.  I named the function with xml in the name to put off the complexity of conversion.  If I were already loading into a ptree it would be different.  Later I can probably create something that parses json directly into libxml.  I don't like the ptree as an intermediary here.

afishbeck:
Fixed build issue.
Also added sending an Accept header from the post-xml script method.
@asselitx @timothyklemm please review.  Hopefully the last time.  Then I'll squash and rebase.

esp/esdllib/esdl_script.cpp
```diff
@@ -727,37 +1089,37 @@ class CEsdlTransformOperationConditional : public CEsdlTransformOperationWithChi
...
+            if (isEmptyString(test))
+                esdlOperationError(ESDL_SCRIPT_MissingOperationAttr, m_tagname, "without test", !m_ignoreCodingErrors);
+            m_test.setown(compileXpath(test));
         }
     }
 
     virtual ~CEsdlTransformOperationConditional(){}
 
-    bool process(IEsdlScriptContext * context, IPropertyTree *content, IXpathContext * xpathContext) override
+    virtual bool process(IEsdlScriptContext * scriptContext, IXpathContext * targetContext, IXpathContext * sourceContext) override
```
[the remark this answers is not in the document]
afishbeck: Who did you have the debate with?  I saw a comment somewhere else that it is stylistically preferred in this project.. so following that recommendation.. and personally prefer it this way.  If there was another consensus we can all follow it, I just don't know of one.

esp/esdllib/esdl_script.cpp
```diff
@@ -903,118 +1480,58 @@ static inline void replaceVariable(StringBuffer &s, IXpathContext *xpathContext,
...
-        treeMap.setValue(mergedTarget, opTree);
-    return opTree;
-}
 
 class CEsdlCustomTransform : public CInterfaceOf<IEsdlCustomTransform>
 {
 private:
-    IArrayOf<IEsdlTransformOperation> m_variables; //keep separate and only at top level for now
     IArrayOf<IEsdlTransformOperation> m_operations;
     Owned<IProperties> namespaces = createProperties(false);
```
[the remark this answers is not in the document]
afishbeck: This PR is to big for genaral cleanup of existing code.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -899,15 +943,29 @@ void EsdlServiceImpl::handleServiceRequest(IEspContext &context,
...
-    handleResultLogging(context, scriptContext, tgtctx.get(), req, soapmsg.str(), origResp.str(), out.str(), logdata.str());
-    context.addTraceSummaryTimeStamp(LogNormal, "end-resLogging");
     ESPLOG(LogMax,"Customer Response: %s", out.str());
 }
 
-bool EsdlServiceImpl::handleResultLogging(IEspContext &espcontext, IEsdlScriptContext *scriptContext, IPropertyTree * reqcontext, IPropertyTree * request,const char *rawreq, const char * rawresp, const char * finalresp, const char * logdata)
+bool EsdlServiceImpl::handleResultLogging(IEspContext &espcontext, IEsdlScriptContext *scriptContext, IEsdlDefService &srvdef, IEsdlDefMethod &mthdef, IPropertyTree * reqcontext, IPropertyTree * request,const char *rawreq, const char * rawresp, const char * finalresp, const char * logdata)
 {
+    StringBuffer temp;
+    if (scriptContext)
```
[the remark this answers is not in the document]
afishbeck: This is conditional on the scripts having been configured.  It's up to the person creating the binding.  Despite the name these scripts could be used for anything.  Calling out to another service at completion for example.
Open another JIRA about TxSummary entries for scripts.

esp/esdllib/esdl_xpath_extensions_libxml.cpp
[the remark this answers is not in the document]
afishbeck: done.

afishbeck:
@asselitx @timothyklemm this needs final sign off.  I saw a couple more comments I missed and made a couple changes.  Please review and sign off.

asselitx (dismissed):
Looks good

timothyklemm (dismissed):

timothyklemm (approved):

afishbeck:
@richardkchapman rebased and squashed.  This PR should be ready to merge.

asselitx (approved):

PR #14334, opened 2020-11-23
HPCC-25006 Add RoxieQueueFilter to TpLogicalClusterQuery

wangkx:
This filter is used to filter targets in a cloud environment.
The filter has three options: All, QueriesOnly, and WorkunitsOnly
(default to All). 

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/scm/ws_topology.ecm
+++ b/esp/scm/ws_topology.ecm
@@ -17,11 +17,11 @@
 
 ////////////////////////////////////////////////////////////
 
-ESPenum DeployOnlyFilter : string
+ESPenum RoxieQueueFilter : string
 {
     All("All"),
-    DeployOnly("DeployOnly"),
-    NotDeployOnly("NotDeployOnly")
+    QueriesOnly("QueriesOnly"),
+    WorkunitsOnly("WorkunitsOnly")
 };
 
 //  ===========================================================================
@@ -69,7 +69,7 @@ ESPStruct TpLogicalCluster
         [min_ver("1.14")] string LanguageVersion;
     [min_ver("1.11")] string Process;
     [min_ver("1.22")] string Type;
-    [min_ver("1.31")] bool DeployOnly;
+    [min_ver("1.31")] bool QueriesOnly;
 };
 //  ===========================================================================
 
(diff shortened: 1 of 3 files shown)


wangkx:
@afishbeck please review

esp/scm/ws_topology.ecm
```diff
@@ -17,6 +17,13 @@
 
 ////////////////////////////////////////////////////////////
 
+ESPenum DeployOnlyFilter : string
+{
+    All("All"),
+    DeployOnly("DeployOnly"),
+    NotDeployOnly("NotDeployOnly")
+};
```
afishbeck, line 25: @richardkchapman Just want to make sure we have the intention and terminology right here.
In many places deploy some ecl is to compile the workunit.. that's not the same thing as publishing a query.  So there are 3 actions.
1. deploy / compile a workunit.
2. run a workunit
3. publish a query.
ECL could be compiled for the roxie (deployed as workunits) but wouldn't end up callable as queries yet.

If the focus is on the workunit it could be:
All, Runnable, CompileOnly.

If the focus is on how things are run it could be:
All, WorkunitsOnly, QueriesOnly

In both cases we really allow publishing/compiling to other cluster types so in practice it would just be All and either CompileOnly or QueriesOnly.

wangkx: I guess we are trying to separate 2 groups of roxies here: the roxies for running (compile and run) workunits and the roxies for publishing queries. If so, we should change to 'All, WorkunitsOnly, QueriesOnly'. But, I could be wrong.

afishbeck: Makes sense to me.

richardkchapman:
I think this is back with @wangkx

wangkx:
@afishbeck please review

afishbeck (approved):
@wangkx looks good.

PR #14345, opened 2020-11-25
HPCC-24262 Introduce an AtomicShared class

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jscm.hpp
+++ b/system/jlib/jscm.hpp
@@ -36,6 +36,8 @@
  #define jlib_thrown_decl DECL_IMPORT
 #endif
 
+#include <atomic>
+
 interface IInterface
 {
     virtual void Link() const = 0;
@@ -62,8 +64,8 @@ public:
     inline ~Shared()                             { ::Release(ptr); }
     inline Shared<CLASS> & operator = (const Shared<CLASS> & other) { this->set(other.get()); return *this;  }
 
-    inline CLASS * operator -> () const         { return ptr; } 
-    inline operator CLASS *() const             { return ptr; } 
+    inline CLASS * operator -> () const         { return ptr; }
+    inline operator CLASS *() const             { return ptr; }
 
     inline void clear()                         { CLASS *temp=ptr; ptr=NULL; ::Release(temp); }
     inline CLASS * get() const                  { return ptr; }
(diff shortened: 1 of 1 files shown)


ghalliday:
@richardkchapman @mckellyln @jakesmith what do you think?  I suspect the number of potential uses is fairly small, it it could be useful in those situations.

system/jlib/jscm.hpp
```diff
@@ -94,6 +109,80 @@ template <class CLASS> class Shared
...
+        CLASS * temp = ptr.exchange(_ptr);
+        ::Release(temp);
+    }
+    inline CLASS * swap(CLASS * _ptr)
+    {
+        return ptr.exchange(_ptr);
+    }
+    inline void swap(AtomicShared<CLASS> & other)
+    {
+        CLASS * temp = other.getClear();
```
ghalliday, line 169: Should be this->getLink(), or do the other way around.

 and should define the semantics so that this is only ever updated once.

system/jlib/jscm.hpp
```diff
@@ -81,6 +83,19 @@ template <class CLASS> class Shared
     }
     inline void set(const Shared<CLASS> &other) { this->set(other.get()); }
     inline void setown(CLASS * _ptr)            { CLASS * temp = ptr; ptr = _ptr; ::Release(temp); }
+    inline bool setownNull(CLASS * _ptr)
```
richardkchapman, line 86: Took me a few moments to process what this was for. Would setownIfNull be a better name?

richardkchapman: Or setownIfUnset

ghalliday: yes, I think I originally thought of calling it setownIfNull().  A slightly unwieldy name is probably better than a confusing name.

system/jlib/jscm.hpp
ghalliday: Now changed and added a comment to document

mckellyln:
ok,sorry, will review this tomorrow.

system/jlib/jscm.hpp
```diff
@@ -94,6 +109,81 @@ template <class CLASS> class Shared
     CLASS * ptr;
 };
 
+// Similar to class Shared<X>, but thread safe versions of the functions (avoid the need for critical sectiions)
```
mckellyln, line 112: spelling of sections

ghalliday:
https://github.com/hpcc-systems/HPCC-Platform/commit/142158ad7383e7569340aa9a69dd4aed78401b5c is an example of how it would be used.

richardkchapman:
Should this target master?

ghalliday:
> Should this target master?

It probably should now.  When it was created it was more suitable for 7.12.x.

ghalliday:
I'll re-target when I squash.

mckellyln (commented):
Looks good.
I am just curious about:
inline AtomicShared(CLASS * _ptr, bool owned);
and
inline void set(CLASS * _ptr);
If there could be any chance for _ptr to change after being set and before Link ?

ghalliday:
> Looks good.
> I am just curious about:
> inline AtomicShared(CLASS * _ptr, bool owned);
> and
> inline void set(CLASS * _ptr);
> If there could be any chance for _ptr to change after being set and before Link ?

I don't think so.. From the caller's point of view the object needs to be guaranteed to be alive while the function call is going on - and the link occurs on the source before the function returns.
For the constructor nothing else can be modifying the object.  For the set() call, it uses setown() which ensures that the pointer is updated atomically.

mckellyln (approved):
ok.
Approved.

PR #14350, opened 2020-11-26
HPCC-18854 Roxie fails with "STATE: activity 11 reset without stop

richardkchapman:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Cloud-compatibility
  - [x] Premature optimization

## Smoketest:

## Testing:
Tested the previously failing case manually. Smoketest should (hopefully) catch any regressions

### DIFF
--- a/roxie/ccd/ccdserver.cpp
+++ b/roxie/ccd/ccdserver.cpp
@@ -2755,6 +2755,15 @@ public:
             return NULL;
     }
 
+    virtual void stop() override
+    {
+        for (unsigned idx = 0; idx < numStreams; idx++)
+        {
+            streamArray[idx]->stop();
+        }
+        CRoxieServerActivity::stop();
+    }
+
     virtual void reset()
     {
         for (unsigned i = 0; i < numInputs; i++)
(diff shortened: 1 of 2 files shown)


richardkchapman:
@ghalliday

richardkchapman:
@ghalliday - This may result in some stops on activities that are already stopped, but that's ok.

roxie/ccd/ccdserver.cpp
```diff
@@ -2755,6 +2755,15 @@ class CRoxieServerMultiInputBaseActivity : public CRoxieServerActivity
             return NULL;
     }
 
+    virtual void stop() override
```
ghalliday, line 2758: @richardkchapman 
Some of the base classes already have some code like this e.g.
NonEmpty, Case - conditionally depending on what has been called already

As long as it is ok to call stop() multiple times then it shouldn't make things worse, but not convinced it is the "coorect" fix.  Worth looking a bit more?

richardkchapman: I did spend a little while looking at whether it was necessary to check that we don't call stop more than once. In general I don't think it's possible to be 100% sure that you do not do so, nor is it necessary given there is code in several places explicitly to cope with having called stop more than once.

ghalliday (approved):

PR #14381, opened 2020-12-03
HPCC-25046 Return hosts in FileSpray.DropZoneFiles

wangkx:
The existing FileSpray.DropZoneFiles returns IP address. The
ESP service method should return hosts.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/services/ws_fs/ws_fsService.cpp
+++ b/esp/services/ws_fs/ws_fsService.cpp
@@ -2981,9 +2981,9 @@ bool CFileSprayEx::onDropZoneFileSearch(IEspContext &context, IEspDropZoneFileSe
         if (!directory.length())
             throw MakeStringException(ECLWATCH_INVALID_INPUT, "DropZone Directory not found for %s.", dropZoneName);
 
-        IpAddress ipToMatch;
-        ipToMatch.ipset(dropZoneServerReq);
-        if (ipToMatch.isNull())
+        IpAddress ipAddress;
+        ipAddress.ipset(dropZoneServerReq);
+        if (ipAddress.isNull())
             throw MakeStringException(ECLWATCH_INVALID_INPUT, "Invalid server %s specified.", dropZoneServerReq);
 
         double version = context.getClientVersion();
@@ -2999,12 +2999,11 @@ bool CFileSprayEx::onDropZoneFileSearch(IEspContext &context, IEspDropZoneFileSe
             if (server.isEmpty())
                 throw MakeStringException(ECLWATCH_INVALID_INPUT, "Invalid server for dropzone %s.", dropZoneName);
 
-            IpAddress ipAddr;
-            ipAddr.ipset(server.str());
-            if (ipAddr.ipequals(ipToMatch))
+            //Do string compare here because the server could be a pseudo-host name.
+            if (strieq(dropZoneServerReq, server))
             {
                 serverFound = true;
-                searchDropZoneFiles(context, ipAddr, directory.str(), nameFilter, files, filesFound);
+                searchDropZoneFiles(context, ipAddress, directory.str(), nameFilter, files, filesFound);
                 break;
             }
         }
@@ -3182,15 +3181,7 @@ bool CFileSprayEx::onDropZoneFiles(IEspContext &context, IEspDropZoneFilesReques
     {
         context.ensureFeatureAccess(FILE_SPRAY_URL, SecAccess_Read, ECLWATCH_FILE_SPRAY_ACCESS_DENIED, "Permission denied.");
 
-        IpAddress ipToMatch;
         const char* netAddress = req.getNetAddress();
-        if (!isEmptyString(netAddress))
-        {
-            ipToMatch.ipset(netAddress);
-            if (ipToMatch.isNull())
-                throw MakeStringException(ECLWATCH_INVALID_INPUT, "Invalid server %s specified.", netAddress);
-        }
-
         bool filesFromALinux = false;
         IArrayOf<IEspDropZone> dropZoneList;
         bool ECLWatchVisibleOnly = req.getECLWatchVisibleOnly();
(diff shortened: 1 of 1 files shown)


wangkx:
@ghalliday please review

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -3224,12 +3224,11 @@ bool CFileSprayEx::onDropZoneFiles(IEspContext &context, IEspDropZoneFilesReques
 
                 IpAddress ipAddr;
                 ipAddr.ipset(server.str());
```
ghalliday, line 3226: Why is the check for ipToMatch done via an ip address, instead of comparing the names?

wangkx: The names may not be same (like 'localhome', '.', or '127.0.0.1').

ghalliday:
@wangkx one question.  I'm not sure why any of the functions in the file compare using ip addresses instead of a textual comparison of the name that is passed in.  Can you remember why it is this way?

wangkx:
@ghalliday please see https://github.com/hpcc-systems/HPCC-Platform/pull/14381#discussion_r535378719.

esp/services/ws_fs/ws_fsService.cpp
ghalliday: But is that ever a problem, since for ecl watch the list has been obtained by calling a service that returns the entries from the environment.
The reason for raising it as an issue is I want to add dropzone targets that do not have associated ips.  The most natural way to support that would be to specify a pseudo-host name, but that will fail if all names are converted to ips.
@jakesmith @GordonSmith what are your thoughts?

GordonSmith: These ESP methods potentially get called from the client tools as well (not 100% if thie method is used specifically).

Suggestion:  Check if name matches via string compare - if not, fall back to IP4 compare.

jakesmith: I don't know why IEspDropZoneFilesRequest uses a netaddres which is currently an IP..
What calls this, passing an IP now?

agree would be better if names, even with bare metal hostnames where he IP could be fluid.

wangkx: > What calls this, passing an IP now?
It could be called by any client through ECLWatch or SOAP call.

wangkx: > Suggestion: Check if name matches via string compare - if not, fall back to IP4 compare.

Does it work? For example, we have to 2 dropzone targets in the list. The first dropzone target has a pseudo-host name. The one for the ipToMatch is the second dropzone target. When comparing it with the first dropzone target, the strings are not the same. And we cannot do the IP4 comparison because the pseudo-host name cannot be converted to ip.

GordonSmith: Do we allow two pseudo-host names to point to the same dropzone target?

wangkx: Is it possible to use some flag inside the pseudo-host names to identify them as pseudo-host?

ghalliday: It could check if the file path was a url, and avoid resolving it if so.  I am going to create a entry in the design discussion, because I think it is worth discussing more generally.

wangkx:
@ghalliday I made the changes of using string compare, including another place with similar problem. Please review.

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -2981,9 +2981,9 @@ bool CFileSprayEx::onDropZoneFileSearch(IEspContext &context, IEspDropZoneFileSe
         if (!directory.length())
             throw MakeStringException(ECLWATCH_INVALID_INPUT, "DropZone Directory not found for %s.", dropZoneName);
 
-        IpAddress ipToMatch;
-        ipToMatch.ipset(dropZoneServerReq);
-        if (ipToMatch.isNull())
+        IpAddress ipAddress;
+        ipAddress.ipset(dropZoneServerReq);
+        if (ipAddress.isNull())
```
ghalliday, line 2986: This check should also probably go, but that should wait for a later PR since searching dropzones that are implemented as blob storage needs more work before it will work.

ghalliday (approved):

ghalliday:
@richardkchapman please can you sanity check this please.

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -2999,12 +2999,11 @@ bool CFileSprayEx::onDropZoneFileSearch(IEspContext &context, IEspDropZoneFileSe
...
 
-            IpAddress ipAddr;
-            ipAddr.ipset(server.str());
-            if (ipAddr.ipequals(ipToMatch))
+            //Do string compare here because the server could be a pseudo-host name.
+            if (strieq(dropZoneServerReq, server))
             {
                 serverFound = true;
-                searchDropZoneFiles(context, ipAddr, directory.str(), nameFilter, files, filesFound);
+                searchDropZoneFiles(context, ipAddress, directory.str(), nameFilter, files, filesFound);
```
richardkchapman, line 3006: Feels like we should be passing dropZoneServerReq here - but perhaps that's for the phase 2 ?

richardkchapman:
@ghalliday One question

esp/services/ws_fs/ws_fsService.cpp
ghalliday: yes I think it should, but that can be in a separate PR - this is a function for listing files onj a drop zone, which isn't implemented for azure blobs yet

PR #14396, opened 2020-12-11
HPCC-25080 Add __IS_LITTLE_ENDIAN__ constant to ECL language

dcamper:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

## Testing:

Manual testing.

### DIFF
--- a/ecl/hql/hqllex.l
+++ b/ecl/hql/hqllex.l
@@ -1072,6 +1072,11 @@ __STAND_ALONE__     {
                         setupdatepos;
                         return (__STAND_ALONE__);
                     }
+__IS_LITTLE_ENDIAN__ {
+                        setupdatepos;
+                        returnToken.setExpr(createConstant(__BYTE_ORDER == __LITTLE_ENDIAN));
+                        return (BOOL_CONST);
+                    }
 __TARGET_PLATFORM__ { RETURNSYM(__TARGET_PLATFORM__); }
 
 {percent}{alphanumcolon}*{percent} {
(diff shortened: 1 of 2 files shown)


dcamper:
@ghalliday Please review.  Thanks!

ghalliday:
@dcamper although it is only a small change I'm not quite sure about it.

a) I think you can already swap bits safely with (unsigned8)TRANSFER((LITTLE_ENDIAN unsigned8)x, BIG_ENDIAN UNSIGNED8)

b) I was expecting __IS_LITTLE_ENDIAN__(type) rather than a flag for the system.  I suspect either implementation is fine though.

c) Are we ever going to run on a big endian system?  It seems little endian is now the defacto standard.

I'll think some more.

dcamper:
@ghalliday Thanks for taking a look.  You have a good point with (c) because, if LE is indeed the standard then everything else is moot.  If it is not standard, I was going after a simple compile-time constant that could be used for conditionally including code.  (a) works without the compile-time test but unless it has its own BE/LE tests -- and it may, I haven't checkd -- it will unnecessarily expensive.

Bottom line, please feel _entirely_ free to reject the Jira and PR.  The whole thing originated from logically considering a coding scenario, not from any actual user-provided problem.

dcamper:
I should also point out that (a) is a "force the integer to BE" but the purpose of my addition is to allow the ECL programmer to "switch byte order to the opposite of native order" using the explicit transfers.  That is a bit different.  But as before, if (c) holds then this whole thing should go away.

ghalliday:
(a)Should swap the byte order whatever the underlying representation.
Closing because I'm not convinced it is needed, or the best way to provide the functionality.  (Sorry!)

PR #14408, opened 2020-12-15
HPCC-25098 Spot invalid single colon logical filenames

jakesmith:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dali/base/dautils.cpp
+++ b/dali/base/dautils.cpp
@@ -417,7 +417,20 @@ void CDfsLogicalFileName::normalizeName(const char *name, StringAttr &res, bool 
         switch (c)
         {
             case '@': ct = s; break;
-            case ':': ct = nullptr; break;
+            case ':':
+            {
+                if (*(s+1) == ':')
+                {
+                    ++s;
+                    char nextc=*(s+1);
+                    if (!validFNameChar(nextc) && (!allowWild || ('*' != nextc) && ('?' != nextc) && ('{' != nextc)))
+                        throw makeStringExceptionV(-1, "Unexpected character '%c' following scope separate in logical name '%s' detected", *(s+1), name);
+                }
+                else
+                    throw makeStringExceptionV(-1, "Invalid single colon in logical name '%s' detected", name);
+                ct = nullptr;
+                break;
+            }
             case '?':
             case '*': wilddetected = true; break;
             case '~':
--- a/testing/regress/ecl/dynamicoptflag.ecl
+++ b/testing/regress/ecl/dynamicoptflag.ecl
@@ -34,7 +34,7 @@ count(p);
 output(p(f='NOT'));
 count(p(f='NOT'));
 
-p2 := PRELOAD(dataset(DYNAMIC(prefix + 'no:_such_file_again'), {string10 f}, FLAT, OPT), 2);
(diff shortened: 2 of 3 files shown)


jakesmith:
@mckellyln - please review

dali/base/dautils.cpp
```diff
@@ -417,7 +417,20 @@ void CDfsLogicalFileName::normalizeName(const char *name, StringAttr &res, bool
...
         {
             case '@': ct = s; break;
-            case ':': ct = nullptr; break;
+            case ':':
+            {
+                if (*(s+1) == ':')
+                {
+                    ++s;
+                    char nextc=*(s+1);
+                    if (!validFNameChar(nextc) && (!allowWild || ('*' != nextc) && ('?' != nextc) && ('{' != nextc)))
```
mckellyln, line 426: Can a valid name end in '::' ?  If so then I think we would incorrectly thorw here ?

mckellyln:
@jakesmith just one question.  Otherwise looks good.

testing/regress/ecl/dynamicoptflag.ecl
```diff
@@ -34,7 +34,7 @@ count(p);
 output(p(f='NOT'));
 count(p(f='NOT'));
 
-p2 := PRELOAD(dataset(DYNAMIC(prefix + 'no:_such_file_again'), {string10 f}, FLAT, OPT), 2);
+p2 := PRELOAD(dataset(DYNAMIC(prefix + 'no_such_file_again'), {string10 f}, FLAT, OPT), 2);
```
mckellyln, line 37: Should we add in a new regression test for this ?

mckellyln (commented):
Just a small question

richardkchapman:
What happens if you want to run a local standalone program that refers to (say) 'C:\AUTOEXEC.BAT' ?

testing/regress/ecl/dynamicoptflag.ecl
jakesmith: will extend unittests to test invlid paths like this

jakesmith:
>What happens if you want to run a local standalone program that refers to (say) 'C:\AUTOEXEC.BAT' ?

This code is with in the logical file handling helper CDfsLogicalFileName.
I think Roxie with direct physical names, must avoid using that, otherwise it would have already hit errors if the name had e.g. '/' or '\' in them.

To access a physical external file in other engines, it has to be encoded as a logical filename using the external file scope.
There is a helper function ExternalLogicalFileName, which can take a plain external form, and return the external logical name equivalent

dali/base/dautils.cpp
jakesmith: no, that would already have been rejected as invalid.

jakesmith:
@mckellyln - the regression suite has picked up an issue where there is valid case of single colon, i.e. ~foreign::10.7.8.9:7010::scope1

Need to special case that 'scope'.

jakesmith:
Closing for now. The code that parses lfn's walks over name multiple times, it's a bit silly and needs tidying up.
To correctly spot the form <type> (e.g. foreign) and that it can have different characters could be done with a kludge, but this code needs revisiting to fix properly.

PR #14411, opened 2020-12-15
HPCC-25101 Clean up some IPropertyTree interfaces

ghalliday:
also

* optimize a couple of functions
* always copy the value of an element when merging
* ensure first item in a list is processed as a list element

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jptree.cpp
+++ b/system/jlib/jptree.cpp
@@ -234,7 +234,7 @@ unsigned ChildMap::getHashFromElement(const void *e) const
     return elem.queryHash();
 }
 
-unsigned ChildMap::numChildren()
+unsigned ChildMap::numChildren() const
 {
     SuperHashIteratorOf<IPropertyTree> iter(*this);
     if (!iter.first()) return 0;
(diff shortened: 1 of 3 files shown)


ghalliday:
@jakesmith please review

system/jlib/jptree.cpp
```diff
@@ -3132,6 +3132,7 @@ void PTree::addLocal(size32_t l, const void *data, bool _binary, int pos)
         IPropertyTree *element1 = detach();
         array = new CPTArray();
         addingNewElement(*element1, ANE_APPEND);
+        static_cast<PTree *>(element1)->setOwner(array);
```
ghalliday, line 3135: Without this YAML generation and merging don't work correctly.  I suspect you may want a dynamic cast instead of a static cast....

jakesmith, line 3135: looks like a good fix, curious what code cause it to be exposed?

jakesmith:
@ghalliday - all looks sensible. 1 question from curiosity

jakesmith (approved):

system/jlib/jptree.cpp
ghalliday: Two things.  When you exported to YAML it generated the first item as a nested structure, and the second as a list.  E.g.

hosts:
  ip: x
hosts:
- ip: x

Also the mergeConfig code misinterpreted the structure.

jakesmith: ok thanks.

PR #14416, opened 2020-12-16
HPCC-24858 Allow dynamic logging manager configurations

timothyklemm:
Update the DESDL service to detect and use a logging manager defined in the
dynamic binding in addition to one defined in the static configuration. The
format of the dynamic configuration matches the format currently expected
from esp.xml.

A dynamically configured manager will always be used. A statically configured
manager will always be used in the absence of a dynamically configured
manager, meaning it persists in the presence of a dynamically configured
manager so it can take effect if the binding is republished without a
configured manager.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/logging/loggingmanager/loggingmanager.hpp
+++ b/esp/logging/loggingmanager/loggingmanager.hpp
@@ -82,7 +82,9 @@ class CLoggingManager : implements ILoggingManager, public CInterface
     bool oneTankFile = false, decoupledLogging = false, initialized = false;
     Owned<ILogFailSafe> logFailSafe;
     CLogContentFilter logContentFilter;
+    unsigned serviceMask = 0;
 
+    inline void setServiceMaskService(LOGServiceType service) { serviceMask |= (1 << service); }
     IEspLogAgent* loadLoggingAgent(const char* name, const char* dll, const char* type, IPropertyTree* cfg);
     bool updateLogImpl(IEspUpdateLogRequestWrap& req, IEspUpdateLogResponse& resp);
     bool saveToTankFile(IEspUpdateLogRequestWrap& req, CLogRequestInFile* reqInFile);
@@ -103,6 +105,7 @@ public:
     virtual bool init(IPropertyTree* cfg, const char* service);
 
     virtual IEspLogEntry* createLogEntry();
+    bool hasService(LOGServiceType service) const override;
     virtual bool updateLog(IEspContext* espContext, IEspUpdateLogRequestWrap& req, IEspUpdateLogResponse& resp);
     virtual bool updateLog(IEspLogEntry* entry, StringBuffer& status);
     virtual bool getTransactionSeed(StringBuffer& transactionSeed, StringBuffer& status);
--- a/esp/platform/pluginloader.hpp
+++ b/esp/platform/pluginloader.hpp
@@ -1,0 +1,184 @@
+/*##############################################################################
+
+    HPCC SYSTEMS software Copyright (C) 2020 HPCC Systems®.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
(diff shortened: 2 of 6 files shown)


esp/platform/pluginloader.hpp
```diff
@@ -0,0 +1,184 @@
...
+                entryPointName = m_entryPointDefault;
+        }
+        return (canFind ? find(libraryName, entryPointName) : nullptr);
+    }
+
+private:
+    using EntryPointMap = std::map<std::string, entry_point_t>;
+    struct Plugin
+    {
+        HINSTANCE library = nullptr;
```
mayx, line 145: Is the shared object freed at some point? Maybe better to use the "SharedObject" jlib class? Need to be careful not to free the library when its procedures are or will still be used.

esp/platform/pluginloader.hpp
```diff
@@ -0,0 +1,184 @@
...
+            if (!missing.isEmpty())
+                missing.append(" and ");
+            missing.append("entry-point function name");
+        }
+        if (!missing.isEmpty())
+            ERRLOG("bad plugin loading configuration - missing %s", missing.str());
+    }
+
+    template <typename instance_t>
+    instance_t* create(const IPTree& configuration, Creator<instance_t> creator)
```
mayx, line 75: I don't see the significance of having the "Creator" functor, which is simply a pass through of the entry_point_t, and entry_point_t is already known as soon as you declare m_loader. My question basically is, why do we need the functor?

mayx (commented):
@timothyklemm a few comments/questions...

esp/platform/pluginloader.hpp
timothyklemm: With this initial use case, the entry point takes no parameters and would be easy to call directly. Not all plugin entry points take no parameters. The functor hides entry point signature differences.

esp/platform/pluginloader.hpp
timothyklemm: Short answer is no. This is essentially a refactoring of existing code to be more generic, and the shared object was not being freed.

The shared object cannot be freed until all instances created by it are released. The public visibility of the entry point functions prevents this class from tracking all instances. Even if it could track instances, coupling the external destruction of a tracked instance to an embedded SharedObject instance could get very messy.

If the now static instance became a member of the service instance, each service instance could clean up after itself. The success of this approach could be dependent upon the order of deletion, which might not be entirely in the service instance's control.

How important is this existing issue and, if important enough to address now, do you have recommendations for safely tracking this relationship?

esp/platform/pluginloader.hpp
mayx: In this case it's fine to keep the functor for future extension.

esp/platform/pluginloader.hpp
timothyklemm: I've created HPCC-25116 to track this concern.

esp/logging/loggingmanager/loggingmanager.hpp
```diff
@@ -84,6 +84,7 @@ class CLoggingManager : implements ILoggingManager, public CInterface
     CLogContentFilter logContentFilter;
     unsigned serviceMask = 0;
 
+    void updateServiceMask(LOGServiceType service, bool value = true);
```
mayx, line 87: There's a recent discussion about avoiding using default params, can you remove it?

mayx (commented):
@timothyklemm one minor comment...

mayx (dismissed):
Looks good.

mayx:
@timothyklemm changes look good, please squash the commits...

mayx (approved):
Looks good.

timothyklemm:

PR #14429, opened 2021-01-05
HPCC-25130 Refactor code signing source to prepare for K8s related changes

shamser:
…anges

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [x] Security
  - [x] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:


shamser:
The change also prepares code for supporting other signing methods.

Also, the user provided fields (userid and passphrase) are now sanitized to secure against command injection.

ecl/hqlcpp/CMakeLists.txt
```diff
@@ -121,6 +121,7 @@ include_directories (
          ./../../common/thorhelper
          ./../../rtl/eclrtl
          ./../../rtl/include
+         ./../../system/codesigner
```
mayx, line 124: Better to use HPCC_SOURCE_DIR

ecl/hqlcpp/hqlres.cpp
```diff
@@ -174,13 +175,16 @@ void ResourceManager::addManifestFile(const char *filename, ICodegenContextCallb
...
     {
         try
         {
-            OwnedHqlExpr sig = checkSignature(fileContents.length(), fileContents.str());
-            useContents = stripSignature(strippedFileContents, fileContents).str();
-            isSigned = true;
+            StringBuffer signer, errmsg;
+            isSigned = queryCodeSigner().verifySignature(fileContents, signer, errmsg);
+            if (isSigned)
+                OwnedHqlExpr sig = createExprAttribute(_signed_Atom,createConstant(signer.str()));
```
mayx, line 185: What's the use of variable sig?

ecl/hqlcpp/hqlres.cpp
```diff
@@ -313,13 +317,16 @@ void ResourceManager::addManifestsFromArchive(IPropertyTree *archive, ICodegenCo
...
         {
             try
             {
-                OwnedHqlExpr sig = checkSignature(manifestContents.length(), manifestContents.str());
-                xml = stripSignature(strippedManifestContents, manifestContents).str();
-                isSigned = true;
+                StringBuffer signer, errmsg;
+                isSigned = queryCodeSigner().verifySignature(manifestContents, signer, errmsg);
+                if (isSigned)
+                    OwnedHqlExpr sig = createExprAttribute(_signed_Atom,createConstant(signer.str()));
```
mayx, line 327: Similar, what's the purpose of having variable sig?

system/codesigner/gpgcodesigner.cpp
```diff
@@ -0,0 +1,278 @@
...
+{
+    return gpgCodeSigner;
+}
+
+void GpgCodeSigner::initGpg(void)
+{
+    if (isGpgInstalled) return;
+
+    CriticalBlock block(crit);
+    StringBuffer cmd, output, errmsg;
```
mayx, line 38: isGpgInstalled should be checked again within critical block. Also strictly it should be an atomic variable.

system/codesigner/gpgcodesigner.cpp
```diff
@@ -0,0 +1,278 @@
...
+GpgCodeSigner gpgCodeSigner;
+
+extern jlib_decl ICodeSigner &queryGpgCodeSigner()
+{
+    return gpgCodeSigner;
+}
+
+void GpgCodeSigner::initGpg(void)
+{
+    if (isGpgInstalled) return;
```
mayx, line 35: Does the name of the variable reflect its use? Here we're just checking the version of the already installed gpg, not installing it. Maybe isInitialized or something like that is better?

system/codesigner/gpgcodesigner.cpp
```diff
@@ -0,0 +1,278 @@
...
+
+StringBuffer & GpgCodeSigner::sanitizeUserId(const char * userId, StringBuffer & sanitizedUserId)
+{
+    sanitizedUserId.clear();
+    sanitizedUserId.ensureCapacity(strlen(userId));
+    const char *p = userId;
+    while (*p && *p==' ') ++p;
+    while (*p)
+    {
+        if (isalnum(*p) || *p==' ' || *p=='.' || *p=='@' || *p =='<' || *p=='>')
```
mayx, line 255: This seems to be overly restrictive, chars like -, _ should be accepted. Would it be better to be exclusive instead of inclusive, and escape as necessary?

system/codesigner/gpgcodesigner.cpp
```diff
@@ -0,0 +1,278 @@
...
+    };
+    return sanitizedUserId;
+}
+
+StringBuffer & GpgCodeSigner::sanitizePassphrase(const char * passphrase, StringBuffer & sanitizedPassphrase)
+{
+    sanitizedPassphrase.clear();
+    sanitizedPassphrase.ensureCapacity(strlen(passphrase));
+    const char *p = passphrase;
+    // Note: strictly speaking any characters should be valid for passphrases. However, fully
```
mayx, line 267: Would it be better to escape the chars that could possibly cause injection? Another option might be to include the whole argument inside quotes, and escape the quote chars maybe?

mayx (commented):
@shamser overall looks good, a few comments/questions...

system/codesigner/gpgcodesigner.cpp
shamser: I see your point.  I see in the original code, you simply rejected any user id containing a quote. That seems like a decent solution: we could reject all user ids that have a quote character, control character or too long (say longer than 2000 characters)?

system/codesigner/gpgcodesigner.cpp
shamser: The user id is already quoted.  We can't quote the passphrases.  We could either filter our passphrases containing quotes, control characters or too long?

system/codesigner/gpgcodesigner.cpp
shamser: I've renamed the field.

ecl/hqlcpp/hqlres.cpp
shamser: This is not necessary. Removed.

ecl/hqlcpp/hqlres.cpp
shamser: This is not necessary. Removed.

system/codesigner/gpgcodesigner.cpp
shamser: Done.

shamser:
@mayx Please can you review the changes.

system/codesigner/gpgcodesigner.cpp
mayx: Yes that should work.

system/codesigner/gpgcodesigner.cpp
mayx: I took a closer look. Passphrase is directly streamed to std input and it doesn't appear any where on the command line, so we don't need to filter it. The sanitize method silently drops chars, that could be a cause of confusion for users. So basically we can just leave the passphrase as it's passed in I think.

mayx:
@shamser a couple of comments about userid and passphrase sanitization. The rest of the changes look good.

system/codesigner/gpgcodesigner.cpp
shamser: Removed.

shamser:
@mayx Should we merge now?

mayx (dismissed):
@shamser looks good.

ecl/hql/hqlmanifest.cpp
```diff
@@ -34,12 +35,12 @@ class ResourceManifest : public CInterface
...
             StringBuffer body;
             // Check for signature
-            if (startsWith(fileContents, "-----BEGIN PGP SIGNED MESSAGE-----"))
+            if (queryCodeSigner().hasSignature(fileContents))
             {
                 // Note - we do not check the signature here - we are creating an archive, and typically that means we
                 // are on the client machine, while the signature can only be checked on the server where the keys are installed.
-                stripSignature(body, fileContents);
-                xml = body.str();
+                StringBuffer  errmsg;
```
ghalliday, line 42: minor: appears to be unused.

ecl/hql/hqlparse.cpp
```diff
@@ -548,7 +549,10 @@ void HqlLex::checkSignature(const attribute & dummyToken)
         return;
     try
     {
-        yyParser->gpgSignature.setown(::checkSignature(text->length(), text->getText()));
+        StringBuffer signer, errmsg;
+        if (!queryCodeSigner().verifySignature(text->getText(), signer, errmsg))
+            throw makeStringExceptionV(0, "gpg: signature not verified");
```
ghalliday, line 554: should this use errmsg?

ecl/hqlcpp/hqlres.cpp
```diff
@@ -174,13 +175,13 @@ void ResourceManager::addManifestFile(const char *filename, ICodegenContextCallb
...
-    if (startsWith(fileContents, "-----BEGIN PGP SIGNED MESSAGE-----"))
+    if (queryCodeSigner().hasSignature(fileContents))
     {
         try
         {
-            OwnedHqlExpr sig = checkSignature(fileContents.length(), fileContents.str());
-            useContents = stripSignature(strippedFileContents, fileContents).str();
-            isSigned = true;
+            StringBuffer signer, errmsg;
+            isSigned = queryCodeSigner().verifySignature(fileContents, signer, errmsg);
```
ghalliday, line 183: no longer throws if there is an error.

ecl/hqlcpp/hqlres.cpp
```diff
@@ -313,13 +314,13 @@ void ResourceManager::addManifestsFromArchive(IPropertyTree *archive, ICodegenCo
...
-        if (startsWith(xml, "-----BEGIN PGP SIGNED MESSAGE-----"))
+        if (queryCodeSigner().hasSignature(xml))
         {
             try
             {
-                OwnedHqlExpr sig = checkSignature(manifestContents.length(), manifestContents.str());
-                xml = stripSignature(strippedManifestContents, manifestContents).str();
-                isSigned = true;
+                StringBuffer signer, errmsg;
+                isSigned = queryCodeSigner().verifySignature(manifestContents, signer, errmsg);
```
ghalliday, line 322: no longer throws?

system/codesigner/gpgcodesigner.cpp
```diff
@@ -0,0 +1,250 @@
...
+    {
+        str = strchr(str, c);
+        if (!str)
+            break;
+        str++;
+    }
+    return str;
+}
+
+StringArray &GpgCodeSigner::getUserIds(StringArray &userIds, StringBuffer &errmsg)
```
ghalliday, line 171: Not sure why errmsg is passed as a parameter - it will never be used by the caller.

system/codesigner/gpgcodesigner.cpp
```diff
@@ -0,0 +1,250 @@
...
+    virtual StringArray &getUserIds(StringArray & userIds, StringBuffer & errmsg) override;
+private:
+    void initGpg(void);
+    bool getKeyGrip(const char * user, StringBuffer & keygrip);
+    void clearPassphrase(const char * key);
+    bool isGpgV1 = false;
+    std::atomic<bool> isInitialized{false};
+    static constexpr const char* signatureMsgHeader = "-----BEGIN PGP SIGNED MESSAGE-----";
+    static constexpr const char* signatureBegin = "-----BEGIN PGP SIGNATURE-----";
+    CriticalSection crit;
```
ghalliday, line 22: clearer if this is before the static members

ecl/hql/hqlutil.hpp
```diff
@@ -251,30 +251,6 @@ extern HQL_API IHqlExpression * createTransformForField(IHqlExpression * field,
 extern HQL_API IHqlExpression * convertScalarToRow(IHqlExpression * value, ITypeInfo * fieldType);
 extern HQL_API bool splitResultValue(SharedHqlExpr & dataset, SharedHqlExpr & attribute, IHqlExpression * value);
 
-/**
- * Check whether GPG signed file is valid.
```
ghalliday, line 255: Any benefit in these comments moving to the methods in the interface?

system/codesigner/gpgcodesigner.cpp
```diff
@@ -0,0 +1,250 @@
...
+};
+
+GpgCodeSigner gpgCodeSigner;
```
ghalliday, line 25: trivial: better if it is static

system/codesigner/gpgcodesigner.cpp
```diff
@@ -0,0 +1,250 @@
+#include "jlib.hpp"
+#include "jexcept.hpp"
+#include "jlog.hpp"
+#include "gpgcodesigner.hpp"
+#include "atomic"
+
+class GpgCodeSigner : implements ICodeSigner
+{
+    virtual bool sign(const char * text, const char * userId, const char * passphrase, StringBuffer & errmsg, StringBuffer & signedText) override;
```
ghalliday, line 9: a slightly pedantic comment which you are welcome to ignore:
these functions should really be const since logically they do not modify the contents of the code signing object, and isGpgv1 and isInitialised should be mutable.

system/codesigner/gpgcodesigner.cpp
```diff
@@ -0,0 +1,250 @@
...
+    virtual bool verifySignature(const char * text, StringBuffer & signer, StringBuffer & errmsg) override;
+    virtual bool hasSignature(const char * text) const override;
+    virtual StringBuffer &stripSignature(const char * text, StringBuffer & unsignedText) const override;
+    virtual StringArray &getUserIds(StringArray & userIds, StringBuffer & errmsg) override;
+private:
+    void initGpg(void);
+    bool getKeyGrip(const char * user, StringBuffer & keygrip);
+    void clearPassphrase(const char * key);
+    bool isGpgV1 = false;
```
ghalliday, line 18: ultra picky: this should also be atomic to avoid and tiny race condition on weak ordering processors.  (Another thread may see the assignment to isInitialised before the update to isGPgV1)

ghalliday (commented):
@shamser in general I like the change - it separates the code much more cleanly.
A few minor comments - mainly where I think exceptions were previously thrown, but aren't any more.

system/codesigner/gpgcodesigner.cpp
shamser: initGpg is non-const.  If sign is const then it wont be able to initGpg.  (In another PR, I use initGpg to also import keys from secrets so that keys are not unnecessarily imported, slowing down compilation.)

shamser:
@ghalliday

ecl/hql/hqlparse.cpp
```diff
@@ -549,9 +549,10 @@ void HqlLex::checkSignature(const attribute & dummyToken)
         return;
     try
     {
-        StringBuffer signer, errmsg;
-        if (!queryCodeSigner().verifySignature(text->getText(), signer, errmsg))
-            throw makeStringExceptionV(0, "gpg: signature not verified");
+        StringBuffer signer;
+        if (!queryCodeSigner().verifySignature(text->getText(), signer))
+                throw makeStringExceptionV(0, "Code sign verify: signature not verified");
```
ghalliday, line 554: trivial: indentation

ecl/hqlcpp/hqlres.cpp
```diff
@@ -174,13 +175,15 @@ void ResourceManager::addManifestFile(const char *filename, ICodegenContextCallb
...
     {
         try
         {
-            OwnedHqlExpr sig = checkSignature(fileContents.length(), fileContents.str());
-            useContents = stripSignature(strippedFileContents, fileContents).str();
-            isSigned = true;
+            StringBuffer signer;
+            isSigned = queryCodeSigner().verifySignature(fileContents, signer);
+            if (!isSigned)
+                throw makeStringExceptionV(0, "Code sign verify: signature not verified");
```
ghalliday, line 185: minor: could use one of the new error codes instead of 0

ghalliday (commented):
@shamser looks good.  Two minor comments.  Please can you squash

ecl/hql/hqlparse.cpp
shamser: I've changed verifySignature to remove errmsg argument.  It will now throw exceptions instead of return error messages in errmsg.

ecl/hql/hqlmanifest.cpp
shamser: removed

ecl/hqlcpp/hqlres.cpp
shamser: I've changed verifySignature to throw exceptions instead of return error messages in errmsg.

ecl/hqlcpp/hqlres.cpp
shamser: I've changed verifySignature to throw exceptions instead of return error messages in errmsg.

system/codesigner/gpgcodesigner.cpp
shamser: I've changed verifySignature to throw exceptions instead of returning error messages in errmsg.

system/codesigner/gpgcodesigner.cpp
shamser: done.

system/codesigner/gpgcodesigner.cpp
shamser: done.

system/codesigner/gpgcodesigner.cpp
shamser: done

ghalliday (approved):

PR #14432, opened 2021-01-05
HPCC-25133 Decrease hard workunit result limit to 100MB

jakesmith:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/common/thorhelper/thorcommon.hpp
+++ b/common/thorhelper/thorcommon.hpp
@@ -29,7 +29,6 @@
 #include "thorhelper.hpp"
 #include "thorxmlwrite.hpp"
 
-static unsigned const defaultDaliResultOutputMax = 2000; // MB
 static unsigned const defaultDaliResultLimit = 10; // MB
 static unsigned const defaultMaxCsvRowSize = 10; // MB
 
--- a/ecl/hqlcpp/hqlcerrors.hpp
+++ b/ecl/hqlcpp/hqlcerrors.hpp
@@ -225,6 +225,7 @@
 #define HQLERR_CannotInterpretRecord            4213
 #define HQLERR_NoVirtualAndAlien                4214
 #define HQLERR_CannotFilterLimitInsideActivity  4215
+#define HQLERR_OutputLimitMaxExceeded           4216
 
 //Warnings....
 #define HQLWRN_PersistDataNotLikely             4500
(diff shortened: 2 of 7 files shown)


common/thorhelper/thorcommon.hpp
```diff
@@ -29,7 +29,7 @@
 #include "thorhelper.hpp"
 #include "thorxmlwrite.hpp"
 
-static unsigned const defaultDaliResultOutputMax = 2000; // MB
+static unsigned const defaultDaliResultOutputMax = 100; // MB
```
richardkchapman, line 32: What happens when the limit is specified higher in a #option?

jakesmith: It fires an exception, e.g.:
                throw MakeStringException(0, "Dali result outputs are restricted to a maximum of %d MB, the current limit is %d MB. A huge dali result usually indicates the ECL needs altering.", defaultDaliResultOutputMax, defaultDaliResultLimit);

jakesmith:

common/thorhelper/thorcommon.hpp
richardkchapman: At compile time? Or at runtime?

(checking the value at compile time would be friendlier, if practical)

jakesmith: The existing checks are at runtime in hthor/roxie/thor.
I think still need runtime check, because limit may come from ECL MAXSIZE(<expr>), which could be runtime dependent.
But I've added a compile time check also.

@ghalliday - can you check, and if constant in eclhelper.hpp is in a suitable place?

ghalliday: The changes look fine.  I'm not sure where that constant should go.  It is a bit out of place in eclhelper.hpp.  I suspect the original place might be better - hqlcpp does include other things from thorhelper already.  I'm not sure it is worth changing though.
I have one example (kel11.eclxml) that sets that options, and it sets it to 1000.  Thant means this change is likely to break that query.

richardkchapman: We are deliberately breaking that query... before that query breaks Dali

ghalliday: yes, just warning that it looks like it will be hit.
@jakesmith needs squashing

richardkchapman: Do you know who we need to warn?

jakesmith: Apart from redbook, I can send an email out to ECLUsers@risk.lexisnexis.com once this is merged/tagged into a build?

jakesmith:
@ghalliday - squashed and rebased to latest candidate-7.12.x

ghalliday (approved):

PR #14440, opened 2021-01-06
HPCC-24837 Report Activity Information for containerized HPCC

wangkx:
1. Add new getContainerClusterInfo() which is similar to the
getEnvironmentClusterInfo() for bare metal environment. For
now, the API only collects the cluster information needed for
reporting the Activity Information.
2. Call the getContainerClusterInfo() to get the cluster
information.
3. Make other changes for reporting the Activity Information
for containerized HPCC.

Signed-off-by: wangkx kevin.wang@lexisnexis.com<kevin.wang@lexisnexis.com>

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/services/ws_smc/ws_smcService.cpp
+++ b/esp/services/ws_smc/ws_smcService.cpp
@@ -237,12 +237,16 @@ struct CActiveWorkunitWrapper: public CActiveWorkunit
 
 void CActivityInfo::createActivityInfo(IEspContext& context)
 {
+    CConstWUClusterInfoArray clusters;
+#ifndef _CONTAINERIZED
     Owned<IEnvironmentFactory> factory = getEnvironmentFactory(true);
     Owned<IConstEnvironment> env = factory->openEnvironment();
 
-    CConstWUClusterInfoArray clusters;
     Owned<IPropertyTree> envRoot= &env->getPTree();
     getEnvironmentClusterInfo(envRoot, clusters);
+#else
+    getContainerWUClusterInfo(clusters);
+#endif
 
     try
     {
@@ -261,7 +265,11 @@ void CActivityInfo::createActivityInfo(IEspContext& context)
     IPropertyTree* serverStatusRoot = connStatusServers->queryRoot();
 
     readTargetClusterInfo(clusters, serverStatusRoot);
+#ifndef _CONTAINERIZED
     readActiveWUsAndQueuedWUs(context, envRoot, serverStatusRoot);
+#else
+    readActiveWUsAndQueuedWUs(context, nullptr, serverStatusRoot);
+#endif
 
     timeCached.setNow();
 }
@@ -322,9 +330,13 @@ void CActivityInfo::readTargetClusterInfo(IConstWUClusterInfo& cluster, IPropert
 
     if (serverStatusRoot)
     {
+#ifndef _CONTAINERIZED
         smcQueue->foundQueueInStatusServer = findQueueInStatusServer(serverStatusRoot, statusServerName.str(), targetCluster->queueName.get());
         if (!smcQueue->foundQueueInStatusServer)
             targetCluster->clusterStatusDetails.appendf("Cluster %s not listening for workunits; ", clusterName.str());
+#else
+        smcQueue->foundQueueInStatusServer = true; //Server is launched dynamically.
+#endif
     }
 
     targetCluster->serverQueue.notFoundInJobQueues = !readJobQueue(targetCluster->serverQueue.queueName.str(), targetCluster->wuidsOnServerQueue, targetCluster->serverQueue.queueState, targetCluster->serverQueue.queueStateDetails);
@@ -418,8 +430,11 @@ void CActivityInfo::readActiveWUsAndQueuedWUs(IEspContext& context, IPropertyTre
     readRunningWUsAndJobQueueforOtherStatusServers(context, serverStatusRoot);
     //TODO: add queued WUs for ECLCCServer/ECLServer here. Right now, they are under target clusters.
 
+#ifndef _CONTAINERIZED
     getDFUServersAndWUs(context, envRoot, serverStatusRoot);
     getDFURecoveryJobs();
+    //For containerized HPCC, we do not know how to find out DFU Server queues, as well as running DFU WUs, for now.
+#endif
 }
 
 void CActivityInfo::readRunningWUsOnStatusServer(IEspContext& context, IPropertyTree* serverStatusRoot, WsSMCStatusServerType statusServerType)
--- a/esp/smc/SMCLib/TpWrapper.cpp
+++ b/esp/smc/SMCLib/TpWrapper.cpp
@@ -25,7 +25,7 @@
 #include "workunit.hpp"
 #include "exception_util.hpp"
 #include "portlist.h"
-
+#include "daqueue.hpp"
 
 const char* MSG_FAILED_GET_ENVIRONMENT_INFO = "Failed to get environment information.";
 
@@ -2050,4 +2050,145 @@ extern TPWRAPPER_API IStringIterator* getContainerTargetClusters(const char* pro
         ret->append_unique(targetName);
     }
     return ret.getClear();
+}
+
+class CContainerWUClusterInfo : public CSimpleInterfaceOf<IConstWUClusterInfo>
+{
+    StringAttr name;
+    StringAttr serverQueue;
+    StringAttr agentQueue;
+    StringAttr thorQueue;
+    ClusterType platform;
+    unsigned clusterWidth;
+
+public:
+    CContainerWUClusterInfo(const char* _name, const char* type, unsigned _clusterWidth)
+        : name(_name), clusterWidth(_clusterWidth)
+    {
+        StringBuffer queue;
+        if (strieq(type, "thor"))
+        {
+            thorQueue.set(getClusterThorQueueName(queue.clear(), name));
+            platform = ThorLCRCluster;
+        }
+        else if (strieq(type, "roxie"))
+        {
+            agentQueue.set(getClusterEclAgentQueueName(queue.clear(), name));
+            platform = RoxieCluster;
+        }
+        else
+        {
+            agentQueue.set(getClusterEclAgentQueueName(queue.clear(), name));
+            platform = HThorCluster;
+        }
+
+        serverQueue.set(getClusterEclCCServerQueueName(queue.clear(), name));
+    }
+
+    virtual IStringVal& getName(IStringVal& str) const override
+    {
+        str.set(name.get());
+        return str;
+    }
+    virtual IStringVal& getAgentQueue(IStringVal& str) const override
+    {
+        str.set(agentQueue);
+        return str;
+    }
+    virtual IStringVal& getServerQueue(IStringVal& str) const override
+    {
+        str.set(serverQueue);
+        return str;
+    }
+    virtual IStringVal& getThorQueue(IStringVal& str) const override
+    {
+        str.set(thorQueue);
+        return str;
+    }
+    virtual ClusterType getPlatform() const override
+    {
+        return platform;
+    }
+    virtual unsigned getSize() const override
+    {
+        return clusterWidth;
+    }
+    virtual bool isLegacyEclServer() const override
+    {
+        return false;
+    }
+    virtual IStringVal& getScope(IStringVal& str) const override
+    {
+        UNIMPLEMENTED;
+    }
+    virtual unsigned getNumberOfSlaveLogs() const override
+    {
+        UNIMPLEMENTED;
+    }
+    virtual IStringVal & getAgentName(IStringVal & str) const override
+    {
+        UNIMPLEMENTED;
+    }
+    virtual IStringVal & getECLSchedulerName(IStringVal & str) const override
+    {
+        UNIMPLEMENTED;
+    }
+    virtual const StringArray & getECLServerNames() const override
+    {
+        UNIMPLEMENTED;
+    }
+    virtual IStringVal & getRoxieProcess(IStringVal & str) const override
+    {
+        UNIMPLEMENTED;
+    }
+    virtual const StringArray & getThorProcesses() const override
+    {
+        UNIMPLEMENTED;
+    }
+    virtual const StringArray & getPrimaryThorProcesses() const override
+    {
+        UNIMPLEMENTED;
+    }
+    virtual const SocketEndpointArray & getRoxieServers() const override
+    {
+        UNIMPLEMENTED;
+    }
+    virtual const char *getLdapUser() const override
+    {
+        UNIMPLEMENTED;
+    }
+    virtual const char *getLdapPassword() const override
+    {
+        UNIMPLEMENTED;
+    }
+    virtual unsigned getRoxieRedundancy() const override
+    {
+        UNIMPLEMENTED;
+    }
+    virtual unsigned getChannelsPerNode() const override
+    {
+        UNIMPLEMENTED;
+    }
+    virtual int getRoxieReplicateOffset() const override
+    {
+        UNIMPLEMENTED;
+    }
+    virtual const char *getAlias() const override
+    {
+        UNIMPLEMENTED;
+    }
+};
+
+extern TPWRAPPER_API unsigned getContainerWUClusterInfo(CConstWUClusterInfoArray& clusters)
+{
+    Owned<IPropertyTreeIterator> queues = queryComponentConfig().getElements("queues");
+    ForEach(*queues)
+    {
+        IPropertyTree& queue = queues->query();
+        Owned<IConstWUClusterInfo> cluster = new CContainerWUClusterInfo(queue.queryProp("@name"),
+            queue.queryProp("@type"), (unsigned) queue.getPropInt("@width", 1));
+        clusters.append(*cluster.getClear());
+    }
+
+    return clusters.ordinality();
 }
--- a/esp/smc/SMCLib/TpWrapper.hpp
+++ b/esp/smc/SMCLib/TpWrapper.hpp
@@ -211,6 +211,9 @@ private:
 };
 
 extern TPWRAPPER_API ISashaCommand* archiveOrRestoreWorkunits(StringArray& wuids, IProperties* params, bool archive, bool dfu);
+
+extern TPWRAPPER_API unsigned getContainerWUClusterInfo(CConstWUClusterInfoArray& clusters);
+
 extern TPWRAPPER_API IStringIterator *getContainerTargetClusters(const char* processType, const char* processName);
 
 #endif //_ESPWIZ_TpWrapper_HPP__


wangkx:
@jakesmith please review

jakesmith:
@wangkx - I have looked in detail, but my feeling is that unless it's a temporary stop-gap, we should try to avoid touching environment.* - since the environment is bare-metal only, with containerized mode using a separate configuration paradigm.
And I suspect the IConstWUClusterInfo and the other environment.* interfaces are not a good fit going forward.

In the container world each engine is a 'target' (simply referred to as a queue), i.e. those generated into queues list.
A EclCCServer will listen to all queues, *unless* it is told to specifically listen to some, via it's "listen" property.

As I say, I haven't looked closely, and it may be that your changes fit the new paradigm into the old target/cluster paradigm, which may be ok for now.

I'll have a go at building a container setup with this PR tomorrow, and use a yaml configuration and see if it reflects the configuration as I'd expect.

Another issue with trying to marry the bare-metal view and cloud view I suspect, is that we should not be exposing or highlighting component IP's, which spin up temporarily.

In the longer term, we probably want a separate view of activity for container mode.

wangkx:
@jakesmith The IConstWUClusterInfo is borrowed to store the queue information read from the ConfigMap (which is based on the yaml configuration). With the minimal changes in this PR, both running WUs and queued WUs can be shown in the existing ECLWatch Activity page. We may say that this is a temporary stop-gap fix before redesigning the whole Activity display. Shall we redesign it now?

jakesmith:
>We may say that this is a temporary stop-gap fix before redesigning the whole Activity display. Shall we redesign it now?

@wangkx - it may be fine for now. Let me look more closely, and play with it (tomorrow).

richardkchapman:
See https://track.hpccsystems.com/browse/HPCC-23548 - not sure we want to be adding new code in environment.cpp at the moment. Is there somewhere else appropriate to add these functions?

common/environment/environment.cpp
```diff
@@ -3215,3 +3215,45 @@ unsigned getEnvironmentThorClusterNames(StringArray &thorNames, StringArray &gro
...
+    {
+        Owned<IPropertyTree> thor = createPTree(name, ipt_caseInsensitive);
+        thors.append(*thor.getClear());
+    }
+    else if (strieq(type, "roxie"))
+    {
+        roxie.setown(createPTree(name, ipt_caseInsensitive));
+    }
+
+    IArrayOf<IPropertyTree> eclServers;
```
jakesmith, line 3235: how are the eclccserver picked up show in container mode?

jakesmith (commented):
I've run with this PR, and from a testing point of view.
I see hthor, thor and roxie(workunit) type jobs in the queues.
I don't however see any pending compile (eclccserver) jobs, though I checked and they were on the queue.

common/environment/environment.cpp
```diff
@@ -3215,3 +3215,45 @@ unsigned getEnvironmentThorClusterNames(StringArray &thorNames, StringArray &gro
...
+    return new CEnvironmentClusterInfo(name, nullptr, nullptr, agent, eclServers, false, nullptr, thors, roxie);
+}
+
+extern ENVIRONMENT_API unsigned getContainerClusterInfo(const char *processType, CConstWUClusterInfoArray &clusters)
+{
+    Owned<IPropertyTreeIterator> queues = queryComponentConfig().getElements("queues");
+    ForEach(*queues)
+    {
+        IPropertyTree &queue = queues->query();
+        if (!isEmptyString(processType))
```
jakesmith, line 3247: I think processType is always empty at the moment, i.e. there's never any filtering, correct?

jakesmith:
@wangkx  - These are quite small changes, that force the new configuration into holders (ConstWUClusterInfo) that the legacy system uses/understands/renders.
However, as I and Richard say, we should avoid extending environment.* - so if we are go keep this as a stop-gap, we should try to abstract it out of environment.*.

I wonder though, whether it wouldn't be simpler to work on a new clean esp service that gets a clean and simpler result for what's running/queued - and as mentioned before - it would be worth rethinking whether the /Servers/Server/Status is the right approach anymore - it may be - but could the jobs in the stats required to be fetched with a quick filter from the workunit interface directly (might depend on the update frequency).

If this stop-gap is nearly there, and can be moved out of environment, then probably worth completing a stop-gap, but open a couple of new JIRA's 1) rework the activity service, 2) to render the new view in eclwatch.

common/environment/environment.cpp
```diff
@@ -3215,3 +3215,45 @@ unsigned getEnvironmentThorClusterNames(StringArray &thorNames, StringArray &gro
     }
     return thorNames.ordinality();
 }
+
+IConstWUClusterInfo *getContainerTargetClusterInfo(IPropertyTree *queue)
+{
+    IArrayOf<IPropertyTree> thors;
+    Owned<IPropertyTree> roxie = nullptr;
```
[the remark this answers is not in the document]
wangkx: I will remove it.

common/environment/environment.cpp
wangkx: Yes for this PR. I thought the function may be used for other calls later.

common/environment/environment.cpp
wangkx: Good question. I haven't found out a way to pick up eclccserver in container mode.

wangkx:
Richard and Jake, thanks for the comments. I am working on this stop-gap because many unknown for a new design: where to read the running WUs, where to read queued WUs, new ECLWatch UI structure, etc.
For this stop-gap, I may create a new structure similar to the ConstWUClusterInfo inside ESP SMC lib and remove the code from environment. I did not pay an attention on pending compile (eclccserver) jobs because most compile jobs run too quick to catch. @jakesmith how do you test it (how to make the job stay long)?

common/environment/environment.cpp
jakesmith: It depends what we want the view to look like.
The topology is different from a legacy system, where a target defined the eclccserver (and agent and thor etc.) that were to be associated with it.

The primary view/info for the activity page we probably want, is to see the targets (the hthor, thor and roxie queues), and the state of the workunits that have been submitted to them.
Each target queue, has several queues associated with it. e.g. a thor has an compile queue (eclccserver), a workflow queue (eclagent), and a thor agent queue (that actually spins up thor instances).
Knowing the state of a workunit, as it transitions through these queues/other states is useful, it's secondary information, that should probably be accessed or linked to under the main list of active workunits under each target.

I think the esp activity service should return a list of the queue targets, with the workunits associated with them, and some state information.

jakesmith:
>how do you test it (how to make the job stay long)?

@wangkx - I reduced maxJobs=1, and submitted lots of jobs to ensure they were taking time to compile.
You could also change useChildProcesses=false, which will cause it to launch a pod instead a process each time, which will dramatically slow it down.

But as I say above, primarily, an overview from the targets point of view is likely what we want, i.e. a list per queue (from the generated yaml 'queues'). Then with state info. attributed to each workunit, and I suspect for the 1st cut, the workunit @state is enough, i.e. it is enough to know it is 'submitted' or 'compiling' etc. Later that state info. can be expanded upon if necessary to give more details, e.g. about which of the multiple queues is blocking it etc.

wangkx:
The new esp activity service: The queryComponentConfig().getElements("queues") can return a list of the queue targets. To get running WUs, how about use the /Status/Servers/Server/Workunits? For queued WUs, is there a way to loop through a queue list to get the WUs? We may use WU's ClusterName attribute to link the WU with its queue target. @jakesmith please comment.

jakesmith:
> To get running WUs, how about use the /Status/Servers/Server/Workunits?
> For queued WUs, is there a way to loop through a queue list to get the WUs? 

for both, any workunits with given state(s), I'd look at what the workunit iterator interfaces can do now, look at how they can be adapted if not suitable, and/or look at their efficiency to filter based on state.
@wangkx

wangkx:
> for both, any workunits with given state(s), I'd look at what the workunit iterator interfaces can do now, look at how they can be adapted if not suitable, and/or look at their efficiency to filter based on state.

I will try it. I think I can use '@state=running' to filter the running WUs. Which filter should be used for queuing WUs? @jakesmith

jakesmith:
> I will try it. I think I can use '@State=running' to filter the running WUs. Which filter should be used for queuing WUs? 

@wangkx - if a workunit is queued, with no component running or picking it up, it should be in a submitted state.

wangkx:
The code has been modified based on the review. Will test for 2 thor's and queuing in eclcc.

wangkx:
I just tested the compiling jobs. All of them are list under a target (because of the @ClusterName of the WU xml). Their states are compiling or submitted.

wangkx:
I did test a 2 thor's environment and the active WUs are shown under each thor. 
@jakesmith any change should I make for the eclccserver job (or others)?

jakesmith:
> I did test a 2 thor's environment and the active WUs are shown under each thor.
> @jakesmith any change should I make for the eclccserver job (or others)?

@wangkx - your changes in this PR to eclccserver (adding a Server/Status entry), is that being used/picked up?

esp/smc/SMCLib/TpWrapper.cpp
```diff
@@ -2050,4 +2050,75 @@ extern TPWRAPPER_API IStringIterator* getContainerTargetClusters(const char* pro
         ret->append_unique(targetName);
     }
     return ret.getClear();
+}
+
+class CContainerWUClusterInfo : public CSimpleInterfaceOf<IConstContainerWUClusterInfo>
```
jakesmith, line 2055: it's ok to use the existing interface (IConstWUClusterInfo) for now, as long as it's not using and/or extending the environment implementation.
This implementation should implement IConstWUClusterInfo, but with most of the methods unimplemented.
Then all the rest of the code is the service is common.

esp/smc/SMCLib/TpWrapper.cpp
```diff
@@ -2050,4 +2050,75 @@ extern TPWRAPPER_API IStringIterator* getContainerTargetClusters(const char* pro
...
+        else
+        {
+            agentQueue.set(getClusterEclAgentQueueName(queue.clear(), name));
+            platform = HThorCluster;
+        }
+
+        serverQueue.set(getClusterEclCCServerQueueName(queue.clear(), name));
+    }
+
+    IStringVal& getName(IStringVal& str) const
```
jakesmith, line 2086: trivial: all virtual overrides should have virtual and override declarations

esp/smc/SMCLib/TpWrapper.hpp
```diff
@@ -211,6 +211,19 @@ class CCluster
...
 extern TPWRAPPER_API ISashaCommand* archiveOrRestoreWorkunits(StringArray& wuids, IProperties* params, bool archive, bool dfu);
+
+interface IConstContainerWUClusterInfo : extends IInterface
+{
+    virtual IStringVal& getName(IStringVal& str) const = 0;
+    virtual IStringVal& getThorQueue(IStringVal& str) const = 0;
+    virtual IStringVal& getAgentQueue(IStringVal& str) const = 0;
+    virtual IStringVal& getServerQueue(IStringVal& str) const = 0;
+    virtual ClusterType getPlatform() const = 0;
+};
```
jakesmith, line 222: See other comment, re. using and re-implementing IConstWUClusterInfo.

esp/smc/SMCLib/TpWrapper.hpp
```diff
@@ -211,6 +211,19 @@ class CCluster
...
+typedef IArrayOf<IConstContainerWUClusterInfo> CConstContainerWUClusterInfoArray;
```
jakesmith, line 223: would be okay to use existing CConstWUClusterInfoArray (IArrayOf<IConstWUClusterInfo>) for this stop-gap implementation too.

jakesmith (changes requested):
@wangkx - please see comments

esp/services/ws_smc/ws_smcService.cpp
```diff
@@ -282,13 +299,19 @@ void CActivityInfo::readTargetClusterInfo(CConstWUClusterInfoArray& clusters, IP
...
+#else
+void CActivityInfo::readTargetClusterInfo(IConstContainerWUClusterInfo& cluster, IPropertyTree* serverStatusRoot, CWsSMCTargetCluster* targetCluster)
+#endif
 {
     SCMStringBuffer clusterName;
     cluster.getName(clusterName);
     targetCluster->clusterName.set(clusterName.str());
     targetCluster->clusterType = cluster.getPlatform();
+#ifndef _CONTAINERIZED
     targetCluster->clusterSize = cluster.getSize();
```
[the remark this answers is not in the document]
wangkx: In my local environment, I saw the width attribute (width=2?) for thor. But, not for roxie.
<queues name="hthor" prefix="hthor" type="hthor"/>
 <queues name="roxie-workunit" prefix="roxie_workunit" type="roxie"/>
 <queues name="roxie" prefix="roxie" type="roxie"/>
 <queues name="thor"
         prefix="thor"
         type="thor"
         width="2"/>

esp/services/ws_smc/ws_smcService.cpp
```diff
@@ -418,7 +445,10 @@ void CActivityInfo::readActiveWUsAndQueuedWUs(IEspContext& context, IPropertyTre
     readRunningWUsAndJobQueueforOtherStatusServers(context, serverStatusRoot);
     //TODO: add queued WUs for ECLCCServer/ECLServer here. Right now, they are under target clusters.
 
+#ifndef _CONTAINERIZED
     getDFUServersAndWUs(context, envRoot, serverStatusRoot);
+    //For containerized HPCC, we do not know how to find out DFU Server queues, as well as running DFU WUs, for now.
+#endif
     getDFURecoveryJobs();
```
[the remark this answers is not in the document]
wangkx: Ok, I will move it.

esp/smc/SMCLib/TpWrapper.cpp
wangkx: Will change

esp/smc/SMCLib/TpWrapper.cpp
wangkx: Will change

wangkx:
> your changes in this PR to eclccserver (adding a Server/Status entry), is that being used/picked up?

It seems the changes in this PR to eclccserver not work. I will do more tests.

wangkx:

I just did another test to check eclccserver (adding a Server/Status entry). I submitted 20 WUs to thor within a short period. All of the 20 WUs are shown under the thor. 2 of them show: 'running [ECLagent on 10.1.27.223]'. Other 18 WUs show: 'submitted [myeclccserver]'. My debug code shows that those submitted WUs are in queue thor.eclserver.

wangkx:
I did more tests. I caught a WU with the following information: 
Instance: "myeclccserver"
Server: "ECLCCserverThread"
State: "compiling"

jakesmith:
>For those submitted WUs, both myeclccserver and queue thor.eclserver are reported to ECLWatch UI. The UI code may revise if we want it show queue thor.eclserver.

@wangkx - where is the 'myeclccserver' queue coming from? No such queue will be created in a container setup. All the queues are based on the queue targets configured, so each Thor has a <thorname>.eclserver queue etc.

esp/services/ws_smc/ws_smcService.cpp
```diff
@@ -385,8 +393,12 @@ const char *CActivityInfo::getStatusServerTypeName(WsSMCStatusServerType type)
 
 bool CActivityInfo::findQueueInStatusServer(IPropertyTree* serverStatusRoot, const char* serverName, const char* queueName)
 {
+#ifndef _CONTAINERIZED
     VStringBuffer path("Server[@name=\"%s\"]", serverName);
     Owned<IPropertyTreeIterator> it(serverStatusRoot->getElements(path.str()));
+#else
+    Owned<IPropertyTreeIterator> it(serverStatusRoot->getElements("Server"));
```
jakesmith, line 400: I'm not clear on why in base-metal it's filtering, and in container mode it isn't ?

jakesmith (commented):
@wangkx - a couple of questions

esp/services/ws_smc/ws_smcService.cpp
wangkx: The function here is to check whether a queue is found in the Status Servers. In a base-metal, the filter (serverName: ThorMaster, HThorServer, etc) is used to narrow down the check. In a non-base-metal, the filter does not work anymore since we do not use the ThorMaster.

wangkx:
> where is the 'myeclccserver' queue coming from? No such queue will be created in a container setup. All the queues are based on the queue targets configured, so each Thor has a .eclserver queue etc.

ESP does not report the 'myeclccserver' as a queue. The 'thor.eclserver' is reported as the queue. As I mentioned before, for those submitted WUs, both myeclccserver and queue thor.eclserver are reported to ECLWatch UI. For some reason, ECLWatch UI chooses to display the myeclccserver, not the queue thor.eclserver. We may ask Gordon to change it if needed.

wangkx:
@jakesmith did I miss any question?

esp/services/ws_smc/ws_smcService.cpp
jakesmith: >In a non-base-metal, the filter does not work anymore since we do not use the ThorMaster.

@wangkx - should be the same as bare-metal, afaik.

wangkx: I think we discussed that before. In non-base-metal, the /Status/Servers does not have the Server which @name is ThorMaster. So, it will not work if the filter is "Server[@name=\"ThorMaster\"]"  ?

jakesmith: @wangkx - The code is the same, all Thor master (in future ThorManager) add a Server[@name="ThorMaster"].
Please try it, with multiple jobs, but I think it should work in same way as before in this respect.

Where it differs and we may have talked before, is that unlike bare-metal - where there'll only ever be exactly 1 @thorname e.g. on a multi-thor cluster in bare-metal, they define e.g. 6 thor's all with unique names, e.g. "thor200_20_a",, "thor200_20_b", ... "thor200_20_f".

There will only ever be one of each @thorname, but all will have Server[@name="ThorMaster"] to categorize their type, so you the pre-existing filter by that type in the service code, should still work.

wangkx: Ok, I will try it. I did have the problem before...

wangkx: @jakesmith this is from my dali (my code is based on 7.12.x from Jan. 12):

<Server cluster="thor"
         mpport="7197"
         name="HThorServer"
         node="10.1.28.209"
         queue="thor.thor"
         started="2021-01-14T18:14:25"/>

I did not see any /Status/Servers/Server with @name="ThorMaster".

jakesmith: @wangkx - were there thor jobs actually running at the time?

wangkx: No running at the time. I will switch back to this project and check it at a running time.

jakesmith: There will be no thor entry (of any kind) when no jobs are running, because in the K8s version, Thor is dynamically launched, multiple Thor's up to ,maxJobs/maxGraphs, when the job is finished, the Thor instance shuts down.

wangkx: If no thor entries when no jobs are running, I should define out the whole findQueueInStatusServer() for non-base-metal. The findQueueInStatusServer() checks the cluster queue entries (including thor queue entries). If an entry is not found, ESP returns 'cluster QUEUENAME not listening for workunits'. For non-base-metal, the error is not valid anymore.

wangkx:
@jakesmith please review my change and let me know if I miss anything.

ecl/eclccserver/eclccserver.cpp
```diff
@@ -900,6 +900,13 @@ int main(int argc, const char *argv[])
         const char *wuid = globals->queryProp("@workunit");
         if (wuid)
         {
+#ifdef _CONTAINERIZED
+            CSDSServerStatus serverstatus("ECLCCserverThread");
+            serverstatus.queryProperties()->setProp("@cluster", globals->queryProp("@name"));
+            serverstatus.queryProperties()->setProp("WorkUnit", wuid);
+            serverstatus.commitProperties();
```
jakesmith, line 907: I don't think this block should be required.

Each active compile thread already has a serverStatus in it for the workunit it's dealing with.
When none are compiling, there are 0.

Because of that and the Thor case, I think this code:

I think the : 
```
        if (!smcQueue->foundQueueInStatusServer)
            targetCluster->clusterStatusDetails.appendf("Cluster %s not listening for workunits; ", clusterName.str());
```

Needs looking at or disabling in container mode.

Does this also impact hthor agents when no hthor jobs are running?

@wangkx

AttilaVamos:
@wangkx I am sure, the schedule1 error is not related to this PR. I will reschedule smoketest for a clear result.

wangkx:
Thanks Attila.

jakesmith (changes requested):
@wangkx - please squash, and I think it's ready.

jakesmith (approved):
@wangkx - looks good.

jakesmith:

wangkx:

PR #14490, opened 2021-01-19
HPCC-25173 Fix 2 target related calls in cloud environment ESP

wangkx:
1. Common out the code for validating target name.
   Both refreshValidClusters() and isValidCluster() are in ws_sql
service and ws_workunits service. The related code for those 2
methods are moved into the SMCLib. The refreshValidClusters() is
renamed to refreshValidTargets(). The isValidCluster() is renamed
as validateTargetName(). Both of them are upgraded to support both
bare metal environment and cloud environment. The validateTargetName()
is upgraded to thorw an exception if the target name is invalid.
The code in ws_sql service and ws_workunits service is modified
to match with those changes.
2. Add getThorClusterNames() which returns both thor target names
and thor queue names for both bare metal environment and cloud
environment. Replace the existing getEnvironmentThorClusterNames()
call in ws_smcService.cpp.
3. Simplify the code in the CWsSMCEx::onBrowseResources().

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/services/ws_sql/ws_sqlService.cpp
+++ b/esp/services/ws_sql/ws_sqlService.cpp
@@ -30,8 +30,6 @@ void CwssqlEx::init(IPropertyTree *_cfg, const char *_process, const char *_serv
         throw MakeStringException(-1, "ws_sqlEx: Problem initiating ECLFunctions structure");
     }
 
-    refreshValidClusters();
-
     setWsSqlBuildVersion(BUILD_TAG);
 }
 
@@ -831,8 +829,7 @@ void CwssqlEx::processMultipleClusterOption(StringArray & clusters, const char  
         hashoptions.appendf("\n#OPTION('AllowedClusters', '%s", targetcluster);
         ForEachItemIn(i,clusters)
         {
-            if (!isValidCluster(clusters.item(i)))
-                throw MakeStringException(-1, "Invalid alternate cluster name: %s", clusters.item(i));
+            validateTargetName(clusters.item(i));
 
             hashoptions.appendf(",%s", clusters.item(i));
         }
@@ -946,11 +943,7 @@ bool CwssqlEx::onExecuteSQL(IEspContext &context, IEspExecuteSQLRequest &req, IE
         if (compiledwuid.length()==0)
         {
             {
-                if (isEmpty(cluster))
-                    throw MakeStringException(-1,"Target cluster not set.");
-
-                if (!isValidCluster(cluster))
-                    throw MakeStringException(-1, "Invalid cluster name: %s", cluster);
+                validateTargetName(cluster);
 
                 if (querytype == SQLTypeCreateAndLoad)
                     clonable = false;
@@ -1018,8 +1011,8 @@ bool CwssqlEx::onExecuteSQL(IEspContext &context, IEspExecuteSQLRequest &req, IE
                 createWUXMLParams(xmlparams, parsedSQL, NULL, cw);
             else if (querytype == SQLTypeSelect)
             {
-                if (notEmpty(cluster) && !isValidCluster(cluster))
-                    throw MakeStringException(ECLWATCH_INVALID_CLUSTER_NAME, "Invalid cluster name: %s", cluster);
+                if (!isEmptyString(cluster))
+                    validateTargetName(cluster);
 
                 createWUXMLParams(xmlparams, parsedSQL->getParamList());
             }
@@ -1232,8 +1225,8 @@ bool CwssqlEx::onExecutePreparedSQL(IEspContext &context, IEspExecutePreparedSQL
        context.ensureFeatureAccess(WSSQLACCESS, SecAccess_Write, -1, "WsSQL::ExecutePreparedSQL: Permission denied.");
 
        const char *cluster = req.getTargetCluster();
-       if (notEmpty(cluster) && !isValidCluster(cluster))
-           throw MakeStringException(ECLWATCH_INVALID_CLUSTER_NAME, "Invalid cluster name: %s", cluster);
+       if (!isEmptyString(cluster))
+           validateTargetName(cluster);
 
        Owned<IWorkUnitFactory> factory = getWorkUnitFactory(context.querySecManager(), context.queryUser());
 
@@ -1450,11 +1443,7 @@ bool CwssqlEx::onPrepareSQL(IEspContext &context, IEspPrepareSQLRequest &req, IE
             }
             else
             {
-                if (isEmpty(cluster))
-                    throw MakeStringException(1,"Target cluster not set.");
-
-                if (!isValidCluster(cluster))
-                    throw MakeStringException(-1/*ECLWATCH_INVALID_CLUSTER_NAME*/, "Invalid cluster name: %s", cluster);
+                validateTargetName(cluster);
 
                 ECLEngine::generateECL(parsedSQL, ecltext);
                 if (hashoptions.length() > 0)
@@ -1686,8 +1675,8 @@ bool CwssqlEx::onCreateTableAndLoad(IEspContext &context, IEspCreateTableAndLoad
         throw MakeStringException(-1, "WsSQL::CreateTableAndLoad: Error: Target TableName is invalid: %s.", targetTableName);
 
     const char * cluster = req.getTargetCluster();
-    if (notEmpty(cluster) && !isValidCluster(cluster))
-        throw MakeStringException(ECLWATCH_INVALID_CLUSTER_NAME, "WsSQL::CreateTableAndLoad: Invalid cluster name: %s", cluster);
+    if (!isEmptyString(cluster))
+        validateTargetName(cluster);
 
     IConstDataSourceInfo & datasource = req.getDataSource();
 
@@ -1979,38 +1968,6 @@ bool CwssqlEx::onGetResults(IEspContext &context, IEspGetResultsRequest &req, IE
     return success;
 }
 
-void CwssqlEx::refreshValidClusters()
-{
-    validClusters.kill();
-#ifdef _CONTAINERIZED
-    Owned<IStringIterator> it = getContainerTargetClusters(nullptr, nullptr);
-#else
-    Owned<IStringIterator> it = getTargetClusters(nullptr, nullptr);
-#endif
-    ForEach(*it)
-    {
-        SCMStringBuffer s;
-        IStringVal &val = it->str(s);
-        if (!validClusters.getValue(val.str()))
-            validClusters.setValue(val.str(), true);
-    }
-}
-
-bool CwssqlEx::isValidCluster(const char *cluster)
-{
-    if (!cluster || !*cluster)
-        return false;
-    CriticalBlock block(crit);
-    if (validClusters.getValue(cluster))
-        return true;
-    if (validateTargetClusterName(cluster))
-    {
-        refreshValidClusters();
-        return true;
-    }
-    return false;
-}
-
 bool CwssqlEx::publishWorkunit(IEspContext &context, const char * queryname, const char * wuid, const char * targetcluster)
 {
     Owned<IWorkUnitFactory> factory = getWorkUnitFactory(context.querySecManager(), context.queryUser());
@@ -2019,7 +1976,7 @@ bool CwssqlEx::publishWorkunit(IEspContext &context, const char * queryname, con
         throw MakeStringException(ECLWATCH_CANNOT_OPEN_WORKUNIT,"Cannot find the workunit %s", wuid);
 
     SCMStringBuffer queryName;
-    if (notEmpty(queryname))
+    if (!isEmptyString(queryname))
         queryName.set(queryname);
     else
         queryName.set(cw->queryJobName());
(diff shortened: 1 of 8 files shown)


wangkx:
@afishbeck please review

esp/smc/SMCLib/TpWrapper.cpp
```diff
@@ -2213,3 +2213,22 @@ extern TPWRAPPER_API void initContainerRoxieTargets(MapStringToMyClass<ISmartSoc
...
 
+extern TPWRAPPER_API unsigned getContainerThorClusterNames(StringArray& targetNames, StringArray& queueNames)
+{
+    Owned<IStringIterator> targets = getContainerTargetClusters("thor", nullptr);
+    ForEach(*targets)
+    {
+        SCMStringBuffer target;
+        targets->str(target);
+        targetNames.append(target.str());
+        StringBuffer queueName(target.str());
```
afishbeck, line 2224: Very minor optimization, but you do have access to target.s and don't have to create a second StringBuffer queueName.

```
SCMStringBuffer target;
targets->str(target);
targetNames.append(target.str());
queueNames.append(target.s.append(THOR_QUEUE_EXT));
```

afishbeck (dismissed):
Other than a trivial optimization, looks good.

wangkx:
@afishbeck please review my change

afishbeck (dismissed):
@wangkx looks good.

wangkx:

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -649,6 +649,17 @@ bool CFileSprayEx::onDFUWUSearch(IEspContext &context, IEspDFUWUSearchRequest &
     {
         context.ensureFeatureAccess(DFU_WU_URL, SecAccess_Read, ECLWATCH_DFU_WU_ACCESS_DENIED, "Access to DFU workunit is denied.");
 
+#ifdef _CONTAINERIZED
+        StringArray clusterNames;
+        Owned<IStringIterator> targets = getContainerTargetClusters(nullptr, nullptr);
```
jakesmith, line 654: is it right to use getContainerTargetClusters() which also seems to get roxie services too?

I'm not clear on what this service method ('DFUWUSearch') is used for?
e.g. why us called DFUWUSearch but returns a list of targets?

esp/smc/SMCLib/TpWrapper.cpp
```diff
@@ -2213,3 +2213,21 @@ extern TPWRAPPER_API void initContainerRoxieTargets(MapStringToMyClass<ISmartSoc
...
+{
+    Owned<IStringIterator> targets = getContainerTargetClusters("thor", nullptr);
+    ForEach(*targets)
+    {
+        SCMStringBuffer target;
+        targets->str(target);
+        targetNames.append(target.str());
+        queueNames.append(target.s.append(THOR_QUEUE_EXT));
+    }
+    return targetNames.ordinality();
```
jakesmith, line 2226: would it be better if this was implemented by using getContainerTargetClusters (with a type filter) ?
It then provides a method to get the queue name, rather than reconstructing it here.

esp/services/ws_sql/ws_sqlService.cpp
```diff
@@ -2003,7 +2003,11 @@ bool CwssqlEx::isValidCluster(const char *cluster)
     CriticalBlock block(crit);
     if (validClusters.getValue(cluster))
         return true;
+#ifndef _CONTAINERIZED
     if (validateTargetClusterName(cluster))
+#else
+    if (validateContainerTargetClusterName(cluster))
+#endif
```
jakesmith, line 2010: minor: good to minimize the amount of conditional code if possible/clean to do so.
As this is used here and in ws_workunitsService below (and prob. other places in future?), it would be cleaner if a common method (e.g. add a 'validateTargetName' to TpWrapper) contained this conditional code, so that each use like this didn't need to have conditional code in it.

esp/services/ws_fs/ws_fsService.cpp
wangkx: The DFUWUSearch retrieves the information for searching DFU WUs. It was built for the legacy ECLWatch page (and may be used in other places). The code here retrieves cluster names (used for searching the WUs by cluster).

esp/smc/SMCLib/TpWrapper.cpp
wangkx: @jakesmith do you mean that I should implement the getContainerWUClusterInfo(with a type filter) ? The getContainerTargetClusters() returns an array of target names.

esp/services/ws_sql/ws_sqlService.cpp
wangkx: I guess I can move the whole isValidCluster() (and the conditional code) into a new validateTargetName().

wangkx: The code has been modified to share the code from both ws_workunitsService and ws_sql.

esp/services/ws_fs/ws_fsService.cpp
afishbeck: Agree the method was badly named.

afishbeck: I think Jake is right though.  In the long run we may want a new section listing clusters rather than queues or services. 
 But to get this working as is for DFU purposes we want the roxie queues names.  Unfortunately we currently use the type name roxie to mean services but there are also queue types of roxie.  So we probably want to add a different type name to getContainerTargetClusters() to get the "roxie-services"... or maybe have a different function for those.
 
Btw a roxie cluster will show up once in the queue list, but might have multiple services pointing to it.

afishbeck: Need to think some more on general that targets can be compileOnly (roxie), runnable (hthor, roxie-workunit, thor), or callable (roxie-services).  And we just need to look at where each type may be needed, and how to add support for that to the functions.. perhaps including the validateTarget function as well.

afishbeck (commented):
Some comments.

esp/services/ws_fs/ws_fsService.cpp
wangkx: In bare metal environment, any DFU WU can run in a roxie? In cloud environment, does it make sure to search DFU WUs by a target (thor target, roxie target, or hthor target)?

afishbeck: The issue is that the roxie can expose multiple services and a queue and all the services show up in the config.

myroxie
   roxie-service1
   roxie-service2

When listing targets to send queries to we want roxie-service1, roxie-service2.
When listing targets to run workunits on we don't want any.  Generally we only want workunit (eclagent roxies).  Technically hat could change later if we have roxie listen on a queue.
For DFU I think we want myroxie.  This is the actual cluster and will be where the files are accessed. 

In the current way it's done that would be the queue of type roxie, not the roxie services.
In the long run we probably want to define clusters explicitly in the configMap.  Clearly showing the difference between clusters and queues... because otherwise different parts of the code could interpret things differently.

wangkx:
Based on the review comments, I have merged duplicated code. I also rolled back my changes in CFileSprayEx::onDFUWUSearch() since, at this time, what to be returned for cloud environment is not clear.  The PR title and message are revised to match with those changes. The commits are squashed into one to make it easy for review. @jakesmith please review

esp/smc/SMCLib/TpWrapper.cpp
```diff
@@ -2213,3 +2213,68 @@ extern TPWRAPPER_API void initContainerRoxieTargets(MapStringToMyClass<ISmartSoc
...
+
+#ifndef _CONTAINERIZED
+    if (!validateTargetClusterName(target))
+#else
+    VStringBuffer xpath("queues[@name=\"%s\"]", target);
+    if (queryComponentConfig().queryPropTree(xpath) == nullptr)
+#endif
+        throw MakeStringException(ECLWATCH_INVALID_CLUSTER_NAME, "Invalid target name: %s", target);
+
+    refreshValidTargets();
```
jakesmith, line 2279: reloaded every time there's a mismatch? That sounds fairly wasteful.
In bare metal once loaded, does it ever need reloading? Only on an environment change? It could detect that..

In containerized mode, currently, if there's any change to the target queues, esp will be auto restarted by K8s.
In near future, we want to optimize that by spotting changes dynamically and refreshing internal structures/data like this. For now I'd just add a comment and never refresh.

esp/smc/SMCLib/TpWrapper.cpp
```diff
@@ -2213,3 +2213,68 @@ extern TPWRAPPER_API void initContainerRoxieTargets(MapStringToMyClass<ISmartSoc
...
+        }
+    }
+}
+
+extern TPWRAPPER_API void validateTargetName(const char* target)
+{
+    if (isEmptyString(target))
+        throw makeStringException(ECLWATCH_INVALID_CLUSTER_NAME, "Empty target name.");
+
+    CriticalBlock block(validTargetSect);
```
jakesmith, line 2266: as per other comment re. 'dirty' flag.
I think better pattern would be to check a flag here (e.g. targetsDirty) and call refresh if set.
That flag would initially be set to true. And other circumstances in future may toggle it.

esp/smc/SMCLib/TpWrapper.cpp
```diff
@@ -2213,3 +2213,68 @@ extern TPWRAPPER_API void initContainerRoxieTargets(MapStringToMyClass<ISmartSoc
...
+        targetNames.append(target.str());
+
+        StringBuffer qName;
+        queueNames.append(getClusterThorQueueName(qName, target.str()));
+    }
+#endif
+    return targetNames.ordinality();
+}
+
+static BoolHash validTargets;
```
jakesmith, line 2236: worth consiidering using a std::set<std::string> instead here.

jakesmith:
@wangkx - looks good/cleaner generally. Added a few minor comments, mostly re. issues that aren't new.

esp/smc/SMCLib/TpWrapper.cpp
wangkx: std::setstd::string ? Any sample?

esp/smc/SMCLib/TpWrapper.cpp
```diff
@@ -2213,3 +2213,68 @@ extern TPWRAPPER_API void initContainerRoxieTargets(MapStringToMyClass<ISmartSoc
...
+    if (found && *found)
+        return;
+
+#ifndef _CONTAINERIZED
+    if (!validateTargetClusterName(target))
+#else
+    VStringBuffer xpath("queues[@name=\"%s\"]", target);
+    if (queryComponentConfig().queryPropTree(xpath) == nullptr)
+#endif
+        throw MakeStringException(ECLWATCH_INVALID_CLUSTER_NAME, "Invalid target name: %s", target);
```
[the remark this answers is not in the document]
wangkx: Will change

esp/services/ws_sql/ws_sqlService.cpp
```diff
@@ -831,8 +831,7 @@ void CwssqlEx::processMultipleClusterOption(StringArray & clusters, const char
         hashoptions.appendf("\n#OPTION('AllowedClusters', '%s", targetcluster);
         ForEachItemIn(i,clusters)
         {
-            if (!isValidCluster(clusters.item(i)))
-                throw MakeStringException(-1, "Invalid alternate cluster name: %s", clusters.item(i));
+            validateTargetName(clusters.item(i));
```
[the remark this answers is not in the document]
wangkx: Did we say that the check and the throw should be moved inside the API call (here validateTargetName()) so that every caller  does not have to check and throw?

esp/services/ws_sql/ws_sqlService.cpp
```diff
@@ -1018,8 +1013,8 @@ bool CwssqlEx::onExecuteSQL(IEspContext &context, IEspExecuteSQLRequest &req, IE
                 createWUXMLParams(xmlparams, parsedSQL, NULL, cw);
             else if (querytype == SQLTypeSelect)
             {
-                if (notEmpty(cluster) && !isValidCluster(cluster))
-                    throw MakeStringException(ECLWATCH_INVALID_CLUSTER_NAME, "Invalid cluster name: %s", cluster);
+                if (notEmpty(cluster))
```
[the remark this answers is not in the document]
wangkx: Will change

esp/smc/SMCLib/TpWrapper.cpp
wangkx: This is from the existing code. I do agree with you. I haven't seen an environment where the whole HPCC system does not need to restart after its environment setting changes (even for a bare metal environment). So, the lines 2271 - 2279 should be removed. Any objection?

esp/smc/SMCLib/TpWrapper.cpp
jakesmith:
```
static std::set<std::string> validTargets;

...
..
   validTargets.insert(atarget);
...
   if (validTargets.end() != validTargets.find(thetarget))
   { // match
       .
   }


```

esp/services/ws_sql/ws_sqlService.cpp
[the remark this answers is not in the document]
jakesmith: possibly. it's fine as is.

esp/smc/SMCLib/TpWrapper.cpp
wangkx: Thanks.

esp/smc/SMCLib/TpWrapper.cpp
jakesmith: In cloud/containers - should load once and not refresh every mismatch at the moment - with a comment re. refreshing dynamically in future, but currently it will auto-restart the container.

In bare-metal, I'm not sure. What if the environment updates? Although wasteful, suspect should leave it as is, reloading on every mismatch in bare metal for now.

wangkx:

jakesmith (commented):

wangkx:
@jakesmith the commits are squashed.

jakesmith (approved):

PR #14516, opened 2021-01-26
HPCC-25251 Increase default spray buffer size and remove explicit default

ghalliday:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dali/ft/daftformat.hpp
+++ b/dali/ft/daftformat.hpp
@@ -22,7 +22,7 @@
 #include "daft.hpp"
 #include "ftbase.ipp"
 
-#define DEFAULT_STD_BUFFER_SIZE         0x10000
+#define DEFAULT_STD_BUFFER_SIZE         0x100000
 #define EFX_BLOCK_HEADER_TYPE   unsigned
 #define EFX_BLOCK_HEADER_SIZE   sizeof(EFX_BLOCK_HEADER_TYPE)
 #define VARIABLE_LENGTH_TYPE    unsigned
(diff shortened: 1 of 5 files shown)


ghalliday:
@jakesmith @mckellyln this is partly for discussion and possible merging.
I think the explicit default in the environment file is a mistake.  (Ideally it should depend on the target, and the explicit default prevents that being decided automatically.)  This could target master instead, I don't have a strong opinion.

However, what should be the size of the transfer size?  I suspect when reading/writing from a local disk a smaller size may reduce latency.  But sprays/copies will always involve at least one remote read or write, and a larger size will help there.  Will 1MB cause any issues e.g. in dafilesrv?  Increasing this is fairly essential for creating azure blobs - it would ideally be 4MB, but 1MB will do for now.

jakesmith:
> I think the explicit default in the environment file is a mistake. (Ideally it should depend on the target, and the explicit default prevents that being decided automatically.)

I agree, removing default from the environment makes sense.

>This could target master instead, I don't have a strong opinion.

Not sure, it _should_ be safer enough for 7.12 - but it probably should not really go into a point build, and is more needed for cloud (8.0).

> However, what should be the size of the transfer size? I suspect when reading/writing from a local disk a smaller size may reduce latency. But sprays/copies will always involve at least one remote read or write, and a larger size will help there.

I suspect the upside of larger writes in general outweighs any benefit of small lower latency writes, which we don't tend to do.

>Will 1MB cause any issues e.g. in dafilesrv? Increasing this is fairly essential for creating azure blobs - it would ideally be 4MB, but 1MB will do for now.

I don't think so, most dafilesrv read sizes will I think already be 1MB, because they are via createRowStream, which in turn is ultimately via CSerialStreamBase I think - which uses DEFAULT_STREAM_BUFFER_SIZE (1MB).

There are likely other default/hard-wired sizes elsewhere in other situations that are used for both dafilesrv and IFile.
I think they can all safely default to 1MB.

jakesmith (commented):
Looks sensible to me.

mckellyln:
```
class CDFUengine: public CInterface, implements IDFUengine
{
    StringBuffer dfuServerName;
    size32_t defaultTransferBufferSize;

    void setDefaultTransferBufferSize(size32_t size)
    {
        defaultTransferBufferSize = size;
    }
```
Do we need to add a check for 0 here and if it is 0 then set it to DEFAULT_STD_BUFFER_SIZE ?

ghalliday:
Yes, filecopy is the only place that it is used.

mckellyln (approved):
Looks fine.
Approved.

PR #14524, opened 2021-01-26
HPCC-25262 Metrics milestone 1

kenrowland:
Implement and collect a gauge metric

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:


kenrowland:
@ghalliday, @rpastrana Please review

devdoc/Metrics.rst
```diff
@@ -0,0 +1,282 @@
...
+  Or even by priority (high, low, sla)
+
+* Current operational levels such as the length of internal queues
+* The latency of operations such as queue results, agent responses, and gateway responses
+
+Roxie also has the need to track internal memory usage beyond the pod/system level capabilities. Tracking the state of its large fixed memory pool is necessary.
+
+The Roxie buddy system also must track how often and who is completing requests. The "I Beat You To It" set of metrics must be collected and exposed in order to detect pending node failure. While specific action on these counts is not known up front, it appears clear that these values are useful and should be collected.
+
+There does not appear to be a need for creating and destroying metrics dynamically. The set of metrics is most likely to be created at startup and remain active through the life of the Roxie. If, however, stats collection seeps into the metrics framework, dynamic creation and destruction of stats metrics is a likely requirement.
```
ghalliday, line 91: This is debatable.  I suspect operations will want to monitor deployed queries individually, and if so that means metrics will need to be created and destroyed dynamically - as queries are deployed.  It will not change very frequently on a production system.

devdoc/Metrics.rst
```diff
@@ -0,0 +1,282 @@
...
+**********************
+Framework Requirements
+**********************
+Not sure if we need this section or not. Might be good to collect high level instead of per area requirements.
+
+Metric Values
+===============
+Metric value requirements.  Capture the following:
+
+* intrinsic value ownership
```
ghalliday, line 124: I am not clear what this means.

devdoc/Metrics.rst
```diff
@@ -0,0 +1,282 @@
...
+Document requirements of sinks to collect and report values to a backend collection systems
+
+Supported Collection Systems
+----------------------------
+Document the initial set of sinks. Probably the following:
+
+* Prometheus
+* Data dog
+* Elasticsearch
+* Null
```
ghalliday, line 143: Do you mean file?  I'm not sure what a null sink means.

devdoc/Metrics.rst
```diff
@@ -0,0 +1,282 @@
...
+
+* Simple methods for the component to update the metric
+* Simple methods for the framework to retrieve metric value(s)
+* Handling of all synchronization between updating and retrieving metric values
+
+In addition, the framework provides the support for retrieving values so that the component does not participate in metric reporting. The component simply creates the metrics it needs, then instruments the component to update the metric whenever its state changes. For example, the component may create a metric that counts the total number of requests it has received. Then, wherever the component receives a request, a corresponding update to the count is added. No where in the component is any code added to retrieve the count as that is handled by the framework.
```
ghalliday, line 160: typo no where -> nowhere

devdoc/Metrics.rst
```diff
@@ -0,0 +1,282 @@
...
+
+****************
+Framework Design
+****************
+
+This section covers the design and architecture of the framework. It discusses the main areas of the design, the interactions between each area, and an overall process model of how the framework operates.
+
+The framework consists of three major areas: metrics, sinks, and the glue logic. These area work together with the platform and the component to provide a reusable metrics collection function.
+
+Metrics represent the quantifiable component state measurements used to track and assess the status of the component and the overall cluster. Metrics are typically scaler values that are easily aggregated by a collection system. Aggregated values provide the necessary input to take component and cluster actions such as scaling up and down. The component is responsible for creating metrics and instrumenting the code. The framework provides the support for collecting and reporting the values. Metrics provide the following
```
ghalliday, line 154: typo scaler -> scalar

devdoc/Metrics.rst
```diff
@@ -0,0 +1,282 @@
...
+Metrics Implementations
+=======================
+
+The sections that follow discuss metric implementations.
+
+Gauge Metric
+------------
+A gauge metric is a continuously updated value representing the current state of an interesting value in the component. For example, the amount of memory used in an internal buffer, or the number of requests waiting on a queue. A gauge metric may increase or decrease in value as needed. Reading the value of a gauge is a stateless operation in that there are no dependencies on the previous reading. The value returned shall always be the current state.
+
+Once created, the component shall update the gauge anytime the state of what is measured is updated. The metric shall provide methods to increase and decrease the value. The sink reads the value during collection an reporting.
```
ghalliday, line 231: typo an->and

devdoc/Metrics.rst
```diff
@@ -0,0 +1,282 @@
...
+
+
+*************
+Configuration
+*************
+This section discusses configuration. Since Helm charts are capable of combining configuration data at a global level into a component's specific configuration, a single config source is all all that is required. The configuration takes the form as shown below. Note that as the design progresses it is expected that there will be additions.
+
+::
+
+  component:
```
ghalliday, line 241: In the Helm file I would expect the normal situation to be for metric sinks to be defined in a common section - i.e. all components report the metrics to the same location.
However you may want to override settings for the sink per component.  If so you may need to think carefully about the structure to allow merging to be done cleanly.

devdoc/Metrics.rst
```diff
@@ -0,0 +1,282 @@
...
+
+    1. Name of the shared object library to load, which may include a full path,
+       otherwise standard library path searching for the OS is used.
+    2. As part of a standard hpcc lib name in "libhpccmetrics_<type>"
+
+metrics.sinks.settings
+    A set of key/value pairs that passed to the sink when initialized. It should contain information necessary for the operation of the sink.
+    Nest YML is supported. Example settings are the prometheus server name, or the collection period for a periodic sink.
+
+metrics.sinks.metrics
```
ghalliday, line 274: By default a component should report all metrics (the goal is for minimal need to modify helm charts to get a system up and running).  Any filtering will need to be more general, but doesn't need to be detailed now.

devdoc/Metrics.rst
```diff
@@ -0,0 +1,282 @@
...
+    Metrics configuration for the component
+
+metrics.sinks
+    List of sinks defined for the component (may have been combined with global config)
+
+metrics.sinks.name
+    A name for the sink. Note this may not be needed, but can provide a way to combine global and component config based on value
+
+metrics.sinks.type
+    The type for the sink. The type defines how to find the sink in order to load it. The value is used as follows:
```
ghalliday, line 264: There should be a simple standard mapping from the type to the name of the file - which should be deployed in our standard location for shared objects.  e.g. prometheus -> libprometheousmetric.so.  I don't think we need to offer options- unless they solve a design issue.

devdoc/Metrics.rst
```diff
@@ -0,0 +1,282 @@
...
+  component:
+    metrics:
+      sinks:
+        - name: <sink name>
+          type: <sink_type>
+          settings:
+            sink_setting1: sink_setting_value1
+            sink_setting2: sink_setting_value2
+          metrics:
+            - name: <metric_name>
```
ghalliday, line 250: worth having a real example to illustrate the structure.  Especially how the structure supports options being set on a component to override a global setting.

devdoc/Metrics.rst
```diff
@@ -0,0 +1,282 @@
...
+
+There does not appear to be a need for creating and destroying metrics dynamically. The set of metrics is most likely to be created at startup and remain active through the life of the Roxie. If, however, stats collection seeps into the metrics framework, dynamic creation and destruction of stats metrics is a likely requirement.
+
+
+ESP
+===
+
+There are some interesting decisions with respect to ESP and collection of metrics. Different applications within ESP present different use cases for collection. Ownership of a given task drives some of these use cases. Take workunit queues. If ownership of the task, with respect to metrics, is WsWorkunits, then uses cases are centric to that component. However, if agents listening on the queue are to report metrics, then a different set of use cases emerge. It is clear that additional work is needed to generate clear ownership of metrics gathered by ESP and/or the tasks it performs.
```
ghalliday, line 97: My impression of this section is that we don't have a clear idea of what metrics esp needs to collect.  That's fine, but it is worth moving the activeTransactions first, and then note the rest is for future investigation/discussion.

devdoc/Metrics.rst
```diff
@@ -0,0 +1,282 @@
...
+  action
+
+* Fault diagnosis and resource monitoring
+
+  Metrics provide historical data useful in diagnosing problems by profiling how demand and
+  usage patterns may change prior to a fault. Predictive analysis can also be applied.
+
+* Analysis of jobs/workunits and profiling
+
+  With proper instrumentation, a robust dynamic metric strategy can track workunit
```
ghalliday, line 32: I think the conclusion was that this area was outside the scope of the metric framework and would be supported via stats/logging.  Possibly move the following line from the esp section here:
"Internal problems with queries should be diagnosed from deep drill down logging"

devdoc/Metrics.rst
```diff
@@ -0,0 +1,282 @@
...
+------------
+A gauge metric is a continuously updated value representing the current state of an interesting value in the component. For example, the amount of memory used in an internal buffer, or the number of requests waiting on a queue. A gauge metric may increase or decrease in value as needed. Reading the value of a gauge is a stateless operation in that there are no dependencies on the previous reading. The value returned shall always be the current state.
+
+Once created, the component shall update the gauge anytime the state of what is measured is updated. The metric shall provide methods to increase and decrease the value. The sink reads the value during collection an reporting.
+
+
+*************
+Configuration
+*************
+This section discusses configuration. Since Helm charts are capable of combining configuration data at a global level into a component's specific configuration, a single config source is all all that is required. The configuration takes the form as shown below. Note that as the design progresses it is expected that there will be additions.
```
ghalliday, line 237: The metrics are configured by a section in the component specific yaml file that is generated from the helm values file.  No real need to say "a single  config source is all all that is required".

devdoc/Metrics.rst
ghalliday: (I'm not 100% sure how to do this... it would be worth discussing)

devdoc/Metrics.rst
```diff
@@ -0,0 +1,282 @@
...
+    A name for the sink. Note this may not be needed, but can provide a way to combine global and component config based on value
+
+metrics.sinks.type
+    The type for the sink. The type defines how to find the sink in order to load it. The value is used as follows:
+
+    1. Name of the shared object library to load, which may include a full path,
+       otherwise standard library path searching for the OS is used.
+    2. As part of a standard hpcc lib name in "libhpccmetrics_<type>"
+
+metrics.sinks.settings
```
ghalliday, line 270: We haven't generally nested other settings - but I don't have a strong objection.

ghalliday (commented):
I think that's a good start.  A few fairly minor comments.  I expect many of them will become clearer, and sections will be filled out, as you implement the first milestone.

devdoc/Metrics.rst
kenrowland: The section in which this lives is more of a place holder to capture any requirements we decide to specifically enumerate. The current text is suggesting items that may be covered in the section. Perhaps it should be removed until such requirements exist and we decide to document them.

devdoc/Metrics.rst
kenrowland: My statement on the need for creating and destroying metrics was from the Roxie point of view based on comments from Richard. I suppose it's based on the internal operation of Roxie, not the logical use of Roxie. I do believe the need for dynamic metrics exists and should be included in the design.

devdoc/Metrics.rst
kenrowland: I should add File to the list as well. I always saw the File sink as more of a development tool and not something that would be used in production. However, it appears that there may be a use for the file sink beyond development. A Null sink would simply allow for enabling the collection and reporting of metrics, but without them actually going anywhere. I see it being useful when duplicating environments and settings but you don't want to fully setup collection. Simply change the sink in the global part of the metrics config to the null sink. Everything runs normally, but the metrics are thrown away. I see this being useful for testing.

devdoc/Metrics.rst
kenrowland: I should have included that. The current framework works that way. All component metrics are reported by default if nothing is explicitly enumerated. I shall update accordingly.

devdoc/Metrics.rst
kenrowland: We can certainly simplify the mapping (using #2 probably). However, @rpastrana originally felt it was important to be able to fully specify the name of the sink in case the sink support was contained in a lib that also provided other support. It also provided a way to quickly test a new sink w/o having to save and restore the old one. Before removing the ability to fully specify the name I just wanted to identify the reasons it was added.

devdoc/Metrics.rst
kenrowland: Need to better understand how helm accomplishes this. The YML presented in the doc is assumed to be the combined config from both global and component pieces and represents what the framework parses.

devdoc/Metrics.rst
kenrowland: Agreed.

devdoc/Metrics.rst
kenrowland: Moved.

devdoc/Metrics.rst
kenrowland: Removed and added the word "combined" to indicate that the framework is parsing a single set of settings that has been combined by an outside process, helm in this case.

kenrowland:
@ghalliday I have noticed that the original editing tool I used only did soft breaks on the lines. That makes it harder to edit in other places. I apologize, but the next commit is going to add line breaks to make it easier to read and modify in other editors. I figure it's best to do this now instead of later.

devdoc/Metrics.rst
ghalliday: That is a good idea to show the merged config, but we also need the user config to provide some documentation, and ensure the structure is correct.

devdoc/Metrics.rst
ghalliday: I'm not really convinced by either example.

devdoc/Metrics.rst
ghalliday: To summarise the discussion.  For the first version we only have a global metric configuration and all components that support metrics use that same configuration.  We revisit once there is a requirement to configure a sink on a per component basis.  (And then merge either in the helm generation or quite possibly within the framework code.)

rpastrana (commented):
@kenrowland added some comments, I don't expect all of them to result in changes to the doc.

devdoc/Metrics.rst
```diff
@@ -0,0 +1,354 @@
...
+applications within ESP present different use cases for collection. Ownership of a given task drives
+some of these use cases. Take workunit queues. If ownership of the task, with respect to metrics, is
+WsWorkunits, then use cases are centric to that component. However, if agents listening on the queue
+are to report metrics, then a different set of use cases emerge. It is clear that additional work is
+needed to generate clear ownership of metrics gathered by ESP and/or the tasks it performs.
+
+ESP needs to report the *activeTransactions* value from the TxSummary class(es). This gives an
+indication of how busy the ESP is in terms of client requests.
+
+Direct measurement of response time in requests may not be useful since the type of request causes
```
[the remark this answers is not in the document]
kenrowland: The notes here are based on a response from Tony. I wanted to capture the thoughts.

devdoc/Metrics.rst
```diff
@@ -0,0 +1,354 @@
...
+  A measurement defined by a component that represent an internal state that is useful in a system
+  reliability engineering function. In the context of the framework, a metric is an object representing
+  the above.
+
+Metric Value
+  The current value of a metric. A corollary is a metric measurement which represents a view of a metric
+  value that may be based on state.
+
+Collection
+  The process of choosing the relevant metrics and retrieving their values.
```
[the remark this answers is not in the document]
kenrowland: I have updated the definitions adding a new term, to hopefully cover the confusion. I wanted to leave any reference to specific responsible components out of the definitions. That way should the framework change, the definitions remain valid.

devdoc/Metrics.rst
```diff
@@ -0,0 +1,354 @@
...
+than metrics since the list of queries to monitor will vary between clusters. Additionally, operational
+metrics solving the cases mentioned above will give a view into the overall health of ESP which will
+affect the execution of queries. Depending on actions taken by these metrics, scaling may solve
+overload conditions to keep cluster responsiveness acceptable.
+
+For Roxie a workunit operates as a service. Measuring service performance using a histogram to capture
+response times as a distribution may be appropriate. Extracting the 95th percentile of response time
+may be useful as well.
+
+There does not be a use case requiring consistency between values of different metrics.
```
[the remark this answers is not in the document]
kenrowland: Updated

devdoc/Metrics.rst
```diff
@@ -0,0 +1,354 @@
...
+
+Dali Use Cases
+==============
+
+From information gathered, Dali plans to keep counts and rates for many of the items it manages.
+
+**********************
+Framework Requirements
+**********************
+Not sure if we need this section or not. Might be good to collect high level instead of per area
```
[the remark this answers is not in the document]
kenrowland: Removing for now. If we decide to explicitly enumerate requirements later, we can decide where they should live.

devdoc/Metrics.rst
```diff
@@ -0,0 +1,354 @@
...
+
+Collection
+  The process of choosing the relevant metrics and retrieving their values.
+
+Trigger
+  What causes the collection of metric values.
+
+Metric Report
+  Formatted metric values ready for ingestion by a collection system.
+
```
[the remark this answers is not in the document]
kenrowland: Added reporting as a term based on previous feedback.

devdoc/Metrics.rst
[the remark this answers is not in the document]
rpastrana: Right, I recall that comment, and I agree "it may not be useful" because the individual time segments that comprise the entire response time will pin-point potentials issues. But at the same time, its importance in monitoring shouldn't be underestimated, specially when all individual time segments aren't available.

devdoc/Metrics.rst
```diff
@@ -0,0 +1,330 @@
...
+       otherwise standard library path searching for the OS is used.
+    2. As part of a standard hpcc lib name in "libhpccmetrics_<type>"
+
+metrics.sinks.settings
+    A set of key/value pairs that passed to the sink when initialized. It should contain information
+    necessary for the operation of the sink. Nested YML is supported. Example settings are the
+    prometheus server name, or the collection period for a periodic sink.
+
+metrics.sinks.metrics
+    Optional list of component defined metrics reported by the sink to the backend during collection
```
JamesDeFabia, line 322: component-defined

JamesDeFabia (commented):
A few comments inline

devdoc/Metrics.rst
```diff
@@ -0,0 +1,330 @@
...
+
+
+*************
+Use Scenarios
+*************
+This section describes how components expect to use the framework. It is not a complete list of all
+requirements but rather a sample.
+
+
+Roxie
```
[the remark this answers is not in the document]
richardkchapman: I disagree (and while the marketing folks may get to overrule me elsewhere, I get to say how it should be spelled here!)

The name Roxie may have been inspired by an acronym but (like laser) it really isn't used as one any more. And it was equally inspired by a girl's name.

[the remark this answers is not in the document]
kenrowland: I shall leave this as Roxie for now and allow the battle to be waged at a later date.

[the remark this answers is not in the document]
JamesDeFabia: I agree with RKC! I was only following what I was told. I am happy to leave it as Roxie.

devdoc/Metrics.rst
```diff
@@ -0,0 +1,330 @@
...
+
+Sinks provide a pluggable interface to hide the specifics of collection systems so that the metrics
+framework is independent of those dependencies. Sinks
+
+* Operate independent of other sinks in the system
+* Convert metric native values into collection system specific measurements and reports
+* Drive the collection and reporting processes
+
+The third area of the framework is the glue logic known as the *MetricsReporter*. It manages the metrics
+system for the component. If provides the following
```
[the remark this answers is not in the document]
richardkchapman: It not If

ghalliday:
@kenrowland I think your branch needs cleaning up - you need to rebase onto master, rather than merging other branches into yours.

metrics/framework/Metrics.hpp
```diff
@@ -0,0 +1,305 @@
...
+        float getValue() const { return gaugeValue; }
+
+        /*
+         * Get current gauge value
+         */
+        void getReportValues(MeasurementVector &mv) const override;
+
+
+    protected:
+        std::atomic<float> gaugeValue{0};
```
ghalliday, line 230: Why float?  That will be inefficient.  I don't think intel (or any other processor) implements incrementing it in hardware.  A terrible type to use for an atomic.

metrics/framework/Metrics.cpp
```diff
@@ -0,0 +1,212 @@
...
+}
+
+
+void GaugeMetric::getReportValues(MeasurementVector &mv) const
+{
+    mv.emplace_back((std::make_shared<Measurement<float>>(name, description, ValueType::METRICS_FLOAT, gaugeValue.load())));
+}
+
+
+IMetricSink *MetricSink::getSinkFromLib(const char *type, const char *sinkName, const IPropertyTree *pSettingsTree)
```
ghalliday, line 30: I'm not sure why this would be a member of MetricSink, rather than a member of the class that calls it.

metrics/framework/Metrics.cpp
```diff
@@ -0,0 +1,212 @@
...
+{
+    return MetricSink::getSinkFromLib(type, "", sinkName, pSettingsTree);
+}
+
+IMetricSink *MetricSink::getSinkFromLib(const char *type, const char *getInstanceProcName, const char *sinkName, const IPropertyTree *pSettingsTree)
```
ghalliday, line 35: not sure why you have two functions.

metrics/framework/Metrics.cpp
```diff
@@ -0,0 +1,212 @@
...
+    // First, treat type as a full library name
+    libName = type;
+    if (libName.find(SharedObjectExtension) == std::string::npos)
+    {
+        libName.append(SharedObjectExtension);
+    }
+    HINSTANCE libHandle = LoadSharedObject(libName.c_str(), true, false);
+
+    //
+    // If type wasn't the lib name, treat it as a type
```
ghalliday, line 49: why?  We can define what the name means.  What is the situation where this is needed?

metrics/framework/Metrics.cpp
```diff
@@ -0,0 +1,212 @@
...
+    //
+    // If able to load the lib, get the instance proc and create the sink instance
+    IMetricSink *pSink = nullptr;
+    if (libHandle != nullptr)
+    {
+        const char *epName = !isEmptyString(getInstanceProcName) ? getInstanceProcName : "getSinkInstance";
+        auto getInstanceProc = (getSinkInstance) GetSharedProcedure(libHandle, epName);
+        if (getInstanceProc != nullptr)
+        {
+            const char *name = isEmptyString(sinkName) ? type : sinkName;
```
ghalliday, line 66: this also seems unnecessary

metrics/framework/Metrics.cpp
```diff
@@ -0,0 +1,212 @@
...
+bool MetricsReporter::initializeSinks(IPropertyTreeIterator *pSinkIt)
+{
+    bool rc = true;
+    for (pSinkIt->first(); pSinkIt->isValid() && rc; pSinkIt->next())
+    {
+        SinkInfo *pSinkInfo;
+        IPropertyTree &sinkTree = pSinkIt->query();
+
+        StringBuffer cfgSinkType, cfgLibName, cfgSinkName, cfgProcName;
+        sinkTree.getProp("@type", cfgSinkType);  // this one is required
```
ghalliday, line 163: minor, code to access strings from property trees would more normally be:

    const char * sinkType = sinkTree.queryProp("type");

metrics/framework/Metrics.cpp
```diff
@@ -0,0 +1,212 @@
...
+        sinkTree.getProp("@libname", cfgLibName);
+        sinkTree.getProp("@name", cfgSinkName);
+        sinkTree.getProp("@instance_proc", cfgProcName);
```
ghalliday, line 166: This structure isn't in the design document

metrics/framework/Metrics.cpp
```diff
@@ -0,0 +1,212 @@
...
+                SinkInfo sinkInfo;
+                sinkInfo.pSink = pSink;
+                auto insertRc = sinks.insert({pSink->getName(), sinkInfo});
+                pSinkInfo = &insertRc.first->second;
+            }
+        }
+
+        //
+        // Retrieve metrics to be reported
+        if (pSinkInfo != nullptr)
```
ghalliday, line 193: this is outside the scope of the 1st milestone

metrics/framework/Metrics.cpp
```diff
@@ -0,0 +1,212 @@
...
+void MetricsReporter::stopCollecting()
+{
+    for (auto const &sinkIt : sinks)
+    {
+        sinkIt.second.pSink->stopCollection();
+    }
+}
+
+
+std::vector<std::shared_ptr<IMetric>> MetricsReporter::getReportMetrics(const std::string &sinkName)
```
ghalliday, line 121: This seems to add an unnecessary extra step - create a list of metrics to report, and then call report on them.  There should be a function to gather the measurements, without returning the temporary list.

metrics/framework/Metrics.hpp
```diff
@@ -0,0 +1,305 @@
...
+    protected:
+        std::string name;
+        std::string type;
+        MetricsReporter *pReporter = nullptr;
+};
+
+
+struct SinkInfo
+{
+    IMetricSink *pSink = nullptr;             // ptr to the sink
```
ghalliday, line 269: minor: This will leak when the program is closed down

metrics/sinks/file/FileSink.cpp
```diff
@@ -0,0 +1,94 @@
...
+        unsigned seconds = pSettingsTree->getPropInt("@period");
+        periodSeconds = std::chrono::seconds(seconds);
+    }
+
+    pSettingsTree->getProp("@filename", fileName);
+    bool clearFile = pSettingsTree->getPropBool("@clear", false);
+
+    //
+    // Clear the file if indicated.
+    if (clearFile)
```
ghalliday, line 43: Should use the jlib functions for managing files.

metrics/sinks/file/FileSink.hpp
```diff
@@ -0,0 +1,42 @@
...
+#include "jptree.hpp"
+#include "jstring.hpp"
+#include <thread>
+
+using namespace hpccMetrics;
+
+class METRICS_API FileMetricSink : public MetricSink
+{
+    public:
+
```
ghalliday, line 27: indent: extra newlines after public and protected

metrics/sinks/file/FileSink.hpp
```diff
@@ -0,0 +1,42 @@
...
+    protected:
+        void writeMeasurementsToFile(const MeasurementVector &values) const;
+        void collectionThread() const;
+
+    protected:
+
+        StringBuffer fileName;
+        std::chrono::seconds periodSeconds;
+        std::thread collectThread;
+        bool stopCollectionFlag = false;
```
ghalliday, line 41: should be atomic if accessed by more than one thread

metrics/sinks/file/FileSink.cpp
```diff
@@ -0,0 +1,94 @@
...
+    pReporter = _pReporter;
+    collectThread = std::thread(&FileMetricSink::collectionThread, this);
+}
+
+
+void FileMetricSink::collectionThread() const
+{
+    while (!stopCollectionFlag)
+    {
+        std::this_thread::sleep_for(periodSeconds);
```
ghalliday, line 65: better to use an interruptable semaphore (jisem.hpp) - otherwise the program has to wait for the timeout to expire before it can terminate.

metrics/sinks/file/FileSink.cpp
```diff
@@ -0,0 +1,94 @@
...
+void FileMetricSink::stopCollection()
+{
+    stopCollectionFlag = true;
+    collectThread.join();
+}
+
+
+void FileMetricSink::writeMeasurementsToFile(const MeasurementVector &values) const
+{
+    auto fh = fopen(fileName.str(), "a");
```
ghalliday, line 86: minor: Should probably use jfile functions.
Is it correct to open and close the file each time?
Minor comments since this is unlikely to be used for anything other than debugging.

metrics/testing/metrictest.cpp
```diff
@@ -0,0 +1,171 @@
+#include <cstdio>
```
ghalliday, line 1: This file should be cleaned up so it only contains tests for the current code.  It should also probably integrate into the cppunit unit test framework - but that can be done as a different pr/milestone.

metrics/sinks/file/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+        FileSink.cpp
+        )
+
+include_directories(
+        ${HPCC_SOURCE_DIR}/metrics/framework
+        ${HPCC_SOURCE_DIR}/system/include
+        ${HPCC_SOURCE_DIR}/system/jlib
+)
+
+add_definitions( -DMETRICSDLL)
```
ghalliday, line 29: There should be an export definition for each separate dll - this is separate from the metrics dll.
More normal is a definition of the form _D<NAME>_EXPORTS

metrics/sinks/file/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+################################################################################
+
+project(hpccmetrics_filesink)
+
+set ( srcs
+        FileSink.cpp
```
ghalliday, line 20: minor: most of our source files do not start with an upper case letter.

metrics/framework/Metrics.hpp
```diff
@@ -0,0 +1,305 @@
...
+        void getReportValues(MeasurementVector &mv) const override;
+
+
+    protected:
+        std::atomic<float> gaugeValue{0};
+};
+
+
+class MetricsReporter;
+interface IMetricSink
```
ghalliday, line 235: Should either be derived from IInterface, or have a virtual destructor, otherwise it can't be cleaned up.

ghalliday (changes requested):
@kenrowland I've gone through and added some initial comments.
My main comment is that the intialization and creation code does not look like it is in the correct place.  The configuration is not coming from the esp configuration file, and there aren't any updates to the helm file definition/schema to reflect the configuration.
(E.g. I would expect a singleton instance of the metrics registry in a base library.)

metrics/framework/Metrics.hpp
kenrowland: Not sure I understand. The MetricSink class which provides a base class for all sinks to inherit, implements the IMetricSink interface and provides a virtual destructor.

metrics/sinks/file/CMakeLists.txt
kenrowland: Changed to lowercase names for consistency.

metrics/sinks/file/CMakeLists.txt
kenrowland: Changed name of export define and added separate one for file sink.

esp/platform/txsummary.hpp
```diff
@@ -26,6 +26,9 @@
 #include "esphttp.hpp"
 #include <list>
 #include <map>
+#include "../../protocols/http/espmetrics.hpp"
```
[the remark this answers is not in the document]
kenrowland: Moved the file for now. Going to discuss instrumentation with Tony.

metrics/framework/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+################################################################################
+
+project(hpccmetrics)
+
+
+set ( srcs
+        Metrics.cpp
```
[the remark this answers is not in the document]
kenrowland: During development is it OK if we keep it separate? Then move to jlib later if it makes sense.

metrics/framework/Metrics.hpp
```diff
@@ -0,0 +1,305 @@
...
+#else
+#define METRICS_API DECL_IMPORT
+#endif
+
+namespace hpccMetrics {
+
+/*
+ * Enumerates the type of the value in a measurement
+ */
+enum ValueType {
```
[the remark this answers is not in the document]
kenrowland: The purpose for the value type is in case a sink has to deserialize a value reported by a metric. Or, should it need to add a suffix to a reported metric name. The type was added after discussions with Rigo in the earlier stages of the framework development.

metrics/framework/Metrics.hpp
```diff
@@ -0,0 +1,305 @@
...
+/*
+ * MeasurementBase
+ *
+ * Base class representing a measurement from a metric. Base class allows creating a vector
+ * of measurements. Class not intended to be instantiated.
+ */
+class MeasurementBase
+{
+    public:
+        std::string getName() const { return name; }
```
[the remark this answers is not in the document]
kenrowland: Changed. Also changed the getDescription.

metrics/framework/Metrics.hpp
```diff
@@ -0,0 +1,305 @@
...
+        std::string getDescription() const { return description; }
+        ValueType getValueType() { return(valueType); }
+        virtual std::string valueToString() const = 0;
+
+    protected:
+        MeasurementBase(std::string _name, ValueType _type, std::string _description) :
+                name{std::move(_name)},
```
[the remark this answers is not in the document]
kenrowland: The sink receives a list of metrics from which measurements are extracted. For simple metrics, the name is simply the metric name and the pair (metric, value) makes sense. However, some future metrics, such as a histogram, are likely to return multiple measurements, each with a unique name. This is the reason they are not paired. Keep in mind that the sink is now responsible for calling each metric to get its measurements, so it in essence has the paring metric available.

metrics/framework/Metrics.hpp
```diff
@@ -0,0 +1,305 @@
...
+        std::string description;
+        ValueType valueType;
+};
+
+/*
+ * Measurement
+ *
+ * Templated class that holds a measurement from a metric.
+ */
+template<typename T>
```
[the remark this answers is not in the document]
kenrowland: Prometheus supports full 64bit floating point numbers for gauges

metrics/framework/Metrics.hpp
```diff
@@ -0,0 +1,305 @@
...
+ * IMetric
+ *
+ * Interface defining a metric
+ */
+interface IMetric
+{
+    /*
+     * Returns the metric name
+     */
+    virtual std::string getName() const = 0;
```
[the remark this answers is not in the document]
kenrowland: Changed.

metrics/framework/Metrics.hpp
```diff
@@ -0,0 +1,305 @@
...
+        MetricType metricType;
+};
+
+class METRICS_API CounterMetric : public Metric
+{
+    public:
+
+        CounterMetric(const char *name, const char *description) :
+                Metric{name, description, MetricType::METRICS_GAUGE}  { }
+        ~CounterMetric() override = default;
```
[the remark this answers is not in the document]
kenrowland: Probably not. Removed.

metrics/framework/Metrics.hpp
kenrowland: Gauges need to support floating point numbers. I know atomic floats are not terribly efficient. I thought about using a fixed decimal representation where the value is 1000 times the actual in order to give 3 decimal places. We should discuss.

metrics/framework/Metrics.cpp
kenrowland: I'm not sure why either. It was copied from an older implementation since the mechanics of what it did has not changed. I moved the code to a single static method in the reporter class.

metrics/framework/Metrics.cpp
kenrowland: This was added during review with Rigo. The thought was that you may package a number of different features into a single lib and this gives a way to extract the sink functionality.

metrics/framework/Metrics.cpp
kenrowland: Added to design doc. Should we decide to not allow overriding of defaults for these values, we can remove them.

metrics/framework/Metrics.cpp
rpastrana: I'm not sure if this is one of the suggestions I provided; but yes I pushed a few enhancements related to plugin library nomenclature flexibility which resolved issues often encountered in practice (mostly in the pluggable security framework). If the criteria is to trim out any/all unnecessary logic, we can certainly employ a strict protocol regarding library names, etc.

metrics/framework/Metrics.hpp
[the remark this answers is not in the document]
ghalliday: But what results that we are going to report will require floating point?  It doesn't matter if the sinks support it if we are never going to generate it.

metrics/framework/Metrics.hpp
ghalliday: yes, it needs discussing.

metrics/framework/Metrics.hpp
ghalliday: Your factory function returns an IMetricSink.  That means you cannot cleanly dispose of the object returned from the function.

metrics/framework/Metrics.cpp
ghalliday: If there is a clear description of the problem it is solving, and that problem seems reasonable, then it is fine.  We load plugins in lots of other places and don't have alternatives.
Since type is going in a helm chart it should be something descriptive, and probably some kind of calculated filename.

metrics/framework/Metrics.hpp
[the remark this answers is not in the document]
ghalliday: The types the framework reports are determined by the metrics we want to report from our platform.  There should need to be a strong reason why we need anything other than integer.

kenrowland:
@afishbeck, @rpastrana, @ghalliday  Please review.

kenrowland:
To enable metrics for ESP, you must add the following section in environment.xml:

`<Environment>
  <Software>
     <metrics name="mymetricsconfig">
       <sinks>
         <sink name="default" type="filesink">
           <settings filename="/var/log/HPCCSystems/espmetrics.txt" clear="true" period="5">
           </settings>
         </sink>
       </sinks>
     </metrics>
  </Software>
</Environment>
`
This is an example section.

esp/platform/txsummary.cpp
```diff
@@ -34,6 +35,7 @@ bool operator < (const StringAttr& a, const StringAttr& b)
 CTxSummary::CTxSummary(unsigned creationTime)
 : m_creationTime(creationTime ? creationTime : msTick())
 {
+    getESPContainer()->getEspMetrics()->pCountRequests->inc(1);
```
afishbeck, line 38: As we've discussed offline, this code could be running outside of ESP in which case getESPContainer() will return nullptr.  You said you wanted to avoid an if check each time.  If that's the case you can wrap the metrics object globally and have it stubbed out when getESPContainer() is null.  Then this call would be a no-op for that scenario.

ghalliday (changes requested):
@kenrowland I added a few minor comments from a quick scan.  

My main comment is that the structure is still wrong.  In particular the class EspMetrics shouldn't exist.

There is no logical requirement for the metrics that are being gathered by the component to be defined in the same place as the class responsible for collecting them.  The metric should be defined and instantiated by the code that will be incrementing it.  Your current structure would not allow a metric to be shared by multiple components (possible examples would be active dali connections, socket/memory/file metrics).  Nor would it allow a metric in a base dll (e.g. jlib) to be reported by the esp component.

I think there are two possible structures that solve the problem:

1. recursive gather.

The esp main() code would instantiate an instance of MetricsReporter.  It would initialize it, and call functions to create any necessary metrics, passing the target MetricsReporter.  Then it would start reporting.

That ensures the metrics can be created by the code that updates them.  Registering is not so clean because it requires esp to know what functions to call to collate the metrics.  It is breaking the encapsulation of which metrics are being gathered.  It does however allow the same metrics to be reported by multiple components, and allows child dlls to be able to add metrics to a component.  I suspect it could be problematic to add dynamic metrics (since the information about the reporter would need to be passed around).

2. Singleton MetricsReporter

This uses a singleton object defined in a base dll (i.e. jlib).  The esp main code calls a function to initialize it (with the configuration) and another to start reporting metrics.  When a metric is created, the code that creates it registers it with the singleton metrics reporter.  This completely decouples the metrics from the component - metrics can be added without having to change any code in esp.  If a metric is being gathered it will automatically be reported by any component that is reporting metrics.  It supports dynamic metrics without any changes.  If a component is not reporting metrics then registering a metric is harmless.

The only complication I can think of for this second approach is that it needs to support metrics being registered before the MetricsReporter has been initialized with the context.  For instance static objects may be created before main() is called.  This ordering may need to be taken into account when filtering is supported.  It will also have an implication for configuring histogram ranges.

If you haven't guessed (2) is my strong preference, and one of the reasons for indicating your metric code should be in jlib.

ghalliday (changes requested):
@kenrowland I still don't think the structure is correct or that the new registry is a good idea.

I have added a few comments on the code, but I have not looked at it in detail because it is more important to get the structure right first.

system/jlib/jmetrics.cpp
```diff
@@ -0,0 +1,243 @@
...
+
+#include "jmetrics.hpp"
+#include "jmutex.hpp"
+#include "jlog.hpp"
+
+using namespace hpccMetrics;
+
+static Singleton<MetricsReporter> metricsReporter;
+static Singleton<MetricsRegistry> metricsRegistry;
+MODULE_INIT(INIT_PRIORITY_JLOG)
```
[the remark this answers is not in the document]
kenrowland: Please suggest the correct priority.

esp/bindings/http/platform/httpprot.cpp
```diff
@@ -455,13 +459,21 @@ bool CHttpThread::onRequest()
...
     initThreadLocal(sizeof(t), &t);
 
     m_httpserver = httpserver;
     httpserver->setSocketReturner(this);
     httpserver->setIsSSL(m_is_ssl);
     httpserver->setShouldClose(m_shouldClose);
+
+    //
+    // Update request count
+    if (isMetricsInitialized())
```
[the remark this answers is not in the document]
kenrowland: Let's have a call with Tony. It appeared to me that other components may use this class.

kenrowland:
Abandoning in favor of new PR that is cleaner

PR #14540, opened 2021-02-03
HPCC-25274 Rationalise secret categories in the Helm charts

shamser:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/helm/hpcc/templates/_helpers.tpl
+++ b/helm/hpcc/templates/_helpers.tpl
@@ -311,12 +311,12 @@ Pass in dict with root
 
 {{/*
 Add the secret volume mounts for a component
+Pass in dict with root and secretsCategories
 */}}
 {{- define "hpcc.addSecretVolumeMounts" -}}
-{{- $component := .component -}}
-{{- $categories := .categories -}}
+{{- $secretsCategories := .secretsCategories -}}
 {{- range $category, $key := .root.Values.secrets -}}
- {{- if (has $category $categories) -}}
+ {{- if (has $category $secretsCategories) -}}
 {{- range $secretid, $secretname := $key -}}
 - name: secret-{{ $secretid }}
   mountPath: /opt/HPCCSystems/secrets/{{ $category }}/{{ $secretid }}
@@ -327,17 +327,18 @@ Add the secret volume mounts for a component
 
 {{/*
 Add Secret volume for a component
+Pass in dict with root and secretsCategories
 */}}
 {{- define "hpcc.addSecretVolumes" -}}
 {{- $component := .component -}}
-{{- $categories := .categories -}}
+{{- $secretsCategories := .secretsCategories -}}
 {{- range $category, $key := .root.Values.secrets -}}
-{{- if (has $category $categories) -}}
+{{- if (has $category $secretsCategories) -}}
 {{- range $secretid, $secretname := $key }}
 - name: secret-{{ $secretid }}
   secret:
     secretName: {{ $secretname }}
-{{ end -}}
+{{- end -}}
 {{- end -}}
 {{- end -}}
 {{- end -}}
@@ -369,10 +370,10 @@ readinessProbe:
 Generate vault info
 */}}
 {{- define "hpcc.generateVaultConfig" -}}
-{{- $categories := .categories -}}
+{{- $secretsCategories := .secretsCategories -}}
 vaults:
 {{- range  $categoryname, $category := .root.Values.vaults -}}
- {{- if (has $categoryname $categories) }}
+ {{- if (has $categoryname $secretsCategories) }}
   {{ $categoryname }}:
   {{- range $vault := . }}
     - name: {{ $vault.name }}
@@ -670,7 +671,7 @@ Generate vault info
 
 {{/*
 Sasha configmap
-Pass in dict with root and me
+Pass in dict with root, me and secretsCategories
 */}}
 {{- define "hpcc.sashaConfigMap" -}}
 {{- $configMapName := printf "sasha-%s" .me.name -}}
@@ -683,11 +684,7 @@ data:
     sasha:
 {{ toYaml (omit .me "logging") | indent 6 }}
 {{- include "hpcc.generateLoggingConfig" . | indent 6 }}
-{{- $categories := list "system" -}}
-{{- if has "data" .me.access -}}
-{{- $categories := append $categories "storage" -}}
-{{- end }}
-{{ include "hpcc.generateVaultConfig" (dict "root" .root "categories" $categories ) | indent 6 }}
+{{ include "hpcc.generateVaultConfig" . | indent 6 }}
 {{- if .me.storage }}
       storagePath: {{ include "hpcc.getVolumeMountPrefix" (dict "root" .root "name" (printf "sasha-%s" .me.name) "me" .me.storage) }}
 {{- end }}
@@ -720,7 +717,7 @@ A template to generate Sasha service containers
 
 {{/*
 A template to generate Sasha service
-Pass in dict with root, me
+Pass in dict with root and me
 */}}
 {{- define "hpcc.addSashaVolumeMounts" }}
 {{- $serviceName := printf "sasha-%s" .me.name -}}
@@ -738,17 +735,13 @@ Pass in dict with root, me
 {{- end }}
 {{- if has "dll" .me.access }}
 {{ include "hpcc.addDllVolumeMount" . -}}
-{{- end }}
-{{ include "hpcc.addSecretVolumeMounts" (dict "root" .root "categories" (list "system" ) ) -}}
-{{- if has "data" .me.access -}}
-{{ include "hpcc.addSecretVolumeMounts" (dict "root" .root "categories" (list "storage" ) ) -}}
-{{- end }}
 {{- end -}}
+{{- end }}
 
 
 {{/*
 A template to generate Sasha service
-Pass in dict with root, me
+Pass in dict with root and me
 */}}
 {{- define "hpcc.addSashaVolumes" }}
 {{- $serviceName := printf "sasha-%s" .me.name -}}
--- a/helm/hpcc/templates/dali.yaml
+++ b/helm/hpcc/templates/dali.yaml
@@ -2,6 +2,8 @@
 {{- if not $dali.disabled -}}
 {{- $commonCtx := dict "root" $ "me" $dali }}
 {{- $sashaServices := $dali.services | default dict -}}
+{{- $daliSecretsCategories := list "system" }}
+{{- $tmpDaliScope := dict "aggregateSashaSecretsCategories" list "aggregateSashaAccess" list -}}
 apiVersion: apps/v1
 kind: Deployment
 metadata:
@@ -35,43 +37,44 @@ spec:
         volumeMounts:
 {{ include "hpcc.addConfigMapVolumeMount" $dali | indent 8 }}
 {{ include "hpcc.addDaliVolumeMount" $commonCtx | indent 8 }}
-{{ include "hpcc.addSecretVolumeMounts" (dict "root" $ "categories" (list "system" ) ) | indent 8 }}
+{{ include "hpcc.addSecretVolumeMounts" (dict "root" $ "secretsCategories" $daliSecretsCategories) | indent 8 }}
 {{- if not $sashaServices.disabled -}}
-{{- range $sashaName, $_sasha := $sashaServices -}}
+{{- range $sashaName, $_sasha := $dali.services -}}
 {{- $sasha := ($_sasha | default dict) -}}
 {{- if not $sasha.disabled -}}
 {{- $_ := set $sasha "name" $sashaName -}}
-{{- with ($sasha | merge (dict "access" (splitList " " (include "hpcc.getSashaServiceAccess" $sasha)))) }}
+{{- $sashaAccess := splitList " " (include "hpcc.getSashaServiceAccess" $sasha) -}}
+{{- $sashaServiceSecretsCategories := append ((or (has "data" $sashaAccess) (has "dalidata" $sashaAccess)) | ternary (list "storage") list) "system" -}}
+{{- $_ := set $tmpDaliScope "aggregateSashaAccess" ((concat $tmpDaliScope.aggregateSashaAccess $sashaAccess) | uniq) -}}
+{{- $_ := set $tmpDaliScope "aggregateSashaSecretsCategories" ((concat $tmpDaliScope.aggregateSashaSecretsCategories $sashaServiceSecretsCategories) | uniq) -}}
+{{- with ($sasha | merge (dict "access" $sashaAccess)) -}}
 {{ include "hpcc.addSashaContainer" (dict "root" $ "me" .) | indent 6 }}
         volumeMounts:
 {{- include "hpcc.addSashaVolumeMounts" (dict "root" $ "me" .) | indent 8 }}
+{{ include "hpcc.addSecretVolumeMounts" (dict "root" $ "secretsCategories" $sashaServiceSecretsCategories) | indent 8 }}
 {{- end }}
 {{- end }}
 {{- end }}
 {{- end }}
       volumes:
 {{ include "hpcc.addConfigMapVolume" $dali | indent 6 }}
-{{- $tmpDaliScope := dict -}}
-{{- $_ := set $tmpDaliScope "sashaAccess" list -}}
 {{- if not $sashaServices.disabled -}}
 {{- range $sashaName, $_sasha := $sashaServices -}}
 {{- $sasha := ($_sasha | default dict) -}}
 {{- if not $sasha.disabled -}}
 {{- $_ := set $sasha "name" $sashaName -}}
-{{- $_ := set $tmpDaliScope "sashaAccess" (concat $tmpDaliScope.sashaAccess (splitList " " (include "hpcc.getSashaServiceAccess" $sasha))) -}}
 {{- include "hpcc.addSashaVolumes" (dict "root" $ "me" $sasha) | indent 6 }}
 {{- end }}
 {{- end }}
 {{- end }}
 {{ include "hpcc.addDaliVolume" $commonCtx | indent 6 }}
-{{- if has "data" $tmpDaliScope.sashaAccess }}
+{{- if has "data" $tmpDaliScope.aggregateSashaAccess }}
 {{ include "hpcc.addDataVolume" $commonCtx | indent 6}}
-{{ include "hpcc.addSecretVolumes" (dict "root" $ "categories" (list "storage" ) ) | indent 6 }}
 {{- end }}
-{{- if has "dll" $tmpDaliScope.sashaAccess }}
+{{- if has "dll" $tmpDaliScope.aggregateSashaAccess }}
 {{ include "hpcc.addDllVolume" $commonCtx | indent 6 -}}
-{{- end }}
-{{ include "hpcc.addSecretVolumes" (dict "root" $ "categories" (list "system" ) ) | indent 6 }}
+{{- end -}}
+{{- include "hpcc.addSecretVolumes" (dict "root" $ "secretsCategories" $tmpDaliScope.aggregateSashaSecretsCategories) | indent 6 }}
 ---
 kind: ConfigMap 
 apiVersion: v1 
@@ -83,7 +86,7 @@ data:
     dali:
 {{ toYaml (omit $dali "logging" "services") | indent 6 }}
 {{- include "hpcc.generateLoggingConfig" $commonCtx | indent 6 }}
-{{ include "hpcc.generateVaultConfig" (dict "root" $ "categories" (list "system" ) ) | indent 6 }}
+{{ include "hpcc.generateVaultConfig" (dict "root" $ "secretsCategories" $daliSecretsCategories ) | indent 6 }}
     global:
 {{ include "hpcc.generateGlobalConfigMap" $ | indent 6 }}
 ---
@@ -94,7 +97,8 @@ data:
 {{- $_ := set $sasha "name" $sashaName -}}
 kind: ConfigMap
 {{- with ($sasha | merge (pick $dali "logging") | merge (dict "inDaliPod" true "access" (splitList " " (include "hpcc.getSashaServiceAccess" $sasha)))) }}
-{{ include "hpcc.sashaConfigMap" (dict "root" $ "me" .) }}
+{{- $sashaSecretsCategories := append ((or (has "data" .access) (has "dalidata" .access)) | ternary (list "storage") list) "system" }}
+{{ include "hpcc.sashaConfigMap" (dict "root" $ "me" . "secretsCategories" $sashaSecretsCategories ) }}
 {{- end }}
 ---
 {{- if $sasha.servicePort -}}
--- a/helm/hpcc/templates/eclagent.yaml
+++ b/helm/hpcc/templates/eclagent.yaml
@@ -16,7 +16,7 @@ data:
     {{ $apptype }}:
 {{ toYaml (omit .me "logging") | indent 6 }}
 {{- include "hpcc.generateLoggingConfig" . | indent 6 }}
-{{ include "hpcc.generateVaultConfig" (dict "root" .root "categories" (list "storage" "ecl" "ecl-user" ) ) | indent 6 }}
+{{ include "hpcc.generateVaultConfig" . | indent 6 }}
     global:
 {{ include "hpcc.generateGlobalConfigMap" .root | indent 6 }}
 {{- if not .me.useChildProcesses }}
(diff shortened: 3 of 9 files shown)


helm/hpcc/templates/eclagent.yaml
```diff
@@ -50,20 +50,22 @@ data:
...
+{{ include "hpcc.addSecretVolumes" (dict "root" .root "categories" .secretsCategoriesSystem ) | indent 10 }}
           restartPolicy: Never
       backoffLimit: 0
 {{- end }}
 {{- end -}}
 
 {{ range $.Values.eclagent -}}
 {{- if not .disabled -}}
-{{- $commonCtx := dict "root" $ "me" . }}
+{{- $secretsCategoriesUser := list "ecl-user" "ecl" "storage" }}
```
ghalliday, line 66: If these lists are different then the names should reflect why they are different e.g. Agent/Job rather than the items that they contain.  From the names I couldn't tell what distinguished them.

helm/hpcc/templates/localroxie.yaml
```diff
@@ -12,14 +12,16 @@ data:
...
+{{ include "hpcc.generateVaultConfig" (dict "root" .root "categories" .secretsCategories ) | indent 6 }}
     global:
 {{ include "hpcc.generateGlobalConfigMap" .root | indent 6 }}
 {{- end -}}
 
 {{ range $roxie := $.Values.roxie -}}
 {{- if not $roxie.disabled  -}}
-{{- $commonCtx := dict "root" $ "me" $roxie }}
+{{- $secretsCategoriesUser := list "storage" "ecl" "ecl-user" }}
+{{- $secretsCategoriesSystem := list "system" "ecl" "ecl-user" }}
```
ghalliday, line 23: I don't think it make sense for these to differ

helm/hpcc/templates/localroxie.yaml
```diff
@@ -12,14 +12,16 @@ data:
...
+{{- $commonCtx := dict "root" $ "me" $roxie "secretsCategoriesUser" $secretsCategoriesUser }}
```
ghalliday, line 24: Passing secretsCategoriesUser, but reading .secretsCategories

helm/hpcc/templates/roxie.yaml
```diff
@@ -15,14 +15,16 @@ data:
...
+{{ include "hpcc.generateVaultConfig" (dict "root" .root "categories" .secretsCategoriesUser ) | indent 6 }}
     global:
 {{ include "hpcc.generateGlobalConfigMap" .root | indent 6 }}
 {{- end -}}
 
 {{ range $roxie := $.Values.roxie -}}
 {{- if not $roxie.disabled -}}
-{{- $commonCtx := dict "root" $ "me" $roxie }}
+{{- $secretsCategoriesUser := list  "storage" "ecl" "ecl-user" }}
+{{- $secretsCategoriesSystem := list "system" "ecl" "storage" }}
```
ghalliday, line 26: I don't think it makes sense for these to differ - the list provided to the vault config should match the volume mounts.

helm/hpcc/templates/thor.yaml
```diff
@@ -18,14 +18,14 @@ data:
...
-{{ include "hpcc.generateVaultConfig" (dict "root" .root "categories" (list "storage" "ecl" "ecl-user" ) ) | indent 6 }}
+{{ include "hpcc.generateVaultConfig" (dict "root" .root "categories" .secretsCategoriesUser ) | indent 6 }}
     {{ $eclAgentType }}: # hthor or roxie
 {{ toYaml $hthorScope | indent 6 }}
       platform:
         type: "thor"
         width: {{ mul ($thorScope.numWorkers | default 1) ( $thorScope.channelsPerWorker | default 1) }}
 {{- include "hpcc.generateLoggingConfig" (dict "root" .root "me" $hthorScope ) | indent 6 }}
-{{ include "hpcc.generateVaultConfig" (dict "root" .root "categories" (list "storage" "ecl" "ecl-user" ) ) | indent 6 }}
+{{ include "hpcc.generateVaultConfig" (dict "root" .root "categories" .secretsCategoriesUser ) | indent 6 }}
```
ghalliday, line 28: @afishbeck Is this in the correct place?  I haven't checked properly, but at first glance it doesn't seem quite right.

helm/hpcc/templates/thor.yaml
```diff
@@ -189,7 +189,9 @@ data:
 
 {{ range $.Values.thor -}}
 {{- if not .disabled -}}
-{{- $commonCtx := dict "root" $ "me" . -}}
+{{- $secretsCategoriesUser := list  "storage" "ecl" "ecl-user" }}
```
ghalliday, line 192: I'm not sure if these do need to be different, but if they are they should be something like secretsCategoriesManager and secretsCategoriesAgent

ghalliday (changes requested):
@shamser a few comments.  Also:
* I think dali.yaml is missing the call to generate the Vault config.
* values.schema.json should be updated so the categories are consistent for vaults and secrets

helm/hpcc/templates/eclagent.yaml
```diff
@@ -50,20 +50,22 @@ data:
 {{ include "hpcc.addConfigMapVolumeMount" .me | indent 12 }}
 {{ include "hpcc.addDataVolumeMount" . | indent 12 }}
 {{ include "hpcc.addDllVolumeMount" .root | indent 12 }}
-{{ include "hpcc.addSecretVolumeMounts" (dict "root" .root "categories" (list "system" "ecl" "storage" ) ) | indent 12 }}
+{{ include "hpcc.addSecretVolumeMounts" (dict "root" .root "categories" .secretsCategoriesSystem ) | indent 12 }}
```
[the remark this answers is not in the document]
jakesmith: I'm still not entirely clear on the distinction logical between vault secrets and non vault secrets.
(and their delination here into secretsCategoriesSystem(non-vault?) and secretsCategoriesUser(vault?)

I understand that vault as it stands has a different structure, and it will enable secrets to be read by individual ecl users (hence "ecl-user"), however any component that needs a "system" or "ecl" or "storage" secret could be under either implementation..

helm/hpcc/templates/localroxie.yaml
jakesmith: why would secretsCategoriesSystem not have "storage" (for example)?

jakesmith: or why do both have "ecl-user" here, but in eclagent only secretsCategoriesUser does?

helm/hpcc/templates/localroxie.yaml
```diff
@@ -12,14 +12,16 @@ data:
     roxie:
 {{ toYaml (omit .me "logging") | indent 6 }}
 {{- include "hpcc.generateLoggingConfig" . | indent 6 }}
-{{ include "hpcc.generateVaultConfig" (dict "root" .root "categories" (list "storage" "ecl" "ecl-user" ) ) | indent 6 }}
+{{ include "hpcc.generateVaultConfig" (dict "root" .root "categories" .secretsCategories ) | indent 6 }}
```
jakesmith, line 15: I think this is mean to be secretsCategoriesUser?

jakesmith:
echoing what Gavin has said I think, but I think most components should have a list of categories of secrets, which is used uniformly and passed to the generator functions.
Or where sub-components need different secrets, e.g. agents vs engine, then they are defined in vars that are reused like 'agentSecrets' and 'workerSecrets'

helm/hpcc/templates/eclagent.yaml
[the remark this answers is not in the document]
shamser: @gavin should there be a single secretsCategories with ecl-user, ecl, storage and ecl-user?  Or should I be adding ecl-user to secretsCategoriesSystem?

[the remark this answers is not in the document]
ghalliday: I think you should have a single set of categories

"system" "ecl-user" "ecl" "storage"

You could possibly use a category set of "system" for the qwait component if child processes are not being used, but I doubt it is worth distinguishing.

"system" is required by all components (see description in values.yaml)

helm/hpcc/templates/localroxie.yaml
shamser: I've remove the need to secretsCategoriesUser - I'm using secretsCategories for everything now.

ghalliday (commented):
@shamser looks good.  I think dali.yaml also needs the same treatment

@jakesmith I think sasha.yaml is missing code to load secrets - all components should load system secrets.

jakesmith (changes requested):
> I think dali.yaml also needs the same treatment

Yes, some secret generation may have been missing before the sasha PR was merged and you rebased, but it now needs making consistent too.

> I think sasha.yaml is missing code to load secrets - all components should load system secrets.

it is there, but via _helpers.tpl, because sasha services are either generated in Dali pod, or in their own pod (see hpcc.sashaConfigMap).
So it also needs to work of a categories list.

@shamser

shamser:
> k sasha.yaml is missing code to load secrets - all components should load system secr

There was code to load the secrets in the helper functions: addSashaVolumeMounts and addSashaVolumeMounts.  I've updated the helper functions to use the dictionary item: secretsCategory.

ghalliday (dismissed):

ghalliday:
@jakesmith please can you sanity check.

helm/hpcc/templates/dali.yaml
```diff
@@ -1,5 +1,10 @@
 {{ range $.Values.dali -}}
 {{- if not .disabled -}}
+{{- $secretsCategories := list "system" }}
+{{- if or (has "data" .access) (has "dalidata" .access) -}}
```
jakesmith, line 4: this won't work, because "access" is set as an internal variable, per dali sasha service.
See : https://github.com/hpcc-systems/HPCC-Platform/blob/c28c89ecd71992ebeb90218292c5caf2f9f18fd7/helm/hpcc/templates/dali.yaml#L40

Each sasha service container has different a access set as returned by getSashaServiceAccess, which is different from Dali's access requirements (which only requires system).

I think needs to be changed so done here just for Dali, and inside the range .services loop for the other containers.

helm/hpcc/templates/sasha.yaml
```diff
@@ -1,6 +1,10 @@
 {{ range $.Values.sasha -}}
 {{- if not .disabled -}}
-{{- $commonCtx := dict "root" $ "me" . }}
+{{- $secretsCategories := list "system" }}
+{{- if or (has "data" .access) -}}
```
jakesmith, line 4: No or condition here.

jakesmith: Test on "access" is too early, it isn't set at this point (it's set on line 10).

jakesmith (changes requested):
@shamser couple of issues relating to "access" in Dali and Sasha

helm/hpcc/templates/thor.yaml
afishbeck: @ghalliday sorry didn't see this comment.  I'll check if this has changed, in later commits, but it looks correct.  This is using categories to know which vaults to give thor access to.  I'll look at the latest commit and see if it looks ok.

helm/hpcc/templates/sasha.yaml
```diff
@@ -3,7 +3,11 @@
 {{ if not $sasha.disabled -}}
 {{ $_ := set $sasha "name" $sashaName -}}
 {{ $_ := set $sasha "access" (splitList " " (include "hpcc.getSashaServiceAccess" $sasha)) -}}
-{{ $commonCtx := dict "root" $ "me" $sasha -}}
+{{ $secretsCategories := list "system" }}
+{{ if has "data" $sasha.access }}
+{{ $secretsCategories := append $secretsCategories "storage" }}
```
jakesmith, line 8: I think there are newline issues here, each of the above lines will introduce a newline.
Some newline suppression needed with {{- or -}} , being careful not to swallow up the preceding or final newline.

Try looking at the output of helm template (or can use startall.sh -t) in this area, to see newlines.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -761,6 +758,7 @@ Pass in dict with root, me
 {{- if has "dll" .me.access }}
 {{ include "hpcc.addDllVolume" . -}}
 {{- end }}
+{{ include "hpcc.addSecretVolumes" (dict "root" .root "categories" .secretsCategories) -}}
```
jakesmith, line 761: trivial: perhaps addSecretVolumes should use "secretsCategories" too, then you could just pass in . instead of constructing dict

helm/hpcc/templates/dali.yaml
```diff
@@ -34,15 +34,20 @@ spec:
...
 {{- range $sashaName, $_sasha := $dali.services -}}
 {{- $sasha := ($_sasha | default dict) -}}
+{{- $secretsCategories2 := list "system" }}
 {{- if not $sasha.disabled -}}
 {{- $_ := set $sasha "name" $sashaName -}}
 {{- with ($sasha | merge (dict "access" (splitList " " (include "hpcc.getSashaServiceAccess" $sasha)))) }}
+{{- if or (has "data" .access) (has "dalidata" .access) -}}
+{{- $secretsCategories2 := append $secretsCategories2 "storage" -}}
+{{- end }}
+{{ include "hpcc.addSecretVolumeMounts" (dict "root" $ "categories" $secretsCategories2 ) | indent 8 }}
```
jakesmith, line 47: this doesn't look right, it's now inside the range and will be repeated(duplicated) per sasha service.
(if you were to add 2 sasha services under dali.services, it would fail I think).

The line that was on line 37, was for the dali container volumeMounts (just "system").
The line that was on line 50, was for volumeMounts for each separate sasha service container.

helm/hpcc/templates/dali.yaml
```diff
@@ -34,15 +34,20 @@ spec:
         volumeMounts:
 {{ include "hpcc.addConfigMapVolumeMount" $dali | indent 8 }}
 {{ include "hpcc.addDaliVolumeMount" $commonCtx | indent 8 }}
-{{ include "hpcc.addSecretVolumeMounts" (dict "root" $ "categories" (list "system" ) ) | indent 8 }}
+{{- $secretsCategories := list "system" }}
 {{- range $sashaName, $_sasha := $dali.services -}}
 {{- $sasha := ($_sasha | default dict) -}}
+{{- $secretsCategories2 := list "system" }}
```
jakesmith, line 40: for clarity should rename this sashaServiceSecretCategories vs the one above, which I'd rename daliSecretsCategories

helm/hpcc/templates/dali.yaml
```diff
@@ -52,21 +57,24 @@ spec:
...
+{{- $ssecretsCategories := list "system" -}}
 {{- if not $sasha.disabled -}}
 {{- $_ := set $sasha "name" $sashaName -}}
 {{- $_ := set $tmpDaliScope "sashaAccess" (concat $tmpDaliScope.sashaAccess (splitList " " (include "hpcc.getSashaServiceAccess" $sasha))) -}}
-{{- include "hpcc.addSashaVolumes" (dict "root" $ "me" $sasha) | indent 6 }}
+{{- if has "data" $tmpDaliScope.sashaAccess }}
+{{- $ssecretsCategories := append $ssecretsCategories "storage" -}}
+{{- end }}
+{{- include "hpcc.addSashaVolumes" (dict "root" $ "me" $sasha "secretsCategories" $ssecretsCategories) | indent 6 }}
+{{- include "hpcc.addSecretVolumes" (dict "root" $ "categories" $ssecretsCategories ) | indent 6 }}
```
jakesmith, line 68: These have also been moved into the range, which is going to cause problems if >1 service defined.
They should be called once the sashaAccess list has been formulated, after the range as before.

helm/hpcc/templates/dali.yaml
```diff
@@ -52,21 +57,24 @@ spec:
 {{- $_ := set $tmpDaliScope "sashaAccess" list -}}
 {{- range $sashaName, $_sasha := $dali.services -}}
 {{- $sasha := ($_sasha | default dict) -}}
+{{- $ssecretsCategories := list "system" -}}
```
jakesmith, line 60: not very clear var. name, maybe aggregateSecretCategories

helm/hpcc/templates/dali.yaml
```diff
@@ -78,7 +86,7 @@ data:
     dali:
 {{ toYaml (omit $dali "logging" "services") | indent 6 }}
 {{- include "hpcc.generateLoggingConfig" $commonCtx | indent 6 }}
-{{ include "hpcc.generateVaultConfig" (dict "root" $ "categories" (list "system" ) ) | indent 6 }}
+{{ include "hpcc.generateVaultConfig" (dict "root" $ "categories" $secretsCategories ) | indent 6 }}
```
jakesmith, line 89: related to other comment re. var nameing, will be clearer if renamed daliSecretsCategoreis

helm/hpcc/templates/eclagent.yaml
```diff
@@ -16,7 +16,7 @@ data:
     {{ $apptype }}:
 {{ toYaml (omit .me "logging") | indent 6 }}
 {{- include "hpcc.generateLoggingConfig" . | indent 6 }}
-{{ include "hpcc.generateVaultConfig" (dict "root" .root "categories" (list "storage" "ecl" "ecl-user" ) ) | indent 6 }}
+{{ include "hpcc.generateVaultConfig" (dict "root" .root "categories" .secretsCategories ) | indent 6 }}
```
jakesmith, line 19: I see "secretsCategories passed into dict when hpcc.agentConfigMap called, but comment at top of hpcc.agentConfigMap should be updated.

helm/hpcc/templates/eclccserver.yaml
```diff
@@ -14,7 +14,7 @@ data:
 {{- include "hpcc.generateLoggingConfig" . | indent 6 }}
       queues:
 {{ include "hpcc.generateConfigMapQueues" .root | indent 6 }}
-{{ include "hpcc.generateVaultConfig" (dict "root" .root "categories" (list "codeVerify") ) | indent 6 }}
+{{ include "hpcc.generateVaultConfig" (dict "root" .root "categories" .secretsCategories ) | indent 6 }}
```
jakesmith, line 17: probably hpcc.*configMap* func comment needs updating to say "secretsCategories" now expected/passed in.

jakesmith (changes requested):
@shamser - a couple of issues with the sasha services / mounts/volumes in ranges.

I added a couple of comments re. helper configmap's now needing/taking "secretsCategories", and the comments need updating, I didn't check/comment on, but suspect that's true in most of them.

helm/hpcc/templates/sasha.yaml
shamser: This section didn't generate extra newlines.  However, I have added {{- and -}} to these lines.  I have found the helper functions creating extra new lines.  I have fixed the ones related to secrets.  However, there are many other helper functions that do generate newlines unnecessarily.  I've created a ticket to address these: https://track.hpccsystems.com/browse/HPCC-25037

jakesmith: ok, thanks.
I think these lines didn't generate a newline because of the pre-existing line -}} on line 5 chomped the newlines they'd otherwise introduce.

richardkchapman:
@shamser Needs rebasing for a clean merge

jakesmith:
I've found a few problems when teting enabling some secrets relating to dali and sasha services.
Will discuss offlie.

jakesmith:
> I've found a few problems when teting enabling some secrets relating to dali and sasha services.
> Will discuss offlie.

For reference: https://github.com/shamser/HPCC-Platform/pull/5/files

shamser:
I have tested the change you suggested.   It looks good.  Shall I squash?

jakesmith (commented):
>Shall I squash?

@shamser - please and I'll scan one last time.

helm/hpcc/templates/dali.yaml
```diff
@@ -2,6 +2,10 @@
 {{- if not $dali.disabled -}}
 {{- $commonCtx := dict "root" $ "me" $dali }}
 {{- $sashaServices := $dali.services | default dict -}}
+{{- $daliSecretsCategories := list "system" }}
+{{- $tmpDaliScope := dict -}}
+{{- $_ := set $tmpDaliScope "aggregateSashaSecretsCategories" list -}}
+{{- $_ := set $tmpDaliScope "aggregateSashaAccess" list -}}
```
jakesmith, line 8: we could change above 3 lines to : 
```
{{- $tmpDaliScope := dict "aggregateSashaSecretsCategories" list "aggregateSashaAccess" "list -}}
```

helm/hpcc/templates/sasha.yaml
```diff
@@ -2,12 +2,16 @@
...
-{{ $_ := set $sasha "access" (splitList " " (include "hpcc.getSashaServiceAccess" $sasha)) -}}
-{{ $commonCtx := dict "root" $ "me" $sasha -}}
-{{ $configSHA := include "hpcc.sashaConfigMap" $commonCtx | sha256sum -}}
-{{ $serviceName := printf "sasha-%s" $sashaName -}}
+{{- if not $sasha.disabled -}}
+{{- $_ := set $sasha "name" $sashaName -}}
+{{- $_ := set $sasha "access" (splitList " " (include "hpcc.getSashaServiceAccess" $sasha)) -}}
+{{- $secretsCategories := list "system" }}
+{{- if or (has "data" $sasha.access) (has "dalidata" $sasha.access) -}}
+{{- $secretsCategories := append $secretsCategories "storage" -}}
```
jakesmith, line 10: I don't think this is going to work, because it's going to treat this $secretsCategories as scoped variable of the if block

To check, the "sasha-wu-archiver " Deployment should have storage secrets generated.

This is similar to the change needed in dali.yaml, where I avoided an if with a ternary clause:
```
{{- $sashaServiceSecretsCategories := append ((or (has "data" $sashaAccess) (has "dalidata" $sashaAccess)) | ternary (list "storage") list) "system" -}}
```

I think this would become:
```
{{- $secretsCategories := append ((or (has "data" $sasha.access) (has "dalidata" $sasha.access)) | ternary (list "storage") list) "system" -}}
```

jakesmith (changes requested):
@shamser - I think there's 1 issue

helm/hpcc/templates/dali.yaml
```diff
@@ -46,8 +44,9 @@ spec:
 {{- if not $sasha.disabled -}}
 {{- $_ := set $sasha "name" $sashaName -}}
 {{- $sashaAccess := splitList " " (include "hpcc.getSashaServiceAccess" $sasha) -}}
-{{- $_ := set $tmpDaliScope "aggregateSashaAccess" ((concat $tmpDaliScope.aggregateSashaAccess $sashaAccess) | uniq) -}}
 {{- $sashaServiceSecretsCategories := append ((or (has "data" $sashaAccess) (has "dalidata" $sashaAccess)) | ternary (list "storage") list) "system" -}}
+{{- $sashaServiceSecretsCategories := list -}}
```
jakesmith, line 48: this line should be removed (it's overwriting the value of previous line.)

jakesmith (changes requested):
@shamser - spotted 1 issue.

jakesmith (commented):
looks good now I think, please squash

helm/hpcc/templates/sasha.yaml
```diff
@@ -2,12 +2,14 @@
...
-{{ if not $sasha.disabled -}}
-{{ $_ := set $sasha "name" $sashaName -}}
-{{ $_ := set $sasha "access" (splitList " " (include "hpcc.getSashaServiceAccess" $sasha)) -}}
-{{ $commonCtx := dict "root" $ "me" $sasha -}}
-{{ $configSHA := include "hpcc.sashaConfigMap" $commonCtx | sha256sum -}}
-{{ $serviceName := printf "sasha-%s" $sashaName -}}
+{{- if not $sasha.disabled -}}
+{{- $_ := set $sasha "name" $sashaName -}}
+{{- $_ := set $sasha "access" (splitList " " (include "hpcc.getSashaServiceAccess" $sasha)) -}}
+
```
jakesmith, line 8: trivial: extra new line introduced?

jakesmith (dismissed):
@shamser - looks good to merge - 1 trivial comment.

helm/hpcc/templates/sasha.yaml
shamser: I've amended the last commit, removing the extra line.

ghalliday (approved):

jakesmith (approved):

PR #14544, opened 2021-02-04
HPCC-25300 Avoid permission issues with kafka build

jakesmith:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:


jakesmith:
@dcamper - Please review.

This is a solution we discussed, and avoids the build failure problems folks have been seeing.
But there may be a more elegant solution, one that avoids this step more thoroughly if nothing has changed or needs building (particularly an issue when deliberately running make install as root)

dcamper:
@jakesmith The upshot of this change would be to rebuild librdkafka every time, even during make install, right?  That does not seem optimal.  Is there a way to test that the 'make install' is active?

jakesmith:
>The upshot of this change would be to rebuild librdkafka every time, even during make install, right? That does not seem optimal. Is there a way to test that the 'make install' is active?

I don't think that's the upshot.
The code without the final delete md5file, i.e. as it was, does:

1) If [the source dir has been copied before] then unconditionally create an md5 file (*NB: before copying over with new*)
2) [unconditionally] copy sources (I think an rsync would work better and could tell if anything had changed using it)
3) Run configure if not run before (btw, might that be problematic if options have changed, and configure is needed? Should it instead run if there were new files..)
4) Use the md5 file created in step 1 (before the latest copy) to check if the files still match (i.e. whether any copied over are different).

So the md5 file is recreated/overwritten every time if the source dir has been copied before.
The old one from a previous step is not needed in this scheme, so it can be deleted..

@dcamper

richardkchapman (approved):

richardkchapman:
@dcamper Given the chaos this is causing for developers I'm merging now to get into today's builds. Bit if you have suggestions for a better fix please do comment.

PR #14555, opened 2021-02-08
HPCC-25326 Remove some ambiguous xpath checking in release builds

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jptree.cpp
+++ b/system/jlib/jptree.cpp
@@ -1022,8 +1022,10 @@ aindex_t PTree::getChildMatchPos(const char *xpath)
     if (!childIter->first())
         return (aindex_t)-1;
     IPropertyTree &childMatch = childIter->query();
+#ifdef _DEBUG
     if (childIter->next())
         AMBIGUOUS_PATH("addPropX", xpath);
+#endif
 
     if (value)
         if (value->isArray())
@@ -1060,7 +1062,9 @@ void PTree::resolveParentChild(const char *xpath, IPropertyTree *&parent, IPrope
         if (this != &pathIter->query())
         {
             IPropertyTree *currentPath = NULL;
+#ifdef _DEBUG
             bool multiplePaths = false;
+#endif
             bool multipleChildMatches = false;
             for (;;)
             {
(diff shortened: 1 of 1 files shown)


ghalliday:
@jakesmith For discussion/review.  A few questions:
* I debated using a different #define to allow the checking to be enabled in release mode if desired.  Would that be better?
* There are a couple of other similar checks, should they also be disabled?
* Which build to target?  The vast majority of the time saved on the original example is in my other fix, but this could affect xml reading/graph parsing/anything that accesses a IPropertyTree.

richardkchapman:
Could argue this should go to 7.8?

ghalliday:
@richardkchapman I think safe enough - so re-targeted.

jakesmith:
The github smoketest failure is due to a bug in the smoketest.
The above result was triggered from when this PR targeted 7.12.x (really I think github should remove the non-7.12.x tests above from the list..)

But it failed because it triggered, and then the PR base changed, before the smoketest had started.
The action pull the branch by name, and got the 7.8.x version, which then failed because the smoketest isn't there (or in particular because one of the scripts it uses isn't there).

I'll open a separate JIRA to fix this smoketest issue.

jakesmith:
@ghalliday is the indexcachebug mismatch genuine / related to this PR?

jakesmith:
>I debated using a different #define to allow the checking to be enabled in release mode if desired. Would that be better?

I think it's sufficient to enable in DEBUG only.

> There are a couple of other similar checks, should they also be disabled?

yes, IMO, the other cases should also be disabled, they might not save as much, but better if it was consistent.

jakesmith (commented):
@ghalliday - see comments, do you want to change other AMBIG. cases so that they are checked in DEBUG only too in this PR?

ghalliday:
@jakesmith please check the new commit

jakesmith (approved):
@ghalliday - looks good

system/jlib/jptree.cpp
```diff
@@ -1083,8 +1087,10 @@ void PTree::resolveParentChild(const char *xpath, IPropertyTree *&parent, IPrope
             }
             if (!parent)
             {
+#ifdef _DEBUG
```
jakesmith, line 1090: @ghalliday - as the compile failure (in Release) highlights, relating to 'multiplePaths' not used,
the check on 1084-1085 can also be avoided (wrapped in a _DEBUG), along with the declaration.

ghalliday:
@jakesmith I think my change needs careful checking.

ghalliday:
@richardkchapman I have double checked and I cannot see any potential changes in behaviour (other than losing ambiguous errors)

PR #14570, opened 2021-02-11
HPCC-25345 Avoid losing write close exceptions

jakesmith:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/thorlcr/activities/indexwrite/thindexwriteslave.cpp
+++ b/thorlcr/activities/indexwrite/thindexwriteslave.cpp
@@ -41,6 +41,7 @@ class IndexWriteSlaveActivity : public ProcessSlaveActivity, public ILookAheadSt
     Owned<IPartDescriptor> partDesc, tlkDesc;
     IHThorIndexWriteArg *helper;
     Owned <IKeyBuilder> builder;
+    OwnedIFileIO builderIFileIO;
     Owned<IRowStream> myInputStream;
     Owned<IPropertyTree> metadata;
     Linked<IEngineRowAllocator> outRowAllocator;
@@ -182,8 +183,8 @@ public:
         if (metadata->getPropBool("_useTrailingHeader", true))
             flags |= USE_TRAILING_HEADER;
         unsigned twFlags = isUrl(partFname) ? TW_Direct : TW_RenameToPrimary;
-        OwnedIFileIO iFileIO = createMultipleWrite(this, partDesc, 0, twFlags, compress, NULL, this, &abortSoon);
-        Owned<IFileIOStream> out = createBufferedIOStream(iFileIO, 0x100000);
+        builderIFileIO.setown(createMultipleWrite(this, partDesc, 0, twFlags, compress, NULL, this, &abortSoon));
+        Owned<IFileIOStream> out = createBufferedIOStream(builderIFileIO, 0x100000);
         if (!needsSeek)
             out.setown(createNoSeekIOStream(out));
 
@@ -239,7 +240,12 @@ public:
         try 
         { 
             metadata.clear();
-            builder.clear(); 
+            builder.clear();
+            if (builderIFileIO)
+            {
+                builderIFileIO->close();
+                builderIFileIO.clear();
+            }
         }
         catch (IException *_e)
         {
--- a/thorlcr/activities/thactivityutil.cpp
+++ b/thorlcr/activities/thactivityutil.cpp
@@ -609,25 +609,15 @@ class CWriteHandler : implements IFileIO, public CInterface
     bool remote;
     CFIPScope fipScope;
     unsigned twFlags;
+    bool closed = false;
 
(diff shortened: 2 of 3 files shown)


jakesmith:
NB: the diffs in thorlcr/activities/thactivityutil.cpp are mostly code movement (code moved from beforeDispose into checkAndHandleClose)

thorlcr/activities/thactivityutil.cpp
```diff
@@ -692,6 +682,26 @@ class CWriteHandler : implements IFileIO, public CInterface
...
+        remote = !rfn.isLocal();
+        rfn.getPath(primaryName);
+        if (globals->getPropBool("@replicateAsync", true))
+            cancelReplicates(&activity, partDesc);
+    }
+    virtual void beforeDispose() override
+    {
+        // Can't throw in destructor...
+        // Note that if we do throw the CWriteHandler object is liable to be leaked...
+        checkAndHandleClose();
```
ghalliday, line 703: I'm thinking about this issue for a similar reason.

I'm not sure this change is safe - e.g., if this is being deleted because the owning object is being destroyed then you will still be throwing an exception within a destructor - and will cause the query to terminate.

jakesmith: I agree, with or without this change this code could cause the process to terminate.
Not quite the issue this change is making though.
It's giving a chance to directly call this code (via overridden close()) before it is indirectly called via beforeDispose/dtor.

Re. possible termination if this code is hit via dtor (because close() not called), as with ~CFileIO, it could be wrapped in a try/catch and log it and continue, but ... in some ways that's worse than terminating with a core and stack trace. You don't know the file(s) state if you continue.

If explicitly called now (as it should be) via close() - then this beforeDispose will be a nop.
@ghalliday

thorlcr/activities/indexwrite/thindexwriteslave.cpp
```diff
@@ -41,6 +41,7 @@ class IndexWriteSlaveActivity : public ProcessSlaveActivity, public ILookAheadSt
     Owned<IPartDescriptor> partDesc, tlkDesc;
     IHThorIndexWriteArg *helper;
     Owned <IKeyBuilder> builder;
+    OwnedIFileIO builderIFileIO;
```
ghalliday, line 44: This will fix the problem, but is the correct fix to add close() to IIOStream, and have builder->finish() call that?
(I am still trying to work out what the correct way of solving this problm is in general.)

thorlcr/activities/thactivityutil.cpp
ghalliday: On rereading I think it is an improvment on the previous code.
It is still potentially a problem throwing an exception from beforeDispose() - it will cause program termination if the Release is being called from a destructor (rather than an explicit clear()).  It would probably be safer to catch any exception and log it instead.

ghalliday (commented):
@jakesmith a couple of questions/comments.  I'm trying to get my head around what we should be doing in general for this.

thorlcr/activities/thactivityutil.cpp
jakesmith: This was supposed to be an improvement - I suspect there are other cases where ~CFileIO will lose exceptions on close.

It could still terminate, as you say, if dtor's via another route.
I agree, in this case, I think it would be better to catch and log and continue, because it should have already flagged the error.
I'll add that.

thorlcr/activities/indexwrite/thindexwriteslave.cpp
jakesmith: I don't think IFileIOStream should ever call the IFileIO::close it's working on 
There might be multiple IFileIOStreams using the same IFileIO..

Alternatively, there could be a variety of createKeyBuilder that was parsed a IFileIO, which it could then close, but I don't think that would be quite right either, the create of the IFileIO should close it.

jakesmith (commented):
@ghalliday - have added try/catch in CWriteHandler beforeDispose to avoid potential exception/termination.

mckellyln (commented):
I agree with the overall improvement and the try/catch in beforeDispose.
I was originally thinking of a try/catch in checkAndHandleClose().

jakesmith:
@ghalliday - squashed.

thorlcr/activities/indexwrite/thindexwriteslave.cpp
ghalliday: I'm not sure it ever makes much sense to have multiple streams writing to a file.  But this change is fine.

ghalliday (approved):

PR #14571, opened 2021-02-11
HPCC-25346 Add WsTopology.TpStoragePlaneQuery

wangkx:
The ESP method is added to report the information about storage planes
in a cloud environment.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/scm/ws_topology.ecm
+++ b/esp/scm/ws_topology.ecm
@@ -316,6 +316,14 @@ ESPStruct TpClusterNameType
     bool   IsDefault;
 };
 
+ESPStruct TpStoragePlane
+{
+    string Name;
+    string Hosts;
+    string Prefix;
+    unsigned NumDevices;
+};
+
 ESPrequest TpListTargetClustersRequest
 {
 };
@@ -646,7 +654,17 @@ ESPresponse [nil_remove, exceptions_inline] TpDropZoneQueryResponse
     ESParray<ESPstruct TpDropZone>    TpDropZones;
 };
 
-ESPservice [auth_feature("DEFERRED"), noforms, version("1.31"), cache_group("ESPWsTP"), exceptions_inline("./smc_xslt/exceptions.xslt")] WsTopology
+ESPrequest TpStoragePlaneQueryRequest
+{
+    string Name;
+};
+
+ESPresponse [nil_remove, exceptions_inline] TpStoragePlaneQueryResponse
+{
+    ESParray<ESPstruct TpStoragePlane> TpStoragePlanes;
+};
+
+ESPservice [auth_feature("DEFERRED"), noforms, version("1.32"), cache_group("ESPWsTP"), exceptions_inline("./smc_xslt/exceptions.xslt")] WsTopology
 {
     ESPmethod [cache_seconds(180), cache_global(1), resp_xsl_default("/esp/xslt/targetclusters.xslt")] TpTargetClusterQuery(TpTargetClusterQueryRequest, TpTargetClusterQueryResponse);
     ESPmethod [cache_seconds(180), cache_global(1), resp_xsl_default("/esp/xslt/topology.xslt")] TpClusterQuery(TpClusterQueryRequest, TpClusterQueryResponse);
@@ -669,6 +687,7 @@ ESPservice [auth_feature("DEFERRED"), noforms, version("1.31"), cache_group("ESP
     ESPmethod [cache_seconds(180), cache_global(1), min_ver(1.25)] TpMachineInfo(TpMachineInfoRequest, TpMachineInfoResponse);
 
     ESPmethod SystemLog(SystemLogRequest, SystemLogResponse);
+    ESPmethod [min_ver(1.32)] TpStoragePlaneQuery(TpStoragePlaneQueryRequest, TpStoragePlaneQueryResponse);
 };
 
 SCMexportdef(WSWU);
--- a/esp/services/ws_topology/ws_topologyService.cpp
+++ b/esp/services/ws_topology/ws_topologyService.cpp
@@ -1954,3 +1954,17 @@ bool CWsTopologyEx::onTpDropZoneQuery(IEspContext &context, IEspTpDropZoneQueryR
     }
     return false;
 }
+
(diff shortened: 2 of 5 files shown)


wangkx:
@GordonSmith ECLWatch UI may use the changes in this PR.

jakesmith:
>https://github.com/hpcc-systems/HPCC-Platform/pull/14571/checks?check_run_id=1883460497

It is not clear what causes the timeout that causes one of the smoketest sections to fail. Roxie appears to have run the query to completion. Unfortunately the regression suite didn't capture the stacks before it was aborted (see: https://track.hpccsystems.com/browse/HPCC-25363)

esp/scm/ws_topology.ecm
```diff
@@ -316,6 +316,15 @@ ESPStruct TpClusterNameType
     bool   IsDefault;
 };
 
+ESPStruct TpStoragePlan
```
jakesmith, line 319: should be "Planes" not "Plan" here and elsewhere.

esp/scm/ws_topology.ecm
```diff
@@ -316,6 +316,15 @@ ESPStruct TpClusterNameType
...
+{
+    string Name;
+    string Hosts;
+    string Prefix;
+    string Mount;
```
jakesmith, line 324: Not sure where this ('Mount') came from?
there the planes are mounted is really an internal matter.

esp/scm/ws_topology.ecm
```diff
@@ -316,6 +316,15 @@ ESPStruct TpClusterNameType
...
+    unsigned NumDevices;
```
jakesmith, line 325: I am not sure this is the information we should be pulling out of the configuration and exposing via a service.
(it may be)

We'll want to present a list of LandingZones I think, which will probably be a select list of storage planes,
but not sure if we want to expose all the planes and/or all their properties.

We might, I'm not sure, but I suspect there comes a point where you just want to view the configuration as a whole, without a strict EspStruct that is incomplete, or can go out of sync.
FWIW the source reference for a storage plane is : https://github.com/hpcc-systems/HPCC-Platform/blob/b1167eeee37e60ac72e51d20e11918ebb3294b78/helm/hpcc/values.schema.json#L303

This is also probably true of other config information.
Do we want a discreet service for each part that duplicates the schema's in esp struct responses, or customized them somehow, or do we only want very select parts for specialized reasons (like a list of LZ's)...
And then for detailed configuration details - embed or view them via a K8s type dashboard.

jakesmith: @ghalliday - can you comment?

esp/scm/ws_topology.ecm
wangkx: Plane or Planes?

esp/scm/ws_topology.ecm
wangkx: I saw it when I dry-run:
planes:
        - name: hpcc-data-plane
          mount: "/var/lib/HPCCSystems/hpcc-data"
        - name: hpcc-spill-plane
          mount: "/var/lib/HPCCSystems/hpcc-spill"

I can remove it since it is for internal.

jakesmith: right, yes it's generated, but used internally, don't think we'd want to return it to a client.

esp/scm/ws_topology.ecm
ghalliday: My question is what is the purpose of this service - I looked in the jira and couldn't see any reasons/requirements.

drop zones lists and targets for sprays should/will be provided by the standard calls.  Who is expected to calls this, and why?

wangkx: The purpose of this service method is to give a user some general information about storage planes. Operation team may check the information from ECLWatch or SOAP calls.

ghalliday: Does the user need to know about storage planes as a logical concept?  They need to know about dropzones (already returned from one function), and locations that data can be copied to/stored.  I would have thought there was a function to provide that as well already.

wangkx: Do we expect some users know about storage planes as a logical concept? Do we already mention the storage planes in our documents (like exportingData.rst)?
The existing TpDropZoneQuery only works for bare metal environment. I may upgrade it to support cloud environment. To do that, I need your advice about how to identify the storage planes used as dropzones from other storage planes.
If we can make the TpDropZoneQuery works for cloud environment, please let me know if you think we do not need this PR any more so that I can close it.

jakesmith: >To do that, I need your advice about how to identify the storage planes used as dropzones from other storage planes.

@shamser is working on a PR, that will tag storage planes as LZ's.
There will then be a helper function that will return those LZ's I think.

shamser: @jakesmith @wangkx I have added an attribute under plane called isDropZone.  By the way, I have made changes to WsTopology::getTpDropZones.  I'll expect to submit a PR today with the changes.

wangkx: Thanks shamser.

shamser: @wangkx Here's my changes relating to dfuserver & dropzones: https://github.com/hpcc-systems/HPCC-Platform/pull/14605

jakesmith:
> In the case of storage planes in general.. I'm not sure there is reason at the moment.
There might come to be, but at the moment I'd leave it until it was clear.

@ghalliday - I'm not sure there's a compelling reason to expose all (or any non-dropzone) planes at the moment.
But one question comes to mind - how do users (via IDE/Eclwatch) see a list of possible targets to write to via PLANE in ecl?
We could have a service that returns only some info. re. the 'data' planes..
(that would also be used to filter by in a logical file view in eclwatch).

wangkx:
I am closing it for now and reopen it when needed.

PR #14579, opened 2021-02-12
HPCC-25351 Add Add WsTopology.TpServiceTargetQuery

wangkx:
The ESP method is added to report the information about services
in a cloud environment.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/scm/ws_topology.ecm
+++ b/esp/scm/ws_topology.ecm
@@ -59,7 +59,15 @@ ESPStruct TpCluster
     [min_ver("1.15")] bool   HasThorSpareProcess;
     [min_ver("1.12")] ESParray<ESPstruct TpMachine> TpMachines;
 };
+
 //  ===========================================================================
+ESPStruct TpServiceTarget
+{
+    string Name;
+    string Type;
+    unsigned Port;
+    bool TLSSecure;
+};
 
 //  ===========================================================================
 ESPStruct TpLogicalCluster
@@ -646,7 +654,18 @@ ESPresponse [nil_remove, exceptions_inline] TpDropZoneQueryResponse
     ESParray<ESPstruct TpDropZone>    TpDropZones;
 };
 
-ESPservice [auth_feature("DEFERRED"), noforms, version("1.31"), cache_group("ESPWsTP"), exceptions_inline("./smc_xslt/exceptions.xslt")] WsTopology
+ESPrequest TpServiceTargetQueryRequest
+{
+    string Type;
+    string Name;
+};
+
+ESPresponse [nil_remove, exceptions_inline] TpServiceTargetQueryResponse
+{
+    ESParray<ESPstruct TpServiceTarget> TpServiceTargets;
+};
+
+ESPservice [auth_feature("DEFERRED"), noforms, version("1.32"), cache_group("ESPWsTP"), exceptions_inline("./smc_xslt/exceptions.xslt")] WsTopology
 {
     ESPmethod [cache_seconds(180), cache_global(1), resp_xsl_default("/esp/xslt/targetclusters.xslt")] TpTargetClusterQuery(TpTargetClusterQueryRequest, TpTargetClusterQueryResponse);
     ESPmethod [cache_seconds(180), cache_global(1), resp_xsl_default("/esp/xslt/topology.xslt")] TpClusterQuery(TpClusterQueryRequest, TpClusterQueryResponse);
@@ -669,6 +688,7 @@ ESPservice [auth_feature("DEFERRED"), noforms, version("1.31"), cache_group("ESP
     ESPmethod [cache_seconds(180), cache_global(1), min_ver(1.25)] TpMachineInfo(TpMachineInfoRequest, TpMachineInfoResponse);
 
     ESPmethod SystemLog(SystemLogRequest, SystemLogResponse);
+    ESPmethod [min_ver(1.32)] TpServiceTargetQuery(TpServiceTargetQueryRequest, TpServiceTargetQueryResponse);
 };
 
 SCMexportdef(WSWU);
--- a/esp/smc/SMCLib/TpWrapper.cpp
+++ b/esp/smc/SMCLib/TpWrapper.cpp
@@ -1970,6 +1970,36 @@ void CTpWrapper::getAttPath(const char* Path,StringBuffer& returnStr)
     JBASE64_Decode(Path, returnStr);
 }
 
+void CTpWrapper::getTpServiceTargets(double version, const char* serviceType, const char* serviceName, IArrayOf<IConstTpServiceTarget>& serviceTargets)
+{
+    Owned<IPropertyTreeIterator> services = queryComponentConfig().getElements("services");
+    ForEach(*services)
+    {
+        IPropertyTree& service = services->query();
+        const char* type = service.queryProp("@type");
+        if (isEmptyString(type) || (!isEmptyString(serviceType) && !strieq(serviceType, type)))
+            continue;
+
+        const char* name = service.queryProp("@name");
+        if (isEmptyString(name) || (!isEmptyString(serviceName) && !strieq(serviceName, name)))
+            continue;
+
+        //Only show the public services for now
+        if (!service.getPropBool("@public"))
+            continue;
+
+        Owned<IEspTpServiceTarget> serviceTarget = createTpServiceTarget();
+        serviceTarget->setName(name);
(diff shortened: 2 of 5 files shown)


wangkx:
@GordonSmith ECLWatch UI may use the changes in this PR.

esp/scm/ws_topology.ecm
```diff
@@ -59,7 +59,17 @@ ESPStruct TpCluster
...
+
 //  ===========================================================================
+ESPStruct TpServiceTarget
+{
+    string Name;
+    string Type;
+    unsigned Port;
+    bool TLSSecure;
+    bool Public;
+    string Target;
```
jakesmith, line 71: what is target? not in any of the service type schema's/or in values.yaml afaics.

esp/scm/ws_topology.ecm
```diff
@@ -59,7 +59,17 @@ ESPStruct TpCluster
...
+};
```
jakesmith, line 72: How is this info. going to be used?
In particular - if not public, then it is an internal service (used only internally).

I think need to 1st be sure what we want and how this/these services are going to be used.
If it's primarily to display on an eclwatch page for example, apart from *perhaps* ops. looking at the system, I don't think it's useful to display internal services.


Not much detail in the JIRA as to how this info. would be used. Can you expand ?

Out of interest, how does bare-metal expose the esp services running on other ports ?

esp/scm/ws_topology.ecm
wangkx: I saw the target from the roxie service:
services:
      - name: roxie
        type: roxie
        port: 9876
        target: roxie

esp/scm/ws_topology.ecm
wangkx: A user may use the port: 8010, 8002, etc. Should a user need to know the TLS?
If we only want to expose public services, I can add the filter. Should we let Admin users know internal services?
In bare-metal mode, we display similar information on Operation/System Servers page.

jakesmith: I suspect we don't want to show internal services by default, not sure in general.
If it's useful, can be acted upon, then it makes sense I think, if not, it either needs to be ignored, or shown separately somehow.
I'd just show the public services for now.

esp/scm/ws_topology.ecm
jakesmith: right ok, I see. The target is the roxie, the service name of this roxie is also named roxie.

wangkx:
@jakesmith I have changed the code to only return public services. I also removed the target attribute since it only exists in roxie service and the roxie service is not a public service.

jakesmith:
>I have changed the code to only return public services. I also removed the target attribute since it only exists in roxie service and the roxie service is not a public service.

Actually, the roxie service is public.. It's the public (client) gateway to the roxie that's deployed.
NB: see kubectl get svc 

But that's effectively a bug in how services are generated, i.e. should generate public for this service too.
Please open a JIRA for that - and link it to this one.

esp/smc/SMCLib/TpWrapper.cpp
```diff
@@ -1970,6 +1970,36 @@ void CTpWrapper::getAttPath(const char* Path,StringBuffer& returnStr)
...
+
+        //Only show the public services for now
+        if (!service.getPropBool("@public"))
+            continue;
+
+        Owned<IEspTpServiceTarget> serviceTarget = createTpServiceTarget();
+        serviceTarget->setName(name);
+        serviceTarget->setType(type);
+        serviceTarget->setPort(service.getPropInt("@port"));
+        if (service.hasProp("@tls")) //if not set, default to ?
```
jakesmith, line 1995: I would remove the comment, I think it's fine to treat absence of this property as false.
So would prob. change to:
```
   if (service.getPropBool("@tls"))
```

jakesmith (commented):
@wangkx - looks fine, 1 minor comment

esp/smc/SMCLib/TpWrapper.cpp
wangkx: Will change

wangkx:
> But that's effectively a bug in how services are generated, i.e. should generate public for this service too. Please open a JIRA for that - and link it to this one.

https://track.hpccsystems.com/browse/HPCC-25382

jakesmith (commented):

wangkx:
@jakesmith I did rebase and tests.

jakesmith:
@wangkx - looks good, thought of one more question though.

is getServiceTargets correct naming?

These are services, I think we're calling the main component queues "Targets".
So it may be better for these methods to be just called "getServices" without conflating with Targets.

Not sure. @ghalliday - what are your thoughts?

wangkx:
The existing WsTopology service already has a method named TpServiceQuery which reports the hpcc components (dali, esps, dfuservers, etc) in bare metal env. Should I merge the code there?

jakesmith:
> The existing WsTopology service already has a method named TpServiceQuery which reports the hpcc components (dali, esps, dfuservers, etc) in bare metal env. Should I merge the code there?

What is that used for/by ? It may not be very applicable to cloud, especially if esp services are now being returned from this new service.

wangkx:
> What is that used for/by ? It may not be very applicable to cloud, especially if esp services are now being returned from this new service.
For example, http://10.173.160.101:8010/#/stub/OPS-DL/SystemServersQuery-DL/SystemServersQueryWidgetIframeWidget
Mainly used by Operation team. The page shows Network Address, working folders, links to log files and configuration file. Gordon may hide it for clouds.

jakesmith:
> For example, http://10.173.160.101:8010/#/stub/OPS-DL/SystemServersQuery-DL/SystemServersQueryWidgetIframeWidget
> Mainly used by Operation team. The page shows Network Address, working folders, links to log files and configuration file. Gordon may hide it for clouds.

ok, then the 'services' in a cloud setup (i.e. those returned by this service and generated in the component config) are different.
And yes TpServiceQuery sounds like it should be suppressed in cloud.
I'm not sure it's just eclwatch though - what sense will this info. make in cloud world, if called by anything else?

If this (and other) services replace TpServiceQuery (and others), then they can be excluded from each build.

I suspect there's quite a few other things in WsTopology that don't quite make sense in cloud anymore, e.g. TpMachine/tpMachineType, or things with paths to logs, etc.
At some point it might make more sense to have a WsContainerServices for services calls like this one, leaving the bare-metal ones as they are.

wangkx:
> At some point it might make more sense to have a WsContainerServices for services calls like this one, leaving the bare-metal ones as they are.

Considering a possible future WsContainerServices (or a better name) service, it may be better not merge the code into the existing WsTopology.TpServiceQuery. We may give the TpServiceTargetQuery method a better name (TpContainerServiceQuery?) and, later on, move the whole method, and other methods (like the DropZone related methods) into the WsContainerServices service.

jakesmith:
> > At some point it might make more sense to have a WsContainerServices for services calls like this one, leaving the bare-metal ones as they are.
> 
> Considering a possible future WsContainerServices (or a better name) service, it may be better not merge the code into the existing WsTopology.TpServiceQuery. We may give the TpServiceTargetQuery method a better name (TpContainerServiceQuery?) and, later on, move the whole method, and other methods (like the DropZone related methods) into the WsContainerServices service.

It depends how much the existing WsTopology services and new container services overlap/diverge.
If it a lot of of WsTopology still makes sense, then it may make sense to be conjoined, trying to minimize the differences.
If they're quite different in many areas, and the containerized services are fewer, smaller and cleaner in a separate service, I suspect it is better to separate.

jakesmith:
@GordonSmith - can you take look at this from a functionality point of view.
This service calls currently returns a list of services that are public, i.e. open to the outside (there are other internal services that it does not show). NB: these are the 'services' generated into the global config.

What will eclwatch + other client want to see etc. ?
Is the info. per service adequate?

I think there's another JIRA re. a separate service call that will return the job queues that clients can submit to.

jakesmith (commented):
@wangkx - should this be closed / does https://github.com/hpcc-systems/HPCC-Platform/pull/14636 replace it?

wangkx:
Looks like this PR should be replaced by https://github.com/hpcc-systems/HPCC-Platform/pull/14636.

PR #14583, opened 2021-02-15
HPCC-25360 Convert sasha service list to a dictionary

jakesmith:
Also fix an issue if secrets configured, and multiple dali
sasha services used them and generated the same volume>1

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/helm/hpcc/templates/_helpers.tpl
+++ b/helm/hpcc/templates/_helpers.tpl
@@ -609,17 +609,19 @@ Generate list of available services
   public: {{ $esp.public }}
 {{ end -}}
 {{- range $dali := $.Values.dali -}}
-{{- range $sasha := $dali.services -}}
+{{- range $sashaName, $_sasha := $dali.services -}}
+{{- $sasha := ($_sasha | default dict) -}}
 {{- if and (not $sasha.disabled) ($sasha.servicePort) -}}
-- name: {{ printf "sasha-%s" $sasha.service }}
+- name: {{ printf "sasha-%s" $sashaName }}
   type: sasha
   port: {{ $sasha.servicePort }}
 {{ end -}}
 {{ end -}}
 {{ end -}}
-{{- range $sasha := $.Values.sasha -}}
+{{- range $sashaName, $_sasha := $.Values.sasha -}}
+{{- $sasha := ($_sasha | default dict) -}}
 {{- if and (not $sasha.disabled) ($sasha.servicePort) -}}
-- name: {{ printf "sasha-%s" $sasha.service }}
+- name: {{ printf "sasha-%s" $sashaName }}
   type: sasha
   port: {{ $sasha.servicePort }}
 {{ end -}}
@@ -664,12 +666,12 @@ Sasha configmap
 Pass in dict with root and me
 */}}
 {{- define "hpcc.sashaConfigMap" -}}
-{{- $serviceName := printf "sasha-%s" .me.service -}}
+{{- $configMapName := printf "sasha-%s" .me.name -}}
 apiVersion: v1
 metadata:
-  name: {{ printf "%s-configmap" $serviceName }}
+  name: {{ printf "%s-configmap" $configMapName }}
 data:
-  {{ $serviceName }}.yaml: |
+  {{ $configMapName }}.yaml: |
     version: 1.0
     sasha:
 {{ toYaml (omit .me "logging") | indent 6 }}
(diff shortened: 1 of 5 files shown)


helm/hpcc/templates/_helpers.tpl
```diff
@@ -609,17 +609,19 @@ Generate list of available services
   public: {{ $esp.public }}
 {{ end -}}
 {{- range $dali := $.Values.dali -}}
-{{- range $sasha := $dali.services -}}
+{{- range $sashaName, $_sasha := $dali.services -}}
+{{- $sasha := ($_sasha | default dict) -}}
```
shamser, line 613: What's the purpose of 'default dict' here?

helm/hpcc/templates/_helpers.tpl
```diff
@@ -609,17 +609,19 @@ Generate list of available services
...
-- name: {{ printf "sasha-%s" $sasha.service }}
+- name: {{ printf "sasha-%s" $sashaName }}
   type: sasha
   port: {{ $sasha.servicePort }}
 {{ end -}}
 {{ end -}}
 {{ end -}}
-{{- range $sasha := $.Values.sasha -}}
+{{- range $sashaName, $_sasha := $.Values.sasha -}}
+{{- $sasha := ($_sasha | default dict) -}}
```
shamser, line 622: Again, is 'default dict' needed here?

helm/hpcc/values.schema.json
```diff
@@ -799,70 +798,81 @@
...
+              "description": "Coalescing will only begin, if the delta size is above this threashold (K)"
+            },
+            "disabled": {},
+            "interval": {},
+            "servicePort": {},
+            "port": {},
+            "at" : {},
+            "throttle": {},
+            "storage": {},
+            "resources": {}
```
shamser, line 817: The attributes above do not have a type or description

helm/hpcc/values.schema.json
```diff
@@ -799,70 +798,81 @@
...
-        "servicePort": {},
-        "port": {},
-        "at" : {},
-        "throttle": {},
-        "storage": {},
-        "resources": {}
-      },
-      "additionalProperties": false
+        {
+          "type": "null"
```
shamser, line 822: what's the purpose of this?

helm/hpcc/values.schema.json
```diff
@@ -887,63 +897,87 @@
...
+            "disabled": {},
+            "interval": {},
+            "servicePort": {},
+            "port": {},
+            "at" : {},
+            "throttle": {},
+            "storage": {},
+            "resources": {},
+            "limit": {},
+            "cutoff": {}
```
shamser, line 915: The attributes above do not have a type or description

helm/hpcc/values.yaml
```diff
@@ -234,70 +234,70 @@ bundles: []
...
-    #  cpu: "1"
-    #  memory: "1G"
-
-  # metaAccess: "shared" # enable if coalescer running in separate pod, and thus needs access to meta over PVC mount
+    coalescer:
+      #servicePort: 8877
+      #interval: 2 # (hours)
+      #at: "* * * * *" # cron type schedule, i.e. Min(0-59) Hour(0-23) DayOfMonth(1-31) Month(1-12) DayOfWeek(0-6)
+      #minDeltaSize: 50 # (Kb) will not start coalescing until delta log is above this threshold
+      #recoverFromIncErrors: false # Try to continue after transaction load failures
```
shamser, line 242: recoverFromIncErrors is not defined in the schema.

helm/hpcc/values.schema.json
shamser: A lot of fields here.  Could we simplify values.yaml by defining defaults for most of these?

shamser (commented):
@jakesmith

helm/hpcc/values.yaml
jakesmith: I didn't mean for this to be here, will remove.

helm/hpcc/values.schema.json
jakesmith: actually, they are still needed.
The schemas are defined and inherited via allOf and "sashacommon".
They are necessary to define here too, because the additionalProperties: false only applies to "properties", so although the schema are inherited, they need defining here to avoid additionalProperties rejecting them when they appear in values.yaml.
For reference see: https://json-schema.org/understanding-json-schema/reference/combining.html.

However, I did notice a mistake in the use of allOf, I'll change in coming commit

jakesmith:
@shamser - please review new commit

shamser (approved):
@richardkchapman Reviewed.

PR #14605, opened 2021-02-23
HPCC-25394 dfuserver despray support in helm v1

shamser:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/common/remote/rmtspawn.cpp
+++ b/common/remote/rmtspawn.cpp
@@ -140,7 +140,12 @@ ISocket * spawnRemoteChild(SpawnKind kind, const char * exe, const SocketEndpoin
         LOG(MCdetailDebugInfo, unknownJob, "Action aborted before connecting to slave (%3d)", replyTag);
         return NULL;
     }
-
+#ifdef _CONTAINERIZED
+    // TODO: replace with K8s job
+    DWORD runcode;
+    if (!invoke_program(cmd.str(), runcode, false))
+        throw makeStringExceptionV(-1,"Error spawning %s", exe);
+#else
     if (SSHusername.isEmpty())
     {
 #if defined(_WIN32)
@@ -162,6 +167,7 @@ ISocket * spawnRemoteChild(SpawnKind kind, const char * exe, const SocketEndpoin
         runssh->init(cmd.str(),SSHidentfilename,SSHusername,SSHpasswordenc,SSHtimeout,SSHretries);
         runssh->exec(childEP,NULL,true); // need workdir? TBD
     }
+#endif
 
     //Have to now try and connect to the child and get back the port it is listening on
     unsigned attempts = 20;
--- a/dali/dfu/dfurun.cpp
+++ b/dali/dfu/dfurun.cpp
@@ -48,6 +48,7 @@ test multiclusteradd with replicate
 #include "dfurun.hpp"
 #include "eventqueue.hpp"
 #include "wujobq.hpp"
+#include "dameta.hpp"
 
 #define SDS_CONNECT_TIMEOUT (5*60*100)
 
@@ -458,10 +459,23 @@ class CDFUengine: public CInterface, implements IDFUengine
         if ((isDotDotString != nullptr) || (isDotString != nullptr))
             throwError3(DFTERR_InvalidFilePath, pfilePath, dotDotString, dotString);
 
-        Owned<IEnvironmentFactory> factory = getEnvironmentFactory(true);
-        Owned<IConstEnvironment> env = factory->openEnvironment();
         StringBuffer netaddress;
         filename.queryIP().getIpText(netaddress);
+#ifdef _CONTAINERIZED
+        Owned<IPropertyTreeIterator> planes = getDropZonePlanesIterator();
+        ForEach(*planes)
+        {
+            IPropertyTree & plane = planes->query();
+            const char * fullDropZoneDir = plane.queryProp("@prefix");
+            assertex(fullDropZoneDir);
+            // note: for bare-metal drop-zones, will need to compare ip address
+            if (startsWith(pfilePath, fullDropZoneDir))
+                return;
+        }
+        throwError1(DFTERR_NoMatchingDropzonePlane, pfilePath);
+#else
+        Owned<IEnvironmentFactory> factory = getEnvironmentFactory(true);
+        Owned<IConstEnvironment> env = factory->openEnvironment();
 
         Owned<IConstDropZoneInfo> dropZone = env->getDropZoneByAddressPath(netaddress.str(), pfilePath);
         if (!dropZone)
@@ -482,6 +496,7 @@ class CDFUengine: public CInterface, implements IDFUengine
                 , (dropZone->isECLWatchVisible() ? "" : "not ")
                 );
         }
+#endif
 #endif
     }
 
--- a/dali/dfu/dfuserver.cpp
+++ b/dali/dfu/dfuserver.cpp
@@ -44,6 +44,7 @@
 #include "jprop.hpp"
 
 #include "dfuerror.hpp"
+#include "daqueue.hpp"
 
 static Owned<IPropertyTree> globals;
 ILogMsgHandler * fileMsgHandler;
@@ -56,8 +57,8 @@ Owned<IDFUengine> engine;
 void usage()
 {
     printf("Usage:\n");
-    printf("  DFUSERVER DALISERVERS=<ip> QUEUE=<q-name>          -- starts DFU Server\n\n");
-    printf("  DFUSERVER DALISERVERS=<ip> STOP=1 QUEUE=<q-name>   -- stops DFU Server\n\n");
+    printf("  dfuserver --daliServers=<ip> --queue=<q-name>          -- starts DFU Server\n\n");
+    printf("  dfuserver --daliServers=<ip> --queue=<q-name> --stop=1 -- stops DFU Server\n\n");
 }
 
 
@@ -68,6 +69,7 @@ static bool exitDFUserver()
     return false;
 }
 
+#ifndef _CONTAINERIZED
 inline void XF(IPropertyTree &pt,const char *p,IProperties &from,const char *fp)
 {
     const char * v = from.queryProp(fp);
@@ -95,7 +97,15 @@ IPropertyTree *readOldIni()
     }
     return ret;
 }
+#endif
 
+static constexpr const char * defaultYaml = R"!!(
+version: "1.0"
+dfuserver:
+  name: dfuserver
+  monitorInterval: 900
+  maxJobs: 1
+)!!";
 
 int main(int argc, const char *argv[])
 {
@@ -113,48 +123,53 @@ int main(int argc, const char *argv[])
 
     NoQuickEditSection xxx;
 
-    Owned<IFile> file = createIFile("dfuserver.xml");
-    if (file->exists())
-        globals.setown(createPTreeFromXMLFile("dfuserver.xml", ipt_caseInsensitive));
-    else
-        globals.setown(readOldIni());
-
-    for (unsigned i=1;i<(unsigned)argc;i++) {
-        const char *arg = argv[i];
-        StringBuffer prop("@");
-        StringBuffer val;
-        while (*arg && *arg != '=')
-            prop.append(*arg++);
-        if (*arg) {
-            arg++;
-            while (isspace(*arg))
-                arg++;
-            val.append(arg);
-            prop.clip();
-            val.clip();
-            if (prop.length()>1)
-                globals->setProp(prop.str(), val.str());
-        }
+    try
+    {
+        globals.setown(loadConfiguration(defaultYaml, argv, "dfuserver", "DFUSERVER", "dfuserver.xml", nullptr));
     }
-    StringBuffer daliServer;
-    StringBuffer queue;
-    if (!globals->getProp("@DALISERVERS", daliServer)||!globals->getProp("@QUEUE", queue)) {
-        usage();
-        globals.clear();
-        releaseAtoms();
+    catch (IException * e)
+    {
+        OERRLOG(e);
+        e->Release();
+        return 1;
+    }
+    catch(...)
+    {
+        OERRLOG("Failed to load configuration");
+        return 1;
+    }
+
+    StringBuffer name, daliServers;
+    if (!globals->getProp("@daliServers", daliServers))
+    {
+        PROGLOG("daliServers not specified in configuration");
+        return 1;
+    }
+    if (!globals->getProp("@name", name))
+    {
+        PROGLOG("name not specified in configuration");
         return 1;
     }
     Owned<IFile> sentinelFile;
-    bool stop = globals->getPropInt("@STOP",0)!=0;
+#ifndef _CONTAINERIZED
+    bool stop = globals->getPropBool("@stop", false);
+#else
+    bool stop = false; // In containerized, this is never true: queues stopped when container stopped
+#endif
     if (!stop) {
         sentinelFile.setown(createSentinelTarget());
         removeSentinelFile(sentinelFile);
+    }
+#ifdef _CONTAINERIZED
+    setupContainerizedLogMsgHandler();
+#else
+    if (!stop)
+    {
         Owned<IComponentLogFileCreator> lf = createComponentLogFileCreator(globals, "dfuserver");
         lf->setMaxDetail(1000);
         fileMsgHandler = lf->beginLogging();
     }
     StringBuffer ftslogdir;
-    const char* name = globals->queryProp("@name");
     if (getConfigurationDirectory(globals->queryPropTree("Directories"),"log","ftslave",name,ftslogdir)) // NB instance deliberately dfuserver's
         setFtSlaveLogDir(ftslogdir.str());
     setRemoteSpawnSSH(
@@ -164,11 +179,11 @@ int main(int argc, const char *argv[])
         globals->getPropInt("SSH/@SSHtimeout",0),
         globals->getPropInt("SSH/@SSHretries",3),
         "run_");
-    bool enableSNMP = globals->getPropInt("@enableSNMP")!=0;
+#endif
     CSDSServerStatus *serverstatus=NULL;
     Owned<IReplicateServer> replserver;
     try {
-        Owned<IGroup> serverGroup = createIGroupRetry(daliServer.str(),DALI_SERVER_PORT);
+        Owned<IGroup> serverGroup = createIGroupRetry(daliServers,DALI_SERVER_PORT);
         initClientProcess(serverGroup, DCR_DfuServer, 0, NULL, NULL, stop?(1000*30):MP_WAIT_FOREVER);
 
         if(!stop)
@@ -188,6 +203,13 @@ int main(int argc, const char *argv[])
 
             IPropertyTree * config = nullptr;
             installDefaultFileHooks(config);
+        }
+        StringBuffer queue, monitorQueue;
+#ifndef _CONTAINERIZED
+        if (!globals->getProp("@queue", queue))
+        {
+            PROGLOG("queue not specified in configuration");
+            return 1;
         }
         const char *q = queue.str();
         for (;;) {
@@ -222,24 +244,45 @@ int main(int argc, const char *argv[])
             if (!*q)
                 break;
         }
-        q = globals->queryProp("@MONITORQUEUE");
-        if (q&&*q) {
+        globals->getProp("@monitorQueue", monitorQueue);
+#else
+        getDfuQueueName(queue, name);
+        unsigned maxJobs = globals->getPropInt("@maxJobs", 1);
+        if (maxJobs<1)
+        {
+            OERRLOG("maxJobs is %u", maxJobs);
+            return 1;
+        }
+        IPropertyTree *t = createPTree();
+        t->setProp("@name",queue);
+        t->setPropInt("@num", maxJobs);
+        serverstatus->queryProperties()->addPropTree("Queue",t);
+        serverstatus->commitProperties();
+        engine->setDefaultTransferBufferSize((size32_t)globals->getPropInt("@transferBufferSize"));
+        for (unsigned i=0; i<maxJobs; i++)
+            engine->startListener(queue,serverstatus);
+        getDfuMonitorQueueName(monitorQueue, name);
+#endif
+        if (monitorQueue.length()>0)
+        {
             if (stop) {
-                stopDFUserver(q);
+                stopDFUserver(monitorQueue);
             }
             else {
-                IPropertyTree *t=serverstatus->queryProperties()->addPropTree("MonitorQueue",createPTree());
-                t->setProp("@name",q);
-                engine->startMonitor(q,serverstatus,globals->getPropInt("@MONITORINTERVAL",60)*1000);
+                IPropertyTree *t=serverstatus->queryProperties()->addPropTree("monitorQueue",createPTree());
+                t->setProp("@name",monitorQueue);
+                int monitorInterval = globals->getPropInt("@monitorInterval", 60);
+                engine->startMonitor(monitorQueue,serverstatus,monitorInterval*1000);
             }
         }
-        q = globals->queryProp("@REPLICATEQUEUE");
-        if (q&&*q) {
+        const char *replicateQueue = globals->queryProp("@replicateQueue");
+        if (!isEmptyString(replicateQueue))
+        {
             if (stop) {
                 // TBD?
             }
             else {
-                replserver.setown(createReplicateServer(q));
+                replserver.setown(createReplicateServer(replicateQueue));
                 replserver->runServer();
             }
         }
--- a/esp/scm/ws_fs.ecm
+++ b/esp/scm/ws_fs.ecm
@@ -414,6 +414,7 @@ ReplicateResponse
 
 ESPrequest Despray
 {
+    string destGroup;
     string sourceLogicalName;
 
     string destIP;
--- a/esp/services/ws_fs/ws_fsService.cpp
+++ b/esp/services/ws_fs/ws_fsService.cpp
@@ -40,6 +40,8 @@
 #include "sacmd.hpp"
 #include "exception_util.hpp"
 #include "LogicFileWrapper.hpp"
+#include "dameta.hpp"
+#include "daqueue.hpp"
 
 #define DFU_WU_URL          "DfuWorkunitsAccess"
 #define DFU_EX_URL          "DfuExceptionsAccess"
@@ -116,8 +118,8 @@ int Schedule::run()
 
 void CFileSprayEx::init(IPropertyTree *cfg, const char *process, const char *service)
 {
+#ifndef _CONTAINERIZED
     StringBuffer xpath;
-
     xpath.clear().appendf("Software/EspProcess[@name=\"%s\"]/EspService[@name=\"%s\"]/QueueLabel", process, service);
     cfg->getProp(xpath.str(), m_QueueLabel);
     StringArray qlist;
@@ -154,13 +156,24 @@ void CFileSprayEx::init(IPropertyTree *cfg, const char *process, const char *ser
     cfg->getProp(xpath.str(), m_MonitorQueueLabel);
 
     directories.set(cfg->queryPropTree("Software/Directories"));
-
-    StringBuffer prop;
-    prop.appendf("queueLabel=%s", m_QueueLabel.str());
-    DBGLOG("%s", prop.str());
-    prop.clear();
-    prop.appendf("monitorQueueLabel=%s", m_MonitorQueueLabel.str());
-    DBGLOG("%s", prop.str());
+#else
+    // Using the first queue for now.
+    // TODO: Re-design support for multiple queues
+    Owned<IPropertyTreeIterator> dfuQueues = queryComponentConfig().getElements("dfuQueues");
+    ForEach(*dfuQueues)
+    {
+        IPropertyTree & dfuQueue = dfuQueues->query();
+        const char * dfuName = dfuQueue.queryProp("@name");
+        if (!isEmptyString(dfuName))
+        {
+            getDfuQueueName(m_QueueLabel, dfuName);
+            getDfuMonitorQueueName(m_MonitorQueueLabel, dfuName);
+            break;
+        }
+    }
+#endif
+    DBGLOG("queueLabel=%s", m_QueueLabel.str());
+    DBGLOG("monitorQueueLabel=%s", m_MonitorQueueLabel.str());
 
     if (!daliClientActive())
     {
@@ -222,13 +235,15 @@ static void DeepAssign(IEspContext &context, IConstDFUWorkUnit *src, IEspDFUWork
     if(src == NULL)
         throw MakeStringException(ECLWATCH_MISSING_PARAMS, "'Source DFU workunit' doesn't exist.");
 
+    StringBuffer tmp;
+    double version = context.getClientVersion();
+    dest.setID(src->queryId());
+
+#ifndef _CONTAINERIZED
     Owned<IEnvironmentFactory> envFactory = getEnvironmentFactory(true);
     Owned<IConstEnvironment> constEnv = envFactory->openEnvironment();
     Owned<IPropertyTree> root = &constEnv->getPTree();
 
-    double version = context.getClientVersion();
-    StringBuffer tmp, encoded;
-    dest.setID(src->queryId());
     if (src->getClusterName(tmp.clear()).length()!=0)
     {
         char *clusterName = (char *)tmp.str();
@@ -282,7 +297,10 @@ static void DeepAssign(IEspContext &context, IConstDFUWorkUnit *src, IEspDFUWork
             dest.setClusterName(clusterNameForDisplay.str());
         }
     }
-
+#else
+    src->getClusterName(tmp.clear());
+    dest.setClusterName(tmp.str());
+#endif
     if ((version > 1.05) && src->getDFUServerName(tmp.clear()).length())
         dest.setDFUServerName(tmp.str());
 
@@ -349,7 +367,9 @@ static void DeepAssign(IEspContext &context, IConstDFUWorkUnit *src, IEspDFUWork
         StringBuffer lfn;
         file->getLogicalName(lfn);
         if (lfn.length() != 0)
+        {
             dest.setSourceLogicalName(lfn.str());
+        }
         else
             dest.setSourceFormat(file->getFormat());
 
@@ -2214,8 +2234,30 @@ bool CFileSprayEx::onReplicate(IEspContext &context, IEspReplicate &req, IEspRep
     return true;
 }
 
+void CFileSprayEx::getDropZoneInfoByDestPlane(double clientVersion, const char* destPlane, const char* destFileIn, StringBuffer& destFileOut, StringBuffer& umask)
+{
+    Owned<IPropertyTree> plane = getDropZonePlane(destPlane);
+    if (!plane)
+        throw makeStringExceptionV(ECLWATCH_DROP_ZONE_NOT_FOUND, "No drop zone matching plane %s", destPlane);
+
+    StringBuffer fullDropZoneDir(plane->queryProp("@prefix"));
+    addPathSepChar(fullDropZoneDir);
+    if (isAbsolutePath(destFileIn))
+    {
+        if (strncmp(fullDropZoneDir, destFileIn, fullDropZoneDir.length()))
+            throw makeStringExceptionV(ECLWATCH_DROP_ZONE_NOT_FOUND, "No drop zone configured for %s:%s", destPlane, destFileIn);
+        destFileOut.set(destFileIn);
+    }
+    else
+    {
+        destFileOut.append(destFileIn);
+    }
+    plane->getProp("@umask", umask);
+}
+
 void CFileSprayEx::getDropZoneInfoByIP(double clientVersion, const char* ip, const char* destFileIn, StringBuffer& destFileOut, StringBuffer& umask)
 {
+#ifndef _CONTAINERIZED
     if (destFileIn && *destFileIn)
         destFileOut.set(destFileIn);
 
@@ -2302,6 +2344,9 @@ void CFileSprayEx::getDropZoneInfoByIP(double clientVersion, const char* ip, con
         else
             LOG(MCdebugInfo, unknownJob, "No valid drop zone found for network address '%s'. Check your system drop zone configuration.", ip);
     }
+#else
+    throw makeStringException(-1, "Internal error: CFileSprayEx::getDropZoneInfoByIP should not be called in containerized environment");
+#endif
 }
 
 static StringBuffer & expandLogicalAsPhysical(StringBuffer & target, const char * name, const char * separator)
@@ -2379,7 +2424,12 @@ bool CFileSprayEx::onDespray(IEspContext &context, IEspDespray &req, IEspDespray
                 throw MakeStringException(ECLWATCH_INVALID_INPUT, "Despray %s: cannot resolve destination network IP from %s.", srcname, destip);
 
             StringBuffer destfileWithPath, umask;
+#ifdef _CONTAINERIZED
+            const char* destPlane = req.getDestGroup();
+            getDropZoneInfoByDestPlane(version, destPlane, destfile, destfileWithPath, umask);
+#else
             getDropZoneInfoByIP(version, destip, destfile, destfileWithPath, umask);
+#endif
             //Ensure the filename is dependent on the file part if parts are being preserved
             if (preserveFileParts && !strstr(destfileWithPath, "$P$"))
                 destfileWithPath.append("._$P$_of_$N$");
--- a/esp/smc/SMCLib/TpWrapper.cpp
+++ b/esp/smc/SMCLib/TpWrapper.cpp
@@ -26,6 +26,7 @@
 #include "exception_util.hpp"
 #include "portlist.h"
 #include "daqueue.hpp"
+#include "dautils.hpp"
 
 const char* MSG_FAILED_GET_ENVIRONMENT_INFO = "Failed to get environment information.";
 
@@ -1664,7 +1665,6 @@ void CTpWrapper::getDropZoneMachineList(double clientVersion, bool ECLWatchVisib
     {
         IWARNLOG("Unknown Exception caught within CTpWrapper::getDropZoneMachineList");
     }
-    
 }
 
 //For a given dropzone or every dropzones (check ECLWatchVisible if needed), read: "@name",
@@ -1674,6 +1674,36 @@ void CTpWrapper::getDropZoneMachineList(double clientVersion, bool ECLWatchVisib
 
 void CTpWrapper::getTpDropZones(double clientVersion, const char* name, bool ECLWatchVisibleOnly, IArrayOf<IConstTpDropZone>& list)
 {
+
+#ifdef _CONTAINERIZED
+    Owned<IPropertyTreeIterator> planes = getDropZonePlanesIterator(name);
+    ForEach(*planes)
+    {
+        IPropertyTree & plane = planes->query();
+        const char * dropzonename = plane.queryProp("@name");
+        const char * prefix = plane.queryProp("@prefix");
+        StringBuffer path(prefix);
+        Owned<IEspTpDropZone> dropZone = createTpDropZone();
+        dropZone->setName(dropzonename);
+        dropZone->setDescription(""); //TODO
+        dropZone->setPath(path);
+        dropZone->setBuild("");
+        dropZone->setECLWatchVisible(plane.getPropBool("@eclWatchVisible", true));
+        StringBuffer localHost;
+        IArrayOf<IEspTpMachine> tpMachines;
+        Owned<IEspTpMachine> machine = createTpMachine();
+        IpAddress ipAddr;
+        ipAddr.ipset("localhost");
+        ipAddr.getIpText(localHost);
+        machine->setNetaddress(localHost.str());
+        machine->setConfigNetaddress("localhost");
+        machine->setDirectory(path);
+        machine->setOS(getPathSepChar(path) == '/' ? MachineOsLinux : MachineOsW2K);
+        tpMachines.append(*machine.getLink());
+        dropZone->setTpMachines(tpMachines);
+        list.append(*dropZone.getLink());
+    }
+#else
     Owned<IEnvironmentFactory> envFactory = getEnvironmentFactory(true);
     Owned<IConstEnvironment> constEnv = envFactory->openEnvironment();
     if (!isEmptyString(name))
@@ -1692,6 +1722,7 @@ void CTpWrapper::getTpDropZones(double clientVersion, const char* name, bool ECL
                 appendTpDropZone(clientVersion, constEnv, dropZoneInfo, list);
         }
     }
+#endif
 }
 
 
--- a/helm/hpcc/templates/_helpers.tpl
+++ b/helm/hpcc/templates/_helpers.tpl
@@ -79,7 +79,7 @@ storage:
   planes:
 {{- /*Generate entries for each data plane (removing the pvc).  Exclude the planes used for dlls and dali.*/ -}}
 {{- range $plane := $planes -}}
- {{- if or (not $plane.labels) (has "data" $plane.labels) }}
+ {{- if or (not $plane.labels) (or (has "data" $plane.labels) (has "lz" $plane.labels)) }}
   - name: {{ $plane.name | quote }}
 {{ toYaml (unset (unset (deepCopy $plane) "name") "pvc")| indent 4 }}
  {{- end }}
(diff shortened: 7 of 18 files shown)


jakesmith:
@shamser - looks like there's a problem starting dfuserver in bare-metal ?

esp/smc/SMCLib/TpWrapper.cpp
```diff
@@ -1674,6 +1674,45 @@ void CTpWrapper::getDropZoneMachineList(double clientVersion, bool ECLWatchVisib
...
 {
+
+#ifdef _CONTAINERIZED
+    IPropertyTree * storage = queryGlobalConfig().queryPropTree("storage");
+    if (storage)
+    {
+        StringBuffer pname("planes[@isDropZone='true']");
+        if (name && *name)
+            pname.appendf("[@name='%s']", name);
+        Owned<IPropertyTreeIterator> planes = storage->getElements(pname.str());
```
jakesmith, line 1685: Feels like getDropZones(const char *optName) should be a utility function for general use, outside of esp code, that esp code calls. And needed by dfurun for example.

Utility method could return a list of matching plane IPT's.

esp/smc/SMCLib/TpWrapper.cpp
```diff
@@ -1674,6 +1674,45 @@ void CTpWrapper::getDropZoneMachineList(double clientVersion, bool ECLWatchVisib
...
+            Owned<IEspTpMachine> machine = createTpMachine();
+            IpAddress ipAddr;
+            ipAddr.ipset("localhost");
+            ipAddr.getIpText(networkAddress);
+            machine->setNetaddress(networkAddress.str());
+            machine->setConfigNetaddress("localhost");
+            machine->setDirectory(path);
+            machine->setOS(getPathSepChar(path) == '/' ? MachineOsLinux : MachineOsW2K);
+            tpMachines.append(*machine.getLink());
+            dropZone->setTpMachines(tpMachines);
```
jakesmith, line 1711: what would happen if this was left unset?
"localhost"/IP isn't right if dropzone plane is a URL, there is no machine in that context.

helm/hpcc/templates/dfuserver.yaml
```diff
@@ -0,0 +1,64 @@
...
+      serviceAccountName: "hpcc-default"
+      initContainers:
+        {{- include "hpcc.checkDataMount" $commonCtx | indent 6 }}
+      containers:
+      - name: {{ .name | quote }}
+        workingDir: /var/lib/HPCCSystems
+        command: [ dfuserver ]
+        args: [
+                {{ include "hpcc.configArg" . }},
+                {{ include "hpcc.daliArg" $ }}
```
jakesmith, line 47: related to comment re. usage should be changed to --daliServers.
These includes generate --<property> <value>

dali/dfu/dfuserver.cpp
```diff
@@ -56,8 +56,8 @@ Owned<IDFUengine> engine;
 void usage()
 {
     printf("Usage:\n");
-    printf("  DFUSERVER DALISERVERS=<ip> QUEUE=<q-name>          -- starts DFU Server\n\n");
-    printf("  DFUSERVER DALISERVERS=<ip> STOP=1 QUEUE=<q-name>   -- stops DFU Server\n\n");
+    printf("  dfuserver daliservers=<ip> queue=<q-name>          -- starts DFU Server\n\n");
+    printf("  dfuserver daliservers=<ip> stop=1 queue=<q-name>   -- stops DFU Server\n\n");
```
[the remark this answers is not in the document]
jakesmith: NB: the yaml generated will use this form via e.g. include hpcc.daliArg

helm/hpcc/values.schema.json
```diff
@@ -140,6 +140,44 @@
...
+            },
+            "enableSNMP": {
+              "type": "boolean",
+              "description": ""
+            },
+            "enableSysLog": {
+              "description": "",
+              "type": "boolean"
+            },
+            "queue": {
```
jakesmith, line 164: I don't think we want to define a queue name separately in container mode.
I think want it to be implicit, just as it is for hthor, roxie, thor etc.
i.e. the queue is based on the name of the dfuserver

helm/hpcc/values.schema.json
```diff
@@ -140,6 +140,44 @@
...
+          {
+            "name": {
+              "type": "string",
+              "description": "The name of the dfuserver process"
+            },
+            "disabled": {
+              "type": "boolean"
+            },
+            "enableSNMP": {
+              "type": "boolean",
```
jakesmith, line 157: I doubt this should be allowed in container, removed from schema if not.

helm/hpcc/values.schema.json
```diff
@@ -140,6 +140,44 @@
...
+              "description": ""
+            },
+            "enableSysLog": {
+              "description": "",
+              "type": "boolean"
```
jakesmith, line 162: not sure about this, is it in any other of our schema's?

helm/hpcc/values.schema.json
```diff
@@ -140,6 +140,44 @@
...
+            },
+            "queue": {
+              "description": "",
+              "type": "string"
+            },
+            "monitorqueue": {
+              "description": "",
+              "type": "string"
```
jakesmith, line 170: does anyone know if and how monitorqueue works?
A quick scan suggests it's all based on scanning dfu workunits who have state DFUstate_monitoring, but I don't see anything that sets that.

@ghalliday @AttilaVamos ?

helm/hpcc/values.schema.json
```diff
@@ -327,6 +365,14 @@
           "description": "which planes (if any) the data is replicated onto (primarily bare metal)",
           "type": "array",
           "items": { "type": "string" }
+        },
+        "isDropZone" : {
+          "description": "true if data plane is dropzone",
+          "type": "boolean"
+        },
+        "umask" : {
```
jakesmith, line 373: why is a umask this necessary/desirable?
Maybe it is, just trying to understand how this fits in with storage/security and non-posix planes.

dali/dfu/dfuserver.cpp
```diff
@@ -113,48 +127,56 @@ int main(int argc, const char *argv[])
...
-        releaseAtoms();
+    if (!globals->getProp("@daliServers", daliServer))
+        globals->getProp("@DALISERVERS", daliServer);
+    if (0==daliServer.length())
+    {
+        PROGLOG("daliservers not specified in dfuserver.xml");
+        return 1;
+    }
+    if (!globals->getProp("@QUEUE", queue))
+        globals->getProp("@queue", queue);
```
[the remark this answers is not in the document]
jakesmith: In container mode, I don't think we should have a configurable "queue" option.
It should be implicit based on name of the dfuserver instance, as it is for other engines/agents etc.
There are utility function in daqueue.hpp, a new one could be added for getDfuServerQueueName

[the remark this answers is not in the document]
jakesmith: Is there going to be a dfuserver instance per active job?
With replicas auto scaled, based on active/queue size? (maybe at some point).

I think until then, we need an option that controls max number of jobs it's going to run in parallel.
At the moment, there's some very odd code https://github.com/hpcc-systems/HPCC-Platform/blob/ff5e5f118e9d77de533725ce86914f4c66339ecf/dali/dfu/dfuserver.cpp#L192, that determines how many jobs it can run in parallel, by seeing how many comma separated queues are defined in @queues, then creating that many listeners.
NB: what our ops tend to do is set bare-metal @queue to e.g. "dfuserver_queue,dfuserver_queue,dfuserver_queue,dfuserver_queue,dfuserver_queue"
which in effect causes it to run up to 5 jobs at one time.

For now, I'd suggest adding a maxJobs config option (similar to other components), and creating that many listeners.

jakesmith (commented):
@shamser - some initial comments, questions and feedback

dali/dfu/dfuserver.cpp
```diff
@@ -113,48 +127,56 @@ int main(int argc, const char *argv[])
...
+    }
+    catch (IException * e)
+    {
+        UERRLOG(e);
+        e->Release();
+        return 1;
+    }
+    catch(...)
+    {
+        OERRLOG("Failed to load configuration");
```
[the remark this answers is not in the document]
shamser: I've changed IException to report as operator message.

dali/dfu/dfuserver.cpp
```diff
@@ -95,7 +96,20 @@ IPropertyTree *readOldIni()
...
     return ret;
 }
-
+#endif
+
+static constexpr const char * defaultYaml = R"!!(
+version: "1.0"
+dfuserver:
+  daliServers: dali
+  enableSNMP: false
```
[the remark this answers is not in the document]
shamser: enableSNMP is unused.  I'm removing this.

dali/dfu/dfuserver.cpp
```diff
@@ -95,7 +96,20 @@ IPropertyTree *readOldIni()
...
+  enableSysLog: true
```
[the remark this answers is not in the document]
shamser: I'll leave this in for now. We can revisit this later.

esp/smc/SMCLib/TpWrapper.cpp
shamser: If the machine is left out, EclWatch despray doesn't display the dropzone.

helm/hpcc/values.schema.json
shamser: The create file uses this umask... I'm not sure if it's necessary.  We may be able to remove this later.

ghalliday: I think a jira to revisit it would make sense.

helm/hpcc/values.yaml
```diff
@@ -368,7 +372,7 @@ roxie:
 thor:
 - name: thor
   prefix: thor
-  numWorkers: 2
+  numWorkers: 4
```
ghalliday, line 375: unrelated change

initfiles/componentfiles/configxml/dfuserver.xsl
```diff
@@ -47,7 +47,7 @@
                <xsl:value-of select="@name"/>
             </xsl:attribute> 
 
-            <xsl:attribute name="daliservers"> <xsl:call-template name="getDaliServers">
+            <xsl:attribute name="daliServers"> <xsl:call-template name="getDaliServers">
```
ghalliday, line 50: this looks sensible - was it deliberate?

ghalliday (commented):
@shamser a few questions.  I'm still wondering whether we should support multiple queues or not - or what it means.  I'll try and think about it some more.

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -2214,8 +2229,37 @@ bool CFileSprayEx::onReplicate(IEspContext &context, IEspReplicate &req, IEspRep
...
+                continue;
+        }
+        else
+        {
+            destFileOut.set(fullDropZoneDir).append(destFileIn);
+        }
+        plane.getProp("@umask", umask);
+        return;
+    }
+    throw MakeStringException(ECLWATCH_DROP_ZONE_NOT_FOUND, "No drop zone configured for %s:%s", destGroup, destFileIn);
```
jakesmith, line 2256: trivial: makeStringException/makeStingExceptionV preferred, should consider MakeStringException deprecated

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -2214,8 +2229,37 @@ bool CFileSprayEx::onReplicate(IEspContext &context, IEspReplicate &req, IEspRep
     return true;
 }
 
+void CFileSprayEx::getDropZoneInfoByDestPlane(double clientVersion, const char* destGroup, const char* destFileIn, StringBuffer& destFileOut, StringBuffer& umask)
+{
+#ifdef _CONTAINERIZED
```
jakesmith, line 2234: probably applicable to bare-metal too eventually, perhaps should compile in, so checked.

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -2335,6 +2382,7 @@ bool CFileSprayEx::onDespray(IEspContext &context, IEspDespray &req, IEspDespray
         PROGLOG("Despray %s", srcname);
         double version = context.getClientVersion();
         const char* destip = req.getDestIP();
+        const char* destGroup = req.getDestGroup();
```
jakesmith, line 2385: is destGroup always a plane in container mode?

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -2214,8 +2229,37 @@ bool CFileSprayEx::onReplicate(IEspContext &context, IEspReplicate &req, IEspRep
     return true;
 }
 
+void CFileSprayEx::getDropZoneInfoByDestPlane(double clientVersion, const char* destGroup, const char* destFileIn, StringBuffer& destFileOut, StringBuffer& umask)
```
jakesmith, line 2232: is destGroup the plane? if so param should be renamed

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -2214,8 +2214,38 @@ bool CFileSprayEx::onReplicate(IEspContext &context, IEspReplicate &req, IEspRep
     return true;
 }
 
+void CFileSprayEx::getDropZoneInfoByDestGroup(double clientVersion, const char* destGroup, const char* destFileIn, StringBuffer& destFileOut, StringBuffer& umask)
```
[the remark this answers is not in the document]
shamser: Other places are referring to this as destGroup and it may be confusing to have the same field refered by 2 different names.  (The ESP services and eclwatch refer to this field as destGroup.)  Shall I rename this to destPlane still?

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -2214,8 +2214,38 @@ bool CFileSprayEx::onReplicate(IEspContext &context, IEspReplicate &req, IEspRep
...
+    VStringBuffer pname("storage/planes[@isDropZone='true'][@name='%s']",destGroup);
+    Owned<IPropertyTreeIterator> planes = queryGlobalConfig().getElements(pname.str());
+    ForEach(*planes)
+    {
+        IPropertyTree & plane = planes->query();
+        StringBuffer fullDropZoneDir(plane.queryProp("@prefix"));
+        addPathSepChar(fullDropZoneDir);
+        if (isDestAbsolutePath)
+        {
+            if (strncmp(fullDropZoneDir, destFileIn, fullDropZoneDir.length())==0)
```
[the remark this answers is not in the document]
shamser: This should work.  For the plane prefix a "azure:storage@container", DestFileIn will be "azure:storage@container/<myfilename>.  The string comparison should work in matching the plane.

esp/scm/ws_fs.ecm
```diff
@@ -414,6 +414,7 @@ ReplicateResponse
 
 ESPrequest Despray
 {
+    string destGroup;
```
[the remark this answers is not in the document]
shamser: Eclwatch has always submitted the dropzone name as the destGroup field.  I have added this field to the Despray esp request structure allows the dropzone name to be accessed from the despray code. (The dropzone name is needed as the ip address is no longer unique.)

esp/services/ws_fs/ws_fsService.cpp
shamser: The destGroup is the dropzone name - in the case of the containerized code, it's the plane. I have not renamed the field as it is used in lots of places as destGroup and I don't want to cause confusion.

esp/services/ws_fs/ws_fsService.cpp
shamser: Yes

esp/smc/SMCLib/TpWrapper.cpp
```diff
@@ -1676,41 +1677,34 @@ void CTpWrapper::getTpDropZones(double clientVersion, const char* name, bool ECL
...
-            machine->setNetaddress(networkAddress.str());
-            machine->setConfigNetaddress("localhost");
-            machine->setDirectory(path);
-            machine->setOS(getPathSepChar(path) == '/' ? MachineOsLinux : MachineOsW2K);
-            tpMachines.append(*machine.getLink());
-            dropZone->setTpMachines(tpMachines);
-            list.append(*dropZone.getLink());
-        }
+    Owned<IPropertyTreeIterator> planes = queryDropZonePlanesIterator(name);
+    ForEach(*planes)
```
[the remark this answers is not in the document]
shamser: There can my multiple matches if name is nullptr.  They are used in two places by the ws_topology ESP service to return a list of dropzones.  All callers will need a list.  (If a single item version was used, the callers would have to create a list and add it to the list before using it as the ESP response.)

helm/hpcc/templates/_helpers.tpl
```diff
@@ -99,6 +99,17 @@ cost:
 {{- end }}
 {{- end -}}
 
+{{/*
+generate dfuserver queues
+*/}}
+{{- define "hpcc.generateConfigDfuQueues" -}}
+ {{- if not .disabled -}}
+{{- range $queue := .root.Values.dfuserver }}
+- name: {{ printf "%s.dfuserver" $queue.name | quote }}
```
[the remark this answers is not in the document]
shamser: We do generate derived values such as for secret, pvc, logging names etc - I think this is similar.

However, the ESP service needs a list of dfu queues.  If we don't provide a list of dfu queues, we'll need to provide a list of dfuservers names in the eclwatch config.

helm/hpcc/values.schema.json
shamser: https://track.hpccsystems.com/browse/HPCC-25579

initfiles/componentfiles/configxml/dfuserver.xsl
shamser: yes, it was.

helm/hpcc/values.yaml
```diff
@@ -386,7 +386,7 @@ roxie:
 thor:
 - name: thor
   prefix: thor
-  numWorkers: 4
+  numWorkers: 2
```
shamser: Undo change inadvertently introduced in a previous commit.

dali/dfu/dfurun.cpp
```diff
@@ -458,10 +459,21 @@ class CDFUengine: public CInterface, implements IDFUengine
...
 
-        Owned<IEnvironmentFactory> factory = getEnvironmentFactory(true);
-        Owned<IConstEnvironment> env = factory->openEnvironment();
+#ifdef _CONTAINERIZED
+        Owned<IPropertyTreeIterator> planes = queryDropZonePlanesIterator();
+        ForEach(*planes)
+        {
+            IPropertyTree & plane = planes->query();
+            StringBuffer fullDropZoneDir(plane.queryProp("@prefix"));
+            if (strncmp(fullDropZoneDir, pfilePath, fullDropZoneDir.length())==0)
```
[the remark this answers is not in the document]
shamser: Bare metal in containerized probably won't work.  Jira for this work: https://track.hpccsystems.com/browse/HPCC-25546

jakesmith (commented):
@shamser - looks good, mostly minor comments/issues/questions.

esp/scm/ws_fs.ecm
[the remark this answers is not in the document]
ghalliday: Since this is a new field it should be called destPlane

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -2379,7 +2429,11 @@ bool CFileSprayEx::onDespray(IEspContext &context, IEspDespray &req, IEspDespray
                 throw MakeStringException(ECLWATCH_INVALID_INPUT, "Despray %s: cannot resolve destination network IP from %s.", srcname, destip);
 
             StringBuffer destfileWithPath, umask;
+#ifdef _CONTAINERIZED
+            getDropZoneInfoByDestPlane(version, destGroup, destfile, destfileWithPath, umask);
```
ghalliday, line 2433: This should be used in bare metal if destPlane is specified - and not supplying it deprecated.

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -2214,8 +2233,35 @@ bool CFileSprayEx::onReplicate(IEspContext &context, IEspReplicate &req, IEspRep
     return true;
 }
 
+void CFileSprayEx::getDropZoneInfoByDestPlane(double clientVersion, const char* destGroup, const char* destFileIn, StringBuffer& destFileOut, StringBuffer& umask)
+{
+    const bool isDestAbsolutePath = isAbsolutePath(destFileIn);
+    Owned<IPropertyTreeIterator> planes = getDropZonePlanesIterator(destGroup);
```
ghalliday, line 2239: Is destGroup required?  If so, isn't there only a single entry for a given plane?

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -2335,6 +2384,7 @@ bool CFileSprayEx::onDespray(IEspContext &context, IEspDespray &req, IEspDespray
         PROGLOG("Despray %s", srcname);
```
ghalliday, line 2384: Is there a jira for gordon to update eclwatch to provide the name?

ghalliday (commented):
@shamser looks good and relatively simple.  A few minor comments.

common/remote/rmtspawn.cpp
```diff
@@ -140,7 +140,13 @@ ISocket * spawnRemoteChild(SpawnKind kind, const char * exe, const SocketEndpoin
         LOG(MCdetailDebugInfo, unknownJob, "Action aborted before connecting to slave (%3d)", replyTag);
         return NULL;
     }
-
+#ifdef _CONTAINERIZED
+    // TODO: In K8s, this won't work in production as there's only a pool of NUM_SLAVE_CONNECT_PORT (20) ports for all ftslave instances
```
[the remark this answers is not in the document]
jakesmith: it's okay, I see now I think.
childEp will always be localhost I think, so it runs ftslave as child process, then connects to it on localhost.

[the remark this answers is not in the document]
shamser: Yes, it does work and the code from 173 is needed.  dfuserver spawns off local ftslave and then it connects to the ftslave via tcp port.  The retries are necessary because the port may have been spawned by a different dfuserver job.  My concern was related to the number of ports available to all the ftslaves - Gavin doesn't think it will be a problem. If another job's ftslave is using the port, the 20 retries will ensure that eventually this job's ftslave will succeed.... I will have another look into this and determine if NUM_SLAVE_CONNECT_PORT needs to be higher for containerized environment: https://track.hpccsystems.com/browse/HPCC-25591

dali/dfu/dfurun.cpp
```diff
@@ -458,10 +459,23 @@ class CDFUengine: public CInterface, implements IDFUengine
...
         StringBuffer netaddress;
         filename.queryIP().getIpText(netaddress);
+#ifdef _CONTAINERIZED
+        Owned<IPropertyTreeIterator> planes = getDropZonePlanesIterator();
+        ForEach(*planes)
+        {
+            IPropertyTree & plane = planes->query();
+            const char * fullDropZoneDir = plane.queryProp("@prefix");
+            assertex(fullDropZoneDir);
+            // note: for bare-metal drop-zones, will need to compare ip address
```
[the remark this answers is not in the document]
shamser: Containerized dfuserver may need to access bare-metal dropzones.

dali/dfu/dfuserver.cpp
```diff
@@ -113,48 +124,49 @@ int main(int argc, const char *argv[])
...
+        return 1;
+    }
+    if (!globals->getProp("@name", name))
+    {
+        PROGLOG("name not specified in configuration");
         return 1;
     }
     Owned<IFile> sentinelFile;
-    bool stop = globals->getPropInt("@STOP",0)!=0;
+    bool stop = globals->getPropBool("@stop", false);
```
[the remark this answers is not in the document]
shamser: It is not used in containerized.  However, all the proceeding code references this field - probably, something we can address in v2.

dali/dfu/dfuserver.cpp
```diff
@@ -222,24 +241,59 @@ int main(int argc, const char *argv[])
...
+        unsigned maxJobs = globals->getPropInt("@maxJobs", 1);
+        if (maxJobs<1)
+        {
+            OERRLOG("maxJobs is %u", maxJobs);
+            return 1;                       
+        }
+        VStringBuffer mask("Queue[@name=\"%s\"][1]", queue.str());
+        if (!stop)
+        {
+            if (serverstatus->queryProperties()->queryPropTree(mask.str()))
```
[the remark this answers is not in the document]
shamser: I'll leave as is for now.  We should re-examine this later.

dali/dfu/dfuserver.cpp
```diff
@@ -222,24 +241,59 @@ int main(int argc, const char *argv[])
...
+                t->setPropInt("@num", maxJobs);
+                serverstatus->queryProperties()->addPropTree("Queue",t);
+            }
+            serverstatus->commitProperties();
+            engine->setDefaultTransferBufferSize((size32_t)globals->getPropInt("@transferBufferSize"));
+            for (unsigned i=0; i<maxJobs; i++)
+                engine->startListener(queue,serverstatus);
+        }
+        else
+            stopDFUserver(queue);
```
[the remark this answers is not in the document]
shamser: I'm removing this code - it doesn't make sense in a container as queues are stopped when container stopped.

esp/services/ws_fs/ws_fsService.cpp
shamser: No changes needed in eclwatch.  This field is already being provided by eclwatch.  It just wasn't defined in the ecm.

esp/services/ws_fs/ws_fsService.cpp
shamser: destGroup (now renamed to destPlane) is required.  Yes, there is only ever one matching.  I'll look into removing the iterator.

esp/services/ws_fs/ws_fsService.cpp
shamser: getDropZoneInfoByIP  uses the environment.xml and getDropZoneInfoByIP expects the dfuserver.yaml configuration.  We can look at combining the functionality of getDropZoneInfoByIP  and getDropZoneInfoByDestPlane in 

There is some additional functionality in getDropZoneInfoByIP which hasn't been implemented in getDropZoneInfoByDestPlane.  There is a isDropZoneRestrictionEnabled configuration field for bare-metal.  

We can do this in this jira: https://track.hpccsystems.com/browse/HPCC-25593

esp/scm/ws_fs.ecm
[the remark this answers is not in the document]
shamser: It's an existing field - it just wasn't defined in the ecm.  Renaming this field will require eclwatch code changes.

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -154,13 +156,24 @@ void CFileSprayEx::init(IPropertyTree *cfg, const char *process, const char *ser
...
+    ForEach(*dfuQueues)
+    {
+        IPropertyTree & dfuQueue = dfuQueues->query();
+        const char * dfuName = dfuQueue.queryProp("@name");
+        if (!isEmptyString(dfuName))
+        {
+            getDfuQueueName(m_QueueLabel, dfuName);
+            getDfuMonitorQueueName(m_MonitorQueueLabel, dfuName);
+            // TODO: not sure how multiple queues would be handled.  (There is no way to select the queue in EclWatch.)
+            // - just use the first one for now
```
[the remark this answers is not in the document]
shamser: Yes, you are right.  I've re-worded the comment and created a jira to look into this later:  https://track.hpccsystems.com/browse/HPCC-25600

dali/dfu/dfuserver.cpp
```diff
@@ -222,24 +241,59 @@ int main(int argc, const char *argv[])
...
             if (stop) {
-                stopDFUserver(q);
+                stopDFUserver(monitorqueue);
             }
             else {
                 IPropertyTree *t=serverstatus->queryProperties()->addPropTree("MonitorQueue",createPTree());
-                t->setProp("@name",q);
-                engine->startMonitor(q,serverstatus,globals->getPropInt("@MONITORINTERVAL",60)*1000);
+                t->setProp("@name",monitorqueue);
+                int monitorInterval = globals->getPropInt("@monitorinterval", 60);
```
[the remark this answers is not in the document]
shamser: The dfuserver.xsd and dfuserver.xsl have the lowercased monitorinterval and monitorqueue.  I didn't change it earlier because I'd have to change the xsd too.  I will change if change doesn't affect backwards compatibility.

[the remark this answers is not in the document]
shamser: I have confirmed that the camel casing these fields affect backward compatibility. I am having to leave the environment.xml unchanged and leaving the .xsd with the non-camel case version.  I'm modifying the  .xsl to generate the camel cased fields from the non-camel cased .xml.

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -282,7 +297,9 @@ static void DeepAssign(IEspContext &context, IConstDFUWorkUnit *src, IEspDFUWork
             dest.setClusterName(clusterNameForDisplay.str());
         }
     }
-
+#else
+    // TODO: work out cluster name
```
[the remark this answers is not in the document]
shamser: tmp is empty at that location.

dali/dfu/dfuserver.cpp
```diff
@@ -241,59 +245,54 @@ int main(int argc, const char *argv[])
...
         unsigned maxJobs = globals->getPropInt("@maxJobs", 1);
         if (maxJobs<1)
         {
             OERRLOG("maxJobs is %u", maxJobs);
-            return 1;                       
+            return 1;
         }
         VStringBuffer mask("Queue[@name=\"%s\"][1]", queue.str());
-        if (!stop)
+        if (serverstatus->queryProperties()->queryPropTree(mask.str()))
```
jakesmith, line 258: don't need [1], could be:
```
VStringBuffer mask("Queue[@name=\"%s\"]", queue.str());
if (serverstatus->queryProperties()->hasProp(mask))
```

dali/dfu/dfuserver.cpp
```diff
@@ -222,24 +245,54 @@ int main(int argc, const char *argv[])
...
+#else
+        getDfuQueueName(queue, name);
+        unsigned maxJobs = globals->getPropInt("@maxJobs", 1);
+        if (maxJobs<1)
+        {
+            OERRLOG("maxJobs is %u", maxJobs);
+            return 1;
+        }
+        VStringBuffer mask("Queue[@name=\"%s\"][1]", queue.str());
+        if (serverstatus->queryProperties()->queryPropTree(mask.str()))
```
jakesmith, line 258: harmless, but per previous [comment](https://github.com/hpcc-systems/HPCC-Platform/pull/14605#discussion_r599484712), this can never be true, because each instance creates it's own status server entry.
There could and will be multiple status server nodes (from different pods), with the same Queue, if replicas>1, but the 1 belonging to this instance will never have a Queue entry until this process sets it.

esp/scm/ws_fs.ecm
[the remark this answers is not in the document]
jakesmith: Confused, in what sense is it an existing field?
Current master definition of 'ESPrequest Despray' doesn't seem to have this field, this changed add +1 new field.
Can you clarify?

helm/hpcc/templates/_helpers.tpl
```diff
@@ -79,7 +79,7 @@ storage:
   planes:
 {{- /*Generate entries for each data plane (removing the pvc).  Exclude the planes used for dlls and dali.*/ -}}
 {{- range $plane := $planes -}}
- {{- if or (not $plane.labels) (has "data" $plane.labels) }}
+ {{- if or (not $plane.labels) (or (has "data" $plane.labels) (has "lz" $plane.labels)) }}
```
jakesmith, line 82: This means all "lz" planes are mounted by all engines.
Should "lz" planes only be accessible to dfuserver?

That doesn't mean to say they can't be both, labels can be : ["data", "lz"]

Incidental thought, I wonder if mounting many planes etc., causes latency when spinning up pods for jobs.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -99,6 +99,18 @@ cost:
 {{- end }}
 {{- end -}}
 
+{{/*
+Generate dfuserver queues
+Pass in root
+*/}}
+{{- define "hpcc.generateConfigDfuQueues" -}}
+{{- range $queue := .root.Values.dfuserver }}
+{{- if not .disabled }}
```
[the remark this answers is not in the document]
shamser: I have moved the disabled check inside the range as per #14605.  Am i missing something?

helm/hpcc/templates/_helpers.tpl
shamser: We are already mounting unnecessary planes e.g dllplanes in dfuserver.  Should generateGlobalConfigMap have types of planes and labels that should be mounted?

richardkchapman: Mounting unneccessary planes should be avoided, I think. Referencing them in the global config file doen;t matter, but mounting them does.

jakesmith: actually, my 1st statement was wrong, this won't mean additional mounts, this is just config.
The data mounts themselves of course are generated via addDataVolumeMount/addDataVolume, and only mount "data" planes.

I don't think it's a big issue, it just means that "lz" planes are seen in the planes list by everyone, not just those that need it (which may just be dfuserver's).

>We are already mounting unnecessary planes e.g dllplanes in dfuserver.

Separate issue, but where? I don't see addDllVolume* used in dfuserver.yaml

jakesmith:
> The existing eclwatch code is calling the topology service setting the field "DestGroup" with the name of the dropzone name. By mentioning that this is an existing field, As this field was not defined in "ESPrequest Despray", it was just being ignored by the ESP.

ok, but as the variable is not in the interface, there are no access methods generated, and therefore the service has never read it, then it's new to the interface, and will have new access methods generated.
So at least should be versioned (min_ver, +interface version bumped) I think.

It's not that not used, the variable name is confusing, even in bare metal.
Clients are setting destGroup to the dropzone name (e.g. "mydropzone"), so group is misleading.
Clients select and set this based on the list returned by TpDropZoneQueue 

The service ignored it though, and only used destIP.

In cloud we can start to use it (and ignore destIP).

So, at least for now, I would leave it as destGroup, service will need changing to start using it in cloud mode.

jakesmith:
I think we need a addLZVolume/addLZVolumeMount to mount the lz's I think, called from dfuserver.yaml,

esp/scm/ws_fs.ecm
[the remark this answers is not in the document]
GordonSmith: FWIW the `destGroup` actually comes from a call to `WsTopology/TpDropZoneQuery` and the response is:
```
{
   "TpDropZoneQueryResponse":{
      "TpDropZones":{
         "TpDropZone":[
            {
               "Name":"mydropzone",
               "Description":"DropZone process",
               "Build":null,
               "Path":"/var/lib/HPCCSystems/mydropzone",
               "ECLWatchVisible":true,
               "UMask":null,
               "TpMachines":{
                  "TpMachine":[
                     {
                        "Name":"ServerList",
                        "Netaddress":"10.0.0.167",
                        "ConfigNetaddress":".",
                        "Domain":null,
                        "Directory":"/var/lib/HPCCSystems/mydropzone",
                        "Type":null,
                        "Available":null,
                        "OS":2,
                        "Path":null,
                        "Port":null,
                        "ProcessNumber":null,
                        "Channels":null
                     }
                  ]
               }
            }
         ]
      }
   }
}
```

helm/hpcc/templates/_helpers.tpl
```diff
@@ -150,17 +150,36 @@ Add ConfigMap volume for a component
...
+{{- $bothLabelsCount := add $planesLabelsCount $includeLabelsCount -}}
+{{- $mergedCount := concat $planeLabels $includeLabels | uniq | len -}}
+{{- ne $bothLabelsCount $mergedCount | printf "%t" -}}
+{{- else -}}
+{{- printf "false" -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Add volume mount
```
jakesmith, line 171: trivial: "mounts" since addVolumesMounts

helm/hpcc/templates/_helpers.tpl
```diff
@@ -174,26 +193,36 @@ If any storage planes are defined that name pvcs they will be mounted
   {{- end }}
  {{- end }}
 {{- end }}
+{{- end -}}
+
+{{/*
+Add data volume mount
+If any storage planes are defined that name pvcs they will be mounted
+*/}}
+{{- define "hpcc.addDataVolumeMount" -}}
```
jakesmith, line 202: not new really, but since [can] mount multiple and calls addVolumeMounts now, this should be Mounts too really

helm/hpcc/templates/dfuserver.yaml
```diff
@@ -0,0 +1,63 @@
...
+          ]
+{{ include "hpcc.addSentinelProbes" . | indent 8 }}
+{{- include "hpcc.addSecurityContext" $commonCtx | indent 8 }}
+{{- include "hpcc.addResources" (dict "me" .resources) | indent 8 }}
+{{ include "hpcc.addImageAttrs" $commonCtx | indent 8 }}
+        volumeMounts:
+{{ include "hpcc.addDataVolumeMount" $commonCtx | indent 8 }}
```
jakesmith, line 53: dfuserver calls addDataVolumeMount .. passing "lz"
I think it would be clearer if addDataVolumeMount called addVoumeMounts with explicit "data" only
and others that wanted different mounts called addVolumeMounts not addData*

helm/hpcc/templates/_helpers.tpl
```diff
@@ -150,17 +150,36 @@ Add ConfigMap volume for a component
...
 {{- /*Create local variables which always exist to avoid having to check if intermediate key values exist*/ -}}
 {{- $storage := (.root.Values.storage | default dict) -}}
 {{- $planes := ($storage.planes | default list) -}}
-{{- $dataStorage := ($storage.dataStorage | default dict) -}}
+{{- $includeLabels := .includeLabels | default list -}}
 {{- range $plane := $planes -}}
  {{- if $plane.pvc -}}
-  {{- if or (not $plane.labels) (has "data" $plane.labels) }}
+  {{- $matchedLabels := include "hpcc.doesStorageLabelsMatch" (dict "plane" $plane "includeLabels" $includeLabels) -}}
+  {{- if or (not $plane.labels) (eq $matchedLabels "true") }}
```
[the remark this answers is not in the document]
shamser: There is not supposed to multiple calls to addVolumeMounts.  It's supposed to be called once with a list of volumes labels to mount.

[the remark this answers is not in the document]
jakesmith: ok yes, as long as not called >1 it's ok, except it should only ever be treating a plane with no labels as data (so should not mount them if just "lz")

helm/hpcc/templates/_helpers.tpl
```diff
@@ -150,36 +150,35 @@ Add ConfigMap volume for a component
...
 {{- /*Create local variables which always exist to avoid having to check if intermediate key values exist*/ -}}
 {{- $storage := (.root.Values.storage | default dict) -}}
+{{- $dataStorage := ($storage.dataStorage | default dict) -}}
 {{- $planes := ($storage.planes | default list) -}}
 {{- $includeLabels := .includeLabels | default list -}}
 {{- range $plane := $planes -}}
  {{- if $plane.pvc -}}
   {{- $matchedLabels := include "hpcc.doesStorageLabelsMatch" (dict "plane" $plane "includeLabels" $includeLabels) -}}
-  {{- if or (not $plane.labels) (eq $matchedLabels "true") }}
+  {{- if ne $matchedLabels "" }}
```
jakesmith, line 181: an empty string is equal to false, so would be more logical/readable as : {{- if $matchedLabels }}

helm/hpcc/templates/_helpers.tpl
```diff
@@ -193,36 +192,35 @@ Pass in root and includeLabels (optional)
   {{- end }}
  {{- end }}
 {{- end }}
+{{- /*Backwards compatibility: when no data planes defined in values, create default pvc*/ -}}
+{{- if and (or (has "data" $includeLabels) (has "" $includeLabels)) (not $dataStorage.plane) }}
```
jakesmith, line 196: I don't think this is the right place for this (in this generalized addVolumeMounts function).
and afaict, it's going to cause calls with includeLabels containing "data" to generate a plane with "data" in main loop and this one as well.

It should only generate this default one if no dataStorage.plane has been defined, but it should do that in addDataVolumeMount 

The special empty ("") case is already being handled by doesStorageLabelsMatch and "" passed in addDataVolumeMount

helm/hpcc/templates/_helpers.tpl
```diff
@@ -193,36 +192,35 @@ Pass in root and includeLabels (optional)
...
 {{- define "hpcc.addDataVolumeMount" -}}
-{{- $storage := (.root.Values.storage | default dict) -}}
-{{- $dataStorage := ($storage.dataStorage | default dict) -}}
-{{- $includeLabels :=  append (.includeLabels | default list) "data" | uniq -}}
-{{- include "hpcc.addVolumeMounts"  (dict "root" .root "includeLabels" $includeLabels) -}}
-{{- if (not $dataStorage.plane) }}
-- name: datastorage
-  mountPath: "/var/lib/HPCCSystems/hpcc-data"
-{{- end }}
+{{- include "hpcc.addVolumeMounts"  (dict "root" .root "includeLabels" (list "data" "")) -}}
```
jakesmith, line 207: trivial: extra space before (dict

helm/hpcc/templates/_helpers.tpl
```diff
@@ -193,36 +192,35 @@ Pass in root and includeLabels (optional)
...
 {{- /*Create local variables which always exist to avoid having to check if intermediate key values exist*/ -}}
 {{- $storage := (.root.Values.storage | default dict) -}}
+{{- $dataStorage := ($storage.dataStorage | default dict) -}}
 {{- $planes := ($storage.planes | default list) -}}
 {{- $includeLabels := .includeLabels -}}
 {{- range $plane := $planes -}}
  {{- if $plane.pvc -}}
   {{- $matchedLabels := include "hpcc.doesStorageLabelsMatch" (dict "plane" $plane "includeLabels" $includeLabels) -}}
-  {{- if or (not $plane.labels) (eq $matchedLabels "true") }}
+  {{- if eq $matchedLabels "true" }}
```
jakesmith, line 223: {{- if $matchedLabels }}

helm/hpcc/templates/_helpers.tpl
```diff
@@ -239,22 +237,20 @@ Pass in root and includeLabels (optional)
   {{- end }}
  {{- end }}
 {{- end -}}
+{{- /*Backwards compatibility: when no data planes defined in values, create default pvc*/ -}}
+{{- if and (or (has "data" $includeLabels) (has "" $includeLabels)) (not $dataStorage.plane) }}
```
jakesmith, line 241: similar to comment in addVolumesMount
shouldn't have "data" specific code in this generalized function, should move to addDataVolume
and "" should already be handled, this is going to generate this when not needed afaics (should only generate if datasStorage.plane is not set)

helm/hpcc/templates/_helpers.tpl
jakesmith: I see now (with Shamser's help) that doesStorageLabelsMatch can generate things like "truetruetrue", so I see why this condition is like it is.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -150,36 +150,35 @@ Add ConfigMap volume for a component
...
 
 {{/*
-Add volume mount
+Add volume mounts
 Pass in root and includeLabels (optional)
 */}}
 {{- define "hpcc.addVolumeMounts" -}}
 {{- /*Create local variables which always exist to avoid having to check if intermediate key values exist*/ -}}
 {{- $storage := (.root.Values.storage | default dict) -}}
+{{- $dataStorage := ($storage.dataStorage | default dict) -}}
```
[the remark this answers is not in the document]
shamser: This is used in line 196

helm/hpcc/templates/_helpers.tpl
shamser: I've commented in the code why this is necessary.

helm/hpcc/templates/_helpers.tpl
jakesmith: or because doesStorageLabelsMatch doesn't return just "true", need to be same as line 181:
```
{{- if ne $matchedLabels "" }}
```

helm/hpcc/templates/_helpers.tpl
shamser: I've commented in the code why this is necessary.

helm/hpcc/templates/_helpers.tpl
shamser: I've added a comment to explain this.

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -2302,6 +2343,9 @@ void CFileSprayEx::getDropZoneInfoByIP(double clientVersion, const char* ip, con
         else
             LOG(MCdebugInfo, unknownJob, "No valid drop zone found for network address '%s'. Check your system drop zone configuration.", ip);
     }
+#else
+    throw makeStringExceptionV(-1, "Internal error: CFileSprayEx::getDropZoneInfoByIP should not be called in containerized environment");
```
jakesmith, line 2347: trivial: should be makeStringException

helm/hpcc/templates/_helpers.tpl
```diff
@@ -162,26 +196,39 @@ If any storage planes are defined that name pvcs they will be mounted
...
+*/}}
+{{- define "hpcc.addVolumes" -}}
 {{- /*Create local variables which always exist to avoid having to check if intermediate key values exist*/ -}}
 {{- $storage := (.root.Values.storage | default dict) -}}
-{{- $planes := ($storage.planes | default list) -}}
 {{- $dataStorage := ($storage.dataStorage | default dict) -}}
-{{- $daliStorage := ($storage.daliStorage | default dict) -}}
-{{- $dllStorage := ($storage.dllStorage | default dict) -}}
+{{- $planes := ($storage.planes | default list) -}}
+{{- $includeLabels := .includeLabels -}}
```
jakesmith, line 227: minor: not needed, or like addVolumeMounts need to default to dict ? : 
```
{{- $includeLabels := .includeLabels | default list -}}
```

jakesmith (approved):
@shamser: 2 v. minor comments, you may or not want to address,
but looks ready to me to be merged.

esp/services/ws_fs/ws_fsService.cpp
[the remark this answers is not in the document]
ghalliday: ?What do you plan to use for this?  Is

    src->getClusterName()

a decent value.  I'm not sure where the results is used.

esp/smc/SMCLib/TpWrapper.cpp
```diff
@@ -1674,6 +1674,38 @@ void CTpWrapper::getDropZoneMachineList(double clientVersion, bool ECLWatchVisib
...
 void CTpWrapper::getTpDropZones(double clientVersion, const char* name, bool ECLWatchVisibleOnly, IArrayOf<IConstTpDropZone>& list)
 {
+
+#ifdef _CONTAINERIZED
+    Owned<IPropertyTreeIterator> planes = getDropZonePlanesIterator(name);
+    ForEach(*planes)
+    {
+        IPropertyTree & plane = planes->query();
+        const char * dropzonename = plane.queryProp("@name");
+        if (isEmptyString(dropzonename))
```
ghalliday, line 1684: Should never occur - name is required for a storage plane/

helm/hpcc/templates/_helpers.tpl
```diff
@@ -138,17 +150,39 @@ Add ConfigMap volume for a component
...
+Pass in plane and includeLabels
+Return: If there is any matching labels, there will be a non-empty string returned.  If there is no matching labels,
+        an empty string will be returned.
 */}}
-{{- define "hpcc.addDataVolumeMount" -}}
+{{- define "hpcc.doesStorageLabelsMatch" -}}
+{{- $plane := .plane -}}
+  {{- range $label := .includeLabels -}}
+    {{- if and (eq $label "") (not $plane.labels) -}}
+      {{- print "true" -}}
```
ghalliday, line 163: minor: might be slightly clearer that non-null means a match if it returned "T" instead of true.

ghalliday (approved):
@shamser a few very minor comments.  Only one that I think definitely should change is the transfer buffer size.

jakesmith (approved):
@shamser - looks good.

The smoketest is failing because it's not rebased on recent 8.0.x and there's a file related to the gh action not present in your branch.
It's really a bug with the smoketest action itself, but rebasing will resolve it.

jakesmith:
> The smoketest is failing because it's not rebased on recent 8.0.x and there's a file related to the gh action not present in your branch.
> It's really a bug with the smoketest action itself, but rebasing will resolve it.

NB: it passed now, because latest smoketest run was a cache miss, and consequently used the standard action/checkout, which does auto-rebase onto latest target branch.

PR #14618, opened 2021-02-25
HPCC-25401 Remove c++98 standard from property for eclregex.cpp

Michael-Gardner:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:


Michael-Gardner:
@xwang2713 please review.  Build is located on jenkins if you want to confirm the logs.

richardkchapman:
@ghalliday Can you review - I'm not convinced this is the right fix but I seem to recall discussing this with you before

rtl/eclrtl/CMakeLists.txt
```diff
@@ -79,16 +79,8 @@ include_directories (
...
-        (ICU_VERSION VERSION_LESS "10.0"))
-        set_source_files_properties(eclregex.cpp PROPERTIES COMPILE_FLAGS -std=c++98)
-    endif ()
-
-    if (WIN32)
-    else ()
-        ADD_DEFINITIONS( -DBOOST_DYN_LINK )
-    endif ()
+if ((USE_BOOST_REGEX) AND (NOT WIN32))
+  ADD_DEFINITIONS( -DBOOST_DYN_LINK )
```
ghalliday, line 83: @richardkchapman 
I think this would mean that we do not support ICU versions < 60 - although earlier versions of the library sources may have been patched to fix the problem.
What is the oldest version that we support (centos6? centos7?) and does that compile with this fix?
It looks like centos 7 may default to v50, which would appear to be likely to hit the issue.

Michael-Gardner: 7 is the oldest CentOS version we're currently supporting and yes, it only has libicu v50 in the repos.  We don't have issues on centOS8 with v60.  If I remove the c++98 standard for that single file, it will compile fine on centos7.

ghalliday (dismissed):
@richardkchapman I think if all our required distributions build then this is good.

richardkchapman (approved):

PR #14642, opened 2021-03-02
HPCC-25358 Rationalize local storage examples.

jakesmith:
1) consolidate on using planes to define storage in local storage example
(consolidate hpcc-localplanes -> hpcc-localfile)

2) extend default to also generate a sasha plane based on a auto-pvc,
and generate settings to use the PVC for sasha services that use storage.

3) add 'labels' and in particular a "data" label. Auto-mount those planes where appropriate
as opposed to previous policy of mounting all except "dll" and "dali".
NB: if labels is blank assume they are "data" (partly for backward compatibility)

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:


jakesmith:
@ghalliday - please review

helm/examples/local/hpcc-localfile/values.yaml
```diff
@@ -1,18 +1,26 @@
...
+  subPath: queries # cannot currently be changed
   size: 1Gi
-  #hostpath: "/var/lib/HPCCSystem/hpcc-data/dali/"
+  rwmany: true
+- name: sasha
+  subPath: sasha
+  size: 1Gi
+  rwmany: true
+- name: data
+  subPath: hpcc-data # cannot currently be changed
```
ghalliday, line 22: is this true?

helm/hpcc/values.schema.json
```diff
@@ -327,6 +327,11 @@
           "description": "which planes (if any) the data is replicated onto (primarily bare metal)",
           "type": "array",
           "items": { "type": "string" }
+        },
+        "labels": {
+          "description": "a list of labels associated with this plane, e.g. lz, data",
```
ghalliday, line 332: should we also include dll, dali?

helm/hpcc/templates/dali.yaml
```diff
@@ -68,6 +68,8 @@ spec:
 {{- end }}
 {{ include "hpcc.addSecretVolumes" (dict "root" $ "categories" (list "system" ) ) | indent 6 }}
 ---
+{{- $storage := ($.Values.storage | default dict) -}}
+{{- $daliStorage := ($storage.daliStorage | default dict) }}
```
ghalliday, line 72: should this default to the first plane with a dali label?  And default dlls/data similarly?

helm/hpcc/templates/dali.yaml
```diff
@@ -77,6 +79,7 @@ data:
     version: 1.0
     dali:
 {{ toYaml (omit $dali "logging" "services") | indent 6 }}
+      dataPath: {{ include "hpcc.getVolumeMountPrefix" (dict "root" $ "name" ($daliStorage.plane | default "dalistorage-pvc") "path" "dalistorage" "me" $daliStorage) }}
```
ghalliday, line 82: is dalistorage-pvc the name of the plane?  If so, that is confusing.

ghalliday (commented):
@jakesmith A few comments from a first look.

helm/examples/local/hpcc-localfile/values.yaml
jakesmith: yes, mainly because there's a few places in the helm chart, but also I think because of getTempFilePath use of hpcc-path.
It could prob. be fixed so can be changed, but perhaps after this.

helm/hpcc/values.schema.json
jakesmith: my thinking was that specific planes, like those two, that they will be named explicitly (as they currently have to be by dllStorage and dataStorage), Dali's could be a property/config option of itself, as it is with the sasha services.

So not sure need additional labels, where named explicitly, and/or can only be 1 (dll).

helm/hpcc/templates/dali.yaml
jakesmith: I wanted to consider getting rid of daliStorage (and prob. dllStorage too),
but thought I'd not require it yet, and leave for separate PR.

helm/hpcc/templates/dali.yaml
jakesmith: that's not right, also not right in a couple of other places.
It will still end up defaulting to the correct path (shouldn't match an existing plane), but should be changed

jakesmith:
@ghalliday - please see new commit

helm/examples/local/hpcc-localfile/values.yaml
ghalliday: added HPCC-25526

helm/hpcc/templates/_helpers.tpl
```diff
@@ -73,13 +73,13 @@ esp:
...
   daliPlane: {{ $daliStoragePlane }}
-  dllsPlane: {{ $daliStoragePlane }}
+  dllsPlane: {{ $dllStoragePlane }}
   dataPlane: {{ $dataStoragePlane }}
   spillPlane: {{ $spillStoragePlane }}
   planes:
 {{- /*Generate entries for each data plane (removing the pvc).  Exclude the planes used for dlls and dali.*/ -}}
 {{- range $plane := $planes -}}
- {{- if and (ne $plane.name $daliStoragePlane) (ne $plane.name $dllStoragePlane) }}
+ {{- if has "data" $plane.labels }}
```
ghalliday, line 82: this will break any existing helm chart that explicitly defines planes.  @richardkchapman is that ok?  At a minimum it needs clear documentation, and (if needed) extending the schema for 7.12.x so the labels can be added ahead of time.

ghalliday: I think it be better if planes defaulted to data (probably the most common) and the test was

    if or (eq 0 (len $planes.labels)) (has "data" $planes.labels)

ghalliday (commented):
@jakesmith  I'm concerned about breaking existing helm charts by requiring the data label.  Should discuss with @richardkchapman  and anyone else interested.

jakesmith:
@ghalliday - as suggested, I've changed it so unlabelled planes are considered default "data" planes, which should mean existing plane definitions in existing values yaml's should behave as before.

jakesmith:
Squashed

helm/examples/local/hpcc-localfile/values.schema.json
```diff
@@ -10,30 +10,55 @@
...
+              "type": "integer"
+            }
+          ]
+        },
+        "rwmany": {
+          "description": "Does the pvc require rw many access (data/dll currently)",
+          "type": "boolean"
+        },
+        "labels": {
+          "description": "a list of labels associated with this plane, e.g. lz, data",
```
ghalliday, line 56: should have a full list?

helm/examples/local/values-localfile.yaml
```diff
@@ -5,16 +5,43 @@
...
+    pvc: dll-localfile-hpcc-localfile-pvc
+    prefix: "/var/lib/HPCCSystems/queries"
+  - name: sasha
+    pvc: sasha-localfile-hpcc-localfile-pvc
+    prefix: "/var/lib/HPCCSystems/sasha"
+  - name: data
+    pvc: data-localfile-hpcc-localfile-pvc
+    prefix: "/var/lib/HPCCSystems/hpcc-data"
+    labels:
+    - data
```
ghalliday, line 28: now out of date - labels missing from other entries?

ghalliday (commented):
@jakesmith one minor comment.

helm/examples/local/values-localfile.yaml
jakesmith: I've now added labels

ghalliday (approved):

PR #14656, opened 2021-03-05
HPCC-25465 Message collator insert method can be n^2

richardkchapman:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Cloud-compatibility
  - [x] Premature optimization

## Smoketest:

### DIFF
--- a/roxie/udplib/udpmsgpk.cpp
+++ b/roxie/udplib/udpmsgpk.cpp
@@ -60,12 +60,20 @@ class PackageSequencer : public CInterface, implements IInterface
 {
     DataBuffer *firstPacket;
     DataBuffer *lastContiguousPacket;
+    DataBuffer *tail = nullptr;
     unsigned metaSize;
     unsigned headerSize;
     const void *header;
+#ifdef _DEBUG
+    unsigned numPackets = 0;
+    unsigned maxSeqSeen = 0;
+    unsigned scans = 0;
+    unsigned overscans = 0;
+#endif
 
     MemoryBuffer metadata;
     InterruptableSemaphore dataAvailable; // MORE - need to work out when to interrupt it!
+
 
 public:
     IMPLEMENT_IINTERFACE;
@@ -122,7 +130,11 @@ public:
         bool res = false;
         assert(dataBuff->msgNext == NULL);
         UdpPacketHeader *pktHdr = (UdpPacketHeader*) dataBuff->data;
-
+        unsigned pktseq = pktHdr->pktSeq;
+#ifdef _DEBUG
+        if ((pktseq & UDP_PACKET_SEQUENCE_MASK) > maxSeqSeen)
+            maxSeqSeen = pktseq & UDP_PACKET_SEQUENCE_MASK;
+#endif
         if (checkTraceLevel(TRACE_MSGPACK, 5))
         {
             StringBuffer s;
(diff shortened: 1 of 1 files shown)


roxie/udplib/udpmsgpk.cpp
```diff
@@ -250,6 +292,12 @@ class PackageSequencer : public CInterface, implements IInterface
         return headerSize;
     }
 
+#ifdef _DEBUG
+    void dump()
+    {
+        DBGLOG("Contains %d packets, lastSeq = %d", numPackets, maxSeqSeen);
```
ghalliday, line 298: really should be %u, instead of %d

ghalliday (commented):
A couple of minor comments, but the code looks good.  II would get Mark to double check,and if he is happy it would be reasonable to include in 6.4.x

ghalliday (approved):

roxie/udplib/udpmsgpk.cpp
```diff
@@ -122,54 +130,85 @@ class PackageSequencer : public CInterface, implements IInterface
...
-                // discard duplicated incoming packet - should be uncommon unless we requested a resend for a packet already in flight
-                if (checkTraceLevel(TRACE_MSGPACK, 5))
-                    DBGLOG("UdpCollator: Discarding duplicate incoming packet");
-                dataBuff->Release();
-                return false;
-            }
-            finger = lastContiguousPacket->msgNext;
-            prev = lastContiguousPacket;
+            finger = nullptr;
+            prev = tail;
```
richardkchapman, line 151: May need tail->msgNext = dataBuffer here...

roxie/udplib/udpmsgpk.cpp
```diff
@@ -178,8 +217,11 @@ class PackageSequencer : public CInterface, implements IInterface
             prev->msgNext = dataBuff;
         }
         else
-            firstPacket = dataBuff;
+            firstPacket = tail = dataBuff;
```
richardkchapman, line 220: This incorrectly overwrites tail if packet 0 is not the first to be seen

roxie/udplib/udpmsgpk.cpp
```diff
@@ -122,54 +130,86 @@ class PackageSequencer : public CInterface, implements IInterface
...
+                }
+                finger = lastContiguousPacket->msgNext;
+                prev = lastContiguousPacket;
             }
-            else if (pktHdr->pktSeq < oldHdr->pktSeq)
+            else
             {
-                break;
+                finger = firstPacket;
+                prev = NULL;
```
jakesmith, line 175: super trivia: could be nullptr

roxie/udplib/udpmsgpk.cpp
```diff
@@ -122,54 +130,86 @@ class PackageSequencer : public CInterface, implements IInterface
...
+                        UdpPacketHeader *oldHdr = (UdpPacketHeader*) lastContiguousPacket->data;
+                        DBGLOG("lastContiguousPacket is at %u , last packet seen is %u", oldHdr->pktSeq & UDP_PACKET_SEQUENCE_MASK, pktHdr->pktSeq & UDP_PACKET_SEQUENCE_MASK);
+                    }
+                    else
+                        DBGLOG("lastContiguousPacket is NULL , last packet seen is %u", pktHdr->pktSeq & UDP_PACKET_SEQUENCE_MASK);
+                    scans = 0;
+                }
+    #endif
+                UdpPacketHeader *oldHdr = (UdpPacketHeader*) finger->data;
+                if (pktHdr->pktSeq == oldHdr->pktSeq)
```
jakesmith, line 196: so finger is walking through from lastContiguousPacket through msgNext,
is this possible, not already rejected at line 161? Or are the chained pktSeq values expected to be > than the lastContiguousPacket->pkgSeq ?

jakesmith (commented):
@richardkchapman - 1 question that probably stems from lack of understanding of impl.

roxie/udplib/udpmsgpk.cpp
richardkchapman: Only the indentation has changed here

roxie/udplib/udpmsgpk.cpp
richardkchapman: The chain is always in order. Newly arriving packets are added via insertion sort. We make this more efficient by (a) tracking the leading already-complete portion and not trying to insert in there and (b) the new code - checking if the addition is at the end - this is true 99.9% of the time, at least if packets arrive in the same order that they were sent.

If packets are NOT received in the same order than they are sent then this change will not really help (but won't make anything any worse either).

jakesmith: makes sense now. Looks fine.

jakesmith (approved):

mckellyln (approved):
Looks good, just a comment about adding scan tracing.

PR #14669, opened 2021-03-10
HPCC-25493 Always create threads at default priority, not inherited from parent

richardkchapman:
Much less error-prone.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jthread.cpp
+++ b/system/jlib/jthread.cpp
@@ -401,6 +401,12 @@ void Thread::startRelease()
                 pthread_attr_setstacksize(&attr, LINUX_STACKSIZE_CAP);
 #endif
         }
+        sched_param param;
+        pthread_attr_getschedparam(&attr, &param);
+        param.sched_priority = 0;
+        pthread_attr_setschedparam(&attr, &param);
+        pthread_attr_setschedpolicy(&attr, SCHED_OTHER);
+        pthread_attr_setinheritsched(&attr, PTHREAD_EXPLICIT_SCHED);
         status = pthread_create(&threadid, &attr, Thread::_threadmain, this);
         if ((status==EAGAIN)||(status==EINTR)) {
             if (numretrys--==0)


mckellyln (approved):
This looks fine.  One minor note about not needing the getschedparam call.

system/jlib/jthread.cpp
```diff
@@ -401,6 +401,12 @@ void Thread::startRelease()
                 pthread_attr_setstacksize(&attr, LINUX_STACKSIZE_CAP);
 #endif
         }
+        sched_param param;
+        pthread_attr_getschedparam(&attr, &param);
```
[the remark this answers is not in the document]
richardkchapman: Is that always true for all systems (now and in the future). The get-then-modify pattern seemed to be the recommended way from my quick searching...

system/jlib/jthread.cpp
```diff
@@ -401,6 +401,12 @@ void Thread::startRelease()
...
+        param.sched_priority = 0;
+        pthread_attr_setschedparam(&attr, &param);
+        pthread_attr_setschedpolicy(&attr, SCHED_OTHER);
```
jakesmith, line 408: If the process has been started with a different riority, and therefore main is at that priority,
does this set new threads to the same priority as the process/main thread, or to a system standard (SCHED_NORMAL) priority?

jakesmith (commented):
@richardkchapman - 1 question.

system/jlib/jthread.cpp
richardkchapman: Testing suggests it ends up same priority as the main thread.

richardkchapman:
@jakesmith Tested via a standalone snippet on a raspberry Pi, it seems to work as desired.

jakesmith (approved):
@richardkchapman looks good to go.

richardkchapman:
@ghalliday Please review/merge.

It's possible to avoid this bug in versions < 8.0 by changing a Roxie setting (to a value that is recommended for production systems anyway) so I think no need to backport

PR #14674, opened 2021-03-10
H25484 Proposal for daliadmin helper functions shared with ESP WsDali service 

wangkx:
## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:


wangkx:
@jakesmith please take a look at the 2 commits here. The 1st commit is the code for building the daliadmin helper functions which are called by daliadmin. Some of those functions may have to be upgraded when used by the WsDali service. The 2nd commit is an example. Please let me know whether this is the correct way to proceed or not. If yes, I may create a separate PR with only the 1st commit for you to review.

dali/daliadmin/daadmin.hpp
```diff
@@ -0,0 +1,89 @@
...
+extern da_decl void dfsreplication(const char *clusterMask, const char *lfnMask, unsigned redundancy, bool dryRun);
+extern da_decl void migrateFiles(const char *srcGroup, const char *tgtGroup, const char *filemask, const char *_options);
+extern da_decl void getxref(const char *dst);
+
+extern da_decl void listworkunits(const char *test, const char *min, const char *max);
+extern da_decl void workunittimings(const char *wuid);
+extern da_decl void dumpWorkunit(const char *wuid, bool includeProgress);
+extern da_decl void dumpProgress(const char *wuid, const char *graph);
+extern da_decl void wuidCompress(const char *match, const char *type, bool compress);
+extern da_decl void dumpWorkunitAttr(IConstWorkUnit *workunit, const WuScopeFilter &filter);
```
jakesmith, line 73: check carefully if function are needed externally, or are truly internal utility methods only.
This one isn't for example, it is only called by the other dumpWorkunitAttr.

I think most are called directly from the command line, so do need to be exported, but worth double checking.

dali/daliadmin/daadmin.cpp
```diff
@@ -274,19 +274,29 @@ void set(const char *path,const char *val)
...
+}
+
+void getValue(const char *path, StringBuffer *value)
 {
     StringBuffer head;
     StringBuffer tmp;
     const char *tail=splitpath(path,head,tmp);
     Owned<IRemoteConnection> conn = querySDS().connect(head.str(),myProcessSession(),RTM_LOCK_READ, daliConnectTimeoutMs);
     if (!conn) {
+        if (value) //called by esp service
```
jakesmith, line 283: rather than contain difference/logic of differences in this function.
I would be better to change get (don't mind if renamed to getValue), to always fill a StringBuffer and change daliadmin cli to do the log "Value of ..."

dali/daliadmin/daadmin.cpp
```diff
@@ -273,12 +273,7 @@ void set(const char *path,const char *val)
 
 //=============================================================================
 
-void get(const char *path)
-{
-    getValue(path, nullptr);
-}
-
-void getValue(const char *path, StringBuffer *value)
+void get(const char *path, StringBuffer *value)
```
jakesmith, line 276: related to 2nd commit comments.
It would be better to unify on a single getValue that always filled a StringBuffer,
then what is done with the result (log vs return to response) is something handled by daliadmin, vs esp service.

jakesmith (commented):
looks good in general, a few comments.
Please address and submit 1st commit as a PR 1st.

dali/daliadmin/daadmin.hpp
wangkx: Sorry, my mistake. Will remove and double check others.

dali/daliadmin/daadmin.cpp
wangkx: Got the idea. The implementation details may be different for each esp service methods. May change the get() in the PR of implementing the esp service get().

wangkx:
The functionalities are implemented. This PR can be closed.

PR #14675, opened 2021-03-10
HPCC-25491  Support Azure LDAP Helm configuration for DALI

RussWhitehead:
Added yaml files for bare metal and azure based configuration

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [x] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:


RussWhitehead:
@afishbeck @jakesmith Please review

AttilaVamos:
I am sure the indexcachebug failure is not related to this PR.

afishbeck:
@RussWhitehead not sure how this works.  ESP uses a scheme like this but it has a mechanism for building a very flexible configuration of services from it's various YAML files.

RussWhitehead:
@afishbeck I followed the ESP pattern, with the expectation that we could vet it out when we have the SDS Migration to Azure complete. Until then, perhaps @jakesmith can clue us in on how the Dali side of Helm configuration works

jakesmith:
@afishbeck @RussWhitehead -  A meta question 1st - shouldn't this be a JIRA/PR about how to configure LDAP in general, with Azure just being a specific example? (and as an example, placed in e.g. helm/examples/ ...)

Secondly, is this essentially attempting to move the legacy configuration out of Dali, into a separate configurable yaml file?

As far as how dali config in cloud mode configuration works - it follows same default pattern that all component configurations follow. There's a defaultYaml in the code, the component calls loadConfiguration and it picks up the component yaml that has been generated based on the component name section from the helm values file/settings/overrides. 

At some point we're going to want to move the whole file service to an esp service, at that point it will be only esp that does any LDAP validation and the Dali calls will be removed.
Therefore we should probably try to avoid tying this to Dali now/with future changes, but we can still generate these LDAP settings so Dali can see them, and later simply move which component/serivce they're generated into.

So I suspect the ldap configuration should be in a separate section in helm values (under the security: section would make sense).

The dali configution (for now, later moved to esp service) can generate this section into its configuration (i.e. see helm/hpcc/templates/dali.yaml configMap generation)

jakesmith (commented):
see comments/questions

afishbeck:
@jakesmith I'm not sure that this current change does anything.  This is essentially a copy of what becomes the ESP defaultYaml for ldap.  Because ESP is a bunch of pluggable components tied together via configuration it's defaultYaml is buit more dynamically.

ESP currently supports helm configuration of it's LDAP security within the "esp:" section, but allowing the LDAP information to be defined in the security section and used globally does make sense.  I think at least for now I'd still allow it to be overridden for a specific ESP just in case.

jakesmith:
@afishbeck - so this is a placeholder?

I think the questions re. whether Azure LDAP is an example, and re. this not specifically being helm config. for Dali  - still hold.

afishbeck:
@jakesmith @RussWhitehead This PR can be closed for now until the real solution reaches PR status.

Hopefully Azure LDAP will be available from the ops team soon so we can test that.  @RussWhitehead are you able to test against your existing LDAP server for that configuration flavor?  There still may be scenarios where we need that even in k8s.

richardkchapman:
Closing as per comments

PR #14693, opened 2021-03-16
HPCC-25448 Remove index qualifiers from subscription notifications

jakesmith:
Each subscriber notify call was sent the absolute xpath that had
triggered the notify.
In practice this was useless, because the index numbers would
likely no longer be valid shortly after they were established.
Additionally, no code used these serialized xpaths, other than
for tracing.

The primary reason for removing them now however, is that a race
condition was exposed, that could cause Dali to crash whilst
these absolute index numbers were being established.
The notify threads creating these absolute paths, could
overlap with other commits, which changed the relative position
of the ptree's involved in the change, and in particular the
'arrayOwner' validity. Therefore the index could be wrong,
or worse hit an invalid arrayOwner ptr and crash.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dali/base/dasds.cpp
+++ b/dali/base/dasds.cpp
@@ -1752,15 +1752,6 @@ void buildNotifyData(MemoryBuffer &notifyData, PDState state, CPTStack *stack, M
                 PTree &child = stack->item(s);
                 const char *str = child.queryName();
                 notifyData.append(strlen(str), str);
-                unsigned pos = child.querySiblingPosition();
-                if (0 == pos) // common case
-                    notifyData.append(3, "[1]");
-                else
-                {
-                    char temp[12];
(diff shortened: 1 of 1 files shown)


jakesmith:
@ghalliday - please review

jakesmith:
Is 7.12.x the right target?
It affects all versions, but has never been seen afaik, other than when testing in K8s.

richardkchapman:
7.12 or 8.0 is fine by me

ghalliday (approved):

PR #14698, opened 2021-03-17
HPCC-25262 Metrics milestone 1

kenrowland:
Implement and collect a counter metric

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [ ] Cloud-compatibility
  - [x] Premature optimization

## Smoketest:


system/jlib/jmetrics.hpp
```diff
@@ -0,0 +1,295 @@
...
+
+    protected:
+        std::string name;
+        std::string type;
+        MetricsReporter *pReporter = nullptr;
+};
+
+extern "C" { typedef hpccMetrics::MetricSink* (*getSinkInstance)(const char *, const IPropertyTree *pSettingsTree); }
+
+struct SinkInfo
```
ghalliday, line 252: better as an opaque type define within the cpp file.

system/jlib/jmetrics.cpp
```diff
@@ -0,0 +1,219 @@
...
+    }
+}
+
+
+std::vector<std::shared_ptr<IMetric>> MetricsReporter::queryMetricsForReport(const std::string &sinkName)
+{
+    std::vector<std::shared_ptr<IMetric>> reportMetrics;
+
+    auto it = sinks.find(sinkName);
+    if (it != sinks.end())
```
ghalliday, line 113: else throw an internal error.
Also - with this implementation the name can't be optional.

system/jlib/jmetrics.cpp
```diff
@@ -0,0 +1,219 @@
...
+    {
+        //
+        // Lock the list of metrics while it's in use
+        std::unique_lock<std::mutex> lock(metricVectorMutex);   // no one else can mess with it for a bit
```
ghalliday, line 117: worth reserving the space in the reportMetrics vector before appending, otherwise this could be O(N^2) for very large numbers of metrics.

system/jlib/jmetrics.cpp
```diff
@@ -0,0 +1,219 @@
...
+{
+    bool rc = true;
+    for (pSinkIt->first(); pSinkIt->isValid() && rc; pSinkIt->next())
+    {
+        SinkInfo *pSinkInfo;
+        Owned<IPropertyTree> pSinkTree = &pSinkIt->get(); // pSinkIt->query();
+
+        StringBuffer cfgSinkType, cfgSinkName;
+        pSinkTree->getProp("@type", cfgSinkType);  // this one is required
+        pSinkTree->getProp("@name", cfgSinkName);
```
ghalliday, line 152: report an error if type (or name) is missing.

system/jlib/jmetrics.cpp
```diff
@@ -0,0 +1,219 @@
...
+bool MetricsReporter::initializeSinks(IPropertyTreeIterator *pSinkIt)
+{
+    bool rc = true;
+    for (pSinkIt->first(); pSinkIt->isValid() && rc; pSinkIt->next())
+    {
+        SinkInfo *pSinkInfo;
+        Owned<IPropertyTree> pSinkTree = &pSinkIt->get(); // pSinkIt->query();
+
+        StringBuffer cfgSinkType, cfgSinkName;
+        pSinkTree->getProp("@type", cfgSinkType);  // this one is required
```
ghalliday, line 151: queryProp() is generally more efficient - preferred if the IPropertyTree will definitely remain alive.

esp/bindings/http/platform/httpprot.cpp
```diff
@@ -455,13 +455,14 @@ bool CHttpThread::onRequest()
         httpserver.setown(new CEspHttpServer(*m_socket, m_apport, m_viewConfig, getMaxRequestEntityLength()));
     }
 
-    time_t t = time(NULL);  
+    time_t t = time(NULL);
     initThreadLocal(sizeof(t), &t);
 
     m_httpserver = httpserver;
     httpserver->setSocketReturner(this);
```
ghalliday, line 462: unrelated change

system/jlib/jmetrics.hpp
```diff
@@ -0,0 +1,295 @@
...
+{
+    public:
+        virtual ~MetricSink() = default;
+        virtual void startCollection(MetricsReporter *pReporter) = 0;
+        virtual void stopCollection() = 0;
+        const std::string &queryName() const { return name; }
+        const std::string &queryType() const { return type; }
+
+    protected:
+        MetricSink(std::string _name, std::string _type) :
```
ghalliday, line 240: not sure why std::string parameters are used here.  (The base classes don't have them.)  More normal is to pass const char *

ghalliday (changes requested):
@kenrowland 
I have reviewed the code.  I commented on most things, but some items are much more significant than others.

I have a few overall comments on the current structure:

* process for reporting metrics:

Currently the process for gathering the metrics is as follows:

    auto reportMetrics = pReporter->queryMetricsForReport(name);
    MeasurementVector measurements;
    for (auto &pMetric: reportMetrics)
        pMetric->collect(measurements);
    writeMeasurementsToFile(measurements);

i.e. a vector of all the metrics is created, and then a vector of measurements.  With large numbers of metrics that will add a significant overhead.

A simpler and more efficient approach would be to directly walk the measurements and report the values by using members of the metric class e.g.

for (auto & cur : measurements)
   report(cur.queryName(), cur.queryValue(), cur....)

which avoids creating the temporary vector of measurements.

Note:
You can take this further and also avoid creating the reportMetrics list by providing a callback interface to report the metrics through.  (This is the approach of my demo branch.)  The disadvantage is that the metrics would need to be locked while the metrics are reported (unlikely to be an issue in practice as long unless the sink can block).

* Simplify types

Unless there is an example where it is useful we should assume all metrics have 8 byte unsigned values.  It would simplify the interface to metrics, no need for template classes and simplify reporting them.

* What objects types are registered.  (Not proposing changing)

I am still in two minds whether metrics should be registered with the metric reporter, or whether it should be a more general metric reporter.  Registering metrics makes it possible to filter metrics more efficiently (if that is a significant requirement), and has fewer complications with object lifetimes.  Registering reporters would reduce the overhead if large numbers of metrics are reported, and may make it easier to integrate stats.   I don't propose we change anything, but there might come a time where we want to revisit this.

devdoc/Metrics.rst
```diff
@@ -0,0 +1,335 @@
...
+This section discusses configuration. Since Helm charts are capable of combining configuration data
+at a global level into a component's specific configuration, The combined configuration takes the
+form as shown below. Note that as the design progresses it is expected that there will be additions.
+
+::
+
+  component:
+    metrics:
+      sinks:
+        sinK:
```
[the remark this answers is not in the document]
kenrowland: This is what I had to do, based on information from Tony, in order to parse the configuration independent of the source, either environment.xml or YML (through my test program).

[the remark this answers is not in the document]
afishbeck: I think my comment was about YAML that matched an existing xml based PTree.  That shouldn't be reflected in the format of the helm values themselves.  But as Gavin says that should become clear when the actual helm integration is done.

kenrowland (commented):
@ghalliday Please look at these changes

ghalliday (commented):
@kenrowland that's looking good.  A few (hopefully) final questions/comments.

ghalliday:
Also, I think you misunderstood my comment about 
add_definitions( -D_METRICS_FILESINK_LIB_EXPORTS)
You do need code as you had - my comment was it should be something simpler like:
add_definitions( -DFILESINK_EXPORTS)

system/jlib/jmetrics.cpp
kenrowland: Here is my thinking. The mutex is protecting access to the map of metrics managed by the reporter. The map contains weak pointers, so it does not maintain a reference count. When a sink needs its list of metrics to report, each metric weak pointer is converted to a shared pointer which is an atomic operation. The shared pointer is placed in a vector and returned to the sink, thus ensuring it has a reference until the sink destroys the vector. During that time if another sink requests its list, the pointer is already locked, so the conversion to shared increments the reference count. 

However, during a report let's say the component destroys the metric by clearing the shared pointer. The metric object will remain alive since the report vector contains a shared pointer to the metric and thus a reference. When the report is completed and the sink destroys the report vector, the ref count goes to 0 and the metric is destroyed (note that the weak pointer remains in the managed list by the reporter). The next time a metric report list is requested, the conversion to a shared pointer will fail and the metric is removed from the list. 

If I am missing something, please let me know.

system/jlib/jmetrics.cpp
```diff
@@ -0,0 +1,219 @@
...
+    auto it = metrics.find(pMetric->queryName());
+    if (it == metrics.end())
+    {
+        metrics.insert({pMetric->queryName(), pMetric});
+    }
+    // todo - need to decide how to handle duplicate named metrics
+}
+
+
+void MetricsReporter::removeMetric(const std::string &metricName)
```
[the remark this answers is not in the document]
kenrowland: Clearing the shared pointer will work (see comments later). Since that will be the most efficient method, I will remove this call.

kenrowland (commented):
@ghalliday Please see latest changes and comments.

system/jlib/jmetrics.cpp
ghalliday: I agree with your overall statement.

My comment was in reference to the statement "A planned optimization is to only build this vector when there is a change".  I assumed that mean you were thinking of preserving the vector from call to call.  If that was done the vector would contain shared_ptrs, which would mean the weak_ptr never returned null.

system/jlib/jmetrics.hpp
```diff
@@ -0,0 +1,203 @@
...
+    std::vector<std::shared_ptr<IMetric>> queryMetricsForReport(const std::string &sinkName);
+
+protected:
+    void initializeSinks(IPropertyTreeIterator *pSinkIt);
+    static MetricSink *getSinkFromLib(const char *type, const char *sinkName, const IPropertyTree *pSettingsTree);
+
+protected:
+    StringBuffer componentPrefix;
+    StringBuffer globalPrefix;
+    std::map<std::string, std::shared_ptr<SinkInfo>> sinks;
```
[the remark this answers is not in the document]
ghalliday: I think use std::unique_ptr until there is a reason to make it shared.

system/jlib/jmetrics.hpp
```diff
@@ -0,0 +1,202 @@
...
+    }
+};
+
+
+
+class jlib_decl MetricSink
+{
+public:
+    virtual ~MetricSink() = default;
+    virtual void startCollection(MetricsReporter *pReporter) = 0;
```
ghalliday, line 148: minor: This looks like it should be passed to the getSinkInstance() function, but I wouldn't change for this PR.

system/metrics/sinks/file/fileSink.hpp
```diff
@@ -0,0 +1,51 @@
...
+#include "jstring.hpp"
+#include "jfile.ipp"
+#include "jsem.hpp"
+#include <thread>
+#include <condition_variable>
+
+
+using namespace hpccMetrics;
+
+class DECL_EXPORT FileMetricSink : public MetricSink
```
ghalliday, line 28: This should be based on a conditional define.  You had code previously to implement it, but I commented that the name was verbose/did not match the general pattern used elsewhere.

ghalliday (commented):
@kenrowland lloking very good.  A few final minor comments.

kenrowland (commented):
@ghalliday Please review latest changes. Thanks.

ghalliday (approved):
@kenrowland great.  Please squash that last commit and I will merge.

PR #14715, opened 2021-03-19
HPCC-25475 Address Coverity scan results in Couchbase plugin

dcamper:
In addition to this PR, we should update the two submodules used by this plugin to point to the most current master branch commit.  The submodule's repos were forked by hpcc-systems:

https://github.com/hpcc-systems/libcouchbase.git
https://github.com/hpcc-systems/libcouchbase-cxx.git

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

## Testing:

Via regression tests.

### DIFF
--- a/plugins/couchbase/couchbaseembed.cpp
+++ b/plugins/couchbase/couchbaseembed.cpp
@@ -181,6 +181,8 @@ namespace couchbaseembed
             auto status = pQcmd->named_param(cbPlaceholder.str(), utf8);
             if (!status.success())
                 failx("Could not bind Param: %s val: %s", cbPlaceholder.str(), utf8);
+            if (utf8)
+                rtlFree(utf8);
         }
         else
             failx("Internal error: detected invalid CouchbaseQueryCommand while attempting to bind to field: %s", cbPlaceholder.str());
@@ -215,6 +217,8 @@ namespace couchbaseembed
             auto status = pQcmd->named_param(cbPlaceholder.str(), (char *)data);
             if (!status.success())
                 failx("Could not bind Param: %s val: %s", cbPlaceholder.str(), (char *)data);
+            if (data)
+                rtlFree(data);
         }
         else
             failx("Internal error: detected invalid CouchbaseQueryCommand while attempting to bind to field: %s", cbPlaceholder.str());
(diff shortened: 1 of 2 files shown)


dcamper:
FYI:  Two Coverity errors for out-of-bounds access were declared false positives.

dcamper:
Also: I recommend that the Coverity findings for the submodules be ignored for now, until we update our submodules to newer code.  It is probable that some or all of those findings have been addressed with newer code.

dcamper:
I should add that these changes were successfully tested against Couchbase Server 7.0.0 (beta) and with both the existing submodules and with updated submodules.

plugins/couchbase/couchbaseembed.cpp
```diff
@@ -912,32 +916,33 @@ namespace couchbaseembed
 
     IPropertyTree * CouchbaseEmbedFunctionContext::nextResultRowTree()
     {
-        for (auto cbrow : *m_pQuery)
+        if (m_pQuery)
         {
-            auto json = cbrow.json().to_string();
+            reportIfQueryFailure(m_pQuery);
```
rpastrana, line 921: this makes sense. Per our offline conv, I'm wondering if reportIfQueryFailure should fail if m_pQuery contains multiple queries?
I'm not sure how often we'll encounter that situation, but I'm concerned about silently dropping the other queries.

rpastrana (commented):
@dcamper looks good, just one comment which might not apply.

plugins/couchbase/couchbaseembed.cpp
dcamper: Well, with the original code, we never supported multiple queries anyway.  If we change anything, I would suggest only documenting this as a known limitation and then address it if the need arises.

rpastrana: Sounds reasonable

rpastrana (dismissed):

dcamper:
@rpastrana Please check my README addendum.

rpastrana (approved):
@dcamper looks reasonable for now

PR #14744, opened 2021-03-25
HPCC-25331 Fix packagemap lookup of cluster information in cloud

afishbeck:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/smc/SMCLib/TpWrapper.cpp
+++ b/esp/smc/SMCLib/TpWrapper.cpp
@@ -2077,6 +2077,7 @@ class CContainerWUClusterInfo : public CSimpleInterfaceOf<IConstWUClusterInfo>
     StringAttr thorQueue;
     ClusterType platform;
     unsigned clusterWidth;
+    StringArray thorProcesses;
 
 public:
     CContainerWUClusterInfo(const char* _name, const char* type, unsigned _clusterWidth)
@@ -2087,6 +2088,7 @@ public:
         {
             thorQueue.set(getClusterThorQueueName(queue.clear(), name));
             platform = ThorLCRCluster;
+            thorProcesses.append(name);
         }
         else if (strieq(type, "roxie"))
         {
@@ -2156,11 +2158,12 @@ public:
     }
     virtual IStringVal & getRoxieProcess(IStringVal & str) const override
     {
-        UNIMPLEMENTED;
+        str.set(name.get());
+        return str;
     }
     virtual const StringArray & getThorProcesses() const override
     {
-        UNIMPLEMENTED;
+        return thorProcesses;
     }
     virtual const StringArray & getPrimaryThorProcesses() const override
     {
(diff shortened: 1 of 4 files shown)


afishbeck:
@richardkchapman I'm adding some new roxie cluster info, can you verify those 3 values?

esp/smc/SMCLib/TpWrapper.cpp
```diff
@@ -2077,21 +2077,30 @@ class CContainerWUClusterInfo : public CSimpleInterfaceOf<IConstWUClusterInfo>
...
+            thorProcesses.append(name);
         }
         else if (strieq(type, "roxie"))
         {
             agentQueue.set(getClusterEclAgentQueueName(queue.clear(), name));
             platform = RoxieCluster;
+            if (cfg)
+            {
+                roxieRedundancy = (unsigned) cfg->getPropInt("@replicas", roxieRedundancy);
+                channelsPerNode = (unsigned) cfg->getPropInt("@numChannels", channelsPerNode);
```
richardkchapman, line 2102: numChannels is total number of channels in the cluster, not per node

channels per node is always 1 in containerized system. frequently 2 but could be more in bare metal

richardkchapman (changes requested):
channelsPerNode code looks wrong

richardkchapman:
@afishbeck

afishbeck:
@richardkchapman updated channelsPerNode.  These values only affect cloud.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -570,6 +570,8 @@ Generate instance queue names
   type: roxie 
   prefix: {{ .prefix | default "null" }}
   queriesOnly: true
+  replicas: {{ .replicas | default 2 }}
```
richardkchapman, line 573: Not convinced that this is right. Apart form anything else the default in roxie.yaml is 1.

I don't know if there's some way to make the default be only specified in one place...

richardkchapman (changes requested):
One comment

helm/hpcc/templates/_helpers.tpl
afishbeck: Actually, I don't think this replicas value has anything to do with what I want in the cluster info.  I think the cluster info is more data replica?  And in cloud we leave that up to the cloud environment?  I'll remove this setting from here.  And in the cluster info should probably just return 1.

afishbeck:
@richardkchapman updated.

richardkchapman (approved):

wangkx (approved):
Looks fine for me

PR #14754, opened 2021-03-26
HPCC-25628 Report compressed part sizes in WsDFU.DFUInfo

wangkx:
A new integer field CompressedSize is added to report the size.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/services/ws_dfu/ws_dfuService.cpp
+++ b/esp/services/ws_dfu/ws_dfuService.cpp
@@ -2042,6 +2042,7 @@ void CWsDfuEx::getFilePartsOnClusters(IEspContext &context, const char *clusterR
     IDistributedFile *df, IEspDFUFileDetail &fileDetails)
 {
     double version = context.getClientVersion();
+    bool compressedFile = isFileKey(df) || df->isCompressed();
     IArrayOf<IConstDFUFilePartsOnCluster>& partsOnClusters = fileDetails.getDFUFilePartsOnClusters();
     ForEachItemIn(i, clusters)
     {
@@ -2062,6 +2063,7 @@ void CWsDfuEx::getFilePartsOnClusters(IEspContext &context, const char *clusterR
             unsigned partIndex = part.queryPartIndex();
 
             __int64 size = -1;
+            __int64 compressedSize = -1;
             StringBuffer partSizeStr;
             IPropertyTree *partPropertyTree = &part.queryProperties();
             if (!partPropertyTree)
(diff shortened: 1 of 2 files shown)


wangkx:
Cc: @GordonSmith

esp/services/ws_dfu/ws_dfuService.cpp
```diff
@@ -2083,6 +2087,13 @@ void CWsDfuEx::getFilePartsOnClusters(IEspContext &context, const char *clusterR
                 FilePart->setPartsize(partSizeStr.str());
                 if (version >= 1.38)
                     FilePart->setPartSizeInt64(size);
+                if (compressedFile && (version >= 1.58))
+                {
+                    if (compressedSize != -1)
+                        FilePart->setCompressedSize(compressedSize);
+                    else
+                        FilePart->setCompressedSize(size);
```
jakesmith, line 2095: It looks like this is to handle the index case (isFileKey(df)==true) ?

Is it correct though, to set both size and compressed size to same in index case? It didn't return this information previously, but Eclwatch does show both.

jakesmith (commented):
@wangkx - 1 question

esp/services/ws_dfu/ws_dfuService.cpp
wangkx: Yes, this is to handle the index case (isFileKey(df)==true).
The code here is based the the existing code line 2285-2286.

wangkx:
@jakesmith sorry, did I answer your questions?

esp/services/ws_dfu/ws_dfuService.cpp
jakesmith: >The code here is based the the existing code line 2285-2286.

ok, so same behaviour as top-level properties, now at the part level.

It's probably ok. Indexes are always compressed, so there isn't really a separate size vs compressedSize - there could be if we wanted to calculate the uncompressed data size, but doubt we're going to do that.
So really only the compressed size is correct here for indexes, however I suspect even if it was a good idea, removing size, leaving only compressed size, would cause compatibility issues.

Leave it as it is.

jakesmith (approved):
@wangkx - looks good to merge

wangkx:

PR #14759, opened 2021-03-29
HPCC-25606 Improve cache behaviour for concurrent page requests

ghalliday:
## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/roxie/ccd/ccdfile.cpp
+++ b/roxie/ccd/ccdfile.cpp
@@ -618,13 +618,20 @@ struct CacheInfoEntry
     {
         struct
         {
+#ifndef _WIN32
             unsigned type: 2;    // disk or the kind of index node
             __uint64 page: 38;   // Support file sizes up to 2^51 i.e. 2PB
             unsigned file: 24;   // Up to 4 million files
+#else
+//Windows does not like packing bitfields with different base types - fails the statck assert
+            __uint64 type: 2;    // disk or the kind of index node
+            __uint64 page: 38;   // Support file sizes up to 2^51 i.e. 2PB
+            __uint64 file: 24;   // Up to 4 million files
+#endif
         } b;
         __uint64 u;
     };
-    static_assert(sizeof(b) == sizeof(u), "Unexpected packing issue");
+    static_assert(sizeof(b) == sizeof(u), "Unexpected packing issue in CacheInfoEntry");
 
     inline CacheInfoEntry() { u = 0; }
     inline CacheInfoEntry(unsigned _file, offset_t _pos, PageType pageType)
--- a/roxie/ccd/ccdstate.cpp
+++ b/roxie/ccd/ccdstate.cpp
@@ -2372,6 +2372,12 @@ private:
                 topology->setPropInt("@leafCacheMem", leafCacheMB);
                 setLeafCacheMem(leafCacheMB * 0x100000);
             }
+            else if (stricmp(queryName, "control:legacyNodeCache")==0)
+            {
+                bool legacyNodeCache = control->getPropBool("@val", true);
+                topology->setPropInt("@legacyNodeCache", legacyNodeCache);
+                setLegacyNodeCache(legacyNodeCache);
+            }
             else if (stricmp(queryName, "control:listFileOpenErrors")==0)
             {
                 // this just creates a delta state file to remove references to Keys / Files we now longer have interest in
--- a/system/jhtree/ctfile.hpp
+++ b/system/jhtree/ctfile.hpp
@@ -193,11 +193,12 @@ class jhtree_decl CNodeBase : public CWritableKeyNode
 {
 protected:
     NodeHdr hdr;
-    byte keyType;
     size32_t keyLen;
     size32_t keyCompareLen;
     CKeyHdr *keyHdr;
+    byte keyType;
     bool isVariable;
+    std::atomic<bool> ready{false}; // is this node read for use?  Can be checked outside a critsec, but only set within one.
 
 private:
     offset_t fpos;
@@ -213,6 +214,8 @@ public:
     inline bool isLeaf() const { return hdr.leafFlag != NodeBranch; }       // actually is-non-branch.  Use should be reviewed.
     inline NodeType getNodeType() const { return (NodeType)hdr.leafFlag; }
 
+    inline bool isReady() const { return ready; }
+    inline void noteReady() { ready = true; }
 public:
     CNodeBase();
     void load(CKeyHdr *keyHdr, offset_t fpos);
--- a/system/jhtree/jhtree.cpp
+++ b/system/jhtree/jhtree.cpp
@@ -598,13 +598,9 @@ typedef OwningSimpleHashTableOf<CNodeMapping, CKeyIdAndPos> CNodeTable;
 #define FIXED_NODE_OVERHEAD (sizeof(CJHTreeNode))
 class CNodeMRUCache : public CMRUCacheOf<CKeyIdAndPos, CJHTreeNode, CNodeMapping, CNodeTable>
 {
-    size32_t sizeInMem, memLimit;
+    std::atomic<size32_t> sizeInMem{0};
+    size32_t memLimit = 0;
 public:
-    CNodeMRUCache(size32_t _memLimit) : memLimit(0)
-    {
-        sizeInMem = 0;
-        setMemLimit(_memLimit);
-    }
     size32_t setMemLimit(size32_t _memLimit)
     {
         size32_t oldMemLimit = memLimit;
@@ -618,6 +614,12 @@ public:
         // remove LRU until !full
         do
         {
+            //Never evict an entry that hasn't yet loaded - otherwise the sizeInMem can become inconsistent
+            CNodeMapping *tail = mruList.tail();
+            assertex(tail);
+            if (!tail->queryElement().isReady() )
+                break;
+
             clear(1);
         }
         while (full());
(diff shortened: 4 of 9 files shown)


ghalliday:
Untested.  It really needs testing in an environment where lots of threads are trying to access the same pages, but they are getting evicted.  It is worth discussing whether we think this will improve things or make things worse.

Commoning up the branch code is likely to be more efficient (smaller code cache) - but does add a couple of extra indirections.

The behaviour if the page is in the cache should be very similar.  The behaviour when it not is different

Before:
Lock, unlock.  Load the page.  Lock, check if it has been added, otherwise add

After:
Lock, add if missing, unlock.   Lock a page-dependent-cr load release lock.

So there likely to be the same number of critical section locks.  Loading a page will contend on a different lock though - so it should reduce contention with loading existing pages - and be less contended.  There will be a limit on the number of nodes concurrently being loaded from memory with the new code, where it was unlimited before.

I'm tempted to rename some of the loadNode() functions since the name is a bit overloaded.

system/jhtree/jhtree.cpp
```diff
@@ -2334,134 +2357,106 @@ extern jhtree_decl void getNodeCacheInfo(ICacheInfoRecorder &cacheInfo)
...
-    leafCache.reportEntries(cacheInfo);
+    for (unsigned i = 0; i < CacheMax; i++)
+        cache[i].reportEntries(cacheInfo);
 }
 
+constexpr StatisticKind addStatId[CacheMax] = { StNumNodeCacheAdds, StNumLeafCacheAdds, StNumBlobCacheAdds };
+constexpr StatisticKind hitStatId[CacheMax] = { StNumNodeCacheHits, StNumLeafCacheHits, StNumBlobCacheHits };
+constexpr RelaxedAtomic<unsigned> * hitMetric[CacheMax] = { &nodeCacheHits, &leafCacheHits, &blobCacheHits };
+constexpr RelaxedAtomic<unsigned> * addMetric[CacheMax] = { &nodeCacheAdds, &leafCacheAdds, &blobCacheAdds };
+constexpr unsigned numLoadCritSects = 16;
```
richardkchapman, line 2368: How did you decide on 16? Feels a little low?

system/jhtree/jhtree.cpp
```diff
@@ -2334,134 +2357,106 @@ extern jhtree_decl void getNodeCacheInfo(ICacheInfoRecorder &cacheInfo)
...
+
     //NOTE: TLK leaf nodes are currently cached along with branches, not with leaves.  It might be better if this was a separate cache.
-    NodeType cacheType = isTLK ? NodeBranch : type;
+    CacheType cacheType = isTLK ? CacheBranch : (CacheType)type;
 
-    // No benefit in caching the following, especially since they will evict useful pages
-    // could also check cache[cacheType] if an array was used and avoid the tests in the critical section (change later)
-    if ((cacheType == NodeMeta) || (cacheType == NodeBloom))
-    {
+    // check cache[cacheType] avoid the critical section (and testign the flag within the critical section)
```
richardkchapman, line 2385: typo in comment

system/jhtree/jhtree.cpp
```diff
@@ -2334,134 +2357,106 @@ extern jhtree_decl void getNodeCacheInfo(ICacheInfoRecorder &cacheInfo)
...
+                cache[cacheType].add(key, *node);
             }
-            if (ctx) ctx->noteStatistic(StNumBlobCacheAdds, 1);
-            blobCacheAdds++;
-            blobCache.add(key, *LINK(node));
+            node->Link();
         }
-        break;
+
+        //Actually load the information outside the critical section
```
richardkchapman, line 2448: This comment feels like it should be lower down. The check for isReady() is likely for the same reason that the check for whether it was found in the cache is likely - i.e. we hope that the cache is effective!

system/jhtree/jhtree.cpp
```diff
@@ -2334,134 +2357,106 @@ extern jhtree_decl void getNodeCacheInfo(ICacheInfoRecorder &cacheInfo)
...
+        if (likely(node->isReady()))
+            return node;
+
+        //Reather than having a critical section in each node, use an array indexed by a hash of the node address
```
richardkchapman, line 2452: Typo Reather

richardkchapman: It's a shame to recalculate the hash that we presumably just calculated above to look up in the cache. Maybe not a big deal...

richardkchapman: It's a shame to recalculate the hash that we presumably just calculated above to look up in the cache. Maybe not a big deal...

system/jhtree/ctfile.hpp
```diff
@@ -193,11 +193,12 @@ class jhtree_decl CNodeBase : public CWritableKeyNode
 {
 protected:
     NodeHdr hdr;
-    byte keyType;
     size32_t keyLen;
     size32_t keyCompareLen;
     CKeyHdr *keyHdr;
+    byte keyType;
     bool isVariable;
+    std::atomic<bool> ready{false};
```
richardkchapman, line 201: A comment on what this means, and that it is updated only within a critical block, might be useful

richardkchapman (approved):
Minor comments only

system/jhtree/jhtree.cpp
ghalliday: no reason.  I will change to 64.

roxie/ccd/ccdfile.cpp
```diff
@@ -618,13 +618,20 @@ struct CacheInfoEntry
     {
         struct
         {
+#ifndef _WIN32
             unsigned type: 2;    // disk or the kind of index node
             __uint64 page: 38;   // Support file sizes up to 2^51 i.e. 2PB
             unsigned file: 24;   // Up to 4 million files
+#else
+//Windows does not like packing bitfields with different base types - fails the statck assert
```
richardkchapman, line 626: typo

ghalliday: I will push a fix.

roxie/ccd/ccdstate.cpp
```diff
@@ -2372,6 +2372,12 @@ class CRoxiePackageSetManager : implements IRoxieQueryPackageManagerSet, impleme
                 topology->setPropInt("@leafCacheMem", leafCacheMB);
                 setLeafCacheMem(leafCacheMB * 0x100000);
             }
+            else if (stricmp(queryName, "control:legacyNodeCache")==0)
+            {
+                bool legacyNodeCache = control->getPropBool("@val", true);
+                topology->setPropInt("@legacyNodeCache", legacyNodeCache);
+                setLegacyNodeCache(legacyNodeCache);
```
mckellyln, line 2379: Would this require existing node/leaf/blob cache flush/re-init ?

system/jhtree/jhtree.cpp
```diff
@@ -2333,148 +2368,156 @@ extern jhtree_decl void getNodeCacheInfo(ICacheInfoRecorder &cacheInfo)
...
+        //The common case is that this flag has already been set (by a previous add).
+        if (likely(node->isReady()))
+            return node;
+
+        //Shame that the hash code is recalculated - it might be possible to remove this.
+        unsigned hashcode = hashc(reinterpret_cast<const byte *>(&key), sizeof(key), 0x811C9DC5);
+        unsigned whichCs = hashcode % numLoadCritSects;
+        Owned<CJHTreeNode> ownedNode(node); // ensure node gets cleaned up if it fails to load
+
+        //Actually load the information outside the critical section
```
jakesmith, line 2492: perhaps clarify comment to say outside of the cache node type crit, to distinguish from crit used to avoid double loads.

system/jhtree/jhtree.cpp
```diff
@@ -618,6 +614,12 @@ class CNodeMRUCache : public CMRUCacheOf<CKeyIdAndPos, CJHTreeNode, CNodeMapping
         // remove LRU until !full
         do
         {
+            //Never evict an entry that hasn't yet loaded - otherwise the sizeInMem can become inconsistent
+            CNodeMapping *tail = mruList.tail();
+            assertex(tail);
+            if (!tail->queryElement().isReady() )
+                break;
```
jakesmith, line 621: not important afaict, but the implication is that that when add is called and full, there could be a few nodes that are in the process of loading at tail, than prevent add->makespace reducing amount used, so goes above mem. limit for a short time... until a future add that calls makespace and reduces below limit again. ?

If true, perhaps a comment here?

jakesmith (approved):
@ghalliday - only v. minor comments. Looks good to me.

system/jhtree/jhtree.cpp
```diff
@@ -2333,148 +2368,156 @@ extern jhtree_decl void getNodeCacheInfo(ICacheInfoRecorder &cacheInfo)
...
+            (*addMetric[cacheType])++;
         }
-        break;
+
+        //The common case is that this flag has already been set (by a previous add).
+        if (likely(node->isReady()))
+            return node;
+
+        //Shame that the hash code is recalculated - it might be possible to remove this.
+        unsigned hashcode = hashc(reinterpret_cast<const byte *>(&key), sizeof(key), 0x811C9DC5);
```
mckellyln, line 2488: I'm not entirely clear on how unique &key hashc are and why the 32bit init value.

mckellyln (commented):
Very good idea.
I'm trying to quickly review, just have two comments/questions so far.

roxie/ccd/ccdstate.cpp
ghalliday: no, I think it can change dynmically.

system/jhtree/jhtree.cpp
```diff
@@ -618,6 +614,12 @@ class CNodeMRUCache : public CMRUCacheOf<CKeyIdAndPos, CJHTreeNode, CNodeMapping
         // remove LRU until !full
         do
         {
+            //Never evict an entry that hasn't yet loaded - otherwise the sizeInMem can become inconsistent
+            CNodeMapping *tail = mruList.tail();
+            assertex(tail);
+            if (!tail->queryElement().isReady() )
```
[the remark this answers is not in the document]
ghalliday: That isn't the same because this doesn't pop something from the list.

PR #14777, opened 2021-03-30
HPCC-25630 Metrics milestone 2 - Add a logging sink

kenrowland:
Added metrics sink to write metric collection reports to the log

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [ ] Cloud-compatibility
  - [x] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jlog.hpp
+++ b/system/jlib/jlog.hpp
@@ -849,6 +849,7 @@ constexpr LogMsgCategory MCdebugInfo(MSGAUD_programmer, MSGCLS_information, Debu
 constexpr LogMsgCategory MCauditInfo(MSGAUD_audit, MSGCLS_information, AudMsgThreshold);
 constexpr LogMsgCategory MCstats(MSGAUD_operator, MSGCLS_progress, ProgressMsgThreshold);
 constexpr LogMsgCategory MCoperatorInfo(MSGAUD_operator, MSGCLS_information, InfoMsgThreshold);
+constexpr LogMsgCategory MCmetrics(MSGAUD_operator, MSGCLS_metric, ErrMsgThreshold);
 
 /*
  * Function to determine log level (detail) for exceptions, based on log message class
--- a/system/metrics/sinks/file/fileSink.cpp
+++ b/system/metrics/sinks/file/fileSink.cpp
@@ -13,7 +13,7 @@
 
 #include "fileSink.hpp"
 #include <cstdio>
-#include <thread>
+#include "platform.h"
 
 using namespace hpccMetrics;
 
@@ -25,82 +25,38 @@ extern "C" MetricSink* getSinkInstance(const char *name, const IPropertyTree *pS
 
 
 FileMetricSink::FileMetricSink(const char *name, const IPropertyTree *pSettingsTree) :
-        MetricSink(name, "file"),
-        collectionPeriodSeconds{60}
+    PeriodicMetricSink(name, "file", pSettingsTree)
 {
-    if (pSettingsTree->hasProp("@period"))
-    {
-        collectionPeriodSeconds = pSettingsTree->getPropInt("@period");
-    }
-
     pSettingsTree->getProp("@filename", fileName);
     clearFileOnStartCollecting = pSettingsTree->getPropBool("@clear", false);
 }
 
 
-FileMetricSink::~FileMetricSink()
-{
-    if (isCollecting)
-    {
-        doStopCollecting();
-    }
-}
-
-
-void FileMetricSink::startCollection(MetricsReporter *_pReporter)
+void FileMetricSink::prepareToStartCollecting()
 {
     fhandle = fopen(fileName.str(), clearFileOnStartCollecting ? "w" : "a");
-    pReporter = _pReporter;
-    isCollecting = true;
-    collectThread = std::thread(&FileMetricSink::collectionThread, this);
 }
 
-
-void FileMetricSink::collectionThread()
+void FileMetricSink::doCollection()
 {
-    //
-    // The initial wait for the first report
-    waitSem.wait(collectionPeriodSeconds * 1000);
-    while (!stopCollectionFlag)
+    auto reportMetrics = pReporter->queryMetricsForReport(name);
+    writeReportHeaderToFile();
+    for (auto &pMetric: reportMetrics)
     {
-        auto reportMetrics = pReporter->queryMetricsForReport(name);
-        writeReportHeaderToFile();
-        for (auto &pMetric: reportMetrics)
-        {
-            writeMeasurementToFile(pMetric->queryName(), pMetric->queryValue(), pMetric->queryDescription());
-        }
-
-        // Wait again
-        waitSem.wait(collectionPeriodSeconds * 1000);
+        writeMeasurementToFile(pMetric->queryName(), pMetric->queryValue(), pMetric->queryDescription());
     }
 }
 
 
-void FileMetricSink::stopCollection()
+void FileMetricSink::collectingHasStopped()
 {
-    if (isCollecting)
-    {
-        doStopCollecting();
-    }
-}
-
-
-void FileMetricSink::doStopCollecting()
-{
-    //
-    // Set the stop collecting flag, then signal the wait semaphore
-    // to wake up and stop the collection thread
-    stopCollectionFlag = true;
-    waitSem.signal();
-    isCollecting = false;
-    collectThread.join();
     fclose(fhandle);
 }
 
 
-void FileMetricSink::writeMeasurementToFile(const std::string &metricName, uint32_t value, const std::string &metricDescription) const
+void FileMetricSink::writeMeasurementToFile(const std::string &metricName, __uint64 value, const std::string &metricDescription) const
 {
-    fprintf(fhandle, "  %s -> %d, %s\n", metricName.c_str(), value, metricDescription.c_str());
+    fprintf(fhandle, "  %s -> %" I64F "d, %s\n", metricName.c_str(), value, metricDescription.c_str());
     fflush(fhandle);
 }
 
--- a/system/metrics/sinks/log/logSink.cpp
+++ b/system/metrics/sinks/log/logSink.cpp
@@ -1,0 +1,45 @@
+/*##############################################################################
+    HPCC SYSTEMS software Copyright (C) 2021 HPCC Systems®.
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+############################################################################## */
+
+#include "logSink.hpp"
+#include "jlog.hpp"
+
+using namespace hpccMetrics;
+
+extern "C" MetricSink* getSinkInstance(const char *name, const IPropertyTree *pSettingsTree)
+{
+    MetricSink *pSink = new LogMetricSink(name, pSettingsTree);
+    return pSink;
+}
+
+
+LogMetricSink::LogMetricSink(const char *name, const IPropertyTree *pSettingsTree) :
+    PeriodicMetricSink(name, "log", pSettingsTree)
+{
+}
+
+
+void LogMetricSink::doCollection()
+{
+    auto reportMetrics = pReporter->queryMetricsForReport(name);
+    for (auto &pMetric: reportMetrics)
+    {
+        writeLogEntry(pMetric);
+    }
+}
+
+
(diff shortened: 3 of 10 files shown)


kenrowland:
@ghalliday @rpastrana Please review.

Few notes:

- Created new periodic sink base class to handle both log and file sinks since they operate the same way
- Removed 'sink' from the sink type when specifying via settings. Now added automatically (this means the _type_ setting value is now _log_ or _file_ instead of _logsink_ or _filesink_ (more intuitive since they are all sinks.
- Some cleanup of unused header references in the metrics lib.

system/metrics/testing/metrictest.cpp
```diff
@@ -64,13 +64,26 @@ const char *testConfigYml = R"!!(component:
...
+
+
 int main(int argc, char *argv[])
 {
     InitModuleObjects();
 
     //
     // Simulate retrieving the component and global config
-    Owned<IPropertyTree> pSettings = createPTreeFromYAMLString(testConfigYml, ipt_none, ptr_ignoreWhiteSpace, nullptr);
+    Owned<IPropertyTree> pSettings = createPTreeFromYAMLString(testFileSinkConfigYml, ipt_none, ptr_ignoreWhiteSpace, nullptr);
```
rpastrana, line 86: couldn't tell, but is this only built/packaged in DEBUG mode?

rpastrana (commented):
@kenrowland a few comments

system/jlib/jlog.hpp
```diff
@@ -845,6 +845,7 @@ constexpr LogMsgCategory MCdebugInfo(MSGAUD_programmer, MSGCLS_information, Debu
 constexpr LogMsgCategory MCauditInfo(MSGAUD_audit, MSGCLS_information, AudMsgThreshold);
 constexpr LogMsgCategory MCstats(MSGAUD_operator, MSGCLS_progress, ProgressMsgThreshold);
 constexpr LogMsgCategory MCoperatorInfo(MSGAUD_operator, MSGCLS_information, InfoMsgThreshold);
+constexpr LogMsgCategory MCmetrics(MSGAUD_operator, MSGCLS_all, CriticalMsgThreshold);
```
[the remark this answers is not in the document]
ghalliday: MSGCLS_all is used for filters, not for logging.  I think there should probably be a new MSGCLS_metric, which should be used here.  I will look at adding one.

system/metrics/sinks/file/CMakeLists.txt
```diff
@@ -18,11 +18,13 @@ project(hpccmetrics_filesink)
 
 set ( srcs
         fileSink.cpp
+        ${HPCC_SOURCE_DIR}/system/metrics/sinks/common/periodicSink.cpp
```
ghalliday, line 21: We try and avoid including the same source file in multiple places.

system/metrics/sinks/common/periodicSink.cpp
```diff
@@ -0,0 +1,83 @@
+/*##############################################################################
+    HPCC SYSTEMS software Copyright (C) 2021 HPCC Systems®.
```
ghalliday, line 2: This makes sense to be commoned up - but should probably live in jmetrics.cpp along with the MetricSink base class.  (It would avoid including it in multiple projects.)

system/metrics/sinks/log/logSink.cpp
```diff
@@ -0,0 +1,45 @@
...
+    for (auto &pMetric: reportMetrics)
+    {
+        writeLogEntry(pMetric);
+    }
+}
+
+
+void LogMetricSink::writeLogEntry(const std::shared_ptr<IMetric> &pMetric)
+{
+    LOG(MCmetrics, "name=%s,value=%ld,description=%s", pMetric->queryName().c_str(), pMetric->queryValue(), pMetric->queryDescription().c_str());
```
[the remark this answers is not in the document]
ghalliday: I don't think you will want the description being logged each time.

ghalliday, line 44: %ld isn't the correct format specifier for a 64bit int.  i think %" I64F "d  or %lld.

system/metrics/sinks/log/logSink.hpp
```diff
@@ -0,0 +1,42 @@
...
+
+using namespace hpccMetrics;
+
+#ifdef LOGSINK_EXPORTS
+#define LOGSINK_API DECL_EXPORT
+#else
+#define LOGSINK_API DECL_IMPORT
+#endif
+
+class LOGSINK_API LogMetricSink : public PeriodicMetricSink
```
ghalliday, line 31: minor: This class could be in the c++ file.  In fact this file could probably be removed with everything in the c++ file.  No need for the LOGSINK_API on the class if it is in the c++ file because it can't be used outside the dll.

system/metrics/testing/metrictest.cpp
```diff
@@ -49,13 +49,13 @@ const char *localConfigYml = R"!!(roxie:
...
 
-const char *testConfigYml = R"!!(component:
+const char *testFileSinkConfigYml = R"!!(component:
   metrics:
     name: config_name
     prefix: component_prefix.
     sinks:
         sink:
-          - type: filesink
+          - type: log
```
ghalliday, line 58: shouldn't this be filesink still?

ghalliday (commented):
@kenrowland a few minor comments, but generally looks good

system/metrics/sinks/log/logSink.cpp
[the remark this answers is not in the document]
rpastrana: I agree @ghalliday it seems wasteful and was going to make the same comment, I didn't make the comment because I know Prometheus seems to support reporting the description, but it would be good to know why @kenrowland decided to report it.

system/metrics/testing/metrictest.cpp
rpastrana: Conditionally exclude the entire testing subproject based on "${CMAKE_BUILD_TYPE}"

AttilaVamos:
@kenrowland I don't think the workflow_9c (parallel=false) failure is related to this PR.

system/metrics/testing/metrictest.cpp
ghalliday: I don't think it is a problem for testing code to be built in release mode - it is often useful.

ghalliday: (provided it does not significantly increase the size of the deployed system.)  This is separate so it wouldn't.

system/metrics/sinks/file/fileSink.cpp
```diff
@@ -25,82 +25,38 @@ extern "C" MetricSink* getSinkInstance(const char *name, const IPropertyTree *pS
...
-    collectThread.join();
     fclose(fhandle);
 }
 
 
-void FileMetricSink::writeMeasurementToFile(const std::string &metricName, uint32_t value, const std::string &metricDescription) const
+void FileMetricSink::writeMeasurementToFile(const std::string &metricName, uint64_t value, const std::string &metricDescription) const
 {
-    fprintf(fhandle, "  %s -> %d, %s\n", metricName.c_str(), value, metricDescription.c_str());
+    fprintf(fhandle, "  %s -> %" PRIu64 ", %s\n", metricName.c_str(), value, metricDescription.c_str());
```
ghalliday, line 59: Unfortunately PRIu64 is a C99 extension, and not part of c++ (it is a bit of a mess).

system/metrics/sinks/file/fileSink.hpp
```diff
@@ -31,27 +27,21 @@ using namespace hpccMetrics;
...
-        bool isCollecting = false;
-        bool clearFileOnStartCollecting = false;
-        FILE *fhandle = nullptr;
-        Semaphore waitSem;
+public:
+    explicit FileMetricSink(const char *name, const IPropertyTree *pSettingsTree);
+    ~FileMetricSink() override = default;
+
+protected:
+    void prepareToStartCollecting() override;
```
ghalliday, line 37: minor: As a general style, we always include the virtual keyword - even when it isn't needed.  Primarily to improve comprehension.

system/metrics/sinks/log/logSink.cpp
```diff
@@ -0,0 +1,46 @@
...
+    for (auto &pMetric: reportMetrics)
+    {
+        writeLogEntry(pMetric);
+    }
+}
+
+
+void LogMetricSink::writeLogEntry(const std::shared_ptr<IMetric> &pMetric)
+{
+    LOG(MCmetrics, "name=%s,value=%" PRIu64, pMetric->queryName().c_str(), pMetric->queryValue());
```
ghalliday, line 45: same comment about PRIu64

system/jlib/jlog.hpp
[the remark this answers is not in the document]
ghalliday: The change has been merged to 8.0.x  please rebase and use the new class.

ghalliday (commented):
@kenrowland very close.  A few last comments.

system/metrics/testing/metrictest.cpp
rpastrana: @ghalliday how do you envision this being used in "release"? I can see perhaps as part of smoke test or regression tests, but I still contend this and code like it shouldn't be packaged in our official release akin to Java junit tests which are built with the project, executed if desired, but not part of the target package.

ghalliday: Well it should be using the cppunit framework that is used by all the other unit tests.  (Which is deployed on a separate option.)  The unit tests are often run in release as well as debug mode.  I agree it shouldn't be deployed as part of our standard release.

kenrowland (commented):
@ghalliday @rpastrana Please take a look at the latest changes. 

Also, note that I changed the name of the templated function that creates a metric and adds it to the reporter. The new name is more descriptive. It required a change to the TxSummary file. Better change just one than multiple.

system/jlib/jlog.hpp
```diff
@@ -849,6 +849,7 @@ constexpr LogMsgCategory MCdebugInfo(MSGAUD_programmer, MSGCLS_information, Debu
 constexpr LogMsgCategory MCauditInfo(MSGAUD_audit, MSGCLS_information, AudMsgThreshold);
 constexpr LogMsgCategory MCstats(MSGAUD_operator, MSGCLS_progress, ProgressMsgThreshold);
 constexpr LogMsgCategory MCoperatorInfo(MSGAUD_operator, MSGCLS_information, InfoMsgThreshold);
+constexpr LogMsgCategory MCmetrics(MSGAUD_operator, MSGCLS_metric, CriticalMsgThreshold);
```
ghalliday, line 852: minor:  It doesn't make too much difference... I'm not sure what detail level it should be, but critical seems wrong.  ErrMsgThreshold might be more appropriate.

system/metrics/sinks/log/logSink.cpp
```diff
@@ -0,0 +1,47 @@
...
+    for (auto &pMetric: reportMetrics)
+    {
+        writeLogEntry(pMetric);
+    }
+}
+
+
+void LogMetricSink::writeLogEntry(const std::shared_ptr<IMetric> &pMetric)
+{
+//    LOG(MCmetrics, "name=%s,value=%" PRIu64, pMetric->queryName().c_str(), pMetric->queryValue());
```
ghalliday, line 44: trivial: Clean up the commented out lines.

system/metrics/testing/metrictest.cpp
```diff
@@ -98,7 +111,7 @@ int main(int argc, char *argv[])
     pEventCountMetric = std::make_shared<CounterMetric>("requests", "The number of requests");
     myReporter.addMetric(pEventCountMetric);
 
-    pQueueSizeMetric = std::make_shared<GaugeMetric>("queuesize", "request queue size");
+    pQueueSizeMetric = createMetricAndAddToReporter<GaugeMetric>("queuesize", "request queue size");
     myReporter.addMetric(pQueueSizeMetric);
```
ghalliday, line 115: I think this should report an error for a duplicate metric.  It would be a good unit test, but I don't think it is intended.

ghalliday (commented):
@kenrowland a few very small comments.  Please can you fix and also squash the commits.  Hopefully it will be ready for merging.

system/jlib/jlog.hpp
kenrowland: Changed to ErrMsgThreshold per suggestion. I used critical prior to ensure that metrics were logged regardless.

system/metrics/sinks/log/logSink.cpp
kenrowland: Removed.

system/metrics/testing/metrictest.cpp
kenrowland: The framework should throw an error if a duplicate named metric is added. However, this was not the intent of the change. I was adding a use of the function template that creates and adds a metric. Removed the extra addMetric call.

kenrowland (commented):
Final changes.

ghalliday (approved):

PR #14789, opened 2021-03-31
HPCC-25489 LDAP credentials should be configurable as secrets

RussWhitehead:
Enabled LDAP Admin username/pwd to be optionally configured to be retrieved from
secrets repo (Vault or local secrets). Added ESP category to secrets in values.yaml,
added secret Key and Vault ID to configuration, and updated LDAP startup to optionally
query values from repo if keys specified in configuration

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/helm/hpcc/values.yaml
+++ b/helm/hpcc/values.yaml
@@ -188,6 +188,10 @@ secrets:
     ## For example, to set the secret associated with the azure storage account "mystorageaccount" use
     ##azure-mystorageaccount: storage-myazuresecret
 
+  authn: {}
+    ## Category for LDAP Admin username/password
+    MyLDAPServerKey: MyLDAPServerValue
+
   ecl: {}
     ## Category for secrets published to all components that run ecl
 
@@ -216,6 +220,8 @@ secrets:
 
 vaults:
   storage:
+
+  authn:
 
   ecl:
 
--- a/initfiles/componentfiles/configschema/xsd/ldapserver.xsd
+++ b/initfiles/componentfiles/configschema/xsd/ldapserver.xsd
@@ -42,6 +42,10 @@
                               hpcc:tooltip="Time in minutes after which the cached security information should expire"/>
                 <xs:attribute name="sharedCache" type="xs:boolean" hpcc:displayName="Shared Cache" hpcc:presetValue="true"
                               hpcc:tooltip="Use a single, shared LDAP cache"/>
+                <xs:attribute name="systemUserVaultID" type="xs:string" hpcc:displayName="System User Vault ID"
+                              hpcc:tooltip="Optional ID of secrets vault containing the secret LDAP AD systemuser username/password."/>
+                <xs:attribute name="systemUserSecretKey" type="xs:string" hpcc:displayName="System User Secret Key"
+                              hpcc:tooltip="Optional secrets repo key name of secret containing LDAP AD systemuser username/password. Specify this instead of systemUser/systemPassword if available."/>
                 <xs:attribute name="systemUser" type="xs:string" hpcc:displayName="System User"
                               hpcc:tooltip="An LDAP administrator account id to be used by HPCC to create and manage HPCC-specific LDAP branches"/>
                 <xs:attribute name="systemPassword" type="xs:string" hpcc:displayName="System User Password" hpcc:modifers="mask,verify"
--- a/system/security/LdapSecurity/ldapconnection.cpp
+++ b/system/security/LdapSecurity/ldapconnection.cpp
@@ -30,6 +30,7 @@
 #include "dautils.hpp"
 #include "dasds.hpp"
 #include "workunit.hpp"
+#include "jsecrets.hpp"
 
 #include <map>
 #include <string>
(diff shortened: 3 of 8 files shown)


richardkchapman:
@RussWhitehead Should this target 8.0.x ?

RussWhitehead:
@richardkchapman Will rebase with next commit. @afishbeck Please review

system/security/LdapSecurity/ldapconnection.cpp
```diff
@@ -351,39 +352,78 @@ class CLdapConfig : implements ILdapConfig, public CInterface
...
+            if (!sb.isEmpty())
+                WARNLOG("Please specify either systemCommonName or systemUserSecretKey in configuration");
+
+            StringBuffer vaultId;
+            cfg->getProp(".//@systemUserVaultID", vaultId);//optional HashiCorp vault ID
+
+            DBGLOG("Retrieving LDAP Admin username/password from secrets repo: %s %s", !vaultId.isEmpty() ? vaultId.str() : "", systemUserSecretKey.str());
+            Owned<IPropertyTree> secretTree;
+            if (!isEmptyString(vaultId.str()))
+                secretTree.setown(getVaultSecret("esp", vaultId, systemUserSecretKey.str(), nullptr));
```
afishbeck, line 376: "esp" as the category worked as a proof of concept, but if components other than ESP (like DALI, any others?) then the category has to be something that identifies the group of components.  We need a new category like "authn".  Because we don't want the secrets to show up anywhere except the exact components that need them... and there is no current category that would get it on only ESP and DALI.  A good example of a category like that is "codeVerify".. it's secrets end up on ESP and eclccserver.

initfiles/componentfiles/configschema/xsd/ldapserver.xsd
```diff
@@ -42,6 +42,10 @@
                               hpcc:tooltip="Time in minutes after which the cached security information should expire"/>
                 <xs:attribute name="sharedCache" type="xs:boolean" hpcc:displayName="Shared Cache" hpcc:presetValue="true"
                               hpcc:tooltip="Use a single, shared LDAP cache"/>
+                <xs:attribute name="systemUserVaultID" type="xs:string" hpcc:displayName="System User Vault ID"
```
afishbeck, line 45: I may have missed something, but these two values need to be settable through the helm chart.  
Well actually we have 2 choices... 1. use a well known name as the secret name and search for it in our new category.  or 2. add helm configuration as you've done here for configmgr.

If you want to use a well defined name like for example "ldapSystemUser" under the new secrets category of "authn" then you just have to change your call to getSecret below to:

secretTree.setown(getSecret("authn", "ldapSystemUser"));

That method will search the category local secrets and vaults to find the secret with that name.

afishbeck (changes requested):

helm/hpcc/values.yaml
```diff
@@ -188,6 +188,10 @@ secrets:
     ## For example, to set the secret associated with the azure storage account "mystorageaccount" use
     ##azure-mystorageaccount: storage-myazuresecret
 
+  authn: {}
+    ## Category for LDAP Admin username/password
+    MyLDAPServerKey: MyLDAPServerValue
```
afishbeck, line 193: This example should be removed or commented out.  If someone uses this and the secret doesn't exist their pods won't run.. so this should only be added when the user really wants to use ldap.

system/security/LdapSecurity/ldapconnection.cpp
```diff
@@ -351,39 +352,82 @@ class CLdapConfig : implements ILdapConfig, public CInterface
...
+        DBGLOG("@@%s",xml.str());//@@
+//@@
+        StringBuffer systemUserSecretKey;
+        cfg->getProp(".//@systemUserSecretKey", systemUserSecretKey);//secrets repo key
+        if (!systemUserSecretKey.isEmpty())
+        {
+            StringBuffer sb;
+            cfg->getProp(".//@systemCommonName", sb);
+            if (!sb.isEmpty())
+                WARNLOG("Please specify either systemCommonName or systemUserSecretKey in configuration");
```
afishbeck, line 372: Warning is a little confusing saying that you have to specify one of them, but really the issue is that they specified both.  Should probably be something like:

 WARNLOG("Please specify either systemCommonName or systemUserSecretKey but not both in configuration");

system/security/LdapSecurity/ldapconnection.cpp
```diff
@@ -351,39 +352,82 @@ class CLdapConfig : implements ILdapConfig, public CInterface
...
+
+            StringBuffer vaultId;
+            cfg->getProp(".//@systemUserVaultID", vaultId);//optional HashiCorp vault ID
+
+            DBGLOG("Retrieving LDAP Admin username/password from secrets repo: %s %s", !vaultId.isEmpty() ? vaultId.str() : "", systemUserSecretKey.str());
```
afishbeck, line 377: StringBuffer::str() won't return nullptr so 
```
!vaultId.isEmpty() ? vaultId.str() : ""
```
is redundant, just "vaultId.str()" should be enough.

system/security/LdapSecurity/ldapconnection.cpp
afishbeck: You've added the authn category above, so you can replace "esp" with "authn" here and below.

afishbeck (changes requested):
@RussWhitehead several comments.

PR #14822, opened 2021-04-06
HPCC-25694 Default file/workunit access problems with new deployment

RussWhitehead:
Create LDAP Workunit/Files features with user permissions, so they can be
applied to "default" permissions when no workunit/file scope exists

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [x] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/security/LdapSecurity/ldapconnection.cpp
+++ b/system/security/LdapSecurity/ldapconnection.cpp
@@ -1551,9 +1551,9 @@ public:
             //Create base LDAP OU tree. Specify PT_ADMINISTRATORS_ONLY to ensure each OU
             //grants access to Administrators only
             createLdapBasedn(NULL, m_ldapconfig->getResourceBasedn(RT_DEFAULT), PT_ADMINISTRATORS_ONLY);
-            createLdapBasedn(NULL, m_ldapconfig->getResourceBasedn(RT_FILE_SCOPE), PT_ADMINISTRATORS_ONLY);
+            createLdapBasedn(NULL, m_ldapconfig->getResourceBasedn(RT_FILE_SCOPE), PT_DEFAULT);
             createLdapBasedn(NULL, m_ldapconfig->getResourceBasedn(RT_VIEW_SCOPE), PT_ADMINISTRATORS_ONLY);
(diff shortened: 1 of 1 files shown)


RussWhitehead:
@wangkx Kevin please review

wangkx (commented):
This fix does not tight security. I am wondering if there is a better fix. For example, default to no access and allow it being modified. @afishbeck any comment?

afishbeck:
Working on other security features and using other platforms I would try and make it so the system would never accidentally be insecure.  That is, it's best if we error on the side of defaulting to something restrictive but provide an easy way of overriding it if the user wants to.  Not sure exactly what that would mean here.  @RussWhitehead can you comment on what that would mean?

RussWhitehead:
@afishbeck @wangkx This Jira is meant to fix legacy behavior that has presumably worked this way since LDAP was implemented, and is documented in the HPCC manuals.  I believe this Jira should be merged to restore that broken feature.
Separately, I agree that the default should be to lock down access and force admins to open it up, but this will take more thought. 
1) We should do the same for file scopes as well as workunit scopes. Perhaps more?
2) We need to provide guidance on how to open the access. It is very frustrating to a new user when the "nothing works" out of the box. Perhaps a wizard tool that walks the admin through initial security configuration
3) Perhaps some of this could be done at initial configuration. A new "Install Securely" VS "Install Permissively"  LDAP option?
4) Documentation will need to be reevaluated and rewritten once we agree on the behaviour

afishbeck:
@RussWhitehead I'll approve with the idea that this fixes the break, but can you open JIRA(s) for implementing "least privileged access control"?

afishbeck (approved):
@RussWhitehead approved with the caveat that we will implement a more least privilege approach when we implement the new green field ldap.

RussWhitehead:
@afishbeck Opened https://track.hpccsystems.com/browse/HPCC-25795
@richardkchapman Please merge

PR #14825, opened 2021-04-07
HPCC-25698 Roxie does not appear to detect client socket closure

richardkchapman:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

## Testing:
Manually on docker desktop environment

### DIFF
--- a/system/jlib/jsocket.cpp
+++ b/system/jlib/jsocket.cpp
@@ -2125,7 +2125,15 @@ EintrRetry:
         rc = sendto(sock, NULL, 0, 0, &u.sa, ul);
     }
     else {
+#ifdef _WIN32
         rc = send(sock, NULL, 0, SEND_FLAGS);
+
+#else
+        char buffer;
+        rc = recv(sock, &buffer, sizeof(buffer), MSG_PEEK | MSG_DONTWAIT);
+        if (!rc)
+            return false;
+#endif
     }
     if (rc < 0) {
         int err=ERRNO();
@@ -2133,7 +2141,11 @@ EintrRetry:
             LOGERR2(err,7,"EINTR retrying");
             goto EintrRetry;
         }
+#ifdef _WIN32
         else
+#else
+        else if (err!=JSE_WOULDBLOCK)
+#endif
             return false;
     }   
     return true;
--- a/roxie/ccd/ccdcontext.cpp
+++ b/roxie/ccd/ccdcontext.cpp
@@ -2764,7 +2764,10 @@ public:
                 {
                     CriticalBlock b(abortLock);
                     if (!protocol->checkConnection())
+                    {
+                        DBGLOG("Client socket close detected");
                         throw MakeStringException(ROXIE_CLIENT_CLOSED, "Client socket closed");
+                    }
                     lastSocketCheckTime = ticksNow;
                 }
             }


system/jlib/jsocket.cpp
```diff
@@ -2125,15 +2125,18 @@ bool CSocket::check_connection()
         rc = sendto(sock, NULL, 0, 0, &u.sa, ul);
     }
     else {
-        rc = send(sock, NULL, 0, SEND_FLAGS);
+        char buffer[4];
+        rc = recv(sock, buffer, sizeof(buffer), MSG_PEEK | MSG_DONTWAIT);
```
ghalliday, line 2129: I don't think MSG_DONTWAIT is implemented in windows.

ghalliday: It might ork if you set non blocking mode, and then read, but what happens if something else is using the socket at the same time?  Do we protect against that?

richardkchapman: As I have no way to test what happens in Windows (and it doesn't really matter), I reverted the windows version of the code to how it was

system/jlib/jsocket.cpp
```diff
@@ -2125,15 +2125,27 @@ bool CSocket::check_connection()
         rc = sendto(sock, NULL, 0, 0, &u.sa, ul);
     }
     else {
+#ifdef _WIN32
         rc = send(sock, NULL, 0, SEND_FLAGS);
+
+#else
+        char buffer;
+        rc = recv(sock, &buffer, sizeof(buffer), MSG_PEEK | MSG_DONTWAIT);
+        if (!rc)
```
mckellyln, line 2134: if rc == 0 then I think you would want to return true as the socket is assumed still connected ?

system/jlib/jsocket.cpp
```diff
@@ -2125,15 +2125,27 @@ bool CSocket::check_connection()
...
     if (rc < 0) {
         int err=ERRNO();
         if ((err==JSE_INTR)&&(retrycount--!=0)) {
             LOGERR2(err,7,"EINTR retrying");
             goto EintrRetry;
         }
+#ifdef _WIN32
         else
+#else
+        else if (err!=JSE_WOULDBLOCK)
```
mckellyln, line 2147: Could it be JSE_WOULDBLOCK or JSE_AGAIN ?

system/jlib/jsocket.cpp
```diff
@@ -2125,15 +2125,27 @@ bool CSocket::check_connection()
         rc = sendto(sock, NULL, 0, 0, &u.sa, ul);
```
mckellyln, line 2125: I am doubtful/curious of a send() of 0 bytes actually doing anything ?

system/jlib/jsocket.cpp
```diff
@@ -2125,15 +2125,27 @@ bool CSocket::check_connection()
...
     }
     else {
+#ifdef _WIN32
         rc = send(sock, NULL, 0, SEND_FLAGS);
+
+#else
+        char buffer;
+        rc = recv(sock, &buffer, sizeof(buffer), MSG_PEEK | MSG_DONTWAIT);
```
mckellyln, line 2133: I might have done a poll() with a 0 timeout and if it returned ready for read then read() with MSG_PEEK but not MSG_DONTWAIT and check for <= 0.
But not sure if that would be any different than here, except for not having to test for EWOULDBLOCK/EAGAIN.

mckellyln:
I haven't tested it, but this may work most times for detecting the other side closed, without having to send -

mckellyln:
I have not tested it and I'm not sure it would be any more robust/reliable, but what we have used before to try and determine if peer closed connection is something like -
        struct pollfd fds[1];
        fds[0].fd = sock;
        fds[0].events = POLLINX; // POLLIN | POLLRDHUP
        fds[0].revents = 0;
        rc = ::poll(fds, 1, 0);
        if (rc == 0)
            return true;
        else if (rc < 0)
        {
            int err=ERRNO();
            if ((err==JSE_INTR)&&(retrycount--!=0))
            {
                LOGERR2(err,7,"EINTR retrying");
                goto EintrRetry;
            }
            else
                return false;
        }
        else // ready for something
        {
            if ( (fds[0].revents & (POLLERR | POLLNVAL)) || (!(fds[0].revents & (POLLINX | POLLHUP))) )
                return false;
        }
        // ready for read
        retrycount=100;
        char buffer;
        rc = recv(sock, &buffer, sizeof(buffer), MSG_PEEK);
        if (rc == 0)
            return false;
        else if (rc < 0)
        {
            int err=ERRNO();
            if ((err==JSE_INTR)&&(retrycount--!=0))
            {
                LOGERR2(err,7,"EINTR retrying");
                goto EintrRetry;
            }
            else
                return false;
        }
        return true;

mckellyln (commented):
Just wondering about the send of 0 bytes but if its always worked before than I assume its fine.
And do we need to care about EAGAIN on the recv() ?
I added some sample code for my idea of checking for peer closed without sending.

system/jlib/jsocket.cpp
richardkchapman: It doesn't appear to - but that's how the code used to "work".

richardkchapman:
If we have existing code that successfully tests for this condition, then it should probably be commoned up with this function and used in place of my new code.

I don't know if the send of 0 bytes has "always worked" - my guess is it worked once but no-one noticed that it stopped working on newer systems. I have no idea if it works on Windows - I left the code that way simply because the new code does not compile on Windows, and it really doesn't matter if this code works on Windows or not.

mckellyln:
@richardkchapman should I put another PR together and test it ?

richardkchapman:
@mckellyln Sure!

mckellyln:
Closing to issue my PR for this issue.

mckellyln:
New PR is: https://github.com/hpcc-systems/HPCC-Platform/pull/14848

PR #14829, opened 2021-04-07
HPCC-25522 Set sizes and compressed sizes in WsDfu.DFUFilePublish

wangkx:
The existing WsDfu.DFUFilePublish does not set the part sizes,
compressed  part sizes, and compressed file size. The issues are
fixed in this PR.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/services/ws_dfu/ws_dfuService.cpp
+++ b/esp/services/ws_dfu/ws_dfuService.cpp
@@ -6553,6 +6553,7 @@ bool CWsDfuEx::onDFUFilePublish(IEspContext &context, IEspDFUFilePublishRequest 
             df->detach(req.getLockTimeoutMs());
         }
 
+        setPublishFileSize(newFileName, fileDesc);
         newFile.setown(queryDistributedFileDirectory().createNew(fileDesc));
         newFile->validate();
         newFile->setAccessed();
@@ -6590,5 +6591,47 @@ bool CWsDfuEx::onDFUFilePublish(IEspContext &context, IEspDFUFilePublishRequest 
     return true;
 }
 
+void CWsDfuEx::setPublishFileSize(const char *lfn, IFileDescriptor *fileDesc)
+{
+    auto funcFilePartSize = [lfn, fileDesc](unsigned partNum)
+    {
+        bool compressed = fileDesc->isCompressed();
+        IPartDescriptor *partDesc = fileDesc->queryPart(partNum);
+        IPropertyTree &props = partDesc->queryProperties();
+        unsigned numCopies = partDesc->numCopies();
+        for (unsigned c = 0; c < numCopies; c++)
+        {
+            RemoteFilename rfn;
+            partDesc->getFilename(c, rfn);
+            try
+            {
+                Owned<IFile> file = createIFile(rfn);
+                if (compressed)
+                {
+                    props.setPropInt64("@compressedSize", file->size());
+                    Owned<IFileIO> io = createCompressedFileReader(file);
+                    if (io)
+                        props.setPropInt64("@size", io->size());
+                }
+                else
+                {
+                    props.setPropInt64("@size", file->size());
+                }
+                break;
+            }
+            catch (IException *e)
+            {
+                VStringBuffer tmp("%s: ", lfn);
+                rfn.getPath(tmp);
+                EXCLOG(e, tmp.str());
+                e->Release();
+            }
+        }
+    };
+
+    //collect and set the sizes for file parts
+    CAsyncForFunc<decltype(funcFilePartSize)> async(funcFilePartSize);
+    async.For(fileDesc->numParts(), 100);
+}
 
 //////////////////////HPCC Browser//////////////////////////
--- a/esp/services/ws_dfu/ws_dfuService.hpp
+++ b/esp/services/ws_dfu/ws_dfuService.hpp
@@ -147,6 +147,7 @@ class CWsDfuEx : public CWsDfu
     static const unsigned defaultMaxFileAccessExpirySeconds=86400; // 24 hours
 
     void dFUFileAccessCommon(IEspContext &context, const CDfsLogicalFileName &lfn, SessionId clientSessionId, const char *requestId, unsigned expirySecs, bool returnTextResponse, unsigned lockTimeoutMs, IEspDFUFileAccessResponse &resp);
+    void setPublishFileSize(const char *logicalName, IFileDescriptor *fileDesc);
     void clearFileProtections(IDistributedFile *df);
     bool changeFileProtections(IEspContext &context, IEspDFUArrayActionRequest &req, IEspDFUArrayActionResponse &resp);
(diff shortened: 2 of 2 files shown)


esp/services/ws_dfu/ws_dfuService.cpp
```diff
@@ -6590,5 +6591,104 @@ bool CWsDfuEx::onDFUFilePublish(IEspContext &context, IEspDFUFilePublishRequest
...
+        {
+            props.setPropInt64("@compressedSize", compressedSize);
+            fileCompressedSize += compressedSize;
+        }
+    }
+
+    //Set file compressed size
+    if (compressed && (fileCompressedSize != (offset_t)-1))
+    {
+        fileDesc->queryProperties().setPropInt64("@compressedSize", fileCompressedSize);
```
jakesmith, line 6690: I think (as with top-level @size), @compressedSize should be auto-computed/populated, as long as parts have this property set. If I'm right, this code is not necessary.
If I'm wrong, I can't see the code that sets it in Thor for example, so may imply a bug.
Can you check?

esp/services/ws_dfu/ws_dfuService.cpp
```diff
@@ -6590,5 +6591,104 @@ bool CWsDfuEx::onDFUFilePublish(IEspContext &context, IEspDFUFilePublishRequest
...
+                {
+                    VStringBuffer tmp("%s: ", lfn.get());
+                    rfn.getPath(tmp);
+                    EXCLOG(e, tmp.str());
+                    e->Release();
+
+                    size = (offset_t)-1;
+                    compressedSize = (offset_t)-1;
+                }
+                if ((size != (offset_t)-1) && (!compressed || (compressedSize != (offset_t)-1)))
```
jakesmith, line 6639: if compressed (and success) compressedSize would be set, would be clearer if just:
```
if ((size != (offset_t)-1) || compressedSize != (offset_t)-1))
```

esp/services/ws_dfu/ws_dfuService.cpp
```diff
@@ -6590,5 +6591,104 @@ bool CWsDfuEx::onDFUFilePublish(IEspContext &context, IEspDFUFilePublishRequest
...
+        offset_t getCompressedSize() { return compressedSize; }
+    };
+
+    //Get the sizes and the compressed sizes for each file parts in parallel
+    CIArrayOf<CFilePartSizeThread> threads;
+    bool compressed = fileDesc->isCompressed();
+    Owned<IPartDescriptorIterator> pi = fileDesc->getIterator();
+    ForEach(*pi)
+    {
+        CFilePartSizeThread* thread = new CFilePartSizeThread(lfn, pi->query(), compressed);
```
jakesmith, line 6655: this will be running an unlimited number of threads, in this context that's ok generally,
but this is a good case for using CAsyncFor (or CAsyncForFunc), e.g. with max 100
and it would significantly reduce code complexity too.

esp/services/ws_dfu/ws_dfuService.cpp
```diff
@@ -6590,5 +6591,104 @@ bool CWsDfuEx::onDFUFilePublish(IEspContext &context, IEspDFUFilePublishRequest
...
+        threads.item(i).join();
+
+    //Set the size and the compressed size for each file parts and calculate file compressed size.
+    offset_t fileCompressedSize = (offset_t)0;
+    ForEachItemIn(ii, threads)
+    {
+        CFilePartSizeThread& thread = threads.item(ii);
+        offset_t size = thread.getSize();
+        offset_t compressedSize = thread.getCompressedSize();
+        if ((size == (offset_t)-1) || (compressed && (compressedSize == (offset_t)-1)))
```
jakesmith, line 6669: clearer if : 
```
if ((size == (offset_t)-1) && (compressedSize == (offset_t)-1))
```

esp/services/ws_dfu/ws_dfuService.cpp
```diff
@@ -6590,5 +6591,104 @@ bool CWsDfuEx::onDFUFilePublish(IEspContext &context, IEspDFUFilePublishRequest
...
+            break;
+        }
+
+        IPartDescriptor *partDesc = fileDesc->queryPart(ii);
+        IPropertyTree &props = partDesc->queryProperties();
+        props.setPropInt64("@size", size);
+        if (compressed)
+        {
+            props.setPropInt64("@compressedSize", compressedSize);
+            fileCompressedSize += compressedSize;
```
jakesmith, line 6683: see other comment below, I _think_ top-level @compressedSize should be automatically populated (if so, no need to accumulate here), and if it is not, there may be a bug elsewhere.

jakesmith (changes requested):
@wangkx - looks fine in principle, but should look at limiting threads and reducing complexity, by doing everything in a function called by CAsyncForFunc, e.g.:
```
auto func = [lfn, fileDesc](unsigned partNum) =
{
 // collect and set
};

CASyncForFunc<decltype(func)> async(func);
async.For(fileDesc->numParts(), 100);
```

esp/services/ws_dfu/ws_dfuService.cpp
wangkx: I will use CAsyncFor.

esp/services/ws_dfu/ws_dfuService.cpp
wangkx: Looks like the auto-computed/populated does not work here. If I do not set the compressedSize here, the Attr/@compressedSize is not set for the file. But, when I create a compressed file using ECL's OUTPUT option ENCRYPT, the Attr/@compressedSize is set.

wangkx: Inside the CDistributedFile constructor, there is some code to calculate/set the @size. But, there is no code for the @compressedSize. Is that a bug?

jakesmith: Yes it is really, or an opportunity to improve and be more consistent. The engines hit the issue, by manually setting compressedSize (and sometimes size) but not consistently.
I've released a fix : https://track.hpccsystems.com/browse/HPCC-25708

esp/services/ws_dfu/ws_dfuService.cpp
jakesmith: will need https://github.com/hpcc-systems/HPCC-Platform/pull/14836, but this code can be removed.

jakesmith (commented):
@wangkx looks good.

Should remove the code that sets top-level @compressedSize which will be handled by https://github.com/hpcc-systems/HPCC-Platform/pull/14836 when it's merged.

wangkx:
@jakesmith the code for the top-level @compressedSize is removed.

jakesmith (approved):

wangkx:

PR #14838, opened 2021-04-09
HPCC-2533 Add support for the ecl scheduler to cloud deployments

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/ecl/eclccserver/eclccserver.cpp
+++ b/ecl/eclccserver/eclccserver.cpp
@@ -805,6 +805,8 @@ void openLogFile()
     envGetConfigurationDirectory("log","eclccserver",globals->queryProp("@name"),logname);
     Owned<IComponentLogFileCreator> lf = createComponentLogFileCreator(logname.str(), "eclccserver");
     lf->beginLogging();
+#else
+    setupContainerizedLogMsgHandler();
 #endif
 }
 
--- a/ecl/eclscheduler/eclscheduler.cpp
+++ b/ecl/eclscheduler/eclscheduler.cpp
@@ -164,14 +164,27 @@ private:
 
 void openLogFile()
 {
+#ifndef _CONTAINERIZED
     Owned<IComponentLogFileCreator> lf = createComponentLogFileCreator(globals, "eclscheduler");
     lf->beginLogging();
+#else
+    setupContainerizedLogMsgHandler();
+#endif
 }
 
 //=========================================================================================
 
+static constexpr const char * defaultYaml = R"!!(
+version: "1.0"
+eclscheduler:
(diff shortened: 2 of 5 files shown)


ghalliday:
Tested by running schedule1.ecl inside the playground, targeting hthor/roxie-workunit/thor.  All passed.

ecl/eclccserver/eclccserver.cpp
```diff
@@ -805,6 +805,8 @@ void openLogFile()
     envGetConfigurationDirectory("log","eclccserver",globals->queryProp("@name"),logname);
     Owned<IComponentLogFileCreator> lf = createComponentLogFileCreator(logname.str(), "eclccserver");
     lf->beginLogging();
+#else
```
ghalliday, line 808: Spotted while making changes to scheduler.  I think this should be included here.

richardkchapman (approved):

ecl/eclscheduler/eclscheduler.cpp
```diff
@@ -223,10 +244,22 @@ int main(int argc, const char *argv[])
...
         CIArrayOf<EclScheduler> schedulers;
+
+#ifdef _CONTAINERIZED
+        Owned<IPTreeIterator> queues = globals->getElements("queues");
+        ForEach(*queues)
+        {
+            IPTree &queue = queues->query();
+            const char *qname = queue.queryProp("@name");
+            DBGLOG("Start listening to queue %s", qname);
+            Owned<EclScheduler> scheduler = new EclScheduler(qname);
```
jakesmith, line 257: This means every listed scheduler, listens to every queue.. 
Should there be an option, like there is for eclccserver's to specify 'listen' set of queues?

jakesmith (approved):
@ghalliday - looks good, 1 minor question.

ecl/eclscheduler/eclscheduler.cpp
ghalliday: That was the intention.  I don't think we should have more than one scheduler defined.

I think eclccserver was implemented that way because compiles for different targets can have vastly different memory requirements.  eclscheduler should use very little memory, and causes problems if there is more than one running at once.

jakesmith: >and causes problems if there is more than one running at once.

should we prevent replicas from being configured to >1 for that reason?

PR #14848, opened 2021-04-13
HPCC-25698 Roxie detect client socket close

mckellyln:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

## Testing:

I tested by stopping testsocket while query was running and checking Roxie log

### DIFF
--- a/system/jlib/jsocket.cpp
+++ b/system/jlib/jsocket.cpp
@@ -1628,6 +1628,7 @@ int CSocket::logPollError(unsigned revents, const char *rwstr)
 int CSocket::wait_read(unsigned timeout)
 {
     int ret = 0;
+    int retrycount = 100;
     while (sock!=INVALID_SOCKET)
     {
 #ifdef _USE_SELECT
@@ -1661,9 +1662,14 @@ int CSocket::wait_read(unsigned timeout)
                 LOGERR2(err,1,"wait_read");
                 break;
             }
+            else
+            {
+                if (retrycount-- <= 0)
+                    break;
+            }
         }
         else if (ret == 0)
-        {   // timeout
+        {   // timeout, no error (timeout can be 0 for a fast check)
             break;
         }
         else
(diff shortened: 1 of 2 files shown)


mckellyln:
Updated to be a little cleaner and should compile on Windows

mckellyln:
After an amend it seems the branch used for smoke test does not match commit and we see errors such as -
```
/home/runner/work/HPCC-Platform/HPCC-Platform/src/system/jlib/jsocket.cpp:2125:1: error: version control conflict marker in file
 2125 | <<<<<<< 3f259053637f93a7d71a64a1a9cb2b1b1f4611c4
```

mckellyln:
Should I close this PR and re-issue it ?

richardkchapman (approved):

system/jlib/jsocket.cpp
```diff
@@ -2116,26 +2122,32 @@ bool CSocket::check_connection()
...
+        return false;
+
+    int retrycount = 100;
+    char buffer;
+
+EintrRecv:
+    rc = recv(sock, &buffer, sizeof(buffer), MSG_PEEK);
+
+    if (rc == 0)
+        return false;
```
jakesmith, line 2138: not sure I understand this, can't the connection be good, but 0 bytes on the wire?
from recv man:
```
       When a stream socket peer has performed an orderly shutdown, the
       return value will be 0 (the traditional "end-of-file" return).

       Datagram sockets in various domains (e.g., the UNIX and Internet
       domains) permit zero-length datagrams.  When such a datagram is
       received, the return value is 0.

       The value 0 may also be returned if the requested number of bytes
       to receive from a stream socket was 0.
```

jakesmith (commented):
@mckellyln - 1 question

jakesmith:
> Should I close this PR and re-issue it ?

no, it's the fault of a regression in the smoketest action, a (2nd) fix is pending: https://github.com/hpcc-systems/HPCC-Platform/pull/14851

system/jlib/jsocket.cpp
richardkchapman: For a UDP socket it could be triggered by someone sending a 0-byte datagram, but not sure this code is supposed to handle this. For TCP it should be ok.

system/jlib/jsocket.cpp
```diff
@@ -2116,26 +2122,32 @@ bool CSocket::check_connection()
...
-        socklen_t  ul = setSockAddr(u,returnep,returnep.port);
-        rc = sendto(sock, NULL, 0, 0, &u.sa, ul);
-    }
-    else {
-        rc = send(sock, NULL, 0, SEND_FLAGS);
-    }
-    if (rc < 0) {
+    int rc = wait_read(0);
+    if (rc == 0)
+        return true;
```
jakesmith, line 2127: am a little confused by this condition.
wait_read returns 0 on timeout:
```
        else if (ret == 0)
        {   // timeout
            break;
        }
...
   return ret;
```
that's via poll():
```
       return value of zero indicates that the system call timed out
       before any file descriptors became read.
```

Or is it that when poll is called with timeout=0, and returns 0, it means 0 file descriptors ready not timeout?

If it's correct to treat wait_read call here as not a timeout, then could benefit from a comment describing reasons/semantics.

system/jlib/jsocket.cpp
jakesmith: Is it ok, because wait_read(0) would have return 0, which usually means timeout, but because poll timeout is 0, means 0 file descriptors ready/0 errors, so has returned true before got here...
So only here because there should have been some date to read, and if not = error   ?
Could benefit from a comment describing semantics.

I guess this is knowingly not thread safe? i.e. if there was data on the socket and got to the recv, but someone else read it before this recv did, then socket is good, but this would return false.
Probably ok, assumes not multiple threads reading/calling check_connection concurrently ?

system/jlib/jsocket.cpp
ghalliday: I agree it could do with some comments - I'm struggling to follow the logic.

system/jlib/jsocket.cpp
mckellyln: Doesn't matter what the poll timeout is - wait_read() returning 0 means socket is not closed at that moment.
If wait_read() returns > 0 then we recv() knowning it wont block - and if that returns 0 then we know socket was closed.

mckellyln: I'll add a comment

mckellyln:
@jakesmith I've pushed a few comments.  thx.

system/jlib/jsocket.cpp
jakesmith: makes sense now, thanks.

mckellyln:
And squashed

jakesmith (approved):
Looks good.

ghalliday (approved):

PR #14856, opened 2021-04-15
HPCC-25734 Remove decode() for normal string in Python3

xwang2713:
This will prevent an unexpected exception

Sign-off by Xiaoming Wang <xiaoming.wang@lexisnexis.com>

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/initfiles/sbin/hpcc/cluster/task.py
+++ b/initfiles/sbin/hpcc/cluster/task.py
@@ -131,7 +131,7 @@ class ScriptTask(Task):
         protect the script from malicious modification.
         '''
         try:
-            self._ip = host.ip.decode('utf-8')
+            self._ip = host.ip
         except AttributeError as e:
             self._ip = host.ip
         cmd = self.script_file + " " + self._ip


initfiles/sbin/hpcc/cluster/task.py
```diff
@@ -131,7 +131,7 @@ def worker(self, host):
         protect the script from malicious modification.
         '''
         try:
-            self._ip = host.ip.decode('utf-8')
+            self._ip = host.ip
```
Michael-Gardner, line 134: @xwang2713 I believe this issue is already properly fixed in 8.0.x.  If we hit an AttributeError here from the decode('utf-8') being present on a string object (which there isn't a guarantee that host.ip actually is a string object because of how python3 does dynamic casting), then we'll fall through the try block and assign host.ip straight to self._ip like you  have on line 134, and hpcc-run.sh will work fine.  I've tested this out on the 160 last week.  Perhaps whoever raised this bug was not using the latest 8.0.x build?

Michael-Gardner (commented):
@xwang2713 I think this Jira is a duplicate of one I fixed last week https://track.hpccsystems.com/browse/HPCC-25693.

We should be able to check on the 160 and ensure that this is no longer an issue by running hpcc-push.sh or any other script that uses the python libraries and we shouldn't get any unhandled AtrributeErrors.

initfiles/sbin/hpcc/cluster/task.py
xwang2713: Do you know which 8.0.x?  I built AMI with 8.0.2-1 but hpcc-run.sh throws an exception for host.ip.decode()

xwang2713: Previously we use Python2.7 which doesn't have this problem. Probably Python2.7 ignore the case.

Michael-Gardner: Earliest branch that I see the commit in is 8.0.4 (4b28584532aac1403ebbdc9e7e94275712faced3)

Python2 didn't have this issue because decode/encode were always available to string objects. They weren't unicode by default, so there was never an attributeError to throw.

xwang2713: OK. Ignore this PR

AttilaVamos:
@xwang2713 I am sure the workflow_contingency_8 (parallel=true) failure is not related to this changes.

PR #14862, opened 2021-04-16
HPCC-25738 Add metrics for ESP to helm chart

kenrowland:
Added metrics config for a logging sink to ESP helm template

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [ ] Thread-safety
  - [x] Cloud-compatibility
  - [x] Premature optimization

## Smoketest:

### DIFF
--- a/esp/platform/espp.cpp
+++ b/esp/platform/espp.cpp
@@ -360,8 +360,14 @@ void initializeMetrics(CEspConfig* config)
     //
     // Initialize metrics
     Owned<IPropertyTree> pMetricsTree = config->queryConfigPTree()->getPropTree("metrics");
+    if (pMetricsTree == nullptr)
+    {
+        pMetricsTree.setown(queryComponentConfig().getPropTree("metrics"));
+    }
+
     if (pMetricsTree != nullptr)
     {
+        PROGLOG("Metrics initializing...");
         MetricsReporter &metricsReporter = queryMetricsReporter();
         metricsReporter.init(pMetricsTree);
         metricsReporter.startCollecting();
--- a/helm/hpcc/templates/esp.yaml
+++ b/helm/hpcc/templates/esp.yaml
@@ -10,8 +10,9 @@ data:
   {{ .me.name }}.yaml: |
     version: 1.0
     esp:
-{{ toYaml (omit .me "logging") | indent 6 }}
+{{ toYaml (omit .me "logging" "metrics") | indent 6 }}
 {{- include "hpcc.generateLoggingConfig" . | indent 6 }}
+{{- include "hpcc.generateMetricsConfig" . | indent 6 }}
 {{- if and .root.Values.certificates .root.Values.certificates.enabled }}
  {{- if not (hasKey .me "tls" )}}
       tls: true
--- a/helm/hpcc/values.schema.json
+++ b/helm/hpcc/values.schema.json
@@ -277,6 +277,15 @@
               "type": "null"
             }
           ]
+        },
+        "metrics": {
+          "type" : "object",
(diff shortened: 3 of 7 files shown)


helm/hpcc/values.schema.json
```diff
@@ -1196,6 +1205,34 @@
...
+    },
+    "sink": {
+      "type" : "object",
+      "description" : "Definition of a metric sink",
+      "properties" : {
+        "type": {
+          "type" : "string",
+          "description" : "The defined metric sink type"
+        },
+        "name" : {
```
ghalliday, line 1223: I'm curious what the name is for.  It is required, but doesn't seem to serve any purpose as far as the user is concerned.

esp/platform/espp.cpp
```diff
@@ -360,8 +360,14 @@ void initializeMetrics(CEspConfig* config)
     //
     // Initialize metrics
     Owned<IPropertyTree> pMetricsTree = config->queryConfigPTree()->getPropTree("metrics");
```
ghalliday, line 362: This is fine as it stands, but I would prefer it coded so that the component config is the default.  The reason is to make it consistent when the code is added to other components (and make it easier to common up).

```
    Owned<IPropertyTree> pMetricsTree = queryComponentConfig().getPropTree("metrics");
    if (pMetricsTree == nullptr)
    {
        pMetricsTree.setown(config->queryConfigPTree()->getPropTree("metrics"));
        queryComponentConfig().setPropTree("metrics", LINK(pMetricsTree));
    }

```

esp/platform/espp.cpp
```diff
@@ -447,11 +453,13 @@ int init_main(int argc, const char* argv[])
         Owned<IPropertyTree> envpt;
         if (application)
         {
+            DBGLOG("Running as application");
```
ghalliday, line 456: not sure if these should be in the final version, depends if @afishbeck thinks they are useful.

ghalliday (commented):
Looks good.  A few minor comments.

esp/platform/espp.cpp
ghalliday: Actually, don't bother with this change... leave it as you have it.

esp/platform/espp.cpp
afishbeck: These could be useful, but I would add the name of the application to clarify which config was being loaded... and below I would say "Running from legacy config" and add the cfgfile file location rather than "Running not as application".

helm/hpcc/templates/esp.yaml
```diff
@@ -12,6 +12,7 @@ data:
     esp:
 {{ toYaml (omit .me "logging") | indent 6 }}
```
afishbeck, line 13: Should metrics be omitted here?

afishbeck (changes requested):
One possibly important and one minor comment.

kenrowland (commented):
@afishbeck Please take a look at the changes from your comments

helm/hpcc/values.schema.json
ghalliday: I'm not convinced it should be required (or even be present).  The helm charts are using lists for the metrics so there isn't a way to override the global metrics, except for completely replacing them.
I think if the framework needs to differentiate the sinks then it should be done internally.
Not significant enough not to merge this PR though, but should be revisited.

ghalliday (approved):
@kenrowland.  Approved, but names on sinks should be revisited.  Please squash.

kenrowland:

afishbeck (approved):
Approved.

PR #14872, opened 2021-04-20
HPCC-25747 DFUInfo returns logical file at rest cost

shamser:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dali/base/dadfs.cpp
+++ b/dali/base/dadfs.cpp
@@ -45,6 +45,7 @@
 #include <vector>
 #include <unordered_map>
 #include <algorithm>
+#include <time.h>
 
 #ifdef _DEBUG
 //#define EXTRA_LOGGING
@@ -174,6 +175,15 @@ const char *normalizeLFN(const char *s,StringBuffer &tmp)
 static IPropertyTree *getEmptyAttr()
 {
     return createPTree("Attr");
+}
+
+static double calcFileCost(const char * cluster, double sizeGB, double fileAgeDays)
+{
+    IPropertyTree * plane = queryStoragePlane(cluster);
+    if (!plane)
+        return 0.0;
+    double storageCostDaily = plane->getPropReal("cost/@storageAtRest", 0.0) * 12 / 365;
+    return storageCostDaily * sizeGB * fileAgeDays;
 }
 
 RemoteFilename &constructPartFilename(IGroup *grp,unsigned partno,unsigned partmax,const char *name,const char *partmask,const char *partdir,unsigned copy,ClusterPartDiskMapSpec &mspec,RemoteFilename &rfn)
@@ -4705,6 +4715,27 @@ public:
         else
             return false;
     }
+    virtual double getCost(const char * cluster) override
+    {
+        CDateTime dt;
+        getModificationTime(dt);
+        double fileAgeDays = difftime(time(nullptr), dt.getSimple())/(24*60*60);
+        double sizeGB = getDiskSize(true, false) / ((double)1024 * 1024 * 1024);
+
+        if (isEmptyString(cluster))
+        {
+            StringArray clusterNames;
+            unsigned countClusters = getClusterNames(clusterNames);
+            double totalCost = 0.0;
+            for (unsigned i = 0; i < countClusters; i++)
+                totalCost += calcFileCost(clusterNames[i], sizeGB, fileAgeDays);
+            return totalCost;
+        }
+        else
+        {
+            return calcFileCost(cluster, sizeGB, fileAgeDays);
+        }
+    }
 };
 
 static unsigned findSubFileOrd(const char *name)
@@ -6653,6 +6684,18 @@ public:
     virtual bool getSkewInfo(unsigned &maxSkew, unsigned &minSkew, unsigned &maxSkewPart, unsigned &minSkewPart, bool calculateIfMissing) override
     {
         return false;
+    }
+
+    virtual double getCost(const char * cluster) override
+    {
+        double totalCost = 0.0;
+        CriticalBlock block (sect);
+        ForEachItemIn(i,subfiles)
+        {
+            IDistributedFile &f = subfiles.item(i);
+            totalCost += f.getCost(cluster);
+        }
+        return totalCost;
     }
 };
 
--- a/esp/services/ws_dfu/ws_dfuService.cpp
+++ b/esp/services/ws_dfu/ws_dfuService.cpp
@@ -45,6 +45,7 @@
 #include "fverror.hpp"
 #include "nbcd.hpp"
 #include "thorcommon.hpp"
+#include "jstats.h"
 
 #include "jstring.hpp"
 #include "exception_util.hpp"
@@ -2643,6 +2644,13 @@ void CWsDfuEx::doGetFileDetails(IEspContext &context, IUserDescriptor *udesc, co
             if (includeJsonTypeInfo)
                 FileDetails.setJsonInfo(jsonLayout);
         }
+    }
+    if (version >= 1.59)
+    {
+        double totalCost = df->getCost(cluster);
+        StringBuffer s;
+        formatMoney(s, money2cost_type(totalCost));
+        FileDetails.setCost(s);
     }
     PROGLOG("doGetFileDetails: %s done", name);
 }
--- a/helm/hpcc/values.schema.json
+++ b/helm/hpcc/values.schema.json
@@ -424,9 +424,20 @@
           "type" : "string"
         },
         "defaultSprayParts" : {
-            "description": "Number of parts sprayed by default",
-            "type" : "integer"
+          "description": "Number of parts sprayed by default",
+          "type" : "integer"
+        },
+        "cost" : {
+          "description": "Costs associated with the storage and use of the plane",
+          "type" : "object",
+          "properties": {
+            "storageAtRest" : {
+              "description": "Storage cost (GiB/month)",
+              "type": "number"
+            }
           }
+        }
+
       },
       "required": [ "name", "prefix" ],
       "additionalProperties": false
--- a/system/jlib/jstats.h
+++ b/system/jlib/jstats.h
@@ -824,5 +824,6 @@ public:
     }
 };
 
+extern jlib_decl StringBuffer & formatMoney(StringBuffer &out, unsigned __int64 value);
 
 #endif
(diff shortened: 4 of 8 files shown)


dali/base/dadfs.cpp
```diff
@@ -4704,6 +4705,20 @@ protected: friend class CDistributedFilePart;
         else
             return false;
     }
+    virtual double getCost(const char * cluster) override
+    {
+        VStringBuffer path("storage/planes[@name='%s']", cluster);
+        Owned<IPropertyTree> plane = queryGlobalConfig().getPropTree(path);
+        if (!plane)
+            return 0.0;
+        double storageCostDaily = plane->getPropReal("cost/@storageAtRest", 0.0) * 12 / 365;
```
jakesmith, line 4714: so /storageAtRest' is supposed to represent a months worth of cost?
Why has this unit been chosen? (because Azure uses it?).

Wouldn't per day (or even per hour) be a better flexible unit for us ?
Then we'd scale whatever units the provider chose to use, into our standard.

dali/base/dadfs.cpp
```diff
@@ -4704,6 +4705,20 @@ protected: friend class CDistributedFilePart;
...
+
+        CDateTime dt;
+        getModificationTime(dt);
+        double fileAgeDays = difftime(time(nullptr), dt.getSimple())/(24*60*60);
+        double sizeGB = getDiskSize(true, true) / ((double)1024 * 1024 * 1024);
```
jakesmith, line 4719: think should be getDiskSize(true, false),  i.e. should use meta info without opening file parts if there.

jakesmith: GB is that granular enough (probably), should be use MB, incase some exotic storage comes along that is super-expensive, but we can't tell if using .5GB of it?

esp/services/ws_dfu/ws_dfuService.cpp
```diff
@@ -2644,6 +2645,13 @@ void CWsDfuEx::doGetFileDetails(IEspContext &context, IUserDescriptor *udesc, co
                 FileDetails.setJsonInfo(jsonLayout);
         }
     }
+    if (version > 1.58)
```
jakesmith, line 2648: trivial: is >= 1.59 (i.e. mentioning the version feature came in on) clearer?

helm/hpcc/values.schema.json
```diff
@@ -421,7 +421,18 @@
         "defaultSprayParts" : {
             "description": "Number of parts sprayed by default",
             "type" : "integer"
+        },
+        "cost" : {
+            "description": "Costs associated with the storage and use of the plane",
+            "type" : "object",
+            "properties": {
+                "storageAtRest" : {
+                    "description": "Storage cost (GiB/month)",
```
jakesmith, line 430: is per month sensible granularity for us, e.g. why not per day?

jakesmith (commented):
generally looks good, a few minor comments/questions.

dali/base/dadfs.cpp
```diff
@@ -4704,6 +4705,20 @@ protected: friend class CDistributedFilePart;
         else
             return false;
     }
+    virtual double getCost(const char * cluster) override
+    {
+        VStringBuffer path("storage/planes[@name='%s']", cluster);
```
ghalliday, line 4710: What happens if cluster is null/empty?  Should it iterate through all the clusters the file lives on?  (I think cluster can be null on the call to dfuInfo).

ghalliday (commented):
A few minor comments.

helm/hpcc/values.schema.json
shamser: The pricing tables provided by the cloud providers are on a monthly basis.  I thought it would make it easier to setup and easier to identify mistakes if the time period was the same.

dali/base/dadfs.cpp
```diff
@@ -4704,6 +4705,20 @@ protected: friend class CDistributedFilePart;
...
+        Owned<IPropertyTree> plane = queryGlobalConfig().getPropTree(path);
+        if (!plane)
+            return 0.0;
+        double storageCostDaily = plane->getPropReal("cost/@storageAtRest", 0.0) * 12 / 365;
+
+        CDateTime dt;
+        getModificationTime(dt);
+        double fileAgeDays = difftime(time(nullptr), dt.getSimple())/(24*60*60);
```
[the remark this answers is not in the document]
shamser: I can make this function more efficient by passing in the cost rate and the current time.  However, this function is not called very frequently.  This can be improved in the later versions.

helm/hpcc/values.schema.json
jakesmith: that's fine - I didn't realize all cloud providers used monthly basis.

helm/hpcc/values.schema.json
```diff
@@ -426,7 +426,19 @@
         "defaultSprayParts" : {
             "description": "Number of parts sprayed by default",
             "type" : "integer"
+        },
+        "cost" : {
+            "description": "Costs associated with the storage and use of the plane",
+            "type" : "object",
+            "default": "0.1193"
```
jakesmith, line 433: missing comma, I think this will fail cause helm install to fail.

Also, this is default for this "object" type, I think this line is supposed to be under "storageAtRest" ?

helm/hpcc/values.schema.json
```diff
@@ -426,7 +426,19 @@
         "defaultSprayParts" : {
             "description": "Number of parts sprayed by default",
             "type" : "integer"
+        },
+        "cost" : {
+            "description": "Costs associated with the storage and use of the plane",
```
jakesmith, line 431: trivial/formatting: inconsistent indentation, 2 space indenting used normally in this file - only other exception I think is 'defautSprayParts'

dali/base/dadfs.cpp
jakesmith: > ghalliday: What happens if cluster is null/empty? Should it iterate through all the clusters the file lives on? (I think cluster can be null on the call to dfuInfo).

I think this is still an issue.
DFUInfo will normally come with a particular cluster specified, but the file is still on 1 or more clusters.

As Gavin says, if cluster is an null/empty, I think needs to iterate through the clusters the file is on, and total up the cost.

jakesmith (changes requested):
@shamser - a couple of issues

dali/base/dadfs.cpp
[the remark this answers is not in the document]
ghalliday: yes, I would ignore for now

helm/hpcc/values.schema.json
shamser: sorry, that was careless.

dali/base/dadfs.cpp
jakesmith: @shamser - can you comment on this one?

ghalliday (approved):

richardkchapman:
Windows compile failing:

ws_dfuService.obj : error LNK2019: unresolved external symbol "class StringBuffer & __cdecl formatMoney(class StringBuffer &,unsigned __int64)" (?formatMoney@@YAAEAVStringBuffer@@AEAV1@_K@Z) referenced in function "private: void __cdecl CWsDfuEx::doGetFileDetails(struct IEspContext &,struct IUserDescriptor *,char const *,char const *,char const *,char const *,char const *,bool,bool,enum CDFUChangeProtection,enum CDFUChangeRestriction,struct IEspDFUFileDetail &)" (?doGetFileDetails@CWsDfuEx@@AEAAXAEAUIEspContext@@PEAUIUserDescriptor@@PEBD2222_N3W4CDFUChangeProtection@@W4CDFUChangeRestriction@@AEAUIEspDFUFileDetail@@@Z) [D:\a\HPCC-Platform\HPCC-Platform\build\esp\services\ws_dfu\ws_dfu.vcxproj]
2074

system/jlib/jstats.h
```diff
@@ -824,5 +824,6 @@ class jlib_decl NullStatisticTarget : implements IStatisticTarget
     }
 };
 
+extern StringBuffer & formatMoney(StringBuffer &out, unsigned __int64 value);
```
richardkchapman, line 827: Needs jlib_decl

richardkchapman (changes requested):

jakesmith (changes requested):
@shamser tagging you (github review tag) re. : https://github.com/hpcc-systems/HPCC-Platform/pull/14872#discussion_r618303706

helm/hpcc/values.schema.json
shamser: @ghalliday (re default values) I haven't been able to provide a default cost value in the values.yaml because planes are not defined in the default values.yaml.  I've used the "default" attribute in the schema but that was ignored.

dali/base/dadfs.cpp
shamser: @jakesmith I've implemented this (re iterating through all clusters) in the latest commit.

helm/hpcc/values.schema.json
shamser: I have removed the "default" attribute because it was being ignored.

shamser:
> @shamser tagging you (github review tag) re. : [#14872 (comment)](https://github.com/hpcc-systems/HPCC-Platform/pull/14872#discussion_r618303706)

I think I've addressed this now.  Please can you have another look.

jakesmith (changes requested):
> > @shamser tagging you (github review tag) re. : [#14872 (comment)](https://github.com/hpcc-systems/HPCC-Platform/pull/14872#discussion_r618303706)
> 
> I think I've addressed this now. Please can you have another look.

@shamser - The comments were on the regular distributed file get cost implementation, but the most commit changes the super file getCost implementation, which I didn need to change (except to remove isEmptyString(cluster) at head)

jakesmith (commented):
@shamser - looks good, please squash

jakesmith (approved):

jakesmith (approved):
@shamser - looks good

jakesmith (approved):
Looks good.

ghalliday (approved):

PR #14873, opened 2021-04-20
HPCC-25752 Roxie may crash at startup if parallel-loading R queries

richardkchapman:
Provide a way to specify which queries are to be preloaded (we may in the
future want to default this list to be empty), and add code to the R plugin to
fail if it is not preloaded.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [x] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/common/dllserver/thorplugin.cpp
+++ b/common/dllserver/thorplugin.cpp
@@ -846,6 +846,40 @@ void SafePluginMap::loadFromList(const char * pluginsList)
     }
 }
 
+bool SafePluginMap::loadNamed(const char * pluginDirectories, const char * plugin)
+{
+    const char *pluginDir = pluginDirectories;
+    for (;*pluginDir;)
+    {
+        StringBuffer thisFile;
+        while (*pluginDir && *pluginDir != ENVSEPCHAR)
+            thisFile.append(*pluginDir++);
+        if(*pluginDir)
+            pluginDir++;
+
+        if(!thisFile.length())
+            continue;
+        Owned<IFile> dir = createIFile(thisFile);
+        if (dir->isDirectory() == fileBool::foundYes)
+        {
+            Owned<IFile> file = createIFile(addPathSepChar(thisFile).append(plugin));
+            if (file->exists())
+            {
+                if (addPlugin(thisFile, plugin))
+                    return true;
+            }
+        }
+        else if (dir->isFile() == fileBool::foundYes)
+        {
+            StringBuffer tail;
+            splitFilename(thisFile, NULL, NULL, &tail, &tail);
+            if (streq(tail, plugin) && addPlugin(thisFile, plugin))
+                return true;
+        }
+    }
+    return false;
+}
+
 void SafePluginMap::loadFromDirectory(const char * pluginDirectory)
 {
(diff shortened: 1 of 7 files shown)


roxie/ccd/ccdstate.cpp
```diff
@@ -2896,7 +2896,21 @@ extern void loadPlugins()
...
-        plugins->loadFromList(pluginDirectory);
+        if (topology->hasProp("preload"))
+        {
+            Owned<IPropertyTreeIterator> preloads = topology->getElements("preload");
+            ForEach(*preloads)
+            {
+                const char *preload = preloads->query().queryProp(".");
+                if (!streq(preload, "none"))
+                {
+                    VStringBuffer soname(SharedObjectPrefix "%s" SharedObjectExtension, preload);
```
ghalliday, line 2907: Are the plugins guaranteed to be on the path?  Also, what will happen about attempting to load a plugin that is not on the path if it isn't preloaded? 
Also not sure how getAdditionalPluginsPath() will fit into this.

helm/hpcc/values.schema.json
```diff
@@ -720,6 +720,11 @@
         "logging": {
           "$ref": "#/definitions/logging"
         },
+        "preload": {
+          "description": "Preloaded plugins",
+          "type": "array",
```
ghalliday, line 725: This probably should be documented somewhere - e.g., that the lib and .so are not needed.

ghalliday (commented):
@richardkchapman not completely sure about this - particularly path related issues.

roxie/ccd/ccdstate.cpp
richardkchapman: I don't think getAdditionalPluginsPath is used any more since we stopped supporting Python 2. The plugins are successfully loaded because Roxie has RUNPATH set to refer to them, so the pluginsPath variable actually ONLY affects what gets preloaded. Could do with some cleaning up though, and I'll have a think about how best this should be done.

common/dllserver/thorplugin.cpp
```diff
@@ -846,6 +846,29 @@ void SafePluginMap::loadFromList(const char * pluginsList)
...
+    {
+        StringBuffer thisFile;
+        while (*pluginDir && *pluginDir != ENVSEPCHAR)
+            thisFile.append(*pluginDir++);
+        if(*pluginDir)
+            pluginDir++;
+
+        if(!thisFile.length())
+            continue;
+        Owned<IFile> file = createIFile(addPathSepChar(thisFile).append(plugin));
```
ghalliday, line 862: The plugin "directory" list can also be a list of files.  Should this also check if endsWith(thisFile, plugin) - and then add in a similar way to loadFromList above

ghalliday (commented):
@richardkchapman looks good.  One question.

common/dllserver/thorplugin.cpp
richardkchapman: Added

ghalliday (approved):

PR #14875, opened 2021-04-20
HPCC-25755 Add general file/folder watcher mechanism

jakesmith:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jfile.cpp
+++ b/system/jlib/jfile.cpp
@@ -16,6 +16,10 @@
 ############################################################################## */
 #include "platform.h"
 
+#include <atomic>
+#include <unordered_set>
+#include <unordered_map>
+
 #ifdef _WIN32
 #include <errno.h>
 //#include <winsock.h>  // for TransmitFile
@@ -34,8 +38,10 @@
 #endif
 
 #if defined (__linux__)
+#include <poll.h>
 #include <sys/vfs.h>
 #include <sys/sendfile.h>
+#include <sys/inotify.h>
 #endif
 #if defined (__APPLE__)
 #include <sys/mount.h>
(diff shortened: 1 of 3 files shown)


system/jlib/jfile.hpp
```diff
@@ -667,4 +671,56 @@ inline bool canRename(const char * filename) { return queryFileSystemProperties(
...
+    opened        = 0x020,
+    movedFrom     = 0x040,
+    movedTo       = 0x080,
+    created       = 0x100,
+    deleted       = 0x200,
+#endif
+    anyChange     = created | movedTo | movedFrom | closedWrite | modified | metaModified
+};
+
+inline FileWatchEvents operator|(FileWatchEvents a, FileWatchEvents b)
```
richardkchapman, line 705: Should this use BITMASK_ENUM ?

system/jlib/jfile.cpp
```diff
@@ -19,6 +19,9 @@
 #ifdef _WIN32
 #include <errno.h>
 //#include <winsock.h>  // for TransmitFile
+#else
+#include <poll.h>
+#include <sys/inotify.h>
```
richardkchapman, line 24: This does not compile on OSX

system/jlib/jfile.cpp
```diff
@@ -7106,3 +7109,230 @@ const FileSystemProperties & queryFileSystemProperties(const char * filename)
...
+        }
+        std::string filename;
+        int watchFd = -1;
+        FileWatchEvents events = FileWatchEvents::none;
+        bool pendingRemoval = false;
+    };
+    int inotifyFd = -1;
+    int pipefd[2] = {-1, -1};
+    pollfd toPoll[2];
+    std::vector<Owned<CMonitorItem>> monitoredFiles;
```
richardkchapman, line 7131: Is this what we are doing instead of using jarray these days?

system/jlib/jfile.cpp
```diff
@@ -7106,3 +7109,230 @@ const FileSystemProperties & queryFileSystemProperties(const char * filename)
...
+
+            do
+            {
+                std::string filename;
+                if (event->len) // if watching directory event->name is populated with file within that dir.
+                    filename = event->name;
+                else
+                {
+                    /* if watching a file, lookup filename from event handle
+                     * linear search, not worth hash table, not expected to be that many.
```
richardkchapman, line 7161: Famous last words... is a hash table really that hard?

system/jlib/jfile.cpp
```diff
@@ -7106,3 +7109,230 @@ const FileSystemProperties & queryFileSystemProperties(const char * filename)
...
+        {
+            Owned<IException> e = makeErrnoException("CFileEventWatcher::stop() - write -> pipefd[1]");
+            EXCLOG(e);
+        }
+        threaded.join();
+    }
+    virtual bool add(const char *filename, FileWatchEvents events) override
+    {
+        CriticalBlock b(crit);
+        for (const auto &item: monitoredFiles)
```
richardkchapman, line 7249: Hash table?

system/jlib/jfile.cpp
```diff
@@ -7106,3 +7109,230 @@ const FileSystemProperties & queryFileSystemProperties(const char * filename)
...
+#ifndef _WIN32
+class CFileEventWatcher : public CInterfaceOf<IFileEventWatcher>, implements IThreaded
+{
+    class CMonitorItem : public CInterface
+    {
+    public:
+        CMonitorItem(const char *_filename, int _watchFd, FileWatchEvents _events) : filename(_filename), watchFd(_watchFd), events(_events)
+        {
+        }
+        std::string filename;
```
richardkchapman, line 7123: StringAttr might be better - I assume immutable

richardkchapman (changes requested):
Mostly style questions, but need to disable (or implement) on OSX before can be merged. For current anticipated uses, disabling is fine.

system/jlib/jfile.hpp
jakesmith: yes, I thought something like that had been introduced, but couldn't find it..
Have switched to it, also amended it to add operator &

system/jlib/jfile.cpp
jakesmith: have moved into __linux__

system/jlib/jfile.cpp
jakesmith: yes in general, for an owned arrays, and maps, tending to use stl.
More flexible, easier to use, and at least for maps, e.g. unordered_map, a fair bit faster if I remember correctly, although not massively time critical in this context.

system/jlib/jfile.cpp
jakesmith: have converted to use ht(s)

system/jlib/jfile.cpp
jakesmith: need 2 (lookup by fd and filename), have changed to use ht's.

system/jlib/jfile.cpp
jakesmith: std::string easier to use with other stl containers.

system/jlib/jfile.hpp
```diff
@@ -667,4 +671,54 @@ inline bool canRename(const char * filename) { return queryFileSystemProperties(
...
+    closed        = 0x004,
+    closedWrite   = 0x008,
+    metaModified  = 0x010,
+    opened        = 0x020,
+    movedFrom     = 0x040,
+    movedTo       = 0x080,
+    created       = 0x100,
+    deleted       = 0x200,
+#endif
+    anyChange     = created | movedTo | movedFrom | closedWrite | modified | metaModified
```
ghalliday, line 702: and deleted?

system/jlib/jfile.cpp
```diff
@@ -7106,3 +7111,244 @@ const FileSystemProperties & queryFileSystemProperties(const char * filename)
...
+            }
+            flags = fcntl(pipefd[i], F_GETFD, 0);
+            if (-1 != flags)
+            {
+                flags |= FD_CLOEXEC;
+                fcntl(pipefd[i], F_SETFD, flags);
+            }
+        }
+
+        toPoll[0].fd = pipefd[0];
```
ghalliday, line 7214: Could do with some explanation about what the pipe is for.  I think it is to allow the watcher to be cancelled, but not at all clear.

system/jlib/jfile.cpp
```diff
@@ -7106,3 +7111,244 @@ const FileSystemProperties & queryFileSystemProperties(const char * filename)
...
+    int inotifyFd = -1;
+    int pipefd[2] = {-1, -1};
+    pollfd toPoll[2];
+    std::unordered_map<int, Linked<CMonitoredItem>> monitoredFilesByFd;
+    std::unordered_map<std::string, Linked<CMonitoredItem>> monitoredFiles;
+    std::unordered_set<int> pendingRemovals;
+    CriticalSection crit;
+    FileWatchFunc callback;
+    CThreaded threaded;
+    bool stopped = true;
```
ghalliday, line 7139: should be an atomic

system/jlib/jfile.cpp
```diff
@@ -7106,3 +7111,244 @@ const FileSystemProperties & queryFileSystemProperties(const char * filename)
...
+                            {
+                                CMonitoredItem *item = iter->second;
+                                if (item->pendingRemoval) // else fd has been reused/readded
+                                {
+                                    monitoredFiles.erase(item->filename);
+                                    monitoredFilesByFd.erase(iter);
+                                }
+                            }
+                        }
+                    }
```
ghalliday, line 7331: Should clear pendingRemovals (outside of cs)

system/jlib/jfile.cpp
```diff
@@ -7106,3 +7111,244 @@ const FileSystemProperties & queryFileSystemProperties(const char * filename)
...
+    }
+    void checkClose(int h)
+    {
+        if (-1 != h)
+            close(h);
+    }
+public:
+    CFileEventWatcher(FileWatchFunc _callback) : callback(_callback), threaded("CFileEventWatcher", this)
+    {
+        inotifyFd = inotify_init1(IN_NONBLOCK);
```
ghalliday, line 7193: Should this also set IN_CLOEXEC?

system/jlib/jfile.cpp
```diff
@@ -7106,3 +7111,244 @@ const FileSystemProperties & queryFileSystemProperties(const char * filename)
...
+
+                        // clear any pending deletes
+                        CriticalBlock b(crit);
+                        for (const auto fd: pendingRemovals)
+                        {
+                            auto iter = monitoredFilesByFd.find(fd);
+                            if (iter != monitoredFilesByFd.end())
+                            {
+                                CMonitoredItem *item = iter->second;
+                                if (item->pendingRemoval) // else fd has been reused/readded
```
ghalliday, line 7324: I don't think this is ever cleared, so I'm not sure the check is required - comment is confusing...

ghalliday (commented):
@jakesmith generally looks good.  A few comments.
It seems to overlap with the monitorDirectory() functionality - which seems to be barely used, but probably also works for windows.

system/jlib/jfile.cpp
jakesmith: yes, agree, I'll comment to explain.

system/jlib/jfile.hpp
jakesmith: yes, I've added

system/jlib/jfile.cpp
jakesmith: will change.

system/jlib/jfile.cpp
jakesmith: I forgot to clear them.. I'll change,
but I need to do within the cs, others could be adding/removing and updating pendingRemovals on other threads.

system/jlib/jfile.cpp
jakesmith: yes think it should, so should the pipe handles, and I see that can use pipe2 to create those without the fcntl code.
Will change.

system/jlib/jfile.cpp
jakesmith: The item is deleted (last ownership by one of the iterators), but pendingRemovals should be clearered too, fixing.

jakesmith: will clarify comment

system/jlib/jfile.cpp
ghalliday: good point

system/jlib/jfile.cpp
```diff
@@ -7106,3 +7112,237 @@ const FileSystemProperties & queryFileSystemProperties(const char * filename)
...
+            close(h);
+    }
+public:
+    CFileEventWatcher(FileWatchFunc _callback) : callback(_callback), threaded("CFileEventWatcher", this)
+    {
+        inotifyFd = inotify_init1(IN_NONBLOCK|IN_CLOEXEC);
+        if (-1 == inotifyFd)
+            throw makeStringException(-1, "CFileEventWatcher - inotify_init1");
+
+
```
ghalliday, line 7198: trivial: extra newline

ghalliday (approved):
Looks good to me - please squash

system/jlib/jfile.cpp
jakesmith: removed

jakesmith:
@ghalliday - squashed.

richardkchapman (approved):

PR #14877, opened 2021-04-20
HPCC-25756 Auto EclccServer queue refresh

jakesmith:
## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/ecl/eclccserver/eclccserver.cpp
+++ b/ecl/eclccserver/eclccserver.cpp
@@ -870,6 +870,43 @@ static void removePrecompiledHeader()
 // Class EclccServer manages a pool of compile threads
 //------------------------------------------------------------------------------------------------------------------
 
+static StringBuffer &getQueues(StringBuffer &queueNames)
+{
+    Owned<IPropertyTree> config = getComponentConfig();
+#ifdef _CONTAINERIZED
+    bool filtered = false;
+    std::unordered_map<std::string, bool> listenQueues;
+    Owned<IPTreeIterator> listening = config->getElements("listen");
+    ForEach (*listening)
+    {
+        const char *lq = listening->query().queryProp(".");
+        if (lq)
+        {
+            listenQueues[lq] = true;
+            filtered = true;
+        }
+    }
+    Owned<IPTreeIterator> queues = config->getElements("queues");
+    ForEach(*queues)
+    {
+        IPTree &queue = queues->query();
+        const char *qname = queue.queryProp("@name");
+        if (!filtered || listenQueues.count(qname))
+        {
+            if (queueNames.length())
+                queueNames.append(",");
+            getClusterEclCCServerQueueName(queueNames, qname);
+        }
+    }
+#else
+    const char * processName = config->queryProp("@name");
+    SCMStringBuffer scmQueueNames;
+    getEclCCServerQueueNames(scmQueueNames, processName);
+    queueNames.append(scmQueueNames.str());
+#endif
+    return queueNames;
+}
+
 class EclccServer : public CInterface, implements IThreadFactory, implements IAbortHandler
 {
     StringAttr queueNames;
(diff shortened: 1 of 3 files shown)


ecl/eclccserver/eclccserver.cpp
```diff
@@ -849,6 +849,43 @@ static void removePrecompiledHeader()
 // Class EclccServer manages a pool of compile threads
 //------------------------------------------------------------------------------------------------------------------
 
+static StringBuffer &getQueues(StringBuffer &queueNames)
```
richardkchapman, line 852: Feels like this should be shared code?

AttilaVamos:
That is strange. I saw similar exception in teststdlibrary.ecl during implement PR-14857 and happened when multiple version executed parallel and two versions imported and used TestStore.ecl in same time and it is not parallel query safe yet (multiple instances used same name space in same store). However this isn't the versioned testdlibrary.ecl.

ecl/eclccserver/eclccserver.cpp
jakesmith: not sure. NB: this is moved code was in EclccServer::run()
Could be a moved to a shared place, but don't know anywhere else that wants a named eclccserver's filtered queue list at the moment.

ecl/eclccserver/eclccserver.cpp
```diff
@@ -881,7 +918,22 @@ class EclccServer : public CInterface, implements IThreadFactory, implements IAb
...
+
+    void configUpdate()
+    {
+        StringBuffer newQueueNames;
+        getQueues(newQueueNames);
+        if (!newQueueNames.length())
+            ERRLOG("No queues found to listen on");
+        if (!strsame(queueNames, newQueueNames))
+        {
+            updatedQueueNames.set(newQueueNames);
```
richardkchapman, line 933: Does this need protecting from a race with the thread that reads it?

richardkchapman (dismissed):
Possible thread safety issue?

ecl/eclccserver/eclccserver.cpp
jakesmith: theoretically it could race with another call to update, have protected and rearrange slightly (so more common code)

esp/smc/SMCLib/TpWrapper.cpp
```diff
@@ -29,6 +29,17 @@
 #include "dautils.hpp"
 #include "dameta.hpp"
 
+static unsigned reloadConifgCBId = (unsigned)-1;
```
ghalliday, line 32: If this is going to happen frequently, I wonder if we should have a helper class that can be used to
a) remove the hook on destruction
b) (conditionally) create the hook from a provided function.

So then would become
```
static ConfigUpdateHook reloadConfigHook;
reloadConfigHook.install(func);

```
or
`reloadConfigHook.installOnce(func);
`?

Could also have a constructor that took a function which called the hook, but couldn't be used everywhere because it is too early if called statically.

ecl/eclccserver/eclccserver.cpp
ghalliday: It still needs a cs - since this thread could be updating updatedQueueNames at the same time as it is being checked.

ecl/eclccserver/eclccserver.cpp
```diff
@@ -881,36 +918,71 @@ class EclccServer : public CInterface, implements IThreadFactory, implements IAb
...
+            queue->cancelAcceptConversation();
+            PROGLOG("Updating queue due to queue names change from '%s' to '%s'", queueNames.str(), newQueueNames.str());
```
ghalliday, line 936: This tracing also needs to be within the critical section

esp/smc/SMCLib/TpWrapper.cpp
```diff
@@ -2410,29 +2421,36 @@ static void refreshValidTargets()
...
         throw makeStringException(ECLWATCH_INVALID_CLUSTER_NAME, "Empty target name.");
 
     CriticalBlock block(validTargetSect);
-    if (targetsDirty)
+#ifdef _CONTAINERIZED
+    if ((unsigned) -1 == reloadConifgCBId) // once only
     {
         refreshValidTargets();
-        targetsDirty = false;
+        reloadConifgCBId = installConfigUpdateHook(configUpdate);
```
ghalliday, line 2442: Should installConfigUpdateHook have a flag to indicate the function should be called when it is registered?
This could then be
reloadConfigHook.installOnce(configUpdate, true);
Could be an option purely on the the (new) hook class, rather than the install function.

ghalliday (dismissed):
a couple of issues and a suggestion for making code easier to write

ecl/eclccserver/eclccserver.cpp
jakesmith: whoops - I was supposed to add there, as part of this change.

ecl/eclccserver/eclccserver.cpp
jakesmith: yes, actually so does 'queue', but will link and and call cancelAcceptConversation outside.

esp/smc/SMCLib/TpWrapper.cpp
jakesmith: Yes, I think worth it, I've implemented it (last commit in https://github.com/hpcc-systems/HPCC-Platform/pull/14966),
but think it should probably be in a separate PR after this one, as it's general, separate from this eclccserver change and modified the existing hooks

esp/smc/SMCLib/TpWrapper.cpp
jakesmith: have implemented in https://github.com/hpcc-systems/HPCC-Platform/pull/14966

jakesmith:
@ghalliday - please see new commit and comments.

ecl/eclccserver/eclccserver.cpp
```diff
@@ -963,7 +1040,13 @@ class EclccServer : public CInterface, implements IThreadFactory, implements IAb
     virtual bool onAbort() 
     {
         running = false;
-        if (queue)
+        Linked<IJobQueue> currentQueue = queue;
```
ghalliday, line 1043: This default value isn't safe.  E.g. it could be released (and set to null) before the critical section is entered.

ecl/eclccserver/eclccserver.cpp
```diff
@@ -881,36 +918,76 @@ class EclccServer : public CInterface, implements IThreadFactory, implements IAb
...
         while (running)
         {
             try
             {
+                bool newQueues = false;
+                {
+                    CriticalBlock b(queueUpdateCS);
+                    if (updatedQueueNames)
+                    {
+                        queueNames.set(updatedQueueNames);
```
ghalliday, line 977: check running again inside the critical section - otherwise there is a small chance the queue might change - leading this to wait on the new queue, while the old one has been cancelled.

esp/smc/SMCLib/TpWrapper.cpp
```diff
@@ -29,6 +29,17 @@
...
 #include "dameta.hpp"
 
+static unsigned reloadConifgCBId = (unsigned)-1;
+MODULE_INIT(INIT_PRIORITY_STANDARD)
+{
+    return true;
+}
+MODULE_EXIT()
+{
+    if ((unsigned)-1 != reloadConifgCBId)
```
ghalliday, line 39: typo conifg -> config

ecl/eclccserver/eclccserver.cpp
```diff
@@ -881,36 +918,76 @@ class EclccServer : public CInterface, implements IThreadFactory, implements IAb
     bool running;
     CSDSServerStatus serverstatus;
     Owned<IJobQueue> queue;
+    CriticalSection queueUpdateCS;
+    StringAttr updatedQueueNames;
+    unsigned reloadConifgCBId = 0;
```
ghalliday, line 923: typo conifg->config

esp/smc/SMCLib/TpWrapper.cpp
ghalliday: It should be a separate PR so that you can clean up the other existing hooks.

ghalliday (commented):
@jakesmith a few minor threading comments

ecl/eclccserver/eclccserver.cpp
jakesmith: yes, that was a mistake, it is set in CS, shouldn't have a default

ecl/eclccserver/eclccserver.cpp
jakesmith: fixed (and others)

ecl/eclccserver/eclccserver.cpp
jakesmith: Added check

jakesmith (commented):
@ghalliday - please see new commit

ghalliday (dismissed):
trivial: Conifg -> config also in tpwrapper.cpp

jakesmith:
> trivial: Conifg -> config also in tpwrapper.cpp

Thanks, now fixed.

ghalliday:
@jakesmith can you squash

jakesmith:
@ghalliday - squashed

ghalliday (approved):

richardkchapman (approved):

PR #14884, opened 2021-04-22
HPCC-25759 Add support for bare-metal storage planes to helm charts

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dali/base/dadfs.cpp
+++ b/dali/base/dadfs.cpp
@@ -39,6 +39,7 @@
 #include "dadfs.hpp"
 #include "eclhelper.hpp"
 #include "seclib.hpp"
+#include "dameta.hpp"
 
 #include <string>
 #include <vector>
@@ -10060,10 +10061,9 @@ public:
         return createClusterGroup(grp_unknown, hosts, path, nullptr, false, false);
     }
 
-    void ensureStorageGroup(bool force, const char * name, unsigned numDevices, const char * path, StringBuffer & messages)
+    void ensureConsistentStorageGroup(bool force, const char * name, IPropertyTree * newClusterGroup, StringBuffer & messages)
     {
         IPropertyTree *existingClusterGroup = queryExistingGroup(name);
-        Owned<IPropertyTree> newClusterGroup = createStorageGroup(name, numDevices, path);
         bool matchExisting = clusterGroupCompare(newClusterGroup, existingClusterGroup);
         if (!existingClusterGroup || !matchExisting)
         {
@@ -10072,14 +10072,14 @@ public:
                 VStringBuffer msg("New cluster layout for cluster %s", name);
                 UWARNLOG("%s", msg.str());
                 messages.append(msg).newline();
-                addClusterGroup(name, newClusterGroup.getClear(), false);
+                addClusterGroup(name, LINK(newClusterGroup), false);
             }
             else if (force)
             {
                 VStringBuffer msg("Forcing new group layout for storageplane %s", name);
                 UWARNLOG("%s", msg.str());
                 messages.append(msg).newline();
-                addClusterGroup(name, newClusterGroup.getClear(), false);
+                addClusterGroup(name, LINK(newClusterGroup), false);
             }
             else
             {
@@ -10090,12 +10090,20 @@ public:
         }
     }
 
+    void ensureStorageGroup(bool force, const char * name, unsigned numDevices, const char * path, StringBuffer & messages)
+    {
+        Owned<IPropertyTree> newClusterGroup = createStorageGroup(name, numDevices, path);
+        ensureConsistentStorageGroup(force, name, newClusterGroup, messages);
+    }
+
     void constructStorageGroups(bool force, StringBuffer &messages)
     {
         IPropertyTree & global = queryGlobalConfig();
         IPropertyTree * storage = global.queryPropTree("storage");
         if (storage)
         {
+            normalizeHostGroups();
+
             Owned<IPropertyTreeIterator> planes = storage->getElements("planes");
             ForEach(*planes)
             {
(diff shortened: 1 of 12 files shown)


ghalliday:
I have tested the following:
Spraying a variable length csv file from bare metal to k8s (fixed bug in esp)
Read files using file::ip::x syntax
Written a file from thor to a bare metal cluster
Read a logical file in hthor/thor/roxie from a bare metal cluster
Added tests to helm testing

Here is some extra definitions I used when testing:
```
storage:
  hostGroups:
  - name: mySillyGroup
    hosts: [ 'rubbish', 'fred', '192.168.86.202', 'jumbo' ]
  # [ 'jumbo', 'fred', '192.168.86.202' ]
  - name: sillySubset
    count: 3
    hostGroup: mySillyGroup
    offset: 1
    delta: 2
  - name: singleValueHostGroup
    count: 1
    hostGroup: sillySubset
    offset: 2
  planes:
  - name: mydesktop   # does this need to be lower case??
    prefix: /home/gavin/temp
    hosts: [ '192.168.86.202' ]
    labels: ['lz', 'data']
  - name: mydesktop2
    prefix: /home/gavin/temp
    hostGroup: singleValueHostGroup
    labels: ['lz']

```
There are several issues that I will open separate jiras for:
hthor can't write to a remote bare metal cluster
roxie does not support cluster output syntax in k8s
upper case plane names cause problems (either fix or add a check in help charts)

dali/base/dameta.cpp
```diff
@@ -38,6 +44,43 @@ IPropertyTree * queryStoragePlane(const char * name)
...
+            unsigned numHosts = cur.getPropInt("@count", hosts.ordinality());
+            unsigned offset = cur.getPropInt("@offset");
+            if (offset + numHosts > hosts.ordinality())
+                throw makeStringExceptionV(-1, "Group %s extends past the end of the base group %s", name, baseGroup);
+
+            unsigned delta = cur.getPropInt("@delta");
+            for (unsigned i=0; i < numHosts; i++)
+            {
+                unsigned baseIndex = offset + (i + delta) % numHosts;
+                cur.addPropTree("hosts")->setProp(nullptr, hosts.item(baseIndex));
```
jakesmith, line 78: could instead be:
```
cur.addProp("hosts", hosts.item(baseIndex));
```

dali/base/dameta.cpp
```diff
@@ -38,6 +44,43 @@ IPropertyTree * queryStoragePlane(const char * name)
...
+
+            unsigned numHosts = cur.getPropInt("@count", hosts.ordinality());
+            unsigned offset = cur.getPropInt("@offset");
+            if (offset + numHosts > hosts.ordinality())
+                throw makeStringExceptionV(-1, "Group %s extends past the end of the base group %s", name, baseGroup);
+
+            unsigned delta = cur.getPropInt("@delta");
+            for (unsigned i=0; i < numHosts; i++)
+            {
+                unsigned baseIndex = offset + (i + delta) % numHosts;
```
jakesmith, line 77: related to comment re. simpler if just 'offset' and 'count' in schema..
this would work without delta, if planes used different offsets, with count and wrapping was implied.

dali/base/dameta.cpp
```diff
@@ -38,6 +44,43 @@ IPropertyTree * queryStoragePlane(const char * name)
...
+        IPropertyTree & cur = hostGroupIter->query();
+        if (!cur.hasProp("hosts"))
+        {
+            const char * name = cur.queryProp("@name");
+            const char * baseGroup = cur.queryProp("@hostGroup");
+            IPropertyTree * match = queryHostGroup(baseGroup, true);
+            StringArray hosts;
+            Owned<IPropertyTreeIterator> hostIter = match->getElements("hosts");
+            ForEach (*hostIter)
+                hosts.append(hostIter->query().queryProp(nullptr));
```
jakesmith, line 64: very picky: I noticed you used a std::vector in dadfs (line 10141) for same purpose. Use same here?

dali/base/dameta.cpp
```diff
@@ -22,13 +22,19 @@
...
+    if (!isEmptyString(name))
+    {
+        VStringBuffer xpath("storage/hostGroups[@name='%s']", name);
+        IPropertyTree & global = queryGlobalConfig();
+        IPropertyTree * match = global.queryPropTree(xpath);
+        if (match)
+            return match;
+    }
+    if (required)
+        throw makeStringExceptionV(0, "No entry found for hostGroup: '%s'", name ? name : "<null>");
```
jakesmith, line 36: minor: are we tending to default to error code 0, or -1?

jakesmith (commented):
@ghalliday looks good, just a few minor/trivial comments.

devdoc/NewFileProcessing.rst
```diff
@@ -11,20 +11,21 @@ storage:
   hostGroups:
   - name: <required>
     hosts: [ .... ]
+  - name: <required>
+    hostGroup: <name>
+    count: <unsigned:#hosts>    # how many hosts within the host group are used ?(default is number of hosts)
+    offset: <unsigned:0>        # index of first host included in the derive group
+    delta: <unsigned:0>         # first host within the range[offset..offset+count-1] in the derived group
```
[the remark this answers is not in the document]
ghalliday: No, they are both needed.  E.g.
(a,b,c,d,e,f)  offset:3, delta:1, count:3 gives (e,f,d).
It is needed to define replication of smaller sized thors.

dali/base/dameta.cpp
ghalliday: see comment above - it is needed.

dali/base/dameta.cpp
ghalliday: We should be defining error codes for all the errors, rather than being lazy!

dali/base/dameta.cpp
ghalliday: I could do - the other code was pre-existing.

devdoc/NewFileProcessing.rst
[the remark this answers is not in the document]
jakesmith: ok thanks, yes that makes sense.

dali/base/dadfs.cpp
```diff
@@ -10104,25 +10112,45 @@ class CInitGroups
...
+
+                    //A bare-metal storage plane defined in terms of a hostGroup
+                    newClusterGroup.setown(createClusterGroup(grp_unknown, hosts, prefix, nullptr, false, false));
+                }
+                else if (plane.hasProp("hostGroup"))
                 {
-                    IPropertyTree *existingClusterGroup = queryExistingGroup(gname);
-                    if (!existingClusterGroup)
-                        UNIMPLEMENTED_X("Bare metal storage planes not yet supported");
+                    throw makeStringExceptionV(-1, "Use 'hosts' rather than 'hostGroup' for inline list of hosts for plane %s", name);
```
shamser, line 10136: This should be caught by the helm's schema validation.

dali/base/dameta.cpp
shamser: I would expect this validation to take place (primarily) in the helm templates.  It would be nice for invalid configuration to cause aborts early rather than starting up and then failing.

shamser (commented):
@ghalliday I'm wondering why we don't support ip address with subnet mask.  It makes managing configuration on the larger clusters easier.  I would have thought it wouldn't be too difficult to implement.

Secondly, it seems a lot be validation could take place in the helm templates.  I would have thought showing an error early was preferable to starting up and failing later.

dali/base/dadfs.cpp
ghalliday: It is caught by the schema.  Extra sanity checking - partly in case we use the same scheme for bare metal.

ghalliday:
@shamser the validation checks in the code are also caught in the schema (see the example helm error test).  I don't support ip ranges or subnets/masks because the old scheme didn't and our tendency will be to name the nodes.  If it is required that can be a separate pr - it could expand them in normalizeHostGroups().

shamser (approved):

dali/base/dameta.cpp
jakesmith: >We should be defining error codes for all the errors, rather than being lazy!

agree, but even when being lazy - better to use a consistent default ;)
otherwise might think 0 has a special meaning

jakesmith (approved):
Only picky comments, looks good to merge

dali/base/dameta.cpp
ghalliday: changed

devdoc/NewFileProcessing.rst
```diff
@@ -11,20 +11,21 @@ storage:
   hostGroups:
   - name: <required>
     hosts: [ .... ]
+  - name: <required>
+    hostGroup: <name>
+    count: <unsigned:#hosts>    # how many hosts within the host group are used ?(default is number of hosts)
+    offset: <unsigned:0>        # index of first host included in the derive group
```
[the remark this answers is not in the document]
ghalliday: fixed

richardkchapman (approved):

PR #14906, opened 2021-04-30
HPCC-25808 Remove remaining instances of atomic_t from the system

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:


ghalliday:
@kenrowland fyi - this PR is removing the remaining legacy atomics

roxie/ccd/ccdserver.cpp
```diff
@@ -25274,16 +25274,16 @@ class CJoinGroup : public CInterface
...
         // NOTE - as soon as endMarkersPending and groupStart->endMarkersPending are decremented to zero this object may get released asynchronously by other threads
         // There must therefore be nothing in this method after them that acceses member variables. Think of it as a delete this...
         // In particular, we can't safely reference groupStart after the dec_and_test of endMarkersPending, hence copy local first 
         CJoinGroup *localGroupStart = groupStart;
-        if (atomic_dec_and_test(&endMarkersPending))
+        if (--endMarkersPending == 0)
         {
             if (localGroupStart)
-                return atomic_dec_and_test(&localGroupStart->endMarkersPending);
+                return localGroupStart->endMarkersPending-- == 0;
```
richardkchapman, line 25286: Is this right?

esp/platform/espprotocol.cpp
```diff
@@ -28,21 +28,21 @@
 
 typedef IXslProcessor * (*getXslProcessor_func)();
 
-static atomic_t gActiveRequests;
+static std::atomic<unsigned> gActiveRequests;
```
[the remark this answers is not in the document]
kenrowland: Concern here is that this will be used a metric (I have a prototype running that accepts a reference to the underlying metric value). We decided to use 64bit values, so my prototype uses std::atomic<__uint64> gActiveRequests;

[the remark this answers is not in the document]
ghalliday: This is a conversion of what is there already.  I suspect your metric classes that use a value that is calculated elsewhere should be an example of a template class - it reports as an uin64, but can read from any value.

ghalliday:
See also https://github.com/hpcc-systems/LN/pull/986

roxie/ccd/ccdfile.cpp
```diff
@@ -314,8 +314,8 @@ class CRoxieLazyFileIO : implements ILazyFileIO, implements IDelayedFile, public
                 }
             }
             lastAccess = msTick();
-            atomic_inc(&numFilesOpen[remote]);
-            if ((unsigned) atomic_read(&numFilesOpen[remote]) > maxFilesOpen[remote])
+            ++numFilesOpen[remote];
```
richardkchapman, line 317: Minor, but this might be clearer and more efficient if the two lines were combined

thorlcr/slave/traceslave.hpp
```diff
@@ -173,15 +173,15 @@ class CTracingStream : public CSimpleInterfaceOf<IEngineRowStream>, implements I
...
+        int bufToDump = rowBufInUse;
         int inactiveBuf = 1 - bufToDump;
 
         // Queue any remaining rows from inactive buffer as the oldest row is skipped
         // when queue out from active buffer to avoid race condition issues (this inactive buffer
         // may have been formerly an active buffer)
         buffers[inactiveBuf].queueOut(rowBufferLogCache, false);
 
-        atomic_set(&rowBufInUse, inactiveBuf);// Swap Active & inactiveBuf
+        rowBufInUse = inactiveBuf;// Swap Active & inactiveBuf
```
richardkchapman, line 184: Not new but is this safe?

fs/dafsserver/dafsserver.cpp
```diff
@@ -2789,13 +2789,13 @@ class CRemoteFileServer : implements IRemoteFileServer, public CInterface
         StructArrayOf<OpenFileInfo> openFiles;
         Owned<IDirectoryIterator> opendir;
         unsigned            lasttick, lastInactiveTick;
-        atomic_t            &globallasttick;
+        std::atomic<unsigned> &globallasttick;
```
richardkchapman, line 2792: Not new, but this feels more like a volatile than an atomic?

richardkchapman (dismissed):
Some minor comments, mostly relating to the code as it was before the change rather than newly-introduced code.

thorlcr/slave/traceslave.hpp
ghalliday: I don't think this code is 100% thread safe.  A row could be being appended at the same time as queueOut is called.  I think that could potentially cause a row to be accessed that is then overwritten, or possibly other issues.  However that is a separate issue.

fs/dafsserver/dafsserver.cpp
ghalliday: I think atomic is correct.  volatile was really meant for memory mapped io devices, where you had to guarantee that data was  re-read.

richardkchapman (approved):

PR #14912, opened 2021-04-30
HPCC-25787 WUQuery very slow if workunit(s) have lots of app values

jakesmith:
The implementation of cached application values was pathologically
slow when large numbers existed (e.g. because a job had sprayed
thousands of files).

The implementation held the parent tree and build an index xpath
to each element ("*[<n>]").
This cause a linear search for each item, for every access method
of each IConstWUAppValue, i.e. each queryName() and queryValue()
call.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/common/workunit/workunit.ipp
+++ b/common/workunit/workunit.ipp
@@ -29,11 +29,10 @@
 
 class WORKUNIT_API CLocalWUAppValue : implements IConstWUAppValue, public CInterface
 {
-    Owned<IPropertyTree> p;
-    StringAttr prop;
+    Owned<const IPropertyTree> owner, props;
 public:
     IMPLEMENT_IINTERFACE;
-    CLocalWUAppValue(IPropertyTree *p,unsigned child);
+    CLocalWUAppValue(const IPropertyTree *_owner, const IPropertyTree *_props);
 
     virtual const char *queryApplication() const;
     virtual const char *queryName() const;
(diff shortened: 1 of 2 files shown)


jakesmith:
@ghalliday - please review - not sure what best target is. Targeting 8.0.x for now.

common/workunit/workunit.ipp
```diff
@@ -66,7 +65,7 @@ public:
 
 //==========================================================================================
 
-template <typename T, typename IT> struct CachedTags
+template <typename T, typename IT, class BASE> struct CachedTags : public BASE
```
ghalliday, line 68: BASE isn't really a base class.   You could make the add() functions static and then avoid adding it as a base class.  Probably even cleaner is to use a function argument:

```
template <typename T, typename IT, void (* ADD)(const IPropertyTree *p, IArrayOf<IT> &dst)> struct CachedTags

void addStatistic(const IPropertyTree * p, IArrayOf<IConstWUStatistic> &dst)
{
    dst.append(*new CLocalWUStatistic(&OLINK(*p)));
}

typedef CachedTags<CLocalWUStatistic, IConstWUStatistic, addStatistic> CachedStatistics;

```

ghalliday (approved):
The change looks good and could be merged.  One suggestion for improving the code structure.

common/workunit/workunit.ipp
jakesmith: that is better, didn't think of a template function arg. for some reason.

jakesmith:
@ghalliday - have changed to use function.

ghalliday (approved):

ghalliday:
@richardkchapman should this be also merged into an earlier branch?

PR #14917, opened 2021-05-04
HPCC-25821 Remove all dependencies on environment in container build

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/common/environment/dalienv.hpp
+++ b/common/environment/dalienv.hpp
@@ -26,12 +26,12 @@
 
 #include "environment.hpp"
 
+#ifndef _CONTAINERIZED
+
 interface IFile;
 class RemoteFilename;
 
 extern ENVIRONMENT_API EnvMachineOS queryOS(const IpAddress & ip);
-
-
 extern ENVIRONMENT_API bool canAccessFilesDirectly(const RemoteFilename & file);
 extern ENVIRONMENT_API bool canAccessFilesDirectly(const IpAddress & ip);
 extern ENVIRONMENT_API bool canAccessFilesDirectly(const char * ipText);
@@ -53,4 +53,10 @@ extern ENVIRONMENT_API void envInstallNASHooks(IPropertyTree *nasPTree, SocketEn
 extern ENVIRONMENT_API IPropertyTree *envGetInstallNASHooks(SocketEndpoint *myEp=NULL);
 extern ENVIRONMENT_API IPropertyTree *envGetInstallNASHooks(IPropertyTree *nasPTree, SocketEndpoint *myEp=NULL);
 
+#else
+
+inline bool canSpawnChildProcess(const IpAddress & ip) { return true; }
+inline EnvMachineOS queryOS(const IpAddress & ip) { return MachineOsLinux; }
+
+#endif
 #endif
--- a/dali/dalidiag/dalidiag.cpp
+++ b/dali/dalidiag/dalidiag.cpp
@@ -302,6 +302,7 @@ void partInfo(const char *name,unsigned copy)
 
 void nodeInfo(const char *ip)
 {
+#ifndef _CONTAINERIZED
     Owned<IEnvironmentFactory> factory = getEnvironmentFactory(false);
     Owned<IConstEnvironment> env = factory->openEnvironment();
 
@@ -331,6 +332,7 @@ void nodeInfo(const char *ip)
     }
     else
         printf("ERROR: cannot find '%s' in Dali Environment\n",ip);
+#endif
 }
 
 void slaveNode(const char *thor,unsigned n)
--- a/dali/dfu/dfurun.cpp
+++ b/dali/dfu/dfurun.cpp
@@ -1075,9 +1075,11 @@ public:
                 : running(_running)
             {
                 if (--running == 0) {
+#ifndef _CONTAINERIZED
                     Owned<IEnvironmentFactory> envf = getEnvironmentFactory(false);
                     Owned<IConstEnvironment> env = envf->openEnvironment();
                     env->clearCache();
+#endif
                 }
             }
             ~CenvClear()
@@ -1263,6 +1265,9 @@ public:
                     {
                         if (options->getPush())
                         {
+#ifdef _CONTAINERIZED
+                            UNIMPLEMENTED_X("CONTAINERIZED(ForeignFileCopy:push)");
+#else
                             // need to set ftslave location
                             StringBuffer progpath;
                             StringBuffer workdir;
@@ -1271,6 +1276,7 @@ public:
                             {
                                 opttree->setProp("@slave",progpath.str());
                             }
+#endif
                         }
                     }
                     if (destination->getMultiCopy()&&!destination->getWrap())
--- a/dali/ft/filecopy.cpp
+++ b/dali/ft/filecopy.cpp
@@ -43,6 +43,25 @@
 #include "dalienv.hpp"
 #include "ftbase.ipp"
 
+#ifdef _CONTAINERIZED
+//Temporary see HPCC-25822
+inline bool canAccessFilesDirectly(const RemoteFilename & file)
+{
+    if (file.queryEndpoint().port!=0)
+        return false;
+    const IpAddress & ip = file.queryIP();
+    if (ip.isLocal()||ip.isNull())  // the isNull check is probably an error but saves time
+        return true;                // I think usually already checked, but another can't harm
+    return false;
+}
+
+inline void setCanAccessDirectly(RemoteFilename & file)
+{
+    setCanAccessDirectly(file,canAccessFilesDirectly(file));
+}
+
+#endif
+
 #define DEFAULT_MAX_CONNECTIONS 800
 #define PARTITION_RECOVERY_LIMIT 1000
 #define EXPECTED_RESPONSE_TIME          (60 * 1000)
@@ -935,6 +954,8 @@ void FileSprayer::beforeTransfer()
     }
 
     throttleNicSpeed = options->getPropInt(ANthrottle, 0);
+#ifndef _CONTAINERIZED
+    //MORE: This is very old windows support code.  We could add support for per-plane throttling if it is required.
     if (throttleNicSpeed == 0 && !usePullOperation() && targets.ordinality() == 1 && sources.ordinality() > 1)
     {
         Owned<IEnvironmentFactory> factory = getEnvironmentFactory(true);
@@ -951,6 +972,7 @@ void FileSprayer::beforeTransfer()
             }
         }
     }
+#endif
 }
 
 
--- a/deployment/configgen/CMakeLists.txt
+++ b/deployment/configgen/CMakeLists.txt
@@ -48,6 +48,10 @@ target_link_libraries ( configgen
         environment 
         )
 
+if (CONTAINERIZED)
+    message(FATAL_ERROR "Configgen is not included in container builds")
+endif()
+
 IF (USE_OPENSSL)
     target_link_libraries ( configgen
     	securesocket
--- a/esp/services/ws_configmgr/CMakeLists.txt
+++ b/esp/services/ws_configmgr/CMakeLists.txt
@@ -69,10 +69,13 @@ target_link_libraries(ws_configmgr
     jlib
     xmllib
     esphttp
-    environment
     dllserver
     configmgr
     )
+
+if (NOT CONTAINERIZED)
+    target_link_libraries ( ws_configmgr environment )
+endif()
 
 IF (USE_OPENSSL)
     target_link_libraries ( ws_configmgr
--- a/esp/services/ws_dfu/ws_dfuService.cpp
+++ b/esp/services/ws_dfu/ws_dfuService.cpp
@@ -101,6 +101,7 @@ short days[12] = {31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31};
 
 CThorNodeGroup* CThorNodeGroupCache::readNodeGroup(const char* _groupName)
 {
+#ifndef _CONTAINERIZED
     Owned<IEnvironmentFactory> factory = getEnvironmentFactory(true);
     Owned<IConstEnvironment> env = factory->openEnvironment();
     Owned<IPropertyTree> root = &env->getPTree();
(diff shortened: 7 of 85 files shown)


ghalliday:
See also https://github.com/hpcc-systems/LN/pull/986

esp/smc/SMCLib/TpContainer.cpp
```diff
@@ -0,0 +1,781 @@
...
+        list.append(*dropZone.getClear());
+    }
+}
+
+void CTpWrapper::getTpSparkThors(double clientVersion, const char* name, IArrayOf<IConstTpSparkThor>& list)
+{
+    UNIMPLEMENTED_X("CONTAINERIZED(CTpWrapper::getTpSparkThors)");
+}
+
+void CTpWrapper::appendTpMachine(double clientVersion, IConstEnvironment* constEnv, IConstInstanceInfo& instanceInfo, IArrayOf<IConstTpMachine>& machines)
```
wangkx, line 401: Should the appendTpMachine() (and other 'common code') be moved to TpCommon.cpp?

wangkx (commented):
@ghalliday I reviewed the changes related to the esp services. Looks good. A few minor questions.

esp/services/ws_ecl/ws_ecl_service.cpp
```diff
@@ -1533,9 +1536,13 @@ int CWsEclBinding::getGenForm(IEspContext &context, CHttpRequest* request, CHttp
     xform->setParameter("includeSoapTest", "1");
     xform->setParameter("useTextareaForStringArray", "1");
 
+#ifdef _CONTAINERIZED
```
afishbeck, line 1539: Could be as simple as:

```
#ifdef _CONTAINERIZED
    bool isRoxie = wsecl->connMap.getValue(wsinfo.qsetname)!=nullptr;
#else
    Owned<IConstWUClusterInfo> clusterInfo = getTargetClusterInfo(wuinfo.qsetname);
    bool isRoxie = clusterInfo && (clusterInfo->getPlatform() == RoxieCluster);
#endif

xform->setParameter("includeRoxieOptions", isRoxie ? "1" : "0");
```

shamser (changes requested):

ghalliday:
A general comment - the intent of this change to to remove all the references, and highlight the code that needs to change.  It isn't aiming to fix any of those issues.  I have opened separate jiras to do that, since many of them will require some thought.

esp/services/ws_dfu/ws_dfuService.cpp
```diff
@@ -182,9 +184,14 @@ void CWsDfuEx::init(IPropertyTree *cfg, const char *process, const char *service
 
     setDaliServixSocketCaching(true);
 
+#ifdef _CONTAINERIZED
+    IERRLOG("CONTAINERIZED(CWsDfuEx::init)");
+    maxFileAccessExpirySeconds = defaultMaxFileAccessExpirySeconds;
```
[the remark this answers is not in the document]
ghalliday: We need to make a lot of changes - but they should be done as separate jiras.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -1236,6 +1245,7 @@ unsigned WsWuInfo::getWorkunitThorLogInfo(IArrayOf<IEspECLHelpFile>& helpers, IE
                 thorLogList.append(*thorLog.getLink());
             }
         }
+        #endif
```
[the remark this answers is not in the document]
ghalliday: I will fix.

esp/smc/SMCLib/TpContainer.cpp
```diff
@@ -0,0 +1,781 @@
...
+
+void CTpWrapper::getClusterMachineList(double clientVersion,
+                                       const char* ClusterType,
+                                       const char* ClusterPath,
+                                       const char* ClusterDirectory,
+                                       IArrayOf<IEspTpMachine> &MachineList,
+                                       bool& hasThorSpareProcess,
+                                       const char* ClusterName)
+{
+    IWARNLOG("UNIMPLEMENTED: CONTAINERIZED(CTpWrapper::getClusterMachineList)");
```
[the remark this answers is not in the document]
ghalliday: Because the code without this change would not have thrown an exception, but would have reported an error.  I tried to make the changes so the behaviour would not change (and break functionality that half works at the moment).

esp/smc/SMCLib/TpContainer.cpp
```diff
@@ -0,0 +1,781 @@
...
+    throw MakeStringExceptionDirect(ECLWATCH_CANNOT_GET_ENV_INFO, MSG_FAILED_GET_ENVIRONMENT_INFO);
+}
+
+void CTpWrapper::getTpEclCCServers(IArrayOf<IConstTpEclServer>& list, const char* serverName)
+{
+    throw MakeStringExceptionDirect(ECLWATCH_CANNOT_GET_ENV_INFO, MSG_FAILED_GET_ENVIRONMENT_INFO);
+}
+
+void CTpWrapper::getTpEclCCServers(IPropertyTree* environmentSoftware, IArrayOf<IConstTpEclServer>& list, const char* serverName)
+{
```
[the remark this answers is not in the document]
ghalliday: see comment above

esp/smc/SMCLib/TpContainer.cpp
ghalliday: It should probably be removed from this file - it is only called from the CONTAINERIZED code.  I didn't go through this file and remove all code that is no longer required.

esp/services/ws_ecl/ws_ecl_service.cpp
ghalliday: quite possibly.  should be fixed as part of HPCC-25850.

common/environment/dalienv.hpp
```diff
@@ -53,4 +53,10 @@ extern ENVIRONMENT_API void envInstallNASHooks(IPropertyTree *nasPTree, SocketEn
 extern ENVIRONMENT_API IPropertyTree *envGetInstallNASHooks(SocketEndpoint *myEp=NULL);
 extern ENVIRONMENT_API IPropertyTree *envGetInstallNASHooks(IPropertyTree *nasPTree, SocketEndpoint *myEp=NULL);
 
+#else
+
+inline bool canSpawnChildProcess(const IpAddress & ip) { return true; }
```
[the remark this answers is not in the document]
ghalliday: I prefer as little conditionally compiled code as possible - it makes the code easier to read, and ensures it compiles with all the different options.  So I prefer a dummy implementation in containerized to having #ifs around the call.
This PR in general is slightly different because it is removing a dependency on a dll, but I would hope some of the conditional compiles can be replaced by general functions instead (that call either bare metal or containerized implementations.)

dali/ft/filecopy.cpp
```diff
@@ -935,6 +954,8 @@ void FileSprayer::beforeTransfer()
     }
 
     throttleNicSpeed = options->getPropInt(ANthrottle, 0);
+#ifndef _CONTAINERIZED
```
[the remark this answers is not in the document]
ghalliday: Why?  It is still a valid option.  This PR is only removing the dependency on the environment.

dali/dfu/dfurun.cpp
```diff
@@ -1075,9 +1075,11 @@ class CDFUengine: public CInterface, implements IDFUengine
                 : running(_running)
             {
                 if (atomic_dec_and_test(&running)) {
+#ifndef _CONTAINERIZED
```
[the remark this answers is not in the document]
ghalliday: That comment is specifically referring to the few lines of code which caused the spray to fail.  i.e.

```
    #ifndef _DEBUG
     StringBuffer gname;
     if (!destination->getRemoteGroupOverride()&&!testLocalCluster(destination->getGroupName(0,gname).str()))
    { throw MakeStringException(-1,"IMPORT cluster %s is not recognized locally",gname.str()); }
    #endif

```

I can't think why you would have any code which is only executed in release.

esp/services/ws_ecl/ws_ecl_service.cpp
ghalliday: Yes - there are separate jiras for fixing the issues.

dali/dalidiag/dalidiag.cpp
```diff
@@ -302,6 +302,7 @@ void partInfo(const char *name,unsigned copy)
 
 void nodeInfo(const char *ip)
 {
+#ifndef _CONTAINERIZED
```
jakesmith, line 305: minor: perhaps should throw an unexpected error, since it should never be called in container mode

deployment/configgen/CMakeLists.txt
```diff
@@ -45,9 +45,12 @@ install ( TARGETS configgen DESTINATION sbin )
 target_link_libraries ( configgen 
         deploy 
         jlib 
-        environment 
         )
 
+if (NOT CONTAINERIZED)
+    target_link_libraries ( configgen environment )
```
jakesmith, line 51: doesn't configgen depend on environment symbols, so will fail to link in containerized mode without them?
I presume it doesn't hit a build error, but wondering why..

Shouldn't we leave them as is, but instead exclude them building at all, i.e. exclude configgen, configenv etc. in containerized mode?

esp/services/ws_configmgr/CMakeLists.txt
```diff
@@ -69,11 +69,14 @@ target_link_libraries(ws_configmgr
     jlib
     xmllib
     esphttp
-    environment
     dllserver
     configmgr
     )
 
+if (NOT CONTAINERIZED)
```
jakesmith, line 76: can't actually see where ws_configmgr is built/included at all.. the subdir cmake file is not included in esp/services/CMakeLists.txt
 but if is, perhaps better to leave as is, and exclude it in containerized mode (as 'ws_config; already is)

esp/services/ws_dfu/ws_dfuService.cpp
```diff
@@ -100,6 +100,7 @@ short days[12] = {31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31};
 
 CThorNodeGroup* CThorNodeGroupCache::readNodeGroup(const char* _groupName)
 {
+#ifndef _CONTAINERIZED
```
jakesmith, line 103: minor: The whole of CThorNodeGroupCache is bare metal only (used in 1 place). I see the calling code is already conditionally compiled.

esp/services/ws_dfu/ws_dfuService.cpp
```diff
@@ -182,9 +184,14 @@ void CWsDfuEx::init(IPropertyTree *cfg, const char *process, const char *service
 
     setDaliServixSocketCaching(true);
 
+#ifdef _CONTAINERIZED
+    IERRLOG("CONTAINERIZED(CWsDfuEx::init)");
+    maxFileAccessExpirySeconds = defaultMaxFileAccessExpirySeconds;
+#else
     factory.setown(getEnvironmentFactory(true));
     env.setown(factory->openEnvironment());
     maxFileAccessExpirySeconds = serviceTree->getPropInt("@maxFileAccessExpirySeconds", defaultMaxFileAccessExpirySeconds);
```
jakesmith, line 193: this line could be outside of #ifdef (there are many other serviceTree->getXX 's outside).
And I _think_ Tony's impl. transforms converts yaml prop. trees to legacy serviceTree's so this will still work if set into the esp service yaml as a property.

jakesmith (commented):
@ghalliday - looks good, some minor comments/questions

deployment/configgen/CMakeLists.txt
ghalliday: You are correct - they are not built in container mode - I will regress

ghalliday:
I think this is ready for re-review.  It could do with merging sooner rather than later - because various other changes are creating merge conflicts.

esp/services/ws_dfu/ws_dfuService.cpp
jakesmith: @ghalliday - just checking, you didn't want to move this #ifndef around whole of CThorNodeGroupCache at this juncture ?

esp/smc/SMCLib/TpCommon.cpp
```diff
@@ -0,0 +1,99 @@
+/*##############################################################################
+
+    HPCC SYSTEMS software Copyright (C) 2012 HPCC Systems®.
```
jakesmith, line 3: should this be changed to 2021?

esp/smc/SMCLib/TpContainer.cpp
```diff
@@ -0,0 +1,804 @@
+/*##############################################################################
+
+    HPCC SYSTEMS software Copyright (C) 2012 HPCC Systems®.
```
jakesmith, line 3: should this be changed to 2021?

jakesmith (approved):
@ghalliday - a couple of trivial comments, looks good to me.

esp/services/ws_dfu/ws_dfuXRefService.cpp
```diff
@@ -553,6 +553,9 @@ bool CWsDfuXRefEx::onDFUXRefList(IEspContext &context, IEspDFUXRefListRequest &r
 {
     try
     {
+#ifdef _CONTAINERIZED
+        IERRLOG("CONTAINERIZED(CWsDfuXRefEx::onDFUXRefList)");
```
richardkchapman, line 557: Why IERRLOG here but UNIMPLEMENTED below?

esp/smc/SMCLib/TpContainer.cpp
ghalliday: fixed (and above)

richardkchapman:
@ghalliday Not sure what the logic determining whether to report sometimes using IERRLOG, sometimes WARNLOG) versus throwing an exception is?

Other than that, looks ok.

esp/services/ws_dfu/ws_dfuXRefService.cpp
ghalliday: I tried to use UNIMPLEMENTED when the code would have previously failed, and IERRLOG when it would have silently ignored it.

PR #14934, opened 2021-05-07
HPCC-25867 Add metrics tag to metrics log entries

kenrowland:
Added 'METRICS:' tag to metrics log entries

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Cloud-compatibility
  - [x] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jlog.hpp
+++ b/system/jlib/jlog.hpp
@@ -481,6 +481,8 @@ inline char const * msgPrefix(LogMsgClass msgClass)
         return "ERROR: ";
     case MSGCLS_warning:
         return "WARNING: ";
+    case MSGCLS_metric:
+        return "METRICS: ";
     default:
         return "";
     }


ghalliday (approved):
I am fine with the change as long as @rpastrana is.

rpastrana (approved):
@kenrowland I'm ok with this change for the short-term. Eventually we'll want to utilize the "class" column (MSGCLS_metric) to distinguish metric log entries.

richardkchapman:
Is there a reason the change Rodrigo suggests can't be made now?

kenrowland:
If the desire is to use the class output option where MSGCLS_metric adds "cls=MET ", then this PR is not needed. I have seen somewhere, but cannot remember where, that ability to add the class output options. 

@rpastrana - if this is preferred, please indicate such and the PR can be abandoned.

rpastrana:
@ghalliday I think this change is fine.
Yes, the alternative is to add the class field by default in containerized mode (we would presumably no longer need the prefix embedded in the message), but as you point out, that would potentially break pre-existing  bare-metal processes.

richardkchapman:
@rpastrana It would not break existing bare-metal processes if the change was in containerized mode only?

I guess my question is, if now is not the right time to make that change, when will be?

ghalliday:
@richardkchapman this may also tie in with a discussion of costs that I was having with Jake and Shamser.  It might be better to hold off on this change and add the class to the container (and non container) logging for 8.2.

rpastrana:
Agreed, this is a good time to make the change.
We can add MSGFIELD_class and remove MSGFIELD_prefix from the MSGFIELD_STANDARD  definition in code:
https://github.com/hpcc-systems/HPCC-Platform/blob/7c6452e741a6e758a2ea5c16b2ddf08dd4f86aae/system/jlib/jlog.hpp#L313

Users can always override those defaults via config: --set global.logging.fields="STD+CLS"

PR #14939, opened 2021-05-09
HPCC-25783 add dropzone storage example settings.

xwang2713:
These include "auto" and "retained" dropzone examples for local, Azure
and AWS.

NFS support either will be added later or directly added to NFS PR which is in reviewing process now.

To enable auto storage creation "storage.dropzoneStorage" is added to HPCC helm chart schema and values.yaml.
Also it is added to global ConfigMap if no dropzone plane supplied.

Tested on local, Azure and AWS

### DIFF
--- a/helm/examples/local/hpcc-localfile/values.schema.json
+++ b/helm/examples/local/hpcc-localfile/values.schema.json
@@ -7,6 +7,9 @@
       "properties": {
         "hostpath": {
           "type": "string"
+        },
+        "mountPrefix": {
+          "type": "string"
         }
       }
     },
--- a/helm/examples/azure/hpcc-azurefile/templates/NOTES.txt
+++ b/helm/examples/azure/hpcc-azurefile/templates/NOTES.txt
@@ -26,6 +26,10 @@ storage:
   dataStorage:
     plane: {{ $plane.name }}
 {{ end -}}
+{{- if eq "dropzone" $plane.name }}
+  dropzoneStorage:
+    plane: {{ $plane.name }}
+{{ end -}}
 {{ end -}}
 
 {{ range $plane := .Values.planes -}}
--- a/helm/examples/azure/hpcc-azurefile/values.yaml
+++ b/helm/examples/azure/hpcc-azurefile/values.yaml
@@ -26,3 +26,9 @@ planes:
   labels: [ "data" ] # NB: all "data" planes will be auto mounted by engine components and others that require access to data
   rwmany: true
   #sku: "Standard_LRS"
+- name: dropzone
+  subPath: dropzone # cannot currently be changed
+  size: 3Gi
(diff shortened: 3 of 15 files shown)


jakesmith (commented):
I think it looks ok.
Not entirely sure we need it though.
You can define an arbitrary number of planes with labels already in the example charts,
and dfuserver should automatically pick up all that are labeled as 'lz' and use them as drop zones.

This change/PR implies there's should only be one (by dropzoneStorage:), a bit like dataStorage: which we're trying to move away from I think, with planes: and engines default storage planes being the primary way you define storage.

However, planes need to point to a pvc, where as *Storage: can point to a storageClass only and auto-instantiate ephemeral storage.

I suspect we should not be adding this to the main hpcc chart, or extending the example.
I think a better route (if anything) would be to add an example that creates some PVC from storage class (includnig "" for ephemeral). Then they could be used by planes direclty.

@ghalliday - my have some thoughts on this.

ghalliday:
I don't really understand this change for a few reasons.

* As jake says the helm charts already support multiple drop zones by adding a 'lz' label to a plane.  I don't think the main helm charts need any change.
* The main reason for a drop zone is so that an external process can copy files to and from the drop zone.  That means they must be externally accessible.  I don't think there is any point in creating an implicit one because it can't be accessed externally.

Have I missed something?

The following might be useful as examples:
* azure blob storage account - users can directly copy to blob storage
* A legacy bare metal storage dropzone (with a dafilesrv running on it)
* Possibly an implicit drop zone for the local file example

I think they would need to be examples to be pasted into an existing values.yaml, not something that can automatically be combined.

ghalliday (changes requested):
See comments above

xwang2713:
Let me try adding plane with label "lz" for ephemeral  storage. I am not sure it will work without same change.

ghalliday:
Following on from yesterdays discussion, it will make sense to have an ephemeral dropzone defined in our standard values.yaml - it would allow drop zone functionality to be tested and files could be uploaded using the esp upload function.

It is possible to define a bare metal drop zone, by using a storage plane that is based on a hostGroup.

dfuplus currently has to use srcip=127.0.0.1 srcfilename=<plane-prefix>/<relative-of-file> nolocal=1, which is ugly.  See HPCC-25910 to improve this.

helm/examples/local/hpcc-localfile/values.schema.json
```diff
@@ -7,6 +7,9 @@
       "properties": {
         "hostpath": {
           "type": "string"
+        },
+        "mountPrefix": {
```
ghalliday, line 11: unused.

xwang2713:
With gavin's change for dropzone in storage examples this PR is not needed.

PR #14948, opened 2021-05-11
HPCC-25845 Fix problems publishing files to roxie in container mode

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dali/dfu/dfuutil.cpp
+++ b/dali/dfu/dfuutil.cpp
@@ -69,7 +69,8 @@ static bool physicalPartCopy(IFile *from,const char *tofile, Owned<IException> &
                 0,
                 (offset_t)-1, // all file
                 NULL,
-                PHYSICAL_COPY_POLL_TIME)) {
+                PHYSICAL_COPY_POLL_TIME,
+                CFflush_rdwr)) {
             // Abort check TBD
         }
     }
@@ -1265,11 +1266,17 @@ public:
     {
         DBGLOG("cloneRoxieSubFile src=%s@%s, dst=%s@%s, prefix=%s, ow=%d, docopy=false", srcLFN, srcCluster, dstLFN, dstCluster, prefix, overwriteFlags);
         CFileCloner cloner;
-        cloner.init(dstCluster, DFUcpdm_c_replicated_by_d, true, NULL, userdesc, foreigndali, NULL, NULL, false, false);
+        bool copyPhysical = false;
+        // MORE: Would the following be better to ensure files are copied when queries are deployed?
+        // bool copyPhysical = isContainerized() && (foreigndali != nullptr);
+        cloner.init(dstCluster, DFUcpdm_c_replicated_by_d, true, NULL, userdesc, foreigndali, NULL, NULL, false, copyPhysical);
         cloner.overwriteFlags = overwriteFlags;
+#ifndef _CONTAINERIZED
+        //In containerized mode there is no need to replicate files to the local disks of the roxie cluster - so don't set the special flag
         cloner.spec1.setRoxie(redundancy, channelsPerNode, replicateOffset);
         if (defReplicateFolder)
             cloner.spec1.setDefaultReplicateDir(defReplicateFolder);
+#endif
         cloner.srcCluster.set(srcCluster);
         cloner.prefix.set(prefix);
         cloner.cloneRoxieFile(srcLFN, dstLFN);
--- a/esp/services/ws_packageprocess/ws_packageprocessService.cpp
+++ b/esp/services/ws_packageprocess/ws_packageprocessService.cpp
@@ -31,6 +31,7 @@
 #include "exception_util.hpp"
 #include "TpWrapper.hpp"
 
+
 void CWsPackageProcessEx::init(IPropertyTree *cfg, const char *process, const char *service)
 {
     packageMapAndSet.subscribe();
@@ -136,6 +137,16 @@ void cloneFileInfoToDali(unsigned updateFlags, StringArray &notFound, IPropertyT
 
     Owned<IReferencedFileList> wufiles = createReferencedFileList(user, password, allowForeignFiles, false);
     wufiles->addFilesFromPackageMap(packageMap);
+
+    Owned<IDFUhelper> helper = createIDFUhelper();
+#ifdef _CONTAINERIZED
+    SCMStringBuffer clusterName;
+    dstInfo->getName(clusterName);
+    StringBuffer targetPlane;
+    getRoxieDefaultPlane(targetPlane, clusterName.str());
+    wufiles->resolveFiles(targetPlane, lookupDaliIp, remotePrefix, srcCluster, !(updateFlags & (DALI_UPDATEF_REPLACE_FILE | DALI_UPDATEF_CLONE_FROM)), false, false);
+    wufiles->cloneAllInfo(updateFlags, helper, true, false, 0, 1, 0, nullptr);
+#else
     SCMStringBuffer processName;
     dstInfo->getRoxieProcess(processName);
     wufiles->resolveFiles(processName.str(), lookupDaliIp, remotePrefix, srcCluster, !(updateFlags & (DALI_UPDATEF_REPLACE_FILE | DALI_UPDATEF_CLONE_FROM)), false, false);
@@ -143,8 +154,8 @@ void cloneFileInfoToDali(unsigned updateFlags, StringArray &notFound, IPropertyT
     StringBuffer defReplicateFolder;
     getConfigurationDirectory(NULL, "data2", "roxie", processName.str(), defReplicateFolder);
 
-    Owned<IDFUhelper> helper = createIDFUhelper();
     wufiles->cloneAllInfo(updateFlags, helper, true, false, dstInfo->getRoxieRedundancy(), dstInfo->getChannelsPerNode(), dstInfo->getRoxieReplicateOffset(), defReplicateFolder);
+#endif
 
     Owned<IReferencedFileIterator> iter = wufiles->getFiles();
     ForEach(*iter)
--- a/esp/services/ws_workunits/ws_workunitsQuerySets.cpp
+++ b/esp/services/ws_workunits/ws_workunitsQuerySets.cpp
@@ -753,7 +753,7 @@ public:
         if (!process.length())
             return;
 #else
-        StringBuffer process(target);
+        process.set(target);
 #endif
 
         ps.setown(createPackageSet(process.str()));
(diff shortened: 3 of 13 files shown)


dali/dfu/dfuutil.cpp
```diff
@@ -1265,11 +1266,17 @@ class CDFUhelper: implements IDFUhelper, public CInterface
     {
         DBGLOG("cloneRoxieSubFile src=%s@%s, dst=%s@%s, prefix=%s, ow=%d, docopy=false", srcLFN, srcCluster, dstLFN, dstCluster, prefix, overwriteFlags);
         CFileCloner cloner;
-        cloner.init(dstCluster, DFUcpdm_c_replicated_by_d, true, NULL, userdesc, foreigndali, NULL, NULL, false, false);
+        bool copyPhysical = false;
+        // MORE: Would the following be better to ensure files are copid when queries are deployed?
```
richardkchapman, line 1270: Typo in comment

richardkchapman (dismissed):

richardkchapman:
One trivial comment. I will wait for others to review too...

mckellyln (dismissed):
Approved

esp/services/ws_workunits/ws_workunitsQuerySets.cpp
```diff
@@ -2952,6 +2962,7 @@ class QueryCloner
                 SCMStringBuffer process;
                 StringBuffer defReplicateFolder;
                 getConfigurationDirectory(NULL, "data2", "roxie", cl->getRoxieProcess(process).str(), defReplicateFolder);
+                //MORE: When is this called, and what should it do if _CONTAINERIZED
```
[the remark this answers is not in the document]
afishbeck: @ghalliday yes, I'm sure it does.  QueryFileCopier copies files for a single query being published, copied, or recreated.  QueryCloner deals with copying or importing entire querysets.. if the file copying flag is set then it gathers the deduped list of files needed by all of the queries, then resolves and clones them using the same methods used above.

dali/dfu/dfuutil.cpp
afishbeck: I think we considered this for baremetal but the roxie copy mechanism was already working and seemingly preferred.  It's definitely worth considering for cloud.. although I'm not yet familiar enough with when we'd want to copying to storage planes to say for sure.  And I've also never tested the flag to see if it actually works.

ghalliday: I tested the code when trying to debug the issue, and it did work.  It might be worth adding a flag to the service to indicate that physical files should be copied.

esp/services/ws_workunits/ws_workunitsQuerySets.cpp
```diff
@@ -766,11 +766,21 @@ class QueryFileCopier
         if (queryname && *queryname)
             queryname = queryid.append(queryname).append(".0").str(); //prepublish dummy version number to support fuzzy match like queries="myquery.*" in package
         files->addFilesFromQuery(cw, pm, queryname);
+        Owned<IDFUhelper> helper = createIDFUhelper();
+#ifdef _CONTAINERIZED
+        //Find the first data plane
+        Owned<IPropertyTreeIterator> dataPlanes = queryGlobalConfig().getElements("storage/planes[labels='data']");
+        if (!dataPlanes->first())
+            throwUnexpectedX("No default data plane defined");
```
ghalliday, line 774: What happens if the file is already local on the target plane?  Should this not be called?  Will this all work fine if it is already there - I need to test.

dali/dfu/dfuutil.cpp
jakesmith: I'm not familiar with how it normally works..
How does roxie see the physical files if they're copyPhysical=false , implying they are not copied at all?
Does that mean they are remotely accessed at the original source location?

esp/services/ws_workunits/ws_workunitsQuerySets.cpp
```diff
@@ -766,11 +766,21 @@ class QueryFileCopier
         if (queryname && *queryname)
             queryname = queryid.append(queryname).append(".0").str(); //prepublish dummy version number to support fuzzy match like queries="myquery.*" in package
         files->addFilesFromQuery(cw, pm, queryname);
+        Owned<IDFUhelper> helper = createIDFUhelper();
+#ifdef _CONTAINERIZED
+        //Find the first data plane
```
jakesmith, line 771: shouldn't the target plane be configurable?
Isn't the 1st plane likely not to be the right one?

esp/services/ws_workunits/ws_workunitsQuerySets.cpp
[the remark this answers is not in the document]
jakesmith: looks like this one is a needed fix

fs/dafsclient/rmtfile.cpp
```diff
@@ -1789,12 +1789,14 @@ extern bool clientAsyncCopyFileSection(const char *uuid,
...
                         ICopyFileProgress *progress,
-                        unsigned timeout)       // returns true when done
+                        unsigned timeout,
+                        CFflags copyFlags)       // returns true when done
 {
     CRemoteFile *cfile = QUERYINTERFACE(from,CRemoteFile);
-    if (!cfile) {
-        // local - do sync
-        from->copySection(to,toOfs,fromOfs,size,progress);
+    if (!cfile || to.isLocal()) {
```
jakesmith, line 1796: curious if the addition of 'to.isLocal()' was actually necessary? i.e. when and by what, for a local file did a CRemoteFile get created?

jakesmith (commented):
@ghalliday - others prob. know better, but am not sure what it means for copyPhysical to be false.
Should the plane it's copying to default to this Roxie's default storage plane, and be configurable?

This probably needs rebasing/retargeting to 8.2/master

fs/dafsclient/rmtfile.cpp
ghalliday: This was triggered when I enabled copyPhysical for the roxie copies.  The files were being copied from a remote cluster to this cluster - where the filenames are local.
Without this check the code executes the copy command on the remote dafilesrv, which then tries to connect to the dafilesrv on the k8s cluster - and then fails.   If there was a dafilesrv running on the cluster it would probably have worked (if the firewall allowed)..  For that reason alone it is better to pull than push.

esp/services/ws_workunits/ws_workunitsQuerySets.cpp
ghalliday: yes it should.  I will open that as a separate jira.  This is struggling to get it to work at all.

esp/services/ws_workunits/ws_workunitsQuerySets.cpp
ghalliday: tested and seems to work

esp/services/ws_workunits/ws_workunitsQuerySets.cpp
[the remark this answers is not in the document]
ghalliday: I will release another commit with this fixed

ghalliday:
I'm not sure about the target.  This is needed for current testing, and I think it should only affect the containerized version

dali/dfu/dfuutil.cpp
afishbeck: In baremetal depending on Roxies configuration it will copy the files itself in the background and access the files at the original source location until the copy is complete.  Roxie can also be configured to always access the files remotely, or to expect the files to have been copied through DFS.

afishbeck:
Maybe outside the scope of this PR, but for the resolveFiles step, what if some of the query files already exist on file planes other than the default?  Can the LFN specify the storage plane or is that always flexible/mapped?  Should packagemaps be able to specify storage planes?

afishbeck:
Sorry if I'm asking obvious questions I haven't looked closely at the storage plane implementation yet.

esp/services/ws_packageprocess/ws_packageprocessService.cpp
```diff
@@ -31,6 +31,25 @@
 #include "exception_util.hpp"
 #include "TpWrapper.hpp"
 
+
+static StringBuffer & getRoxieDefaultPlane(StringBuffer & plane, const char * roxieName)
```
ghalliday, line 35: NOTE: This function should be commoned up.  Tony has suggested a file it can probably live in - I will try moving it.

ghalliday:
@kturlapathi you may be interested in these fixes.

esp/services/ws_packageprocess/ws_packageprocessService.cpp
ghalliday: now moved and extended to allow the plane to be configured

ghalliday:
> Maybe outside the scope of this PR, but for the resolveFiles step, what if some of the query files already exist on file planes other than the default? Can the LFN specify the storage plane or is that always flexible/mapped? Should packagemaps be able to specify storage planes?

I think it would be good if esp supported this, but the current code isn't structured to make that very easy.  Ideally roxie would have a list of acceptable storage planes, or the planes could be tagged with a label to indicate they were suitable to use with roxie data.

richardkchapman (dismissed):

esp/services/ws_workunits/ws_workunitsQuerySets.cpp
jakesmith: @ghalliday - was a JIRA opened for this?

jakesmith (dismissed):
Looks good afaict.

esp/services/ws_workunits/ws_workunitsQuerySets.cpp
ghalliday: It was implemented by the last commit - it now takes it from the roxie storagePlane if defined.

jakesmith: ok thanks

helm/hpcc/values.schema.json
```diff
@@ -763,6 +763,11 @@
           "description": "The default storage plane to write data files to",
           "type": "string"
         },
+        "storagePlanes": {
```
ghalliday, line 766: @richardkchapman this isn't actually used - should I remove it, or leave it there for use by HPCC-25930?

richardkchapman: It's fine to leave.

richardkchapman (approved):

PR #14955, opened 2021-05-13
HPCC-25744 Prometheus metrics sink

rpastrana:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/metrics/sinks/prometheus/CMakeLists.txt
+++ b/system/metrics/sinks/prometheus/CMakeLists.txt
@@ -1,4 +1,4 @@
-###############################################################################
+################################################################################
 #    HPCC SYSTEMS software Copyright (C) 2021 HPCC Systems®.
 #
 #    Licensed under the Apache License, Version 2.0 (the "License");
@@ -14,6 +14,26 @@
 #    limitations under the License.
 ################################################################################
 
-HPCC_ADD_SUBDIRECTORY (file)
-HPCC_ADD_SUBDIRECTORY (log)
-HPCC_ADD_SUBDIRECTORY (prometheus)
+project(hpccmetrics_prometheus)
+
+set ( srcs
+      prometheusSink.cpp
+    )
+
+set ( INCLUDES
+      ${HPCC_SOURCE_DIR}/system/httplib/httplib.h
+    )
+
+include_directories(
+        ${HPCC_SOURCE_DIR}/system/include
+        ${HPCC_SOURCE_DIR}/system/jlib
+        ${HPCC_SOURCE_DIR}/system/httplib
+)
+
+ADD_DEFINITIONS(  -DPROMETHEUSSINK_EXPORTS )
+HPCC_ADD_LIBRARY( hpccmetrics_prometheussink SHARED ${srcs} )
+TARGET_LINK_LIBRARIES( hpccmetrics_prometheussink
+                       jlib
+                     )
+
+INSTALL ( TARGETS hpccmetrics_prometheussink RUNTIME DESTINATION ${EXEC_DIR} LIBRARY DESTINATION ${LIB_DIR} )
--- a/system/metrics/sinks/prometheus/prometheusSink.cpp
+++ b/system/metrics/sinks/prometheus/prometheusSink.cpp
@@ -1,0 +1,148 @@
+/*################################################################################
+#    HPCC SYSTEMS software Copyright (C) 2021 HPCC Systems®.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+################################################################################
+*/
+
+#include "jlog.hpp"
+#include "prometheusSink.hpp"
+
+using namespace hpccMetrics;
+
+extern "C" MetricSink* getSinkInstance(const char *name, const IPropertyTree *pSettingsTree)
+{
+    return new PrometheusMetricSink(name, pSettingsTree);
+}
+
+PrometheusMetricSink::PrometheusMetricSink(const char *name, const IPropertyTree *pSettingsTree) :
+    MetricSink(name, PROMETHEUS_REPORTER_TYPE)
+{
+    m_metricsReporter = nullptr;
+    m_metricsSinkName = name;
+    m_processing = false;
+    m_port = DEFAULT_PROMETHEUS_METRICS_SERVICE_PORT;
+    m_verbose = false;
+
+    if (pSettingsTree)
+    {
+        m_verbose = pSettingsTree->getPropBool("@verbose", false);
+        m_port = pSettingsTree->getPropInt("@port", DEFAULT_PROMETHEUS_METRICS_SERVICE_PORT);
+
+        pSettingsTree->getProp("@serviceName", m_metricsServiceName);
+
+        if (m_metricsServiceName.isEmpty())
+            m_metricsServiceName = DEFAULT_PROMETHEUS_METRICS_SERVICE_NAME;
+        else
+            if (m_metricsServiceName.charAt(0) != '/')
+                m_metricsServiceName.insert(0, '/');
+
+
+        m_server.set_error_handler([](const Request& req, Response& res)
+        {
+            StringBuffer msg(detail::status_message(res.status));
+            if (res.status == 500)
+            {
+                if (res.has_header(HTTPLIB_ERROR_MESSAGE_HEADER_NAME))
+                    msg.append(" - ").append(res.get_header_value(HTTPLIB_ERROR_MESSAGE_HEADER_NAME).c_str());
+                else
+                    msg.append(" - ").append("encountered unknown error!");
+
+                LOG(MCinternalError, "PrometheusMetricsService: %s", msg.str());
+            }
+
+            VStringBuffer respmessage(PROMETHEUS_METRICS_HTTP_ERROR, msg.str(), req.path.c_str(), res.status);
+            res.set_content(respmessage.str(), PROMETHEUS_METRICS_SERVICE_RESP_TYPE);
+
+            LOG(MCuserInfo, "TxSummary[status=%d;user=@%s:%d;contLen=%ld;req=%s;]\n", res.status, req.remote_addr.c_str(), req.remote_port, req.content_length, req.method.c_str());
+        });
+
+        m_server.Get(m_metricsServiceName.str(), [&](const Request& req, Response& res)
+        {
+            StringBuffer payload;
+            if (!m_metricsReporter)
+                throw std::runtime_error("NULL MetricsReporter detected!");
+            else
+                toPrometheusMetrics(m_metricsReporter->queryMetricsForReport(std::string(m_metricsSinkName.str())), payload, m_verbose);
+
+            res.set_content(payload.str(), PROMETHEUS_METRICS_SERVICE_RESP_TYPE);
+            LOG(MCuserInfo, "TxSummary[status=%d;user=@%s:%d;contLen=%ld;req=GET;]\n",res.status, req.remote_addr.c_str(), req.remote_port, req.content_length);
+        });
+    }
+}
+
+const char * PrometheusMetricSink::mapHPCCMetricTypeToPrometheusStr(MetricType type)
+{
+    switch (type)
+    {
+    case hpccMetrics::METRICS_COUNTER:
+        return "counter";
+    case hpccMetrics::METRICS_GAUGE:
+        return "gauge";
+    default:
+        LOG(MCdebugInfo, "Encountered unknown metric - cannot map to Prometheus metric!");
+        return nullptr;
+    }
+}
+
+void PrometheusMetricSink::toPrometheusMetrics(std::vector<std::shared_ptr<IMetric>> reportMetrics, StringBuffer & out, bool verbose)
+{
+    /*
+     * [# HELP <metric name> <metric summary>\n]
+     * [# TYPE <metric name> <metric type, if missing, metric is un-typed>\n]
+     * <metric name> [{<label name>=<label value>, ...}] <metric value>\n
+     *
+     * where metric name := [a-zA-Z_:][a-zA-Z0-9_:]*
+     */
+
+    for (auto &pMetric: reportMetrics)
+    {
+        const std::string & name = pMetric->queryName();
+        if (verbose)
+        {
+            if (!pMetric->queryDescription().empty())
+                out.append("# HELP ").append(name.c_str()).append(" ").append(pMetric->queryDescription().c_str()).append("\n");
+
+            MetricType type = pMetric->queryMetricType();
+            const char * promtype = mapHPCCMetricTypeToPrometheusStr(type);
+            if (promtype)
+                out.append("# TYPE ").append(name.c_str()).append(" ").append(promtype).append("\n");
+        }
+
+        out.append(name.c_str()).append(" ").append(pMetric->queryValue()).append("\n");
+    }
+}
+
+void PrometheusMetricSink::startCollection(MetricsReporter *_pReporter)
+{
+    if (!_pReporter)
+        throw MakeStringException(-1, "PrometheusMetricsService: NULL MetricsReporter detected!");
+
+    m_metricsReporter = _pReporter;
+
+    m_collectThread = std::thread(collectionThread, this);
+    m_processing = true;
+}
+
+void PrometheusMetricSink::stopCollection()
+{
+    m_processing = false;
+    m_server.stop();
+    m_collectThread.join();
+}
+
+void PrometheusMetricSink::startServer()
+{
+    LOG(MCoperatorProgress, "PrometheusMetricsService started:  port: '%i' uri: '%s' sinkname: '%s'\n", m_port, m_metricsServiceName.str(), m_metricsSinkName.str());
+    m_server.listen("localhost", m_port);
+}
--- a/system/metrics/sinks/prometheus/prometheusSink.hpp
+++ b/system/metrics/sinks/prometheus/prometheusSink.hpp
@@ -1,0 +1,102 @@
+/*################################################################################
+#    HPCC SYSTEMS software Copyright (C) 2021 HPCC Systems®.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+################################################################################
+*/
+
+
+#pragma once
+
+#include "jmetrics.hpp"
+#include "jptree.hpp"
+#include "jstring.hpp"
+#include <thread>
+
+//including cpp-httplib single header file REST client
+//  doesn't work with format-nonliteral as an error
+//
+#if defined(__clang__) || defined(__GNUC__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wformat-nonliteral"
+#endif
+
+#undef INVALID_SOCKET
+#include "httplib.h"
+
+#if defined(__clang__) || defined(__GNUC__)
+#pragma GCC diagnostic pop
+#endif
+
+#ifdef _USE_OPENSSL
+#include <openssl/x509v3.h>
+#endif
+
+using namespace hpccMetrics;
+using namespace httplib;
+
+#ifdef PROMETHEUSSINK_EXPORTS
+#define PROMETHEUSSINK_API DECL_EXPORT
+#else
+#define PROMETHEUSSINK_API DECL_IMPORT
+#endif
+
+class PROMETHEUSSINK_API PrometheusMetricSink : public MetricSink
+{
+public:
+    explicit PrometheusMetricSink(const char *name, const IPropertyTree *pSettingsTree);
+    ~PrometheusMetricSink() override = default;
+    void startServer();
+private:
+    std::thread       m_collectThread;
+    MetricsReporter * m_metricsReporter;
+    StringBuffer      m_metricsSinkName;
+
+    std::atomic<bool>               m_processing{false};
+    int                             m_port;
+    StringBuffer                    m_metricsServiceName;
+    bool                            m_verbose;
+
+    static constexpr const char * PROMETHEUS_REPORTER_TYPE = "prometheus";
+    static constexpr const char * PROMETHEUS_METRICS_HTTP_PAGE_TITLE = "HPCC Systems - Prometheus Metrics Service";
+    static constexpr const char * PROMETHEUS_METRICS_SERVICE_RESP_TYPE = "text/html; charset=UTF-8";
+    static constexpr int          DEFAULT_PROMETHEUS_METRICS_SERVICE_PORT = 8767;
+    static constexpr const char * DEFAULT_PROMETHEUS_METRICS_SERVICE_NAME = "/metrics";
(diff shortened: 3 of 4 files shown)


rpastrana:
@kenrowland take a look at a high level

system/metrics/sinks/prometheus/prometheusReporterService.cpp
```diff
@@ -0,0 +1,354 @@
...
+void PrometheusReporterService::toPrometheusMetric(std::vector<std::shared_ptr<IMetric>> reportMetrics, StringBuffer & out)
+{
+    for (auto &pMetric: reportMetrics)
+    {
+        std::string name = pMetric->queryName();
+        if (!pMetric->queryDescription().empty())
+            out.append("# HELP ").append(name.c_str()).append(" ").append(pMetric->queryDescription().c_str()).append("\n");
+
+        MetricType type = pMetric->queryMetricType();
+        if (mapMockMetricTypeToPrometheusStr(type))
```
kenrowland, line 224: Appears you do the conversion of the type twice. What if the conversion routine just returned something like "custom" and you reported the value as such. That way you'd always report all values. Not sure makes sense.

system/metrics/sinks/prometheus/prometheusReporterService.cpp
```diff
@@ -0,0 +1,354 @@
...
+            out.append("# TYPE ").append(name.c_str()).append(" ").append(mapMockMetricTypeToPrometheusStr(type)).append("\n");
+
+        out.append(name.c_str()).append(" ").append(std::to_string(pMetric->queryValue()).c_str()).append("\n");
+
+        LOG(MCmetrics, "name=%s,value=%" I64F "d", pMetric->queryName().c_str(), pMetric->queryValue());
```
kenrowland, line 229: Not sure if logging is appropriate or not. I feel that if you want metrics output in the log as well that you would load the log sink in addition to the prometheus sink.

system/metrics/sinks/prometheus/prometheusReporterService.cpp
```diff
@@ -0,0 +1,354 @@
...
+}
+
+bool PrometheusReporterService::onMetrics(CHttpRequest * request, CHttpResponse * response)
+{
+    StringBuffer payload;
+    if (!m_metricsReporter)
+        onMetricsUnavailable(request, response);
+    else
+    {
+        toPrometheusMetric(m_metricsReporter->queryMetricsForReport(std::string(m_metricsSinkName)), payload);
```
kenrowland, line 240: Trivial, but toPrometheusMetric implies a single metric is handled when it's actually a vector.

system/metrics/sinks/prometheus/prometheusReporterService.cpp
```diff
@@ -0,0 +1,354 @@
...
+        default:
+            return nullptr;
+    }
+}
+
+void PrometheusReporterService::toPrometheusMetric(std::vector<std::shared_ptr<IMetric>> reportMetrics, StringBuffer & out)
+{
+    for (auto &pMetric: reportMetrics)
+    {
+        std::string name = pMetric->queryName();
```
kenrowland, line 219: You may be able to make this a const reference to avoid a copy.

kenrowland (commented):
Please take a look at my comments and see if any warrant a change.

system/metrics/sinks/prometheus/prometheusReporterService.cpp
rpastrana: Agreed, this was mostly for my benefit during debugging. BTW is it possible to have multiple sinks concurrently?

rpastrana:
@kenrowland made changes, please perform full review.

rpastrana:
Hi @AttilaVamos I'm not sure what's going on w/ the checks on this PR, but I assume the failures are false negatives

system/metrics/sinks/prometheus/prometheusReporterService.cpp
```diff
@@ -0,0 +1,359 @@
...
+            WARNLOG("Encountered unknown metric - cannot map to Prometheus metric!");
+            return nullptr;
+    }
+}
+
+void PrometheusReporterService::toPrometheusMetrics(std::vector<std::shared_ptr<IMetric>> reportMetrics, StringBuffer & out)
+{
+    for (auto &pMetric: reportMetrics)
+    {
+        const std::string name = pMetric->queryName();
```
kenrowland, line 220: In all actuality, testing for this value being empty is not necessary, all metrics added to a reporter are required to have a unique name. The value will never be empty.

system/metrics/sinks/prometheus/prometheusReporterService.cpp
kenrowland: The framework allows multiple sinks and treats them independently.

kenrowland (dismissed):
Looks fine. I would make the one last change to remove the check for an empty metric name. I should have caught that last review. After that a merge is fine. I shall mark approved.

AttilaVamos:
@rpastrana That error in teststdlibrary (tmod=DataPatterns) is a strange one and it caused by a "Disk Full" error. I saw there is a kind of test code as well. Does that using extensive disk space?
I rescheduled this PR to see whether that problem is persistent.

rpastrana:
@AttilaVamos looks like the tests have passed now, thanks.

rpastrana:
@kenrowland made the last change you recommended, but I also want to add the helm chart configuration schema for this sink in this PR

system/metrics/sinks/prometheus/prometheusReporterService.cpp
```diff
@@ -0,0 +1,354 @@
...
+        settings->getProp("@serviceName", m_metricsServiceName);
+        if (m_metricsServiceName.isEmpty())
+            m_metricsServiceName = DEFAULT_PROMETHEUS_METRICS_SERVICE_NAME;
+
+        if(m_use_ssl)
+        {
+               m_sslconfig = settings->getBranch("sslConfig");
+#ifdef _USE_OPENSSL
+            if(m_sslconfig != nullptr)
+                m_ssctx.setown(createSecureSocketContextEx2(m_sslconfig, ServerSocket));
```
rpastrana, line 52: @afishbeck just realized the createSecureSocketContextEx2 API operates on a "legacy" style config tree and wondered if we should be doing something special to tie into the containerized secrets management. 

sslConfig :
            cipherList:  !el "ECDH+AESGCM:DH+AESGCM:ECDH+AES256:DH+AES256:ECDH+AES128:DH+AES:ECDH+3DES:DH+3DES:RSA+AESGCM:RSA+AES:RSA+3DES:!aNULL:!MD5"
            passphrase: !el someencodedpassphrase
            certificate: !el /home/hpcc/certificate/key.pem
            privatekey: !el /home/hpcc/certificate/public.key.pem
            verify:
            - enable: true
              address_match: true
              accept_selfsigned: true
              trusted_peers: !el a|b|c
              ca_certificates:
              - path: /home/hpcc/certificate

system/metrics/sinks/prometheus/prometheusReporterService.cpp
```diff
@@ -0,0 +1,354 @@
...
+			out.append("# TYPE ").append(name.c_str()).append(" ").append(promtype).append("\n");
+
+		out.append(name.c_str()).append(" ").append(std::to_string(pMetric->queryValue()).c_str()).append("\n");
+    }
+}
+
+bool PrometheusReporterService::onMetrics(CHttpRequest * request, CHttpResponse * response)
+{
+    StringBuffer payload;
+    if (!m_metricsReporter)
```
ghalliday, line 236: This should be an internal error, caught earlier.  It shouldn't be possible to start the sink without a valid place to get metrics.

system/metrics/sinks/prometheus/prometheusReporterService.cpp
```diff
@@ -0,0 +1,354 @@
...
+
+        response->send();
+    }
+
+    return true;
+}
+
+bool PrometheusReporterService::onPost(CHttpRequest * request, CHttpResponse * response)
+{
+    IEspContext& context = *request->queryContext();
```
ghalliday, line 253: unused - should be caught by -Wall=error

system/metrics/sinks/prometheus/prometheusReporterService.cpp
```diff
@@ -0,0 +1,354 @@
...
+
+    response->setVersion(HTTP_VERSION);
+    response->addHeader("Expires", "0");
+    response->setStatus(HTTP_STATUS_OK);
+    response->setContentType(DEFAULT_PROMETHEUS_METRICS_SERVICE_RESP_TYPE);
+
+    IEspContext& context = *request->queryContext();
+    const char * serviceName = context.queryServiceName(nullptr); //why does this method require in param?
+
+    if (!stricmp(serviceName, m_metricsServiceName.str()))
```
ghalliday, line 271: trivial  if (strieq()) is clearer to read.

system/metrics/sinks/prometheus/prometheusReporterService.cpp
```diff
@@ -0,0 +1,354 @@
...
+
+        //EspAuthState authState=authUnknown;
+        sub_service stype = sub_serv_unknown;
+
+        StringBuffer pathEx;
+        StringBuffer serviceName;
+        StringBuffer methodName;
+
+        //perhaps this is too ESP specific???
+        request->getEspPathInfo(stype, &pathEx, &serviceName, &methodName, false);
```
ghalliday, line 324: I don't think you should be calling any function with esp in the name.

system/metrics/sinks/prometheus/prometheusReporterService.cpp
```diff
@@ -0,0 +1,354 @@
...
+    return onGetNotFound(response, serviceName);
+}
+
+void PrometheusReporterService::handleRequest(ISocket* clientSocket)
+{
+    Owned<CHttpRequest>         request;
+    Owned<CHttpResponse>        response;
+
+    request.setown(new CHttpRequest(*clientSocket));
+    IEspContext* ctx = createEspContext(createHttpSecureContext(request.get()));
```
ghalliday, line 283: References to IEspContext seem wrong.  This code is not running within esp, and that looks like a significant chunk of functionality being pulled in.

system/metrics/sinks/prometheus/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+        ${HPCC_SOURCE_DIR}/esp/clients
+        ${HPCC_SOURCE_DIR}/esp/bindings/SOAP/xpp
+        ${HPCC_SOURCE_DIR}/system/xmllib
+)
+
+ADD_DEFINITIONS(  -DPROMETHEUSSINK_EXPORTS )
+HPCC_ADD_LIBRARY( hpccmetrics_prometheussink SHARED ${srcs} )
+TARGET_LINK_LIBRARIES( hpccmetrics_prometheussink
+                       jlib
+                       esphttp
```
ghalliday, line 46: I think esphttp is pulling in much too much code:

   ${XALAN_LIBRARIES}
    ${XERCES_LIBRARIES}
    jlib
    xmllib
    dalibase
    thorhelper      # this should not be dependent on the engine helper dll
    workunit        # neither should it be dependent on the wokunit dll
 
I think it will need a lighter weight library to process the requests.  @afishbeck any suggestions?

system/metrics/sinks/prometheus/prometheusReporterService.cpp
```diff
@@ -0,0 +1,354 @@
...
+const char * mapMockMetricTypeToPrometheusStr(MetricType type)
+{
+    switch (type)
+    {
+        case hpccMetrics::METRICS_COUNTER:
+            return "counter";
+        case hpccMetrics::METRICS_GAUGE:
+            return "gauge";
+        default:
+            WARNLOG("Encountered unknown metric - cannot map to Prometheus metric!");
```
ghalliday, line 211: This could potentially generate a lot of logging.

system/metrics/sinks/prometheus/prometheusReporterService.cpp
```diff
@@ -0,0 +1,354 @@
...
+            return nullptr;
+    }
+}
+
+void PrometheusReporterService::toPrometheusMetrics(std::vector<std::shared_ptr<IMetric>> reportMetrics, StringBuffer & out)
+{
+    for (auto &pMetric: reportMetrics)
+    {
+        const std::string name = pMetric->queryName();
+        if (!pMetric->queryDescription().empty())
```
ghalliday, line 221: Should the #HELP only be output the first time?  Should that be an option?  Same for the # TYPE

system/metrics/sinks/prometheus/prometheusReporterService.cpp
```diff
@@ -0,0 +1,354 @@
...
+{
+    StringBuffer payload;
+    if (!m_metricsReporter)
+        onMetricsUnavailable(request, response);
+    else
+    {
+        toPrometheusMetrics(m_metricsReporter->queryMetricsForReport(std::string(m_metricsSinkName)), payload);
+
+        response->setContent(payload.length(), payload.str());
+        response->setContentType(DEFAULT_PROMETHEUS_METRICS_SERVICE_RESP_TYPE);
```
ghalliday, line 243: minor: Clearer without the DEFAULT_ prefix - unless it can be changed.

system/metrics/sinks/prometheus/prometheusReporterService.cpp
```diff
@@ -0,0 +1,354 @@
...
+            WARNLOG("Encountered unknown metric - cannot map to Prometheus metric!");
+            return nullptr;
+    }
+}
+
+void PrometheusReporterService::toPrometheusMetrics(std::vector<std::shared_ptr<IMetric>> reportMetrics, StringBuffer & out)
+{
+    for (auto &pMetric: reportMetrics)
+    {
+        const std::string name = pMetric->queryName();
```
ghalliday, line 220: Should be a const & otherwise it clones the string each time

ghalliday (changes requested):
@rpastrana some initial comments.

system/metrics/sinks/prometheus/prometheusReporterService.cpp
```diff
@@ -0,0 +1,354 @@
...
+    m_processing = false; //this might need to be guarded
+    return 0;
+}
+
+bool PrometheusReporterService::isProcessing()
+{
+    return m_processing;
+}
+
+int PrometheusReporterService::start(MetricsReporter * pReporter)
```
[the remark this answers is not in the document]
kenrowland: @ghalliday One of the reasons I hold off on giving the reporter to the sink until collection starts is that sink init takes place before the component has fully added all metrics. It was a way to enforce preventing the sink from accessing the reporter until all init had completed. If we don't want to enforce that, we can make the change, however, sinks should know to wait until collection is started before calling the reporter.

system/metrics/sinks/prometheus/prometheusReporterService.cpp
rpastrana: Agreed

system/metrics/sinks/prometheus/prometheusReporterService.cpp
```diff
@@ -0,0 +1,354 @@
...
+#include "espcontext.hpp"
+#include "httpsecurecontext.hpp"
+#include "htmlpage.hpp"
+
+using namespace hpccMetrics;
+
+PrometheusReporterService::PrometheusReporterService(const char * sinkname, const IPropertyTree * settings)
+{
+    m_metricsReporter = nullptr;
+    m_metricsSinkName = sinkname;
```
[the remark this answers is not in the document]
ghalliday: Actually it is because it is called from another class that saves it.  I added a separate question about why there are 2 classes.

system/metrics/sinks/prometheus/prometheusReporterService.cpp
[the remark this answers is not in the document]
ghalliday: @kenrowland We support dynamic metrics - so the sinks already need to support metrics being added after startCollection is called.  I can't think of any reason why they would call the reporter before startCollection() is called, and that can be a requirement if necessary.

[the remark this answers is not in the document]
kenrowland: @ghalliday I agree that sinks must support metrics being added dynamically. I was looking for a way to enforce no sink calling the reporter until it had finished with its initialization. If a requirement on the sink is sufficient, then let me know and I will open a Jira to address it. I suggest we wait until after this PR is merged so that all sinks and the framework are handled in the same PR.

system/metrics/sinks/prometheus/prometheusReporterService.cpp
rpastrana: I'm conflicted about this one...
@kenrowland is there a way for the sink to query the framework for metrics to be reported at init time?
I know metrics can be added dynamically, but this would allow the sink to report potential problems once at init time, with out needed to keep track of metrics, and performing look-ups at runtime.

system/metrics/sinks/prometheus/prometheusReporterService.cpp
rpastrana: Agreed, it seems wasteful to report it every time, however their examples and literature seem to indicate the help is expected on every output (if available). The sink won't keep track of the unique requester(s) nor sessions, and therefore only reporting it once might not be particularly helpful. Perhaps expose a 'verbose' option to expose/suppress the help?

system/metrics/sinks/prometheus/prometheusReporterService.cpp
rpastrana: Agreed, unfortunately the CHttpRequest/CHttpResponse utilize that object to keep some generic, and some ESP specific context info - perhaps a bit of an overkill. It feels like there should be base httpcontext, extended by the IEspContext, but as is the ESPContext is referenced in the http req/resp logic.
This particular service def doesn't need to create the "httpSecureContext" and will pass nullptr instead.

ghalliday: @afishbeck what is the best solution for this problem.  You added system/httplib for the Hashicorp vault secrets.  Should that be used?  Should we compile that into jlib somehow so that the code is shared?

system/metrics/sinks/prometheus/CMakeLists.txt
ghalliday: Note, dafilesrv is one of the components that is likely to have metrics added.  It should be lightweight, and it would not pull in workunit, xmllib or xalan/xerces.

kenrowland: Bringing in esphttp will include some of the base ESP metrics such as active request count. That would mean all components would get an ESP related active request count as part of their metrics as the metric is created statically.

system/metrics/sinks/prometheus/prometheusSink.cpp
```diff
@@ -0,0 +1,53 @@
...
+    return pSink;
+}
+
+
+PrometheusMetricSink::PrometheusMetricSink(const char *name, const IPropertyTree *pSettingsTree) :
+    MetricSink(name, REPORTER_TYPE)
+{
+	try
+	{
+		m_server.set(new PrometheusReporterService(this->name.c_str(), pSettingsTree));
```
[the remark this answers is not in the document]
rpastrana: It seemed like a reasonable logic abstraction layer... will any other component ever reference either class individually?? maybe not.

system/metrics/testing/metrictest.cpp
```diff
@@ -47,7 +72,8 @@ int main(int argc, char *argv[])
 
         //
         // Simulate retrieving the component and global config
-        Owned<IPropertyTree> pSettings = createPTreeFromYAMLString(testLogSinkConfigYml, ipt_none, ptr_ignoreWhiteSpace, nullptr);
+        //Owned<IPropertyTree> pSettings = createPTreeFromYAMLString(testLogSinkConfigYml, ipt_none, ptr_ignoreWhiteSpace, nullptr);
+        Owned<IPropertyTree> pSettings = createPTreeFromYAMLString(testPrometheusSinkConfigYml, ipt_none, ptr_ignoreWhiteSpace, nullptr);
```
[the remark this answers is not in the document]
kenrowland: I will start looking at converting the test program to a set of unit tests for the reporter and for each individual sink. Eventually the test program should be removed.

rpastrana:
@ghalliday please review httplib rendition

system/metrics/sinks/prometheus/prometheusSink.hpp
```diff
@@ -0,0 +1,116 @@
...
+#endif
+
+#undef INVALID_SOCKET
+#include "httplib.h"
+
+#if defined(__clang__) || defined(__GNUC__)
+#pragma GCC diagnostic pop
+#endif
+
+#undef INVALID_SOCKET
```
ghalliday, line 41: lines 41-46 look like they are duplicates of 34-39.

system/metrics/sinks/prometheus/prometheusSink.hpp
```diff
@@ -0,0 +1,116 @@
...
+    StringBuffer                    m_metricsServiceName;
+    bool                            m_verbose;
+
+    static constexpr const char * PROMETHEUS_METRICS_HTTP_PAGE_TITLE = "HPCC Systems - Prometheus Metrics Service";
+    static constexpr const char * PROMETHEUS_METRICS_SERVICE_RESP_TYPE = "text/html; charset=UTF-8";
+    static constexpr int          DEFAULT_PROMETHEUS_METRICS_SERVICE_PORT = 8767;
+    static constexpr bool         DEFAULT_PROMETHEUS_METRICS_SERVICE_SSL = false;
+    static constexpr const char * DEFAULT_PROMETHEUS_METRICS_SERVICE_NAME = "/metrics";
+    static constexpr const char * HTTPLIB_ERROR_MESSAGE_HEADER_NAME = "EXCEPTION_WHAT";
+    static constexpr const char * PROMETHEUS_METRICS_HTTP_ERROR = "\
```
ghalliday, line 84: I don't think this is a legal c++ string.  I'm not sure why the compiler is accepting it.  You could either use string concatenation:

const char * s = "abc"
    "def\n"
    "...";

or use the raw string syntax if you want embedded newlines, and no need to quote ".

const char * s = R"!!(<html>
...
)!!";

system/metrics/sinks/prometheus/prometheusSink.hpp
```diff
@@ -0,0 +1,116 @@
...
+#else
+#define PROMETHEUSSINK_API DECL_IMPORT
+#endif
+
+class PROMETHEUSSINK_API PrometheusMetricSink : public MetricSink
+{
+public:
+    explicit PrometheusMetricSink(const char *name, const IPropertyTree *pSettingsTree);
+    ~PrometheusMetricSink() override = default;
+    static constexpr const char * PROMETHEUS_REPORTER_TYPE = "prometheus";
```
ghalliday, line 66: trivial - not quite sure why this isn't with the other strings (or inlined since only used once)

ghalliday (changes requested):
@rpastrana looks very close to ready.  A few minor comments.

rpastrana:
@ghalliday please confirm lastest changes and I'll squash

system/metrics/sinks/prometheus/prometheusSink.hpp
```diff
@@ -75,28 +67,22 @@ class PROMETHEUSSINK_API PrometheusMetricSink : public MetricSink
...
-        <H1>Error: %s</H1>\
-        <H2>Path: %s</H2>\
-        <H2>Status: %d</H2>\
-        </body>\
-        \
-        </html>\
-        ";
+    static constexpr const char * PROMETHEUS_METRICS_HTTP_ERROR = R"!!(
+<html><head>
+<meta http-equiv=\"Content-Type\" content=\"text/html; charset=utf-8\">
```
ghalliday, line 78: Have you tested this?  I don't think you want the \ in front of the "s.  Also, will it be a problem having a newline before \< html >?

ghalliday (changes requested):
@rpastrana

system/metrics/sinks/prometheus/prometheusSink.hpp
rpastrana: I passed my eye-ball test: 
image
but you're right, I'll update the string.

rpastrana:
@ghalliday corrected the escape chars, and added proper html doctype declaration.

ghalliday (approved):
@rpastrana please squash and I will mrge

rpastrana:
@ghalliday squashed.

ghalliday (approved):

PR #14958, opened 2021-05-13
HPCC-25833 Add more changes in ESP FS service for cloud

wangkx:
1. Add #ifndef _CONTAINERIZED for the code only used by legacy
ECLWatch.
2. When FileSpray service is loaded, call the getTpDropZones()
to retrieve dropzone information for both containerized
environment and bare metal environment.
3. Modify the code to support containerized environment.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/services/ws_fs/ws_fsService.cpp
+++ b/esp/services/ws_fs/ws_fsService.cpp
@@ -611,9 +611,10 @@ bool CFileSprayEx::ParseLogicalPath(const char * pLogicalPath, StringBuffer &tit
 void setRoxieClusterPartDiskMapping(const char *clusterName, const char *defaultFolder, const char *defaultReplicateFolder,
                                bool supercopy, IDFUfileSpec *wuFSpecDest, IDFUoptions *wuOptions)
 {
-#ifdef _CONTAINERIZED
-    UNIMPLEMENTED_X("CONTAINERIZED(setRoxieClusterPartDiskMapping)");
-#else
+    ClusterPartDiskMapSpec spec;
+    spec.setDefaultBaseDir(defaultFolder);
+#ifndef _CONTAINERIZED
+    //In containerized mode, there is no need to replicate files to the local disks of the roxie cluster.
     Owned<IEnvironmentFactory> envFactory = getEnvironmentFactory(true);
     Owned<IConstEnvironment> constEnv = envFactory->openEnvironment();
 
@@ -641,8 +642,6 @@ void setRoxieClusterPartDiskMapping(const char *clusterName, const char *default
 
     unsigned numDataCopies = process.getPropInt("@numDataCopies", 1);
 
-    ClusterPartDiskMapSpec spec;
-    spec.setDefaultBaseDir(defaultFolder);
     if (strieq(slaveConfig, "overloaded"))
     {
         channelsPerNode = process.getPropInt("@channelsPernode", 1);
@@ -663,11 +662,11 @@ void setRoxieClusterPartDiskMapping(const char *clusterName, const char *default
         replicate = true;
     }
     spec.setRoxie (redundancy, channelsPerNode, replicateOffset);
+    wuOptions->setReplicate(replicate);
+#endif
     if (!supercopy)
         spec.setRepeatedCopies(CPDMSRP_lastRepeated,false);
     wuFSpecDest->setClusterPartDiskMapSpec(clusterName,spec);
-    wuOptions->setReplicate(replicate);
-#endif
 }
 
 StringBuffer& constructFileMask(const char* filename, StringBuffer& filemask)
@@ -964,14 +963,14 @@ bool CFileSprayEx::onGetDFUWorkunits(IEspContext &context, IEspGetDFUWorkunits &
             clusterReq.append(clusterName);
         }
 
-        StringArray targetClusters, clusterProcesses;
-#ifdef _CONTAINERIZED
-        IERRLOG("CONTAINERIZED(CFileSprayEx::onGetDFUWorkunits)");
-#else
+#ifndef _CONTAINERIZED
+        //The code is used to find out a target cluster from a process cluster.
+        //It is not needed for containerized environment.
         Owned<IEnvironmentFactory> envFactory = getEnvironmentFactory(true);
         Owned<IConstEnvironment> constEnv = envFactory->openEnvironment();
         Owned<IPropertyTree> root = &constEnv->getPTree();
 
+        StringArray targetClusters, clusterProcesses;
         Owned<IPropertyTreeIterator> clusters= root->getElements("Software/Topology/Cluster");
         if (clusters->first())
         {
@@ -1019,7 +1018,6 @@ bool CFileSprayEx::onGetDFUWorkunits(IEspContext &context, IEspGetDFUWorkunits &
             } while (clusters->next());
         }
 #endif
-
         __int64 pagesize = req.getPageSize();
         __int64 pagefrom = req.getPageStartFrom();
         __int64 displayFrom = 0;
@@ -1132,8 +1130,13 @@ bool CFileSprayEx::onGetDFUWorkunits(IEspContext &context, IEspGetDFUWorkunits &
             resultWU->setUser(wu->getUser(user).str());
 
             const char* clusterName = wu->getClusterName(cluster).str();
-            if (clusterName)
+            if (!isEmptyString(clusterName))
             {
+#ifdef _CONTAINERIZED
+                resultWU->setClusterName(clusterName);
+#else
+                //For bare metal environment, the wu->getClusterName() may return a process
+                //cluster. It has to be converted to a target cluster.
                 StringBuffer clusterForDisplay(clusterName);
 
                 if (clusterProcesses.ordinality())
@@ -1149,6 +1152,7 @@ bool CFileSprayEx::onGetDFUWorkunits(IEspContext &context, IEspGetDFUWorkunits &
                     }
                 }
                 resultWU->setClusterName(clusterForDisplay.str());
+#endif
             }
 
             resultWU->setIsProtected(wu->isProtected());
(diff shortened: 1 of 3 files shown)


esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -2944,21 +2959,27 @@ bool CFileSprayEx::checkDropZoneIPAndPath(double clientVersion, const char* drop
...
+            continue;
+        if (strnicmp(path, prefix, strlen(prefix)) != 0)
+            continue;
+
+        IArrayOf<IConstTpMachine>& tpMachines = dropZone.getTpMachines();
+        ForEachItemIn(ii, tpMachines)
+        {
+            IConstTpMachine& tpMachine = tpMachines.item(ii);
+            const char* netAddress = tpMachines.item(ii).getNetaddress();
+            if (streq(netAddr, netAddress))
```
afishbeck, line 2982: I think in the cloud the user should be dealing with DNS names rather than IPs.  Can we make sure if they are providing DNS name we match that?  We can still let them provide an IP for now just in case, but we should probably frown on it and only be using DNS names.  Unless I'm missing something.

wangkx:
@afishbeck the PR has been rebased to 8.2.0 (the first commit) and revised (the second commit) based on the review comment. Please re-review.

wangkx:
@ghalliday the PR has been rebased to the latest 8.2.0. Please review.

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -3037,13 +3044,29 @@ bool CFileSprayEx::onFileList(IEspContext &context, IEspFileListRequest &req, IE
...
+    ForEachItemIn(i, allTpDropZones)
+    {
+        IConstTpDropZone& dropZone = allTpDropZones.item(i);
+        if (!isEmptyString(dropZoneName) && !streq(dropZoneName, dropZone.getName()))
+            continue;
+
+        const char* prefix = dropZone.getPath();
+        if (strnicmp(path, prefix, strlen(prefix)) != 0)
+            continue;
+
```
ghalliday, line 3061: if netAddr is blank should return true?

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -3153,8 +3176,35 @@ bool CFileSprayEx::onDropZoneFileSearch(IEspContext &context, IEspDropZoneFileSe
         if (!validNameFilter)
             throw MakeStringException(ECLWATCH_INVALID_INPUT, "Invalid Name Filter '*'");
 
+        double version = context.getClientVersion();
```
ghalliday, line 3179: Should the dropZoneServerReq be allowed to be null if there is only one server associated with the drop zone (likely to be very common)?
It could also mean all servers for a given name.

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -3153,8 +3176,35 @@ bool CFileSprayEx::onDropZoneFileSearch(IEspContext &context, IEspDropZoneFileSe
...
+                continue;
+            if (req.getECLWatchVisibleOnly() && !dropZone.getECLWatchVisible())
+                continue;
+
+            IArrayOf<IConstTpMachine>& tpMachines = dropZone.getTpMachines();
+            ForEachItemIn(ii, tpMachines)
+            {
+                IConstTpMachine& tpMachine = tpMachines.item(ii);
+                const char* netAddress = tpMachine.getNetaddress();
+                if (isEmptyString(netAddress) || !matchNetAddressRequest(dropZoneServerReq, isIPAddressReq, tpMachine))
```
ghalliday, line 3199: Would it be simpler to code matchNetAddressRequest() as:

```
extern TPWRAPPER_API bool matchNetAddressRequest(const char* netAddressReg, IConstTpMachine& tpMachine)
{
    return streq(netAddressReg, tpMachine.getNetaddress()) || 
               streq(netAddressReg, tpMachine.getConfigNetaddress());
}
```

ghalliday, line 3199: curious: What is the check for an empty netAddress for?

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -3380,9 +3425,40 @@ bool CFileSprayEx::onDropZoneFiles(IEspContext &context, IEspDropZoneFilesReques
         bool filesFromALinux = false;
         IArrayOf<IEspDropZone> dropZoneList;
         bool ECLWatchVisibleOnly = req.getECLWatchVisibleOnly();
-
 #ifdef _CONTAINERIZED
-        IERRLOG("CONTAINERIZED(CFileSprayEx::onDropZoneFileSearch)");
+        bool isIPAddressReq = !isEmptyString(netAddress) && isIPAddress(netAddress);
+        ForEachItemIn(i, allTpDropZones)
```
ghalliday, line 3430: Could this code be used in both bare metal and containerized code?

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -3153,8 +3176,35 @@ bool CFileSprayEx::onDropZoneFileSearch(IEspContext &context, IEspDropZoneFileSe
         if (!validNameFilter)
             throw MakeStringException(ECLWATCH_INVALID_INPUT, "Invalid Name Filter '*'");
 
+        double version = context.getClientVersion();
+        bool serverFound = false;
+        unsigned filesFound = 0;
+        IArrayOf<IEspPhysicalFileStruct> files;
 #ifdef _CONTAINERIZED
-        UNIMPLEMENTED_X("CONTAINERIZED(CFileSprayEx::onDropZoneFileSearch)");
+        bool isIPAddressReq = isIPAddress(dropZoneServerReq);
```
ghalliday, line 3184: Could this code be used for bare-metal and containerized?

ghalliday (commented):
@wangkx a few minor comments

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -182,6 +182,9 @@ void CFileSprayEx::init(IPropertyTree *cfg, const char *process, const char *ser
         throw MakeStringException(-1, "No Dali Connection Active. Please Specify a Dali to connect to in your configuration file");
     }
 
+    CTpWrapper tpWrapper;
+    tpWrapper.getTpDropZones(9999, nullptr, false, allTpDropZones); //version 9999: get the latest information about dropzone
```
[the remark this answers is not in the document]
wangkx: Can we assume that an ESP has to be restarted if the settings are changed? If not, I may need to call the getTpDropZones() every times when the dropzone information is needed.

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -3037,13 +3044,29 @@ bool CFileSprayEx::onFileList(IEspContext &context, IEspFileListRequest &req, IE
 
 bool CFileSprayEx::checkDropZoneIPAndPath(double clientVersion, const char* dropZoneName, const char* netAddr, const char* path)
 {
-#ifdef _CONTAINERIZED
-    IERRLOG("CONTAINERIZED(CFileSprayEx::checkDropZoneIPAndPath)");
-    return false;
-#else
     if (isEmptyString(netAddr) || isEmptyString(path))
```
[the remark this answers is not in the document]
wangkx: The checkDropZoneIPAndPath() is called before dropzone files can be accessed. The code is not from this PR. In bare metal environment, one dropzone may have multiple servers (network addresses). The network address (not dropzone name) has to be specified in order to access dropzone files. We may access all files within a dropzone if network address is not specified. That will change the original design. Should we try it in another PR?

esp/services/ws_fs/ws_fsService.cpp
wangkx: Related to the comment above

esp/services/ws_fs/ws_fsService.cpp
wangkx: Yes, I will change it

esp/services/ws_fs/ws_fsService.cpp
wangkx: Yes, I will change it

esp/services/ws_fs/ws_fsService.cpp
wangkx: The netAddress is used in line 3203. So, it is checked here.

esp/services/ws_fs/ws_fsService.cpp
wangkx: The code is not from this PR. I think it is a good enhancement to allow dropZoneServerReq be null. Changed it in another PR?

wangkx:
@AttilaVamos is the Smoketest error related to my changes in this PR?

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -3661,15 +3662,15 @@ bool CFileSprayEx::onGetDFUServerQueues(IEspContext &context, IEspGetDFUServerQu
...
     try
     {
-#ifdef _CONTAINERIZED
-        UNIMPLEMENTED_X("CONTAINERIZED(CFileSprayEx::onGetDFUServerQueues)");
-#else
         context.ensureFeatureAccess(FILE_SPRAY_URL, SecAccess_Read, ECLWATCH_FILE_SPRAY_ACCESS_DENIED, "Permission denied.");
 
         StringArray qlist;
+#ifdef _CONTAINERIZED
+        qlist.append(m_QueueLabel); // TODO: re-design support for multiple queues (see the comment in init()).
```
ghalliday, line 3669: This may as well be implemented using code similar to init().  Something like:

```
    Owned<IPropertyTreeIterator> dfuQueues = getComponentConfigSP()->getElements("dfuQueues");
    ForEach(*dfuQueues)
    {
        IPropertyTree & dfuQueue = dfuQueues->query();
        const char * dfuName = dfuQueue.queryProp("@name");
        assertex(!isEmptyString(dfuName)); //  redundant since checked by the schema...
        StringBuffer queueLabel;
        getDfuQueueName(queueLabel, dfuName);
        qlist.append(queueLabel);
    }

```

ghalliday (commented):
@wangkx I think this is close, but I have a few concerns

* That commoning up the code doesn't break the existing bare-metal implementation.  (In the end this was the reason for not merging into 8.2.0 because the changes were last minute.)
* That the k8s version works sensibly.  (E.g. no need to pass in a ip or a directory)
* That the bare-metal and k8s versions work consistently.

If it doesn't break bare-metal then it would be ok to merge, provided the other two issues are addressed soon afterwards.  Please can you test the bare-metal versions to ensure that the code that has been commoned up hasn't introduced any issues.

esp/services/ws_fs/ws_fsService.cpp
wangkx: OK, I will change.

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -3153,47 +3176,37 @@ bool CFileSprayEx::onDropZoneFileSearch(IEspContext &context, IEspDropZoneFileSe
...
-            if (server.isEmpty())
-                throw MakeStringException(ECLWATCH_INVALID_INPUT, "Invalid server for dropzone %s.", dropZoneName);
+        bool isIPAddressReq = isIPAddress(dropZoneServerReq);
+        IArrayOf<IConstTpDropZone> allTpDropZones;
+        CTpWrapper tpWrapper;
+        tpWrapper.getTpDropZones(9999, nullptr, false, allTpDropZones); //version 9999: get the latest information about dropzone
+        ForEachItemIn(i, allTpDropZones)
+        {
+            IConstTpDropZone& dropZone = allTpDropZones.item(i);
+            const char* name = dropZone.getName();
```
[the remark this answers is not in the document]
wangkx: This will behave the same: compare both name and address. Before this PR, the name is used to retrieve the dropzone information at line 3161.

wangkx:
@ghalliday the changes in this PR are tested in a bare-metal environment. The https://track.hpccsystems.com/browse/HPCC-26165 is created for the other two issues. Please review.

ghalliday:
How do I test the DropZoneFileSearch function?  I can't find anything in eclwatch that seems to correspond to it.
@GordonSmith I found CreateLandingZonesFilterStore in the FileSpray.ts, but it doesn't appear to be used.

wangkx:
I tested the DropZoneFileSearch function using http://your_esp_ip:8010/FileSpray/DropZoneFileSearch?form.

GordonSmith:
@ghalliday in ECL Watch select the landing zones page and then enter a filter.

ghalliday:
@GordonSmith thanks.  I couldn't see that - even though I was looking for it.  I think the UI could do with some improvement - in this case separating commands from functions that affect the view.  Something to think about for the new version (if you haven't already).

ghalliday (approved):
@wangkx please can you squash.

wangkx:
@ghalliday I just squashed.

wangkx:
The changes are rebased to the latest 8.2.x.

PR #14963, opened 2021-05-14
HPCC-25764 Roxie target planes - fix unknown

jakesmith:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dali/base/dautils.cpp
+++ b/dali/base/dautils.cpp
@@ -3194,7 +3194,7 @@ public:
         return dfile.get(); 
     }
 
-    bool init(const char *fname,IUserDescriptor *user,bool onlylocal,bool onlydfs, bool write, bool isPrivilegedUser)
+    bool init(const char *fname,IUserDescriptor *user,bool onlylocal,bool onlydfs, bool write, bool isPrivilegedUser, const StringArray *clusters)
     {
         fileExists = false;
         if (!onlydfs)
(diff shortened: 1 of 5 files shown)


dali/base/dautils.cpp
```diff
@@ -3245,9 +3245,25 @@ class CLocalOrDistributedFile: implements ILocalOrDistributedFile, public CInter
...
+#ifdef _CONTAINERIZED
+            // NB: this may work equally as well in bare-metal
+            if (clusters)
+            {
+                if (clusters->ordinality()>1)
+                    throw makeStringExceptionV(0, "Container mode does not yet support output to multiple clusters while writing file %s)", fname);
+                cluster.append(clusters->item(0));
+            }
+            else
+                getDefaultStoragePlane(cluster);
```
ghalliday, line 3260: Not directly related, but this function implementation looks wrong.
Shouldn't it be using @ storagePlane

    const char * plane = config.queryProp("@storagePlane");

dali/base/dautils.cpp
```diff
@@ -3245,9 +3245,25 @@ class CLocalOrDistributedFile: implements ILocalOrDistributedFile, public CInter
                 if (dfile.get())
                     return true;
             }
+
+            StringBuffer dir;
+            StringBuffer cluster;
```
ghalliday, line 3250: cluster should be inside the #ifdef - otherwise it will be an unreferenced object

dali/base/dautils.cpp
```diff
@@ -3245,9 +3245,25 @@ class CLocalOrDistributedFile: implements ILocalOrDistributedFile, public CInter
...
+#ifdef _CONTAINERIZED
+            // NB: this may work equally as well in bare-metal
```
ghalliday, line 3252: except the default directory in bare metal also includes the component

ghalliday (commented):
A couple of comments

dali/base/dautils.cpp
jakesmith: agreed looks wrong. As existing separate issued, I've opened a separate JIRA : https://track.hpccsystems.com/browse/HPCC-25903

dali/base/dautils.cpp
jakesmith: I'll take comment out for now

dali/base/dautils.cpp
jakesmith: right, I'll move it.
gcc doesn't complain - does clang? Is it because compiler not smart enough to realize because object has ctor etc. ?

jakesmith:
@ghalliday - went ahead and squashed in removal of comment and move of StringBuffer cluster; info #ifdef.

richardkchapman (approved):

ghalliday (approved):

PR #14969, opened 2021-05-18
HPCC-24059 Additional SOAPCALL option to log text at end of soap request

afishbeck:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/ecl/hql/hqlgram.y
+++ b/ecl/hql/hqlgram.y
@@ -3768,6 +3768,12 @@ soapFlag
                             parser->normalizeExpression($3, type_string, false);
                             $$.setExpr(createExprAttribute(logAtom, $3.getExpr()), $1);
                         }
+    | TOK_LOG '(' expression ',' expression ')'
+                        {
+                            parser->normalizeExpression($3, type_string, false);
+                            parser->normalizeExpression($5, type_string, false);
+                            $$.setExpr(createExprAttribute(logAtom, $3.getExpr(), $5.getExpr()), $1);
+                        }
     | XML_TOKEN         {
                             $$.setExpr(createAttribute(xmlAtom));
                             $$.setPosition($1);
--- a/ecl/hqlcpp/hqlhtcpp.cpp
+++ b/ecl/hqlcpp/hqlhtcpp.cpp
@@ -17965,7 +17965,7 @@ ABoundActivity * HqlCppTranslator::doBuildActivitySOAP(BuildCtx & ctx, IHqlExpre
 
     IHqlExpression * namespaceAttr = expr->queryAttribute(namespaceAtom);
     IHqlExpression * responseAttr = expr->queryAttribute(responseAtom);
-    IHqlExpression * logText = NULL;
+    HqlExprArray logTextArray;
     bool logMin = false;
     bool logXml = false;
     ForEachChildFrom(i, expr, 1)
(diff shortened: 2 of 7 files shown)


afishbeck:
@ghalliday not sure what to do about the eclhelper interface changed test when it's intentional.

richardkchapman:
This needs to target master

afishbeck:
@richardkchapman Retargeted to master.

ecl/hql/hqlgram.y
```diff
@@ -3768,6 +3768,12 @@ soapFlag
                             parser->normalizeExpression($3, type_string, false);
                             $$.setExpr(createExprAttribute(logAtom, $3.getExpr()), $1);
                         }
+    | TOK_LOG '(' expression ',' expression ')'
+                        {
+                            parser->normalizeExpression($3, type_string, false);
+                            parser->normalizeExpression($5, type_string, false);
+                            $$.setExpr(createExprAttribute(logAtom, createComma($3.getExpr(), $5.getExpr())), $1);
```
ghalliday, line 3775: minor: Normally you would not have a comma inside an attribute.  i.e:

    createExprAttribute(logAtom, $3.getExpr(), $5.getExpr())

and then use unwindChildren() to extract them

ecl/hqlcpp/hqlhtcpp.cpp
```diff
@@ -18198,10 +18213,16 @@ ABoundActivity * HqlCppTranslator::doBuildActivityHTTP(BuildCtx & ctx, IHqlExpre
...
 
-    if (logText)
+    if (logTextArray.length())
     {
         MemberFunction func(*this, instance->startctx, "virtual void getLogText(size32_t & __lenResult, char * & __result, const void * _left) override");
-        doBuildFunctionReturn(func.ctx, unknownStringType, logText);
+        doBuildFunctionReturn(func.ctx, unknownStringType, &logTextArray.item(0));
+    }
+
+    if (logTextArray.length()>1)
```
ghalliday, line 18222: trivial: duplicate changes suggest these functions should be commoned up.  Not for this PR though.

ghalliday (changes requested):
@afishbeck one fairly minor comment

ghalliday:
@afishbeck you should also increment 
#define ACTIVITY_INTERFACE_VERSION      652
since the helper has changed.

afishbeck:
@ghalliday I'll increment the ACTIVITY_INTERFACE_VERSION.  But just for discussion, since we're setting and checking a flag, that would seem to make it backward compatible.  The extra logging just wouldn't appear on older builds.  I imagine this is to prevent confusion with unexpected/missing behavior, although in this case it's fairly minor?

ecl/hql/hqlgram.y
afishbeck: Gotcha.  I think I remember how a comma might be used now, it's been a while.  I'll fix it.

ghalliday:
We're mainly concerned about whether old queries will run without recompiling on new systems - and increasing that number will not prevent that.  It will probably prevent new code running on old systems, but that isn't generally an issue.

ecl/hql/hqlgram.y
ghalliday: The commas are mainly used for datasets.  At some point I may get rid of most/all of them and use initializer lists instead.

afishbeck:
@ghalliday removed createComma, and updated ACTIVITY_INTERFACE_VERSION.

ghalliday (approved):

richardkchapman (approved):

PR #14971, opened 2021-05-19
HPCC-25835 Fix environment accesses in ws_dfuXRefService for cloud

wangkx:
In WsDfuXRef.DFUXRefUnusedFiles(), use preload ISmartSocket to send
roxie command.
In WsDfuXRef.DFUXRefList, conditionally compile the code related to
process cluster.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/scm/ws_dfuXref.ecm
+++ b/esp/scm/ws_dfuXref.ecm
@@ -213,6 +213,7 @@ ESPrequest DFUXRefUnusedFilesRequest
     string ProcessCluster;
     bool CheckPackageMaps;
     [min_ver("1.01")] bool GetFileDetails(false);
+    [min_ver("1.03")] string TargetCluster;
 };
 
 ESPresponse [exceptions_inline] DFUXRefUnusedFilesResponse
@@ -223,7 +224,7 @@ ESPresponse [exceptions_inline] DFUXRefUnusedFilesResponse
 };
 
 //  ===========================================================================
-ESPservice [version("1.02"), default_client_version("1.02"), auth_feature("DEFERRED"), exceptions_inline("./smc_xslt/exceptions.xslt")] WsDFUXRef
+ESPservice [version("1.03"), default_client_version("1.03"), auth_feature("DEFERRED"), exceptions_inline("./smc_xslt/exceptions.xslt")] WsDFUXRef
 {
     ///ESPmethod [resp_xsl_default("./smc_xslt/xref_main.xslt")] DFUXRefList(DFUXRefListRequest, DFUXRefListResponse);
     ESPmethod [resp_xsl_default("/esp/xslt/xref_main.xslt")] DFUXRefList(DFUXRefListRequest, DFUXRefListResponse);
--- a/esp/services/ws_dfu/ws_dfuXRefService.cpp
+++ b/esp/services/ws_dfu/ws_dfuXRefService.cpp
@@ -115,6 +115,10 @@ void CWsDfuXRefEx::init(IPropertyTree *cfg, const char *process, const char *ser
         throw MakeStringException(-1, "No Dali Connection Active. Please Specify a Dali to connect to in you configuration file");
     }
 
+#ifdef _CONTAINERIZED
+    initContainerRoxieTargets(roxieConnMap);
+#endif
+
     XRefNodeManager.setown(CreateXRefNodeFactory());
 
     //Start out builder thread......
@@ -555,12 +559,15 @@ bool CWsDfuXRefEx::onDFUXRefList(IEspContext &context, IEspDFUXRefListRequest &r
     {
         context.ensureFeatureAccess(FEATURE_URL, SecAccess_Read, ECLWATCH_DFU_XREF_ACCESS_DENIED, "WsDfuXRef::DFUXRefList: Permission denied.");
 
+        Owned<IPropertyTree> xrefNodeTree = createPTree("XRefNodes");
+
+#ifndef _CONTAINERIZED
+        //In bare metal environment, each physical cluster is an XRef node.
         CConstWUClusterInfoArray clusters;
         getEnvironmentClusterInfo(clusters);
 
         BoolHash uniqueProcesses;
-        Owned<IPropertyTree> xrefNodeTree = createPTree("XRefNodes");
-        ForEachItemIn(c, clusters)
+        ForEachItemIn(c, clusters) //loop through every target cluster
         {
             IConstWUClusterInfo &cluster = clusters.item(c);
             switch (cluster.getPlatform())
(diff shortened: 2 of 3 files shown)


esp/services/ws_dfu/ws_dfuXRefService.cpp
```diff
@@ -555,12 +559,15 @@ bool CWsDfuXRefEx::onDFUXRefList(IEspContext &context, IEspDFUXRefListRequest &r
     {
         context.ensureFeatureAccess(FEATURE_URL, SecAccess_Read, ECLWATCH_DFU_XREF_ACCESS_DENIED, "WsDfuXRef::DFUXRefList: Permission denied.");
 
+        Owned<IPropertyTree> xrefNodeTree = createPTree("XRefNodes");
+
+#ifndef _CONTAINERIZED
+        //In bare metal environment, each physical cluster is an XRef node.
```
afishbeck, line 565: I'm not that familiar with how this service interacts with XREF.  This change correctly removes the process clusters (which should match DFU groups) from the list.  

@ghalliday @jakesmith 
But should the storage planes be on the list instead?  Or is that not necessary in the cloud?

jakesmith:
@wangkx - xref isn't supported in cloud, we're not entirely sure of the future of it.

So I'm not sure this PR is required (for anything else) ?

wangkx:
@jakesmith Yes. The other fix in this PR is for WsDfuXRef.DFUXRefUnusedFiles() which was added by https://github.com/hpcc-systems/HPCC-Platform/commit/c135674bd13337bfa85ed330df8589a7fb72d98a.

esp/scm/ws_dfuXref.ecm
```diff
@@ -213,6 +213,7 @@ ESPrequest DFUXRefUnusedFilesRequest
     string ProcessCluster;
     bool CheckPackageMaps;
     [min_ver("1.01")] bool GetFileDetails(false);
+    [min_ver("1.03")] string TargetCluster;
```
jakesmith, line 216: @wangkx - I'm not familiar with this service, what is the difference between the pre-existing ProcessCluster and the new TargetCluster ?

wangkx: The pre-existing ProcessCluster was used in bare metal environment for specifying the physical cluster. For cloud environment, a user should specify a roxie queue target. The new TargetCluster is added for specifying the roxie queue target. For backward compatible, it is allowed to specify the roxie queue target using pre-existing ProcessCluster.

jakesmith: If ProcessCluster can/does mean target roxie, then an additional field doesn't clarify IMO.
I am not familiar with the interface/way this request is populated, but don't think we should have 2 fields in same request with similar/overlapping meanings.

afishbeck: The process cluster is the physical cluster... which maps to the DFS group name.
The target cluster is a configuration object that combines [process cluster, eclccserver, queryset, etc].
You could have more than one target object configured per physical cluster.
For example the physical cluster roxieprod1 can have the following target clusters configured:

roxieprod1-fromgitrepo
roxieprod1-fromsqlrepo
roxieprod1-x

But all of those reference the same actual physical (process) cluster.

afishbeck: So the interface allows you to specify the physical cluster directly, or we can figure it out if you provide the target name.

afishbeck: In the cloud there is currently no real distinction.

jakesmith: in cloud there is no physical cluster/dfs group, only a target.
in bare-metal (prior to this change) only the physical target was specified.

@afishbeck, is there any case where both need to be specified? If not I think it would be better not to add 2nd type of target field to request.

afishbeck: Agreed in cloud target and process cluster names are interchangeable, so no need to add new target parameter. For this service it really goes back to the question of what xref will become though, since we are really using this as the DFS groupname.
For other services we do also have to be careful where process is mapped to groupname.. we'll have to decide if we mean storage plane on a case by case basis.

jakesmith: There will be planes, but not groups in the longer term - as the moment groups are auto-created based on planes - for backward compatibility with existing code which expect them, but they are pseudo groups really (i.e. they are padded with localhosts representing the partition width of the file that links to them).

It's not clear what DFS XREF will do in future, there's still a role to play with checking the meta is correct and the underlying physical data is there, and we might target a storage plane much in the same we target a group now (but a plane might be backed by a azure blob url)

I'm not sure what the Roxie XREF aspect does (in fact I didn't know it existed until Kevin pointed it our in an earlier comment). From a quick read of the JIRA , I was left the impression that it wasn't to do with integrity of physical files vs logical, but to do with which logical files were still in use in packages, is that true or ?
But if that's true, I'm not sure how in bare-metal a ProcessCluster that mapped to a physical cluster, mapped back to a roxie cluster deployed packaged?

afishbeck: The roxie XREF returns a list of the files which DFS thinks are on the roxie, but no queries (or optionally packagemap) claims to be using.
It's useful for operations to check if they've missed any files that may need to be deleted after they've removed some queries.

afishbeck: I think the roxie one should be made to work in the cloud.

jakesmith: >The roxie XREF returns a list of the files which DFS thinks are on the roxie

It sounds like this still needs to deal with the equivalent of a group, i.e. a plane in cloud world, and not a target roxie.


And in bare-metal - it really should have been told both the target roxie, and the group, but has been assuming they are both the same (roxie could be called myroxie, and writing files to mystorage).

I think the situation is same whether bare-metal or cloud - it should not assume same, request needs both.

Also wondering about the situation where the roxie has queires/packages which are using files on multiple group/planes ...

jakesmith: I think so, as you say, it kinda worked/was useful in baremetal because the storage was dedicated.
It could still work if the a plane was dedicated to a roxie, but where there are multiple roxie's using shared palnes, or 1 or more roxies using multiple planes, the implementation falls short.

So yes, I agree, think should be disabled for now.

Thinking ahead to alternative implementations, would it make sense for the roxie queries that use files to add an ownership tag or similar to the meta of a file, so we knew which logical files were in use by which deployed roxies (and/or down to the query/package level)?

wangkx:
I am closing this PR because all of the XREF functionality needs to be re-considered and should be disabled on the cloud for now.

PR #14978, opened 2021-05-21
HPCC-25841 Build code for WU log only if not container build

wangkx:
In containerized HPCC, the log is accessed from the 3rd party tools.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/services/ws_workunits/ws_workunitsHelpers.cpp
+++ b/esp/services/ws_workunits/ws_workunitsHelpers.cpp
@@ -561,7 +561,7 @@ void WsWuInfo::getHelpers(IEspECLWorkunit &info, unsigned long flags)
             for (unsigned i = 0; i < FileTypeSize; i++)
                 getHelpFiles(query, (WUFileType) i, helpers, flags, helpersCount);
         }
-
+#ifndef _CONTAINERIZED
         getWorkunitThorLogInfo(helpers, info, flags, helpersCount);
 
         if (cw->getWuidVersion() > 0)
@@ -625,6 +625,7 @@ void WsWuInfo::getHelpers(IEspECLWorkunit &info, unsigned long flags)
                 break;
             }
         }
+#endif
 
         info.setHelpers(helpers);
         info.setHelpersCount(helpersCount);
@@ -1063,6 +1064,7 @@ void WsWuInfo::getInfo(IEspECLWorkunit &info, unsigned long flags)
     getServiceNames(info, flags);
 }
 
+#ifndef _CONTAINERIZED
 unsigned WsWuInfo::getWorkunitThorLogInfo(IArrayOf<IEspECLHelpFile>& helpers, IEspECLWorkunit &info, unsigned long flags, unsigned& helpersCount)
 {
     unsigned countThorLog = 0;
@@ -1249,6 +1251,7 @@ unsigned WsWuInfo::getWorkunitThorLogInfo(IArrayOf<IEspECLHelpFile>& helpers, IE
 
     return countThorLog;
 }
+#endif
 
 bool WsWuInfo::getClusterInfo(IEspECLWorkunit &info, unsigned long flags)
 {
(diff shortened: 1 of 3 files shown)


esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -3760,13 +3770,20 @@ int CWsWuFileHelper::zipAFolder(const char* folder, bool gzip, const char* zipFi
     return (system(zipCommand.str()));
 }
 
+#ifdef _CONTAINERIZED
```
afishbeck, line 3773: We discussed this in our triage meeting.  People seem to agree that keeping the extra variable is probably worth it to greatly simplify the containerized IFDEFs. Rather than changing the function parameter list, and maybe even object variable just make it so that variable isn't used or has a value that makes it obsolete. That affects several similar changes in this PR.

afishbeck (changes requested):
@wangkx consensus is that we simplify the ifdefs and don't change the function parameters.. just make it so the value gets ignored in containerized.

wangkx:
@afishbeck I like the idea about simplifying the ifdefs. Please review and let me know whether I did enough.

afishbeck (approved):
@wangkx yes that looks much cleaner.

richardkchapman:
It's still possible for some components to associate a log file as a "associated file" on a workunit - for example eclcc preserves the compiler log files and attaches them. If I'm reading this right, you have disabled the ability to fetch these?

wangkx:
The "associated file" is still fetched into the ZAP report for a cloud environment. It is fetched by createZAPWUQueryAssociatedFiles() in ws_workunitsHelpers.cpp.

richardkchapman:
But can I still view the log file other than via a zap download?

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -510,7 +510,9 @@ EnumMapping queryFileTypes[] = {
    { FileTypeResText, "res" },
    { FileTypeHintXml, "hint" },
    { FileTypeXml, "xml" },
+#ifndef _CONTAINERIZED
```
richardkchapman, line 513: Not convinced that this should be removed

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -4009,7 +4021,9 @@ void CWsWuFileHelper::readWUFile(const char* wuid, const char* workingFolder, Ws
         writeToFileIOStream(workingFolder, "ArchiveQuery.xml", mb);
         break;
     case CWUFileType_CPP:
+#ifndef _CONTAINERIZED
```
richardkchapman, line 4024: Not convinced this should be removed

esp/services/ws_workunits/ws_workunitsService.cpp
```diff
@@ -3074,7 +3074,11 @@ bool CWsWorkunitsEx::onWUFile(IEspContext &context,IEspWULogFileRequest &req, IE
                 winfo.getWorkunitArchiveQuery(mb);
                 openSaveFile(context, opt, req.getSizeLimit(), "ArchiveQuery.xml", HTTP_TYPE_APPLICATION_XML, mb, resp);
             }
+#ifdef _CONTAINERIZED
```
richardkchapman, line 3077: Not sure about this

richardkchapman (commented):
I'm not sure about the implications for the eclcc compiler log

esp/services/ws_workunits/ws_workunitsHelpers.cpp
wangkx: I guess I should not remove. Will the log file be stored in a persistent storage?

wangkx:
@richardkchapman the changes for File_Log are rolled back. Please review.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
richardkchapman: It's stored in the dll location

richardkchapman (approved):

PR #14981, opened 2021-05-24
HPCC-25817 Fix getSashaNodes() in workunitservices for cloud

wangkx:
Add the services into thor yaml file and eclagent yaml file.
Change the getSashaNodes() to get sasha node information from the
yaml files.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/helm/hpcc/templates/thor.yaml
+++ b/helm/hpcc/templates/thor.yaml
@@ -25,11 +25,15 @@ data:
 {{ toYaml $thorScope | indent 6 }}
 {{- include "hpcc.generateLoggingConfig" (dict "root" .root "me" $thorScope) | indent 6 }}
 {{ include "hpcc.generateVaultConfig" . | indent 6 }}
+      services:
+{{ include "hpcc.generateConfigMapServices" .root | indent 6 }}
     {{ $eclAgentType }}: # hthor or roxie
 {{ toYaml $hthorScope | indent 6 }}
       platform:
         type: "thor"
         width: {{ mul ($thorScope.numWorkers | default 1) ( $thorScope.channelsPerWorker | default 1) }}
+      services:
+{{ include "hpcc.generateConfigMapServices" .root | indent 6 }}
 {{- include "hpcc.generateLoggingConfig" (dict "root" .root "me" $hthorScope ) | indent 6 }}
 {{ include "hpcc.generateVaultConfig" . | indent 6 }}
     eclagent: # main agent Q handler
--- a/helm/hpcc/templates/eclagent.yaml
+++ b/helm/hpcc/templates/eclagent.yaml
@@ -18,6 +18,8 @@ data:
 {{ toYaml (omit .me "logging" "env") | indent 6 }}
 {{- include "hpcc.generateLoggingConfig" . | indent 6 }}
 {{ include "hpcc.generateVaultConfig" . | indent 6 }}
+      services:
+{{ include "hpcc.generateConfigMapServices" .root | indent 6 }}
     global:
 {{ include "hpcc.generateGlobalConfigMap" .root | indent 6 }}
 {{- if not .me.useChildProcesses }}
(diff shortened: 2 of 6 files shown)


helm/hpcc/templates/thor.yaml
```diff
@@ -30,6 +30,8 @@ data:
       platform:
         type: "thor"
         width: {{ mul ($thorScope.numWorkers | default 1) ( $thorScope.channelsPerWorker | default 1) }}
+      services:
+{{ include "hpcc.generateConfigMapServices" .root | indent 6 }}
```
jakesmith, line 34: I think this/a services: section may also be needed under:
```
{{ $eclAgentType }}: # hthor or roxie 
```

because this section is going to be used by Thor (manager), but there could be calls to WorkunitServices from the job agent (outside of Thor).

Did you hit the Thor scenario with some test ECL code?

jakesmith (commented):
@wangkx - 1 question about need for this section elsewhere in thor,yaml

Also, I think needs similar for roxie.

helm/hpcc/templates/thor.yaml
wangkx: Is it under the {{ $eclAgentType }}: # hthor or roxie  now? In generated /etc/config/thor.yaml, the services section is under the hthor.
 
I tested hthor, roxie-workunit, and thor using:
OUTPUT(STD.System.Workunit.WorkunitExists('W20210522-211941', false, true));

wangkx:
@jakesmith Looks like that my changes in eclagent.yaml work for roxie-workunit. I run the 'WorkunitExists' test on my roxie-workunit queue and it works.

helm/hpcc/templates/thor.yaml
jakesmith: That ECL won't actually run a Thor/in Thor or Roxie, because it's too simple.
It will instead generate that call in line and run in the eclagent.

You'll need to call inside an activity to test, from an APPLY would work I think.

wangkx: Now, I learn how to force some ECL code running on thor (not roxie yet).
You are correct. The fix does not work for thor. I got the error: System error: 20069: Graph graph1[1], SLAVE #1 [10.1.8.130:20100]: Sasha 'sasha-wu-archiver' server not found.

wangkx: The system error (20069: Graph graph1[1], SLAVE #1 [10.1.8.130:20100]: Sasha 'sasha-wu-archiver' server not found) is 'from' eclagent. My debug code shows that the getComponentConfigSP() returns empty when the error is generated. I am not sure how that happens. For this case, which hpcc component or program (ECL WU dll?) loads the workunitservices plugin (with no ComponentConfig)? @jakesmith @ghalliday any suggestion?

jakesmith: I can look again, but I think it's because the services: are not being generated into all the component configuration sections that need them, i.e. under thor: in this case.

wangkx: I can understand that the getComponentConfigSP() here does not return the services. But, why does it return nothing (no services, and no other item like analyzeWorkunit="true" config="/etc/config/thor.yaml" daliServers="mydali")?

wangkx: Using the same test ecl code (with the Apply and logical file) on roxie, no additional pod is generated. My test code shows  the getSashaNodes() is called successfully inside the log of the roxie-workunit pod (same as using the simple OUTPUT ecl). Not sure other ECL test I should use to test roxie.

jakesmith (commented):
@wangkx - looks like there are now some merge conflicts and needs rebasing

jakesmith (commented):
@wangkx - the change looks good for thor, hthor and roxie-workunit.
I assume workunit services work in bare-metal from deployed roxie queries, so should also work in coud.

roxie.yaml and localroxie.yaml will similarly need updating to services, for them to work.

wangkx:
@jakesmith the PR is rebased. Please review.

wangkx:
@jakesmith I just added the services to roxie.yaml and localroxie.yaml.

wangkx:
This PR is replaced by https://github.com/hpcc-systems/HPCC-Platform/pull/15016.

PR #14983, opened 2021-05-24
HPCC-25935 Allow environment variables to be defined in the values.yaml

richardkchapman:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/helm/hpcc/templates/eclagent.yaml
+++ b/helm/hpcc/templates/eclagent.yaml
@@ -12,10 +12,10 @@ data:
   {{ .me.name }}.yaml:
     version: 1.0
     eclagent:
-{{ toYaml (omit .me "logging") | indent 6 }}
+{{ toYaml (omit .me "logging" "env") | indent 6 }}
 {{- include "hpcc.generateLoggingConfig" . | indent 6 }}
     {{ $apptype }}:
-{{ toYaml (omit .me "logging") | indent 6 }}
+{{ toYaml (omit .me "logging" "env") | indent 6 }}
 {{- include "hpcc.generateLoggingConfig" . | indent 6 }}
 {{ include "hpcc.generateVaultConfig" . | indent 6 }}
     global:
@@ -50,6 +50,10 @@ data:
 {{- include "hpcc.addResources" (dict "me" .me.resources) | indent 12 }}
 {{- $appCmd := printf "%s %s %s _HPCC_ARGS_" $apptype (include "hpcc.configArg" .me) (include "hpcc.daliArg" .root) }}
 {{ include "hpcc.addCommandAndLifecycle" (. | merge (dict "command" $appCmd)) | indent 12 }}
+{{- if .env }}
+            env:
+{{ include "hpcc.mergeEnvironments" .env | indent 12 }}
+{{- end }}
             workingDir: /var/lib/HPCCSystems
             volumeMounts:
 {{ include "hpcc.addConfigMapVolumeMount" .me | indent 12 }}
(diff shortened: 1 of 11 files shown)


helm/hpcc/templates/eclagent.yaml
```diff
@@ -50,6 +50,13 @@ data:
 {{- include "hpcc.addResources" (dict "me" .me.resources) | indent 12 }}
 {{- $appCmd := printf "%s %s %s _HPCC_ARGS_" $apptype (include "hpcc.configArg" .me) (include "hpcc.daliArg" .root) }}
 {{ include "hpcc.addCommandAndLifecycle" (. | merge (dict "command" $appCmd)) | indent 12 }}
+{{- if index .me "env" }}
```
jakesmith, line 53: thought index was supposed to take an index # e.g. index .me 0
may need to be hasKey .me "env" here?
Or just :
{{- if .env }}

helm/hpcc/templates/eclagent.yaml
```diff
@@ -50,6 +50,13 @@ data:
...
+            env:
+{{- range .me.env }}
```
jakesmith, line 55: can you do {{ toYaml .me.env | indent 2 }} instead of range?

helm/hpcc/templates/eclagent.yaml
```diff
@@ -79,7 +86,7 @@ data:
 {{- $apptype := .type | default "hthor" -}}
 {{- $secretsCategories := list "system" "ecl-user" "ecl" "storage" }}
 {{- $commonCtx := dict "root" $ "me" . "secretsCategories" $secretsCategories }}
-{{- $configSHA := include "hpcc.getConfigSHA" ($commonCtx | merge (dict "configMapHelper" "hpcc.agentConfigMap" "component" "eclagent" "excludeKeys" (print "global," $apptype ".replicas"))) }}
+{{- $configSHA := include "hpcc.getConfigSHA" ($commonCtx | merge (dict "configMapHelper" "hpcc.agentConfigMap" "component" "eclagent" "excludeKeys" (print "global," $apptype ".replicas" ".env"))) }}
```
jakesmith, line 89: I may be missing something.. if .env changes, pod will restart anyway, so valid to ignore, but also no point?

helm/hpcc/templates/eclagent.yaml
richardkchapman: I could on this one - I can't on any where we need to add the SENTINEL entry (though I suppose I could do that by adding to the env dictionary). Eventually I expect the resulting env may well be a merge of a global and a local one.

helm/hpcc/templates/eclagent.yaml
richardkchapman: I wanted to suppress it from the config file because confusing to have it there when it isn't read from there.

jakesmith: ok, it will still be in the configmap with this change, just not part of the calculated SHA.
Would need an alternative form of generateConfig (based on getConfigSHA) that generated the filtered output, instead of piping into sha256sum for that.

helm/hpcc/values.schema.json
```diff
@@ -572,6 +575,22 @@
...
+      "type": "array",
+      "items": {
+        "type": "object",
+        "properties": {
+          "name": {
+            "type": "string",
+            "description": "Environment variable name"
+          },
+          "value": {
+            "type": ["integer", "string", "boolean"],
```
jakesmith, line 588: should this be "number" instead of "integer" , to allow for decimals?

jakesmith (commented):
@richardkchapman - looks good, 1 small question.
PR tick boxes need checking (when changed from draft)

jakesmith (approved):
@richardkchapman - looks good.

There's a "WIP" in the PR comment you might want to remove though.

richardkchapman:

helm/hpcc/templates/thor.yaml
```diff
@@ -267,6 +280,10 @@ spec:
                 {{ include "hpcc.daliArg" $ }},
                 {{ printf "\"--name=%s\"" .name }}
               ]
+        env:
+{{ include "hpcc.mergeEnvironments" $env | indent 8 -}}
```
ghalliday, line 284: These sentinal variables are new.  (Also localroxie).  Could it have caused problems before, or are they unused, but here for completeness?

ghalliday (approved):

PR #14984, opened 2021-05-25
HPCC-25945 Simplify config update hook installation

jakesmith:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dali/base/dafdesc.cpp
+++ b/dali/base/dafdesc.cpp
@@ -3339,17 +3339,8 @@ static void generateHosts(IPropertyTree * storage, GroupInfoArray & groups)
     }
 }
 
-
+static CConfigUpdateHook configUpdateHook;
 static std::atomic<unsigned> normalizeHostGroupUpdateCBId{(unsigned)-1};
-MODULE_INIT(INIT_PRIORITY_STANDARD)
-{
-    return true;
-}
-MODULE_EXIT()
-{
-    if ((unsigned)-1 != normalizeHostGroupUpdateCBId)
-        removeConfigUpdateHook(normalizeHostGroupUpdateCBId);
-}
 static CriticalSection storageCS;
 static void doInitializeStorageGroups(bool createPlanesFromGroups)
 {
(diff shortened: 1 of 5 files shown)


dali/base/dafdesc.cpp
```diff
@@ -3339,17 +3339,8 @@ static void generateHosts(IPropertyTree * storage, GroupInfoArray & groups)
     }
 }
 
-
+static CConfigUpdateHook configUpdateHook;
 static std::atomic<unsigned> normalizeHostGroupUpdateCBId{(unsigned)-1};
```
ghalliday, line 3343: no longer used

system/jlib/jptree.cpp
```diff
@@ -8677,6 +8687,24 @@ void removeConfigUpdateHook(unsigned notifyFuncId)
...
+{
+    unsigned id = configCBId.exchange((unsigned)-1);
+    if ((unsigned)-1 != id)
+        removeConfigUpdateHook(id);
+}
+
+void CConfigUpdateHook::installOnce(ConfigUpdateFunc callbackFunc, bool callWhenInstalled)
+{
+    unsigned uninitialized = (unsigned)-1;
+    if (configCBId.compare_exchange_strong(uninitialized, 0))
```
ghalliday, line 8700: One slight problem with this function is that it can return before the callbackfunc has been called if it can be called concurrently on multiple threads.  That probably isn't what is expected/desired.
I don't think it will cause any issue with the current code (TpWrapper is the only potential).  Adding a critsec to the class is a resonable amount of overhead.  Sharing one between all instances is robably ok, since only used if not already installed.

(It would be slightly more efficient to read the value first and only perform the compare_exchange_strong if equal -1.)

ghalliday (commented):
A couple of minor comments.  In general the code looks cleaner.

system/jlib/jptree.cpp
jakesmith: right, yes there's a window it can return before called, I agree that's a problem.
I don't see an issue with adding a CS to this class to protect against that though.

jakesmith:

jakesmith:
@ghalliday - squashed

ghalliday (approved):

PR #14985, opened 2021-05-25
HPCC-25749 Keep track of the count of every read and write to a file

shamser:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/thorlcr/activities/indexwrite/thindexwrite.cpp
+++ b/thorlcr/activities/indexwrite/thindexwrite.cpp
@@ -279,6 +279,22 @@ public:
                 VStringBuffer pval("%f", info->getBloomProbability());
                 bloom->setProp("@bloomProbability", pval.str());
             }
+            // Update numDiskWrites
+            StatsScopeId rootScope(SSTfile, 0U);
+            Owned<IStatisticGatherer> fileStats;
+            rootScope.setFileId(fileName);
+            fileStats.setown(createStatisticsGatherer(SCTall, "", rootScope));
+            statsCollection.getStats(*fileStats);
+            Owned<IStatisticCollection> stats = fileStats->getResult();
+            Owned<IStatisticCollectionIterator> iter = &stats->getScopes(nullptr, false);
+            __int64 numDiskWrites = 0;
+            ForEach(*iter)
+            {
+                IStatisticCollection & cur = iter->query();
+                numDiskWrites += cur.queryStatistic(StNumDiskWrites);
+            }
+            props.setPropInt64("@numDiskWrites", numDiskWrites);
+
             container.queryTempHandler()->registerFile(fileName, container.queryOwner().queryGraphId(), 0, false, WUFileStandard, &clusters);
             if (!dlfn.isExternal())
                 queryThorFileManager().publish(container.queryJob(), fileName, *fileDesc);
--- a/thorlcr/activities/thdiskbase.cpp
+++ b/thorlcr/activities/thdiskbase.cpp
@@ -192,6 +192,7 @@ void CWriteMasterBase::init()
         if (blockCompressed)
             props.setPropBool("@blockCompressed", true);
         props.setProp("@kind", "flat");
+        props.setPropInt64("@numDiskWrites", 0);
         if (((TAKdiskwrite == container.getKind()) || (TAKspillwrite == container.getKind())) &&
                 (0 != (diskHelperBase->getFlags() & TDXtemporary)) && container.queryOwner().queryOwner() && (!container.queryOwner().isGlobal())) // I am in a child query
         { // do early, because this will be local act. and will not come back to master until end of owning graph.
@@ -204,11 +205,28 @@ void CWriteMasterBase::publish()
 {
     if (published) return;
     published = true;
+
+    const char *logicalFileName = dlfn.get();
+    StatsScopeId rootScope(SSTfile, 0U);
+    Owned<IStatisticGatherer> fileStats;
+    rootScope.setFileId(logicalFileName);
+    fileStats.setown(createStatisticsGatherer(SCTall, "", rootScope));
+    statsCollection.getStats(*fileStats);
+    Owned<IStatisticCollection> stats = fileStats->getResult();
+    Owned<IStatisticCollectionIterator> iter = &stats->getScopes(nullptr, false);
+    __int64 numDiskWrites = 0;
+    ForEach(*iter)
+    {
+        IStatisticCollection & cur = iter->query();
+        numDiskWrites += cur.queryStatistic(StNumDiskWrites);
+    }
+
     if (!(diskHelperBase->getFlags() & (TDXtemporary|TDXjobtemp)))
         updateActivityResult(container.queryJob().queryWorkUnit(), diskHelperBase->getFlags(), diskHelperBase->getSequence(), fileName, recordsProcessed);
 
     IPropertyTree &props = fileDesc->queryProperties();
     props.setPropInt64("@recordCount", recordsProcessed);
+    props.setPropInt64("@numDiskWrites", numDiskWrites);
     if (0 == (diskHelperBase->getFlags() & TDXtemporary) || container.queryJob().queryUseCheckpoints())
     {
         if (0 != (diskHelperBase->getFlags() & TDWexpires))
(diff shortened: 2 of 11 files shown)


shamser:
This tracks the count of most read and writes to a logical file.  It does not track the number of reads from an index file (this will be done in separate jira).  The code has been tested with by examining the logical file's attributes using daliadmin <dali> dfsfile.

thorlcr/activities/thdiskbase.cpp
```diff
@@ -204,11 +205,28 @@ void CWriteMasterBase::publish()
 {
     if (published) return;
     published = true;
+
+    const char *logicalFileName = dlfn.get();
+    StatsScopeId rootScope(SSTfile, 0U);
+    Owned<IStatisticGatherer> fileStats;
+    rootScope.setFileId(logicalFileName);
+    fileStats.setown(createStatisticsGatherer(SCTall, "", rootScope));
```
jakesmith, line 213: not sure you need nesting as this write master knows the filename already, so think serialize stat. from each slave can be non-nested.

thorlcr/activities/thdiskbaseslave.cpp
```diff
@@ -299,6 +299,13 @@ void CDiskReadSlaveActivityBase::serializeStats(MemoryBuffer &mb)
         stats.merge(activeStats);
     }
     stats.setStatistic(StNumDiskRowsRead, diskProgress);
+    if (!isEmptyString(logicalFilename))
+    {
+        unsigned __int64 numDiskReads = activeStats.getStatisticValue(StNumDiskReads);
+        StatsScopeId scope(SSTfile, logicalFilename);
```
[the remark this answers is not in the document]
jakesmith: It's less of a concern if these are only serialized when done (because only used at moment when graph is done).
However, volume of stats could still be an issue, if these acts are in rapidly repeatedly executing child queries.

jakesmith (commented):
@shamser - a few comments/questions - mainly about efficiency/frequency of serialization.

Is there a separate JIRA to cover the other activities that read files?
e.g. IndexRead, KeyedJoin , Fetch , IndexDistribute

Would also be worth opening JIRA to implement same in hthor and roxie.

shamser:
> a few comments/questions - mainly about efficiency/frequency of serialization.
> 
> Is there a separate JIRA to cover the other activities that read files?
> e.g. IndexRead, KeyedJoin , Fetch , IndexDistribute
> 
> Would also be worth opening JIRA to implement same in hthor and roxie.

I've submitted this PR with partial implementation (1) to check that the I'm on the right track as far as the implementation (2) to break up a large change set into a number of smaller PRs

thorlcr/activities/indexwrite/thindexwrite.cpp
```diff
@@ -279,6 +279,22 @@ class IndexWriteActivityMaster : public CMasterActivity
                 VStringBuffer pval("%f", info->getBloomProbability());
                 bloom->setProp("@bloomProbability", pval.str());
             }
+            // Update numDiskWrites
+            StatsScopeId rootScope(SSTfile, 0U);
+            Owned<IStatisticGatherer> fileStats;
+            rootScope.setFileId(fileName);
+            fileStats.setown(createStatisticsGatherer(SCTall, "", rootScope));
```
ghalliday, line 286: A couple of issues.  You would normally want to use a CRuntimeStatisticCollection to serialize information from a slave to the master.  the createStatisticsGatherer() is generally used once the stats are going to be published.
Also SCTall is not correct - in this case it would probably be SCTthor.
I also agree with Jake's comment below.

ghalliday: @jakesmith I'm slightly surprised the information isn't already serialized in the activity stats.

thorlcr/activities/thdiskbase.cpp
ghalliday, line 213: Similar code to the index write above

thorlcr/activities/thdiskbaseslave.cpp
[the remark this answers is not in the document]
ghalliday: @jakesmith how do superfiles fit in to this?

thorlcr/graph/thgraphmaster.cpp
```diff
@@ -540,9 +540,26 @@ void CMasterActivity::getEdgeStats(IStatisticGatherer & stats, unsigned idx)
...
+
+    StatsScopeId rootScope(SSTfile, 0U);
+    Owned<IStatisticGatherer> fileStats;
     ForEachItemIn(s, readFiles)
     {
         IDistributedFile &file = readFiles.item(s);
+        StringBuffer fileName;
+        file.getLogicalName(fileName);
+        rootScope.setFileId(fileName.str());
+        fileStats.setown(createStatisticsGatherer(SCTall, "", rootScope));
```
ghalliday, line 552: There should probably be different helper functions to allow you to walk through CRuntimeStatistics

ghalliday (commented):
@shamser a few comments.  If you do need to created nested items for the disk read size then talk to me about changes to the base classes to avoid the need for an IStatisticGatherer.

jakesmith:
@shamser  @ghalliday -

**Frequency of sending these stats.**

I think it was discussed, and decided that as long as abort could be relied on to send stats., then updating once at end was ok.
Meaning, if you watch a long running job, then you will not see these new stats, until it has finished the subgraph.
Is that the conclusion we came to in the meeting we had together?

**Persistence of these stats**

We touched on auditing this info. to logging, so it's persistent and lives beyond the life of the logical file meta data.
I'm not sure we came to a final decision though?
If we can audit this to logging, similar to how we already audit other file read/written info, that attributes the file with owner,workunit etc, then I think it will allow ELK to find it by file, or owner, or job etc.
Then do we need to publish it to dali meta data per logical file at all, i.e. should it just be logged with appropriate audience, info.?
I suspect that can be left to next stage, but thoughts at this juncture?

ghalliday:
> @shamser @ghalliday -
> 
> **Frequency of sending these stats.**
> 
> I think it was discussed, and decided that as long as abort could be relied on to send stats., then updating once at end was ok.
> Meaning, if you watch a long running job, then you will not see these new stats, until it has finished the subgraph.
> Is that the conclusion we came to in the meeting we had together?
I think so.  The information will not be published until the subgraph has completed - so there is no need to send it until the end of the subgraph/abort.

> **Persistence of these stats**
> 
> We touched on auditing this info. to logging, so it's persistent and lives beyond the life of the logical file meta data.
> I'm not sure we came to a final decision though?
> If we can audit this to logging, similar to how we already audit other file read/written info, that attributes the file with owner,workunit etc, then I think it will allow ELK to find it by file, or owner, or job etc.
> Then do we need to publish it to dali meta data per logical file at all, i.e. should it just be logged with appropriate audience, info.?
> I suspect that can be left to next stage, but thoughts at this juncture?
My recollection is that we should publish the information so that we can quickly display the cost of a file/workunit, but we should also be logging the costs to an audit file so that we can account for items that are deleted etc.  

I don't think we agreed any details of the auditing - although we did talk about logging the raw information used to calculate the cost as well as the calculated cost.  Possibly logging with a message audience of Audit, and a (new) category of cost.  The audit log is something to return to later.

jakesmith:
>I think so. The information will not be published until the subgraph has completed - so there is no need to send it until the end of the subgraph/abort.

If the workers do not send until the end it will mean that on occasion long running (and expensive) job metrics that go 'bad' (e.g. crash) will be lost.
We could still push the data from the slave to the master periodically (along with other stats), but only publish at end, or on abort.
The master is less likely to suffer a fatal fault, and more likely to be able push final state.

So, as long as the serialized data isn't too verbose, I think I'd advocate that it was serialized by the workers along with other stats, but published by master at end.

jakesmith:
> I don't think we agreed any details of the auditing - although we did talk about logging the raw information used to calculate the cost as well as the calculated cost. Possibly logging with a message audience of Audit, and a (new) category of cost. The audit log is something to return to later.

Agreed, something to revisit.

shamser:
I'm breaking this Jira into smaller tasks and closing this.

shamser:
I'm breaking this Jira into smaller tasks and closing this.

PR #14987, opened 2021-05-26
HPCC-25952 Allow planes to create implicit persistent volume claims

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:


ghalliday:
HPCC-25959 contains the documentation.
The PR is suitable for initial review - but needs more testing before ready to merge (e.g., the forcePermissions options).  Further PRs will add spill configuration and create the planes from the daliStorage (and remove that as the preferred way of configuring).

ghalliday:
@jakesmith I think this is ready for review.
@richardkchapman How can we improve the automated testing.  I think it might be useful to have a test that brings up a k8s system with various different values files, but I don't have any good ideas how to to achieve that since some of the changes that we want to check depend on the execution environment.

helm/hpcc/templates/datastorage-pvc.yaml
```diff
@@ -3,3 +3,5 @@
 {{- if (and (not $dataStorage.existingClaim) (not $dataStorage.plane)) }}
 {{ include "hpcc.addPVC" (dict "root" $ "name" "datastorage" "me" $dataStorage) }}
 {{- end }}
+{{- include "hpcc.addPVCsFromPlanes" (dict "root" $ "label" "data") }}
+{{- include "hpcc.addPVCsFromPlanes" (dict "root" $ "label" "lz") }}
```
jakesmith, line 7: rather than have a hpcc.addPVCsFromPlanes in each data/dll/dalistorage (feels like these files should be retired) for some specific labels (e.g. "data" and "lz" here in datastorage-pvc.yaml),
would it be better to have a storage.yaml, that creates PVC for all planes defined regardless of labels, as long as they define storageClass ?

helm/hpcc/templates/_helpers.tpl
```diff
@@ -905,6 +928,7 @@ Pass in dict with root and me
 {{- $serviceName := printf "sasha-%s" .me.name -}}
 {{- if .me.storage }}
 {{ include "hpcc.addVolume" (dict "root" .root "name" $serviceName "me" .me.storage) -}}
+{{ include "hpcc.addVolumes" (dict "root" .root "includeLabels" (list "sasha") ) }}
```
jakesmith, line 931: prior to this change, each sasha service might only mount 1 pvc, now using planes and "sasha" label, it would mount all, which would be a problem if ReadWriteOnce type.

If the/a, service names a plane, and it's a pvc, I think should only mount that one

helm/hpcc/templates/_helpers.tpl
```diff
@@ -236,7 +237,7 @@ Add data volume mount
 Pass in root
 */}}
 {{- define "hpcc.addDataVolumeMount" -}}
-{{- include "hpcc.addVolumeMounts" (dict "root" .root "includeLabels" (list "data" "")) -}}
+{{- include "hpcc.addVolumeMounts" (dict "root" .root "includeLabels" (list "data" "" "lz")) -}}
```
jakesmith, line 240: This change means all engines will auto mount/expose lz planes.
Is that intended/needed/related to this PR?

jakesmith (commented):
@ghalliday - a couple of questions

helm/hpcc/templates/_helpers.tpl
ghalliday: It was intended - so that they can use the plane::x:: syntax to directly read data from the landing zones - otherwise you would have to spray before you can read data from a lz.  I'm happy to debate whether this is a sensible change...

jakesmith: ok, reasoning makes sense.
It may be more that we want to restrict access to LZ's by default, but by user authorization of planes.

helm/hpcc/templates/datastorage-pvc.yaml
ghalliday: I think that can/should be a separate pr.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -928,7 +931,7 @@ Pass in dict with root and me
 {{- $serviceName := printf "sasha-%s" .me.name -}}
 {{- if .me.storage }}
 {{ include "hpcc.addVolume" (dict "root" .root "name" $serviceName "me" .me.storage) -}}
-{{ include "hpcc.addVolumes" (dict "root" .root "includeLabels" (list "sasha") ) }}
+{{ include "hpcc.addVolumes" (dict "root" .root "includeLabels" (list "sasha") "name" .me.storage.plane ) }}
```
jakesmith, line 934: might this fail if no .me.storage dict ? i.e. does it need to have a $storage := .me.storage | default dict

helm/hpcc/templates/dali.yaml
```diff
@@ -124,7 +124,7 @@ kind: ConfigMap
 {{- if and (not $sasha.storage.existingClaim) (not $sasha.storage.plane) }}
 {{ include "hpcc.addPVC" (dict "root" $ "name" (printf "sasha-%s" $sashaName) "me" $sasha.storage) }}
 {{- end }}
-{{- include "hpcc.addPVCsFromPlanes" (dict "root" $ "label" "sasha") }}
+{{- include "hpcc.addPVCsFromPlanes" (dict "root" $ "label" "sasha" "name" $sasha.storage.plane) }}
```
jakesmith, line 127: does this to check $sasha.storage is valid first (and default to empty dict to protect)

ghalliday: It does - in line 123.

helm/hpcc/templates/_helpers.tpl
ghalliday: no, it works.  It sets it to a null value.

ghalliday: (I tested with and without it set.  I need to add some test cases though...)

helm/hpcc/templates/dali.yaml
jakesmith: ok, missed that, thanks

jakesmith (approved):
@ghalliday - looks good, approving, but needs squashing.

jakesmith:
@ghalliday - why did eclhelper check fail (complaining about getLogTailText), perhaps ran before it was rebased or something?

ghalliday:
@jakesmith please check the last commit.  The system starts up with the two new test files supplied as values files.  The sasha code is not right - I think it would be better if you tackled that - and I suspect as you suggest moving all the pvc creation out into a separate storage.yaml would be cleaner.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -45,19 +45,47 @@ Translate a port list to a comma-separated list
...
+{{- define "hpcc.getFirstPlaneForLabel" -}}
+{{- $root := .root -}}
+{{- $label := .label -}}
+{{- $storage := ($root.Values.storage | default dict) -}}
+{{- $planes := ($storage.planes | default list) -}}
+{{- $firstPlane := dict -}}
+{{- range $plane := $planes -}}
+{{- $labels := $plane.labels | default (list "data") -}}
+{{- if (has $label $labels) -}}
+{{- $_ := set $firstPlane "plane" $plane.name -}}
```
jakesmith, line 57: may be missing something, shouldn't something be checking if set from previous iteration, otherwise $firstPlane.plane is going to be set again and become the last matching plane, not first.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -45,19 +45,47 @@ Translate a port list to a comma-separated list
...
 {{- $storage := ($.Values.storage | default dict) -}}
-{{- $planes := ($storage.planes | default list) -}}
-{{- $firstPlane := dict -}}
-{{- range $plane := $planes -}}
- {{- if and (not $firstPlane.plane) (or (not $plane.labels) (has "data" $plane.labels)) }}
- {{- $_ := set $firstPlane "plane" $plane.name -}}
- {{- end -}}
+{{- $dataStorage := ($storage.dataStorage | default dict) -}}
+{{- $firstPlane := (include "hpcc.getFirstPlaneForLabel" (dict "root" $ "label" "data")) -}}
+{{- $dataStorage.plane | default $firstPlane | default "hpcc-data-plane" -}}
```
jakesmith, line 78: all feels a little too complicated still in code/template/auto.

Perhaps would be simpler not to default to a implicit "hpcc-data-plane" if nothing defined,
but require at least one (can be ephemeral) and for default data planes (and same for other storage types) to be required.

The code here, assumptions, documentation and explanation would be simpler I think.

helm/hpcc/values.yaml
```diff
@@ -79,6 +79,14 @@ security:
 
 ## storage:
 ##
+## 1. If a component has the storagePlane property set, then that plane will be the default data location for that component.
+## 2. If storage.dataStorage.plane is defined then that plane will be the default location
+## 3. If there is a plane definition with no label or a label of "data" then the first matching plane will be the default data location
+## 4. Otherwise the default data location will be derived from the properties of storage.dataStorage.
```
jakesmith, line 85: I'm not sure 3 and 4 are helpful, I think they may complicate and confuse.
If we make it a requirement that there the global default data plane must be defined storage.dataStorage.plane, that that plane must exist. I think that will be clearer and template code simpler.

helm/hpcc/templates/sasha.yaml
```diff
@@ -52,6 +52,7 @@ kind: ConfigMap
 {{- if and (not $sasha.storage.existingClaim) (not $sasha.storage.plane) }}
```
jakesmith, line 52: I think sasha should have an [optional] storagePlane, that must exist.
If ephemeral it a PVC will have been created by addPVCsFromPlanes

jakesmith (commented):
@ghalliday - a few comments/questions

helm/hpcc/templates/_helpers.tpl
ghalliday: yes you're correct, I will fix.

helm/hpcc/templates/_helpers.tpl
ghalliday: I think that is a good goal - but we will break most existing values files if we make that change - so I would delay that until we are confident this method works (e.g., all examples and our own values file are moved over to defining planes).

helm/hpcc/values.yaml
ghalliday: (3) is could be avoided.  Same also applies to dali, dlls spill etc..  It avoids having to potentially specify the same thing in two different places.
(4) is needed until all examples and documentation switches over to defining a list of planes.

helm/hpcc/templates/sasha.yaml
ghalliday: I think I will leave sasha to you - it seems to behave differently from the other examples.

ghalliday:
@jakesmith squashed and rebased previous changes.  New commit fixing the firstplane issues.  Responded to your other comments.

jakesmith (approved):
@ghalliday - looks ready bar squashing

ghalliday:
@richardkchapman squashed.  Please review/sanity check/merge

helm/hpcc/templates/_helpers.tpl
```diff
@@ -68,19 +68,49 @@ Translate a port list to a comma-separated list
  {{- end -}}
 {{- end -}}
 
+{{- define "hpcc.getFirstPlaneForLabel" -}}
+{{- $root := .root -}}
```
richardkchapman, line 72: style: Not sure what you gain by this rather than just using .root in the code below?

helm/hpcc/templates/_helpers.tpl
```diff
@@ -68,19 +68,49 @@ Translate a port list to a comma-separated list
...
+{{- end -}}
+{{- end -}}
+{{- if $firstPlane.plane -}}
+{{- $firstPlane.plane -}}
+{{- end -}}
+{{- end -}}
+
+{{- define "hpcc.hasPlaneForLabel" -}}
+{{- if (include "hpcc.getFirstPlaneForLabel" .) -}}
+true
```
richardkchapman, line 92: I think this will end up as a string rather than a boolean value if called via include, as you do elsewhere...

helm/hpcc/templates/_helpers.tpl
```diff
@@ -121,20 +151,24 @@ storage:
...
 {{- if not $dataStorage.plane }}
+{{- if not (include "hpcc.hasPlaneForLabel" (dict "root" $ "label" "data")) }}
   - name: hpcc-data-plane
     labels:
     - data
     prefix: {{ .Values.global.defaultDataPath | default "/var/lib/HPCCSystems/hpcc-data" | quote }}
 {{- end }}
+{{- end }}
 {{- if not $spillStorage.plane }}
+{{- if not (include "hpcc.hasPlaneForLabel" (dict "root" $ "label" "spill")) }}
```
richardkchapman, line 167: Does this work? hasPlaneForLabel will return empty string or "true" rather than a boolean, but perhaps that's ok for an if ...

helm/hpcc/templates/_helpers.tpl
```diff
@@ -590,6 +638,26 @@ A kludge to ensure mounted storage (e.g. for nfs, minikube or docker for desktop
...
+*/}}
+{{- define "hpcc.changePlaneMountPerms" -}}
+{{- $storage := (.root.Values.storage | default dict) -}}
+{{- $planes := ($storage.planes | default list) -}}
+  {{- $includeLabels := .includeLabels | default list -}}
+{{- range $plane := $planes -}}
+ {{- if and ($plane.forcePermissions) (or ($plane.pvc) (hasKey $plane "storageClass")) -}}
+  {{- $mountpath := $plane.prefix -}}
+  {{- $matchedLabels := include "hpcc.doesStorageLabelsMatch" (dict "plane" $plane "includeLabels" $includeLabels) }}
+  {{- if ne $matchedLabels "" }}
```
richardkchapman, line 652: If you have to compare vs "" here it suggests the earlier code I flagged may be iffy!

richardkchapman (commented):
Some style issues and a slight concern over template functions returning booleans

richardkchapman (approved):

PR #14996, opened 2021-06-01
HPCC-25972 Fix initial set of dependencies on queryEnvironmentConf()

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dali/server/daserver.cpp
+++ b/dali/server/daserver.cpp
@@ -665,6 +665,7 @@ int main(int argc, const char* argv[])
 // Audit logging
         StringBuffer auditDir;
         {
+            //MORE: Does this need to change in CONTAINERIZED mode?
             Owned<IComponentLogFileCreator> lf = createComponentLogFileCreator(serverConfig, "dali");
             lf->setLogDirSubdir("audit");//add to tail of config log dir
             lf->setName("DaAudit");//override default filename
--- a/fs/dafsclient/rmtclient.cpp
+++ b/fs/dafsclient/rmtclient.cpp
@@ -91,11 +91,15 @@ static unsigned dafsConnectFailRetryTimeMs = defaultDafsConnectFailRetrySeconds 
 
 MODULE_INIT(INIT_PRIORITY_DAFSCLIENT)
 {
+#ifdef _CONTAINERIZED
+    //MORE: This function is called too soon to read them from the configuration file.
+#else
     const IProperties &confProps = queryEnvironmentConf();
     dafsConnectTimeoutMs = confProps.getPropInt("dafsConnectTimeoutSeconds", defaultDafsConnectTimeoutSeconds) * 1000;
     dafsConnectRetries = confProps.getPropInt("dafsConnectRetries", defaultDafsConnectRetries);
     dafsMaxReceiveTimeMs = confProps.getPropInt("dafsMaxReceiveTimeSeconds", defaultDafsMaxRecieveTimeSeconds);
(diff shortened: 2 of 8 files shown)


ghalliday:
@jakesmith I suspect there are more, but this is an initial set of changes to look at.  I suspect you will need better fixes when you integrate dafilesrv....

ghalliday:
@richardkchapman you may prefer a smaller fix which removes the errors, and then have this as a separate PR that addresses some of the reasons it was failing.

ghalliday:
How could this regression have been avoided?
- Checking that container version had been built and run
- Me being clearer on what stage the PR was at.

dali/server/daserver.cpp
```diff
@@ -665,6 +665,7 @@ int main(int argc, const char* argv[])
 // Audit logging
         StringBuffer auditDir;
         {
+            //MORE: Does this need to change in CONTAINERIZED mode?
```
jakesmith, line 668: I think it should just output to standard logging (stderr).
The audience field should be enough to discriminate it when filtering/viewing in ES etc.

fs/dafsclient/rmtclient.cpp
```diff
@@ -91,11 +91,15 @@ static unsigned dafsConnectFailRetryTimeMs = defaultDafsConnectFailRetrySeconds
 
 MODULE_INIT(INIT_PRIORITY_DAFSCLIENT)
 {
+#ifdef _CONTAINERIZED
+    //MORE: This function is called too soon to read them from the configuration file.
```
jakesmith, line 95: there is no clear use-case at the moment for a dafilesrv client inside HPCC (only outside from a different client, e.g. Java/Spark).
So at the moment the whole of rmtclient.cpp/rmtfile.cpp - could be compiled out if we wanted I think.
But, we may want to go via dafilesrv client connection in future for some reason, e.g. routing through for security reasons

roxie/ccd/ccdmain.cpp
```diff
@@ -771,12 +771,14 @@ int CCD_API roxie_main(int argc, const char *argv[], const char * defaultYaml)
...
 
         Linked<IPropertyTree> directoryTree = topology->queryPropTree("Directories");
+#ifndef _CONTAINERIZED
         if (!directoryTree)
         {
             Owned<IPropertyTree> envFile = getHPCCEnvironment();
             if (envFile)
                 directoryTree.set(envFile->queryPropTree("Software/Directories"));
         }
+#endif
```
jakesmith, line 781: Don't think it's an option to have "Directories" in roxie config in k8s/helm, so this #ifndef _CONTAINERIZED block may be better to around lines 773 to 795

system/jlib/jutil.cpp
```diff
@@ -2453,7 +2453,10 @@ static CriticalSection envConfCrit;
 jlib_decl const IProperties &queryEnvironmentConf()
 {
 #if defined(_CONTAINERIZED) && defined(_DEBUG)
-    throwUnexpectedX("queryEnvironmentConf() callled from container system");
+    //The following line is currently hit by too many examples.  Re-enable the exception when more
+    //work has been done removing calls to getConfigurationDirectory() and other related functions.
+    //throwUnexpectedX("queryEnvironmentConf() callled from container system");
+    IERRLOG("queryEnvironmentConf() callled from container system");
```
jakesmith, line 2459: trivial: pre-existing typo 'callled'

jakesmith (approved):
@ghalliday - all looks sensible, minor comments only.

PR #15000, opened 2021-06-02
HPCC-25834 Conditionally compile WsDfu code for cloud

wangkx:
1. CThorNodeGroup related code is for bare metal only.
2. Add UNIMPLEMENTED_X for DFUFileAccess, DFUFileCreate, and
DFUFilePublish because they are not ready for cloud. Do not
compile their related code for cloud.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/services/ws_dfu/ws_dfuService.cpp
+++ b/esp/services/ws_dfu/ws_dfuService.cpp
@@ -98,10 +98,10 @@ const unsigned MAX_KEY_ROWS = 20;
 
 short days[12] = {31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31};
 
-
+#ifndef _CONTAINERIZED
+//CThorNodeGroupCache is for bare metal only.
 CThorNodeGroup* CThorNodeGroupCache::readNodeGroup(const char* _groupName)
 {
-#ifndef _CONTAINERIZED
     Owned<IEnvironmentFactory> factory = getEnvironmentFactory(true);
     Owned<IConstEnvironment> env = factory->openEnvironment();
     Owned<IPropertyTree> root = &env->getPTree();
@@ -114,7 +114,6 @@ CThorNodeGroup* CThorNodeGroupCache::readNodeGroup(const char* _groupName)
         if (groupName.length() && strieq(groupName.str(), _groupName))
             return new CThorNodeGroup(_groupName, cluster.getCount("ThorSlaveProcess"), cluster.getPropBool("@replicateOutputs", false));
     }
-#endif
 
     return NULL;
 }
@@ -132,6 +131,7 @@ CThorNodeGroup* CThorNodeGroupCache::lookup(const char* groupName, unsigned time
 
     return e.getClear();
 }
+#endif
 
 void CWsDfuEx::init(IPropertyTree *cfg, const char *process, const char *service)
 {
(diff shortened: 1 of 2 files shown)


afishbeck:
@jakesmith can you give this a quick look as a sanity check since it covers DFS functionality?

esp/services/ws_dfu/ws_dfuService.cpp
```diff
@@ -6093,12 +6102,16 @@ void CWsDfuEx::dFUFileAccessCommon(IEspContext &context, const CDfsLogicalFileNa
...
 }
+#endif
 
 // NB: deprecated from ver >= 1.50
 bool CWsDfuEx::onDFUFileAccess(IEspContext &context, IEspDFUFileAccessRequest &req, IEspDFUFileAccessResponse &resp)
 {
     try
     {
+#ifdef _CONTAINERIZED
+        UNIMPLEMENTED_X("CONTAINERIZED(DFUFileAccess)"); //See JIRA issue HPCC-25172
```
ghalliday, line 6113: Why is this disabled in containerized?  What is the function it is calling that will fail?

esp/services/ws_dfu/ws_dfuService.cpp
```diff
@@ -6450,6 +6478,9 @@ bool CWsDfuEx::onDFUFileCreateV2(IEspContext &context, IEspDFUFileCreateV2Reques
 
 bool CWsDfuEx::onDFUFilePublish(IEspContext &context, IEspDFUFilePublishRequest &req, IEspDFUFilePublishResponse &resp)
 {
+#ifdef _CONTAINERIZED
+    UNIMPLEMENTED_X("CONTAINERIZED(DFUFilePublish)"); //See JIRA issue HPCC-25172
```
ghalliday, line 6482: Why is this disabled? - The jira reference is useful (assuming it is correct), but the comments could do with detailing what function is needed.
Do these exceptions help?  The code should already throw an exception if it accesses something it shouldn't in the containerized builds.  It gives a list of places that should be tested before shipping 8.2, but it also means that if the code that they call is fixed you need to then remove all exceptions like this.  You could end up with code failing just because the exception hasn't been removed.

esp/services/ws_dfu/ws_dfuService.hpp
```diff
@@ -27,6 +27,7 @@
 #include <atomic>
 #include "ws_dfuHelpers.hpp"
 
+#ifndef _CONTAINERIZED
```
ghalliday, line 30: I prefer minimizing the number of #ifdefs - because it means there is less likelihood of breaking the code and not spotting it.  I would be inclined to keep this compiled in both - unless there are functions that call code that is not available (e.g. environment).

ghalliday (changes requested):
@wangkx main comment is that this needs to target 8.2.0 - which it will probably then have some clashes (since some of the functions have already been changed).

wangkx:
I will re-target it to 8.2.0.

wangkx:
@ghalliday the PR has been re-targeted to 8.2.0. The original changes were started before https://github.com/hpcc-systems/HPCC-Platform/commit/adef7a318f86d2c23047fa091f1e1f2bcd3ffc4c is merged. After read your comments, I think this PR can be closed. That is because it does not fix anything and not give more information.

ghalliday:
Unfortunately I think you are correct  I will close it.

PR #15005, opened 2021-06-02
HPCC-25950 Institute lower default log level

rpastrana:
- Updates global default log level
- Updates helm chart global default log level

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dali/sasha/saserver.cpp
+++ b/dali/sasha/saserver.cpp
@@ -298,8 +298,6 @@ static constexpr const char * defaultYaml = R"!!(
 version: 1.0
 sasha:
   name: sasha
-  logging:
-    detail: 100
 )!!";
 
 int main(int argc, const char* argv[])
--- a/dali/server/daserver.cpp
+++ b/dali/server/daserver.cpp
@@ -398,8 +398,6 @@ version: 1.0
 dali:
   name: dali
   dataPath: "/var/lib/HPCCSystems/dalistorage"
-  logging:
-    detail: 100
 )!!";
 
 
@@ -652,7 +650,7 @@ int main(int argc, const char* argv[])
         Owned<IMPServer> mpServer = getMPServer();
         Owned<IAllowListHandler> allowListHandler = createAllowListHandler(populateAllowListFromEnvironment, formatDaliRole);
         mpServer->installAllowListCallback(allowListHandler);
-        setMsgLevel(fileMsgHandler, serverConfig->getPropInt("SDS/@msgLevel", 100));
+        setMsgLevel(fileMsgHandler, serverConfig->getPropInt("SDS/@msgLevel", DebugMsgThreshold));
 #endif
         startLogMsgChildReceiver();
         startLogMsgParentReceiver();
--- a/helm/examples/logging/README.md
+++ b/helm/examples/logging/README.md
@@ -46,7 +46,7 @@ The available target categories include disaster(DIS), error(ERR), warning(WRN),
     helm install myhpcc ./hpcc --set dali[0].logging.classes="ALL-PRO" --set dali[0].name="mydali"
 
 ### Log Detail Level Configuration
-Log output verbosity can be adjusted from "critical messages only" (1) up to "report all messages" (100). By default, the log level is set highest (100). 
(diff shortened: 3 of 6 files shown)


jakesmith (commented):
@rpastrana - changes look good, but there some other places that currently also have a default of 100, which should be changed at same time.
e.g. the defaultYaml in roxie, daserver and saserver.

I think we should just remove those override setting, to match others, and to pick up global default.

rpastrana:
@jakesmith please review.

rpastrana:
@AttilaVamos the check failure seems unrelated:
"CMake Error at cmake_modules/commonSetup.cmake:756 (MESSAGE):
-- Found Bison v3.5.1
  You need Gnu c++ version 7.3.0 or later to build this project (version
-- Configuring incomplete, errors occurred!"

mckellyln:
The error "You need Gnu c++ version 7.3.0 or later to build this project ..." occurs sometimes on additional commits or a squash.  It is something to do with re-running cmake and its cache - it is not related to anything specific to this PR.
@jakesmith has looked into it and I have as well but it is difficult to reproduce.
@AttilaVamos fyi

dali/dfu/dfuserver.cpp
```diff
@@ -104,6 +104,8 @@ version: "1.0"
   name: dfuserver
   monitorInterval: 900
   maxJobs: 1
+  logging:
+    detail: 100
```
jakesmith, line 108: not sure why this was added?

jakesmith (changes requested):
@rpastrana - There's a couple of other components that override the default (used to be same as default) unnecessarily, that should also be removed.

rpastrana:
> 
> 
> @rpastrana - There's a couple of other components that override the default (used to be same as default) unnecessarily, that should also be removed.

I didn't feel comfortable overriding those components' explicit logging levels because there is evidence (at least in bare metal) they are meant to log at the highest level. 
Looks like you @jakesmith were involved in the sasha baremetal and container logging detail:
https://github.com/hpcc-systems/HPCC-Platform/blame/b895abd1e8d54ae9d6d4a3675037d405b576e4a1/dali/sasha/saserver.cpp#L357
https://github.com/hpcc-systems/HPCC-Platform/blob/b895abd1e8d54ae9d6d4a3675037d405b576e4a1/dali/sasha/saserver.cpp#L357
If you confirm it should default, I'll go ahead and remove.

@richardkchapman can you confirm this Roxie config entry can be removed and allowed to default?
https://github.com/hpcc-systems/HPCC-Platform/blob/b4bdebe56a7c40af5631559bb5d9b9581a5604ca/roxie/roxie/roxie.cpp#L55

jakesmith:
> > @rpastrana - There's a couple of other components that override the default (used to be same as default) unnecessarily, that should also be removed.
> 
> I didn't feel comfortable overriding those components' explicit logging levels because there is evidence (at least in bare metal) they are meant to log at the highest level.
> Looks like you @jakesmith were involved in the sasha baremetal and container logging detail:
> https://github.com/hpcc-systems/HPCC-Platform/blame/b895abd1e8d54ae9d6d4a3675037d405b576e4a1/dali/sasha/saserver.cpp#L357
> https://github.com/hpcc-systems/HPCC-Platform/blob/b895abd1e8d54ae9d6d4a3675037d405b576e4a1/dali/sasha/saserver.cpp#L357
> 
> 
> If you confirm it should default, I'll go ahead and remove.
> @richardkchapman can you confirm this Roxie config entry can be removed and allowed to default?
> https://github.com/hpcc-systems/HPCC-Platform/blob/b4bdebe56a7c40af5631559bb5d9b9581a5604ca/roxie/roxie/roxie.cpp#L55

@rpastrana - please go ahead and remove the sasha one.

Before this PR, these per component overrides (of detail: 100) were effectively no-op's, setting the component default to what the global default already specified. I don't think they were intentional, just there as a sort of self-documenting attribute.
Sasha doesn't want to raise the global logging default above the global logging default.
I don't think anywhere does.

rpastrana:
@jakesmith updated per your suggestions.

jakesmith (approved):
@rpastrana - looks good.

PR #15015, opened 2021-06-04
HPCC-25561 Roxie udp layer needs to handle a server disappearing 

richardkchapman:
## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Cloud-compatibility
  - [x] Premature optimization

## Smoketest:

### DIFF
--- a/roxie/topo/toposerver.cpp
+++ b/roxie/topo/toposerver.cpp
@@ -62,9 +62,15 @@ static void topo_server_usage()
     printf("\n");
 }
 
+struct TopologyEntry
+{
+    unsigned lastSeen = 0;
+    time_t instance = 0;
+};
+
 unsigned traceLevel = 0;
 unsigned topoPort = TOPO_SERVER_PORT;
-std::map<std::string, unsigned> topology;
+std::map<std::string, TopologyEntry> topology;
 StringBuffer cachedResponse;
 StringBuffer cachedDigest;
 bool responseDirty = true;
@@ -73,6 +79,7 @@ unsigned lastTopologyReport = 0;
 const unsigned timeoutCheckInterval = 1000;
 const unsigned heartbeatInterval = 5000;
 const unsigned timeoutHeartbeatInterval = 10000;
+const unsigned removeHeartbeatInterval = 120000;
 const unsigned topologyReportInterval = 60000;
 bool aborted = false;
 Semaphore stopping;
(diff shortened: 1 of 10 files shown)


richardkchapman:
Marking this ready for review as it was nearly forgotten about as a draft...

It handles - and has been tested for - cases where a server restarts on same IP as before. I need to confirm behaviour when a server disappears and does not come back (e.g. on an austoscaling event).

richardkchapman:
@ghalliday Ready for review. I have tested the autoscale case too

roxie/udplib/udptopo.cpp
```diff
@@ -184,18 +193,38 @@ CTopologyServer::CTopologyServer(const char *topologyInfo)
...
+                    DBGLOG("Deleting pending data for server %s which has terminated or restarted", ep.getUrlStr(s).str());
+                    ROQ->abortPendingData(ep);
+                }
+                if (instance)
+                {
+                    servers[ep.port].append(ep);
+                    serverInstances[ep] = instance;
+                }
+                else
+                    serverInstances.erase(ep);
```
ghalliday, line 223: I don't think this will be there anyway - because you are creating a new entry?

roxie/udplib/udptopo.cpp
```diff
@@ -393,17 +436,28 @@ void TopologyManager::setRoles(const std::vector<RoxieEndpointInfo> &myRoles)
...
 }
 
-void TopologyManager::closedown()
+void TopologyManager::closedown(const std::vector<RoxieEndpointInfo> &myRoles)
 {
-    topoBuf.replaceString("server|", "-server|");
-    topoBuf.replaceString("agent|", "-agent|");
+    topoBuf.clear();
+    _setRoles(myRoles, true);
+    myInstance = 0;
```
ghalliday, line 459: I'm not sure why there are two calls.  Could do with more explanation.

ghalliday (commented):
@richardkchapman I think it is ok, but could do with some more comments to explain some details.

roxie/topo/toposerver.cpp
```diff
@@ -142,16 +150,22 @@ void timeoutTopology()
...
-        unsigned lastSeen = it->second;
+        unsigned lastSeen = it->second.lastSeen;
         if (now-lastSeen > timeoutHeartbeatInterval)
         {
             if (traceLevel)
             {
                 DBGLOG("No heartbeat for %u ms for %s", now-lastSeen, it->first.c_str());
             }
-            it = topology.erase(it);
+            if (now-lastSeen > removeHeartbeatInterval)
```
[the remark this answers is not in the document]
richardkchapman: Clearing the instance means that the clients will be easily able to detect that this server has disappeared, rather than having to detect the fact by comparing the new list of servers it gets with the list of servers it knows about.

Once sufficient time has elapsed that everyone should know it has gone, we remove it entirely so that the size of the information doesn't keep growing if servers are continuously replaced.

roxie/udplib/udptopo.cpp
richardkchapman: The first one generates the text to ask the server to remove the old information. The second generates the text to give toposerver the new information (with instance=0).

roxie/udplib/udptopo.cpp
richardkchapman: I suspect you are right

roxie/udplib/udptopo.cpp
richardkchapman: Could probably drop the second if the server treated a "remove" as setting the instance to 0

ghalliday (approved):
@richardkchapman please squash

richardkchapman:
@ghalliday Squashed

PR #15017, opened 2021-06-07
HPCC-25819 Implement WsFileIO.CheckServerAccess in container mode

wangkx:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/services/ws_fileio/CMakeLists.txt
+++ b/esp/services/ws_fileio/CMakeLists.txt
@@ -58,6 +58,7 @@ include_directories (
          ./../../smc/SMCLib 
          ./../../bindings/SOAP/xpp 
          ${HPCC_SOURCE_DIR}/common/thorhelper
+         ${HPCC_SOURCE_DIR}/dali/sasha
     )
 
 ADD_DEFINITIONS( -D_USRDLL )
@@ -74,6 +75,7 @@ target_link_libraries ( ws_fileio
          ${XALAN_LIBRARIES} ${XERCES_LIBRARIES}
          xmllib 
          esphttp 
+         SMCLib
     )
 
 if (NOT CONTAINERIZED)
--- a/esp/services/ws_fileio/ws_fileioservice.cpp
+++ b/esp/services/ws_fileio/ws_fileioservice.cpp
@@ -30,15 +30,45 @@
 
 void CWsFileIOEx::init(IPropertyTree *cfg, const char *process, const char *service)
 {
+    CTpWrapper tpWrapper;
+    tpWrapper.getTpDropZones(9999, nullptr, false, allTpDropZones); //version 9999: get the latest information about dropzone
 }
 
-bool CWsFileIOEx::CheckServerAccess(const char* targetDZNameOrAddress, const char* relPath, StringBuffer& netAddr, StringBuffer& absPath)
+bool CWsFileIOEx::CheckServerAccess(const char* targetDZNameOrAddress, const char* netAddrReq, const char* relPath, StringBuffer& netAddr, StringBuffer& absPath)
 {
     if (!targetDZNameOrAddress || (targetDZNameOrAddress[0] == 0) || !relPath || (relPath[0] == 0))
         return false;
 
 #ifdef _CONTAINERIZED
-    UNIMPLEMENTED_X("CONTAINERIZED(CWsFileIOEx::CheckServerAccess)");
+    bool isIp4Req = isIPAddress(netAddrReq);
+    ForEachItemIn(i, allTpDropZones)
+    {
+        IConstTpDropZone& dropZone = allTpDropZones.item(i);
+        if (!dropZone.getECLWatchVisible())
+            continue;
+
+        const char* name = dropZone.getName();
+        if (isEmptyString(name) || !streq(targetDZNameOrAddress, name))
+            continue;
+
+        const char* prefix = dropZone.getPath();
+        if (isEmptyString(prefix))
+            continue;
+
+        IArrayOf<IConstTpMachine>& tpMachines = dropZone.getTpMachines();
+        ForEachItemIn(ii, tpMachines)
+        {
+            IConstTpMachine& tpMachine = tpMachines.item(ii);
+            if (!isEmptyString(netAddrReq) && !matchNetAddressRequest(netAddrReq, isIp4Req, tpMachine))
+                continue;
+
+            netAddr.set(tpMachine.getNetaddress());
+            absPath.set(prefix);
+            addPathSepChar(absPath);
+            absPath.append(relPath);
+            return true;
+        }
+    }
 #else
     netAddr.clear();
     Owned<IEnvironmentFactory> factory = getEnvironmentFactory(true);
@@ -120,9 +150,9 @@ bool CWsFileIOEx::CheckServerAccess(const char* targetDZNameOrAddress, const cha
         }
 
     }
+#endif
 
     return false;
-#endif
 }
 
 bool CWsFileIOEx::onCreateFile(IEspContext &context, IEspCreateFileRequest &req, IEspCreateFileResponse &resp)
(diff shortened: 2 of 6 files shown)


esp/smc/SMCLib/TpCommon.cpp
```diff
@@ -97,3 +97,16 @@ extern TPWRAPPER_API IStringIterator* getContainerTargetClusters(const char* pro
     }
     return ret.getClear();
 }
+
+extern TPWRAPPER_API StringBuffer& readNetAddressRequest(const char* addressReg, StringBuffer& address)
+{
+    if (isEmptyString(addressReg))
```
afishbeck, line 103: Does this mean the drop zone address has to be an IP and not a DNS name?

esp/services/ws_fileio/ws_fileioservice.cpp
```diff
@@ -149,7 +181,7 @@ bool CWsFileIOEx::onCreateFile(IEspContext &context, IEspCreateFileRequest &req,
 
     StringBuffer destAbsPath;
     StringBuffer destNetAddr;
-    if (!CheckServerAccess(server, destRelativePath, destNetAddr, destAbsPath))
+    if (!CheckServerAccess(server, req.getDestNetAddress(), destRelativePath, destNetAddr, destAbsPath))
```
afishbeck, line 184: @wangkx can you explain the difference between the old "name or address way" vs. the separate name and address way?  It might have been a little confusing but wouldn't looking at name first and then address still work?

esp/smc/SMCLib/TpCommon.cpp
wangkx: No, the drop zone address can be a DNS name.

afishbeck: I could be missing something.  But, the reason I ask is that if you configure the drop zone with a DNS name, but then try and look it up by DNS name, this routine seems to convert to IP before comparison.  Have you tested with configured host name and lookup by host name?

wangkx: If the drop zone with a DNS name is configured, the DNS name will be converted to IP (both the DNS name and the IP are stored into the IEspTpMachine object) when the DNS name is read from the configuration. Please read the createHostTpMachine() in TpContainer.cpp for details. When a DNS name is looked up, it is converted into IP here before comparing it with the IP from the IEspTpMachine object.

wangkx:
@afishbeck the PR has been rebased to 8.2.0 (the first commit) and revised (the second commit) based on the review comment. Please re-review.

esp/smc/SMCLib/TpCommon.cpp
```diff
@@ -97,3 +97,12 @@ extern TPWRAPPER_API IStringIterator* getContainerTargetClusters(const char* pro
     }
     return ret.getClear();
 }
+
+extern TPWRAPPER_API bool validateNetAddressRequest(const char* netAddressReg, bool ipReq, IConstTpMachine& tpMachine)
```
afishbeck, line 101: Minor, but I would change the name of this function, you are matching or comparing the value, not validating it.  

@ghalliday Doesn't have to prevent merging though if we want this to make it into 8.2.

esp/smc/SMCLib/TpCommon.cpp
```diff
@@ -97,3 +97,12 @@ extern TPWRAPPER_API IStringIterator* getContainerTargetClusters(const char* pro
...
+{
+    if (ipReq && streq(netAddressReg, tpMachine.getNetaddress()))
```
afishbeck, line 103: very minor, but could be simplified.
```
if (ipReq)
  return streq(netAddressReg, tpMachine.getNetaddress());
return streq(netAddressReg, tpMachine.getConfigNetaddress());
```

afishbeck (commented):
@ghalliday some minor comments, but doesn't have to prevent merging if we want this to make it into 8.2.

ghalliday:
@wangkx are you planning to change as Tony suggested, or shall I merge.  (Needs squashing first)

esp/services/ws_fileio/ws_fileioservice.cpp
```diff
@@ -30,15 +30,45 @@
...
+
+            netAddr.set(tpMachine.getNetaddress());
+            absPath.set(prefix);
+            addPathSepChar(absPath);
+            absPath.append(relPath);
+            return true;
+        }
+    }
 #else
     netAddr.clear();
```
ghalliday, line 73: Should this code also check netAddrReq?

esp/services/ws_fileio/ws_fileioservice.cpp
```diff
@@ -30,15 +30,45 @@
...
+        IConstTpDropZone& dropZone = allTpDropZones.item(i);
+        if (!dropZone.getECLWatchVisible())
+            continue;
+
+        const char* name = dropZone.getName();
+        if (isEmptyString(name) || !streq(targetDZNameOrAddress, name))
+            continue;
+
+        const char* prefix = dropZone.getPath();
+        if (isEmptyString(prefix))
```
ghalliday, line 55: This can never occur

esp/services/ws_fileio/ws_fileioservice.cpp
```diff
@@ -30,15 +30,45 @@
...
-    UNIMPLEMENTED_X("CONTAINERIZED(CWsFileIOEx::CheckServerAccess)");
+    bool isIp4Req = isIPAddress(netAddrReq);
+    ForEachItemIn(i, allTpDropZones)
+    {
+        IConstTpDropZone& dropZone = allTpDropZones.item(i);
+        if (!dropZone.getECLWatchVisible())
+            continue;
+
+        const char* name = dropZone.getName();
+        if (isEmptyString(name) || !streq(targetDZNameOrAddress, name))
```
ghalliday, line 51: name can never be blank

esp/services/ws_fileio/CMakeLists.txt
```diff
@@ -58,6 +58,7 @@ include_directories (
          ./../../smc/SMCLib 
          ./../../bindings/SOAP/xpp 
          ${HPCC_SOURCE_DIR}/common/thorhelper
+         ${HPCC_SOURCE_DIR}/dali/sasha
```
ghalliday, line 61: This suggests the header files are not very clean.  Better would be for ISashaCommand to be an opaque type.

esp/services/ws_fileio/ws_fileioservice.cpp
ghalliday: Could the container code also be used for bare-metal?

ghalliday (commented):
A few minor comments.  I don't think any are significant enough to prevent it being merged.
Also, what documentation impact is there?

wangkx:
Those are used by APIs (not used by ECLWatch UI).  So far, there is no documentation impact.

esp/services/ws_fileio/ws_fileioservice.cpp
wangkx: > Should this code also check netAddrReq?
A user may use the targetDZNameOrAddress to pass in a network address request.  

> Could the container code also be used for bare-metal?
The container code does not handle the legacy structure for bare-metal.

ghalliday: But the same request should work, whether or not it is targeting a bare metal or k8s system.  I suspect in most cases the name will be passed an no ip, so not a big issue.

Why doesn't the container code handle the legacy structure for baremetal?  What would go wrong if the container code was always used.

wangkx: The legacy structure is referred to the line 100. In many settings for bare metal, a DropZone setting does not have the ServerList section. But, the DropZone setting has an '@computer' attribute. Based on that attribute, the network address of the DropZone can be found from the Hardware section in the environment.xml. If the container code is used for those bare metal settings, the network address of the DropZone cannot be found.

richardkchapman (approved):

PR #15025, opened 2021-06-08
HPCC-25722 Allow annotations on Roxie services

richardkchapman:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [x] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/helm/hpcc/templates/_helpers.tpl
+++ b/helm/hpcc/templates/_helpers.tpl
@@ -840,13 +840,13 @@ Generate list of available services
 {{- range $roxie := $.Values.roxie -}}
  {{- if not $roxie.disabled -}}
   {{- range $service := $roxie.services -}}
-   {{- if ne (int $service.port) 0 -}}
+   {{- if ne (int $service.servicePort) 0 -}}
 - name: {{ $service.name }}
   class: roxie
   type: roxie
-  port: {{ $service.port }}
+  port: {{ $service.servicePort }}
   target: {{ $roxie.name }}
-  public: {{ $service.external }}
+  public: {{ (ne ( include "hpcc.isVisibilityPublic" (dict "root" $ "visibility" $service.visibility)) "") | ternary "true" "false" }}
    {{- end -}}
   {{- end }}
 {{ end -}}
@@ -855,24 +855,26 @@ Generate list of available services
 - name: {{ $esp.name }}
   class: esp
   type: {{ $esp.application }}
-  port: {{ $esp.servicePort }}
+  port: {{ $esp.service.servicePort }}
   {{- if hasKey $esp "tls" }}
   tls: {{ $esp.tls }}
   {{- else }}
   tls: {{ ($.Values.certificates | default dict).enabled }}
   {{- end }}
-  public: {{ $esp.public }}
+  public: {{ (ne ( include "hpcc.isVisibilityPublic" (dict "root" $ "visibility" $esp.service.visibility))  "") | ternary "true" "false" }}
 {{ end -}}
 {{- range $dali := $.Values.dali -}}
 {{- $sashaServices := $dali.services | default dict -}}
 {{- if not $sashaServices.disabled -}}
 {{- range $sashaName, $_sasha := $sashaServices -}}
 {{- $sasha := ($_sasha | default dict) -}}
-{{- if and (not $sasha.disabled) ($sasha.servicePort) -}}
+{{- if hasKey $sasha "service" -}}
+{{- if and (not $sasha.disabled) ($sasha.service.servicePort) -}}
 - name: {{ printf "sasha-%s" $sashaName }}
   class: sasha
   type: {{ $sashaName }}
-  port: {{ $sasha.servicePort }}
+  port: {{ $sasha.service.servicePort }}
+{{ end -}}
 {{ end -}}
 {{ end -}}
 {{ end -}}
@@ -881,11 +883,13 @@ Generate list of available services
 {{- if not $sashaServices.disabled -}}
 {{- range $sashaName, $_sasha := $sashaServices -}}
 {{- $sasha := ($_sasha | default dict) -}}
-{{- if and (not $sasha.disabled) ($sasha.servicePort) -}}
+{{- if and (not $sasha.disabled) (hasKey $sasha "service") -}}
+{{- if $sasha.service.servicePort -}}
 - name: {{ printf "sasha-%s" $sashaName }}
   class: sasha
   type: {{ $sashaName }}
-  port: {{ $sasha.servicePort }}
+  port: {{ $sasha.service.servicePort }}
+{{ end -}}
 {{ end -}}
 {{ end -}}
 {{- end -}}
@@ -1023,23 +1027,71 @@ Pass in dict with root and me
 {{- end -}}
 
 {{/*
-A template to generate Sasha service
-Pass in dict me
+A template to generate the type of a service based on the visibility setting
+Pass in dict with .root, .visibility defined
 */}}
-{{- define "hpcc.addSashaService" }}
-{{- $serviceName := printf "sasha-%s" .me.name }}
+{{- define "hpcc.isVisibilityPublic" }}
+{{- if and (hasKey . "visibility") .visibility -}}
+ {{- if hasKey .root.Values.global "visibilities" -}}
+  {{- if hasKey .root.Values.global.visibilities .visibility -}}
+   {{- $globalServiceInfo := get .root.Values.global.visibilities .visibility -}}
+   {{- if (not (eq $globalServiceInfo.type "ClusterIP")) -}}
+   true
+   {{- end -}}
+  {{- else -}}
+   {{- required (printf "Specified service visibility %s not found in global visibilities section" .visibility) nil -}}
+  {{- end -}}
+ {{- else -}}
+  {{- required "global visibilities section not found" nil -}}
+ {{- end -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+A template to generate a service
+Pass in dict with .root, .name, .service, .defaultPort, .selector defined
+*/}}
+{{- define "hpcc.addService" }}
+{{- $lvars := dict "type" "ClusterIP" "labels" dict "annotations" dict -}}
+{{- if hasKey . "service" -}}
+ {{- if hasKey .service "labels" -}}{{- $_ := set $lvars "labels" (merge $lvars.labels .service.labels) -}}{{- end -}}
+ {{- if hasKey .service "annotations" -}}{{- $_ := set $lvars "annotations" (merge $lvars.annotations .service.annotations) -}}{{- end -}}
+ {{- if hasKey .service "visibility" -}}
+  {{- if hasKey .root.Values.global "visibilities" -}}
+   {{- if hasKey .root.Values.global.visibilities .service.visibility -}}
+    {{- $globalServiceInfo := get .root.Values.global.visibilities .service.visibility -}}
+    {{- if hasKey $globalServiceInfo "labels" -}}{{- $_ := set $lvars "labels" (merge $lvars.labels $globalServiceInfo.labels) -}}{{- end -}}
+    {{- if hasKey $globalServiceInfo "annotations" -}}{{- $_ := set $lvars "annotations" (merge $lvars.annotations $globalServiceInfo.annotations) -}}{{- end -}}
+    {{- $_ := set $lvars "type" $globalServiceInfo.type -}}
+   {{- else -}}
+    {{- required (printf "Specified service visibility %s not found in global visibilities section" .service.visibility) nil -}}
+   {{- end -}}
+  {{- else -}}
+   {{- required "global visibilities section not found" nil -}}
+  {{- end -}}
+ {{- end -}}
+{{- end -}}
 apiVersion: v1
 kind: Service
 metadata:
-  name: {{ $serviceName | quote }}
+  name: {{ .name | quote }}
+  labels:
+    helmVersion: 8.2.0-closedown
+{{- if $lvars.labels }}
+{{ toYaml $lvars.labels | indent 4 }}
+{{- end }}
+{{- if $lvars.annotations }}
+  annotations:
+{{ toYaml $lvars.annotations | indent 4 }}
+{{- end }}
 spec:
   ports:
-  - port: {{ .me.servicePort }}
+  - port: {{ required "servicePort must be specified" .service.servicePort }}
     protocol: TCP
-    targetPort: {{ .port | default 8877 }}
+    targetPort: {{ .service.port | default .defaultPort }}
   selector:
-    run: {{ .selectorName | default $serviceName | quote }}
-  type: ClusterIP
+    server: {{ .selector | quote }}
+  type: {{ $lvars.type }}
 {{- end -}}
 
 
@@ -1193,7 +1245,7 @@ so that we can recognize the signature of our own CA.
 */}}
 {{- define "hpcc.addCertificate" }}
 {{- if (.root.Values.certificates | default dict).enabled -}}
-{{- $externalCert := and (hasKey . "external") .external -}}
+{{- $externalCert := or (and (hasKey . "external") .external) (ne (include "hpcc.isVisibilityPublic" .) "") -}}
 {{- $issuer := ternary .root.Values.certificates.issuers.public .root.Values.certificates.issuers.local $externalCert -}}
 {{- if $issuer -}}
 {{- $namespace := .root.Release.Namespace -}}
@@ -1230,7 +1282,7 @@ spec:
   - {{ .servicename }}.{{ $domain }}
  {{- /* if service parameter is passed in we are using the component config as a service config entry */ -}}
  {{- else if .service -}}
-   {{- $public := and (hasKey .service "public") .service.public -}}
+   {{- $public := and (hasKey .service "visibility") (not (eq .service.visibility "cluster")) -}}
    {{- if eq $public $externalCert }}
   - {{ .service.name }}.{{ $domain }}
    {{- end }}
@@ -1311,7 +1363,7 @@ spec:
 Add a certficate volume mount for a component
 */}}
 {{- define "hpcc.addCertificateVolumeMount" -}}
-{{- $externalCert := and (hasKey . "external") .external -}}
+{{- $externalCert := or (and (hasKey . "external") .external) (ne (include "hpcc.isVisibilityPublic" .) "") -}}
 {{- $exposure := ternary "public" "local" $externalCert }}
 {{- /*
     A .certificate parameter means the user explictly configured a certificate to use
@@ -1333,7 +1385,7 @@ Add a certficate volume mount for a component
 Add a secret volume for a certificate
 */}}
 {{- define "hpcc.addCertificateVolume" -}}
-{{- $externalCert := and (hasKey . "external") .external -}}
+{{- $externalCert := or (and (hasKey . "external") .external) (ne (include "hpcc.isVisibilityPublic" .) "") -}}
 {{- $exposure := ternary "public" "local" $externalCert -}}
 {{- /*
     A .certificate parameter means the user explictly configured a certificate to use
--- a/helm/hpcc/templates/dali.yaml
+++ b/helm/hpcc/templates/dali.yaml
@@ -37,13 +37,22 @@ spec:
   selector:
     matchLabels:
       run: {{ $dali.name | quote }}
+      server: {{ $dali.name | quote }}
       app: dali
   template:
     metadata:
       labels:
         run: {{ $dali.name | quote }}
+        server: {{ $dali.name | quote }}
         app: dali
         helmVersion: 8.2.0-closedown
+{{- if hasKey $dali "labels" }}
+{{ toYaml $dali.labels | indent 8 }}
+{{- end }}
+{{- if hasKey $dali "annotations" }}
+      annotations:
+{{ toYaml $dali.annotations | indent 8 }}
+{{- end }}
     spec:
       {{- include "hpcc.placementsByPodTargetType" (dict "root" $ "pod" $dali.name "type" "dali") | indent 6 }}
       serviceAccountName: "hpcc-default"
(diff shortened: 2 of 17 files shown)


helm/hpcc/templates/_helpers.tpl
```diff
@@ -763,39 +763,44 @@ Generate list of available services
...
+  port: {{ $service.servicePort }}
   target: {{ $roxie.name }}
-  public: {{ $service.external }}
+  public: {{ $service.public }}
    {{- end -}}
   {{- end }}
 {{ end -}}
 {{- end -}}
 {{- range $esp := $.Values.esp -}}
+{{- if not (hasKey $esp "service") -}}
```
jakesmith, line 778: out of curiosity, are there esp services without a public/internal service?

helm/hpcc/templates/esp.yaml
```diff
@@ -73,6 +73,12 @@ data:
 {{- if not .disabled -}}
 {{- $env := concat ($.Values.global.env | default list) (.env | default list) -}}
 {{- $application := .application | default "eclwatch" -}}
+{{- if not (hasKey . "service") -}}
+{{- $_ := set . "service" (dict "port" (.port | default 0) "servicePort" (.servicePort | default 0) "public" (.public | default false) ) -}}
+{{- $_ := unset . "port" -}}
+{{- $_ := unset . "servicePort" -}}
+{{- $_ := unset . "public" -}}
```
jakesmith, line 80: is above for backward compatibility, for old value yamls that don't define the new .service.servicePort etc. spec?
If so, need similar for Sasha and Roxie to support their old format?

helm/hpcc/templates/esp.yaml
```diff
@@ -135,15 +141,23 @@ apiVersion: v1
...
   ports:
-  - port: {{ required "servicePort must be specified" .servicePort }}
+  - port: {{ required "servicePort must be specified" .service.servicePort }}
     protocol: TCP
-    targetPort: {{ .port | default 8880 }}
+    targetPort: {{ .service.port | default 8880 }}
   selector:
     run: {{ .name | quote }}
-  type: {{ .public | ternary "LoadBalancer" "ClusterIP" }}
+  type: {{ .service.public | ternary "LoadBalancer" "ClusterIP" }}
```
jakesmith, line 159: The whole of this Service generation before/now looks very generic, only thing that's specific to esp perhaps is the default ports.
It would be good to move it into _helpers.tpl, as a generateService call, and use in other places too.
(think it's same for example as on in Roxie apart from the default port aspect)

helm/hpcc/templates/localroxie.yaml
```diff
@@ -121,21 +121,28 @@ spec:
...
   ports:
-  - port: {{ $service.port }}
+  - port: {{ $service.servicePort }}
     protocol: TCP
-    targetPort: {{ $service.port }}
+    targetPort: {{ $service.servicePort }}
   selector:
     roxie-server: {{ $servername | quote }}
-  type: {{ if $service.external -}} LoadBalancer {{- else -}} ClusterIP {{- end }}
+  type: {{ if $service.public -}} LoadBalancer {{- else -}} ClusterIP {{- end }}
```
jakesmith, line 145: see comment in esp.yaml - looks like generate of a Service is very similar, could be commoned up.

helm/hpcc/values.yaml
```diff
@@ -250,8 +250,7 @@ dali:
 - name: mydali
   services: # internal house keeping services
     coalescer:
-      #servicePort: 8877
-      #interval: 2 # (hours)
+      interval: 2 # (hours)
```
jakesmith, line 253: intentionally commented in?

helm/hpcc/values.yaml
```diff
@@ -300,14 +301,18 @@ sasha:
     #throttle: 0 # throttle ratio (0-99, 0 no throttling, 50 is half speed)
 
   dfurecovery-archiver:
-    #disabled: true
+    disabled: true
+    service:
```
jakesmith, line 305: I assume not intended, just disabled while testing?

helm/hpcc/values.yaml
```diff
@@ -300,14 +301,18 @@ sasha:
...
+      servicePort: 8877
     #limit: 20 # threshold number of DFU recovery items before archiving starts (0 disables)
     #cutoff: 4 # minimum DFU recovery item age to archive (days)
     #interval: 12 # minimum interval between running DFU recovery archiver(in hours, 0 disables)
     #at: "* * * * *" # schedule to run DFU recovery archiver (cron format)
 
   file-expiry:
-    #disabled: true
+    disabled: true
```
jakesmith, line 313: I assume not intended, just disabled while testing?

helm/hpcc/templates/esp.yaml
richardkchapman: Yes. Not sure whether it's a good idea or not, or whether should just require people to change. Given the schema changes I suspect this code may be moot anyway.

roxie/ccd/ccdmain.cpp
```diff
@@ -739,7 +739,7 @@ int CCD_API roxie_main(int argc, const char *argv[], const char * defaultYaml)
                     // Makes debugging easier...
                     IPropertyTree *service = topology->addPropTree("services");
                     service->setProp("@name", "query");
-                    service->setPropInt("@port", 9876);
+                    service->setPropInt("@servicePort", 9876);
```
jakesmith, line 742: is this correct? Shouldn't it be listening to @port (the internally defined port to use), which should default to @servicePort, if @port was not defined in values yaml

jakesmith: or maybe this isn't being used to listen to here,
but is being used contact the exposed service ?

helm/hpcc/templates/esp.yaml
jakesmith: yeah, given the schema, I think no point, though disruptive to users potentially.

If this code is removed though, I think there is other code that node checks .service.X - that will need to be careful to check if key "service" exists first.

helm/hpcc/templates/localroxie.yaml
```diff
@@ -121,21 +121,28 @@ spec:
...
+  annotations:
+{{ toYaml $service.annotations | indent 4 }}
+{{ end -}}
 spec:
   ports:
-  - port: {{ $service.port }}
+  - port: {{ $service.servicePort }}
     protocol: TCP
-    targetPort: {{ $service.port }}
+    targetPort: {{ $service.servicePort }}
```
jakesmith, line 142: shouldn't this remain as port?

richardkchapman: Roxie does not allow them to be different. I'm not sure why other components do, TBH.

roxie/ccd/ccdmain.cpp
richardkchapman: Roxie uses a single port to specify both internal and external - they are not allowed to differ.

helm/hpcc/values.yaml
richardkchapman: reverted

helm/hpcc/values.yaml
richardkchapman: reverted

helm/hpcc/values.yaml
richardkchapman: reverted

helm/hpcc/templates/localroxie.yaml
richardkchapman: Done

helm/hpcc/templates/esp.yaml
richardkchapman: Done

roxie/ccd/ccdmain.cpp
```diff
@@ -1326,7 +1326,7 @@ int CCD_API roxie_main(int argc, const char *argv[], const char * defaultYaml)
                     unsigned numThreads = roxieFarm.getPropInt("@numThreads", 0);
                     if (!numThreads)
                         numThreads = numServerThreads;
-                    unsigned port = roxieFarm.getPropInt("@port", ROXIE_SERVER_PORT);
```
richardkchapman, line 1329: This breaks bare-metal...

jakesmith: are other places ok to default in bare metal (e.g. line 1205) ?

Perhaps instead of conditional code here, check @port in bare-metal of farmers upfront, e.g. in block around line 656

helm/hpcc/values.schema.json
```diff
@@ -953,11 +1005,29 @@
           "type": "integer",
           "default": 1
         },
-        "servicePort": {
-          "type": "integer"
-        },
-        "port": {
-          "type": "integer"
+        "service": {
```
jakesmith, line 1008: I think this service section is identical to esp's apart from no "public" field - can it be commoned up?

roxie/ccd/ccdmain.cpp
richardkchapman: Others are inside #ifdef's that make them containerized only.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -945,23 +952,31 @@ Pass in dict with root and me
...
-  - port: {{ .me.servicePort }}
+  - port: {{ required "servicePort must be specified" .service.servicePort }}
     protocol: TCP
-    targetPort: {{ .port | default 8877 }}
+    targetPort: {{ .service.port | default .defaultPort }}
   selector:
-    run: {{ .selectorName | default $serviceName | quote }}
-  type: ClusterIP
+    server: {{ .selector | quote }}
+  type: {{ and ( hasKey .service "public") .service.public | ternary "LoadBalancer" "ClusterIP" }}
```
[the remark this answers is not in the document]
richardkchapman: Not for sasha...

helm/hpcc/values.schema.json
richardkchapman: The lack of public is enough reason to not want to

ghalliday (commented):
@richardkchapman looks good.  a few minor comments.

richardkchapman:
"even if public was missing the hasKey test would not be needed since it would evaluate to nil/false."

You would think that, but ternary seems to object to the first arg not being boolean, rather than treating nil as false.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -763,39 +763,44 @@ Generate list of available services
...
   target: {{ $roxie.name }}
-  public: {{ $service.external }}
+  public: {{ $service.public }}
    {{- end -}}
   {{- end }}
 {{ end -}}
 {{- end -}}
 {{- range $esp := $.Values.esp -}}
+{{- if not (hasKey $esp "service") -}}
+{{- $_ := set $esp "service" (dict "port" (.port | default 0) "servicePort" (.servicePort | default 0) "public" (.public | default false) ) -}}
```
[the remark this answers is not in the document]
richardkchapman: Yes, though I think this change is now redundant since service is compulsory in the schema. I will remove it.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -804,11 +809,13 @@ Generate list of available services
 {{- if not $sashaServices.disabled -}}
 {{- range $sashaName, $_sasha := $sashaServices -}}
 {{- $sasha := ($_sasha | default dict) -}}
-{{- if and (not $sasha.disabled) ($sasha.servicePort) -}}
+{{- if and (not $sasha.disabled) (hasKey $sasha "service") -}}
+{{- if $sasha.service.servicePort -}}
```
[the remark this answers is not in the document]
richardkchapman: Sasha is a bit different as service is optional

helm/hpcc/templates/dali.yaml
```diff
@@ -37,11 +37,13 @@ spec:
   selector:
     matchLabels:
       run: {{ $dali.name | quote }}
```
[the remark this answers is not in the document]
richardkchapman: I used a server: label for associating services with pods, rather than a run: label - they don't always match (e.g. sasha services that are done inside the dali, and roxie clusters where not every node is a server).

I left the run: label in place as that seems to be something of a helm convention, and has its uses for interrogating the state of the system via kubectl etc. I don't think we actually use it ourselves now though.

helm/hpcc/values.schema.json
```diff
@@ -709,6 +734,42 @@
...
+              "description": "The local port used by the pod",
+              "default": 8880
+            },
+            "servicePort": {
+              "type": "integer",
+              "description": "The port that this service will be exposed on"
+            },
+            "visibility": {
+              "type": "string",
+              "description": "Should this service be exposed outside the cluster, locally or to the internet"
```
ghalliday, line 762: "enum": [ "cluster", "local", "global" ],?

helm/hpcc/values.schema.json
ghalliday: Does this change now you have local/global?

helm/hpcc/values.yaml
```diff
@@ -63,6 +63,21 @@ global:
   #   postJobCommand: "kill $(pgrep linkerd2-proxy)"
   #   postJobCommandViaSidecar: true
 
+  ## visibilities section can be used to set labels, annotations and service type for any service with the specified visibility
+  visibilities:
+    cluster:
+      type: ClusterIP
+    local:
+      annotations:
+        service.beta.kubernetes.io/azure-load-balancer-internal: "true"
```
ghalliday, line 72: Should this be commented out by default?

helm/hpcc/values.yaml
```diff
@@ -365,48 +383,59 @@ esp:
...
+  service:
+    ## port can be used to change the local port used by the pod. If omitted, the default port (8880) is used
+    port: 8888
+    ## servicePort controls the port that this service will be exposed on, either internally to the cluster, or externally
+    servicePort: 8010
+    ## Specify visibility: local (or global) if you want the service available from outside the cluster. Typically, eclwatch and wsecl are published externally, while eclservices is designed for internal use.
+    visibility: local
+    ## Annotations can be specified on a service - for example to specify provider-specific information such as service.beta.kubernetes.io/azure-load-balancer-internal-subnet
+    annotations:
+      service.beta.kubernetes.io/azure-load-balancer-internal-subnet: "mysubnet"
```
ghalliday, line 395: comment out? or remove entirely now you have your visibility section?  Same for the label below.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -945,23 +949,71 @@ Pass in dict with root and me
...
 
 {{/*
-A template to generate Sasha service
-Pass in dict me
+A template to generate the type of a service based on the visibility setting
+Pass in dict with .root, .visibility defined
 */}}
-{{- define "hpcc.addSashaService" }}
-{{- $serviceName := printf "sasha-%s" .me.name }}
+{{- define "hpcc.getVisibilityType" }}
```
ghalliday, line 955: Is this the correct name for the function?  getServiceType()?

ghalliday (commented):
@richardkchapman I like the visibility section....  A few comments

helm/hpcc/values.schema.json
richardkchapman: It was originally, but it's now possible for people to add their own visibility classes if they want

helm/hpcc/templates/_helpers.tpl
richardkchapman: It is mapping a visibility value to a service type value. I could rename it if you think a different name is less confusing. Though I suspect if I can get boolean returns working I will probably replace it with something like "isVisibilityExternal"

helm/hpcc/values.schema.json
richardkchapman: I don't think so - we don't want to allow sasha services to be made visible.

helm/hpcc/values.yaml
richardkchapman: No - because we want by default the system to work on azure without exposing things to the internet (thus keeping Flavio happy). It will be harmless on non-azure systems.

helm/hpcc/values.yaml
richardkchapman: I should comment it out. The idea is that you can override on individual services as well as setting up visibility "classes" of service.

ghalliday (approved):
Looks ok.  Jake should also check.

helm/hpcc/values.schema.json
ghalliday: Should the restriction be removed for the other place where visibility is defined?

richardkchapman: Yes - well spotted. I will push another fix

ghalliday (approved):
Looks good to me.

esp/smc/SMCLib/TpContainer.cpp
```diff
@@ -500,7 +500,7 @@ void CTpWrapper::getServices(double version, const char* serviceType, const char
     {
         IPropertyTree& service = itr->query();
         //Only show the public services for now
-        if (!service.getPropBool("@public"))
+        if (!service.getPropBool("@public", false))
```
jakesmith, line 503: super trivial: we don't normally add the default param in cases like this.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -840,13 +840,13 @@ Generate list of available services
...
-   {{- if ne (int $service.port) 0 -}}
+   {{- if ne (int $service.servicePort) 0 -}}
 - name: {{ $service.name }}
   class: roxie
   type: roxie
-  port: {{ $service.port }}
+  port: {{ $service.servicePort }}
   target: {{ $roxie.name }}
-  public: {{ $service.external }}
+  public: {{ (ne ( include "hpcc.isVisibilityPublic" (dict "root" $ "visibility" $service.visibility))  "") | ternary "true" "false" }}
```
jakesmith, line 849: super trivial: extra space before ""

helm/hpcc/templates/_helpers.tpl
```diff
@@ -855,24 +855,26 @@ Generate list of available services
...
   type: {{ $esp.application }}
-  port: {{ $esp.servicePort }}
+  port: {{ $esp.service.servicePort }}
   {{- if hasKey $esp "tls" }}
   tls: {{ $esp.tls }}
   {{- else }}
   tls: {{ ($.Values.certificates | default dict).enabled }}
   {{- end }}
-  public: {{ $esp.public }}
+  public: {{ (ne ( include "hpcc.isVisibilityPublic" (dict "root" $ "visibility" $esp.service.visibility))  "") | ternary "true" "false" }}
```
jakesmith, line 864: trivial: if isVisibilityPublic returned only "true" or "false" (never ""), could you simplify?

helm/hpcc/templates/_helpers.tpl
```diff
@@ -1023,23 +1027,71 @@ Pass in dict with root and me
...
 {{/*
-A template to generate Sasha service
-Pass in dict me
+A template to generate the type of a service based on the visibility setting
+Pass in dict with .root, .visibility defined
 */}}
-{{- define "hpcc.addSashaService" }}
-{{- $serviceName := printf "sasha-%s" .me.name }}
+{{- define "hpcc.isVisibilityPublic" }}
+{{- if and (hasKey . "visibility") .visibility -}}
```
jakesmith, line 1034: isn't :
```
{{- if .visibility -}}
```

enough here?

helm/hpcc/templates/_helpers.tpl
```diff
@@ -1023,23 +1027,71 @@ Pass in dict with root and me
...
+ {{- if hasKey .root.Values.global "visibilities" -}}
+  {{- if hasKey .root.Values.global.visibilities .visibility -}}
+   {{- $globalServiceInfo := get .root.Values.global.visibilities .visibility -}}
+   {{- if (not (eq $globalServiceInfo.type "ClusterIP")) -}}
+   true
```
jakesmith, line 1039: I think really, should not indent here, i.e. not return "    true"

helm/hpcc/templates/_helpers.tpl
```diff
@@ -1023,23 +1027,71 @@ Pass in dict with root and me
...
+   {{- end -}}
+  {{- else -}}
+   {{- required (printf "Specified service visibility %s not found in global visibilities section" .service.visibility) nil -}}
```
jakesmith, line 1042: trivia: is required ... nil , preferable to fatal .. ?

jakesmith (approved):
@richardkchapman - trivial comments only

helm/hpcc/templates/_helpers.tpl
richardkchapman: No - because it would be returning them as strings, not as a boolean

helm/hpcc/templates/_helpers.tpl
richardkchapman: Doesn't matter given the -}} above. I decided it was more legible this way

helm/hpcc/templates/_helpers.tpl
richardkchapman: It seems to give a fairly nice error message. I can try fatal and see if it seems better

esp/smc/SMCLib/TpContainer.cpp
richardkchapman: I was trying to revert the code to the way it was (and failed). I'll take it out

helm/hpcc/templates/_helpers.tpl
richardkchapman: Will remove

helm/hpcc/templates/_helpers.tpl
richardkchapman: Maybe. But I tested it like this...

helm/hpcc/templates/_helpers.tpl
richardkchapman: I can't find any documentation on fatal

richardkchapman: (though I did find a typo in the error reporting line when I tried to test it so that was handy!)

PR #15046, opened 2021-06-14
HPCC-25722 Allow annotations on services and pods

richardkchapman:
Fix issue with Sasha and ESP ignoring local port overrides (and thus failing
to connect to external LB)

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [x] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

## Testing:
Tested overridden local ports for a couple of services.

### DIFF
--- a/dali/sasha/saserver.cpp
+++ b/dali/sasha/saserver.cpp
@@ -360,7 +360,7 @@ int main(int argc, const char* argv[])
     #endif
         DBGLOG("Build %s", hpccBuildInfo.buildTag);
 
-        unsigned short port = serverConfig->getPropInt("@port");
+        unsigned short port = serverConfig->getPropInt("service/@port", serverConfig->getPropInt("@port"));
         if (!port)
         {
             if (!stop && !coalescer)
--- a/esp/platform/application_config.cpp
+++ b/esp/platform/application_config.cpp
@@ -295,7 +295,7 @@ bool addProtocol(IPropertyTree *legacyEsp, IPropertyTree *app)
 void addServices(IPropertyTree *legacyEsp, IPropertyTree *appEsp, const char *application, const char *auth, bool tls)
 {
     Owned<IPropertyTreeIterator> services = appEsp->getElements("application/services");
-    int port = appEsp->getPropInt("@port", 8880);
+    int port = appEsp->getPropInt("service/@port", appEsp->getPropInt("@port", 8880));
     const char *netAddress = appEsp->queryProp("@netAddress");
     if (!netAddress)
         netAddress = ".";
--- a/esp/applications/eclservices/esp.yaml
+++ b/esp/applications/eclservices/esp.yaml
@@ -6,7 +6,8 @@ esp:
    auth: ldap
    tls: true
    wuTokens: true
-   port: 8880
+   service:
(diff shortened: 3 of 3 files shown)


dali/sasha/saserver.cpp
```diff
@@ -360,7 +360,7 @@ int main(int argc, const char* argv[])
     #endif
         DBGLOG("Build %s", hpccBuildInfo.buildTag);
 
-        unsigned short port = serverConfig->getPropInt("@port");
+        unsigned short port = serverConfig->getPropInt("service/@port", serverConfig->getPropInt("@port"));
```
jakesmith, line 363: doesn't the schema prevent "@port" being valid, and so is there any point in defaulting up legacy spec. ?

esp/platform/application_config.cpp
```diff
@@ -295,7 +295,7 @@ bool addProtocol(IPropertyTree *legacyEsp, IPropertyTree *app)
 void addServices(IPropertyTree *legacyEsp, IPropertyTree *appEsp, const char *application, const char *auth, bool tls)
 {
     Owned<IPropertyTreeIterator> services = appEsp->getElements("application/services");
-    int port = appEsp->getPropInt("@port", 8880);
+    int port = appEsp->getPropInt("service/@port", appEsp->getPropInt("@port", 8880));
```
jakesmith, line 298: I think these application yaml's are part of the image e.g. /opt/HPCCSystems/componentfiles/applications/eclwatch/

I guess picking up the legacy "@port" as a default here, ensures still works if the used with an older image, that also had @port set differently in it's image copy of esp.yaml.
But I don't think that can/has happened in reality.

Perhaps would be clearer not to have code pick up legacy "@port" here.

jakesmith (commented):
a couple of minor questions

esp/platform/application_config.cpp
afishbeck: Those application yamls can be thought of as ESPs "default.yaml".  The values can still be provided in values.yaml which needs this backward compatibility to work with any existing ones.

dali/sasha/saserver.cpp
richardkchapman: Depends on what bare-metal looks like...

richardkchapman:
I codeed it as I did to avoid breaking bare metal cases.

afishbeck (commented):
espservices is just one of several application flavors that could use the yaml change.  
esp/applications/*/esp.yaml
The backward compatability check should mean they would continue to work, but they also serve as a form of documentation about how users should set up their values.yaml for ESP so should be updated.

esp/platform/application_config.cpp
jakesmith: if this code is used in bare-metal and it's possible it could pick up a old legacy format.
In cloud the schema will not allow the legacy spec.

jakesmith (approved):
The defaults are harmless anyway.

ghalliday (approved):
test failures are a bit concerning, but unrelated to this change.

PR #15051, opened 2021-06-15
HPCC-25910 Improve the syntax for spraying from a k8s landingzone

shamser:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dali/dfu/dfurun.cpp
+++ b/dali/dfu/dfurun.cpp
@@ -468,7 +468,6 @@ class CDFUengine: public CInterface, implements IDFUengine
             IPropertyTree & plane = planes->query();
             const char * fullDropZoneDir = plane.queryProp("@prefix");
             assertex(fullDropZoneDir);
-            // note: for bare-metal drop-zones, will need to compare ip address
             if (startsWith(pfilePath, fullDropZoneDir))
                 return;
         }
@@ -1317,7 +1316,7 @@ public:
 #ifdef _CONTAINERIZED
                         StringBuffer clusterName;
                         destination->getGroupName(0, clusterName);
-                        Owned<IPropertyTree> plane = getDropZonePlane(clusterName);
+                        Owned<IPropertyTree> plane = getStoragePlane(clusterName);
                         if (plane)
                         {
                             if (plane->hasProp("@defaultSprayParts"))
--- a/dali/dfuplus/dfuplus.cpp
+++ b/dali/dfuplus/dfuplus.cpp
@@ -351,7 +351,9 @@ int CDfuPlusHelper::doit()
     return 0;
 }
 
-bool CDfuPlusHelper::fixedSpray(const char* srcxml,const char* srcip,const char* srcfile,const MemoryBuffer &xmlbuf,const char* dstcluster,const char* dstname,const char *format, StringBuffer &retwuid, StringBuffer &except)
+bool CDfuPlusHelper::fixedSpray(const char* srcxml,const char* srcip,const char* srcfile,const char* srcplane,
+                                const MemoryBuffer &xmlbuf,const char* dstcluster,const char* dstname,
+                                const char *format, StringBuffer &retwuid, StringBuffer &except)
 {
     int recordsize;
     if(stricmp(format, "recfmvb") == 0) {
@@ -374,18 +376,23 @@ bool CDfuPlusHelper::fixedSpray(const char* srcxml,const char* srcip,const char*
     }
 
     Owned<IClientSprayFixed> req = sprayclient->createSprayFixedRequest();
-    if(srcxml == nullptr)
+    if(isEmptyString(srcxml))
     {
-        req->setSourceIP(srcip);
+        info("\nFixed spraying from %s on %s to %s\n", srcfile, srcplane?srcplane:srcip, dstname);
+        if (!isEmptyString(srcip)) req->setSourceIP(srcip);
+        if (!isEmptyString(srcplane)) req->setSourcePlane(srcplane);
         req->setSourcePath(srcfile);
     }
     else
+    {
+        info("\nSpraying to %s\n", dstname);
         req->setSrcxml(xmlbuf);
+    }
 
     if(recordsize != 0)
         req->setSourceRecordSize(recordsize);
 
-    if(dstcluster != nullptr)
+    if(!isEmptyString(dstcluster))
         req->setDestGroup(dstcluster);
     req->setDestLogicalName(dstname);
     req->setOverwrite(globals->getPropBool("overwrite", false));
@@ -429,11 +436,6 @@ bool CDfuPlusHelper::fixedSpray(const char* srcxml,const char* srcip,const char*
     if(globals->hasProp("expireDays"))
         req->setExpireDays(globals->getPropInt("expireDays"));
 
-    if(srcxml == nullptr)
-        info("\nFixed spraying from %s on %s to %s\n", srcfile, srcip, dstname);
-    else
-        info("\nSpraying to %s\n", dstname);
-
     Owned<IClientSprayFixedResponse> result = sprayclient->SprayFixed(req);
     const char *wuid = result->getWuid();
     if (!wuid||!*wuid) {
@@ -444,21 +446,26 @@ bool CDfuPlusHelper::fixedSpray(const char* srcxml,const char* srcip,const char*
     return true;
 }
 
-bool CDfuPlusHelper::variableSpray(const char* srcxml,const char* srcip,const char* srcfile,const MemoryBuffer &xmlbuf,const char* dstcluster,const char* dstname,const char *format,StringBuffer &retwuid, StringBuffer &except)
+bool CDfuPlusHelper::variableSpray(const char* srcxml,const char* srcip,const char* srcfile,const char* srcplane,
+                                   const MemoryBuffer &xmlbuf,const char* dstcluster,const char* dstname,
+                                   const char *format,StringBuffer &retwuid, StringBuffer &except)
 {
     Owned<IClientSprayVariable> req = sprayclient->createSprayVariableRequest();
-    if(srcxml == nullptr)
+    if(isEmptyString(srcxml))
     {
-        req->setSourceIP(srcip);
+        info("\nVariable spraying from %s on %s to %s\n", srcfile, srcplane?srcplane:srcip, dstname);
+        if (!isEmptyString(srcip)) req->setSourceIP(srcip);
+        if (!isEmptyString(srcplane)) req->setSourcePlane(srcplane);
         req->setSourcePath(srcfile);
     }
     else
     {
+        info("\nSpraying to %s\n", dstname);
         req->setSrcxml(xmlbuf);
     }
 
     const char* mrsstr = globals->queryProp("maxRecordSize");
-    if(mrsstr != nullptr)
+    if(!isEmptyString(mrsstr))
         req->setSourceMaxRecordSize(atoi(mrsstr));
     else
         req->setSourceMaxRecordSize(8192);
@@ -480,23 +487,23 @@ bool CDfuPlusHelper::variableSpray(const char* srcxml,const char* srcip,const ch
     }
     else if(stricmp(format, "xml") == 0)
     {
-        if(encoding == nullptr)
+        if(isEmptyString(encoding))
             encoding = "utf8";
         else if(stricmp(encoding, "ascii") == 0)
             throw MakeStringException(-1, "xml format only accepts utf encodings");
-        if(rowtag == nullptr || *rowtag == '\0')
+        if(isEmptyString(rowtag))
             throw MakeStringException(-1, "rowtag not specified.");
-        if(rowpath && *rowpath)
+        if(!isEmptyString(rowpath))
             throw MakeStringException(-1, "You can't use rowpath option with xml format");
     }
     else if(stricmp(format, "csv") == 0)
     {
-        if(encoding == nullptr)
+        if(isEmptyString(encoding))
             encoding = "ascii";
 
-        if(rowtag != nullptr && *rowtag != '\0')
+        if(!isEmptyString(rowtag))
             throw MakeStringException(-1, "You can't use rowtag option with csv/delimited format");
-        if(rowpath && *rowpath)
+        if(!isEmptyString(rowpath))
             throw MakeStringException(-1, "You can't use rowpath option with csv/delimited format");
 
         const char* separator = globals->queryProp("separator");
@@ -509,7 +516,7 @@ bool CDfuPlusHelper::variableSpray(const char* srcxml,const char* srcip,const ch
                 req->setNoSourceCsvSeparator(true);
         }
         const char* terminator = globals->queryProp("terminator");
-        if(terminator && *terminator)
+        if(!isEmptyString(terminator))
             req->setSourceCsvTerminate(terminator);
         const char* quote = globals->queryProp("quote");
         if(quote)
@@ -519,17 +526,17 @@ bool CDfuPlusHelper::variableSpray(const char* srcxml,const char* srcip,const ch
             req->setSourceCsvQuote(quote);
         }
         const char* escape = globals->queryProp("escape");
-        if(escape && *escape)
+        if(!isEmptyString(escape))
             req->setSourceCsvEscape(escape);
     }
     else
         encoding = format; // may need extra later
 
     req->setSourceFormat(CDFUfileformat::decode(encoding));
-    if(rowtag != nullptr)
+    if(!isEmptyString(rowtag))
         req->setSourceRowTag(rowtag);
 
-    if(dstcluster != nullptr)
+    if(!isEmptyString(dstcluster))
         req->setDestGroup(dstcluster);
     req->setDestLogicalName(dstname);
     req->setOverwrite(globals->getPropBool("overwrite", false));
@@ -578,13 +585,10 @@ bool CDfuPlusHelper::variableSpray(const char* srcxml,const char* srcip,const ch
     if(globals->hasProp("expireDays"))
         req->setExpireDays(globals->getPropInt("expireDays"));
 
-    if(srcxml == nullptr)
-        info("\nVariable spraying from %s on %s to %s\n", srcfile, srcip, dstname);
-    else
-        info("\nSpraying to %s\n", dstname);
     Owned<IClientSprayResponse> result = sprayclient->SprayVariable(req);
     const char *wuid = result->getWuid();
-    if (!wuid||!*wuid) {
+    if (isEmptyString(wuid))
+    {
         result->getExceptions().errorMessage(except);
         return false;
     }
@@ -594,31 +598,36 @@ bool CDfuPlusHelper::variableSpray(const char* srcxml,const char* srcip,const ch
 
 int CDfuPlusHelper::spray()
 {
+    bool usingSrcPlane = false;
     const char* srcxml = globals->queryProp("srcxml");
     const char* srcip = globals->queryProp("srcip");
+    const char* srcplane =  globals->queryProp("srcPlane");
     const char* srcfile = globals->queryProp("srcfile");
 
     bool nowait = globals->getPropBool("nowait", false);
 
+    if (!isEmptyString(srcplane))
+        usingSrcPlane=true;
     MemoryBuffer xmlbuf;
 
-    if(srcxml == nullptr)
+    if(isEmptyString(srcxml))
     {
-        if(srcfile == nullptr)
+        if(isEmptyString(srcfile))
             throw MakeStringException(-1, "srcfile not specified");
-        if(srcip == nullptr) {
+        if(!usingSrcPlane && isEmptyString(srcip))
+        {
 #ifdef DAFILESRV_LOCAL
             progress("srcip not specified - assuming spray from local machine\n");
             srcip = ".";
 #else
-            throw MakeStringException(-1, "srcip not specified");
+            throw MakeStringException(-1, "Neither srcip nor srcplane specified");
 #endif
         }
     }
     else
     {
-        if(srcip != nullptr || srcfile != nullptr)
-            throw MakeStringException(-1, "srcip/srcfile and srcxml can't be used at the same time");
+        if(!isEmptyString(srcip) || !isEmptyString(srcfile) || usingSrcPlane)
+            throw MakeStringException(-1, "srcip/srcfile/srcplane and srcxml can't be used at the same time");
         StringBuffer buf;
         buf.loadFile(srcxml);
         int len = buf.length();
@@ -626,10 +635,10 @@ int CDfuPlusHelper::spray()
     }
 
     const char* dstname = globals->queryProp("dstname");
-    if(dstname == nullptr)
+    if(isEmptyString(dstname))
         throw MakeStringException(-1, "dstname not specified");
     const char* dstcluster = globals->queryProp("dstcluster");
-    if(dstcluster == nullptr)
+    if(isEmptyString(dstcluster))
         throw MakeStringException(-1, "dstcluster not specified");
     const char* format = globals->queryProp("format");
     if(format == nullptr)
@@ -637,17 +646,20 @@ int CDfuPlusHelper::spray()
     else if (stricmp(format, "delimited") == 0)
         format="csv";
 
-    SocketEndpoint localep;
-    StringBuffer localeps;
-    if (checkLocalDaFileSvr(srcip,localep))
-        srcip = localep.getUrlStr(localeps).str();
+    if (!usingSrcPlane)
+    {
+        SocketEndpoint localep;
+        StringBuffer localeps;
+        if (checkLocalDaFileSvr(srcip,localep))
+            srcip = localep.getUrlStr(localeps).str();
+    }
     StringBuffer wuid;
     StringBuffer errmsg;
     bool ok;
     if ((stricmp(format, "fixed") == 0)||(stricmp(format, "recfmvb") == 0)||(stricmp(format, "recfmv") == 0)||(stricmp(format, "variablebigendian") == 0))
-        ok = fixedSpray(srcxml,srcip,srcfile,xmlbuf,dstcluster,dstname,format,wuid,errmsg);
+        ok = fixedSpray(srcxml,srcip,srcfile,srcplane,xmlbuf,dstcluster,dstname,format,wuid,errmsg);
     else if((stricmp(format, "csv") == 0)||(stricmp(format, "xml") == 0)||(stricmp(format, "json") == 0)||(stricmp(format, "variable") == 0))
-        ok = variableSpray(srcxml,srcip,srcfile,xmlbuf,dstcluster,dstname,format,wuid, errmsg);
+        ok = variableSpray(srcxml,srcip,srcfile,srcplane,xmlbuf,dstcluster,dstname,format,wuid, errmsg);
     else
         throw MakeStringException(-1, "format %s not supported", format);
     if (!ok) {
@@ -716,21 +728,22 @@ int CDfuPlusHelper::replicate()
 int CDfuPlusHelper::despray()
 {
     const char* srcname = globals->queryProp("srcname");
-    if(srcname == nullptr)
+    if(isEmptyString(srcname))
         throw MakeStringException(-1, "srcname not specified");
 
     const char* dstxml = globals->queryProp("dstxml");
     const char* dstip = globals->queryProp("dstip");
+    const char* dstplane = globals->queryProp("dstplane");
     const char* dstfile = globals->queryProp("dstfile");
 
     bool nowait = globals->getPropBool("nowait", false);
 
     MemoryBuffer xmlbuf;
-    if(dstxml == nullptr)
+    if(isEmptyString(dstxml))
     {
-        if(dstip == nullptr) {
+        if (isEmptyString(dstplane) && isEmptyString(dstip)) {
 #ifdef DAFILESRV_LOCAL
-            progress("dstip not specified - assuming spray from local machine\n");
+            progress("dstip and dstplane not specified - assuming despray to local machine\n");
             dstip = ".";
 #else
             throw MakeStringException(-1, "dstip not specified");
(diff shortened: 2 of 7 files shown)


shamser:
Documentation jira: https://track.hpccsystems.com/browse/HPCC-26044

dali/dfu/dfurun.cpp
```diff
@@ -1317,7 +1317,7 @@ class CDFUengine: public CInterface, implements IDFUengine
 #ifdef _CONTAINERIZED
                         StringBuffer clusterName;
                         destination->getGroupName(0, clusterName);
-                        Owned<IPropertyTree> plane = getDropZonePlane(clusterName);
+                        Owned<IPropertyTree> plane =  getStoragePlane(clusterName);
```
shamser, line 1320: This is a bug fix.  The original code should have used getStoragePlane() not getDropzonePlane.

dali/dfuplus/dfuplus.cpp
```diff
@@ -351,7 +351,9 @@ int CDfuPlusHelper::doit()
     return 0;
 }
 
-bool CDfuPlusHelper::fixedSpray(const char* srcxml,const char* srcip,const char* srcfile,const MemoryBuffer &xmlbuf,const char* dstcluster,const char* dstname,const char *format, StringBuffer &retwuid, StringBuffer &except)
+bool CDfuPlusHelper::fixedSpray(const char* srcxml,const char* srcip,const char* srcfile,const char* srcplane,
```
ghalliday, line 354: How does dstplane differ from dstcluster?

dali/dfuplus/dfuplus.cpp
```diff
@@ -721,14 +746,15 @@ int CDfuPlusHelper::despray()
...
+    const char* dstplane = globals->queryProp("dstplane");
     const char* dstfile = globals->queryProp("dstfile");
 
     bool nowait = globals->getPropBool("nowait", false);
 
     MemoryBuffer xmlbuf;
     if(dstxml == nullptr)
     {
-        if(dstip == nullptr) {
+        if (!isEmptyString(dstplane) && isEmptyString(dstip)) {
```
ghalliday, line 757: Shouldn't this be isEmptyString(dstplane)?

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -1814,17 +1814,28 @@ bool CFileSprayEx::onGetDFUExceptions(IEspContext &context, IEspGetDFUExceptions
...
     {
+        if (!isEmptyString(srcPlane))
+        {
+            Owned<IPropertyTree> plane =  getStoragePlane(srcPlane);
+            if (!plane)
+                throw MakeStringException(ECLWATCH_INVALID_INPUT, "Unknown storage plane: %s", srcPlane);
+            const char * storagePlanePath = plane->queryProp("@prefix");
+            if (!startsWith(srcPath,storagePlanePath))
+                throw MakeStringException(ECLWATCH_INVALID_INPUT, "Invalid source path");
+            sourceIPReq.set("127.0.0.1"); // with storage planes, all locations are local
```
ghalliday, line 1833: Unfortunately not true... we also support bare-metal drop zones (on cloud and bare-metal).  See https://github.com/hpcc-systems/HPCC-Platform/pull/15038

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -1865,41 +1876,52 @@ bool CFileSprayEx::onSprayFixed(IEspContext &context, IEspSprayFixed &req, IEspS
...
-
         StringBuffer destFolder, destTitle, defaultFolder, defaultReplicateFolder;
+        StringBuffer gName, ipAddr;
+        bool useStoragePlane = false;
 
         const char* destNodeGroup = req.getDestGroup();
-        if(destNodeGroup == NULL || *destNodeGroup == '\0')
-            throw MakeStringException(ECLWATCH_INVALID_INPUT, "Destination node group not specified.");
+        const char* destPlane = req.getDestPlane();
+        if (!isEmptyString(destPlane))
```
ghalliday, line 1885: What functionality does destPlane provide?  I think it is identical to destCluster.

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -2406,22 +2440,28 @@ bool CFileSprayEx::onDespray(IEspContext &context, IEspDespray &req, IEspDespray
...
 
             //If the destination filename is not provided, calculate a relative filename from the logical filename
             if(!destfile || !*destfile)
             {
                 expandLogicalAsPhysical(implicitDestFile, srcname, "/");
                 destfile = implicitDestFile;
             }
+            if (destPlane)
+            {
+                destip="127.0.0.1";  // storage planes are all local
```
ghalliday, line 2463: not necessarily - see comment above.

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -1814,17 +1814,28 @@ bool CFileSprayEx::onGetDFUExceptions(IEspContext &context, IEspGetDFUExceptions
...
     StringBuffer sourcePath(srcPath);
-    sourceIPReq.set(srcIP);
-    sourceIPReq.trim();
     sourcePath.trim();
+
     if(srcxml.length() == 0)
     {
+        if (!isEmptyString(srcPlane))
+        {
+            Owned<IPropertyTree> plane =  getStoragePlane(srcPlane);
```
ghalliday, line 1827: I'm slightly concerned about security implications.  This needs to be restricted to drop zones - not any storage plane, otherwise it would allow arbitrary files to be "sprayed" from an existing storage plane.

ghalliday (changes requested):
@shamser looks like a reasonable set of changes.  A few general comments
- needs to target 8.2.0
- not sure what difference there is between destCluster and destPlane.
- need to check it doesn't introduce a security hole - does it check that it is a dropzone
- drop zones can have associated ips
-

jakesmith:
@shamser - is it best to wait to review until the regression suite failures are resolved?

dali/dfuplus/dfuplus.cpp
shamser: The despray ESP service and the despray action require the dstplane.  I added this to spray for consistency.  I didn't want users to be confused by having to used dstcluster with spray and dstplane for despray.

dali/dfuplus/dfuplus.cpp
shamser: no.  This section should not be executed if dstplane has a value. This code section is for handling the case where the dstip is empty so assume that the destination ip is localhost.   If the dstplane has a value then the destination isn't localhost so this section should not be executed.

esp/services/ws_fs/ws_fsService.cpp
shamser: It is the same at the moment. Added for consistency.  I can remove in the service but would you want it to be removed from dfuplus too?

esp/services/ws_fs/ws_fsService.cpp
shamser: What should sourceIPReq be set to? If it's empty, how should RemoteMultiFilename be handled (ws_fsService.cpp lines 1943-1948)?

shamser:
@jakesmith you can wait till I've addressed the regression.

shamser:
@ghalliday

esp/services/ws_fs/ws_fsService.cpp
ghalliday: See the linked PR for accessing the ip if it is non localhost

esp/services/ws_fs/ws_fsService.cpp
ghalliday: I don't think it is worth having two parameters to the service which do the same thing - it is adding complexity for no real benefit.  I don't have a strong opinion on the dfuplus command.  I suspect it  is simpler without.  At some point all the parameters could do with reviewing, but probably not now...

dali/dfuplus/dfuplus.cpp
ghalliday: Agreed - but your test *will* execute it if dstplane has a value.

shamser: sorry, you're right.

jakesmith (commented):
@shamser - please refresh the github review status once regression suite passed/ready for review.

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -1814,15 +1814,47 @@ bool CFileSprayEx::onGetDFUExceptions(IEspContext &context, IEspGetDFUExceptions
...
+                addNonEmptyPathSepChar(sourcePath);
+                sourcePath.append(srcPath).trim();
+            }
+            const char * hostGroup = dropZone->queryProp("@hostGroup");
+            const char * host = "127.0.0.1";
+            if (hostGroup)
+            {
+               VStringBuffer xpath("storage/hostGroups[@name='%s']", srcPlane);
+               IPropertyTree * match = getGlobalConfigSP()->queryPropTree(xpath);
+               if (match)
```
jakesmith, line 1848: if the plane has defined a hostGroup, and it's not defined, I think this should fire an error, not proceed and use localhost default ?

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -2260,25 +2292,36 @@ bool CFileSprayEx::onReplicate(IEspContext &context, IEspReplicate &req, IEspRep
...
+    }
+    destFileOut.append(destFileIn).trim();
+    dropZone->getProp("@umask", umask);
+    const char * hostGroup = dropZone->queryProp("@hostGroup");
+    const char * host = "127.0.0.1";
+    if (hostGroup)
+    {
+        VStringBuffer xpath("storage/hostGroups[@name='%s']", destPlane);
+        IPropertyTree * match = getGlobalConfigSP()->queryPropTree(xpath);
+        if (match)
```
jakesmith, line 2321: similar questions here as per comments here: https://github.com/hpcc-systems/HPCC-Platform/pull/15051#discussion_r652797724

dali/dfuplus/dfuplus.cpp
shamser: do you mean rename dstcluster to dstplane?  I agree dstplane makes more sense.  However, The FileSpray ESP service and dfuplus still uses dstcluster.  I think it better to change it every where in one go.  May be do this everywhere in a separate jira?

dali/dfuplus/dfuplus.cpp
```diff
@@ -374,9 +376,10 @@ bool CDfuPlusHelper::fixedSpray(const char* srcxml,const char* srcip,const char*
     }
 
     Owned<IClientSprayFixed> req = sprayclient->createSprayFixedRequest();
-    if(srcxml == nullptr)
+    if(isEmptyString(srcxml))
     {
-        req->setSourceIP(srcip);
+        if (!isEmptyString(srcip)) req->setSourceIP(srcip);
+        if (!isEmptyString(srcplane)) req->setSourcePlane(srcplane);
```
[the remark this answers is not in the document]
shamser: This is checked by the caller i.e. CDfuPlusHelper::spray() will error if both is supplied.

dali/dfuplus/dfuplus.cpp
```diff
@@ -739,25 +756,34 @@ int CDfuPlusHelper::despray()
...
+    }
     else
         info("\nDespraying %s\n", srcname);
 
     Owned<IClientDespray> req = sprayclient->createDesprayRequest();
     req->setSourceLogicalName(srcname);
     if(dstxml == nullptr)
     {
-        req->setDestIP(dstip);
+        if (!isEmptyString(dstplane))
```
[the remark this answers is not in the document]
shamser: I'd prefer to leave as is because (1) the extra calls isn't going to impact performance of the dfuplus cli (2) if I change for dstplane, I should be doing the same for the other similar ones and I'd prefer not to make to code changes.

dali/dfuplus/dfuplus.hpp
```diff
@@ -66,8 +66,8 @@ class CDfuPlusHelper : public CInterface, implements IInterface
     int listhistory();
     int erasehistory();
 
-    bool fixedSpray(const char* srcxml,const char* srcip,const char* srcfile,const MemoryBuffer &xmlbuf,const char* dstcluster,const char* dstname,const char *format,StringBuffer &retwuid,StringBuffer &except );
-    bool variableSpray(const char* srcxml,const char* srcip,const char* srcfile,const MemoryBuffer &xmlbuf,const char* dstcluster,const char* dstname,const char *format,StringBuffer &retwuid,StringBuffer &except );
+    bool fixedSpray(const char* srcxml,const char* srcip,const char* srcfile,const char* srcplane,const MemoryBuffer &xmlbuf,const char* dstcluster,const char* dstname,const char *format,StringBuffer &retwuid,StringBuffer &except );
+    bool variableSpray(const char* srcxml,const char* srcip,const char* srcfile,const char* srcplane,const MemoryBuffer &xmlbuf,const char* dstcluster,const char* dstname,const char *format,StringBuffer &retwuid,StringBuffer &except );
```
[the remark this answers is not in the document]
shamser: Please see https://github.com/hpcc-systems/HPCC-Platform/pull/15051/files#r653435195

dali/dfuplus/dfuplus.cpp
```diff
@@ -594,60 +600,70 @@ bool CDfuPlusHelper::variableSpray(const char* srcxml,const char* srcip,const ch
...
+        {
 #ifdef DAFILESRV_LOCAL
             progress("srcip not specified - assuming spray from local machine\n");
             srcip = ".";
 #else
-            throw MakeStringException(-1, "srcip not specified");
+            throw MakeStringException(-1, "Neither srcip nor srcplane specified");
 #endif
         }
+        if(usingSrcPlane && !isEmptyString(srcip))
```
ghalliday, line 628: In bare metal I think you can define a single drop zone with multiple IPs.  I don't know if it is used (need to check with tony K).  If so this test may be too restrictive.

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -1814,15 +1814,47 @@ bool CFileSprayEx::onGetDFUExceptions(IEspContext &context, IEspGetDFUExceptions
...
+                sourcePath.append(dropZonePlanePath);
+                addNonEmptyPathSepChar(sourcePath);
+                sourcePath.append(srcPath).trim();
+            }
+            const char * hostGroup = dropZone->queryProp("@hostGroup");
+            const char * host = "127.0.0.1";
+            if (hostGroup)
+            {
+               VStringBuffer xpath("storage/hostGroups[@name='%s']", srcPlane);
+               IPropertyTree * match = getGlobalConfigSP()->queryPropTree(xpath);
```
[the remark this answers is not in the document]
ghalliday: Could use getHostGroup() in dameta.hpp.

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -1814,15 +1814,47 @@ bool CFileSprayEx::onGetDFUExceptions(IEspContext &context, IEspGetDFUExceptions
...
+               if (match)
+                   host = match->queryProp("hosts[1]");
```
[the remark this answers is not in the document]
ghalliday: Needs some input from existing users to know if they actually use this feature in bare metal.

dali/dfuplus/dfuplus.cpp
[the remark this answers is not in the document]
ghalliday: happy for these changes to be made later - otherwise it is hard to know where to stop

ghalliday (changes requested):
Some minor comments.  Jake's comment about invalid host group pointer definitely needs addressing.

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -1840,13 +1842,11 @@ void CFileSprayEx::readAndCheckSpraySourceReq(MemoryBuffer& srcxml, const char*
...
-            const char * host = "127.0.0.1";
+            const char * host = "localhost"; // by default assume storage plane mounted on dfuserver
             if (hostGroup)
             {
-               VStringBuffer xpath("storage/hostGroups[@name='%s']", srcPlane);
-               IPropertyTree * match = getGlobalConfigSP()->queryPropTree(xpath);
-               if (match)
-                   host = match->queryProp("hosts[1]");
+                Owned<IPropertyTree> match = getHostGroup(hostGroup,true);
+                host = match->queryProp("hosts[1]");
```
ghalliday, line 1849: Theoretically this could be freed when the scope is left.  Either assign in the block or clone when saving.

Same on other call below.

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -1814,19 +1814,51 @@ bool CFileSprayEx::onGetDFUExceptions(IEspContext &context, IEspGetDFUExceptions
...
+    StringBuffer sourcePath;
+
     if(srcxml.length() == 0)
     {
+        if (!isEmptyString(srcPlane))
+        {
+            Owned<IPropertyTree> dropZone = getDropZonePlane(srcPlane);
+            if (!dropZone)
+                throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "Unknown landing zone: %s", srcPlane);
+            if (!isEmptyString(srcIP))
```
ghalliday, line 1829: I'm not 100% sure this is illegal.  I think it should probably be allowed if there is a "host" element within the hostGroup that matches.  I think I have seen examples with multiple ips.

ghalliday (commented):
@shamser looking good.  A couple of (hopefully) final comments.

dali/dfuplus/dfuplus.cpp
jakesmith: not critical, for me because a 'plane' param is being introduced in same method that there's a 'cluster' based param, this is the time to change this one, but in general others can be tackled incrementally.

esp/services/ws_fs/ws_fsService.cpp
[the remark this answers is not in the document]
jakesmith: For how since it is possible to define a plane without hosts, and/or there may be greater than one host,
perhaps it should error if there isn't exactly one hosts entry, e.g. if (1 != match->getCount("hosts")) ...

jakesmith (commented):
1 comment from me: https://github.com/hpcc-systems/HPCC-Platform/pull/15051#discussion_r653591909

esp/services/ws_fs/ws_fsService.cpp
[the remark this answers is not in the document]
shamser: @ghalliday mentioned that multiple host ip's are being used.  To support this,  both srcplane and srcip may be provided to dfuplus so that the speciifc host ip can be selected (from this list of ip's).  @jakesmith please see last commit.

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -2313,13 +2329,29 @@ void CFileSprayEx::getDropZoneInfoByDestPlane(double clientVersion, const char*
...
         Owned<IPropertyTree> match = getHostGroup(hostGroup,true);
-        host = match->queryProp("hosts[1]");
+        if (!hostip.isEmpty())
+        {
+            Owned<IPropertyTreeIterator> hostIter = match->getElements("hosts");
+            ForEach (*hostIter)
+            {
+                const char *knownIP = hostIter->query().queryProp(nullptr);
+                if (strcmp(knownIP, hostip.str())==0)
+                    return;
```
jakesmith, line 2342: should this be a break; , to fall through so that sourcePathReq is filled in?

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -2313,13 +2329,29 @@ void CFileSprayEx::getDropZoneInfoByDestPlane(double clientVersion, const char*
...
+            }
```
jakesmith, line 2343: minor: I think iterator code could be replaced by:
```
VStringBuffer xexpr(".[hosts='%s']", hostip.str());
if (match->hasProp(xexpr))
```

esp/services/ws_fs/ws_fsService.cpp
jakesmith: above comment in wrong place, will move

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -1842,13 +1840,31 @@ void CFileSprayEx::readAndCheckSpraySourceReq(MemoryBuffer& srcxml, const char*
...
+                if (!isEmptyString(srcIP))
+                {
+                    Owned<IPropertyTreeIterator> hostIter = match->getElements("hosts");
+                    ForEach (*hostIter)
+                    {
+                        const char *knownIP = hostIter->query().queryProp(nullptr);
+                        if (strcmp(knownIP, srcIP)==0)
+                        {
+                            sourceIPReq.set(knownIP);
+                            return;
```
jakesmith, line 1855: should this be a break; , to fall through so that sourcePathReq is filled in?

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -1842,13 +1840,31 @@ void CFileSprayEx::readAndCheckSpraySourceReq(MemoryBuffer& srcxml, const char*
...
             {
                 Owned<IPropertyTree> match = getHostGroup(hostGroup,true);
-                host = match->queryProp("hosts[1]");
+                if (!isEmptyString(srcIP))
+                {
+                    Owned<IPropertyTreeIterator> hostIter = match->getElements("hosts");
+                    ForEach (*hostIter)
+                    {
+                        const char *knownIP = hostIter->query().queryProp(nullptr);
+                        if (strcmp(knownIP, srcIP)==0)
```
jakesmith, line 1852: minor: I think this iterator code could be replaced by:
```
VStringBuffer xexpr(".[hosts='%s']", hostip.str());
if (match->hasProp(xexpr))
```

jakesmith (commented):
1 question, 1 minor comment

dali/dfuplus/dfuplus.cpp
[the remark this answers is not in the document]
ghalliday: no, it is ok for both to be supplied.

esp/services/ws_fs/ws_fsService.cpp
ghalliday: agreed

ghalliday (commented):
@shamser very close.  One item left I think

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -1845,16 +1845,10 @@ void CFileSprayEx::readAndCheckSpraySourceReq(MemoryBuffer& srcxml, const char*
...
-                        if (strcmp(knownIP, srcIP)==0)
-                        {
-                            sourceIPReq.set(knownIP);
-                            return;
-                        }
-                    }
+                    VStringBuffer xexpr(".[hosts='%s']", srcIP);
+                    if (!match->hasProp(xexpr))
+                        throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "srcip %s is not valid storage plane %s", srcIP, srcPlane);
+                    sourceIPReq.set(match->queryProp(srcIP));
```
ghalliday, line 1851: I think you mean

    sourceIPReq.set(srcIP);

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -2334,14 +2328,10 @@ void CFileSprayEx::getDropZoneInfoByDestPlane(double clientVersion, const char*
...
-            {
-                const char *knownIP = hostIter->query().queryProp(nullptr);
-                if (strcmp(knownIP, hostip.str())==0)
-                    return;
-            }
-            throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "destip %s is not valid storage plane %s", hostip.str(), destPlane);
+            VStringBuffer xexpr(".[hosts='%s']", hostip.str());
+            if (!match->hasProp(xexpr))
+                throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "destip %s is not valid storage plane %s", hostip.str(), destPlane);
+            return;
```
ghalliday, line 2334: probably better not to include the return - in case someone added something to the end of the function.

ghalliday (commented):
@shamser a couple of comments

ghalliday (approved):

esp/services/ws_fs/ws_fsService.cpp
jakesmith: @ghalliday @shamser - looks like this still needs to be fixed?

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -1814,19 +1814,61 @@ bool CFileSprayEx::onGetDFUExceptions(IEspContext &context, IEspGetDFUExceptions
...
+                sourcePath.append(srcPath).trim();
+            }
+            const char * hostGroup = dropZone->queryProp("@hostGroup");
+            if (hostGroup)
+            {
+                Owned<IPropertyTree> match = getHostGroup(hostGroup,true);
+                if (!isEmptyString(srcIP))
+                {
+                    VStringBuffer xexpr(".[hosts='%s']", srcIP);
+                    if (!match->hasProp(xexpr))
```
jakesmith, line 1849: > minor: I think iterator code could be replaced by:
> 
> ```
> VStringBuffer xexpr(".[hosts='%s']", hostip.str());
> if (match->hasProp(xexpr))
> ```

I did say I think .. I've just tested an xpath test like this and it fails.
And I don't think there's any way to test the value (as opposed to the value of an attribute or subprop) in a qualifier like this (there is in the full xpath spec. you can do hosts[texst() = <value>], but we don't support that.
So unfortunately, you'll have to revert this to an iterator to find match.

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -2260,25 +2302,45 @@ bool CFileSprayEx::onReplicate(IEspContext &context, IEspReplicate &req, IEspRep
...
+    }
+    destFileOut.append(destFileIn).trim();
+    dropZone->getProp("@umask", umask);
+    const char * hostGroup = dropZone->queryProp("@hostGroup");
+    if (hostGroup)
+    {
+        Owned<IPropertyTree> match = getHostGroup(hostGroup,true);
+        if (!hostip.isEmpty())
+        {
+            VStringBuffer xexpr(".[hosts='%s']", hostip.str());
```
jakesmith, line 2331: See : https://github.com/hpcc-systems/HPCC-Platform/pull/15051/files#r656257078

esp/services/ws_fs/ws_fsService.cpp
jakesmith: sorry, looks like I was commend on an out of date commit. Ignore above comment.

jakesmith (changes requested):
@shamser - the .[hosts=..] xpath syntax doesn't work after all.

ghalliday (approved):

jakesmith (approved):
@shamser - looks good to merge after squashing.

dali/dfuplus/dfuplus.cpp
```diff
@@ -509,27 +516,27 @@ bool CDfuPlusHelper::variableSpray(const char* srcxml,const char* srcip,const ch
                 req->setNoSourceCsvSeparator(true);
         }
         const char* terminator = globals->queryProp("terminator");
-        if(terminator && *terminator)
+        if(!isEmptyString(terminator))
             req->setSourceCsvTerminate(terminator);
         const char* quote = globals->queryProp("quote");
-        if(quote)
+        if(!isEmptyString(quote))
```
ghalliday, line 522: Unfortunately this change is wrong - see the comment on the next line

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -2260,25 +2313,56 @@ bool CFileSprayEx::onReplicate(IEspContext &context, IEspReplicate &req, IEspRep
...
+            if (!ipAddressMatches)
+                throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "destip %s is not valid storage plane %s", hostip.str(), destPlane);
+        }
+        else
+            hostip.set(match->queryProp("hosts[1]"));
     }
     else
     {
-        destFileOut.append(destFileIn);
+        if (!hostip.isEmpty())
```
ghalliday, line 2362: Should "localhost" be allowed?

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -1814,19 +1814,72 @@ bool CFileSprayEx::onGetDFUExceptions(IEspContext &context, IEspGetDFUExceptions
...
+                    if (!ipAddressMatches)
+                        throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "srcip %s is not valid storage plane %s", srcIP, srcPlane);
+                    sourceIPReq.set(srcIP);
+                }
+                else
+                    sourceIPReq.set(match->queryProp("hosts[1]"));
+            }
+            else
+            {
+                if (!isEmptyString(srcIP))
```
ghalliday, line 1869: should "localhost" be allowed?

ghalliday (changes requested):
@shamser two final issues (blank quote and allowing localhost)

esp/services/ws_fs/ws_fsService.cpp
shamser: I think it would be confusing to allow the user to use "localhost" parameter with a mounted storage plane.  It implies that the storage plane may be mounted on their local machine.

esp/services/ws_fs/ws_fsService.cpp
shamser: Please see previous comment on this: https://github.com/hpcc-systems/HPCC-Platform/pull/15051#discussion_r657054620

shamser: Eclwatch does provide the srcIP address so I'll remove this check.

esp/services/ws_fs/ws_fsService.cpp
shamser: Eclwatch does provide the srcIP address so I'll remove this check.

ghalliday (approved):

PR #15072, opened 2021-06-17
HPCC-25824 Remove remaining references to dkc

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:


dali/dfu/dfurun.cpp
```diff
@@ -54,10 +54,6 @@ test multiclusteradd with replicate
 
 extern ILogMsgHandler * fileMsgHandler;
 
-extern void doKeyDiff(IFileDescriptor *oldf,IFileDescriptor *newf,IFileDescriptor *patchf); // patchf out
```
richardkchapman, line 57: Keydiff/keypatch are not same as dkc are they?

ghalliday: I think the dfu functions were implemented with it - see the deleted file.  @jakesmith is that right?
Pushed again because failed to check in a couple of changes.

jakesmith: I wouldn't have known, but yes it looks that way.

jakesmith (commented):
@ghalliday - It looks fine as far it goes.
Are we sure we want to remove this functionality?

No idea whether works in bare-metal and doubt if ever used, so that may be reason enough.
But to remove implies we will never want to revisit.

Is it that we feel that key diff/patch is not viable, in which case what about the activity implementations ?
Or just that dfuserver implementation is redundant or at least would need reimplementing if we ever wanted it?

ghalliday:
I'm not sure it is actually removing any functionality.  As far as I read the code the keydiff/keypatch is implemented by executing dkcslave as a process in a similar way to ftslave.
I can't find any version of the sources that actually build dkcslave - even back to 3.2.x.  I suspect if we had the pre oss source code I could find it.
If I am mis reading it then yes we should keep the code.

ghalliday:
I think it does still work - I'll close and revisit what can be removed.  A much smaller subset.

PR #15081, opened 2021-06-18
HPCC-26083 Containerized Roxie should not copy files multiple times 

richardkchapman:
## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Cloud-compatibility
  - [x] Premature optimization

## Smoketest:

### DIFF
--- a/roxie/ccd/ccdfile.cpp
+++ b/roxie/ccd/ccdfile.cpp
@@ -169,7 +169,15 @@ public:
         }
         copying = _copying; 
     }
-
+    virtual void dump() const
+    {
+        CriticalBlock b(crit);
+        DBGLOG("LazyFileIO object %s has %d sources:", queryFilename(), sources.ordinality());
+        ForEachItemIn(idx, sources)
+        {
+            DBGLOG("%c %s", idx==currentIdx ? '*' : ' ', sources.item(idx).queryFilename());
+        }
+    }
     virtual bool isCopying() const
     {
         CriticalBlock b(crit);
@@ -401,7 +409,7 @@ public:
     virtual void setSize(offset_t size) { throwUnexpected(); }
     virtual offset_t appendFile(IFile *file,offset_t pos,offset_t len) { throwUnexpected(); return 0; }
 
-    virtual const char *queryFilename() { return logical->queryFilename(); }
+    virtual const char *queryFilename() const { return logical->queryFilename(); }
     virtual bool isAliveAndLink() const { return CInterface::isAliveAndLink(); }
 
     virtual IMemoryMappedFile *getMappedFile() override
@@ -478,17 +486,26 @@ public:
 
     void copyComplete()
     {
+        CriticalBlock b(crit);
+        setFailure(); // lazyOpen will then reopen it...
+        currentIdx = 0;
+        remote = false;
+        copying = false;
+        sources.kill();
+        sources.add(*logical.getLink(), 0);
+        if (!lazyOpen)
+            _checkOpen();
+    }
+
+    bool checkCopyComplete()
+    {
+        CriticalBlock b(crit);
+        if (logical->exists())   // MORE - do we need to check data/size etc? do we have the info to do so?
         {
-            CriticalBlock b(crit);
-            setFailure(); // lazyOpen will then reopen it...
-            currentIdx = 0;
-            remote = false;
-            copying = false;
-            sources.kill();
-            sources.add(*logical.getLink(), 0);
-            if (!lazyOpen)
-                _checkOpen();
+            copyComplete();
+            return true;
         }
+        return false;
     }
 
     virtual IFile *querySource() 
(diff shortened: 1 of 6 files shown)


roxie/ccd/ccdfile.cpp
```diff
@@ -1418,9 +1427,49 @@ class CRoxieFileCache : implements IRoxieFileCache, implements ICopyFileProgress
...
+                        buddyChecking = true;
+                    }
+                    ForEachItemIn(idx, checkBuddies)
+                    {
+                        ILazyFileIO &check = checkBuddies.item(idx);
+                        if (check.isRemote())
+                        {
+                            check.setCopying(false);
+                        }
+                        if (check.isRemote())
```
ghalliday, line 1458: Should have a comment to indicate why it is being rechecked.  (Could be nested within the previous if to make it clearer it isn't a mistake.)

roxie/ccd/ccdmain.cpp
```diff
@@ -102,6 +102,8 @@ bool oneShotRoxie = false;
 unsigned udpMulticastBufferSize = 262142;
 #ifndef _CONTAINERIZED
 bool roxieMulticastEnabled = true;
+#else
+unsigned myChannel;
```
ghalliday, line 106: yuk, but I guess it is ok.

ghalliday (commented):
@richardkchapman looks like ti should work.  One minor comment to clarify the code.

roxie/ccd/ccdfile.cpp
richardkchapman: Done

roxie/ccd/ccdfile.cpp
```diff
@@ -1418,11 +1459,63 @@ class CRoxieFileCache : implements IRoxieFileCache, implements ICopyFileProgress
...
+                        {
+                            DBGLOG("No more data files to copy");
+                            reportedFilesToCopy = false;
+                        }
+                    }
+                }
+#endif
+                unsigned elapsed = msTick()-start;
+                if (elapsed >= 10*60*1000)
+                {
```
ghalliday, line 1515: Is it a problem, that this is unlikely to be executed in containerized?  Does start need ressetting inside this if instead of at the head of the loop?

roxie/ccd/ccdfile.cpp
```diff
@@ -1418,11 +1459,63 @@ class CRoxieFileCache : implements IRoxieFileCache, implements ICopyFileProgress
...
+                        if (traceRemoteFiles)
+                            DBGLOG("Checking whether someone has copied file %s for me", check.queryFilename());
+                        if (check.isRemote())
+                        {
+                            if (traceRemoteFiles)
+                                check.dump();
+                            if (!check.checkCopyComplete())   // Recheck whether there is a local file we can open
+                            {
+                                CriticalBlock b1(crit);
+                                buddyCopying.append(check);
```
ghalliday, line 1496: likely to be better to add all the files that are still being copied in a single critical section?

ghalliday (commented):
@richardkchapman looks good.  A couple of minor comments.

roxie/ccd/ccdmain.cpp
```diff
@@ -675,6 +682,13 @@ int CCD_API roxie_main(int argc, const char *argv[], const char * defaultYaml)
...
             }
+#ifdef _CONTAINERIZED
+            if (agentChannels.size() != 1)
+                throw makeStringExceptionV(ROXIE_INTERNAL_ERROR, "Invalid channel specification %s - single channel expected", channels);
+            myChannel = agentChannels[0].first;
+            if (myChannel > numChannels)
+                throw makeStringExceptionV(ROXIE_INTERNAL_ERROR, "Invalid channel specification %s - value out of range", channels);
+#endif
         }
 #ifdef _CONTAINERIZED
```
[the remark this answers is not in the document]
richardkchapman: I suspect it should be an error if channels not specified - but if it defaults to 1 then so should myChannel. I will look.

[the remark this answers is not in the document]
richardkchapman: Actually it may be correct for channels not to be specified, if in separate server mode. Leaving it at 0 is ok there, I suppose.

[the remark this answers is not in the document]
richardkchapman: It does mean that the server-only nodes will not participate in the copying - but that's ok.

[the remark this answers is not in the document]
richardkchapman: Slightly more concerning is that the localAgent case will also be set to 0 so no-one will ever copy.

roxie/ccd/ccdfile.cpp
richardkchapman: Yes - will fix

roxie/ccd/ccdfile.cpp
richardkchapman: Not sure it will make enough difference to worry about.

roxie/ccd/ccdmain.cpp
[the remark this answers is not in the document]
richardkchapman: I think that's actually ok - you need a separate "pull roxie" set up for that case.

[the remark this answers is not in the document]
ghalliday: Would be worth a comment.

ghalliday (approved):
I think probably ok, although not so sure about the crit contention.

PR #15088, opened 2021-06-21
HPCC-26022 Add metrics unit tests

kenrowland:
Added unit tests for metric framework and base periodic sink class

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jmetrics.cpp
+++ b/system/jlib/jmetrics.cpp
@@ -28,18 +28,10 @@ MODULE_EXIT()
 }
 
 
-struct hpccMetrics::SinkInfo
-{
-    explicit SinkInfo(MetricSink *_pSink) : pSink{_pSink} {}
-    MetricSink *pSink = nullptr;             // ptr to the sink
-    std::vector<std::string> reportMetrics;   // vector of metrics to report (empty for none)
-};
-
 MetricsReporter &hpccMetrics::queryMetricsReporter()
 {
     return *metricsReporter.query([] { return new MetricsReporter; });
 }
-
 
 MetricsReporter::~MetricsReporter()
 {
@@ -200,6 +192,23 @@ MetricSink *MetricsReporter::getSinkFromLib(const char *type, const char *sinkNa
         throw MakeStringException(MSGAUD_operator, "getSinkFromLib - Unable to load sink lib (%s)", libName.c_str());
     }
     return pSink;
+}
+
+// Method for use when testing
+void MetricsReporter::addSink(MetricSink *pSink, const char *name)
+{
+    //
+    // Add the sink if it does not already exist, otherwise delete the sink because
+    // we are taking ownership.
+    auto sinkIt = sinks.find(name);
+    if (sinkIt == sinks.end())
+    {
+        sinks.insert({std::string(name), std::unique_ptr<SinkInfo>(new SinkInfo(pSink))});
+    }
+    else
+    {
+        delete pSink;
+    }
 }
 
 
--- a/system/jlib/jmetrics.hpp
+++ b/system/jlib/jmetrics.hpp
@@ -120,7 +120,7 @@ class jlib_decl CounterMetric : public MetricVal
 public:
     CounterMetric(const char *name, const char *description) :
         MetricVal{name, description, MetricType::METRICS_COUNTER}  { }
-    void inc(uint64_t val)
+    void inc(uint64_t val = 1)
     {
         value.fetch_add(val);
     }
@@ -218,7 +218,12 @@ protected:
 
 extern "C" { typedef hpccMetrics::MetricSink* (*getSinkInstance)(const char *, const IPropertyTree *pSettingsTree); }
 
-struct SinkInfo;
+struct SinkInfo
+{
+    explicit SinkInfo(MetricSink *_pSink) : pSink{_pSink} {}
+    MetricSink *pSink = nullptr;             // ptr to the sink
+    std::vector<std::string> reportMetrics;   // vector of metrics to report (empty for none)
+};
 
 class jlib_decl MetricsReporter
 {
@@ -226,6 +231,7 @@ public:
     MetricsReporter() = default;
     ~MetricsReporter();
     void init(IPropertyTree *pMetricsTree);
+    void addSink(MetricSink *pSink, const char *name);  // for use by unit tests
     void addMetric(const std::shared_ptr<IMetric> &pMetric);
     void startCollecting();
     void stopCollecting();
--- a/testing/unittests/metrics/MetricFrameworkTests.cpp
+++ b/testing/unittests/metrics/MetricFrameworkTests.cpp
@@ -1,0 +1,189 @@
+/*##############################################################################
+
+    HPCC SYSTEMS software Copyright (C) 2021 HPCC Systems®.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+############################################################################## */
+
+#ifdef _USE_CPPUNIT
+
+#include <cppunit/TestFixture.h>
+#include "unittests.hpp"
+#include <algorithm>
+
+#include "jmetrics.hpp"
+
+using namespace hpccMetrics;
+
+class MetricFrameworkTestSink : public MetricSink
+{
+public:
+    explicit MetricFrameworkTestSink(const char *name) :
+        MetricSink(name, "test") { }
+
+    ~MetricFrameworkTestSink() override = default;
+
+    void startCollection(MetricsReporter *_pReporter) override
+    {
+        pReporter = _pReporter;
+        isCollecting = true;
+    }
+
+    void stopCollection() override
+    {
+        isCollecting = false;
+    }
+
+    std::vector<std::shared_ptr<IMetric>> getReportMetrics() const
+    {
+        return pReporter->queryMetricsForReport(name);
+    }
+
+public:
+    bool isCollecting = false;
+};
+
+
+class MetricFrameworkTests : public CppUnit::TestFixture
+{
+public:
+    MetricFrameworkTests()
+    {
+        pTestSink = new MetricFrameworkTestSink("testsink");
+        frameworkTestReporter.addSink(pTestSink, "testsink");
+    }
+
+    ~MetricFrameworkTests() = default;
+
+    CPPUNIT_TEST_SUITE(MetricFrameworkTests);
+        CPPUNIT_TEST(Test_counter_metric_increments_properly);
+        CPPUNIT_TEST(Test_gauge_metric_updates_properly);
+        CPPUNIT_TEST(Test_custom_metric);
+        CPPUNIT_TEST(Test_reporter_calls_sink_to_start_and_stop_collection);
+        CPPUNIT_TEST(Test_reporter_manages_metrics_properly);
+    CPPUNIT_TEST_SUITE_END();
+
+protected:
+
+
+    void Test_counter_metric_increments_properly()
+    {
+        std::shared_ptr<CounterMetric> pCounter = std::make_shared<CounterMetric>("test-counter", "description");
+        CPPUNIT_ASSERT_EQUAL(0, static_cast<int>(pCounter->queryValue()));
+
+        //
+        // Test default increment (by 1)
+        pCounter->inc();
+        int counterValue = pCounter->queryValue();
+        CPPUNIT_ASSERT_EQUAL(1, counterValue);
+
+        //
+        // Test increment by > 1
+        pCounter->inc(2);
+        counterValue = pCounter->queryValue();
+        CPPUNIT_ASSERT_EQUAL(3, counterValue);
+    }
+
+
+    void Test_gauge_metric_updates_properly()
+    {
+        std::shared_ptr<GaugeMetric> pGauge = std::make_shared<GaugeMetric>("test-gauge", "description");
+        int gaugeValue = pGauge->queryValue();
+        CPPUNIT_ASSERT_EQUAL(0, gaugeValue);
+
+        //
+        // Test initial setting of gauge
+        pGauge->set(25);
+        gaugeValue = pGauge->queryValue();
+        CPPUNIT_ASSERT_EQUAL(25, gaugeValue);
+
+        //
+        // Test updating gauge
+        pGauge->add(10);
+        gaugeValue = pGauge->queryValue();
+        CPPUNIT_ASSERT_EQUAL(35, gaugeValue);
+
+        pGauge->add(-5);
+        gaugeValue = pGauge->queryValue();
+        CPPUNIT_ASSERT_EQUAL(30, gaugeValue);
+    }
+
+
+    void Test_custom_metric()
+    {
+        int customCounter = 0;
+        std::shared_ptr<CustomMetric<int>> pCustomCounter = std::make_shared<CustomMetric<int>>("custom-counter", "description", METRICS_COUNTER, customCounter);
+        int customCounterValue = pCustomCounter->queryValue();
+        CPPUNIT_ASSERT_EQUAL(0, customCounterValue);
+
+        customCounter++;
+        customCounterValue = pCustomCounter->queryValue();
+        CPPUNIT_ASSERT_EQUAL(1, customCounterValue);
+    }
+
+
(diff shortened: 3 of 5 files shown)


AttilaVamos:
@kenrowland Maybe I am wrong, but it seems your unittests are not part of the platform unittests collection. If this is true, what is the plan to execute them?

kenrowland:
@AttilaVamos These are new tests. Once reviewed and approved I was going to ask you to add them to the list that is run.

testing/unittests/metrics/sinks/periodic/PeriodicSinkTests.cpp
```diff
@@ -0,0 +1,122 @@
...
+            // from the wait period multiple used, then we are close enough
+            int multiple = 5;
+            reporter.startCollecting();
+
+            //
+            // Check that the sink is collecting
+            CPPUNIT_ASSERT_MESSAGE("Expected sink to report it was collecting", pPeriodicTestSink->isCurrentlyCollectiing());
+
+            // wait... then stop collecing
+            sleep(multiple * period);
```
ghalliday, line 102: This will is probably too long for a standard unit test.  May need to mark it as a stress test.

ghalliday (changes requested):
@kenrowland good to have some tests.  A few summary comments:
- need to have it integrated with the normal unit test program
- Would be better to cut down on the boilerplate/verbosity and combine tests

AttilaVamos:
@kenrowland  Please rebase your code. There is a merged fix for fetch.ecl failing on Thor in GitHub Smoketest Action.

kenrowland:
@ghalliday One reason these tests are separate is that the current unit tests load libraries that instantiate metrics as part of their static initialization. In order to fully test metrics, we need a clean environment. Therefore I separated these unit tests into separate programs.

ghalliday:
> @ghalliday One reason these tests are separate is that the current unit tests load libraries that instantiate metrics as part of their static initialization. In order to fully test metrics, we need a clean environment. Therefore I separated these unit tests into separate programs.
> 
> Once these tests are approved, they should be added to the set of unit tests that are run by the smoke test.

Yes, and I don't think that separate programs is a viable solution - it will break as soon as a metric is added to jlib.  You need to use a different mechanism.  See my comment in one of the (hidden?) conversations.

kenrowland:
@ghalliday To keep the tests all in the same executable, I will investigate simply instantiating a private metrics reporter object for testing purposes. Perhaps the ability to add a sink directly and not through a lib can be an added function of that class as opposed to being part of the base reporter class.

kenrowland:
@ghalliday Please see changes. 

@AttilaVamos Rebased to latest master. Let me know if there are any problems. Also, tests were moved to run as part of regular unit tests so no need to add them separately.

ghalliday:
@kenrowland the default branch should be 8.2.x for tests/fixes, master would be for new major functionality (for 8.4.0)

system/jlib/jmetrics.hpp
```diff
@@ -218,14 +218,20 @@ class jlib_decl PeriodicMetricSink : public MetricSink
 
 extern "C" { typedef hpccMetrics::MetricSink* (*getSinkInstance)(const char *, const IPropertyTree *pSettingsTree); }
 
-struct SinkInfo;
+struct SinkInfo
```
ghalliday, line 221: Should be in the cpp file as it was before.

testing/unittests/metrics/framework/MetricFrameworkTests.cpp
```diff
@@ -0,0 +1,256 @@
...
+        CPPUNIT_TEST(Test_removes_destroyed_metric_after_collecting_started);
+    CPPUNIT_TEST_SUITE_END();
+
+protected:
+
+
+    void Test_it_increments_counter_metric_by_1_by_default()
+    {
+        std::shared_ptr<CounterMetric> pCounter = std::make_shared<CounterMetric>("test-counter", "description");
+        CPPUNIT_ASSERT_EQUAL_MESSAGE("Counter metric expected initial value to be 0", 0, static_cast<int>(pCounter->queryValue()));
```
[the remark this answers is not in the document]
ghalliday: I think the compelling reason is the last of those - it takes me at least twice as long to understand the long statement.  It is also testing something that will almost certainly never fail, so little point in explaining it in detail.

testing/unittests/metrics/MetricFrameworkTests.cpp
```diff
@@ -0,0 +1,190 @@
...
+static MetricFrameworkTestSink *pTestSink = new MetricFrameworkTestSink("testsink");
+bool sinkAdded = false;
+
+
+class MetricFrameworkTests : public CppUnit::TestFixture
+{
+public:
+    MetricFrameworkTests()
+    {
+        if (!sinkAdded)
```
ghalliday, line 73: minor: Cleaner to have the reporter as a member, and avoid the need for test/reset

testing/unittests/metrics/MetricFrameworkTests.cpp
```diff
@@ -0,0 +1,190 @@
...
+        frameworkTestReporter.addMetric(pCounter);
+        frameworkTestReporter.addMetric(pGauge);
+        numAdded = 2;
+
+        frameworkTestReporter.startCollecting();
+
+        //
+        // Make sure the initial list is correct
+        auto numMetrics = frameworkTestReporter.queryMetricsForReport("testsink").size();
+        CPPUNIT_ASSERT_EQUAL_MESSAGE(VStringBuffer("Expected %d metrics to be returned", numAdded).str(), numAdded, static_cast<int>(numMetrics));
```
ghalliday, line 163: This is a good example of an assert that is much harder to read because of the message.

CPPUNIT_ASSERT_EQUAL(numAdded, static_cast<int>(numMetrics));

even better if numAdded had the same type (size_t?) so there would be no need for the cast.

testing/unittests/metrics/MetricFrameworkTests.cpp
```diff
@@ -0,0 +1,190 @@
...
+    {
+        isCollecting = false;
+    }
+
+public:
+    bool isCollecting = false;
+};
+
+
+MetricsReporter frameworkTestReporter;
```
ghalliday, line 62: should be static (or moved into the class - see below)

testing/unittests/metrics/PeriodicSinkTests.cpp
```diff
@@ -0,0 +1,123 @@
...
+    public:
+        PeriodicSinkTests()
+        {
+            if (pPeriodicTestSink == nullptr)
+            {
+                //
+                // Load the settings then set the period using the global var
+                Owned<IPropertyTree> pSettings = createPTreeFromYAMLString(periodicSinkSettingsTestYml, ipt_none, ptr_ignoreWhiteSpace, nullptr);
+                pSettings->setPropInt("@period", period);
+                pPeriodicTestSink = new PeriodicTestSink("periodic_test_sink", pSettings);
```
ghalliday, line 81: This will leak, although it will probably avoid being reported as a leak because it is assigned as a static.

Same comments about making the MetricsReporter a member.

testing/unittests/metrics/PeriodicSinkTests.cpp
```diff
@@ -0,0 +1,123 @@
...
+    int numCollections = 0;
+};
+
+
+MetricsReporter periodicSinkTestReporter;
+
+const char *periodicSinkSettingsTestYml = R"!!(period: 2
+)!!";
+
+int period = 2;
```
ghalliday, line 66: static (const) to avoid clashing with other unit tests that happen to use a variable with the same name.

ghalliday (commented):
@kenrowland they look better.  A few comments.

testing/unittests/metrics/MetricFrameworkTests.cpp
kenrowland: I removed all message asserts along with all explicit casts of test values.

testing/unittests/metrics/PeriodicSinkTests.cpp
kenrowland: Done.

testing/unittests/metrics/PeriodicSinkTests.cpp
kenrowland: Moved.

kenrowland:
@ghalliday Please see latest changes.

testing/unittests/metrics/MetricFrameworkTests.cpp
```diff
@@ -0,0 +1,193 @@
...
+    bool isCollecting = false;
+};
+
+
+class MetricFrameworkTests : public CppUnit::TestFixture
+{
+public:
+    MetricFrameworkTests()
+    {
+        pFrameworkTestReporter = new MetricsReporter();
```
ghalliday, line 62: minor: curious why you have a pointer, rather than an object member.  (It would allow you to use a factory method...)

system/jlib/jmetrics.cpp
```diff
@@ -28,19 +28,18 @@ MODULE_EXIT()
 }
 
 
-struct hpccMetrics::SinkInfo
-{
-    explicit SinkInfo(MetricSink *_pSink) : pSink{_pSink} {}
-    MetricSink *pSink = nullptr;             // ptr to the sink
-    std::vector<std::string> reportMetrics;   // vector of metrics to report (empty for none)
-};
+//struct hpccMetrics::SinkInfo
```
ghalliday, line 31: Should be deleted if moved to the hpp

system/jlib/jmetrics.hpp
ghalliday: ok.  Could be avoided with a factory function, but probably not worth it.

ghalliday (commented):
@kenrowland only change needed is to delete the commented out code.  Please squash as well.
Also, I suspect this be targeting 8.2.x rather than master?

testing/unittests/metrics/MetricFrameworkTests.cpp
kenrowland: As discussed, an artifact from previous work. Moved it to be a member variable.

kenrowland:
@ghalliday Please look at what are hopefully final changes. If good, I will squash for merging.

system/jlib/jmetrics.cpp
```diff
@@ -202,6 +194,18 @@ MetricSink *MetricsReporter::getSinkFromLib(const char *type, const char *sinkNa
...
 
+// Method for use when testing
+void MetricsReporter::addSink(MetricSink *pSink, const char *name)
+{
+    //
+    // Add the sink if it does not already exist
+    auto sinkIt = sinks.find(name);
+    if (sinkIt == sinks.end())
+    {
+        sinks.insert({std::string(name), std::unique_ptr<SinkInfo>(new SinkInfo(pSink))});
```
ghalliday, line 205: Will pSink leak?  It looks like that is generally true (for initializeSinks as well).  If this function takes ownership then it should delete pSink if it isn't added.

ghalliday (commented):
@kenrowland looks good.  Only comment is about leaking the sinks (which I think was a pre-existing problem.)

system/jlib/jmetrics.cpp
kenrowland: The destructor for the reporter deletes the registered sinks.

testing/unittests/metrics/MetricFrameworkTests.cpp
```diff
@@ -0,0 +1,189 @@
...
+        gaugeValue = pGauge->queryValue();
+        CPPUNIT_ASSERT_EQUAL(25, gaugeValue);
+
+        //
+        // Test updating gauge
+        pGauge->add(10);
+        gaugeValue = pGauge->queryValue();
+        CPPUNIT_ASSERT_EQUAL(35, gaugeValue);
+
+        pGauge->add(-5);
```
jakesmith, line 116: this will need rebasing once your other PR is merged.

testing/unittests/metrics/MetricFrameworkTests.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+
+    void Test_reporter_manages_metrics_properly()
+    {
+        int numAdded;
+        std::shared_ptr<CounterMetric> pCounter = std::make_shared<CounterMetric>("test-counter", "description");
+        std::shared_ptr<GaugeMetric> pGauge = std::make_shared<GaugeMetric>("test-gauge", "description");
+        frameworkTestReporter.addMetric(pCounter);
+        frameworkTestReporter.addMetric(pGauge);
+        numAdded = 2;
```
jakesmith, line 151: style: might as well declare here.

testing/unittests/metrics/PeriodicSinkTests.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+    void doCollection() override
+    {
+        numCollections++;
+    }
+
+public:
+    bool prepareCalled = false;
+    bool stopCollectionNotificationCalled = false;
+    int numCollections = 0;
```
jakesmith, line 57: should prob. be an unsigned

testing/unittests/metrics/PeriodicSinkTests.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+
+class PeriodicSinkTests : public CppUnit::TestFixture
+{
+public:
+    PeriodicSinkTests()
+    {
+        //
+        // Load the settings then set the period using the global var
+        Owned<IPropertyTree> pSettings = createPTreeFromYAMLString(periodicSinkSettingsTestYml, ipt_none, ptr_ignoreWhiteSpace, nullptr);
```
jakesmith, line 74: params are defaults, so not needed.

testing/unittests/metrics/PeriodicSinkTests.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+        //
+        // Check that the sink called to prepare for collection
+        CPPUNIT_ASSERT(pPeriodicTestSink->prepareCalled);
+
+        // wait... then stop collecting
+        sleep(multiple * period);
+        periodicSinkTestReporter.stopCollecting();
+
+        int numCollections = pPeriodicTestSink->numCollections;
```
jakesmith, line 105: trivial: unsigned?

testing/unittests/metrics/PeriodicSinkTests.cpp
```diff
@@ -0,0 +1,123 @@
...
+        CPPUNIT_ASSERT_EQUAL(true, numReportsCorrect);
+
+        //
+        // Verify collection was stopped
+        CPPUNIT_ASSERT(pPeriodicTestSink->stopCollectionNotificationCalled);
+    }
+
+protected:
+    MetricsReporter periodicSinkTestReporter;
+    PeriodicTestSink *pPeriodicTestSink;
```
jakesmith, line 117: style/consistency: we normally initialize all members upfront, even if ctor is about to.
The compiler usually spots, so no less efficient and makes it clearer/harder to accidentally introduce uninitialized members

jakesmith (commented):
@kenrowland - some minor comments, but agree with Gavin's comment re. leak - see my follow comments.

Also please squash and rebase/retarget 8.2.x

testing/unittests/metrics/PeriodicSinkTests.cpp
kenrowland: Changed to unsigned

testing/unittests/metrics/PeriodicSinkTests.cpp
kenrowland: Removed

testing/unittests/metrics/PeriodicSinkTests.cpp
kenrowland: Changed all related values to unsigned.

testing/unittests/metrics/PeriodicSinkTests.cpp
kenrowland: Added init of pPeriodicTestSink to nullptr.

kenrowland (commented):
@jakesmith Addressed your comments. Please take a look again. I will rebase, probably against candidate-8.2.x, once you are OK with it.

system/jlib/jmetrics.cpp
jakesmith: initializeSinks - agreed - 'sinks' find() match should guarentee the insert succeeds

testing/unittests/metrics/MetricFrameworkTests.cpp
jakesmith: @kenrowland - missed this comment?

jakesmith (commented):
@kenrowland - looks good afaics, please squash and rebase for final review

kenrowland:
Closing in favor of new PR since can't easily rebase to candidate-8.2.x

PR #15092, opened 2021-06-22
HPCC-26092 Tighten up default plane validation

richardkchapman:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [x] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/helm/hpcc/templates/_helpers.tpl
+++ b/helm/hpcc/templates/_helpers.tpl
@@ -478,21 +478,27 @@ vaults:
 {{- end -}}
 
 {{/*
-Return a value indicating whether a storage plane is defined or not.
+Check whether a storage plane is defined or not.
 */}}
-{{- define "hpcc.isValidStoragePlane" -}}
+{{- define "hpcc.checkValidStoragePlane" -}}
 {{- $search := .search -}}
+{{- $category := .category -}}
 {{- $storage := (.root.Values.storage | default dict) -}}
 {{- $planes := ($storage.planes | default list) -}}
 {{- $dataStorage := ($storage.dataStorage | default dict) -}}
 {{- /* If storage.dataStorage.plane is defined, the implicit plane hpcc-dataplane is not defined */ -}}
-{{- $done := dict "matched" (and (not $dataStorage.plane) (eq $search "hpcc-dataplane")) -}}
+{{- $done := dict "matched" (and (not $dataStorage.plane) (eq $search "hpcc-dataplane")) "all" "" -}}
 {{- range $plane := $planes -}}
- {{- if eq $search $plane.name -}}
- {{- $_ := set $done "matched" true -}}
+ {{- if eq $category $plane.category -}}
+  {{- if eq $search $plane.name -}}
+   {{- $_ := set $done "matched" true -}}
+  {{- end -}}
+  {{- $_ := set $done "all" ( printf "%s \"%s\"" $done.all $plane.name) -}}
  {{- end -}}
 {{- end -}}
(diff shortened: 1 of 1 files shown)


dockerfiles/startall.sh
```diff
@@ -95,10 +105,11 @@ if [[ -n ${PERSIST} ]] ; then
   mkdir -p ${PERSIST_PATH}/hpcc-data
```
ghalliday, line 105: minor: could also execute this if -n ${PERSISTVALUES}

helm/hpcc/templates/_helpers.tpl
```diff
@@ -514,15 +520,11 @@ Check that the storage and spill planes for a component exist
 {{- define "hpcc.checkDefaultStoragePlane" -}}
 {{- if (hasKey .me "storagePlane") }}
  {{- $search := .me.storagePlane -}}
- {{- if ne (include "hpcc.isValidStoragePlane" (dict "search" $search "root" .root)) "true" -}}
-  {{- $_ := fail (printf "storage data plane %s for %s is not defined" $search .me.name ) }}
- {{- end -}}
+ {{- include "hpcc.isValidStoragePlane" (dict "search" $search "root" .root "category" "data" "type" "storage data" "for" .me.name) -}}
```
ghalliday, line 523: minor, better to rename isValid to checkValid

ghalliday: Alternative implementation would be to have a function that returns a list of all the storage planes, and include it in the fail message.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -491,21 +491,27 @@ vaults:
...
 {{- $storage := (.root.Values.storage | default dict) -}}
 {{- $planes := ($storage.planes | default list) -}}
 {{- $dataStorage := ($storage.dataStorage | default dict) -}}
 {{- /* If storage.dataStorage.plane is defined, the implicit plane hpcc-dataplane is not defined */ -}}
-{{- $done := dict "matched" (and (not $dataStorage.plane) (eq $search "hpcc-dataplane")) -}}
+{{- $done := dict "matched" (and (not $dataStorage.plane) (eq $search "hpcc-dataplane")) "all" "" -}}
 {{- range $plane := $planes -}}
- {{- if eq $search $plane.name -}}
- {{- $_ := set $done "matched" true -}}
+ {{- if has $category $plane.labels -}}
```
ghalliday, line 505: better to reverse the tests, and report an error if the category isn't in the labels.  Also need to be careful at the moment that a blank labels list is treated as equivalent to [data]

ghalliday (commented):
@richardkchapman some minor comments

helm/hpcc/templates/_helpers.tpl
richardkchapman: Except I'm scanning through all planes with this category in order to give the candidate list in the error message

dockerfiles/startall.sh
richardkchapman: Should really be creating directories that are named in the values file in that case

richardkchapman:
@ghalliday I don't know if this is still useful given other changes

ghalliday:
I think it probably is.  There probably need to be more tests like this, but this is a useful start.

ghalliday (approved):

PR #15097, opened 2021-06-22
HPCC-26100 Fix SOAPCALL chunked encoding reads with persistent connection

afishbeck:
…tions

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/common/thorhelper/thorsoapcall.cpp
+++ b/common/thorhelper/thorsoapcall.cpp
@@ -1858,7 +1858,7 @@ private:
                                 chunkSize = (chunkSize*16) + 10 + (toupper(ch) - 'A');
                             dataProvider->getBytes(&ch, 1);
                         }
-                        while (chunkSize && ch != '\n')//consume chunk-extension and CRLF
+                        while (ch != '\n')//consume chunk-extension and CRLF
                             dataProvider->getBytes(&ch, 1);
                         while (chunkSize)
                         {
@@ -1885,9 +1885,20 @@ private:
                                     chunkSize = (chunkSize*16) + 10 + (toupper(ch) - 'A');
                                 dataProvider->getBytes(&ch, 1);
                             }
-                            while(chunkSize && ch != '\n')//consume chunk-extension and CRLF
+                            while(ch != '\n')//consume chunk-extension and CRLF
                                 dataProvider->getBytes(&ch, 1);
                         }
+                        //to support persistent connections we need to consume all trailing bytes for each message
+                        dataProvider->getBytes(&ch, 1);
+                        //the standard allows trailing HTTP headers to be included at the of the chunked message, the following loop will consume those
+                        while ( ch != '\r' ) //if there is content before the CRLF it is a trailing HTTP header
+                        {
+                            while (ch != '\n')//consume the CRLF of the trailing header
+                                dataProvider->getBytes(&ch, 1);
+                            dataProvider->getBytes(&ch, 1);
+                        }
+                        while (ch != '\n') //consume final CRLF
+                            dataProvider->getBytes(&ch, 1);
                     }
                     break;
                 }


afishbeck:
@stuartort Can someone help test this in the real world?
Perhaps put it somewhere for Vlad to test.  Like upgrading a dev 1 node roxie or eclagent where Vlad can 1st reproduce his problem and then have the problem go away after this fix?  Just make sure the maxPersistentRequests is not configured and set to 0 which would bypass all of this.

common/thorhelper/thorsoapcall.cpp
```diff
@@ -1897,6 +1897,15 @@ class CWSCAsyncFor : implements IWSCAsyncFor, public CInterface, public CAsyncFo
                             while(chunkSize && ch != '\n')//consume chunk-extension and CRLF
                                 dataProvider->getBytes(&ch, 1);
                         }
+                        //to support persistent connections we need to consume all trailing bytes for each message
+                        while ( ch != '\r' ) //consume trailers
```
ghalliday, line 1901: Is this the correct behaviour with the example:

```
7\r\n
chunks.\r\n  (data)
0\r\n        (final byte - 0)
\r\n         (end message)
```
Lines up to 1870 will read up to the 0, but line 1870/1871 will not skip the \r.  This code will then match and consume the \r\n after  the 0 but the end of message \r\n will not be consumed.

ghalliday (commented):
@afishbeck one question

ghalliday:
Also, I don't think master is the correct target.  8,.2.0 (or earlier?)

common/thorhelper/thorsoapcall.cpp
afishbeck: I suspect I may have to consume one more crlf.  I'll take a look.

ghalliday: An alternative is to remove the chunkSize test from line 1870.  I may be completely wrong - I only did a quick web search since I knew nothing about chunked encoding.

afishbeck:
@ghalliday I should now properly consume the CRLF of the 0 chunkSize as well as the rest.  I also better tested that I have consumed all the bytes at least for my test.

I've rebased to 8.2.0 for now but can target even earlier (the issue can arise in 7.12.x) is we want.

Still want to try get someone to do a real world test if possible.  But please review.

richardkchapman:
@afishbeck I think we should target 7.12.x - but need to be confident that this has been tested before merging. What's the status/plan on that?

afishbeck:
@richardkchapman retargeted 7.12.x.  I'm coordinating with Chris, Ops, and Vlad to test.

common/thorhelper/thorsoapcall.cpp
ghalliday: I think the change will fix the common case.  It does not cope with the case where there is no output at all - so the total message is
```
0\r\n
\r\n
```
The spec also refers to needing to be able to skip trailing header fields

See https://datatracker.ietf.org/doc/html/rfc7230#section-4.1

The final termination looks like it is a single \r\n terminated line

afishbeck: @ghalliday This loop is meant to consume those trailing headers (I was calling them "trailers"):
```
while ( ch != '\r' ) //consume trailers
{
  while (ch != '\n')
     dataProvider->getBytes(&ch, 1);
  dataProvider->getBytes(&ch, 1);
}
```
Basically if there are lines with content before the final close, consume those first.  Do you see any problems with it?  Wasn't able to really test.. and doubt it's used much.

ghalliday: Yes you're right.  Maybe a comment before the while (indicating what it is skippoing), and another one before the next while
I think the chunkSize test in 1870 may also need to go to avoid problems with a completely zero length result.

afishbeck: Ok I'll update.  But I'll let Vlad test as is as that won't test the zero length scenario anyway.

afishbeck: Updated with fix for completely zero length result and added comments about consuming the trailing HTTP headers.

Test build should be deployed tomorrow morning and Vlad will test.

afishbeck:
@richardkchapman @ghalliday Vlad's test succeeded this morning, and the logging verified that it was using the new code.

ghalliday (approved):

ghalliday:
@richardkchapman this looks good to merge

PR #15102, opened 2021-06-24
HPCC-26088 Clean up uses of getConfigurationDirectory

ghalliday:
## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jutil.cpp
+++ b/system/jlib/jutil.cpp
@@ -2624,12 +2624,9 @@ jlib_decl bool queryMtlsBareMetalConfig()
 }
 #endif
 
+#ifndef _CONTAINERIZED
 static IPropertyTree *getOSSdirTree()
 {
-#ifdef _CONTAINERIZED
-    IERRLOG("getOSSdirTree() called from container system");
-    return nullptr;
-#endif
     Owned<IPropertyTree> envtree = getHPCCEnvironment();
     if (envtree) {
         IPropertyTree *ret = envtree->queryPropTree("Software/Directories");
@@ -2638,6 +2635,8 @@ static IPropertyTree *getOSSdirTree()
     }
     return NULL;
 }
+#endif
+
 
 StringBuffer &getFileAccessUrl(StringBuffer &out)
 {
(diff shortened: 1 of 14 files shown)


ghalliday:
@jakesmith @richardkchapman for initial discussion.

system/jlib/jutil.cpp
```diff
@@ -2663,8 +2662,79 @@ StringBuffer &getFileAccessUrl(StringBuffer &out)
...
+
+    Owned<IPropertyTree> storagePlane = getStoragePlane(planeName);
+    return storagePlane->getProp("@prefix", ret);
+}
+#endif
+
 bool getConfigurationDirectory(const IPropertyTree *useTree, const char *category, const char *component, const char *instance, StringBuffer &dirout)
 {
+#ifdef _CONTAINERIZED
+    if (streq(category, "data"))
```
richardkchapman, line 2694: Perhaps assert that useTree is null?

ghalliday: It is non null in too many places at the moment - possibly once all the calls have been cleaned up.

system/jlib/jutil.cpp
```diff
@@ -2663,8 +2662,79 @@ StringBuffer &getFileAccessUrl(StringBuffer &out)
...
+    {
+        return getDefaultPlaneDirectory(dirout, "@daliPlane", "storage/@daliPlane");
+    }
+    if (streq(category, "query"))
+    {
+        return getDefaultPlaneDirectory(dirout, "@dllPlane", "storage/@dllPlane");
+    }
+    if (streq(category, "lock"))
+    {
+        //Should this be on the dali plane?  Does this even make any sense?
```
jakesmith, line 2727: I don't think it does make sense in cloud.
It is only used in Dali to prevent to dali's in same process space, that will never happen..
That use (NamedMutex) is already in a #ifndef _CONTAINERIZED block

jakesmith (approved):
@ghalliday looks good, 1 minor comment

system/jlib/jutil.cpp
```diff
@@ -2663,8 +2662,79 @@ StringBuffer &getFileAccessUrl(StringBuffer &out)
...
+    return storagePlane->getProp("@prefix", ret);
+}
+#endif
+
 bool getConfigurationDirectory(const IPropertyTree *useTree, const char *category, const char *component, const char *instance, StringBuffer &dirout)
 {
+#ifdef _CONTAINERIZED
+    if (streq(category, "data"))
+    {
+        Owned<IPropertyTree> storagePlane = getStoragePlane(instance);
```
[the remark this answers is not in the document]
jakesmith: There's code in Thor in thmastermain.cpp (820-831) which before this PR would always return null, and not call setBaseDirectory as consequence - so that code did nothing.
With this change it might return something if a plane matches the Thor name.
I think that code should be #ifdef 'd out in container.

Thor default to the default plane, if there are no CLUSTER/PLANE directives on an output.

richardkchapman (approved):

PR #15104, opened 2021-06-24
HPCC-26111 CMake smoketest build issue with CC/CXX

mckellyln:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

## Testing:

smoketest

### DIFF
--- a/.github/workflows/smoketest.yml
+++ b/.github/workflows/smoketest.yml
@@ -272,6 +272,8 @@ jobs:
       if: ${{ steps.check-cache.outputs.state != '' }} # if 'state' set to anything, then step needed
       run: |
         cd build
+        echo "removing CMakeCache.txt"
+        rm -f CMakeCache.txt
         if [[ ! -f cache-timestamp ]]; then touch cache-timestamp; fi
         touch ../cmake-timestamp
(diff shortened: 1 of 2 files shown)


mckellyln:
Removing the CMakeCache.txt appears to, so far, resolve this issue, at least with many PR updates.
Since the cmake command specifies many/most/all required options, deleting the CMakeCache.txt beforehand should not alter the created Makefiles.

jakesmith (commented):
> Removing the CMakeCache.txt appears to, so far, resolve this issue, at least with many PR updates.
> Since the cmake command specifies many/most/all required options, deleting the CMakeCache.txt beforehand should not alter the created Makefiles.

it's hitting this problem (sometimes) because of a rerun from a previously build, because it's restored from a cache.

Deleting CMakeCache.txt will invalidate the build cache and cause a full rebuild I believe, that would defeat the point of the rerun/caching.

richardkchapman:
I don't think removing CMakeCache.txt means that cpp recompiles happen (I just tried it). It does mean that CMake has to be run, but that's quick.

jakesmith:
> I don't think removing CMakeCache.txt means that cpp recompiles happen (I just tried it). It does mean that CMake has to be run, but that's quick.

ok, maybe it's fine then. I thought it touched something in the process, perhaps that's before the build variables were moved from a changing header that was included everywhere, into 1 file (jutil.cpp) in jlib and picked up programmatically.
Does jutil.cpp recompile when CMakeCache.txt is removed/rebuilt?

mckellyln:
cmake is already always run specifying needed values/arguments.
I do not think it causes any additional compiles or linking, but I will double check.
I think this is an issue within cmake, and to debug it fully would, I believe, require adding a debug cmake with much verbosity.  This PR is a quick attempt at a work-around.

jakesmith:
> cmake is already always run specifying needed values/arguments.
> I do not think it causes any additional compiles or linking, but I will double check.
> I think this is an issue within cmake, and to debug it fully would, I believe, require adding a debug cmake with much verbosity. This PR is a quick attempt at a work-around.

as long as it doesn't cause a lot rebuild on a cache hit should be fine.

richardkchapman (approved):

mckellyln:
Just did an amend with a title change to verify what gets rebuilt

jakesmith (approved):

PR #15109, opened 2021-06-25
HPCC-26005 Create numDiskReads statistic for logical files (thor flat)

shamser:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/thorlcr/activities/diskread/thdiskreadslave.cpp
+++ b/thorlcr/activities/diskread/thdiskreadslave.cpp
@@ -411,6 +411,7 @@ void CDiskRecordPartHandler::close(CRC32 &fileCRC)
     if (partStream)
     {
         mergeStats(fileStats, partStream);
+        activity.mergeSubFileStats(partDesc, partStream);
         partStream->stop(&fileCRC);
     }
 }
--- a/thorlcr/activities/thdiskbase.cpp
+++ b/thorlcr/activities/thdiskbase.cpp
@@ -68,19 +68,20 @@ void CDiskReadMasterBase::init()
             mapping.setown(getFileSlaveMaps(file->queryLogicalName(), *fileDesc, container.queryJob().queryUserDescriptor(), container.queryJob().querySlaveGroup(), local, false, hash, file->querySuperFile()));
             addReadFile(file, temp);
         }
+        IDistributedSuperFile *super = file->querySuperFile();
+        unsigned numsubs = super?super->numSubFiles(true):0;
         if (0 != (helper->getFlags() & TDRfilenamecallback)) // only get/serialize if using virtual file name fields
         {
-            IDistributedSuperFile *super = file->querySuperFile();
-            if (super)
+            for (unsigned s=0; s<numsubs; s++)
             {
-                unsigned numsubs = super->numSubFiles(true);
-                unsigned s=0;
-                for (; s<numsubs; s++)
-                {
-                    IDistributedFile &subfile = super->querySubFile(s, true);
-                    subfileLogicalFilenames.append(subfile.queryLogicalName());
-                }
+                IDistributedFile &subfile = super->querySubFile(s, true);
+                subfileLogicalFilenames.append(subfile.queryLogicalName());
             }
+        }
+        if (0==(helper->getFlags() & TDXtemporary))
+        {
+            for (unsigned i=0; i<numsubs; i++)
+                subFileStats.push_back(new CThorStatsCollection(diskReadRemoteStatistics));
         }
         void *ekey;
         size32_t ekeylen;
@@ -116,6 +117,34 @@ void CDiskReadMasterBase::serializeSlaveData(MemoryBuffer &dst, unsigned slave)
         CSlavePartMapping::serializeNullMap(dst);
 }
 
+void CDiskReadMasterBase::done()
+{
+    if (!subFileStats.empty())
+    {
+        unsigned numSubFiles = subFileStats.size();
+        for (unsigned i=0; i<numSubFiles; i++)
+        {
+            IDistributedFile *file = queryReadFile(i);
+            if (file)
+                file->addAttrValue("@numDiskReads", subFileStats[i]->getStatisticSum(StNumDiskReads));
+        }
+    }
+    else
+    {
+        IDistributedFile *file = queryReadFile(0);
+        if (file)
+            file->addAttrValue("@numDiskReads", statsCollection.getStatisticSum(StNumDiskReads));
+    }
+    CMasterActivity::done();
+}
+
+void CDiskReadMasterBase::deserializeStats(unsigned node, MemoryBuffer &mb)
+{
+    CMasterActivity::deserializeStats(node, mb);
+
+    for (auto &stats: subFileStats)
+        stats->deserialize(node, mb);
+}
 /////////////////
 
 void CWriteMasterBase::init()
--- a/thorlcr/activities/thdiskbase.ipp
+++ b/thorlcr/activities/thdiskbase.ipp
@@ -32,12 +32,15 @@ protected:
     Owned<CSlavePartMapping> mapping;
     IHash *hash;
     StringAttr fileName;
+    std::vector<OwnedPtr<CThorStatsCollection>> subFileStats;
 
 public:
     CDiskReadMasterBase(CMasterGraphElement *info);
-    virtual void init();
-    virtual void serializeSlaveData(MemoryBuffer &dst, unsigned slave);
+    virtual void init() override;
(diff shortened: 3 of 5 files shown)


richardkchapman:
Please retarget to 8.2.x

thorlcr/activities/thdiskbase.cpp
```diff
@@ -116,6 +114,36 @@ void CDiskReadMasterBase::serializeSlaveData(MemoryBuffer &dst, unsigned slave)
...
 }
 
+void CDiskReadMasterBase::done()
+{
+    if (!subFileStats.empty())
+    {
+        unsigned numSubFiles = subFileStats.size();
+        for (unsigned i=0; i<numSubFiles; i++)
+        {
+            IDistributedFile *file = queryReadFile(i);
```
jakesmith, line 124: has this been tested?
I don't think this works, afair, there will be only 1 logical file for the super, not 1 for every sub of the super.

thorlcr/activities/thdiskbase.ipp
```diff
@@ -32,12 +32,15 @@ protected:
     Owned<CSlavePartMapping> mapping;
     IHash *hash;
     StringAttr fileName;
+    std::vector<OwnedPtr<CThorStatsCollection>> subFileStats;
 
 public:
     CDiskReadMasterBase(CMasterGraphElement *info);
     virtual void init();
     virtual void serializeSlaveData(MemoryBuffer &dst, unsigned slave);
+    virtual void done() override;
```
jakesmith, line 41: trivial: good change, for consistency would be good if other 'override' were added at same time.

thorlcr/activities/thdiskbaseslave.cpp
```diff
@@ -300,6 +324,10 @@ void CDiskReadSlaveActivityBase::serializeStats(MemoryBuffer &mb)
     }
     stats.setStatistic(StNumDiskRowsRead, diskProgress);
     PARENT::serializeStats(mb);
+    unsigned numSubFiles = subFileStats.size();
+    mb.append(numSubFiles);
```
jakesmith, line 328: master should know this already, so could avoid serializing?

thorlcr/activities/diskread/thdiskreadslave.cpp
```diff
@@ -411,6 +411,7 @@ void CDiskRecordPartHandler::close(CRC32 &fileCRC)
     if (partStream)
     {
         mergeStats(fileStats, partStream);
+        activity.mergeSubFileStats(partDesc, partStream);
```
jakesmith, line 414: this means I think, that we have full disk read stats for the whole super (from prev. line), plus stats for each sub.

That's ok/good, but am wondering if there are supers with dozens of subs, is it getting to expensive to serialize detailed stats on every sub, especially as at the moment as only the StNumDiskReads is going to be used.

@ghalliday - what do you think?

Or if we are going to use them, should the other stats in the collection be preserved per sub, and published to the activity too?

thorlcr/activities/thdiskbase.cpp
```diff
@@ -68,20 +68,18 @@ void CDiskReadMasterBase::init()
...
-                for (; s<numsubs; s++)
-                {
-                    IDistributedFile &subfile = super->querySubFile(s, true);
-                    subfileLogicalFilenames.append(subfile.queryLogicalName());
-                }
+                IDistributedFile &subfile = super->querySubFile(s, true);
+                subfileLogicalFilenames.append(subfile.queryLogicalName());
             }
         }
+        for (unsigned i=0; i<numsubs; i++)
```
jakesmith, line 81: So the number is in sync. with the slaves (and avoid need to serialize count etc.),
should avoid creating the objects if a temporary file (as you do in slaves).

jakesmith (commented):

thorlcr/activities/thdiskbase.cpp
shamser: I have tested it by reading a superfile with 2 subfiles.  The subfiles numDiskReads attributes for each file was updated- each files numDiskReads value reflected its size.  Is there specific circumstance that this should not work that I should test?

jakesmith: ah, I see now, I had forgotten how this worked. addSubFile is called once so 1 entry for the super,
but in CMasterActivity::preStart it expands the readFiles array if it contains super files with the subfiles.

jakesmith:
@shamser - the JIRA/PR title suggests this is for all logical files, but it's for flat files only at the moment. i.e. not csv, xml, indexes, files read by fetch, or kj.
The rest to be tackled in future JIRA's, but it would be good to change the JIRA/PR title to reflect this is just for Thor flat files, and open subtasks for others if they don't exist.

thorlcr/activities/thdiskbase.cpp
```diff
@@ -116,6 +117,34 @@ void CDiskReadMasterBase::serializeSlaveData(MemoryBuffer &dst, unsigned slave)
...
+            file->addAttrValue("@numDiskReads", statsCollection.getStatisticSum(StNumDiskReads));
+    }
+    CMasterActivity::done();
+}
+
+void CDiskReadMasterBase::deserializeStats(unsigned node, MemoryBuffer &mb)
+{
+    CMasterActivity::deserializeStats(node, mb);
+
+   for (auto &stats: subFileStats)
```
jakesmith, line 145: trivial/formatting: indent off by 1 space.

thorlcr/activities/thdiskbaseslave.cpp
```diff
@@ -235,13 +234,36 @@ void CDiskReadSlaveActivityBase::init(MemoryBuffer &data, MemoryBuffer &slaveDat
...
+        }
     }
     gotMeta = false; // if variable filename and inside loop, need to invalidate cached meta
 }
-
+void CDiskReadSlaveActivityBase::mergeSubFileStats(IPartDescriptor *partDesc, IExtRowStream *partStream)
+{
+    if (subFileStats.size()>0)
+    {
+        ISuperFileDescriptor * superFDesc =  partDesc->queryOwner().querySuperFileDescriptor();
```
jakesmith, line 260: trivial/formatting: extra space before partDesc

thorlcr/activities/thdiskbaseslave.cpp
```diff
@@ -235,13 +234,36 @@ void CDiskReadSlaveActivityBase::init(MemoryBuffer &data, MemoryBuffer &slaveDat
...
+                partDescs.item(0).queryOwner().setDefaultDir(queryTempDir());
+        }
+        else
+        {
+            ISuperFileDescriptor *super = partDescs.item(0).queryOwner().querySuperFileDescriptor();
+            if (super)
+            {
+                unsigned numSubFiles = super->querySubFiles();
+                for (unsigned i=0; i<numSubFiles; i++)
+                    subFileStats.push_back(new CRuntimeStatisticCollection(diskReadRemoteStatistics));
```
jakesmith, line 250: Each worker may only be handling parts from some sub files, e.g. a super with 2 subs of 2 parts, being read on a 4-way. Meaning this would serialize metrics for subs that aren't being read by this slave.

But not a big issue and don't think should change, just an observation.

jakesmith (commented):
@shamser - looks good to me, 2 trivial formatting comments. Please squash.

@Gavin - could you comment re. https://github.com/hpcc-systems/HPCC-Platform/pull/15109/files#r661634999 ?

jakesmith (approved):
@shamser - looks good to me.
But would like Gavin's thoughts on : https://github.com/hpcc-systems/HPCC-Platform/pull/15109#discussion_r661634999 before merging.

thorlcr/activities/diskread/thdiskreadslave.cpp
ghalliday: I've checked and I don't think the overhead will be that large.  It only serializes stats that have been filled in, and also packs the values so they are likely to only take a few bytes each.
I don't think there is an advantage to publishing the individual file stats to the activity.  Possibly to the workunit summary, but even then I doubt it.

thorlcr/activities/thdiskbaseslave.cpp
ghalliday: If so I think that will end up serializing a single 0 byte, so trivial overhead.

ghalliday (approved):

PR #15135, opened 2021-07-04
HPCC-26156 Enable parallel build for AWS SDK

GordonSmith:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/aws/CMakeLists.txt
+++ b/system/aws/CMakeLists.txt
@@ -8,13 +8,19 @@ if(USE_AWS)
         set(EXTERNAL_SQS_CXX_FLAGS "-Wno-error=implicit-fallthrough -Wno-error=int-in-bool-context -Wno-error=format-nonliteral -Wno-error=maybe-uninitialized" )
     endif()
 
+    include(ProcessorCount)
+    ProcessorCount(N)
+    if(NOT N EQUAL 0 AND CMAKE_MAKE_PROGRAM STREQUAL "/usr/bin/make")
+        set(AWS_BUILD_FLAGS -j${N})
+    endif()
+
     include(ExternalProject)
     ExternalProject_Add(
         aws-sdk-cpp
         SOURCE_DIR ${AWS_SDK_CPP_SOURCE_DIR}
         BINARY_DIR ${CMAKE_CURRENT_BINARY_DIR}/aws-sdk-cpp
         CMAKE_ARGS -DCUSTOM_MEMORY_MANAGEMENT:string=0 -DCMAKE_CXX_FLAGS=${EXTERNAL_SQS_CXX_FLAGS} -DCMAKE_SHARED_LINKER_FLAGS_INIT=-Wl,-rpath-link,${LIB_PATH}
-        BUILD_COMMAND ${CMAKE_MAKE_PROGRAM} aws-cpp-sdk-sqs aws-cpp-sdk-s3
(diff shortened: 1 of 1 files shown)


system/aws/CMakeLists.txt
```diff
@@ -8,13 +8,19 @@ if(USE_AWS)
...
+    endif()
+
     include(ExternalProject)
     ExternalProject_Add(
         aws-sdk-cpp
         SOURCE_DIR ${AWS_SDK_CPP_SOURCE_DIR}
         BINARY_DIR ${CMAKE_CURRENT_BINARY_DIR}/aws-sdk-cpp
         CMAKE_ARGS -DCUSTOM_MEMORY_MANAGEMENT:string=0 -DCMAKE_CXX_FLAGS=${EXTERNAL_SQS_CXX_FLAGS} -DCMAKE_SHARED_LINKER_FLAGS_INIT=-Wl,-rpath-link,${LIB_PATH}
-        BUILD_COMMAND ${CMAKE_MAKE_PROGRAM} aws-cpp-sdk-sqs aws-cpp-sdk-s3
+        BUILD_COMMAND ${CMAKE_MAKE_PROGRAM} ${AWS_BUILD_FLAGS} aws-cpp-sdk-sqs aws-cpp-sdk-s3
```
ghalliday, line 23: @GordonSmith 
This change looks reasonable, and probably worth including.  However, wouldn't it be better to use
${CMAKE_COMMAND} --build
which I think the aws libraries support.  Then it would cleanly support ninja and other build systems.

GordonSmith: That would still only build (by default) with -- -j1?

GordonSmith: @ghalliday this variable will already be set to Ninja (and other build systems):  https://cmake.org/cmake/help/latest/variable/CMAKE_MAKE_PROGRAM.html

ghalliday: Yes, but there seems to be a difference in the way build commands are handled if it invokes the make program directly, or uses cmake.  I had one instance where one way it executed in parallel, but the other it did not.  I'll try and test.

ghalliday: I just tested this, and it seemed to make no difference to the compile time (with lots of cores).

ghalliday: Adding the option to enable unity builds did change it (it slowed it down for me).

ghalliday: @GordonSmith testing this again.
If you use make -j with no number then it doesn't speed it up at all, but if you use make -j <n> then it does help.

It looks like cloning the repositories or whatever it is doing actually takes a large chunk of the time once parallel building is enabled.

Building with ninja saved a further 20% - even with the cloning in place.

ghalliday (approved):

ghalliday:
@richardkchapman please check you think this is appropriate to merge into 8.6.x.
We should revisit how AWS is integrated.  I suspect we should disable it by default in the builds - I think it is only used by the s3file plugin, and that is more a proof of concept than production ready.

GordonSmith:
@ghalliday if it is disabled by default, then this PR becomes mostly moot.

richardkchapman (approved):

PR #15153, opened 2021-07-08
HPCC-26168 Fix issues using multiple dll storage planes

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/helm/hpcc/templates/_helpers.tpl
+++ b/helm/hpcc/templates/_helpers.tpl
@@ -194,22 +194,21 @@ storage:
   dataPlane: {{ include "hpcc.getDefaultDataPlane" . }}
   planes:
 {{- /*Generate entries for each data plane (removing the pvc).  Exclude the planes used for dlls and dali.*/ -}}
-{{- range $plane := $planes -}}
- {{- if or (eq "data" $plane.category) (eq "lz" $plane.category) }}
+{{- range $plane := $planes }}
   - name: {{ $plane.name | quote }}
-  {{- $planeYaml := omit $plane "name" "pvc" "storageClass" "storageSize" "subPath" -}}
-  {{- if $plane.subPath -}}
-   {{- $_ := set $planeYaml "prefix" (printf "%s/%s" $planeYaml.prefix $plane.subPath) -}}
-  {{- end -}}
-  {{- if and (eq "data" $plane.category) (not $plane.defaultSprayParts) -}}
-   {{- $_ := set $planeYaml "defaultSprayParts" (include "hpcc.getMaxNumWorkers" $ | int) -}}
-  {{- end -}}
-  {{- toYaml $planeYaml | nindent 4 }}
- {{- end }}
+ {{- $planeYaml := omit $plane "name" "pvc" "storageClass" "storageSize" "subPath" -}}
+ {{- if $plane.subPath -}}
+  {{- $_ := set $planeYaml "prefix" (printf "%s/%s" $planeYaml.prefix $plane.subPath) -}}
+ {{- end -}}
+ {{- if and (eq "data" $plane.category) (not $plane.defaultSprayParts) -}}
+  {{- $_ := set $planeYaml "defaultSprayParts" (include "hpcc.getMaxNumWorkers" $ | int) -}}
+ {{- end -}}
+ {{- toYaml $planeYaml | nindent 4 }}
 {{- end }}
 {{- if not (include "hpcc.hasPlaneForCategory" (dict "root" $ "category" "spill")) }}
   - name: hpcc-spill-plane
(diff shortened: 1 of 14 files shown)


roxie/ccd/ccdfile.cpp
```diff
@@ -909,9 +909,20 @@ class CRoxieFileCache : implements IRoxieFileCache, implements ICopyFileProgress
...
-                assertex(dllserver_root != nullptr);
+                StringBuffer cacheRootDirectory;
+                if (isContainerized())
+                {
+                    if (!getConfigurationDirectory(nullptr, "query", nullptr, nullptr, cacheRootDirectory))
+                        throwUnexpected();
+                }
+                else
+                {
+                    const char* dllserver_root = getenv("HPCC_DLLSERVER_PATH");
```
richardkchapman, line 922: Feels like this should be retrieved via environment directories section...

roxie/ccd/ccdfile.cpp
```diff
@@ -1852,16 +1863,23 @@ class CRoxieFileCache : implements IRoxieFileCache, implements ICopyFileProgress
...
-#endif
+        StringBuffer cacheRootDirectory;
+        if (isContainerized())
+        {
+            if (!getConfigurationDirectory(nullptr, "query", nullptr, nullptr, cacheRootDirectory))
+                throwUnexpected();
+        }
+        else
+        {
+            //Default behaviour is to not load or saving anything on bare metal
```
richardkchapman, line 1874: Either this comment is wrong, or the previous code that asserted the env var was set is wrong.

roxie/ccd/ccdmain.cpp
```diff
@@ -723,7 +723,7 @@ int CCD_API roxie_main(int argc, const char *argv[], const char * defaultYaml)
 #ifdef _CONTAINERIZED
         getDefaultStoragePlane(defaultPlane);
         {
-            Owned<IStoragePlane> plane = getStoragePlane(defaultPlane, true);
+            Owned<IStoragePlane> plane = getDataStoragePlane(defaultPlane, true);
```
richardkchapman, line 726: Not new, but the fact that this throws an exception when not set is a bit of a pain for standalone executable usage. Probably only affects me though.

richardkchapman (changes requested):
Not sure the issues highlighted are new, but may be worth addressing as part of this change

roxie/ccd/ccdfile.cpp
ghalliday: The assert was if containerized was true - this is if it is false.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -155,19 +155,18 @@ storage:
...
-  {{- if $plane.subPath -}}
-   {{- $_ := set $planeYaml "prefix" (printf "%s/%s" $planeYaml.prefix $plane.subPath) -}}
-  {{- end -}}
-    {{- toYaml $planeYaml | nindent 4 }}
- {{- end }}
+ {{- $planeYaml := omit $plane "name" "pvc" "storageClass" "storageSize" "subPath" -}}
+ {{- if $plane.subPath -}}
+  {{- $_ := set $planeYaml "prefix" (printf "%s/%s" $planeYaml.prefix $plane.subPath) -}}
+ {{- end -}}
+   {{- toYaml $planeYaml | nindent 4 }}
```
jakesmith, line 164: trivial/formatting: indentation looks off here?

jakesmith (commented):
@ghalliday - one trivial/formatting issue.

But I think this PR should now also remove the HPCC_DLLSERVER_PATH's from the dockerfiles/platform-core*/Dockerfile files

ghalliday:
rebased and env entries removed from the docker files.
Should possibly target 8.4 at this point, but I wonder if some of the changes might be needed to allow temp files to go to NVMe.

roxie/ccd/ccdfile.cpp
richardkchapman: I think you have left this code in so that it behaves the same as previously (i.e. only saves the cache info in bare-metal if environment var is set up (which it will not normally be). However, since that feature in bare metal is undocumented, untested, and unused (It may cause Roxie to randomly core priot to 8.2.x so I'm pretty sure no-one is relying on it) it would make sense to change the env var name to something more meaningful at this point?

roxie/ccd/ccdfile.cpp
richardkchapman: Same comment re env var name as before

richardkchapman (approved):

PR #15173, opened 2021-07-14
HPCC-26202 Certificate and MTLS related fixes

jakesmith:
+ Fix a couple of regressions that stem from helm "service" changes,
where the certificate calls need to be passed the service to
use for its name.
+ Missing certificate generation in dfuserver and sasha components
+ Add a mtls option to disabled mtls (default true), so can
enable certificates, without enabling mtls.
+ Remove global cached ISecureSocketContext server context, which
doesn't make sense if the component is creating multiple contexts.
e.g. in case of esp it creates a context with 1 config for MP,
and another context for public secure esp connections.
The single module global cache instance was not useful anyway,
the creators typically create once or cache themselves.
+ Provide means for a specific issuer to be used with
the addCertificate* calls (this will be used for a signing issuer)

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/helm/hpcc/templates/esp.yaml
+++ b/helm/hpcc/templates/esp.yaml
@@ -74,8 +74,13 @@ data:
 {{- $secretsCategories := ternary (list "storage" "esp" "codeSign" "codeVerify")  (list "storage" "esp") (eq $application "eclwatch") -}}
 {{- $includeStorageCategories := ternary (list "lz" "data") (list "data") (eq $application "eclwatch") -}}
 {{- $commonCtx := dict "root" $ "me" . "secretsCategories" $secretsCategories  "includeCategories" $includeStorageCategories "env" $env -}}
-{{- $configSHA := include "hpcc.getConfigSHA" ($commonCtx | merge (dict "configMapHelper" "hpcc.espConfigMap" "component" "esp" "excludeKeys" "global,esp.queues")) }}
-
+{{- $configSHA := include "hpcc.getConfigSHA" ($commonCtx | merge (dict "configMapHelper" "hpcc.espConfigMap" "component" "esp" "excludeKeys" "global,esp.queues")) -}}
+{{- if (ne (include "hpcc.isVisibilityPublic" (dict "root" $ "visibility" .service.visibility)) "") }}
+ {{- /* If esp is using a public cert, this flag is used later to add the local MTLS mount and cert */ -}}
+ {{- $_ := set $commonCtx "externalCert" true -}}
+{{- else -}}
+ {{- $_ := set $commonCtx "externalCert" false -}}
+{{- end -}}
 apiVersion: apps/v1
 kind: Deployment
 metadata:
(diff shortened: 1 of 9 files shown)


helm/hpcc/templates/esp.yaml
```diff
@@ -144,6 +154,9 @@ kind: ConfigMap
 {{- end }}
 {{- end }}
 
-{{ include "hpcc.addCertificate" (dict "root" $ "name" .name "service" . "component" $application "visibility" .service.visibility) }}
+{{ include "hpcc.addCertificate" (dict "root" $ "name" .name "service" .service "component" $application "visibility" .service.visibility) }}
+{{- if $commonCtx.externalCert }}
```
afishbeck, line 158: These 3 sections took me a while to work out.  May just want a comment explaining that these are to create and mount the local mtls certificate for the ESP when the service is public.

system/security/securesocket/securesocket.cpp
```diff
@@ -1824,9 +1823,7 @@ SECURESOCKET_API ISecureSocketContext* createSecureSocketContext(SecureSocketTyp
     }
     else
     {
-        if(server_securesocket_context.get() == NULL)
-            server_securesocket_context.setown(new securesocket::CSecureSocketContext(sockettype));
-        return server_securesocket_context.getLink();
+        return new securesocket::CSecureSocketContext(sockettype);
```
afishbeck, line 1826: Can we get rid of the if entirely?  Both outcomes now seem to look the same.

system/security/securesocket/securesocket.cpp
```diff
@@ -1839,9 +1836,7 @@ SECURESOCKET_API ISecureSocketContext* createSecureSocketContextEx(const char* c
     }
     else
     {
-        if(server_securesocket_context.get() == NULL)
-            server_securesocket_context.setown(new securesocket::CSecureSocketContext(certfile, privkeyfile, passphrase, sockettype));
-        return server_securesocket_context.getLink();
+        return new securesocket::CSecureSocketContext(certfile, privkeyfile, passphrase, sockettype);
```
afishbeck, line 1839: Same, get rid of if?

system/security/securesocket/securesocket.cpp
```diff
@@ -1857,9 +1852,7 @@ SECURESOCKET_API ISecureSocketContext* createSecureSocketContextEx2(IPropertyTre
     }
     else
     {
-        if(server_securesocket_context.get() == NULL)
-            server_securesocket_context.setown(new securesocket::CSecureSocketContext(config, sockettype));
-        return server_securesocket_context.getLink();
+        return new securesocket::CSecureSocketContext(config, sockettype);
```
afishbeck, line 1855: Same.

afishbeck (dismissed):
A couple of minor comments.

system/security/securesocket/securesocket.cpp
jakesmith: yes, I should have noticed. Can also get rid of the CS which was only there to protect the global.
Changed.

system/security/securesocket/securesocket.cpp
jakesmith: Changed.

system/security/securesocket/securesocket.cpp
jakesmith: Changed.

system/security/securesocket/securesocket.cpp
```diff
@@ -1813,54 +1812,23 @@ class CRsaCertificate : implements ICertificate, public CInterface
...
-    {
-        return new securesocket::CSecureSocketContext(sockettype);
-    }
-    else
-    {
-        if(server_securesocket_context.get() == NULL)
-            server_securesocket_context.setown(new securesocket::CSecureSocketContext(sockettype));
-        return server_securesocket_context.getLink();
-    }
+    return new securesocket::CSecureSocketContext(sockettype);
```
mckellyln, line 1818: I think we decided to try and use the same context.  Why this change ?

mckellyln: ok, answered

mckellyln (approved):
Looks good.
Thor sort ok.

system/security/securesocket/securesocket.cpp
jakesmith: yes - description in commit comment

PR #15181, opened 2021-07-17
HPCC-26214 Check acceptable file path in ESP services

wangkx:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/services/ws_fileio/ws_fileioservice.cpp
+++ b/esp/services/ws_fileio/ws_fileioservice.cpp
@@ -37,6 +37,9 @@ bool CWsFileIOEx::CheckServerAccess(const char* targetDZNameOrAddress, const cha
     if (!targetDZNameOrAddress || (targetDZNameOrAddress[0] == 0) || !relPath || (relPath[0] == 0))
         return false;
 
+    if (containsRelPaths(relPath)) //Detect a path like: a/../../../f
+        throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "Invalid file path %s", relPath);
+
     netAddr.clear();
     Owned<IEnvironmentFactory> factory = getEnvironmentFactory(true);
     Owned<IConstEnvironment> env = factory->openEnvironment();
--- a/esp/services/ws_fs/ws_fsBinding.cpp
+++ b/esp/services/ws_fs/ws_fsBinding.cpp
@@ -394,7 +394,7 @@ int CFileSpraySoapBindingEx::downloadFile(IEspContext &context, CHttpRequest* re
         request->getParameter("OS", osStr);
         request->getParameter("Path", pathStr);
         request->getParameter("Name", nameStr);
-        
+
 #if 0
         StringArray files;
         IProperties* params = request->queryParameters();
@@ -432,6 +432,9 @@ int CFileSpraySoapBindingEx::downloadFile(IEspContext &context, CHttpRequest* re
         if (*(pathStr.str() + pathStr.length() -1) != pathSep)
             pathStr.append( pathSep );
 
+        if (!validateDropZonePath(nullptr, netAddressStr, pathStr)) //The pathStr should be the absolute path for the dropzone.
+            throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "Invalid file path %s", pathStr.str());
+
         StringBuffer fullName;
         fullName.appendf("%s%s", pathStr.str(), nameStr.str());
 
@@ -468,6 +471,17 @@ int CFileSpraySoapBindingEx::downloadFile(IEspContext &context, CHttpRequest* re
         response->handleExceptions(xslp, me, "FileSpray", "DownloadFile", StringBuffer(getCFD()).append("./smc_xslt/exceptions.xslt"));
     }
     return 0;
+}
+
+int CFileSpraySoapBindingEx::onStartUpload(IEspContext& ctx, CHttpRequest* request, CHttpResponse* response, const char* serv, const char* method)
+{
+    StringBuffer netAddress, path;
+    request->getParameter("NetAddress", netAddress);
+    request->getParameter("Path", path);
+    if (!validateDropZonePath(nullptr, netAddress, path)) //The path should be the absolute path for the dropzone.
+        throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "Invalid Landing Zone path %s", path.str());
+
+    return EspHttpBinding::onStartUpload(ctx, request, response, serv, method);
 }
 
 int CFileSpraySoapBindingEx::onFinishUpload(IEspContext &ctx, CHttpRequest* request, CHttpResponse* response,   const char *service, const char *method, StringArray& fileNames, StringArray& files, IMultiException *me)
--- a/esp/services/ws_fs/ws_fsBinding.hpp
+++ b/esp/services/ws_fs/ws_fsBinding.hpp
@@ -74,8 +74,9 @@ public:
         ensureNavLink(*folder, "XRef", "/WsDFUXRef/DFUXRefList", "View Xref result details or run the Xref utility", NULL, NULL, 12);
     }
 
-    int onGetInstantQuery(IEspContext &context, CHttpRequest* request, CHttpResponse* response, const char *service, const char *method);
-    int onFinishUpload(IEspContext &ctx, CHttpRequest* request, CHttpResponse* response, const char *service, const char *method, StringArray& fileNames, StringArray& files, IMultiException *me);
+    virtual int onGetInstantQuery(IEspContext &context, CHttpRequest* request, CHttpResponse* response, const char *service, const char *method) override;
+    virtual int onStartUpload(IEspContext& ctx, CHttpRequest* request, CHttpResponse* response, const char* serv, const char* method) override;
+    virtual int onFinishUpload(IEspContext &ctx, CHttpRequest* request, CHttpResponse* response, const char *service, const char *method, StringArray& fileNames, StringArray& files, IMultiException *me) override;
     virtual bool canDetachFromDali() override
     {
         return false;
--- a/esp/services/ws_fs/ws_fsService.cpp
+++ b/esp/services/ws_fs/ws_fsService.cpp
@@ -2815,6 +2815,9 @@ bool CFileSprayEx::onFileList(IEspContext &context, IEspFileListRequest &req, IE
                 throw MakeStringException(ECLWATCH_ACCESS_TO_FILE_DENIED, "Only cfg or log file allowed.");
         }
 
+        if (!validateDropZonePath(nullptr, netaddr, sPath) && !validateConfigurationDirectory(nullptr, "log", nullptr, nullptr, sPath)) //The path should be the absolute path for the dropzone or log file.
+            throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "Invalid file path %s", sPath.str());
+
         RemoteFilename rfn;
         SocketEndpoint ep;
 #ifdef MACHINE_IP
--- a/esp/services/ws_topology/ws_topologyService.cpp
+++ b/esp/services/ws_topology/ws_topologyService.cpp
@@ -236,6 +236,16 @@ bool CWsTopologyEx::onTpLogFile(IEspContext &context,IEspTpLogFileRequest  &req,
     return true;
 }
 
+void CWsTopologyEx::validateFilePath(const char *file, const char *fileType, const char *compType, const char *compName)
+{
+    StringBuffer actualPath;
+    splitUNCFilename(file, nullptr, &actualPath, nullptr, nullptr);
+    if (containsRelPaths(actualPath)) //Detect a path like: /home/lexis/runtime/var/log/HPCCSystems/myesp/../../../
+        throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "Invalid file path %s", actualPath.str());
+    if (!validateConfigurationDirectory(nullptr, fileType, compType, compName, actualPath))
+        throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "Invalid file path %s", actualPath.str());
+}
+
 bool CWsTopologyEx::onSystemLog(IEspContext &context,IEspSystemLogRequest  &req, IEspSystemLogResponse &resp)
 {
     try
@@ -254,6 +264,7 @@ bool CWsTopologyEx::onSystemLog(IEspContext &context,IEspSystemLogRequest  &req,
         }
         else
         {
+            validateFilePath(name, "log", nullptr, nullptr);
             logname = name;
         }
 
@@ -1117,7 +1128,10 @@ void CWsTopologyEx::readTpLogFile(IEspContext &context,const char* fileName, con
 {
     StringBuffer logname;
     if (strcmp(fileType,"thormaster_log"))
+    {
+        validateFilePath(fileName, "log", nullptr, nullptr);
         logname = fileName;
+    }
     else
         logname.append(CCluster(fileName)->queryRoot()->queryProp("LogFile"));
 
@@ -1743,6 +1757,8 @@ bool CWsTopologyEx::onTpGetComponentFile(IEspContext &context, IEspTpGetComponen
 
         StringBuffer uncPath;
         uncPath.append(pathSepChar).append(pathSepChar).append(netAddress).append(sDir).append(fileName);
+
+        validateFilePath(uncPath, (stricmp(fileType, "log") == 0) ? "log" : "run", nullptr, nullptr);
 
         if (stricmp(fileType, "log") == 0)
         {
--- a/esp/services/ws_workunits/ws_workunitsHelpers.cpp
+++ b/esp/services/ws_workunits/ws_workunitsHelpers.cpp
@@ -4011,6 +4011,19 @@ IFileIOStream* CWsWuFileHelper::createWUFileIOStream(IEspContext& context, const
     return createIOStreamWithFileName(zipFileNameWithPath.str(), IFOread);
 }
 
+void CWsWuFileHelper::validateFilePath(const char *file, bool UNCFileName, const char *fileType, const char *compType, const char *compName)
+{
+    StringBuffer actualPath;
+    if (UNCFileName)
+        splitUNCFilename(file, nullptr, &actualPath, nullptr, nullptr);
+    else
+        splitFilename(file, nullptr, &actualPath, nullptr, nullptr);
+    if (containsRelPaths(actualPath)) //Detect a path like: /home/lexis/runtime/var/log/HPCCSystems/myesp/../../../
+        throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "Invalid file path %s", actualPath.str());
+    if (!validateConfigurationDirectory(nullptr, fileType, compType, compName, actualPath))
+        throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "Invalid file path %s", actualPath.str());
+}
+
 void CWsWuFileHelper::readWUFile(const char* wuid, const char* workingFolder, WsWuInfo& winfo, IConstWUFileOption& item,
     StringBuffer& fileName, StringBuffer& fileMimeType)
 {
@@ -4028,11 +4041,18 @@ void CWsWuFileHelper::readWUFile(const char* wuid, const char* workingFolder, Ws
     case CWUFileType_CPP:
     case CWUFileType_LOG:
     {
-        const char *tail=pathTail(item.getName());
-        fileName.set(tail ? tail : item.getName());
+        const char *file=item.getName();
+#ifndef _CONTAINERIZED
+        validateFilePath(file, false, "run", nullptr, nullptr);
+#else
+        validateFilePath(file, false, "query", nullptr, nullptr);
+#endif
+
+        const char *tail=pathTail(file);
+        fileName.set(tail ? tail : file);
         fileMimeType.set(HTTP_TYPE_TEXT_PLAIN);
         fileNameWithPath.set(workingFolder).append(PATHSEPCHAR).append(fileName.str());
-        winfo.getWorkunitCpp(item.getName(), item.getDescription(), item.getIPAddress(), mb, true, fileNameWithPath.str());
+        winfo.getWorkunitCpp(file, item.getDescription(), item.getIPAddress(), mb, true, fileNameWithPath.str());
         break;
     }
     case CWUFileType_DLL:
@@ -4052,11 +4072,16 @@ void CWsWuFileHelper::readWUFile(const char* wuid, const char* workingFolder, Ws
         writeToFileIOStream(workingFolder, fileName.str(), mb);
         break;
     case CWUFileType_ThorLog:
+    {
+        const char *file=item.getName();
+        validateFilePath(file, true, "log", nullptr, nullptr);
+
         fileName.set("thormaster.log");
         fileMimeType.set(HTTP_TYPE_TEXT_PLAIN);
         fileNameWithPath.set(workingFolder).append(PATHSEPCHAR).append(fileName.str());
-        winfo.getWorkunitThorMasterLog(nullptr, item.getName(), mb, fileNameWithPath.str());
+        winfo.getWorkunitThorMasterLog(nullptr, file, mb, fileNameWithPath.str());
         break;
+    }
     case CWUFileType_ThorSlaveLog:
     {
         fileName.set("ThorSlave.log");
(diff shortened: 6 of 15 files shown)


esp/services/ws_topology/ws_topologyService.cpp
```diff
@@ -236,6 +236,16 @@ bool CWsTopologyEx::onTpLogFile(IEspContext &context,IEspTpLogFileRequest  &req,
     return true;
 }
 
+void CWsTopologyEx::checkFilePath(const char *file, const char *fileType, const char *compType, const char *compName)
```
afishbeck, line 239: minor, really just personal preference, and feel free to disagree, I would name the functions "validate" rather than "check" when their primary purpose is security related.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -4011,6 +4011,19 @@ IFileIOStream* CWsWuFileHelper::createWUFileIOStream(IEspContext& context, const
     return createIOStreamWithFileName(zipFileNameWithPath.str(), IFOread);
 }
 
+void CWsWuFileHelper::checkFilePath(const char *file, bool UNCFileName, const char *fileType, const char *compType, const char *compName)
+{
+    StringBuffer actualPath;
+    if (UNCFileName)
```
afishbeck, line 4017: You are splitting the path, validating one part, and then using the orinal path.  Is there any chance the other parts can be used to hack the system.  It may be the way they are parsed they won't have at least "../" but it's worth considering.

esp/services/ws_topology/ws_topologyService.cpp
wangkx: I will change it.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
wangkx: The other parts are machine part (url or IP) and the tail part (the part after the last PathSepChar). Any idea how to check them?

wangkx: The other parts are the machine part (url or IP) and the tail part (the part after the last PathSepChar). I haven't got an idea for the tail part. For the machine part, we may consider (in another PR?) to read the allowed machine list (with/without the allowed folder list for each machine) when the ESP is started (and when environment changes).

GordonSmith: If its available (c++17) you should probably run the full path through: 
 std::filesystem::canonical (https://en.cppreference.com/w/cpp/filesystem/canonical) as it will clean it up and give you the "real" path, including symlinks etc.

wangkx: I just tried:
#include <iostream>
#include <filesystem>
namespace fs = std::filesystem;

I got: error: ‘filesystem’ is not a namespace-name

GordonSmith: Its ultimately a wrapper around realpath:  https://pubs.opengroup.org/onlinepubs/9699919799/functions/realpath.html

Which we do use already...

GordonSmith: Actually looks like jfile has wrapped it in `makeAbsolutePath`?

wangkx: After using std::experimental::filesystem::path, I got:

/home/lexis/src3/HPCC-Platform/esp/services/ws_fs/ws_fsService.cpp:2800: undefined reference to `std::experimental::filesystem::v1::current_path[abi:cxx11]()'
/usr/bin/ld: /home/lexis/src3/HPCC-Platform/esp/services/ws_fs/ws_fsService.cpp:2800: undefined reference to `std::experimental::filesystem::v1::canonical(std::experimental::filesystem::v1::__cxx11::path const&, std::experimental::filesystem::v1::__cxx11::path const&)'
/usr/bin/ld: CMakeFiles/ws_fs.dir/ws_fsService.cpp.o: in function `std::experimental::filesystem::v1::__cxx11::path::path<char [16], std::experimental::filesystem::v1::__cxx11::path>(char const (&) [16])':
/usr/include/c++/9/experimental/bits/fs_path.h:212: undefined reference to `std::experimental::filesystem::v1::__cxx11::path::_M_split_cmpts()'
collect2: error: ld returned 1 exit status

wangkx: I just tried the makeAbsolutePath:
```
    StringBuffer s1, s2, s3, s4;
    makeAbsolutePath("//10.0.1.15/a/b/../c/./d/e", s1);
    makeAbsolutePath("/a/b/../c/./d/e", s2);
    makeAbsolutePath("~/a/b/../c/./d/e", s3);
    makeAbsolutePath("\\x\\y\\..\\..\\a\\b\\c", s4);
    DBGLOG("####results s1=(%s), s2=(%s), s3=(%s), s4=(%s)", s1.str(), s2.str(), s3.str(), s4.str());
```
I got: "####results s1=(//10.0.1.15/a/b/../c/./d/e), s2=(/a/b/../c/./d/e), s3=(/home/lexis/a/b/../c/./d/e), `s4=(\x\y\..\..\a\b\c)"
I am not sure it helps...

GordonSmith: Hmmm the comments suggest it should be handling those...   If you debug it does it hit "realpath" at anypoint?

wangkx: Yes for s2, s3, and s4.

wangkx: After adding the link_libraries(stdc++fs) to CMakeLists, the std::experimental::filesystem::canonical works. But, it does not work for a remote file.

jakesmith (changes requested):
@wangkx - please see comments

esp/smc/SMCLib/TpWrapper.cpp
```diff
@@ -2191,4 +2191,74 @@ extern TPWRAPPER_API unsigned getContainerWUClusterInfo(CConstWUClusterInfoArray
...
+        {
+            if (matchNetAddressRequest(netAddr, isIPAddressReq, tpMachines.item(ii)))
+                return true;
+        }
+    }
+#else
+    Owned<IEnvironmentFactory> envFactory = getEnvironmentFactory(true);
+    Owned<IConstEnvironment> constEnv = envFactory->openEnvironment();
+    Owned<IConstDropZoneInfoIterator> dropZoneItr = constEnv->getDropZoneIteratorByAddress(netAddr);
+    ForEach(*dropZoneItr)
```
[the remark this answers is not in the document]
wangkx: Not for7.12.x. That change was introduced to 8.0.0 by https://github.com/hpcc-systems/HPCC-Platform/commit/18ea181ac341f83c7c75258e1dd0a24b3ef4451f.

esp/services/ws_fileio/ws_fileioservice.cpp
```diff
@@ -37,6 +37,9 @@ bool CWsFileIOEx::CheckServerAccess(const char* targetDZNameOrAddress, const cha
     if (!targetDZNameOrAddress || (targetDZNameOrAddress[0] == 0) || !relPath || (relPath[0] == 0))
         return false;
 
+    if (hasSpecialPathCharacters(relPath)) //Detect a path like: a/../../../f
```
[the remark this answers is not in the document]
wangkx: I guess I have to check the char after a dot.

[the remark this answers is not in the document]
jakesmith: it's more that it/they appear on it's own after a path separator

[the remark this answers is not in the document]
wangkx: Right. Is it enough just to check the char before the dot?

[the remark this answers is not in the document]
wangkx: If the dot is the first char, it also should be invalid.

wangkx:
@jakesmith I made the changes based on your comments. Please review.

system/jlib/jfile.cpp
```diff
@@ -5252,12 +5252,18 @@ bool hasSpecialPathCharacters(const char *path)
...
+    char sepChar = getPathSepChar(path);
+    char previousChar = '\0';
     const char *cur = path;
     for (;;)
     {
-        switch (*cur++)
+        switch (*cur)
         {
         case '.':
+            if ((previousChar=='\0') || (previousChar==sepChar)) //the dot is the frist char or after a path separator.
```
jakesmith, line 5264: I think ok, in the sense that the context hasSpecialPathCharacters() is used it's unlikely to trip up,
but this could still produce false negatives, e.g. /.mydir/ is a valid path (& and % are valid chars really too).

If the main objective is to spot relative paths, perhaps this method should be renamed (e.g. containsRelPaths), and it explitly walks and looks for only '~', '.' and '..'

wangkx: Ok, I will: 1. rename; 2. only looks for '~', '.' and '..'; 3. check the char after a dot.

wangkx: Is the /..mydir/ a valid path?

jakesmith: yes "." and ".." only mean anything if they're a path directive, leading/trailing or between path separators.

jakesmith: and technically '~' is only meaningful as leading, e.g. really 'a~b' is a legal posix path/filename

wangkx:
@jakesmith please review

jakesmith (approved):
Looks ok to me.
@afishbeck - could you give it one more look please

afishbeck (approved):
@wangkx yes, looks fine to me.

wangkx:
@richardkchapman this PR has been approved.

afishbeck (approved):
Additional change looks fine.

wangkx:
I just did squash.

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -2815,6 +2815,9 @@ bool CFileSprayEx::onFileList(IEspContext &context, IEspFileListRequest &req, IE
                 throw MakeStringException(ECLWATCH_ACCESS_TO_FILE_DENIED, "Only cfg or log file allowed.");
         }
 
+        if (!validateDropZonePath(nullptr, netaddr, sPath) && !validateConfigurationDirectory(nullptr, "log", nullptr, nullptr, sPath)) //The path should be the absolute path for the dropzone or log file.
```
ghalliday, line 2818: I know this has generally been the case, but requiring an absolute name for the dropzone or log path seems wrong - since it is exposing internal information that shouldn't be required.  Should probably be revisited in a separate PR

esp/services/ws_topology/ws_topologyService.cpp
```diff
@@ -236,6 +236,16 @@ bool CWsTopologyEx::onTpLogFile(IEspContext &context,IEspTpLogFileRequest  &req,
     return true;
 }
 
+void CWsTopologyEx::validateFilePath(const char *file, const char *fileType, const char *compType, const char *compName)
+{
+    StringBuffer actualPath;
+    splitUNCFilename(file, nullptr, &actualPath, nullptr, nullptr);
+    if (containsRelPaths(actualPath)) //Detect a path like: /home/lexis/runtime/var/log/HPCCSystems/myesp/../../../
+        throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "Invalid file path %s", actualPath.str());
+    if (!validateConfigurationDirectory(nullptr, fileType, compType, compName, actualPath))
```
ghalliday, line 245: This is only going to work for log files if there isn't an override for a particular instance in the environment file, and then also only works because [INSTANCE] is the trailing component.  It probably works for most existing environment files, but not for all.

ghalliday, line 245: Potential performance issues because this will reread environment.xml each time it is called?  (getOSSdirTree() is not cached.)

esp/scm/ws_topology.ecm
```diff
@@ -575,6 +576,7 @@ ESPrequest [nil_remove] SystemLogRequest
     string Name;
     string Type;
     int   Zip;
+    [min_ver("1.32")] string CompType;
```
ghalliday, line 579: This should have some kind of documentation.  Does eclwatch need to be updated to provide this argument, or will it mean that the legacy SystemLog option will no longer work.

esp/scm/ws_topology.ecm
```diff
@@ -543,6 +543,7 @@ ESPrequest [nil_remove] TpLogFileRequest
     [min_ver("1.05")] int PageNumber(0);
     [min_ver("1.05")] bool LoadData(false);
     [min_ver("1.29")] bool IncludeLogFieldNames(true);
+    [min_ver("1.32")] string CompType;
```
ghalliday, line 546: This will require changes to eclwatch (e.g. WsTopology.ts, tplog.xslt) for it to work.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -4028,11 +4041,18 @@ void CWsWuFileHelper::readWUFile(const char* wuid, const char* workingFolder, Ws
     case CWUFileType_CPP:
     case CWUFileType_LOG:
     {
-        const char *tail=pathTail(item.getName());
-        fileName.set(tail ? tail : item.getName());
+        const char *file=item.getName();
+#ifndef _CONTAINERIZED
+        validateFilePath(file, false, "run", nullptr, nullptr);
+#else
+        validateFilePath(file, false, "query", nullptr, nullptr);
```
ghalliday, line 4048: This is potentially not going to work for containerized systems - you can have c++ files and dlls stored in multiple locations.  An unusual case though.
The idea that you can get a single location and then compare it as a prefix is likely to have other instances (like log overrides in bare metal) where it doesn't really work.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -4052,11 +4072,16 @@ void CWsWuFileHelper::readWUFile(const char* wuid, const char* workingFolder, Ws
         writeToFileIOStream(workingFolder, fileName.str(), mb);
         break;
     case CWUFileType_ThorLog:
+    {
+        const char *file=item.getName();
+        validateFilePath(file, true, "log", nullptr, nullptr);
```
ghalliday, line 4077: doesn't this need a component type?

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -4068,11 +4093,16 @@ void CWsWuFileHelper::readWUFile(const char* wuid, const char* workingFolder, Ws
         break;
     }
     case CWUFileType_EclAgentLog:
+    {
+        const char *file=item.getName();
+        validateFilePath(file, true, "log", nullptr, nullptr);
```
ghalliday, line 4098: also requires a component type?

ghalliday (commented):
@wangkx a few comments.  I've been very picky since this is being merged into 7.12.x.
Especially looking for places where this will break existing functionality.  I think on fairly standard configurations it should work, but weird environment.xml file mappings are not going to work.
Why is CompType added to the topology calls?  If it is required there are quite a few other calls that are missing it.
The general assumption that you can get a single directory, and then check it is a prefix is likely to not work in various contexts.

esp/scm/ws_topology.ecm
wangkx: The CompType is used to set the component variable inside the getConfigurationDirectory(). So far, I haven't found which ConfigurationDirectory needs that variable. I add the CompType here just in case we need it some time later. Do you know when the component variable has to be specified?

esp/services/ws_topology/ws_topologyService.cpp
wangkx: Do we know where those special log file could be located? Should we add some settings to specify the paths to be allowed for retrieving the log files (and force users to modify their environment.xml)?

wangkx: Do you know why getOSSdirTree() is not cached?

esp/services/ws_fs/ws_fsBinding.cpp
```diff
@@ -470,6 +473,17 @@ int CFileSpraySoapBindingEx::downloadFile(IEspContext &context, CHttpRequest* re
     return 0;
 }
 
+int CFileSpraySoapBindingEx::onStartUpload(IEspContext& ctx, CHttpRequest* request, CHttpResponse* response, const char* serv, const char* method)
+{
+    StringBuffer netAddress, path;
+    request->getParameter("NetAddress", netAddress);
+    request->getParameter("Path", path);
+    if (!validateDropZonePath(nullptr, netAddress, path)) //The path should be the absolute path for the dropzone.
```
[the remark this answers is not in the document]
wangkx: That code in the EspHttpBinding::onStartUpload is for adding package, not for dropzone file.

system/jlib/jutil.cpp
```diff
@@ -2715,6 +2715,22 @@ bool replaceConfigurationDirectoryEntry(const char *path,const char *frommask,co
...
+{
+    if (isEmptyString(dirToValidate))
+        return false;
+
+    StringBuffer configDir;
+    if (!getConfigurationDirectory(useTree, category, component, instance, configDir))
+        return false;
+    
+    addPathSepChar(configDir);
+    unsigned configDirLen = configDir.length();
```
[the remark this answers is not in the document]
wangkx: Good to know. I will change.

system/jlib/jfile.cpp
```diff
@@ -5248,6 +5248,47 @@ bool isRemotePath(const char *path)
...
+            break;
+        }
+        case '\0':
+            return false;
+        }
+
+        previousChar = *cur;
+        cur++;
+    }
+    return false;
```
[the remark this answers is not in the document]
wangkx: Will remove.

esp/smc/SMCLib/TpWrapper.cpp
```diff
@@ -2191,4 +2191,74 @@ extern TPWRAPPER_API unsigned getContainerWUClusterInfo(CConstWUClusterInfoArray
...
+{
+    if (ipReq)
+        return streq(netAddressReg, tpMachine.getNetaddress());
+    return streq(netAddressReg, tpMachine.getConfigNetaddress());
+}
+
+extern TPWRAPPER_API bool validateDropZonePath(const char* dropZoneName, const char* netAddr, const char* pathToCheck)
+{
+    if (isEmptyString(netAddr) || isEmptyString(pathToCheck))
+        throw MakeStringException(ECLWATCH_INVALID_INPUT, "NetworkAddress or Path not defined.");
```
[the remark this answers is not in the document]
wangkx: Will change

esp/services/ws_fs/ws_fsBinding.hpp
```diff
@@ -75,6 +75,7 @@ class CFileSpraySoapBindingEx : public CFileSpraySoapBinding
     }
 
     int onGetInstantQuery(IEspContext &context, CHttpRequest* request, CHttpResponse* response, const char *service, const char *method);
+    int onStartUpload(IEspContext& ctx, CHttpRequest* request, CHttpResponse* response, const char* serv, const char* method);
```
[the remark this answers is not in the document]
wangkx: I will add them.

esp/scm/ws_topology.ecm
ghalliday: My conclusion was for the default environment file none of default definitions need it to be set (because it is the trailing part of the file path).  Someone could define a strange environment where all cases will need it to be defined.  (It would also be required if a component-specific override was defined.)
ComponentType would be a better name than CompType, slightly clearer what it corresponds to.

esp/services/ws_fs/ws_fsBinding.cpp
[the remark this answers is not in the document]
ghalliday: I asked because you are calling through to that base definition, which suggests it is the function that would have been called previously.

esp/scm/ws_topology.ecm
wangkx: I haven't found a place where the CompType has to be set so far.

esp/services/ws_topology/ws_topologyService.cpp
ghalliday: Because no one has implemented it.  I noticed, and nearly add caching when I was modifying code in that area.  It needs to be a time limited cache - and should be a separate PR.

esp/scm/ws_topology.ecm
ghalliday: If it is not specified then the user will be able to get to any logs.

esp/services/ws_topology/ws_topologyService.cpp
ghalliday: They could theoretically be located anywhere.  I don't know if any real-life environments actually hit this.

ghalliday (commented):
@wangkx a few responses.
I'm not sure what should be done about component type.  For a default environment I don't think it is needed.  To cope with all environments you would need a different approach since you can also have exceptions for particular instances (generate a list of all potential paths and check if it matches any of them).
I'm not sure if adding CompType is worthwhile - since it raises the question of why it isn't set in other equivalent calls.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
ghalliday: ignore - see later comment

system/jlib/jfile.cpp
```diff
@@ -5248,6 +5248,47 @@ bool isRemotePath(const char *path)
...
 
+bool containsRelPaths(const char *path)
+{
+    if (isEmptyString(path))
+        return false;
+    if (*path == '~')
+        return true;
+
+    const char *cur = path;
+    char sepChar = getPathSepChar(path);
```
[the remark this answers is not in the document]
wangkx: How about I add a getPathSepCharEx() to implement this logic?

esp/services/ws_fs/ws_fsBinding.cpp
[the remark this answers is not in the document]
wangkx: Sorry, I think I did not explain it clearly. When adding a package, the CWsPackageProcessSoapBindingEx (not CFileSpraySoapBindingEx) is used. The onStartUpload is not overrided in CWsPackageProcessSoapBindingEx.  So, the EspHttpBinding::onStartUpload will be called directly for adding package. Both NetAddress and Path are empty when adding package.

esp/services/ws_topology/ws_topologyService.cpp
wangkx: As I know, the function here is used for accessing the configuration files and log files of HPCC components (most likely through ECLWatch). Any suggestion about how to do with those special log files?

esp/services/ws_workunits/ws_workunitsHelpers.cpp
wangkx: Any suggestion about how to do with the c++ files and dlls stored in multiple locations?

wangkx:
Since the CompType raises the question and no use case has been found yet, I should remove it for this PR.

ghalliday:
I am inclined to agree.  We should either pass null in all places, or allow the comp to be supplied in all places - currently it is a bit of a mixture.  With the standard environment it makes no difference since the component is the trailing entry on the path.  With a completely custom environment this solution may not work anyway (it would also need the instance).  It also potentially has issues with different instances storing information in different locations (unusual k8s configurations).

I would remove comp from the services, and accept that this change does a reasonable job on standard environments, but could fail on custom environments.  Possibly revisit the implementation if multiple dll storage planes (or something else that isn't supported) turns out to be something that is used.

wangkx:
@ghalliday please review. If I miss anything, please let me know.

esp/services/ws_topology/ws_topologyService.cpp
ghalliday: Created HPCC-26316

system/jlib/jfile.cpp
```diff
@@ -5248,6 +5248,46 @@ bool isRemotePath(const char *path)
...
+    for (;;)
+    {
+        switch (*cur)
+        {
+        case '.':
+        {
+            if (previousChar == sepChar)
+            {
+                cur++;
+                if ((*cur == sepChar) || (*cur == '\0')) // '.'
```
ghalliday, line 5270: @jakesmith should this be disallowed?  @GordonSmith could this cause any problems?
It means \var\lib\HPCCSystems\myhthor\.\hthor.log
will be disallowed.  Could there be any existing code where .\hthor.log is passed in as a relative path?  (Not just for logs but any of these affected functions.)

system/jlib/jfile.hpp
```diff
@@ -491,6 +491,19 @@ inline char getPathSepChar(const char *dir)
     return s?(*s):((*dir&&(dir[1]==':'))?'\\':PATHSEPCHAR);
 }
 
+inline char getPathSepCharEx(const char *s)
+{
+    bool foundBackslash=false;
+    if (s) while (*s) {
```
ghalliday, line 497: I see that the code is cloned and modified, but...
non standard nesting of if/while and bracket placement.  Probably better with the definition in the c++ file (reduces compile times and complication of the header.

ghalliday (commented):
@wangkx a couple of minor comments.  I think it would be ok to merge as-is, but I would like to double check that /./ will not cause problems, and just about worth cleaning up the jfile function.

PR #15188, opened 2021-07-20
HPCC-26238 Add scoped classes to manage gauge metrics

kenrowland:
Added classes to manage gauges within a scope

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Cloud-compatibility
  - [x] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jmetrics.hpp
+++ b/system/jlib/jmetrics.hpp
@@ -136,10 +136,7 @@ public:
     GaugeMetric(const char *name, const char *description) :
         MetricVal{name, description, MetricType::METRICS_GAUGE}  { }
 
-    /*
-     * Update the value as indicated
-     */
-    void add(int64_t delta)
+    void adjust(int64_t delta)
     {
         value += delta;
     }
@@ -267,5 +264,49 @@ std::shared_ptr<CustomMetric<T>> createCustomMetricAndAddToReporter(const char *
     queryMetricsReporter().addMetric(pMetric);
     return pMetric;
 }
+
+
+class jlib_decl ScopedGaugeUpdater
+{
+public:
+    explicit ScopedGaugeUpdater(GaugeMetric &_pGauge, int64_t _amount=1)
+        : gauge{_pGauge}, amount{_amount}
+    {
+        gauge.adjust(amount);
+    }
+    ScopedGaugeUpdater(const ScopedGaugeUpdater&) = delete;
+    ScopedGaugeUpdater(ScopedGaugeUpdater&) = delete;
+    ScopedGaugeUpdater& operator=(const ScopedGaugeUpdater&) = delete;
+    ScopedGaugeUpdater& operator=(ScopedGaugeUpdater&) = delete;
+    ~ScopedGaugeUpdater()
+    {
+        gauge.adjust(-amount);
+    }
+
+protected:
+    GaugeMetric &gauge;
+    int64_t amount;
+};
+
+
+class jlib_decl ScopedGaugeDecrementer
+{
+public:
+    explicit ScopedGaugeDecrementer(GaugeMetric &_pGauge, int64_t _amount=1)
+        : gauge{_pGauge}, amount{_amount}
+    { }
+    ScopedGaugeDecrementer(const ScopedGaugeDecrementer&) = delete;
+    ScopedGaugeDecrementer(ScopedGaugeDecrementer&) = delete;
+    ScopedGaugeDecrementer& operator=(const ScopedGaugeDecrementer&) = delete;
+    ScopedGaugeDecrementer& operator=(ScopedGaugeDecrementer&) = delete;
+    ~ScopedGaugeDecrementer()
+    {
+        gauge.adjust(-amount);
+    }
+
+protected:
+    GaugeMetric &gauge;
+    int64_t amount;
+};
 
 }
--- a/system/metrics/testing/metrictest.cpp
+++ b/system/metrics/testing/metrictest.cpp
@@ -130,12 +130,12 @@ void processThread(int numLoops, unsigned delay, bool addDynamic, const std::str
 
         if (pDynamicMetric)
         {
-            pDynamicMetric->add(1);
+            pDynamicMetric->adjust(1);
         }
 
         pEventCountMetric->inc(2u);
-        pQueueSizeMetric->add(3);
+        pQueueSizeMetric->adjust(3);
         std::this_thread::sleep_for(std::chrono::seconds(delay));
-        pQueueSizeMetric->add(-1);
+        pQueueSizeMetric->adjust(-1);
     }
 }
--- a/testing/unittests/metrics/MetricFrameworkTests.cpp
+++ b/testing/unittests/metrics/MetricFrameworkTests.cpp
@@ -109,11 +109,11 @@ protected:
 
         //
         // Test updating gauge
-        pGauge->add(10);
+        pGauge->adjust(10);
         gaugeValue = pGauge->queryValue();
         CPPUNIT_ASSERT_EQUAL(35, gaugeValue);
 
-        pGauge->add(-5);
+        pGauge->adjust(-5);
         gaugeValue = pGauge->queryValue();
         CPPUNIT_ASSERT_EQUAL(30, gaugeValue);
     }


kenrowland:
@jakesmith Please review.

I created two scoped classes. One handles the increment and decrement of the gauge and is designed for use in a scope where the gauge should increment on entry and decrement on exit (even if an exception is thrown)

The second handles decrement of the gauge for cases where the increment happened prior to entering the scope.

Also changed the methods on the gauge to inc and dec. Left add temporarily for when the metric unit tests are merged. Once that is done I will remove it and update the test cases.

No ESP changes required since the current gauge in its instrumentation is a custom metric.

You may have better class name suggestions.

Thanks.

system/jlib/jmetrics.hpp
```diff
@@ -262,4 +273,42 @@ std::shared_ptr<CustomMetric<T>> createCustomMetricAndAddToReporter(const char *
...
+        pGauge->inc(amount);
+    }
+
+    ~ScopedGaugeUpdater()
+    {
+        pGauge->dec(amount);
+    }
+
+protected:
+    std::shared_ptr<GaugeMetric> pGauge;
```
jakesmith, line 293: I'm concerned about the efficiency here.
If ScopedGaugeUpdater is used to track very rapid transactions, then the overhead of linking/using a shared_ptr is going to be more than the inc/dec itself.

Does it need to be a shared_ptr at this point? and/or can it just be a reference to one?

system/jlib/jmetrics.hpp
```diff
@@ -262,4 +273,42 @@ std::shared_ptr<CustomMetric<T>> createCustomMetricAndAddToReporter(const char *
...
+    AutoGaugeDecrement(const std::shared_ptr<GaugeMetric> &_pGauge, int64_t _amount=1):
+        pGauge{_pGauge},
+        amount{_amount} { }
+    ~AutoGaugeDecrement()
+    {
+        pGauge->dec(amount);
+    }
+
+protected:
+    std::shared_ptr<GaugeMetric> pGauge;
```
jakesmith, line 310: as per comment above, can this be a GaugeMetric or a reference to a shared_ptr to avoid the overhead?

jakesmith (commented):
a couple of comments - main concern is cost of using these.

system/jlib/jmetrics.hpp
kenrowland: @jakesmith 
I have been hesitant to store shared pointers as references since the reference does not increment the reference count. Should the owner of the scoped object decide to delete the object, the scoped object may fail when decrementing. I see some solutions:
1. use a reference and don't worry about it (i.e. trust that it won't be deleted)
2. Use 1, but add a null check prior to the decrement (if (pGauge) { do the decrement })
3. Live with the overhead (probably not acceptable)
4. Use a try-finally.

What do you think about (2)? It eliminates the copy and provides the safety in case the metric were deleted. I would also ensure the copy and assignment constructors are deleted for additional safety.

jakesmith: The component created the metric and the component is using the guage updater to update it.
It needs a very low overhead way of doing so.
I think 1 is fine - the semantics are that during the live of the scoped object, we assert the metric is valid.
The componet/creator would have to invalidate it for that not to be so.

2. would still be additional wasted overhead, you'd need a weak pointer / lock to avoid potential race condition - but as per above - there shouldn't be one.

kenrowland: I'll go with 1 and make the change. Since use of the class is the responsibility of the component, there is also no need to delete the copy and assignment constructors.

kenrowland:
@jakesmith Updated per our conversation.

system/jlib/jmetrics.hpp
```diff
@@ -262,4 +273,50 @@ std::shared_ptr<CustomMetric<T>> createCustomMetricAndAddToReporter(const char *
     return pMetric;
 }
 
+
+class jlib_decl ScopedGaugeUpdater
+{
+public:
+    explicit ScopedGaugeUpdater(const std::shared_ptr<GaugeMetric> &_pGauge, int64_t _amount=1) :
+        gauge{*_pGauge},
```
jakesmith, line 281: I doesn't seem to any good reason to keep this taking a shared_ptr type, instead of the underlying type itself.

As it stands, someone can't pass in a GaugeMetric type, forcing the usage to unecessarily use shared_ptr type.
Imagine a sequence of code that wants to take a metric and pass it around and then somewhere deeper use an updater, it doesn't necessarily want to pass around a reference to a shared_ptr. It can if it wants, it depends on the use case, but the updater doesn't need it.

You could define it as:
```
    explicit ScopedGaugeUpdater(GaugeMetric *_pGauge, int64_t _amount=1) :
        gauge{*_pGauge},
```
and that would be more correct/flexible than having it take a reference to a shared_ptr,
but shared_ptr doesn't define an overload operator to pointer, so you would have to call .get() on the shared_ptr, e.g.:
```
    hpccMetrics::ScopedGaugeUpdater updater(myGauge.get());
```
which is fine, but you could (and I think I would) make it take a reference to the underlying type, i.e.:
```
    explicit ScopedGaugeUpdater(GaugeMetric &_pGauge, int64_t _amount=1) :
        gauge{_pGauge},
```
And since shared_ptr does overload the dereference operator, you can then use as, e.g:
```
    hpccMetrics::ScopedGaugeUpdater updater(*myGauge);
```

NB: we tend to use Owned in our code base, and passing in an Owned would automatically invoke the operator and pass in the underlying ptr to the object.

system/jlib/jmetrics.hpp
```diff
@@ -262,4 +273,50 @@ std::shared_ptr<CustomMetric<T>> createCustomMetricAndAddToReporter(const char *
...
+        gauge.dec(amount);
+    }
+
+protected:
+    GaugeMetric &gauge;
+    int64_t amount;
+};
+
+
+class jlib_decl AutoGaugeDecrement
```
jakesmith, line 302: it's still a scoped utiltiy class, to be consistent with other one (ScopedGaugeUpdater), I think this should be called ScopedGagueDecrement

jakesmith (changes requested):
@kenrowland - looks fine except the shared_ptr's as arguments to the updaters.

kenrowland:
@jakesmith Please see latest changes.

system/jlib/jmetrics.hpp
```diff
@@ -262,4 +273,50 @@ std::shared_ptr<CustomMetric<T>> createCustomMetricAndAddToReporter(const char *
...
+
+class jlib_decl ScopedGaugeUpdater
+{
+public:
+    explicit ScopedGaugeUpdater(GaugeMetric &_pGauge, int64_t _amount=1) :
+        gauge{_pGauge},
+        amount{_amount}
+        {
+            gauge.inc(amount);
+        }
```
jakesmith, line 285: super trivial/style: level of indentation looks a little unusual, i.e. this is at a higher depth than other implementations.
We tend to put the member initializers either on the same line at the ctor (if there's a few), or grouped on next lines, but the inline method impl. at same level as the ctor, e.g.:
```
explicit ScopedGaugeUpdater(GaugeMetric &_pGauge, int64_t _amount=1)
    : gauge{_pGauge}, amount{_amount}
{
    gauge.inc(amount);
} 
```

jakesmith (commented):
@kenrowland - looks good, 1 trivial comment. Please check and squash.

kenrowland:
@jakesmith Made formatting change and squashed. Please approve.

jakesmith (dismissed):
@kenrowland - looks good

kenrowland:
@richardkchapman Please merge

system/jlib/jmetrics.hpp
```diff
@@ -136,14 +136,25 @@ class jlib_decl GaugeMetric : public MetricVal
...
+    void add(int64_t delta)
+    {
+        value += delta;
+    }
+
     /*
      * Update the value as indicated
      */
-    void add(int64_t delta)
+    void inc(uint64_t delta)
```
ghalliday, line 147: minor: What is the benefit of the new functions, rather than calling add(amount) and add(-amount)?

system/jlib/jmetrics.hpp
```diff
@@ -262,4 +273,48 @@ std::shared_ptr<CustomMetric<T>> createCustomMetricAndAddToReporter(const char *
...
+        gauge.dec(amount);
+    }
+
+protected:
+    GaugeMetric &gauge;
+    int64_t amount;
+};
+
+
+class jlib_decl ScopedGaugeDecrement
```
ghalliday, line 300: picky: For consistency either this should be ScopedGaugeDecrementer, or previous class should be ScopedGaugeUpdate

ghalliday (commented):
@kenrowland looks like a useful change.  A couple of minor comments.

system/jlib/jmetrics.hpp
kenrowland: Jake asked for more meaningful method names.

system/jlib/jmetrics.hpp
kenrowland: I can make that change. It's been a journey to find reasonable names.

system/jlib/jmetrics.hpp
ghalliday: From your initial comment:

> Also changed the methods on the gauge to inc and dec. Left add temporarily for when the metric unit tests are merged. Once that is done I will remove it and update the test cases.

I don't think Jake made any comment on these functions.

kenrowland: Most likely lost when rebased. Please get with him.

ghalliday: As far as I can see this hasn't been reopened, so it was part of your initial commit.

ghalliday: Jake says he did comment on it.  I would suggest one of the following

* have a single function, but rename it so it is neutral.  E.g. adjust()

* have two functions.  But parameters should be signed value so you can still say add(delta) where delta can be negative.

(I don't mind which.)

If you opt for 2 functions the original 3rd function should be deleted, because otherwise the public interface is confusing.

jakesmith: not sure where the comment was, I thought in PR, but perhaps it was offline.
I think from offline chat a single method as it was, but with a more generic name, e.g ."adjust" is preferred.

system/jlib/jmetrics.hpp
richardkchapman: "There are only two hard things in Computer Science: cache invalidation and naming things."

kenrowland (commented):
Added update to unit tests to match new gauge adjust method

kenrowland:
@ghalliday Can you please do a final review? I will rebase and request a merge once approved. Thanks.

ghalliday (dismissed):
@kenrowland please squash.

kenrowland:
@richardkchapman  please merge. Thanks.

richardkchapman (approved):

PR #15201, opened 2021-07-26
HPCC-26249 Removed xterm dependencies in the platform

Michael-Gardner:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:


Michael-Gardner:
tested on 20.04 and found no runtime issues after removing the xterm dependency

richardkchapman:
Why is this functionality not needed on newer Ubuntu systems?

richardkchapman:
@Michael-Gardner One question. Also at this point 8.4.x seems a more appropriate target

Michael-Gardner:
@richardkchapman The information returned by hpcc-run.sh running hpcc-init on older ubuntu distro's and systemd on newer distros has changed, so the output differs depending on what the underlying init system is.  It doesn't really 'display' the output of starting the platform on the newer distributions. The reason behind using xterm was so we could launch and display the old hpcc-init output to the end user so they knew something was being done.  If this tool is still relevant to people we could keep it around and I could rewrite it to output something for when we're using systemd.

We already have it conditionally being excluded for Containerized builds, so there's no extra bloat there.

richardkchapman:
@Michael-Gardner Please retarget to master/8.4.x

richardkchapman (approved):

richardkchapman:
@Michael-Gardner Looks ok (though it will need to target master now since it has been stalled for so long), but wasn't the point of this change to remove the dependency on xterm? And if so should there be some change to a dependency list and/or a change to what is installed into containers?

richardkchapman:
Actually I see that one of the deleted files was responsible for adding the xterm dependency (and this code is not enabled in containerized). So ignore my question above.

PR #15217, opened 2021-07-30
HPCC-26292 New ESDL script configuration parsing

timothyklemm:
Add wrapper of xpp::XmlPullParser to handle complex configuration markup
with minimal impact existing parsing logic.

- Declare an interface from xpp::XmlPullparser's public API and make the
  wrapper extend this interface.
- Support detection and use of external XML fragments based on start tags.
- Support detection and use of embedded XML fragments based on start tags
  and content.
- Support generic conversion of property tree configuration nodes into input
  for the wrapper. Ensure that serialized property tree nodes can be fixed
  up to be compatible with the ESDL service engine's serialization.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/bindings/SOAP/xpp/fxpp/FragmentedXmlAssistant.hpp
+++ b/esp/bindings/SOAP/xpp/fxpp/FragmentedXmlAssistant.hpp
@@ -1,0 +1,175 @@
+/*##############################################################################
+
+    HPCC SYSTEMS software Copyright (C) 2021 HPCC Systems®.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+############################################################################## */
+
+#ifndef FXPP_FRAGMENTED_XML_ASSISTANT_H
+#define FXPP_FRAGMENTED_XML_ASSISTANT_H
+
+#include "fxpp/FragmentedXmlPullParser.hpp"
+
+namespace fxpp
+{
+
+namespace fxa
+{
+/**
+ * The `fxpp::fxa` namespace represents one possible implementation of `IFragmentedXmlAssistant`.
+ * All declared interfaces may be reused by alternate implementations if such reuse makes sense,
+ * but this is not required.
+ *
+ * To support the detection of embedded fragment content, one or more calls to
+ * `IAssistant::registerEmbeddedFragment` must be made prior to parsing markup. Each call
+ * designates an additional element URI and name combination that is allowed to contain embedded
+ * content.
+ *
+ * To support the detection of external fragment references, one or more calls to
+ * `IAssistant::registerExternalFragment` must be combined with calls to
+ * `IAssistant::setExternalDetector` and `IAssistant::setExternalLoader`. The assistant knows when
+ * to look for an external reference but delegates locating references to the registered
+ * `IExternalDetector` instance and loading referenced content to the registered
+ * `IExternalContentLoader` instance. `IExternalDetection` is used to report the detector's
+ * findings to the loader. `IExternalContentInjector` is used by the assistant to accept content
+ * from the loader.
+ */
+
+/**
+ * Abstraction representing the "identity" of an external fragment. The identity of a fragment is
+ * the information located by a detector that can be used by a loader to get the fragment data. It
+ * is assumed that a fragment may be uniquely identified by at least one key and at most two keys.
+ * The first, or location, may be a file path or resource bundle identifier. The second, if used,
+ * may be an XPath referring to a nodeset within a broader XML document or some other identifier
+ * used to refine the data identified by the primary key.
+ */
+interface IExternalDetection : extends IInterface
+{
+    virtual int         queryDetectedType() const = 0;
+    virtual const char* queryContentLocation() const = 0;
+    virtual const char* queryContentLocationSubset() const = 0;
+    virtual const char* queryUID() const = 0;
+
+    inline bool matchesType(int type) const { return queryDetectedType() == type; }
+};
+
+/**
+ * A collection of some well known external reference detection types. This collection may not be
+ * comprehensive. It gives names to values with meaning to the implementation and may be extended
+ * to include other values, but such extension is not required.
+ *
+ * - Unknown:  A value has not been assigned. Registered detection keys must not use this value.
+ * - Resource: The location value identifies a resource bundle's resource. The subset value is
+ *             not currently significant.
+ * - Path:     The location value identifies the file system path of an XML file. The subset value
+ *             is not currently significant.
+ */
+enum ExternalDetectionType
+{
+    Unknown,
+    Resource,
+    Path,
+};
+
+/**
+ * Abstraction describing a delegate responsible for locating external fragment reference values
+ * within parser start tag structures.
+ */
+interface IExternalDetector : extends IInterface
+{
+    virtual IExternalDetection* detectExternalReference(const StartTag& stag) const = 0;
+};
+
+/**
+ * Abstraction describing a helper responsible for accepting fragment content from an
+ * `IExternalContentLoader` instance. An assistant will provide an instance of this interface to
+ * a loader instance as the only means of propagating loaded content back to the parser.
+ */
+interface IExternalContentInjector : extends IInterface
+{
+    virtual bool injectContent(const char* content, int contentLength) = 0;
+};
+
+/**
+ * Abstraction describing a delegate responsible for loading detected content for an assistant.
+ * It is separate from the assistant to support alternate sources; a unit test might access in-
+ * memory data while other use cases might load files or perform resource bundle lookups.
+ */
+interface IExternalContentLoader : extends IInterface
+{
+    virtual bool loadAndInject(const IExternalDetection& detected, IExternalContentInjector& injector) const = 0;
+};
+
+/**
+ * Extension of the `IFragmentedXmlAssistant` interface adding awareness of the delegate
+ * interfaces and supported markup knowledge.
+ *
+ * `registerEmbeddedFragment` is the only required method to support embedded fragments.
+ *
+ * `setExternalLoader` and `registerExternalFragment` are required to support external fragment
+ * references. `setExternalDetector` is also required if any registered fragments do not register
+ * an element-specific detector. 
+ */
+interface IAssistant : extends IFragmentedXmlAssistant
+{
+    virtual void setExternalDetector(const IExternalDetector* detector) = 0;
+    virtual void setExternalLoader(const IExternalContentLoader* loader) = 0;
+    virtual void registerExternalFragment(const char* nsUri, const char* name, XmlFragmentRules rules) = 0;
+    virtual void registerExternalFragment(const char* nsUri, const char* name, XmlFragmentRules rules, const IExternalDetector& detector) = 0;
+    virtual void registerEmbeddedFragment(const char* nsUri, const char* name, XmlFragmentRules rules) = 0;
+};
+
+/**
+ * Definition of a single rule used by an external detector to locate external fragment references.
+ * A collection of these may be provided to `createExternalDetector` to initialize the default
+ * detector with alternate rules.
+ */
+struct DetectionKey
+{
+    const char* location = nullptr;
+    const char* subset = nullptr;
+    int         type = ExternalDetectionType::Unknown;
+
+    DetectionKey(const char* _location, const char* _subset, int _type) : location(_location), subset(_subset), type(_type) {}
+};
+
+/**
+ * Create an instance of the default `IAssistant` implementation.
+ */
+extern IAssistant* createAssistant();
+
+/**
+ * Create an instance of the default `IExternalDetector` implementation. The detection keys soecify
+ * all markup-specific detection rules used by the instance. Default rules are not merged with these
+ * keys.
+ */
+extern IExternalDetector* createExternalDetector(const std::initializer_list<DetectionKey>& keys);
+
+/**
+ * Create an instance of the default `IExternalDetector` implementation. The absence of detection
+ * keys implies a default configuration will be used. The default configuration is:
+ *
+ * | Location | Subset | Type |
+ * | :-: | :-: | :- |
+ * | resource | | ExternalDetectionType::Resource |
+ * | resource | match | ExternalDetectionType::Resource |
+ * | path | | ExternalDetectionType::Path |
+ * | path | match | ExternalDetectionType::Path |
+ */
+inline IExternalDetector* createExternalDetector() { return createExternalDetector({}); }
+
+} // namespace fxa
+
+} // namespace fxpp
+
+#endif // FXPP_FRAGMENTED_XML_ASSISTANT_H
--- a/esp/bindings/SOAP/xpp/fxpp/FragmentedXmlPullParser.cpp
+++ b/esp/bindings/SOAP/xpp/fxpp/FragmentedXmlPullParser.cpp
@@ -1,0 +1,819 @@
+/*##############################################################################
+
+    HPCC SYSTEMS software Copyright (C) 2021 HPCC Systems®.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+############################################################################## */
+
+#include <cstring>
+#include "fxpp/FragmentedXmlPullParser.hpp"
+#include <list>
+#include <map>
+#include <memory>
+#include <set>
+#include <vector>
+
+using namespace xpp;
+
+namespace fxpp
+{
+
+/**
+ * Named type for internal parser rules related to fragment injection. Internal rules do not
+ * use bits that overlap potential external rules, to allow an internal combination of both.
+ */
+using XmlInternalFragmentRules = uint16_t;
+
+/**
+ * Instruct the parser to require a non-empty unique identifier for a fragment. The identifier is
+ * used to detect circular references between fragments which means that the same identifier must
+ * be generated for each fragment.
+ *
+ * Identifiers are required for all fragments with the exception of the initial parser input and
+ * embedded fragments. In both of these cases, the fragment content cannot close a loop and content
+ * leading to a loop will be detected when an external fragment is applied.
+ */
+static const XmlInternalFragmentRules FXPP_REQUIRE_FRAGMENT_UID     = 0x0100;
+
+/**
+ * Instruct the parser to enforce the injection restriction requiring an in-progress element.
+ * Fragments set as parser input and requested from an assistant do not not require this check.
+ */
+static const XmlInternalFragmentRules FXPP_REQUIRE_CURRENT_ELEMENT  = 0x0200;
+
+/**
+ * `IFagmentedXmlPullParser` implementation that uses a stack of `XmlPullParser` instances to
+ * support parsing fragments within fragments. To a great extent, it is a wrapper of
+ * `XmlPullParser`, and what works with the original parser should work with this class with no
+ * required changes. There are exceptions:
+ * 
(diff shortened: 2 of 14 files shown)


esp/bindings/SOAP/xpp/fxpp/FragmentedXmlPullParser.h
```diff
@@ -0,0 +1,2145 @@
...
+ * This framework attempts to address several complexity concerns introduced by some current and
+ * future configurable components, such as ESDL integration scripts:
+ *
+ * - If a configuration is accessed through a property tree, is the configuration given as a
+ *   property tree element hierarchy or as an element containing text that is the configuration?
+ * - Is it reasonable to assume that the previous question need only be answered once for a
+ *   complete configuration? With configuration reuse encouraged and component reuse happening,
+ *   it does seem likely that the configuration of one component may contain the configuration
+ *   of a second component, defined as embedded markup within.
+ * - If multiple elements in a configuration may contain embedded markup, how many locations
```
asselitx, line 63: 'locations' meaning different places in the platform code or in plugins to the platform?

timothyklemm: Places in code. Consider ESDL integration script parsing. The parser gets passed around to all of the operations, which are responsible for parsing themselves. How many operations might have to be changed for each one to handle embedded markup?

This might not be the best example, because I wouldn't recommend many operations be enabled for this, but I expect to have script markup, with possible embedded markup, within markup for other components, which might also have embedded markup.

esp/bindings/SOAP/xpp/fxpp/FragmentedXmlPullParser.h
```diff
@@ -0,0 +1,2145 @@
...
+{
+    constexpr static const char* SYNTHESIZED_ROOT = "synthesized_root";
+    constexpr static const char* SYNTHESIZED_GROUP = "synthesized_group";
+}
+
+/**
+ * `IXmlFragment` implementation that assumes the character pointers given in its constructor will
+ * remain valid for the life of the instance.
+ */
+class CSimpleXmlFragment : public CInterfaceOf<IXmlFragment>
```
afishbeck, line 341: I think this class would lend itself to unsafe behavior in real world implementations.  It ok for testing and internal use, but maybe it should be local static implementations of this type of class in those cpp files.  Don't want to make people think this class is safe to use.  By unsafe I mean that we don't know how long the black box will be referencing these pointers and could for example cause shut down issues.  Another class that did reference counting around the content would be better.

esp/bindings/SOAP/xpp/fxpp/FragmentedXmlPullParser.h
```diff
@@ -0,0 +1,2145 @@
...
+ * `IFagmentedXmlPullParser` implementation that uses a stack of `XmlPullParser` instances to
+ * support parsing fragments within fragments. To a great extent, it is a wrapper of
+ * `XmlPullParser`, and what works with the original parser should work with this class with no
+ * required changes. There are exceptions:
+ * 
+ * - Mixed content, the ability for an element to contain both text and child element content, is
+ *   explicitly not supported. Attempting to enable this mode will result in an exception.
+ * - Whitespace content events are ignored by default.
+ */
+class CFragmentedXmlPullParser : public IFragmentedXmlPullParser
```
afishbeck, line 373: Since we use interfaces it would be better to put these classes in *.cpp files and expose a create function so that externally only interfaces are used.  Interfaces would need to be declared outside of any classes.  Any classes that require the user to derive there own classes from them should go in *.ipp files.

esp/bindings/SOAP/xpp/fxpp/FragmentedXmlPullParser.h
```diff
@@ -0,0 +1,2145 @@
...
+     * Abstraction representing the "identity" of an external fragment. The identity of a fragment
+     * is the information located by a detector that can be used by a loader to get the fragment
+     * data. It is assumed that a fragment may be uniquely identified by at least one key and at
+     * most two keys. The first, or primary, key may be a file path or resource bundle identifier.
+     * The second, if used, may be an XPath referring to a nodeset within a broader XML document.
+     */
+    interface IExternalDetection : extends IInterface
+    {
+        virtual int         queryDetectedType() const = 0;
+        virtual const char* queryPrimaryKey() const = 0;
```
afishbeck, line 1118: Not a fan of calling something that might be a filepath a "key" even if it may be a look up when loading from a resource pool.  Id rather either be explicit with these function names... queryFilePath() or only have one queryLocation() and have the meaning depend on the type.

esp/bindings/SOAP/xpp/fxpp/FragmentedXmlPullParser.h
```diff
@@ -0,0 +1,2145 @@
...
+
+    /**
+     * Abstraction describing a helper (an assistant's assistant) responsible for locating fragment
+     * identifiers within parser start tag structures. The detection of fragments is separated from
+     * the assistant to allow multiple implementations to support different markup language
+     * requirements.
+     */
+    interface IExternalDetector : extends IInterface
+    {
+        virtual IExternalDetection* detect(const StartTag& stag) const = 0;
```
afishbeck, line 1133: How about?:

virtual IDetectedContent *detectExternalContent(const StartTag& stag) const = 0;

If this is implemented in a class that implements multiple interfaces, detect may not be a unique enough name to make it intuitive.

esp/bindings/SOAP/xpp/fxpp/FragmentedXmlPullParser.h
```diff
@@ -0,0 +1,2145 @@
...
+    };
+
+    /**
+     * Abstraction describing a helper responsible for loading detected fragments for an assistant.
+     * It is separate from the assistant to support alternate sources; a unit test might access in-
+     * memory data while other use cases might load files or perform resource bundle lookups.
+     */
+    interface IExternalLoader : extends IInterface
+    {
+        virtual IXmlFragment* load(const IExternalDetection& detected, XmlFragmentRules rules) const = 0;
```
afishbeck, line 1143: As we discussed in our meeting consider having the loader just call back into the calling object and passing the pointers and other info that currently goes into the IXmlFragment object.

For example (and names are just examples and feel free to change them):

 ```
   interface IFragementLocation : extends IInterface
   {
      virtual void addContent(const char *id, const char *content, unsigned length, , XmlFragmentRules rules) const = 0;
   };

   interface IExternalLoader : extends IInterface
   {
      virtual void load(IFragmentLocation &location, const IDetectedContent& detected, XmlFragmentRules rules) const = 0;
   };
```
As we discussed, to make this useful, you would have to figure out a way to make it efficient such as immediately create the XPP parser so it immediately has the copy of the data.

File: esp/bindings/SOAP/xpp/fxpp/FragmentedXmlPullParser.h [hunk 7]
```diff
@@ -0,0 +1,2145 @@
...
+    IMPLEMENT_IINTERFACE;
+    IMPLEMENT_IXMLFRAGMENT_WITH(CBaseXmlFragment);
+
+    CSafeXmlFragment(const char* uid, const char* content, int length, XmlFragmentRules rules)
+        : CBaseXmlFragment(uid, content, length, rules)
+    {
+    }
+};
+
+class CBaseXmlFragmentGroup : public CBaseXmlFragment
```
afishbeck, line 1445: I would hope that once the classes are in the cpp and only the external interfaces are exposed that the concept of FragmentGroup would be internal to the cpp file and not exposed outside the black box?

esp/bindings/SOAP/xpp/fxpp/FragmentedXmlPullParser.h
```diff
@@ -0,0 +1,2145 @@
...
+ *   the serialization of the child elements. No trigger is included and no group members
+ *   are created.
+ * - If the node contains child elements and tag is not omitted, the group content contains
+ *   the serialization of the node, in its entirety, with no trigger and no group members.
+ * - If the node is empty and the tag is omitted, the group content a synthesized placeholder
+ *   with no group members. It is low overhead enabling any node to be represented without
+ *   pre-validating its content.
+ * - If the node is empty and the tag is not omitted, the group content is the group name
+ */
+class CPTreeXmlFragment : implements IXmlFragmentGroup, public CBaseXmlFragmentGroup
```
afishbeck, line 1636: As we discussed I think that in general while we may use a ptree at the top level, we would only ever want nested content to be loaded from child content that is already parsed as a ptree in cases needed for backward compatibility.  No new functionality should support that.  The reason being that we shouldn't have to worry about ordering issues in our content.

esp/bindings/SOAP/xpp/fxpp/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+################################################################################
```
afishbeck, line 1: Does an empty CMakeLists.txt file serve a purpose?

esp/bindings/SOAP/xpp/fxpp/FragmentedXmlPullParser.hpp
```diff
@@ -0,0 +1,186 @@
...
+     * option should rarely be necessary.
+     */
+    virtual void injectFragment(const char* uid, const char* content, int contentLength, XmlFragmentRules rules) = 0;
+    virtual void injectFragment(const char* uid, const char* prefix, int prefixLength, const char* body, int bodyLength, const char* suffix, int suffixLength, XmlFragmentRules rules) = 0;
+
+    /**
+     * Parser extension controlling whether content events for content containing whitespace
+     * characters only are ignored (false) or retained (true).
+     */
+    virtual void keepWhitespace(bool enabled) = 0;
```
afishbeck, line 134: trivial: this is an action, I would name the parameter "enable".

esp/bindings/SOAP/xpp/fxpp/FragmentedXmlPullParser.hpp
```diff
@@ -0,0 +1,186 @@
...
+
+    /**
+     * Parser extension controlling whether event data returned from `readStartTag`, `readContent`,
+     * and `readEndTag` remains valid until the parser is destroyed or its input is reset. If true,
+     * pointers returned from these methods are guaranteed to remain valid while parsing. If false,
+     * the caller should not rely on the pointers after the next call to `next`.
+     */
+    virtual void setIndefiniteEventDataRetention(bool enabled) = 0;
```
afishbeck, line 142: same.

esp/bindings/SOAP/xpp/fxpp/FragmentedXmlPullParser.hpp
```diff
@@ -0,0 +1,186 @@
...
+     * parser's event consumer.
+     */
+    constexpr static const char* SYNTHESIZED_ROOT = "synthesized_root";
+}
+
+extern IFragmentedXmlPullParser* createParser();
+
+} // namespace fxpp
+
+#endif // FXPP_FRAGMENTED_XML_FULL_PARSER_H_
```
afishbeck, line 186: Add newline.

esp/bindings/SOAP/xpp/xpp/StartTag.h
```diff
@@ -39,13 +39,28 @@ namespace xpp {
...
+    }
+
     ~StartTag() {
       if(attArr != NULL) {
         delete [] attArr;
       }
       attArr = NULL;
     }
 
+    StartTag& operator = (const StartTag& src) {
```
afishbeck, line 53: I understand why you added this.  It would be nice if you could have another way of creating your shallow StartTag copy.  This operator= could easily be used by accident by someone who didn't understand the implications.  So, same thing, but in function that is less likely to be misused.

afishbeck (commented):
A couple comments, otherwise looks good.

esp/bindings/SOAP/xpp/fxpp/CMakeLists.txt
timothyklemm: I can't think of one. I just followed the lead of the xpp, xpp/xpp, and xpp/sxt directories, which also include empty files. I figured it gave me a 75% chance of being right, since xpp/xjx is the only parser directory that doesn't have an empty file.

esp/bindings/SOAP/xpp/xpp/StartTag.h
timothyklemm: This doesn't enable something that would otherwise be impossible. It explicitly makes something that has always been possible safe by deep-copying `attArr`.  Are you asking for an accidental copying to crash the system, which is what happens without this operator?

afishbeck: Do we use the existing copy anywhere?  I'd add a private operator= to prevent it's use and avoid confusion.  But if we use it and this is a slight improvement it might be ok.

timothyklemm: The parser reads start tags before returning from `next`. `readStartTag` can only be called once before the next call to `next`. Assignment is used to update the output parameter with the previously cached data.

afishbeck: Ok sounds like an improvement then.

esp/bindings/SOAP/xpp/fxpp/FragmentedXmlPullParser.cpp
```diff
@@ -0,0 +1,818 @@
...
+                // should not need to be parsed.
+                hasChildren = true;
+            }
+            m_deferred.clear();
+        }
+
+        DataFrame* df = nullptr;
+        // Expect an equal number of start and end tags before a start tag has been read, and one
+        // extra end tag after reading a start tag.
+        int level = (m_activeElements.empty() ? 0 : 1);
```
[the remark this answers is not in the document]
timothyklemm: We never need the count of active elements here. We only need to know if there is is an active element. If there is an active element, parsing has begun and we need to see one more end tag than we see start tags to balance the already parsed start tag. If there is no active element, parsing has not started and we're skipping the entire document content.

We don't tend to skip before we begin parsing, but nothing prevents it from happening and this is an inexpensive check to handle it.

esp/bindings/SOAP/xpp/fxpp/FragmentedXmlPullParser.h
```diff
@@ -0,0 +1,2145 @@
...
+
+/**
+ * Instruct the parser to enforce the injection restriction requiring an in-progress element.
+ */
+#define FXPP_REQUIRE_CURRENT_ELEMENT  XmlInternalFragmentRules(0x0200)
+
+/**
+ * Reserved element names used internally by the parser.
+ */
+namespace reserved
```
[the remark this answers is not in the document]
timothyklemm: I opted for documentation and programmer beware, but did consider reserving a GUID. I wasn't convinced the collision risk was great enough to worry about it.

esp/bindings/SOAP/xpp/fxpp/FragmentedXmlPullParser.h
```diff
@@ -0,0 +1,2145 @@
...
+
+/**
+ * An extension of IXmlFragment enabling serialized processing of a variable number of discrete
+ * fragmants. Intended primarily for initial parser input, its use is not limited to this use case.
+ * A group is defined by three parts, with each implementation defining which are required or
+ * optional.
+ * 
+ * # A group may include any consumer-visible document content it deems necessary. Built-in
+ *   implementations use this to define a parent of all group members.
+ * # A group should include a synthesized trigger element that the parser will recognize as the
```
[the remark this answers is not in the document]
timothyklemm: N/A

testing/unittests/fxpptests.cpp
```diff
@@ -0,0 +1,1390 @@
...
+      const char* expectedGroupContent = R"!!!(
+        <PTree><synthesized_group/></PTree>
+      )!!!";
+      std::vector<const char*> expectedMemberContents({
+        R"!!!(<Embedded/>)!!!",
+      });
+      compareFragment("PTree/embedded/encoded", *fragment, expectedGroupContent, expectedMemberContents);
+    }
+
+    void testPTree_Children_RetainRoot()
```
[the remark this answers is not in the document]
timothyklemm: N/A

esp/bindings/SOAP/xpp/fxpp/FragmentedXmlPullParser.h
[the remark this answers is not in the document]
timothyklemm: N/A

testing/unittests/fxpptests.cpp
```diff
@@ -0,0 +1,1390 @@
...
+      cfg.expectEquivalence = false;
+      cfg.useAssistant = false;
+      runTest(cfg, nullptr, nullptr);
+    }
+    void testEmptyInput()
+    {
+      TestConfig cfg("empty-input");
+      cfg.expectEquivalence = false;
+      cfg.useAssistant = false;
+      runTest(cfg);
```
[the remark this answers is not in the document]
timothyklemm: Yes

esp/bindings/SOAP/xpp/fxpp/FragmentedXmlPullParser.h
```diff
@@ -0,0 +1,2145 @@
...
+ * the parser's default behavior is to replace the original element, in its entirety, with the
+ * fragment. This instructs the parser to retain the fragment identifying element and inject the
+ * fragment markup as descendents of that element.
+ */
+#define FXPP_RETAIN_FRAGMENT_PARENT   XmlFragmentRules(0x01)
+
+/**
+ * Make all namespaces in effect at the time of injection available to the injected content. Each
+ * namespace URI and prefix is defined in the fragment, eliminating the need to repeat definitions
+ * in each fragment.
```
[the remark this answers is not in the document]
timothyklemm: Added elaboration to the comment.

esp/bindings/SOAP/xpp/fxpp/FragmentedXmlPullParser.hpp
```diff
@@ -0,0 +1,186 @@
...
+ *
+ * To parse multiple fragments with manual control of additional fragments, an instance of
+ * `fxpp::IFragmentedXmlPullParser` will accept new fragments with the `injectFragment` method.
+ * This method cannot be called before the first `START_TAG` is reported, nor can it be called
+ * after the balancing `END_TAG` is reported.
+ *
+ * To parse multiple fragments with automatic detection of external fragment references, an instance
+ * of `fxpp::IFragmentedXmlAssistant` must be registered with the `fxpp::IFragmentedXmlPullParser`
+ * instance. The assistant implementation is responsible for recognizing which markup elements may
+ * specify external fragment references, which of those that may actually do specify external
```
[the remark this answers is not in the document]
timothyklemm: No, but will rephrase.

esp/bindings/SOAP/xpp/fxpp/FragmentedXmlPullParser.cpp
```diff
@@ -0,0 +1,818 @@
...
+            suffixBuf.append(suffix);
+        prepareInputSuffix(suffixBuf);
+        parser->setInput(prefixBuf, int(prefixBuf.length()), body, bodyLength, suffixBuf, int(suffixBuf.length()));
+        
+        return (new FragmentFrame(parser, uid, rules));
+    }
+
+    void checkProposedFragment(const char* uid, const char* prefix, int& prefixLength, const char* body, int& bodyLength, const char* suffix, int& suffixLength, XmlInternalFragmentRules rules)
+    {
+        #define LENGTH_ADJUSTMENT(c, l) if (-1 == (l) && (c)) l = int(strlen(c))
```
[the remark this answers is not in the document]
timothyklemm: Added comments.

esp/bindings/SOAP/xpp/fxpp/FragmentedXmlAssistant.hpp
```diff
@@ -0,0 +1,175 @@
...
+ *
+ * `setExternalLoader` and `registerExternalFragment` are required to support external fragment
+ * references. `setExternalDetector` is also required if any registered fragments do not register
+ * an element-specific detector. 
+ */
+interface IAssistant : extends IFragmentedXmlAssistant
+{
+    virtual void setExternalDetector(const IExternalDetector* detector) = 0;
+    virtual void setExternalLoader(const IExternalContentLoader* loader) = 0;
+    virtual void registerExternalFragment(const char* uri, const char* name, XmlFragmentRules rules) = 0;
```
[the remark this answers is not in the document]
timothyklemm: With `ns` being a common prefix for namespace, will rename to `nsUri`.

asselitx (commented):
These change look good to me, but I'll let Tony make final approval.

afishbeck (approved):
@timothyklemm approved.

richardkchapman:
@timothyklemm Should this target 8.4.x ?

timothyklemm:
@richardkchapman, it should be able to target 8.4.x. I've built and tested against master and am unaware of any issues that would prevent it. I'm not able to do anything this week to fix problems if I'm wrong; I'm on PTO in Nevada, with a very packed week.

PR #15250, opened 2021-08-11
HPCC-26381 Enable Prometheus metric service autodiscovery

rpastrana:
- Updates examples README
- Creates helper template fuctions to add metrics port
- Creates helper template fuctions to add prometheus specific annotations

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/helm/examples/metrics/README.md
+++ b/helm/examples/metrics/README.md
@@ -20,3 +20,36 @@ For example, the following adds the Prometheus configuration as the collection s
 ```code
 helm install mycluster ./hpcc -f <path>/prometheus_metrics.yml
 ```
+### Prometheus Collection System
+When enabled and properly configured, the HPCC metrics for Prometheus feature exposes an HTTP service on each pod of any HPCC component
+which reports metrics. It also annotates all HPCC pods to be discoverable by Prometheus servers and provides the connectivity inforation required, such as
+the port to liston on, and the url path on which to serve Prometheus formatted metrics. Autodiscovery annotations can be disabled via metrics.sinks[type=prometheus].settings.autodiscovery
+
+####Prometheus Metric Service Discovery
+On the Prometheus server side, an "additionalScrapeConfigs" entry is necessary to link the HPCC metrics services to Prometheus
+
+For example, the following Prometheus scrape job can be applied to Prometheus deployments as part of the 'additionalScrapeConfigs' configuration:
+
+```code
+    - job_name: 'hpcc-per-pod-metrics'
+      scrape_interval: 5s
+      metrics_path: /metrics
+      kubernetes_sd_configs:
+      - role: pod
+      relabel_configs:
+      - source_labels: [ __meta_kubernetes_pod_annotation_prometheus_io_scrape ]
+        action: keep
+        regex: true
+      - source_labels: [ __meta_kubernetes_pod_annotation_prometheus_io_path ]
+        action: replace
+        target_label: __metrics_path__
+        regex: (.+)
+      - source_labels: [ __address__, __meta_kubernetes_pod_annotation_prometheus_io_port ]
+        action: replace
+        regex: ([^:]+)(:\d+)?;(\d+)
+        replacement: ${1}:${3}
+        target_label: __address__
+```
+
+ More information on the "additionalScrapeConfigs" options provided by Prometheus here: https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config
+ 3rd Party write up on the relabel_configs used in the additionalscrapeconfigs here: https://gist.github.com/reachlin/a98b90afcbff4604c90c183a0169474f
--- a/helm/hpcc/templates/_helpers.tpl
+++ b/helm/hpcc/templates/_helpers.tpl
@@ -429,6 +429,28 @@ Pass in dict with root and secretsCategories
 {{- end -}}
 
 {{/*
+Generate Prometheus scrape annotations
+Enables selfdiscovery of metrics service on configured path/port
+Requires sinks[type=prometheus]
+Pass in dict with sinks
+*/}}
+{{- define "hpcc.addPrometheusScrapeAnnotations" -}}
+{{- if hasKey . "sinks" }}
+ {{ range $sink := .sinks -}}
+  {{- if eq (get $sink "type") "prometheus" }}
+   {{- if and (hasKey $sink "settings") ( hasKey $sink.settings "autodiscovery") }}
+    {{- if (eq $sink.settings.autodiscovery true ) }}
+prometheus.io/scrape: 'true'
+prometheus.io/path: {{ $sink.settings.path | default "/metrics" }}
+prometheus.io/port: {{ $sink.settings.port | default 8767 | quote }}
+    {{ end }}
+   {{ end }}
+  {{ end }}
+ {{ end}}
+{{ end}}
+{{- end -}}
+
+{{/*
 Add Secret volume for a component
 Pass in dict with root and secretsCategories
 */}}
@@ -1023,7 +1045,22 @@ spec:
   type: {{ $lvars.type }}
 {{- end -}}
 
-
+{{/*
+Generate prometheusMetricsReporter label if metrics.sinks[type=prometheus].
+Ranges over metric sinks map
+Pass in dict with sinks
+*/}}
+{{- define "hpcc.generateMetricsReporterLabel" }}
+ {{ range $sink := .sinks -}}
+  {{- if eq (get $sink "type") "prometheus" }}
+   {{- if and (hasKey $sink "settings") ( hasKey $sink.settings "autodiscovery") }}
+    {{- if (eq $sink.settings.autodiscovery true ) }}
+prometheusMetricsReporter: "yes"
+    {{ end }}
+   {{ end }}
+  {{ end }}
+ {{ end }}
+{{- end -}}
 {{/*
 Return access permssions for a given service
 */}}
--- a/helm/hpcc/templates/dali.yaml
+++ b/helm/hpcc/templates/dali.yaml
@@ -46,12 +46,20 @@ spec:
         server: {{ $dali.name | quote }}
         app: dali
         helmVersion: 8.3.0-trunk0
+{{- if hasKey $.Values.global "metrics" }}
+ {{- include "hpcc.generateMetricsReporterLabel" $.Values.global.metrics | nindent 8 }}
+{{- end }}
 {{- if hasKey $dali "labels" }}
 {{ toYaml $dali.labels | indent 8 }}
 {{- end }}
-{{- if hasKey $dali "annotations" }}
+{{- if or (hasKey $dali "annotations") (hasKey $.Values.global "metrics") }}
       annotations:
-{{ toYaml $dali.annotations | indent 8 }}
+ {{- if hasKey $.Values.global "metrics" }}
+  {{- include "hpcc.addPrometheusScrapeAnnotations" $.Values.global.metrics | nindent 8 }}
+ {{- end }}
+ {{- if hasKey $dali "annotations" }}
+  {{- toYaml $dali.annotations | nindent 8 }}
+ {{- end }}
 {{- end }}
     spec:
       {{- include "hpcc.placementsByPodTargetType" (dict "root" $ "pod" $dali.name "type" "dali") | indent 6 }}
--- a/helm/hpcc/templates/esp.yaml
+++ b/helm/hpcc/templates/esp.yaml
@@ -102,11 +102,17 @@ spec:
         accessDali: "yes"
         app: {{ $application }}
         helmVersion: 8.3.0-trunk0
+{{- if hasKey $.Values.global "metrics" }}
+ {{- include "hpcc.generateMetricsReporterLabel" $.Values.global.metrics | nindent 8 }}
+{{- end }}
 {{- if hasKey . "labels" }}
 {{ toYaml .labels | indent 8 }}
 {{- end }}
       annotations:
         checksum/config: {{ $configSHA }}
+{{- if hasKey  $.Values.global "metrics" }}
+ {{- include "hpcc.addPrometheusScrapeAnnotations" $.Values.global.metrics | nindent 8 }}
+{{- end }}
 {{- if hasKey . "annotations" }}
 {{ toYaml .annotations | indent 8 }}
 {{- end }}
--- a/helm/hpcc/templates/network.yaml
+++ b/helm/hpcc/templates/network.yaml
@@ -23,6 +23,7 @@
 */}}
 {{- $egress := (.Values.global.egress | default dict) -}}
 {{- $egressRestricted := $egress.restricted | default false -}}
+{{- $metrics := .Values.global.metrics -}}
 ---
 ##
 ## Default to no ingress, and egress only to dns servers
@@ -155,3 +156,29 @@ spec:
     - podSelector:
         matchLabels:
           app: eclservices
+---
+##
+## Provides access to pods labeled prometheusMetricsReporter: "yes"
+## onto port declared in global.metrics.sinks[type=prometheus].configuration.port
+## Only created if global.metrics.sinks[type=prometheus].settings.autodiscovery true
+{{- if hasKey $metrics "sinks" }}
+ {{ range $sink := $metrics.sinks -}}
+  {{- if eq (get $sink "type") "prometheus" }}
+   {{- if and (hasKey $sink "settings") (hasKey $sink.settings "autodiscovery") }}
+    {{- if (eq $sink.settings.autodiscovery true ) }}
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: hpcc-metrics-prometheus
+spec:
+  podSelector:
+    matchLabels:
+      prometheusMetricsReporter: "yes"
+  ingress:
+  - ports:
+    - port: {{ $sink.settings.port | default 8767 }}
+     {{ end }}
+    {{ end }}
+   {{ end }}
+  {{ end }}
+{{ end}}
--- a/helm/hpcc/templates/roxie.yaml
+++ b/helm/hpcc/templates/roxie.yaml
@@ -105,11 +105,17 @@ spec:
         run: {{ $commonCtx.toponame | quote }}
         roxie-cluster: {{ $roxie.name | quote }}
         helmVersion: 8.3.0-trunk0
+{{- if hasKey $.Values.global "metrics" }}
+ {{- include "hpcc.generateMetricsReporterLabel"  $.Values.global.metrics | nindent 8}}
+{{- end }}
 {{- if hasKey $toposerver "labels" }}
 {{ toYaml $toposerver.labels | indent 8 }}
 {{- end }}
       annotations:
         checksum/config: {{ $topoconfigSHA }}
+{{- if hasKey $.Values.global "metrics" }}
+ {{- include "hpcc.addPrometheusScrapeAnnotations"  $.Values.global.metrics | nindent 8 }}
+{{- end }}
 {{- if hasKey $toposerver "annotations" }}
 {{ toYaml $toposerver.annotations | indent 8 }}
 {{- end }}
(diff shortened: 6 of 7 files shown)


rpastrana:
@richardkchapman the point of this change is to expose the "metrics" http service as port on esp/dali/roxie k8s services and adds the appropriate annotations which is consumed by prometheus to determine which endpoints should be scraped. Right now the metrics config seems to be "global" only and likely requires making it per/component for better control of exposed ports.
```
kubectl get ep
NAME                                      ENDPOINTS                                                     AGE
eclqueries                                10.1.20.43:8880,10.1.20.43:8767                               3m16s
eclservices                               10.1.20.62:8880,10.1.20.62:8767                               3m16s
eclwatch                                  10.1.20.49:8767,10.1.20.49:8010                               3m16s
esdl-sandbox                              10.1.20.52:8880,10.1.20.52:8767                               3m16s
mydali                                    10.1.20.48:7070,10.1.20.48:8767                               3m16s
roxie                                     10.1.20.44:9876,10.1.20.51:9876,10.1.20.54:9876 + 5 more...   3m16s
roxie-toposerver                          10.1.20.50:8767,10.1.20.50:9004                               3m16s
sasha-coalescer                           10.1.20.48:8767,10.1.20.48:8877                               3m16s
sasha-dfuwu-archiver                      10.1.20.56:8767,10.1.20.56:8877                               3m16s
sasha-wu-archiver                         10.1.20.45:8767,10.1.20.45:8877                               3m16s
sql2ecl                                   10.1.20.53:8880,10.1.20.53:8767                               3m16s
```

richardkchapman:
I may have misunderstood, but it seems this is adding a second port to every published service, where I would have expected a new service on anything that wanted prometheus to be able to get metrics for.

rpastrana:
> 
> 
> I may have misunderstood, but it seems this is adding a second port to every published service, where I would have expected a new service on anything that wanted prometheus to be able to get metrics for.

I meant to point out the current metrics configuration strategy doesn't allow per/component metric settings. If we want to enable/disable Prometheus related Helm settings on a per component basis, that behavior would need to change.

Also, we can define a new k8s service to coincide with each of the background http services we standup within the same pod as the HPCC component being monitored. Those new k8s services would declare the same metricsport and use the base SVC as its selector (all traffic routed to base SVC). In my prototyping, this approach bloated our list of services and chose what I though was a less intrusive solution. The other concern I had, was being able to create NetworkPolicies targeting the metrics endpoint, which based on documentation we should be able to control even if we don't place the port on a new k8s service. Let me know if you want to create the new k8s services. @richardkchapman

richardkchapman:
I just can't see this approach working:

1. You will only be able to gather metrics for components that export a service - not all do 
2. A service may refer to multiple pods (load balanced) where the metrics wants to be per pod.

Seems to me (from 30 seconds scanning prometheus docs, so you almost certainly understand it far better than I do) that what we want to do is add annotations on all pods (that have implemented something for metrics):

annotations:
  prometheus.io/port: 9216
  prometheus.io/scrape: true

Then add common code into all processes that we want instrumented to answer http requests on that port with metrics info.

Probably will also need to add something in to the network rules to ensure that ingress from the prometheus server to these pods is allowed on that port.

helm/examples/metrics/README.md
```diff
@@ -20,3 +20,37 @@ For example, the following adds the Prometheus configuration as the collection s
...
+        - role: pod
+      relabel_configs:
+      - source_labels: [ __meta_kubernetes_pod_annotation_prometheus_io_scrape ]
+        action: keep
+        regex: true
+      - source_labels: [ __meta_kubernetes_pod_annotation_prometheus_io_path ]
+        action: replace
+        target_label: __metrics_path__
+        regex: (.+)
+      - source_labels: [ __address__, __meta_kubernetes_pod_annotation_prometheus_io_port ]
```
rpastrana, line 48: Doesn't seem to apply when __address__ (pre-label) doesn't contain a port. Port seems to be drawn from the pod's containerpod.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -429,6 +429,34 @@ Pass in dict with root and secretsCategories
...
+Generate Prometheus scrape annotations
+Enables selfdiscovery of metrics service on configured path/port
+Requires metrics.sinks[type=prometheus]
+Pass in metric config
+*/}}
+{{- define "hpcc.addPrometheusScrapeAnnotations" -}}
+{{- if hasKey . "sinks" }}
+ {{ range $sink := .sinks -}}
+  {{ range $sinkattindex, $sinkatts := $sink -}}
+   {{- if and  (eq $sinkattindex "type") }}
```
rpastrana, line 442: there has to be a straightforward way to test .sinks[type="prometheus"]

helm/hpcc/templates/esp.yaml
```diff
@@ -113,6 +115,13 @@ spec:
       serviceAccountName: "hpcc-default"
       containers:
       - name: {{ .name | quote }}
+{{ if hasKey . "service" }}
+ {{ if hasKey .service "servicePort" }}
```
rpastrana, line 119: Why don't ESP containers report containerport? From my reading, it's mostly "informative". We find ourselves requiring a containerport on ESP pod for Prometheus pod discovery. Any concerns?

rpastrana:
@jakesmith I'd like to get your take on the worthiness of the second commit (a66dc1b1febc97ae6c7671c0416fc64bf618165b), which attempts to add a single networkpolicy ( for all pods declared as 'prometheusMetricsReporter=yes').

helm/hpcc/templates/_helpers.tpl
jakesmith: I think you can simply the "type" test to:
```
{{ range $sink := .sinks -}}
 {{- if eq (get $sink "type") "prometheus" }}
```

get will return "" if missing.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -429,6 +429,34 @@ Pass in dict with root and secretsCategories
...
+    {{- if eq $sinkatts "prometheus" }}
+     {{- if hasKey $sink "settings" }}
+      {{- if hasKey $sink.settings "autodiscovery" }}
+       {{- if eq $sink.settings.autodiscovery true }}
```
jakesmith, line 446: This looks like a case where 'dig' would be useful.
I hadn't used before, but just tested it, can express the above as:
```
{{- if (dig "settings" "autodiscovery" false $sink) }}
```
i.e. all as:
```
{{ range $sink := .sinks -}}
 {{- if eq (get $sink "type") "prometheus" }}
  {{- if (dig "settings" "autodiscovery" false $sink) }}
        prometheus.io/scrape: 'true'
        prometheus.io/path: {{ default "/metrics" $sink.settings.path }}
        prometheus.io/port: {{ default 8767 $sink.settings.port | quote }}
   {{ end }}
  {{ end }}
{{ end}}
```

helm/hpcc/templates/_helpers.tpl
```diff
@@ -429,6 +429,34 @@ Pass in dict with root and secretsCategories
...
+        prometheus.io/scrape: 'true'
+        prometheus.io/path: {{ default "/metrics" $sink.settings.path }}
+        prometheus.io/port: {{ default 8767 $sink.settings.port | quote }}
```
jakesmith, line 449: Should not indent here (in functions), because that fixes the level of indentation and might not be right for each context it's used in, but instead use ' | indent <n>' where function used. 
Hence in all our other helper functions there is no indent of what they generate.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -1016,7 +1044,29 @@ spec:
...
+{{- define "hpcc.generateMetricsReporterLabel" }}
+ {{- if hasKey . "sinks" }}
+  {{ range $sink := .sinks -}}
+   {{ range $sinkattindex, $sinkatts := $sink -}}
+    {{- if and  (eq $sinkattindex "type") }}
+     {{- if eq $sinkatts "prometheus" }}
+      {{- if hasKey $sink "settings" }}
+       {{- if hasKey $sink.settings "autodiscovery" }}
+        {{- if eq $sink.settings.autodiscovery true }}
+        prometheusMetricsReporter: "yes"
```
jakesmith, line 1060: see prev. comments

helm/hpcc/templates/dali.yaml
```diff
@@ -49,8 +49,11 @@ spec:
 {{- if hasKey $dali "labels" }}
 {{ toYaml $dali.labels | indent 8 }}
 {{- end }}
-{{- if hasKey $dali "annotations" }}
       annotations:
+{{- if hasKey $.Values.global "metrics" }}
+{{ include "hpcc.addPrometheusScrapeAnnotations" $.Values.global.metrics }}
```
jakesmith, line 54: see prev. indent comment - this should indent, e.g.:
```
{{- include "hpcc.addPrometheusScrapeAnnotations" $.Values.global.metrics | nindent 8 -}}
```

helm/hpcc/templates/dali.yaml
```diff
@@ -49,8 +49,11 @@ spec:
 {{- if hasKey $dali "labels" }}
 {{ toYaml $dali.labels | indent 8 }}
 {{- end }}
-{{- if hasKey $dali "annotations" }}
       annotations:
+{{- if hasKey $.Values.global "metrics" }}
```
jakesmith, line 53: to avoid blank annotations: , perhaps this could be:
```
{{- if or (hasKey $dali "annotations") (hasKey $.Values.global "metrics") }}
      annotations:
```

helm/hpcc/templates/esp.yaml
```diff
@@ -100,11 +99,17 @@ spec:
         accessDali: "yes"
         app: {{ $application }}
         helmVersion: 8.3.0-trunk0
+{{- if hasKey $.Values.global "metrics" }}
+    {{ include "hpcc.generateMetricsReporterLabel" $.Values.global.metrics }}
```
jakesmith, line 103: should use | indent or | nindent

helm/hpcc/templates/network.yaml
```diff
@@ -155,3 +156,36 @@ spec:
...
+## Only created if global.metrics.sinks[type=prometheus].settings.autodiscovery true
+##
+{{- if hasKey $metrics "sinks" }}
+  {{ range $sink := $metrics.sinks -}}
+   {{ range $sinkattindex, $sinkatts := $sink -}}
+    {{- if and  (eq $sinkattindex "type") }}
+     {{- if eq $sinkatts "prometheus" }}
+      {{- if hasKey $sink "settings" }}
+       {{- if hasKey $sink.settings "autodiscovery" }}
+        {{- if eq $sink.settings.autodiscovery true }}
```
jakesmith, line 172: think can be simplified with pick and dig (see prev. comment)

helm/hpcc/templates/network.yaml
```diff
@@ -155,3 +156,36 @@ spec:
...
+kind: NetworkPolicy
+metadata:
+  name: hpccmetrics-prometheus
+spec:
+  podSelector:
+    matchLabels:
+      prometheusMetricsReporter: "yes"
+  ingress:
+  - ports:
+    - port: {{ default 8767 $sink.settings.port }}
```
jakesmith, line 183: I think this should be:
```
    - port: {{ $sink.settings.port | default 8767 }}
```

helm/hpcc/templates/roxie.yaml
```diff
@@ -219,6 +222,9 @@ spec:
 {{- end }}
       annotations:
         checksum/config: {{ $configSHA }}
+{{- if hasKey $.Values.global "metrics" }}
+{{ include "hpcc.addPrometheusScrapeAnnotations" $.Values.global.metrics }}
```
jakesmith, line 226: should use | indent or | nindent

helm/hpcc/templates/roxie.yaml
```diff
@@ -310,6 +316,9 @@ spec:
...
```
jakesmith, line 320: should use | indent or | nindent

jakesmith (changes requested):
@rpastrana - looks good, main issue is the indentation issue, but looks like 'dig' would be particularly good and simplifying the template code here.

helm/examples/metrics/README.md
```diff
@@ -20,3 +20,37 @@ For example, the following adds the Prometheus configuration as the collection s
...
+        action: replace
+        regex: ([^:]+)(:\d+)?;(\d+)
+        replacement: ${1}:${3}
+        target_label: __address__
```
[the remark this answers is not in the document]
rpastrana: Good idea, but the Prometheus configuration process is a little more involved than I think we want document.

helm/hpcc/templates/_helpers.tpl
rpastrana: That's much cleaner, thanks for the pointers. 
The only problem is dig apparently doesn't show until more recent versions of HELM (I'm on v3.2.4 and apparently dig is not available).
Do we have a stated minimum helm version supported?
Updated helm and applied dig successfully

helm/hpcc/templates/esp.yaml
```diff
@@ -100,11 +99,18 @@ spec:
         accessDali: "yes"
         app: {{ $application }}
         helmVersion: 8.3.0-trunk0
+{{- if hasKey $.Values.global "metrics" }}
```
rpastrana, line 102: This should be based on the metrics block pontentially created on line 38: https://github.com/hpcc-systems/HPCC-Platform/blob/7cf6eff4459add2daed0091b5838eb1e32690747/helm/hpcc/templates/esp.yaml#L38
How to reference generated output from template call?

jakesmith: you can use fromYaml to get back the dict from what's generated, e.g:
```
$metrics := {{ fromYaml (include generateMetricsConfig  .) }}
```
you might need a default dict after it, to cope with it not generating anything, e.g.:
```
$metrics := {{ fromYaml (include generateMetricsConfig  .) | default dict }}
```
(would need testing)

helm/hpcc/templates/dali.yaml
```diff
@@ -46,12 +46,20 @@ spec:
         server: {{ $dali.name | quote }}
         app: dali
         helmVersion: 8.3.0-trunk0
+{{- if hasKey $.Values.global "metrics" }}
```
jakesmith, line 49: does this and other places that reference hasKey $.Values.global, need to use the merged metrics from generateMetricsConfig too? (see other comment re. fromYaml).
If so, you could set $mymetrics once and reference elsewhere within the template.

helm/hpcc/templates/network.yaml
```diff
@@ -23,6 +23,7 @@
 */}}
 {{- $egress := (.Values.global.egress | default dict) -}}
 {{- $egressRestricted := $egress.restricted | default false -}}
+{{- $metrics := .Values.global.metrics }}
```
jakesmith, line 26: wondering how this works if individual components have overriden some settings, resulting in different metrics per component (i.e. the result of generateMetricsConfig) ?

helm/hpcc/templates/roxie.yaml
```diff
@@ -109,6 +109,9 @@ spec:
 {{- end }}
       annotations:
         checksum/config: {{ $topoconfigSHA }}
+{{- if hasKey $.Values.global "metrics" }}
+{{ include "hpcc.addPrometheusScrapeAnnotations" $.Values.global.metrics | indent 8 }}
```
jakesmith, line 113: trivial: we're trying to indent the template blocks (1 space) for clarity, e.g.:
```
{{- if hasKey $.Values.global "metrics" }}
 {{- include "hpcc.addPrometheusScrapeAnnotations" $.Values.global.metrics | nindent 8 }}
{{- end }}
```

helm/hpcc/templates/roxie.yaml
```diff
@@ -219,6 +222,9 @@ spec:
 {{- end }}
       annotations:
         checksum/config: {{ $configSHA }}
+{{- if hasKey $.Values.global "metrics" }}
+{{ include "hpcc.addPrometheusScrapeAnnotations" $.Values.global.metrics | indent 8 }}
```
jakesmith, line 226: indentation as per last comment

helm/hpcc/templates/roxie.yaml
```diff
@@ -310,6 +316,9 @@ spec:
...
```
jakesmith, line 320: indentation as per last comment

helm/hpcc/templates/_helpers.tpl
jakesmith: hm, I'm not sure re. minimal version. I didn't know that dig was relatively new.
Unfortunately, I think it may be too new, and we should avoid for now. I say that, because I just checked what version comes by default in the Azure portal shell, and it's v3.4.0 that doesn't support dig :(

rpastrana: Agreed. It's too bad, I like the streamlined optics of the function. Will revert.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -429,6 +429,23 @@ Pass in dict with root and secretsCategories
...
+Requires metrics.sinks[type=prometheus]
+Pass in metric config
+*/}}
+{{- define "hpcc.addPrometheusScrapeAnnotations" -}}
+{{ range $sink := .sinks -}}
+ {{- if eq (get $sink "type") "prometheus" }}
+ {{- if (dig "settings" "autodiscovery" false $sink) }}
+prometheus.io/scrape: 'true'
+prometheus.io/path: {{ default "/metrics" $sink.settings.path }}
+prometheus.io/port: {{ default 8767 $sink.settings.port | quote }}
```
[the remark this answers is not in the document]
rpastrana: I believe it does work (fairly certain I explicitly tested it) but I don't mind re-arraigning if nothing else for the sake of consistency w/ pre-existing code...

helm/hpcc/templates/dali.yaml
rpastrana: It doesn't appear dali nor roxie merge global.metrics into local metrics.

helm/hpcc/templates/esp.yaml
rpastrana: I'm now referencing the "local" metrics. However, it appears the schema doesn't currently allow metrics configuration outside of global, adding temporary values for illustration...

helm/hpcc/templates/roxie.yaml
rpastrana: Indeed. I attempted to create a dedicated networkpolicy per pod which declares itself prometheusReporter. However it seems (for now) per component configuration isn't allowed (per schema)

helm/hpcc/templates/roxie.yaml
rpastrana: I actually tried to follow existing convention, see L:220-222 just above

jakesmith: right, it's a bit of a inconsistent mess at the moment. But we agreed that indenting is clearer, so for new or modified code we should indent blocks.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -429,6 +429,85 @@ Pass in dict with root and secretsCategories
...
+    {{ end }}
+   {{ end }}
+  {{ end }}
+ {{ end}}
+{{ end}}
+{{- end -}}
+
+
+{{/*
+Generate NetworkPolicy exposing if global.metrics.sink[type=prometheus].settins.port differs from
```
rpastrana, line 458: Odd function... the concept is, a networkpolicy is created for global.metrics.sinks[type=prometheus].settings.port.
If <component>.metrics.sinks[type=prometheus].settings.port exists and differs, create dedicated networkpolicy. The podSelector likely should change.

jakesmith (approved):
@rpastrana - 2 trivial formatting issues, looks good, ready to merge once squashed.

richardkchapman (approved):

PR #15254, opened 2021-08-12
HPCC-26383 Add utility classes for tracking blocked time

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jdebug.cpp
+++ b/system/jlib/jdebug.cpp
@@ -645,6 +645,56 @@ MODULE_EXIT()
 
 //===========================================================================
 
+
+void BlockedTimeTracker::noteWaiting()
+{
+    CriticalBlock block(cs);
+    numWaiting++;
+    timeStampTally -= get_cycles_now();
+}
+
+void BlockedTimeTracker::noteComplete()
+{
+    CriticalBlock block(cs);
+    numWaiting--;
+    timeStampTally += get_cycles_now();
+}
+
+__uint64 BlockedTimeTracker::getWaitingNs() const
+{
+    unsigned active;
+    cycle_t tally;
+    {
+        CriticalBlock block(cs);
+        active = numWaiting;
+        tally = timeStampTally;
+    }
+
+    if (active != 0)
+    {
+        cycle_t now = get_cycles_now();
+        tally += active * now;
+    }
+
+    return cycle_to_nanosec(tally);
+}
+
+
+__uint64 LightweightBlockedTimeTracker::getWaitingNs() const
+{
+    __uint64 tally = timeStampTally.load(std::memory_order_acquire); // read the value once atomically
+    unsigned active = tally % MAX_ACTIVE;
+    if (active != 0)
+    {
+        cycle_t now = get_cycles_now();
+        tally += active * now * MAX_ACTIVE;
+    }
+
+    return cycle_to_nanosec(tally / MAX_ACTIVE);
+}
+
+//===========================================================================
+
 #ifdef _WIN32
 
 typedef enum _PROCESSINFOCLASS {
--- a/system/jlib/jdebug.hpp
+++ b/system/jlib/jdebug.hpp
@@ -22,6 +22,7 @@
 
 #include "jiface.hpp"
 #include "jstats.h"
+#include <atomic>
 
 #define TIMING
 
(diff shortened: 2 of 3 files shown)


ghalliday:
@kenrowland I think this is likely to be useful when you add a request blocked time to dali.

system/jlib/jdebug.hpp
```diff
@@ -225,6 +226,107 @@ class jlib_decl HiresTimer
...
+private:
+    BlockedTimeTracker & tracker;
+};
+
+//Lightweight version that uses a single atomic, but has a limit on the number of active blocked items
+//Easier to understand by looking at the code for the BlockedTimeTracker class
+//Could be a template class (for MAX_ACTIVE), but I doubt it is worth it.
+class jlib_decl LightweightBlockedTimeTracker
+{
+    static constexpr unsigned MAX_ACTIVE = 64;  //MAX_ACTIVE should be a power of 2 for efficiency
```
richardkchapman, line 289: At 2^32 cycles/sec (approx) you can measure about 2^24 seconds before you start to get issues with loss of resolution, I think... 

If so MAX_ACTIVE could safely be a bit higher, I suspect.

richardkchapman:
Very clever, and looks like it should work. But I'm still not 100% sure I understand where we would use it or what it represents.

system/jlib/jdebug.hpp
ghalliday: I think it currently gives about 2 years accumulated blocked time.  Hard to know at the moment what a sensible value would be.

ghalliday:
For usage, the immediate example is for processing one type of dali messages.  When they are received they are added to a queue and then processed sequentially form that queue.  If noteWaiting() is called when they are received, and noteComplete() when they are read off the queue this class will provide the total time those messages spend waiting before they are processed.

richardkchapman:
64 feels a little short for the upper limit of a queue. Though I guess in the dali queue case you may use the heavier one

kenrowland:
Just a general comment that these classes can be used to track time for more than just blocked items. Any reason it should not be used for processing time as well?

ghalliday:
Yes they can be used for any elapsed time.  They are particular interesting for times that are commonly not completed at the time the elapsed time is requested.  @jakesmith They may be particularly relevant to Thor for monitoring time waiting for responses from other nodes.  I did wonder about renaming, but decided to leave as it is because that is the most common case.

ghalliday:
It may well be worth introducing a single threaded version that could avoid the atomic operations, but something for another day if we have a use.

system/jlib/jdebug.cpp
```diff
@@ -643,6 +643,56 @@ MODULE_EXIT()
 }
 
 
+//===========================================================================
+
+
+void BlockedTimeTracker::noteWaiting()
```
jakesmith, line 649: is jtime.* a better place for these?

system/jlib/jdebug.hpp
```diff
@@ -225,6 +226,108 @@ class jlib_decl HiresTimer
...
+private:
+    LightweightBlockedTimeTracker & tracker;
+};
+
+
+// Utility class and function for performing an action when the object is destroyed.
+// Example of use:
+// auto action = OnLeave([&](){ timer.noteComplete(); });
+template <class DESTRUCTOR>
+class OnLeaveAction
```
jakesmith, line 317: This is very similar to the existing COnScopeExit.

jakesmith (commented):
Nice.
A couple of auxillary comments.

system/jlib/jdebug.cpp
kenrowland: I kind of agree with Jake since these classes are more than likely to be used by metrics to measure latency and not in a debug only mode.

system/jlib/jdebug.hpp
ghalliday: Yes, that is doing the same - I will remove mine.  How do we make it clear where classes like that can be found.

system/jlib/jdebug.cpp
ghalliday: I'm not sure.

It is in jdebug because that is where the other classes for measuring times currently live - TimeSection, CycleTimer etc.

jdebug isn't a great name - since most (none?) of the functions don't have much to do with debugging.  They could all move to jtime, or create a new jtiming.hpp.  For the moment I would leave here.

ghalliday:
@richardkchapman removed duplicate class and switched unit test to use COnScopeExit instead.

ghalliday:
@richardkchapman I think this is now ready to merge

richardkchapman (approved):

PR #15257, opened 2021-08-12
HPCC-26281 First pass of meta data and units

kenrowland:
Naming support added to metrics framework

Singed-off-by: Ken Rowland kenneth.rowland@lexisnexisrisk.com

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/devdoc/Metrics.rst
+++ b/devdoc/Metrics.rst
@@ -328,6 +328,50 @@ metrics.sinks[].settings
     necessary for the operation of the sink. Nested YML is supported. Example settings are the
     prometheus server name, or the collection period for a periodic sink.
 
+*************
+Metric Naming
+*************
+
+Metric names shall follow a convention as outlined in this section. Because different collection systems
+have different requirements for how metric value reports are generated, naming is split into two parts.
+
+First, each metric is given a base name that describes what the underlying value is. Second, meta data
+is assigned to each metric to further qualify the value. For example, a set of metrics may count the
+number of requests a component has received. Each metric would have the same base name, but meta data would
+separate types of request (GET vs POST), or disposition such as pass or fail.
+
+Base Name
+=========
+The following convention defines how metric names are formed:
+
+* Names consist of parts separated by a period (.)
+* Each part shall use snake case (allows for compound names in each part)
+* Names for metric types shall be named as follows:
+
+  Gauges: <plural-noun>.<state>   requests.waiting, status_requests.waiting
+
+  Counters:  <plural-noun>.<past-tense-verb>   requests.failed, gateway_requests.queued
+
+  Time:    <singular-noun>.<state or active-verb>.time  request.blocked.time, request.process.time
+
+Meta Data
+=========
+Meta data further qualifies a metric value. This allows metrics to have the same name, but different scopes
+or categories. Generally, meta data is only used to furher qualify metrics that would have the same base
+name, but need further distinction. An example best describes a use case for meta data. Consider a
+component that accepts HTTP requests, but needs to track GET and POST requests separately. Instead of
+defining metrics with names *post_requests.received* and *get_requests.received*, the component creates two
+metrics with the base name *requests.received* and attaches meta data describing the request type of
+POST to one and GET to the other.
+
+Use of meta data allows aggregating both types of requests into a single combined count of received
+requests while allowing a breakdown by type.
+
+Meta data is represented as a key/value pair and is attached to the metric by the component during
+metric creation. The sink is responsible for converting meta data into useful information for the
+collection system during reporting.
+
+The *Component Instrumentation* section covers how meta data is added to a metric.
 
 *************************
 Component Instrumentation
@@ -404,3 +448,38 @@ Where:
 
   A reference to the underlying event state which must be a scalar value convertable to a 64bit unsigned
   integer (__uint64)
+
+Adding Metric Meta Data
+=======================
+A component, depending on requirements, may attach meta data to further qualify created metrics.
+Meta data takes the form of key value pairs. The base metric class *MetricBase* constructor defines
+a parameter for a vector of meta data. Metric subclasses also define meta data as a constructor parameter,
+however an empty vector is the default. The *IMetric* interface defines a method for retrieving the meta data.
+
+Meta data is order dependent.
+
+Below are two examples of constructing a metric with meta data. One creates the vector and passes it as a
+parameter, the other constructs the vector in place.
+
+::
+
+    MetricMetaData metaData1{{"key1", "value1"}};
+    std::shared_ptr<CounterMetric> pCounter1 =
+        std::make_shared<CounterMetric>("requests.completed", "description", SMeasureCount, metaData1);
+
+    std::shared_ptr<CounterMetric> pCounter2 =
+        std::make_shared<CounterMetric>("requests.completed", "description", SMeasureCount, MetricMetaData{{"key1", "value2"}});
+
+Metric Units
+============
+Metric units are treated separately from the base name and meta data. The reason is to allow the sink to
+translate based on collection system requirements. The base framework provides a convenience method for
+converting units into a string. However, the sink is free to do any conversions, both actual units and
+the string representation, as needed.
+
+Metric units are defined using a subset of the *StaticsMeasure*  enumeration values defined
+in **jstatscodes.h**. The current values are used:
+
+* SMeasureTimeNs - A time measurement in nanoseconds
+* SMeasureCount - A count of events
+* SMeasureSize - Size in bytes
--- a/system/jlib/jmetrics.cpp
+++ b/system/jlib/jmetrics.cpp
@@ -33,6 +33,7 @@ MetricsManager &hpccMetrics::queryMetricsManager()
     return *metricsManager.query([] { return new MetricsManager; });
 }
 
+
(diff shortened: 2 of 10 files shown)


kenrowland:
@ghalliday, @rpastrana - Please review. I'm sure there will be some terminology changes.

system/jlib/jmetrics.hpp
```diff
@@ -44,6 +45,15 @@ enum MetricType
 };
 
 
+struct MetricMetaData
```
ghalliday, line 48: naming: Should this be named MetricMetaDataEntry/Item?  

Possibly
`    using MetricMetaData = std::vector<MetricMetaDataEntry>;
`

ghalliday (commented):
@kenrowland generally looks ok - a few comments

devdoc/Metrics.rst
```diff
@@ -404,3 +448,38 @@ Where:
...
+============
+Metric units are treated separately from the base name and meta data. The reason is to allow the sink to
+translate based on collection system requirements. The base framework provides a convenience method for
+converting units into a string. However, the sink is free to do any conversions, both actual units and
+the string representation, as needed.
+
+Metric units are defined using a subset of the *StaticsMeasure*  enumeration values defined
+in **jstatscodes.h**. The current values are used:
+
+* SMeasureNone - No defined units, name or meta data should contain relevant information
```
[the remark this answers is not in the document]
kenrowland: I cannot provide an immediate example, however I wanted to give the component the option to provide the information another way and keep the sink for automatically adding the units to the report name. We can remove for now and add it back should the need arise. That is what I will do.

A follow-on question is whether we should enforce setting units to one of a predefined set. Thoughts?

system/jlib/jmetrics.cpp
```diff
@@ -50,28 +51,40 @@ void MetricsManager::init(IPropertyTree *pMetricsTree)
...
+    bool rc = false;
     std::unique_lock<std::mutex> lock(metricVectorMutex);
-    auto it = metrics.find(pMetric->queryName());
+    std::string name = pMetric->queryName();
+    auto metaData = pMetric->queryMetaData();
+    for (auto &metaDataIt: metaData)
+    {
+        name.append(".").append(metaDataIt.value);
+    }
+    name.append(queryUnitsString(pMetric->queryUnits(), "."));
```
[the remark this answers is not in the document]
kenrowland: Removed.

kenrowland:
@ghalliday @rpastrana Please see changes

testing/unittests/metrics/MetricFrameworkTests.cpp
```diff
@@ -209,13 +210,62 @@ class MetricFrameworkTests : public CppUnit::TestFixture
             ScopedGaugeDecrementer gaugeDecrementer(*pGauge, 1);
         }
         CPPUNIT_ASSERT_EQUAL(0, static_cast<int>(pGauge->queryValue()));
+    }
+
+    void Test_metric_meta_data()
+    {
+        MetricMetaData data1{MetricMetaDataItem{"key1", "value1"}};
```
ghalliday, line 217: minor: the keyword MetricMetaDataItem can be removed.

`MetricMetaData data1{{"key1", "value1"}};`

main reason for mentioning it is that this code is likely to serve as an example of how it should be called.

You can also include it directly on the call:

`std::make_shared<CounterMetric>("requests.completed", "description", SMeasureCount, MetricMetaData{{"key1", "value1"}})
`

but for some reason the MetricMetaData cannot be removed (even though it can be removed from a call to the constructor.  Looks like an oddity of template parameter forwarding.

system/jlib/jmetrics.hpp
```diff
@@ -114,26 +107,25 @@ class jlib_decl MetricBase : public IMetric
...
-    void setMetaData(const std::vector<MetricMetaData> &_metaData) { metaData = _metaData; }
-    const std::vector<MetricMetaData> &queryMetaData() const { return metaData; }
+    const MetricMetaData &queryMetaData() const { return metaData; }
     StatisticMeasure queryUnits() const override { return units; }
 
 
 protected:
     // No one should be able to create one of these
-    MetricBase(const char *_name, const char *_desc, MetricType _metricType, StatisticMeasure _units) :
+    MetricBase(const char *_name, const char *_desc, MetricType _metricType, StatisticMeasure _units, const MetricMetaData &_metaData) :
```
ghalliday, line 116: I slightly wonder about the efficiency of this - e.g. whether there should be a version with a && parameter.  I think leave for the moment, but might be worth revisiting at some point.

ghalliday (commented):
@kenrowland a few more minor comments.

kenrowland (commented):
@ghalliday @rpastrana Please changes from review.

system/jlib/jmetrics.cpp
```diff
@@ -50,28 +60,39 @@ void MetricsManager::init(IPropertyTree *pMetricsTree)
...
         {
 #ifdef _DEBUG
-            throw MakeStringException(MSGAUD_operator, "addMetric - Attempted to add duplicate named metric with name '%s'", pMetric->queryName().c_str());
+            throw MakeStringException(MSGAUD_operator, "addMetric - Attempted to add duplicate named metric with name '%s'", name.c_str());
 #else
-            OERRLOG("addMetric - Adding a duplicate named metric '%s', old metric replaced", pMetric->queryName().c_str());
+            OERRLOG("addMetric - Adding a duplicate named metric '%s', old metric replaced", name.c_str());
 #endif
         }
         it->second = pMetric;
```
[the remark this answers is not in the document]
ghalliday: If it does return a value, then it should be clearly defined.  In this case it should be the same as whether an exception would not be thrown.  That means it should also set it to true if match is non null (otherwise your regression tests would fail in release) - and should be extended to cover this case.

system/jlib/jmetrics.cpp
```diff
@@ -212,6 +223,18 @@ void MetricsManager::addSink(MetricSink *pSink, const char *name)
 }
 
 
+const char *MetricsManager::queryUnitsString(StatisticMeasure units) const
+{
+    switch (units)
+    {
+        case SMeasureCount:  return "count";
```
ghalliday, line 230: style: we don't generally indent case statements within a switch.

system/jlib/jmetrics.hpp
```diff
@@ -225,14 +255,15 @@ struct SinkInfo
...
+    MetricsManager() {}
     ~MetricsManager();
     void init(IPropertyTree *pMetricsTree);
     void addSink(MetricSink *pSink, const char *name);  // for use by unit tests
-    void addMetric(const std::shared_ptr<IMetric> &pMetric);
+    bool addMetric(const std::shared_ptr<IMetric> &pMetric);
     void startCollecting();
     void stopCollecting();
     std::vector<std::shared_ptr<IMetric>> queryMetricsForReport(const std::string &sinkName);
+    const char * queryUnitsString(StatisticMeasure units) const;
```
ghalliday, line 266: minor: can become a global function.

testing/unittests/metrics/MetricFrameworkTests.cpp
```diff
@@ -209,13 +210,62 @@ class MetricFrameworkTests : public CppUnit::TestFixture
...
+        //
+        // Test that two metrics with the same base name and meta data fail adding. Again note both methods for passing meta data
+        MetricMetaData metaData2{{"key1", "value1"}};
+        std::shared_ptr<CounterMetric> pCounter3 = std::make_shared<CounterMetric>("requests.queued", "description", SMeasureCount, metaData2);
+        std::shared_ptr<CounterMetric> pCounter4 = std::make_shared<CounterMetric>("requests.queued", "description", SMeasureCount,
+                                                                                   MetricMetaData{{"key1", "value1"}});
+
+        frameworkTestManager.addMetric(pCounter3);
+
+        bool exceptionThrown = false;
```
ghalliday, line 248: Simpler and more consistent would be:

```
        bool success = true;
        try
        {
            // Expect a return of false if not debug
            success = frameworkTestManager.addMetric(pCounter4);
        }
        catch (IException * e)
        {
            success = false;   // this is expected if built in debug mode
            e->Release();
        }

        CPPUNIT_ASSERT(!success);

```
It would also be worth adding another test:

`pCounter3.reset();
`
and then checking that adding the metric now succeeds.

ghalliday (commented):
@kenrowland a few (hopefully final) comments.  Looking close to ready.

testing/unittests/metrics/MetricFrameworkTests.cpp
kenrowland: Added a new test.

system/jlib/jmetrics.cpp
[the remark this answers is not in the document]
kenrowland: Added a comment that the return value indicates whether the name was unique when the metric was added.

ghalliday (approved):
@kenrowland please squash

kenrowland:
@ghalliday Please merge

ghalliday:
@rpastrana do you have any problems with this being merged?

rpastrana (approved):

PR #15263, opened 2021-08-15
HPCC-26395 require setting resources requests/limits explicitly.

xwang2713:
Previously only limits can be set. Now requests are added.

resources.limits will not guarantee a pod/job will get the required cpu or memory.
If memory allocation failed the pod/job will be in state of "OOMKilled" and will be restarted but normally it will fail again and again.
The fix is to ask user explicitly set "requests" or "limits". If resources.requestes is set the cpu/memory will be allocated when the pod/job is started. User will check with "kubectl describe node -o wide"

Signed off by Xiaoming Wang <xiaoming.wang@lexisnexis.com>

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

## Testing:

Tested on Azure and it works
On local even it help eliminate "OOMKilled" for thor-worker but the GNN training still fails when images are more than certain numbers such 350+. It cloud be related each PC resources.

### DIFF
--- a/helm/hpcc/templates/_helpers.tpl
+++ b/helm/hpcc/templates/_helpers.tpl
@@ -820,8 +820,7 @@ Pass in a dictionary with me defined
 {{- define "hpcc.addResources" }}
 {{- if .me  }}
 resources:
-  limits:
-{{ toYaml .me | indent 4 }}
+{{ toYaml .me | indent 2 }}
 {{- end }}
 {{- end -}}
 
--- a/helm/hpcc/values.yaml
+++ b/helm/hpcc/values.yaml
@@ -498,14 +498,26 @@ thor:
   maxJobs: 4
   maxGraphs: 2
   #managerResources:
-  #  cpu: "1"
-  #  memory: "2G"
+  #  requests:
+  #    cpu: "1"
+  #    memory: "2G"
+  #  limits:
+  #    cpu: "2"
+  #    memory: "4G"
   #workerResources:
-  #  cpu: "4"
-  #  memory: "4G"
+  #  requests:
+  #    cpu: "4"
+  #    memory: "4G"
+  #  limits:
+  #    cpu: "8"
+  #    memory: "8G"
   #eclAgentResources:
-  #  cpu: "1"
-  #  memory: "2G"
+  #  requests:
+  #    cpu: "1"
+  #    memory: "2G"
+  #  limits:
+  #    cpu: "2"
+  #    memory: "4G"
 
 eclscheduler:
 - name: eclscheduler


jakesmith (commented):
@xwang2713 - not convinced this is right/needed.

We intentionally want limits == request, and intentionally only generate limits, which should set request to same.
```
Note: If a Container specifies its own memory limit, but does not specify a memory request, Kubernetes automatically assigns a memory request that matches the limit. Similarly, if a Container specifies its own CPU limit, but does not specify a CPU request, Kubernetes automatically assigns a CPU request that matches the limit.
```
from https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/

I've verified on my system and requests do auto match the generated limits, e.g. Dali:

  Namespace                   Name                                        CPU Requests  CPU Limits  Memory Requests  Memory Limits  Age
  ---------                   ----                                        ------------  ----------  ---------------  -------------  ---
  default                     mydali-765d64b49d-dd8vj                     1 (6%)        1 (6%)      1G (4%)          1G (4%)        52s

xwang2713:
@Jake, I still think we should allow explicitly set "requests" and "limits" in resources. The reasons:
1) Kubernetes define "requests" and "limits" with reasons. Otherwise it only need have "limits"
2) It is pretty much up to cloud provider to implement Kubernetes spec such as resource "requests" and "limits".  Also user may need only set "requests" instead of "limits". It could be only for debug or explore the application and the memory usage. For example in Central India set "requests: 8GB" works but not "limits: 8GB"
3) For future resources spec support. It could be new resources type introduced or behavior changed for "requests" and "limits".

All of these are just my opinions.

xwang2713:
I rerun 2 thor workers on centralindia
1) with set only resources.limits: cpu 4  memory 8GB it did allocate 4 cpu and memory 8GB and limits are set also from "kubectl describe node <node id>". But thorworkers result "OOMKilled"
2) with set resources.requests: cpu 4  memory 8GB 
```code
default                     thorworker-w20210821-215012-graph3-bsz2t     4 (25%)       0 (0%)      8G (13%)         0 (0%)         4m10s
```
The trainning run OK:
```code
ming@focal-1:~/work/HPCC/HPCC-Tests/aks/Azure/Kubernetes/default-spot-gnn$ kubectl logs thorworker-w20210821-215012-graph3-bsz2t | grep ====
1/1 [==============================] - 10s 10s/step - batch: 0.0000e+00 - size: 1.0000 - loss: 0.9127 - binary_accuracy: 0.0000e+00
1/1 [==============================] - 8s 8s/step - batch: 0.0000e+00 - size: 1.0000 - loss: 0.6379 - binary_accuracy: 0.7714
1/1 [==============================] - 8s 8s/step - batch: 0.0000e+00 - size: 1.0000 - loss: 0.7757 - binary_accuracy: 0.4571
1/1 [==============================] - 8s 8s/step - batch: 0.0000e+00 - size: 1.0000 - loss: 0.9784 - binary_accuracy: 0.6429
1/1 [==============================] - 8s 8s/step - batch: 0.0000e+00 - size: 1.0000 - loss: 0.7012 - binary_accuracy: 0.6571
1/1 [==============================] - 7s 7s/step - batch: 0.0000e+00 - size: 1.0000 - loss: 0.7301 - binary_accuracy: 0.3619
1/1 [==============================] - 8s 8s/step - batch: 0.0000e+00 - size: 1.0000 - loss: 0.4759 - binary_accuracy: 0.9381
1/1 [==============================] - 8s 8s/step - batch: 0.0000e+00 - size: 1.0000 - loss: 0.4770 - binary_accuracy: 0.7143
1/1 [==============================] - 8s 8s/step - batch: 0.0000e+00 - size: 1.0000 - loss: 0.5984 - binary_accuracy: 0.4667
1/1 [==============================] - 8s 8s/step - batch: 0.0000e+00 - size: 1.0000 - loss: 0.2469 - binary_accuracy: 1.0000
1/1 [==============================] - 7s 7s/step - batch: 0.0000e+00 - size: 1.0000 - loss: 0.2958 - binary_accuracy: 0.7905
1/1 [==============================] - 7s 7s/step - batch: 0.0000e+00 - size: 1.0000 - loss: 0.1499 - binary_accuracy: 1.0000
```

jakesmith:
>with set resources.requests: cpu 4 memory 8GB

From your output:
{noformat}
Memory Requests = 8G
Memory Limits = 0 (unlimited)
{noformat}

Because you've changed it to request and limit of zero. Meaning k8s can resource it, but it's not binding the pod to any limit, i.e. it can use as much as it likes beyond 8G, e.g. 23GB as one of your prev. outputs showed:
```
Warning  Evicted              9m9s   kubelet            The node was low on resource: memory. Container thorworker-w20210817-163652-graph3 was using 23567124Ki, which exceeds its request of 8G.
```

The unattributed 3rd party (Python/TensorFlow) memory usage isn't being account for.
It will "work" until the node if overcommitted, because it's been allowed to resource to a/the node due to requests only, but will fail and start evicting pods once the node actually runs out of memory.

We want pods to only use what they say they are going to use, hence it is deliberate that our memory limit = request, such that if a pod uses more than what it requested, it will fail, as opposed to using more, overcommitting and causing it and other random pods that have been allocated to the node to be evicted.

>with set only resources.limits: cpu 4 memory 8GB it did allocate 4 cpu and memory 8GB and limits are set also from "kubectl describe node ". But thorworkers resulted "OOMKilled" during the model traning

The issue is not that the limit should be removed, but that the configuration and pod should stay within a defined limit.
8G was not enough for this HPCC+TensorFlow setup. 
See fuller description of the problem in : https://track.hpccsystems.com/browse/HPCC-26394

jakesmith:
> @jake, I still think we should allow explicitly set "requests" and "limits" in resources. The reasons:
> 
> 1. Kubernetes define "requests" and "limits" with reasons. Otherwise it only need have "limits"
> 2. It is pretty much up to cloud provider to implement Kubernetes spec such as resource "requests" and "limits".  Also user may need only set "requests" instead of "limits". It could be only for debug or explore the application and the memory usage. For example in Central India set "requests: 8GB" works but not "limits: 8GB"
> 3. For future resources spec support. It could be new resources type introduced or behavior changed for "requests" and "limits".
> 
> All of these are just my opinions.

As discussed, it's "working" if you remove the limit completely, but that is because the config/memory usage is ill defined, and will lead to unstable system as it starts to overcommit.

For memory in HPCC, we want to ensure that it does not exceed the request, in other words we want the limit to be set and equal to the request.
Conceivably, for other resource types, it may make more sense to have request < limit (don't think it makes sense for cpu, because it's not dynamic like that).

I can imagine in other systems, it's more acceptable to over commit, because k8s has resourced according to the 'request', but pods have used over that request up to the 'limit', thereby causing pods to be evicted regularly as the nodes resources are dynamically exhausted. That's not what we want with HPCC.

In all the cases this has been an issue, it's because the request/limit hasn't been defined properly for the pod's actual usage - underlying issue = Python/TensorFlow using unbound amount of memory and in effect 0 given to it. As discussed in HPCC-26394, the fix is a) try to bound the amount of mem. it uses b) take account of HPCC memory *_and_* Python/TensorFlow memory in the defined limits.

Can discuss more with @richardkchapman when he's back, re. justification for request being different from limit in HPCC pod specs.

jakesmith (commented):
@richardkchapman - could you comment? See discussion above.

richardkchapman:
As far as I can see this change just allows people to set configurations that won't work. Not sure how that benefits anyone.

xwang2713:
Close this PR. If needed will re-address it in the future

PR #15265, opened 2021-08-16
HPCC-26280 Fix bug caused by inconsistant subfile count from ISuperFileDescriptor and IDistributedSuperFile

shamser:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/thorlcr/activities/fetch/thfetch.cpp
+++ b/thorlcr/activities/fetch/thfetch.cpp
@@ -75,12 +75,12 @@ public:
             }
             else if (encrypted)
                 throw MakeActivityException(this, 0, "File '%s' was published as encrypted but no encryption key provided", fetchFile->queryLogicalName());
-            IDistributedSuperFile *super = fetchFile->querySuperFile();
-            unsigned numsubs = super?super->numSubFiles(true):0;
+            ISuperFileDescriptor *superDesc = fileDesc->querySuperFileDescriptor();
+            unsigned numsubs = superDesc?superDesc->querySubFiles():0;
             for (unsigned i=0; i<numsubs; i++)
                 subFileStats.push_back(new CThorStatsCollection(diskReadActivityStatistics));
 
-            mapping.setown(getFileSlaveMaps(fetchFile->queryLogicalName(), *fileDesc, container.queryJob().queryUserDescriptor(), container.queryJob().querySlaveGroup(), container.queryLocalOrGrouped(), false, NULL, super));
+            mapping.setown(getFileSlaveMaps(fetchFile->queryLogicalName(), *fileDesc, container.queryJob().queryUserDescriptor(), container.queryJob().querySlaveGroup(), container.queryLocalOrGrouped(), false, NULL, fetchFile->querySuperFile()));
             mapping->serializeFileOffsetMap(offsetMapMb);
             addReadFile(fetchFile);
         }
--- a/thorlcr/activities/indexread/thindexread.cpp
+++ b/thorlcr/activities/indexread/thindexread.cpp
@@ -243,8 +243,11 @@ public:
                 if (super)
                 {
                     unsigned numSubFiles = super->numSubFiles();
-                    for (unsigned i=0; i<numSubFiles; i++)
-                        subIndexFileStats.push_back(new CThorStatsCollection(indexReadActivityStatistics));
+                    if (numSubFiles>1)
+                    {
+                        for (unsigned i=0; i<numSubFiles; i++)
+                            subIndexFileStats.push_back(new CThorStatsCollection(indexReadActivityStatistics));
+                    }
                 }
             }
             //MORE: Change index getFormatCrc once we support projected rows for indexes.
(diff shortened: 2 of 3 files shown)


thorlcr/activities/fetch/thfetch.cpp
```diff
@@ -75,8 +75,8 @@ class CFetchActivityMaster : public CMasterActivity
             }
             else if (encrypted)
                 throw MakeActivityException(this, 0, "File '%s' was published as encrypted but no encryption key provided", fetchFile->queryLogicalName());
-            IDistributedSuperFile *super = fetchFile->querySuperFile();
```
jakesmith, line 78: 'super' is needed by line 83 - will fail to compile without it.

thorlcr/activities/thdiskbase.cpp
```diff
@@ -80,6 +80,10 @@ void CDiskReadMasterBase::init()
         }
         if (0==(helper->getFlags() & TDXtemporary))
         {
+            // Use the querySubFiles() from ISuperFileDescriptor to ensure that the same subfile count as used in the slaves
+            // (fileDesc->querySuperFileDescriptor() returns nullptr if subfile==0, which is ok as some is used in slaves)
+            ISuperFileDescriptor *superDesc = fileDesc->querySuperFileDescriptor();
+            unsigned numsubs = superDesc?superDesc->querySubFiles():0;
```
jakesmith, line 86: this will workaround the serialization mismatch, but it will cause the code in CDiskReadMasterBase::done() to incorrectly set @numDiskReads into he super, instead of the 1 subfile.
In fact, CDiskReadMasterBase::done() should never set into the super (it will at moment if 0 subs).

jakesmith: It could be fixed by adding additional code to CDiskReadMasterBase::done() to check case where is super, but 1 sub, and set statsCollection.getStatisticSum(StNumDiskReads) into the sub instead..

But, I think the real bug is that the IFileDescriptor coming back from IDistributeFile for a super of 1 sub, is a IFileDescriptor instead of a ISuperFileDescriptor..

This is the code (in dadfs.cpp) which is causing it:
```
    virtual IFileDescriptor *getFileDescriptor(const char *clustername) override
    {
        CriticalBlock block (sect);
        if (subfiles.ordinality()==1)
            return subfiles.item(0).getFileDescriptor(clustername);
```

NB: it will still return a ISuperFileDescriptor if 0 subs.

I think that 'optimization' is wrong, and should be removed.
It would then fix the anomaly that's causing this bug (should still change code to avoid if 0 setting property into super, i.e. should never set into super)

thorlcr/activities/indexread/thindexread.cpp
```diff
@@ -243,8 +243,11 @@ class CIndexReadBase : public CMasterActivity
                 if (super)
                 {
                     unsigned numSubFiles = super->numSubFiles();
-                    for (unsigned i=0; i<numSubFiles; i++)
-                        subIndexFileStats.push_back(new CThorStatsCollection(indexReadActivityStatistics));
+                    if (numSubFiles>1)
+                    {
+                        for (unsigned i=0; i<numSubFiles; i++)
+                            subIndexFileStats.push_back(new CThorStatsCollection(indexReadActivityStatistics));
+                    }
```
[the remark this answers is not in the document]
shamser: Yes.  It is to make sure the subfile count was consistent with the slave indexread process.

thorlcr/activities/thdiskbase.cpp
shamser: Regarding updating superfile with one subfile: although it seems the superfile would be updated, it is actually updating the subfile. I have checked again and confirm this is the case:  queryFile(0) is returning the subfile rather than the superfile.

In CDiskReadSlaveActivityBase::init it was calling partDescs.item(0).queryOwner().querySuperFileDescriptor() and that returns nullptr when there is one subfile.  Does querySuperFileDescriptor() call getFileDescriptor?  Have .querySuperFileDescriptor() return a file descriptor even when the subfile.ordinality()==1 does fix this issue?  However, would this cause problems elsewhere or changes existing behavior?  There must have been a reason for doing this before.

shamser:
@jakesmith I'm tracing the code to see how/if querySuperFileDescriptor() makes use of getFileDescriptor() to see how your suggested change would work.  Please can you see my comments.

thorlcr/activities/thdiskbase.cpp
jakesmith: right, yes I see now CDiskReadMasterBase::done() is ok, and queryFile(0) is returning the sub, so that's what's updated in case of super with 1.

jakesmith: >Have .querySuperFileDescriptor() return a file descriptor even when the subfile.ordinality()==1 does fix this issue? However, would this cause problems elsewhere or changes existing behavior? There must have been a reason for doing this before.

it's wrong that 1 vs 0 vs >1 is treated differently by DFS getFileDescriptor.
You should get back a ISuperFileDescriptor so you can tell it's a super even if 1.
I suspect it was considered a shortcut to avoid processing/creating the ISuperFileDescriptor if only 1.

Removing the special-case for 1, does fix the issue (https://github.com/hpcc-systems/HPCC-Platform/pull/15271).

I've targeted 8.4(master), it may still be best to add the workaround to 8.2, just in case the DFS change exposes any other unknown bugs etc.

shamser:
This will be addressed by fixing dfs bug (getFileDescriptor() returns IFileDescriptor rather than ISuperFileDescriptor if the subfile count is 1).  Hence, closing this PR.

PR #15277, opened 2021-08-19
HPCC-26406 Scale stub memory resource by number of instances

jakesmith:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/helm/hpcc/templates/eclagent.yaml
+++ b/helm/hpcc/templates/eclagent.yaml
@@ -163,7 +163,7 @@ spec:
 {{- if .useChildProcesses }}
 {{- include "hpcc.addResources" (dict "me" .resources) | indent 8 }}
 {{- else if not $.Values.global.privileged }}
-{{- include "hpcc.addStubResources" . | indent 8 }}
+{{- include "hpcc.addStubResources" (dict "instances" .maxActive) | indent 8 }}
 {{- end }}
 {{ include "hpcc.addImageAttrs" $commonCtx | indent 8 }}
         volumeMounts:
--- a/helm/hpcc/values.schema.json
+++ b/helm/hpcc/values.schema.json
@@ -766,12 +766,18 @@
     },
     "eclccserver": {
       "type": "object",
-      "required": [ "name" ],
+      "required": [ "name", "maxActive" ],
       "additionalProperties": { "type": ["integer", "string", "boolean"] },
       "properties": {
         "name": {
           "type": "string",
           "description": "The name of the eclccserver process"
+        },
+        "maxActive": {
+          "type": "integer",
+          "minimum": 1,
+          "description": "Maximum number of compile jobs that can be active simultaneously",
+          "default": 4
         },
         "replicas": {
           "type": "integer"
@@ -909,6 +915,7 @@
     "eclagent": {
       "type": "object",
       "additionalProperties": { "type": ["integer", "string", "boolean"] },
+      "required": [ "maxActive" ],
       "properties": {
         "replicas": {
           "type": "integer"
--- a/testing/helm/tests/env.yaml
+++ b/testing/helm/tests/env.yaml
@@ -38,6 +38,7 @@ dfuserver:
     
 eclagent:
 - name: eclagent
+  maxActive: 1
   env:
   - name: ealv1
     value: ealv1v
@@ -58,6 +59,7 @@ thor:
 
 eclccserver: 
 - name: eclccserver
+  maxActive: 1
   env:
   - name: esav1
     value: esav1v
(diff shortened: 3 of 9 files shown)


helm/hpcc/templates/eclagent.yaml
```diff
@@ -163,7 +163,7 @@ spec:
 {{- if .useChildProcesses }}
 {{- include "hpcc.addResources" (dict "me" .resources) | indent 8 }}
 {{- else if not $.Values.global.privileged }}
-{{- include "hpcc.addStubResources" . | indent 8 }}
+{{- include "hpcc.addStubResources" (dict "instances" .maxActive) | indent 8 }}
```
ghalliday, line 166: What happens if maxActive is not defined?

jakesmith: good point, I meant default to 1

jakesmith: @ghalliday - max made maxActive properties required

helm/hpcc/values.schema.json
```diff
@@ -799,13 +799,19 @@
...
       "properties": {
         "name": {
           "type": "string",
           "description": "The name of the eclccserver process"
         },
+        "maxActive": {
+          "type": "integer",
+          "minimum": 1,
+          "description": "Maximum number of compile jobs that can be active simultaneously",
+          "default": 4
```
ghalliday, line 813: That does this do?

testing/helm/tests/env.yaml
```diff
@@ -58,6 +59,7 @@ thor:
 
 eclccserver: 
 - name: eclccserver
+  maxActive: 1
```
ghalliday, line 62: Previously this defaulted to 4.  Should discuss what the correct value is.

testing/helm/tests/env.yaml
```diff
@@ -38,6 +38,7 @@ dfuserver:
     
 eclagent:
 - name: eclagent
+  maxActive: 1
```
ghalliday, line 41: Previously defaulted to 100 (!?).  Should discuss what the correct value is.

ghalliday (commented):
@jakesmith change looks ok, but have a few questions, which are probably worth discussing with Richard.
I would be tempted to include in 8.2.x because it must be currently causing problems, but it also is a breaking change for existing values files (since maxActive is now required), so that wouldn't be a good idea.
There is a separate issue of what the defaults should be (which is probably related to auto-scaling)

helm/hpcc/values.schema.json
jakesmith: This mirrors the default in code (in eclccserver.cpp):
```
#ifdef _CONTAINERIZED
            queryCodeSigner().initForContainer();

            bool useChildProcesses = globals->getPropInt("@useChildProcesses", false);
            unsigned maxThreads = globals->getPropInt("@maxActive", 4);
#else

```

testing/helm/tests/env.yaml
jakesmith: as it's just a schema validation file, I thought 1 (min) was fine for this test.

testing/helm/tests/env.yaml
jakesmith: as it's just a schema validation file, I thought 1 (min) was fine for this test.

Shall I change to match defaults?

ghalliday: no, I didn't spot they were validation files - they're fine as they are.

helm/hpcc/values.schema.json
ghalliday: Does it do anything though (do some editors use it?)  Genuine question...

jakesmith: ah, no idea.

jakesmith:
> @jakesmith change looks ok, but have a few questions, which are probably worth discussing with Richard.
> I would be tempted to include in 8.2.x because it must be currently causing problems, but it also is a breaking change for existing values files (since maxActive is now required), so that wouldn't be a good idea.

Although it wasn't enforced in 8.2, maxActive is present in our stock values.yaml.
So yes, a custom values.yaml definition for eclagent or eclccserver that had maxActive missing, would now fail schema, but if they hadn't defined those, then it would continue to pass.

> There is a separate issue of what the defaults should be (which is probably related to auto-scaling)

Agreed, e.g. in-code 100 seems high for default for agentexec, however our stock values.yaml overrides it to 4.
As you say, slightly separate issue though.

ghalliday:
ok.  I suspect it is ok to merge as-is to 8.2.x, but will let @richardkchapman look at it first.

richardkchapman (approved):

PR #15291, opened 2021-08-25
HPCC-26422 Initial Dali metrics

kenrowland:
Added some initial metrics to Dali

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Cloud-compatibility
  - [x] Premature optimization

## Smoketest:

### DIFF
--- a/dali/base/danqs.cpp
+++ b/dali/base/danqs.cpp
@@ -28,19 +28,13 @@
 #include "mputil.hpp"
 #include "daserver.hpp"
 #include "danqs.hpp"
+#include "jmetrics.hpp"
 
 #ifdef _MSC_VER
 #pragma warning (disable : 4355)
 #endif
 
-// # HELP nq_requests The total number of Dali NQ requests handled
-// # TYPE nq_requests counter
-// Probably requires this counter to be held in a __int64 scalar.
-
-// # HELP nq_active_requests Current number of active NQ requests being handled.
-// # TYPE nq_active_requests gauge
-// A unsigned scalar should suffice, i.e. never going to have that many active transactions
-
+static auto pNqRequestsCount = hpccMetrics::registerCounterMetric("dali.nq_requests.received", "The total number of Dali NQ requests received", SMeasureCount);
 
 enum MQueueRequestKind { 
     MQR_ADD_QUEUE,
@@ -529,14 +523,6 @@ public:
         stopped = false;
         while (!stopped) {
             try {
-                // 1) Need to increment nq_requests here
-                // NB: it will never be decremented. This is total for life of this instance.
-
-                // 2) Need to increment nq_active_requests here
-                // and ensure it's scoped, such that it is guaranteed
-                // to decrement when processMessage() is complete.
-                // NB: NQ activeRequests are always synchronously handled by processMessage()
-
                 mb.clear();
                 if (coven.recv(mb,RANK_ALL,MPTAG_DALI_NAMED_QUEUE_REQUEST,NULL)) {
                     processMessage(mb); // synchronous to ensure queue operations handled in correct order
@@ -556,6 +542,8 @@ public:
 
     void processMessage(CMessageBuffer &mb)
     {
+        pNqRequestsCount->fastInc(1);
+
         // single threaded by caller
         ICoven &coven=queryCoven();
         mb.read(fn);
@@ -1010,5 +998,3 @@ IDaliServer *createDaliNamedQueueServer()
     daliNamedQueueServer = new CDaliNamedQueueServer();
     return daliNamedQueueServer;
 }
-
-
--- a/dali/base/dasds.cpp
+++ b/dali/base/dasds.cpp
@@ -31,6 +31,7 @@
 #include "jqueue.tpp"
 #include "dautils.hpp"
 #include "dadfs.hpp"
+#include "jmetrics.hpp"
 
 #define DEBUG_DIR "debug"
 #define DEFAULT_KEEP_LASTN_STORES 10 // should match default in dali.xsd
@@ -80,19 +81,10 @@ static unsigned readWriteTimeout = 60000;
 #define DEFAULT_LCIDLE_RATE 1          // 1 write transactions per idle period. <= this rate is deemed idle (suitable for save)
 #define STORENOTSAVE_WARNING_PERIOD 72 // hours
 
-static unsigned msgCount=(unsigned)-1;
-
-// # HELP sds_requests The total number of Dali SDS requests handled
-// # TYPE sds_requests counter
-// Probably requires this counter to be held in a __int64 scalar.
-
-// # HELP sds_active_requests Current number of active SDS requests being handled.
-// # TYPE sds_active_requests gauge
-// A unsigned scalar should suffice, i.e. never going to have that many active transactions
-
-// # HELP sds_pending_requests Current number of pending SDS requests.
-// # TYPE sds_pending_requests gauge
-// A unsigned scalar should suffice, i.e. never going to have that many pending transactions
+static auto pSdsRequestsReceived = hpccMetrics::registerCounterMetric("dali.sds_requests.received", "The total number of Dali SDS requests received", SMeasureCount);
+static auto pSdsRequestsStarted = hpccMetrics::registerCounterMetric("dali.sds_requests.started", "The total number of Dali SDS requests started", SMeasureCount);
+static auto pSdsRequestsCompleted = hpccMetrics::registerCounterMetric("dali.sds_requests.completed", "The total number of Dali SDS requests completed", SMeasureCount);
+static auto pSdsRequestsPending = hpccMetrics::registerGaugeFromCountersMetric("dali.sds_requests.pending", "Current number of pending SDS requests", SMeasureCount, pSdsRequestsReceived, pSdsRequestsStarted);
 
 // #define TEST_NOTIFY_HANDLER
 
@@ -3747,15 +3739,7 @@ int CSDSTransactionServer::run()
             mb.clear();
             if (coven.recv(mb, RANK_ALL, MPTAG_DALI_SDS_REQUEST, NULL))
             {
-                // 1) Need to increment sds_requests here
-                // NB: it will never be decremented. This is total for life of this instance.
-                
-                // 2) Need to increment sds_active_requests here
-                // and ensure it's scoped, such that it is guaranteed
-                // to decrement when handling is complete.
-                // NB: most transactions are handled asynchronously, so decrement will need to happen on another thread
-
-                msgCount++;
+                pSdsRequestsReceived->fastInc(1);
                 try
                 {
                     SdsCommand action;
@@ -3778,11 +3762,6 @@ int CSDSTransactionServer::run()
                         case DAMP_SDSCMD_GETELEMENTSRAW:
                         case DAMP_SDSCMD_GETCOUNT:
                         {
-                            // 1) Need to increment sds_pending_requests here.
-                            // NB: pending requests are those received, but not yet being handled (not active)
-                            // e.g. because thread pool limits are blocking them.
-                            // sds_pending_requests should be decremented, when the transactions starts in processMessage()
-
                             mb.reset();
                             handler.handleMessage(mb);
                             mb.clear(); // ^ has copied mb
@@ -4091,12 +4070,9 @@ bool translateOldFormat(CServerRemoteTree *parentServerTree, IPropertyTree *pare
 
 void CSDSTransactionServer::processMessage(CMessageBuffer &mb)
 {
-    // 1) Decrement sds_pending_requests here.
-    // NB: request has started, no longer pending
-    // See CSDSTransactionServer::run() for where needs to be incremented. 
-
-    // 2) Ensure sds_active_requests is decremented, when processMessage() is finished
-    // e.g. use a scoped object here, and decrement in dtor.
+    pSdsRequestsStarted->inc(1);
+    // Ensure that number of completed requests is incremented when the function completes.
+    COnScopeExit incCompletedOnExit([&](){ pSdsRequestsCompleted->inc(1); });
 
     TimingBlock xactTimingBlock(xactTimingStats);
     ICoven &coven = queryCoven();
--- a/dali/server/daserver.cpp
+++ b/dali/server/daserver.cpp
@@ -25,6 +25,7 @@
 #include "jptree.hpp"
 #include "jmisc.hpp"
 #include "jutil.hpp"
+#include "jmetrics.hpp"
 
 #include "mpbase.hpp"
 #include "mpcomm.hpp"
@@ -53,8 +54,6 @@
 #ifndef _NO_LDAP
 #include "daldap.hpp"
 #endif
-
-
 
 Owned<IPropertyTree> serverConfig;
 static IArrayOf<IDaliServer> servers;
@@ -400,6 +399,21 @@ dali:
   dataPath: "/var/lib/HPCCSystems/dalistorage"
 )!!";
 
+//
+// Initialize metrics
+static void initializeMetrics(IPropertyTree *pConfig)
+{
+    //
+    // Initialize metrics if present
+    Owned<IPropertyTree> pMetricsTree = pConfig->getPropTree("metrics");
+    if (pMetricsTree != nullptr)
+    {
+        PROGLOG("Metrics initializing...");
+        hpccMetrics::MetricsManager &metricsManager = hpccMetrics::queryMetricsManager();
+        metricsManager.init(pMetricsTree);
+        metricsManager.startCollecting();
+    }
+}
 
 int main(int argc, const char* argv[])
 {
@@ -751,6 +765,7 @@ int main(int argc, const char* argv[])
         }
         if (ok) {
             writeSentinelFile(sentinelFile);
+            initializeMetrics(serverConfig);
             covenMain();
             removeAbortHandler(actionOnAbort);
         }
--- a/esp/platform/espprotocol.cpp
+++ b/esp/platform/espprotocol.cpp
@@ -29,11 +29,8 @@
 
 static RelaxedAtomic<unsigned> gActiveRequests;
 
-static auto pActiveRequests = hpccMetrics::createCustomMetricAndAddToManager("activerequests",
-                                                                             "Number of active requests",
-                                                                             hpccMetrics::METRICS_GAUGE,
-                                                                             gActiveRequests,
-                                                                             SMeasureCount);
+static auto pActiveRequests = hpccMetrics::registerCustomMetric("esp.requests.active", "Number of active requests",
+                                                                hpccMetrics::METRICS_GAUGE, gActiveRequests, SMeasureCount);
 
 typedef IXslProcessor * (*getXslProcessor_func)();
 
--- a/esp/platform/txsummary.cpp
+++ b/esp/platform/txsummary.cpp
@@ -24,9 +24,7 @@
 
 #define MATCH_ENTRY [&](const EntryValue& e) {return strieq(e.get()->name, pathPart);}
 
-static auto pRequestCount = hpccMetrics::createMetricAndAddToManager<hpccMetrics::CounterMetric>("requests",
-                                                                                                 "Number of Requests",
-                                                                                                 SMeasureCount);
+static auto pRequestCount = hpccMetrics::registerCounterMetric("esp.requests.received", "Number of requests received", SMeasureCount);
 
 inline bool validate(const char* k)
 {
--- a/system/jlib/jmetrics.hpp
+++ b/system/jlib/jmetrics.hpp
@@ -27,6 +27,7 @@
 #include "jptree.hpp"
 #include "platform.h"
 #include "jstatcodes.h"
+#include "jatomic.hpp"
 
 
 namespace hpccMetrics {
@@ -138,7 +139,7 @@ protected:
     MetricVal(const char *name, const char *desc, MetricType metricType, StatisticMeasure _units, const MetricMetaData &_metaData) :
         MetricBase(name, desc, metricType, _units, _metaData) {}
 
-    std::atomic<__uint64> value{0};
+    RelaxedAtomic<__uint64> value{0};
 };
 
 
@@ -153,6 +154,11 @@ public:
     void inc(uint64_t val = 1)
     {
         value.fetch_add(val);
+    }
+
+    void fastInc(uint16_t val = 1)
+    {
+        value.fastAdd(val);
     }
 };
 
@@ -171,6 +177,11 @@ public:
         value += delta;
     }
 
+    void fastAdjust(int64_t delta)
+    {
+        value.fastAdd(delta);
+    }
+
     /*
      * Set the value
      */
@@ -179,6 +190,32 @@ public:
         value = val;
     }
 };
+
+
+class jlib_decl GaugeMetricFromCounters : public MetricVal
+{
+public:
+    GaugeMetricFromCounters(const char *name, const char *description, StatisticMeasure _units,
+                            const std::shared_ptr<CounterMetric> &_pBeginCounter, const std::shared_ptr<CounterMetric> &_pEndCounter,
+                            const MetricMetaData &_metaData = MetricMetaData()) :
+        MetricVal{name, description, MetricType::METRICS_GAUGE, _units, _metaData},
+        pBeginCounter{_pBeginCounter},
+        pEndCounter{_pEndCounter}
+    {
+        assert(pBeginCounter->queryUnits() == pEndCounter->queryUnits());
+    }
+
+    virtual __uint64 queryValue() const override
+    {
+        auto endValue = pEndCounter->queryValue();
+        return pBeginCounter->queryValue() - endValue;
+    }
+
+protected:
+    std::shared_ptr<CounterMetric> pBeginCounter;
+    std::shared_ptr<CounterMetric> pEndCounter;
+};
+
 
 template<typename T>
 class CustomMetric : public MetricBase
@@ -289,8 +326,30 @@ std::shared_ptr<T> createMetricAndAddToManager(const char *name, const char* des
 }
 
 
+inline std::shared_ptr<CounterMetric> registerCounterMetric(const char *name, const char* desc, StatisticMeasure units, const MetricMetaData &metaData = MetricMetaData())
+{
+    return createMetricAndAddToManager<CounterMetric>(name, desc, units, metaData);
+}
+
+
+inline std::shared_ptr<GaugeMetric> registerGaugeMetric(const char *name, const char* desc, StatisticMeasure units, const MetricMetaData &metaData = MetricMetaData())
+{
+    return createMetricAndAddToManager<GaugeMetric>(name, desc, units, metaData);
+}
+
+
+inline std::shared_ptr<GaugeMetricFromCounters> registerGaugeFromCountersMetric(const char *name, const char* desc, StatisticMeasure units,
+                                                                                const std::shared_ptr<CounterMetric> &pBeginCounter, const std::shared_ptr<CounterMetric> &pEndCounter,
+                                                                                const MetricMetaData &metaData = MetricMetaData())
+{
+    std::shared_ptr<GaugeMetricFromCounters> pMetric = std::make_shared<GaugeMetricFromCounters>(name, desc, units, pBeginCounter, pEndCounter, metaData);
+    queryMetricsManager().addMetric(pMetric);
+    return pMetric;
+}
+
+
 template <typename T>
-std::shared_ptr<CustomMetric<T>> createCustomMetricAndAddToManager(const char *name, const char *desc, MetricType metricType, T &value, StatisticMeasure units, const MetricMetaData &metaData = MetricMetaData())
+std::shared_ptr<CustomMetric<T>> registerCustomMetric(const char *name, const char *desc, MetricType metricType, T &value, StatisticMeasure units, const MetricMetaData &metaData = MetricMetaData())
 {
     std::shared_ptr<CustomMetric<T>> pMetric = std::make_shared<CustomMetric<T>>(name, desc, metricType, value, units, metaData);
     queryMetricsManager().addMetric(pMetric);
@@ -328,9 +387,10 @@ public:
         : gauge{_pGauge}, amount{_amount}
     { }
     ScopedGaugeDecrementer(const ScopedGaugeDecrementer&) = delete;
-    ScopedGaugeDecrementer(ScopedGaugeDecrementer&) = delete;
+    ScopedGaugeDecrementer(ScopedGaugeDecrementer&&) = delete;
     ScopedGaugeDecrementer& operator=(const ScopedGaugeDecrementer&) = delete;
-    ScopedGaugeDecrementer& operator=(ScopedGaugeDecrementer&) = delete;
+    ScopedGaugeDecrementer& operator=(ScopedGaugeDecrementer&&) = delete;
+
     ~ScopedGaugeDecrementer()
     {
         gauge.adjust(-amount);
--- a/testing/unittests/metrics/MetricFrameworkTests.cpp
+++ b/testing/unittests/metrics/MetricFrameworkTests.cpp
@@ -71,8 +71,9 @@ public:
         CPPUNIT_TEST(Test_custom_metric);
         CPPUNIT_TEST(Test_manager_calls_sink_to_start_and_stop_collection);
         CPPUNIT_TEST(Test_manager_manages_metrics_properly);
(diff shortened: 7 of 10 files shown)


jakesmith (commented):
@kenrowland - please don't squash until review is over, it makes it harder to see which review comments have been address/what has changed.

it looks fine I think, but should remove the redundant code in daserver.cpp lines 778-781, and make the ScopedGaugeUpdater usage consistent.

dali/base/danqs.cpp
```diff
@@ -41,6 +42,8 @@
 // # TYPE nq_active_requests gauge
 // A unsigned scalar should suffice, i.e. never going to have that many active transactions
 
+static auto pNqRequestsCount = hpccMetrics::createMetricAndAddToManager<hpccMetrics::CounterMetric>("nq_requests.received", "The total number of Dali NQ requests received", SMeasureCount);
```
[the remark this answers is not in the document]
ghalliday: It should either be hpcc.nq_requests or dali.nq_requests.  Probably slight preference for the latter.

@jake would you commonly want to aggregate nq and sds requests together or not?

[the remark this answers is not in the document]
ghalliday: A bit verbose, I wonder whether 
    hpccMetrics::registerCounterMetric(...)
would be better to improve readability.

dali/base/danqs.cpp
```diff
@@ -556,6 +560,8 @@ class CDaliNamedQueueServer: public IDaliServer, public Thread, implements IConn
 
     void processMessage(CMessageBuffer &mb)
     {
+        hpccMetrics::ScopedGaugeUpdater gaugeUpdater(*pNqActiveRequests, 1);
```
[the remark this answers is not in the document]
ghalliday: This isn't very useful.  Its value will be either 0 or 1, indicating whether dali is currently processing a message.

dali/base/danqs.cpp
```diff
@@ -531,6 +534,7 @@ class CDaliNamedQueueServer: public IDaliServer, public Thread, implements IConn
             try {
                 // 1) Need to increment nq_requests here
                 // NB: it will never be decremented. This is total for life of this instance.
+                pNqRequestsCount->inc(1);
```
ghalliday, line 537: @jakesmith I think this is in the wrong place - this is before the message has been received, so will generally be too high.

dali/base/danqs.cpp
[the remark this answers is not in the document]
ghalliday: A general efficiency concern - this involves two atomic operations.  If you recorded the "dali.nq_requests.processed" instead it would only involve 1 atomic operation.  You could publish dali.nq_requests.active by subtracting them.  I suspect it is more useful to publish a counter of the number processed.
@rpastrana how easy is it in prometheus to subtract two counters and display the result?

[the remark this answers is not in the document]
ghalliday: A different efficiency concern is that this code is single-threaded so none of the metric increments need to be atomic.  There should be a version of the increment that avoids the atomic penalty.  (approx 10ns on intel for reference, different on ARM/POWER8)

dali/base/dasds.cpp
```diff
@@ -94,6 +95,10 @@ static unsigned msgCount=(unsigned)-1;
 // # TYPE sds_pending_requests gauge
 // A unsigned scalar should suffice, i.e. never going to have that many pending transactions
 
+static auto pSdsRequestsCount = hpccMetrics::createMetricAndAddToManager<hpccMetrics::CounterMetric>("sds_requests.received", "The total number of Dali SDS requests handled", SMeasureCount);
```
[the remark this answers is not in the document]
ghalliday: Number of SDS requests received, rather than handled.  Also needs a "dali." prefix.

dali/base/dasds.cpp
```diff
@@ -94,6 +95,10 @@ static unsigned msgCount=(unsigned)-1;
...
+static auto pSdsActiveRequests = hpccMetrics::createMetricAndAddToManager<hpccMetrics::GaugeMetric>("sds_requests.active", "Current number of active SDS requests being handled", SMeasureCount);
+static auto pSdsPendingRequests = hpccMetrics::createMetricAndAddToManager<hpccMetrics::GaugeMetric>("sds_requests.pending", "Current number of pending SDS requests", SMeasureCount);
```
ghalliday, line 100: I'm not sure how useful this metrics are for alerting or scaling.  I am happy for them to be published - because they will probably look pretty in a graph, but they're only a snapshot.  Publishing the counts allows you to work out the average load over the period, rather than the instantaneous load at the point the metrics were gathered.)

Better to publish "dai.sqs_requests.started", "dai.sqs_requests.processed" or "dai.sqs_requests.completed" (should be consistent with nq above, and same pattern should be used for all types of requests processed in the system.

dali/base/dasds.cpp
```diff
@@ -3755,6 +3760,8 @@ int CSDSTransactionServer::run()
                 // to decrement when handling is complete.
                 // NB: most transactions are handled asynchronously, so decrement will need to happen on another thread
 
+                pSdsRequestsCount->inc(1);
```
ghalliday, line 3763: This doesn't need to be atomic.  It also replaces msgCount, so that code can be removed.

dali/base/dasds.cpp
```diff
@@ -4098,6 +4106,9 @@ void CSDSTransactionServer::processMessage(CMessageBuffer &mb)
     // 2) Ensure sds_active_requests is decremented, when processMessage() is finished
     // e.g. use a scoped object here, and decrement in dtor.
 
+    pSdsPendingRequests->adjust(-1);
+    hpccMetrics::ScopedGaugeUpdater sdsActiveGaugeUpdater(*pSdsActiveRequests);
```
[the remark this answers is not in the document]
ghalliday: This can also be non atomic.

If you only use counters I think all 3 counters can currently be non-atomic.

ghalliday (commented):
@kenrowland some initial comments

kenrowland:
@ghalliday @jakesmith Please review. A couple of things:
- I left some old code commented for reference to what was being done before
- I added a scoped increment of a count for the synthesized pending gauge (in order to ensure the number of completed requests is updated
- I did not change metric updates to the fast versions. I figure we can change them once we agree fully on the new names and which updates can be the fast versions (probably all if I remember correctly).

dali/server/daserver.cpp
```diff
@@ -765,3 +767,24 @@ int main(int argc, const char* argv[])
...
+    {
+        pMetricsTree.setown(getComponentConfigSP()->getPropTree("metrics"));
+    }
+
+    if (pMetricsTree != nullptr)
+    {
+        PROGLOG("Metrics initializing...");
+        hpccMetrics::MetricsManager &metricsManager = hpccMetrics::queryMetricsManager();
+        metricsManager.init(pMetricsTree);
+        metricsManager.startCollecting();
```
ghalliday, line 788: I don't see a matching call to stopCollecting()

ghalliday: Thinking about generalizing this to more components.  Is it worth adding an 

initAndStartCollecting(IPropertyTree * optTree);

function so that instead of this if() block being duplicated in each component it could say

hpccMetrics::queryMetricsManager().initAndStartCollecting(pMetricsTree);

Also, since the code code for getting the pMetricsTree is likely to be identical for each component, even better might be:

hpccMetrics::queryMetricsManager().initAndStartCollecting();

esp/platform/txsummary.cpp
```diff
@@ -24,9 +24,7 @@
 
 #define MATCH_ENTRY [&](const EntryValue& e) {return strieq(e.get()->name, pathPart);}
 
-static auto pRequestCount = hpccMetrics::createMetricAndAddToManager<hpccMetrics::CounterMetric>("requests",
-                                                                                                 "Number of Requests",
-                                                                                                 SMeasureCount);
+static auto pRequestCount = hpccMetrics::registerCounterMetric("esp.requests.received", "Number of Requests", SMeasureCount);
```
ghalliday, line 27: description could be more verbose.  Same for others that have been modified.

system/jlib/jmetrics.hpp
```diff
@@ -341,4 +395,25 @@ class jlib_decl ScopedGaugeDecrementer
...
 
+
+class jlib_decl ScopedCountIncrement
+{
+public:
+    explicit ScopedCountIncrement(CounterMetric &_pCount, int64_t _amount=1)
+    : count{_pCount}, amount{_amount}
+    { }
+    ScopedCountIncrement(const ScopedGaugeDecrementer&) = delete;
+    ScopedCountIncrement(ScopedGaugeDecrementer&) = delete;
```
ghalliday, line 406: is this useful?  I think it is implied by copy constructor on the  line above.  Same for the second assignment.  Did you mean to use &&?  I'm not sure it is likely to catch real world problems.

ghalliday (commented):
@kenrowland A few comments, but looks closer to being ready.

system/jlib/jmetrics.hpp
kenrowland: After some further research, the non const copy constructor and assignment operator should be sufficient. Move copy constructors and assignment are not automatically generated, so they were left out. However, since we are deleting the non move copy constructor, best to ensure they are deleted as well.

testing/unittests/metrics/MetricFrameworkTests.cpp
```diff
@@ -82,18 +83,28 @@ class MetricFrameworkTests : public CppUnit::TestFixture
...
         CPPUNIT_ASSERT_EQUAL(0, static_cast<int>(pCounter->queryValue()));
+        int expectedValue = 0;
 
         //
         // Test default increment (by 1)
         pCounter->inc();
+        expectedValue++;
         int counterValue = pCounter->queryValue();
-        CPPUNIT_ASSERT_EQUAL(1, counterValue);
+        CPPUNIT_ASSERT_EQUAL(expectedValue, counterValue);
```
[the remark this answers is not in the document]
kenrowland: I found that as I added more updates to the metric, keeping track of what the expected result was harder. An earlier test in the sequence forces a change in all subsequent asserts. So I added a variable to keep track of the expected changes. If you feel that a constant is better, let me know and I will change it back.

testing/unittests/metrics/MetricFrameworkTests.cpp
```diff
@@ -210,8 +221,15 @@ class MetricFrameworkTests : public CppUnit::TestFixture
             ScopedGaugeDecrementer gaugeDecrementer(*pGauge, 1);
         }
         CPPUNIT_ASSERT_EQUAL(0, static_cast<int>(pGauge->queryValue()));
+
+        std::shared_ptr<CounterMetric> pCount = std::make_shared<CounterMetric>("test-count", "description", SMeasureCount);
+        {
+            ScopedCountIncrement(*pCount, 1);
+        }
```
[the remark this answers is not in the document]
kenrowland: ScopedCountIncrement has been removed in favor of using COnScopeExit

kenrowland:
@ghalliday @jakesmith Please see latest changes. A couple of open issues for Jake to comment and some questions.

@afishbeck Some minor changes to the implementation in ESP should you want to take a look.

dali/base/danqs.cpp
```diff
@@ -531,6 +533,7 @@ class CDaliNamedQueueServer: public IDaliServer, public Thread, implements IConn
             try {
                 // 1) Need to increment nq_requests here
                 // NB: it will never be decremented. This is total for life of this instance.
+                pNqRequestsCount->fastInc(1);
```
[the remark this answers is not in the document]
jakesmith: yes, comments are in the wrong place.
Comment+code should be after the recv, or better still to be inside and at start of processMessage

dali/base/dasds.cpp
```diff
@@ -3755,6 +3761,8 @@ int CSDSTransactionServer::run()
                 // to decrement when handling is complete.
                 // NB: most transactions are handled asynchronously, so decrement will need to happen on another thread
 
+                pSdsRequestsReceived->inc(1);
```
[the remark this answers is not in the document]
jakesmith: yes, this is the only thread that receives requests/can update.

dali/base/danqs.cpp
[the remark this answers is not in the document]
jakesmith: >@jake would you commonly want to aggregate nq and sds requests together or not?

I don't think so. They are different beasts.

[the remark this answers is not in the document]
jakesmith: > A bit verbose, I wonder whether
> hpccMetrics::registerCounterMetric(...)
> would be better to improve readability.

I agree, think that is more readable.

dali/base/dasds.cpp
```diff
@@ -4091,12 +4088,12 @@ bool translateOldFormat(CServerRemoteTree *parentServerTree, IPropertyTree *pare
...
-    // 1) Decrement sds_pending_requests here.
+    // 1) Increment sds_requests.started here.
     // NB: request has started, no longer pending
-    // See CSDSTransactionServer::run() for where needs to be incremented. 
+    // See CSDSTransactionServer::run() for where needs to be incremented.
 
-    // 2) Ensure sds_active_requests is decremented, when processMessage() is finished
-    // e.g. use a scoped object here, and decrement in dtor.
+    pSdsRequestsStarted->inc(1);  // this is where the active count was incremented before
+    COnScopeExit incCompletedOnExit([&](){ pSdsRequestsCompleted->inc(); });
```
jakesmith, line 4096: inc(1) vs inc() inconsistency.

Unless @ghalliday disagrees, let's take out the default param value altogether and force it to [consistently] require to be inc(1)

dali/base/danqs.cpp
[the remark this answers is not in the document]
kenrowland: Moved to inside processMessage

dali/base/dasds.cpp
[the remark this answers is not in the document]
kenrowland: Changed to fastInc;

dali/base/dasds.cpp
kenrowland, line 4096: @jakesmith Could the inc here also be a fastInc ?

dali/base/dasds.cpp
```diff
@@ -3749,11 +3755,7 @@ int CSDSTransactionServer::run()
...
                 // 1) Need to increment sds_requests here
                 // NB: it will never be decremented. This is total for life of this instance.
-                
-                // 2) Need to increment sds_active_requests here
-                // and ensure it's scoped, such that it is guaranteed
-                // to decrement when handling is complete.
-                // NB: most transactions are handled asynchronously, so decrement will need to happen on another thread
+                pSdsRequestsReceived->inc(1);
 
                 msgCount++;
```
jakesmith, line 3760: > This doesn't need to be atomic. It also replaces msgCount, so that code can be removed.

all references to msgCount should now be removed (2), as replaced by pSdsRequestsReceived

jakesmith:
>If you only use counters I think all 3 counters can currently be non-atomic.

comment from changed code in sds, but..

@ghalliday , the started/completed do need to atomic afaics. There are multiple 'processMessage' threads that will be updating these counters.

dali/base/danqs.cpp
```diff
@@ -41,6 +42,7 @@
 // # TYPE nq_active_requests gauge
 // A unsigned scalar should suffice, i.e. never going to have that many active transactions
```
jakesmith, line 43: the above comment block re. active should be removed as we're not intending to supply this (as can only be 0 or 1) for NQ.

dali/base/dasds.cpp
```diff
@@ -94,6 +95,11 @@ static unsigned msgCount=(unsigned)-1;
 // # TYPE sds_pending_requests gauge
 // A unsigned scalar should suffice, i.e. never going to have that many pending transactions
 
+static auto pSdsRequestsReceived = hpccMetrics::registerCounterMetric("dali.sds_requests.received", "The total number of Dali SDS requests received", SMeasureCount);
+static auto pSdsRequestsStarted = hpccMetrics::registerCounterMetric("dali.sds_requests.started", "The total number of Dali SDS requests started", SMeasureCount);
+static auto pSdsRequestsCompleted = hpccMetrics::registerCounterMetric("dali.sds_requests.completed", "The total number of Dali SDS requests completed", SMeasureCount);
+static auto pSdsRequestsPending = hpccMetrics::registerGaugeFromCountersMetric("dali.sds_requests.pending", "Current number of pending SDS requests", SMeasureCount, pSdsRequestsReceived, pSdsRequestsStarted);
```
jakesmith, line 101: did we drop the idea of tracking pending request?
I think that makes sense, because we know # received, number started and number completed so can calculate number blocked pending.
But if so, this line should be deleted.

dali/base/dasds.cpp
jakesmith: no, I don't see how it can, there will multiple concurrent threads at this point (in processMessage via thread pool)

system/jlib/jmetrics.hpp
```diff
@@ -154,6 +154,11 @@ class jlib_decl CounterMetric : public MetricVal
     {
         value.fetch_add(val);
     }
+
+    void fastInc(uint16_t val = 1)
+    {
+        value = value + val;
```
jakesmith, line 160: Isn't this actually slower ..?
Because it's still an atomic "value + val" will require 'value' to be read atomically, and then the assignment requires an atomic assignment .
So potentially twice as slow..

@ghalliday - should the fast version instead use memory_order_relaxed order? (but would it actually be any quicker, given 2 ops).

And/or for places where you know the counter is only every going to be single threaded,
should there be a different implementation, e.g. CounterNonAtomicMetric

system/jlib/jmetrics.hpp
```diff
@@ -171,6 +176,11 @@ class jlib_decl GaugeMetric : public MetricVal
         value += delta;
     }
 
+    void fastAdjust(int64_t delta)
+    {
+        value = value + delta;
```
jakesmith, line 181: as above, I think this it's actually slower.

system/jlib/jmetrics.hpp
ghalliday: @jakesmith this is actually faster on intel.   The reason is that read and write are actually always atomic (unless they are misaligned, I think over a cache line), and it has a naturally strong memory ordering.  An atomic increment uses a LOCK ADD - which has various effects - mainly synchronizing the cache line over different cores, and has an overhead of ~18cycles as a result (read https://www.agner.org/optimize/instruction_tables.pdf if you really care!).

On non intel chips with a weak memory model there is likely to be a difference between using memory_order_relaxed - which is what the RelaxedAtomic defaults to - because the default std::atomic mode will probably require extra work to ensure the results are consistent (which isn't needed here).

So yes, this is faster, but even better would be to use RelaxedAtomic, or use the memory_order_relaxed on the access of the metric values,

dali/base/dasds.cpp
ghalliday: @jakesmith this line automatically calculates and publishes the number.

system/jlib/jmetrics.hpp
jakesmith: ok, it's less than obvious from a casual glance at the code.

dali/base/dasds.cpp
jakesmith: right ok, yes I'd missed what it was doing.

ghalliday:
@kenrowland after more thought and research I think we need to get the memory order parameter correct - and make sure it is relaxed.  (It does not matter if these atomic reads and writes are reordered in relation to other reads and writes.)

dali/base/dasds.cpp
```diff
@@ -4091,12 +4088,12 @@ bool translateOldFormat(CServerRemoteTree *parentServerTree, IPropertyTree *pare
...
 {
-    // 1) Decrement sds_pending_requests here.
+    // 1) Increment sds_requests.started here.
     // NB: request has started, no longer pending
-    // See CSDSTransactionServer::run() for where needs to be incremented. 
+    // See CSDSTransactionServer::run() for where needs to be incremented.
 
-    // 2) Ensure sds_active_requests is decremented, when processMessage() is finished
-    // e.g. use a scoped object here, and decrement in dtor.
+    pSdsRequestsStarted->inc(1);  // this is where the active count was incremented before
```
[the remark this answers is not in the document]
kenrowland: Comment was informational only for the review in case anyone wanted to know what change this represented. I intended on removing it for the final commit. I will remove it now.

dali/base/danqs.cpp
kenrowland: Removed.

dali/base/dasds.cpp
```diff
@@ -3749,11 +3755,7 @@ int CSDSTransactionServer::run()
             {
                 // 1) Need to increment sds_requests here
```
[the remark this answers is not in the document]
kenrowland: Removed

dali/base/dasds.cpp
kenrowland: Removed msgCount.

dali/base/dasds.cpp
kenrowland: Since that is change to the base framework, I would like to address it in a separate Jira after this is merged. @ghalliday, comments?

kenrowland: It shall remain as is.

esp/platform/espprotocol.cpp
```diff
@@ -29,11 +29,8 @@
 
 static RelaxedAtomic<unsigned> gActiveRequests;
 
-static auto pActiveRequests = hpccMetrics::createCustomMetricAndAddToManager("activerequests",
-                                                                             "Number of active requests",
-                                                                             hpccMetrics::METRICS_GAUGE,
-                                                                             gActiveRequests,
-                                                                             SMeasureCount);
+static auto pActiveRequests = hpccMetrics::registerCustomMetric("esp.requests.active", "Number of active requests",
+                                                                hpccMetrics::METRICS_GAUGE,gActiveRequests, SMeasureCount);
```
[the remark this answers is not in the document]
kenrowland: Added.

system/jlib/jmetrics.hpp
```diff
@@ -180,6 +190,31 @@ class jlib_decl GaugeMetric : public MetricVal
     }
 };
 
+
+class jlib_decl GaugeMetricFromCounters : public MetricVal
+{
+public:
+    GaugeMetricFromCounters(const char *name, const char *description, StatisticMeasure _units,
+                            const std::shared_ptr<CounterMetric> &_pEventCount, const std::shared_ptr<CounterMetric> &_pEnterStateCount,
```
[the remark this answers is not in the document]
kenrowland: Let's go with being and end for now.

system/jlib/jmetrics.hpp
```diff
@@ -180,6 +190,31 @@ class jlib_decl GaugeMetric : public MetricVal
...
+                            const MetricMetaData &_metaData = MetricMetaData()) :
+        MetricVal{name, description, MetricType::METRICS_GAUGE, _units, _metaData},
+        pEventCount{_pEventCount},
+        pEnterStateCount{_pEnterStateCount}
+        {
```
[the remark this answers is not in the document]
kenrowland: Moved

kenrowland:
@ghalliday In order to get the memory order correct per your comment, I went ahead and changed the type for the atomic member variable to a RelaxedAtomic instead of adding the order parameter to each individual operation.

dali/base/dasds.cpp
```diff
@@ -80,8 +81,6 @@ static unsigned readWriteTimeout = 60000;
 #define DEFAULT_LCIDLE_RATE 1          // 1 write transactions per idle period. <= this rate is deemed idle (suitable for save)
 #define STORENOTSAVE_WARNING_PERIOD 72 // hours
 
-static unsigned msgCount=(unsigned)-1;
-
 // # HELP sds_requests The total number of Dali SDS requests handled
```
ghalliday, line 84: Minor: The following comments are also redundant and should go.

dali/base/dasds.cpp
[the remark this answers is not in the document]
ghalliday: minor: not removed yet.

dali/base/dasds.cpp
ghalliday: I don't mind if this pr, or the next.  The next should also revisit all the existing metric names and make sure they match the agreed naming scheme.  (I think they may already, but worth double checking.)

dali/server/daserver.cpp
```diff
@@ -752,6 +766,7 @@ int main(int argc, const char* argv[])
         if (ok) {
             writeSentinelFile(sentinelFile);
             covenMain();
+            initializeMetrics(serverConfig);
             removeAbortHandler(actionOnAbort);
         }
         stopServer();
```
ghalliday, line 772: @jakesmith would it be suitable to call stopCollecting() after this call to stopServer()?

system/jlib/jmetrics.hpp
```diff
@@ -138,7 +139,8 @@ class jlib_decl MetricVal : public MetricBase
     MetricVal(const char *name, const char *desc, MetricType metricType, StatisticMeasure _units, const MetricMetaData &_metaData) :
         MetricBase(name, desc, metricType, _units, _metaData) {}
 
-    std::atomic<__uint64> value{0};
+    RelaxedAtomic<__uint64> value{0};
+    //std::atomic<__uint64> value{0};
```
ghalliday, line 143: remove comment

system/jlib/jmetrics.hpp
ghalliday: should now call value.fastAdd(val);

system/jlib/jmetrics.hpp
```diff
@@ -171,6 +178,11 @@ class jlib_decl GaugeMetric : public MetricVal
         value += delta;
     }
 
+    void fastAdjust(int64_t delta)
+    {
+        value = value + delta;
+    }
```
ghalliday, line 184: should now call value.fastAdd()

system/jlib/jmetrics.hpp
```diff
@@ -180,6 +192,31 @@ class jlib_decl GaugeMetric : public MetricVal
...
+        MetricVal{name, description, MetricType::METRICS_GAUGE, _units, _metaData},
+        pBeginCounter{_pBeginCounter},
+        pEndCounter{_pEndCounter}
+    {
+        assert(pBeginCounter->queryUnits() == pEndCounter->queryUnits());
+    }
+
+    virtual __uint64 queryValue() const override
+    {
+        return pBeginCounter->queryValue() - pEndCounter->queryValue();
```
ghalliday, line 211: really picky (I debated even mentioning it).
Better as

```
endValue = pEndCounter->queryValue()
return pBeginCounter->queryValue() - endValue;

```
Why?  The evaluation order is undefined in the previous code.  If begin->value was called, then both begin and end were incremented before end->value() was called you could end up with a -ve count.  Making the order explicit will ensure it is positive.

ghalliday (commented):
@kenrowland a few comments.  Looks like it is close to being ready.

dali/base/dasds.cpp
kenrowland: In my opinion these comments are now no longer needed. The instantiation of the metric itself has the description which should suffice. If additional comments are needed referencing the individual metrics, I can restore them.

dali/base/dasds.cpp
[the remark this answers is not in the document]
kenrowland: I actually removed it this time.

dali/base/dasds.cpp
kenrowland: Opened https://track.hpccsystems.com/browse/HPCC-26501 to address this.

system/jlib/jmetrics.hpp
kenrowland: Removed (yes actually removed).

system/jlib/jmetrics.hpp
kenrowland: Updated to use relaxed atomic class methods.

system/jlib/jmetrics.hpp
kenrowland: Updated.

system/jlib/jmetrics.hpp
kenrowland: An explicit order is better. The value can be 0 (which some argue is included with a positive designation).

kenrowland (commented):
Updated based on comments from @ghalliday.

@jakesmith There are a few open questions to you.

dali/base/dasds.cpp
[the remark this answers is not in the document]
jakesmith: I don't think any of the remaining comments above useful (and slightly confusing)

dali/base/dasds.cpp
jakesmith: >Since that is change to the base framework, I would like to address it in a separate Jira after this is merged.

without chanigng the methods, can we at least make the call consistent? i.e. make both the above calls inc(1).

dali/server/daserver.cpp
```diff
@@ -752,6 +766,7 @@ int main(int argc, const char* argv[])
         if (ok) {
             writeSentinelFile(sentinelFile);
             covenMain();
+            initializeMetrics(serverConfig);
```
jakesmith, line 769: this isn't going to work here, covenMain enters a loop and will not return until Dali starts to quit.
Should be moved before covenMain call

kenrowland: @jakesmith Can it simply be moved up one line?

dali/server/daserver.cpp
jakesmith: yes, think it should be.
If it isn't I suspect there could be potential problem (crashes) during closedown whilst the sink threads are still running and objects they are reporting on are destructed.

kenrowland: @jakesmith The framework only maintains weak pointers to metric objects, so if Dali is shutting down and destroying any metric objects, the framework will detect that the object has been deleted and will delete it from its internal collection.

ghalliday: It can still cause a problem - if the framework has a strong pointer to an object, and that object is an instance of a class defined in a dll which is then unloaded.  Then the thread can crash when it calls a virtual method of the class whose definition is no longer in memory.  We have even seen instances of this in the past...

jakesmith (changes requested):
@kenrowland - a few comments.
The significant one is that dali metrics can't be working at the moment afaics,
because initializeMetrics is only being called after Dali starts exiting.

dali/server/daserver.cpp
kenrowland: That is a danger when using custom metrics and perhaps we should better document the concern. Or, custom metrics should be a temporary solution with an impetus to migrate to framework defined classes.

dali/server/daserver.cpp
jakesmith: yes that is enough afaics.

dali/server/daserver.cpp
jakesmith: If you add stopCollecting() call, the threads I assume will be stopped, therefore avoiding the problem being discussed.

testing/unittests/metrics/MetricFrameworkTests.cpp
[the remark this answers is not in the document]
ghalliday: I don't really understand this - if you add tests on the end then you don't need to change any other code, and the previously value is the one in the assert.
Not worth the hassle of updating though.

ghalliday (commented):
@kenrowland a few notes about out of date comments, but looks very close to ready.  Should be squashed once others have approved.

dali/base/dasds.cpp
```diff
@@ -4091,12 +4070,12 @@ bool translateOldFormat(CServerRemoteTree *parentServerTree, IPropertyTree *pare
 
 void CSDSTransactionServer::processMessage(CMessageBuffer &mb)
 {
-    // 1) Decrement sds_pending_requests here.
+    // 1) Increment dali.sds_requests.started here.
     // NB: request has started, no longer pending
-    // See CSDSTransactionServer::run() for where needs to be incremented. 
+    // See CSDSTransactionServer::run() for where needs to be incremented.
```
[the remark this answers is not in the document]
kenrowland: Add comment to make it clear what is happening

kenrowland:
@ghalliday Rebased with latest comment changes.
@jakesmith If you want a final review...

ghalliday (approved):

kenrowland:
@richardkchapman Please merge

jakesmith (approved):

PR #15295, opened 2021-08-26
HPCC-26439 Allow string terminator in WU result when converting to CSV

wangkx:
Inside a WU result string, covert \0 to "\000".

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/rtl/eclrtl/rtlformat.cpp
+++ b/rtl/eclrtl/rtlformat.cpp
@@ -1402,17 +1402,25 @@ unsigned CommonCSVWriter::getChildrenMaxColumnID(CCSVItem* item, unsigned& maxCo
     return maxColumnID;
 }
 
-void CommonCSVWriter::escapeQuoted(unsigned len, char const* in, StringBuffer& out)
+void CommonCSVWriter::escapeString(unsigned len, char const* in, StringBuffer& out)
 {
     char const* finger = in;
     while (len--)
     {
-        //RFC-4180, paragraph "If double-quotes are used to enclose fields, then a double-quote
-        //appearing inside a field must be escaped by preceding it with another double quote."
-        //unsigned newLen = 0;
-        if (*finger == '"')
-            out.append('"');
-        out.append(*finger);
+        if (*finger != '\0')
+        {
+            //RFC-4180, paragraph "If double-quotes are used to enclose fields, then a double-quote
+            //appearing inside a field must be escaped by preceding it with another double quote."
+            if (*finger == '"')
+                out.append('"');
+            out.append(*finger);
+        }
+        else
+        {
+            //null bytes should not occur, but encode them if they do to prevent problems elsewhere
+            out.append("\000");
+        }
+
         finger++;
     }
 }
@@ -1544,7 +1552,7 @@ void CommonCSVWriter::addStringField(unsigned len, const char* field, const char
 {
     StringBuffer v;
     v.append(csvQuote);
-    escapeQuoted(len, field, v);
+    escapeString(len, field, v);
     v.append(csvQuote);
     addContentField(v.str(), fieldName);
 }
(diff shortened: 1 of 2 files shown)


esp/bindings/http/platform/httptransport.cpp
```diff
@@ -941,14 +941,14 @@ int CHttpMessage::startSend()
     return 0;
 }
 
-int CHttpMessage::sendChunk(const char *chunk)
```
ghalliday, line 944: These changes do not seem to be used, and I doubt that including null characters in http responses is a good idea.

rtl/eclrtl/rtlformat.cpp
```diff
@@ -1054,7 +1054,7 @@ void CommonCSVWriter::outputBool(bool field, const char* fieldName)
 {
     if (!checkHeaderName(fieldName))
         return;
-    addContentField((field) ? "true" : "false", fieldName);
```
ghalliday, line 1057: I would retain the old function as well as the new function - passing in the string explicitly is a bit error prone.

rtl/eclrtl/rtlformat.cpp
```diff
@@ -1546,7 +1546,7 @@ void CommonCSVWriter::addStringField(unsigned len, const char* field, const char
     v.append(csvQuote);
     escapeQuoted(len, field, v);
```
ghalliday, line 1547: Would a better fix be to convert \0 to a better ascii representation.  E.g. "\000" or similar?  Including null characters in csv text is dubious, and probably best not to directly include them in the result in case they cause other follow on problems.  What was the example that provoked the change (the Jira doesn't have any more details).

ghalliday (commented):
@wangkx I'm not sure about the approach - see my comments.  It would be good to have more details in the jira of the initial example.

rtl/eclrtl/rtlformat.cpp
wangkx: It is a good idea to covert \0 to "\000". 
I just uploaded a ZAP report to the Jira. The ZAP report gives the example.

wangkx:
@ghalliday please review the new commit.

rtl/eclrtl/rtlformat.cpp
```diff
@@ -1402,17 +1404,22 @@ unsigned CommonCSVWriter::getChildrenMaxColumnID(CCSVItem* item, unsigned& maxCo
...
     char const* finger = in;
     while (len--)
     {
-        //RFC-4180, paragraph "If double-quotes are used to enclose fields, then a double-quote
-        //appearing inside a field must be escaped by preceding it with another double quote."
-        //unsigned newLen = 0;
-        if (*finger == '"')
-            out.append('"');
-        out.append(*finger);
+        if (*finger == '\0')
```
ghalliday, line 1412: I would reverse this test - simpler to have the common (non null) branch 1st (and the compiler will also assume that is the case).
This should also have a comment e.g.
//null bytes should not occur, but encode them if they do to prevent problems elsewhere

rtl/eclrtl/rtlformat.cpp
```diff
@@ -1402,17 +1404,22 @@ unsigned CommonCSVWriter::getChildrenMaxColumnID(CCSVItem* item, unsigned& maxCo
...
+            out.append("\000");
+        else
+        {
+            //RFC-4180, paragraph "If double-quotes are used to enclose fields, then a double-quote
+            //appearing inside a field must be escaped by preceding it with another double quote."
+            //unsigned newLen = 0;
```
ghalliday, line 1418: unrelated: can be deleted.

ghalliday (commented):
@wangkx a couple of comments

wangkx:
@ghalliday the code is revised. Please review.

ghalliday (approved):

PR #15318, opened 2021-09-03
HPCC-26448 Logical Files Filter on Index

jeclrsg:
Fix issue where checking Index in the LogicalFiles filter did not properly set ContentType = "key" in the request to DFUQuery

Signed-off-by: Jeremy Clements <jeremy.clements@lexisnexisrisk.com>

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/src/src-react/components/Files.tsx
+++ b/esp/src/src-react/components/Files.tsx
@@ -29,7 +29,12 @@ const FilterFields: Fields = {
     "EndDate": { type: "datetime", label: nlsHPCC.ToDate, placeholder: "" },
 };
 
-function formatQuery(filter) {
+function formatQuery(_filter) {
+    const filter = { ..._filter };
+    if (filter.Index) {
+        filter.ContentType = "key";
+        delete filter.Index;
+    }
     if (filter.StartDate) {
         filter.StartDate = new Date(filter.StartDate).toISOString();
     }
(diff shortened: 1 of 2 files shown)


esp/src/src-react/components/Files.tsx
```diff
@@ -30,6 +30,9 @@ const FilterFields: Fields = {
 };
 
 function formatQuery(filter) {
+    if (filter.Index) {
+        filter.ContentType = "key";
```
GordonSmith, line 34: You should also `delete filter.Index` as well.

GordonSmith (changes requested):

esp/src/src-react/components/Files.tsx
jeclrsg: If filter.Index is removed, the checkbox will not show as checked if you were to land on the url & then open the Filter to modify it. Changing the name of the field to "ContentType" also has the same issue. If the value is correct for the request the checkbox won't be checked when repopulated via the url, since it is expecting a boolean.

GordonSmith: What I do in the `workunits.tsx` component is:

```
function formatQuery(_filter) {
    const filter = { ..._filter };
  ...
    return filter;
}
```

jeclrsg: Thanks Gordon. Changed as requested & also rebased.

GordonSmith (approved):
@jeclrsg can you rebase onto candidate-8.4.x otherwise looks good.

PR #15319, opened 2021-09-03
HPCC-26476 FTSlave cleanup

richardkchapman:
## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Cloud-compatibility
  - [x] Premature optimization

## Smoketest:

## Testing:
Compiled.


richardkchapman:
@jakesmith @ghalliday 

Main concern would be if the removed functionality (a) might be needed in the near future or (b) is potentially required by clients from old systems connecting to new ones. I don't THINK that either are likely to be the case.

jakesmith:
>Main concern would be if the removed functionality (a) might be needed in the near future or (b) is potentially required by clients from old systems connecting to new ones. I don't THINK that either are likely to be the case.

For (a) , I'd say even if it is, which I doubt, given I think this code has not been used in a long time, we can re-introduce what we need, and in the meantime the code is more maintainable without the noise of unused code.

For (b) I suspect that can't be true for any modern system, if it's in use, it's probably only in v. old versions.

The code changes/removal looks good to me.

jakesmith (approved):
Looks good to me

esp/services/ws_dfu/ws_dfuService.cpp
```diff
@@ -2230,7 +2230,7 @@ void CWsDfuEx::doGetFileDetails(IEspContext &context, IUserDescriptor *udesc, co
         df->setRestrictedAccess(changeRestriction==CDFUChangeRestriction_Restrict);
     }
 
-    offset_t size=queryDistributedFileSystem().getSize(df), recordSize=df->queryAttributes().getPropInt64("@recordSize",0);
+    offset_t size=df->getFileSize(true, false), recordSize=df->queryAttributes().getPropInt64("@recordSize",0);
```
ghalliday, line 2233: @richardkchapman this isn't directly connected to removing the old code.  Although the code for the two calls looks identical, do you want to keep this change?

ghalliday (commented):
@richardkchapman looks good - one question.

esp/services/ws_dfu/ws_dfuService.cpp
richardkchapman: It is required to remove the link to the dalift dll from esp services.

ghalliday (approved):

PR #15320, opened 2021-09-03
HPCC-26457 Prevent wuidread resolving result too early.

jakesmith:
Workunit read should not try to get the result when it is
constructed, because conditional branches which aren't used
are still constructed and the result may be from an unexected
dependent subgraph.
Delay checking for result until preparing graph to execute.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/thorlcr/master/thactivitymaster.cpp
+++ b/thorlcr/master/thactivitymaster.cpp
@@ -80,28 +80,82 @@ MODULE_INIT(INIT_PRIORITY_STANDARD)
 
 CActivityBase *createGroupActivityMaster(CMasterGraphElement *container);
 
+static bool getWorkunitResultFilename(CGraphElementBase &container, StringBuffer & diskFilename, const char * wuid, const char * stepname, int sequence)
+{
+    try
+    {
+        Owned<IConstWUResult> result;
+        if (wuid)
+            result.setown(container.queryCodeContext()->getExternalResult(wuid, stepname, sequence));
+        else
+            result.setown(container.queryCodeContext()->getResultForGet(stepname, sequence));
+        if (!result)
+            throw MakeThorException(TE_FailedToRetrieveWorkunitValue, "Failed to find value %s:%d in workunit %s", stepname?stepname:"(null)", sequence, wuid?wuid:"(null)");
+
+        SCMStringBuffer tempFilename;
+        result->getResultFilename(tempFilename);
+        if (tempFilename.length() == 0)
+            return false;
+
+        diskFilename.append("~").append(tempFilename.str());
+        return true;
+    }
+    catch (IException * e) 
+    {
+        StringBuffer text; 
+        e->errorMessage(text); 
+        e->Release();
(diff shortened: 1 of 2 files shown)


jakesmith:
@ghalliday - please review.

This is not an ideal fix, it's essentially moving the check from factory time (which caused the problem) to when the graph is actually used and serialized to slaves.
It really shouldn't be creating the activities that are down conditional branches that aren't [currently] being executed at all, until they're needed, and I tried changing so that was so, but there are too many knock on effects of the way the graphs are initialized/reinitialized and wired together. This is a much more pragmatic fix.

jakesmith:
NB: this issue is ancient - I guess it's quite rare for a wuidread to be in a child query, and dependent on a result, down a conditional arm that isn't executed via some other means.
The issue was hit in 7.6.40-1. I'm currently targetting 8.2.x.

thorlcr/master/thactivitymaster.cpp
```diff
@@ -80,28 +80,82 @@ MODULE_INIT(INIT_PRIORITY_STANDARD)
...
+        e->Release();
+        throw MakeThorException(TE_FailedToRetrieveWorkunitValue, "Failed to find value %s:%d in workunit %s [%s]", stepname?stepname:"(null)", sequence, wuid?wuid:"(null)", text.str());
+    }
+    return false;
+}
+
+
 class CGenericMasterGraphElement : public CMasterGraphElement
 {
+    StringBuffer wuidDiskFilename;
```
ghalliday, line 116: Doesn't seem like it is quite in the right place.

Also seems a shame to add it to every activity in the master.  Probably doesn't actually matter.

ghalliday (commented):
The change looks ok, but wondering if there is an easy change so that the logic was a bit cleaner.  I suspect the answer is no.

thorlcr/master/thactivitymaster.cpp
jakesmith: I could move into a specialized derived class for TAKworkunitread, but wasn't sure it was worth it, or would make it clearer.

jakesmith:
> The change looks ok, but wondering if there is an easy change so that the logic was a bit cleaner. I suspect the answer is no.

I looked at various ways of tackling this.
1) prevent creation of activities down [so far] unused arms of conditionsl, i.e. delay until needed.
But that leads to issues/complications with rewiring the graphs as conditionals change, and also complications with partly started activities like spitter. Solvable, and potentially cleaner in the long run, but would require a lot more changes and more risky, also may not be as effiecient for child queries as they would need to re-wire below conditionals if there were changes.
2) an implementation of an act. that conditionally behaved as a diskread (by using a contained proxy act.), it worked but less clean/more changes.
3) this impl.

This seemed simplest most pragmatic solution to the issue.

ghalliday (approved):

PR #15321, opened 2021-09-06
HPCC-26432 Roxie to copy multiple files in parallel

richardkchapman:
If useRemoteResources is false, Roxie will copy any missing files as they are
opened - this will give multiple threads and make much better use of dafilesrv
socket caching.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/roxie/ccd/ccdfile.cpp
+++ b/roxie/ccd/ccdfile.cpp
@@ -289,12 +289,14 @@ public:
                         {
                             if (traceLevel > 5)
                                 DBGLOG("Opening %s", sourceName);
-                            disconnectRemoteIoOnExit(current);
+                            if (useRemoteResources)
+                                disconnectRemoteIoOnExit(current);
                             break;
                         }
     //                  throwUnexpected();  - try another location if this one has the wrong version of the file
                     }
-                    disconnectRemoteFile(f);
+                    if (useRemoteResources)
+                        disconnectRemoteFile(f);
                 }
                 catch (IException *E)
                 {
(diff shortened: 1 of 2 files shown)


jakesmith (approved):
@richardkchapman - looks sensible to me.

I hadn't appreciated when reviewing the changes initially, that this mechanism will be being called from N queries in parallel, and therefore there will be N parallel copies.

And discussed there's a slight caveat that if a 2nd query comes along and looks up a file a previous query is copying, then it will return from lookup and continue to use it remotely. But I think with the limit on parallel queries loading combined with this mechism , combined with the fact it's unlikely a query like that will continue to actually read any data soon, because it will be blocked on other copies, it's fine.

roxie/ccd/ccdfile.cpp
```diff
@@ -1786,6 +1788,14 @@ class CRoxieFileCache : implements IRoxieFileCache, implements ICopyFileProgress
                         reportedFilesToCopy = true;
                         if (iShouldCopy)
                         {
+                            if (!useRemoteResources)
+                            {
+                                b.leave();
```
ghalliday, line 1793: What happens if multiple threads/queries try to copy the same file at the same time?  Won't they try and copy the file multiple times?

ghalliday (commented):
@richardkchapman one question

roxie/ccd/ccdfile.cpp
richardkchapman: In theory, the copy only happens when a file is added to the cache. One thread will add to the cache and do the copy, the others should find the file already in the cache and get a (lazy, initially remote) IFileIO to it.

roxie/ccd/ccdfile.cpp
```diff
@@ -895,7 +897,7 @@ class CRoxieFileCache : implements IRoxieFileCache, implements ICopyFileProgress
             ~AutoDisconnector() { if (f) disconnectRemoteFile(f); }
         private:
             IFile *f;
-        } autoDisconnector(f, autoDisconnect);
+        } autoDisconnector(f, autoDisconnect && useRemoteResources);
```
ghalliday, line 900: Took a while to understand this change.  I assume it is because this is called from checkOpen and you are just about to copy the file, so you don't want it closed.  It would benefit from a comment.

ghalliday (approved):
@richardkchapman looks ok, but would benefit from an extra comment

roxie/ccd/ccdfile.cpp
richardkchapman: Added a comment

richardkchapman:
@ghalliday Please merge

PR #15325, opened 2021-09-08
HPCC-26487 Read query dlls directly in cloud

jakesmith:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dali/base/dacoven.cpp
+++ b/dali/base/dacoven.cpp
@@ -936,6 +936,7 @@ ICoven &queryCoven()
 {
     if (coven==NULL)
     {
+        PrintStackReport();
         Owned<IException> e = MakeStringException(-1, "No access to Dali - this normally means a plugin call is being called from a thorslave");
         EXCLOG(e, NULL);
         throw e.getClear();
--- a/thorlcr/graph/thgraphmaster.hpp
+++ b/thorlcr/graph/thgraphmaster.hpp
@@ -48,7 +48,9 @@ interface IJobManager : extends IInterface
     virtual void setWuid(const char *wuid, const char *cluster=NULL) = 0;
     virtual IDeMonServer *queryDeMonServer() = 0;
     virtual void fatal(IException *e) = 0;
+#ifndef _CONTAINERIZED
     virtual void addCachedSo(const char *name) = 0;
+#endif
     virtual void updateWorkUnitLog(IWorkUnit &workunit) = 0;
 };
 
--- a/thorlcr/master/thmastermain.cpp
+++ b/thorlcr/master/thmastermain.cpp
@@ -66,7 +66,9 @@
 #include "thexception.hpp"
 #include "thmem.hpp"
 
+#ifndef _CONTAINERIED
 #define DEFAULT_QUERY_SO_DIR "sodir"
+#endif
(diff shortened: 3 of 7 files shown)


dali/base/dacoven.cpp
```diff
@@ -936,6 +936,7 @@ ICoven &queryCoven()
 {
     if (coven==NULL)
     {
+        PrintStackReport();
```
jakesmith, line 939: not directly related, but added to debug something I hit when implementing this PR, and think it is worth leaving in.

richardkchapman: No danger of crating excessive tracing? Probably not if this sutuation is always fatal...

jakesmith: Should always be fatal.
It would still be useful if in _DEBUG only..
leave it or make it _DEBUG only?

thorlcr/master/thmastermain.cpp
```diff
@@ -836,6 +838,16 @@ int main( int argc, const char *argv[]  )
...
         roxiemem::setTotalMemoryLimit(gmemAllowHugePages, gmemAllowTransparentHugePages, gmemRetainMemory, ((memsize_t)mmemSize) * 0x100000, 0, thorAllocSizes, NULL);
 
+        char thorPath[1024];
+        if (!GetCurrentDirectory(1024, thorPath))
+        {
+            OERRLOG("ThorMaster::main: Current directory path too big, setting it to null");
+            thorPath[0] = 0;
+        }
+        unsigned l = strlen(thorPath);
+        if (l) { thorPath[l] = PATHSEPCHAR; thorPath[l+1] = '\0'; }
```
ghalliday, line 848: not new code, but potentially buffer overflow error if path is 1023 characters long.

thorlcr/graph/thgraphmaster.hpp
```diff
@@ -48,7 +48,9 @@ interface IJobManager : extends IInterface
     virtual void setWuid(const char *wuid, const char *cluster=NULL) = 0;
     virtual IDeMonServer *queryDeMonServer() = 0;
     virtual void fatal(IException *e) = 0;
+#ifndef _CONTAINERIZED
```
ghalliday, line 51: I wonder if it is better to include this in the container build for a few reasons
- code is simpler, and compile errors are caught for both versions
- we might want to store dlls on storage that can't be mounted - then there would be a need to copy locally again.
- might we want to copy locally anyway to avoid re-read costs, and ensure it is read all in one go.  At least explore the relative performance it was written to a local nvme drive.
-

richardkchapman: If dlls are on storage that can't be mounted / it cost money to keep reading them from, then we would probably want to rethink Roxie behaviour too...

richardkchapman (approved):

thorlcr/graph/thgraphmaster.hpp
ghalliday: Yes we probably should...

PR #15335, opened 2021-09-09
HPCC-26477 Stop returning schemaLocation by default

asselitx:
Change the default behavior to omit returning the schemaLocation for normal
method responses (HTTP OK). It is returned unconditionally in Soap Fault
responses.

Correctly encode the value when it is returned. Change the returned value to
conform to the standard, which is a list of URI pairs of namespace and schema
location. The hostname of the location URI is pulled from the
wsdlServiceAddress attribute of the service binding, and if not present the
schemaLocation is omitted.

Add a DESDL binding configuration attribute at
Definition[@returnSchemaLocationOnOK] to enable schemaLocation return in normal
method responses.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Cloud-compatibility
  - [x] Premature optimization

## Smoketest:

### DIFF
--- a/esp/bindings/SOAP/Platform/soapbind.cpp
+++ b/esp/bindings/SOAP/Platform/soapbind.cpp
@@ -87,26 +87,13 @@ const char * CHttpSoapBinding::getTransportType()
     return "http";
 }
 
-static CSoapFault* makeSoapFault(CHttpRequest* request, IMultiException* me, const char *ns)
+static CSoapFault* makeSoapFault(CHttpRequest* request, IMultiException* me, const char *ns, const char* schemaLocation)
 {
-    const char* svcName = request->queryServiceName();
-    if (svcName && *svcName)
+    if (!isEmptyString(schemaLocation))
     {
-        const char* method = request->queryServiceMethod();
-        StringBuffer host;
-        const char* wsdlAddr = request->queryParameters()->queryProp("__wsdl_address");
-        if (wsdlAddr && *wsdlAddr)
-            host.append(wsdlAddr);
-        else
-        {
-            host.append(request->queryHost());
-            if (request->getPort()>0)
-              host.append(":").append(request->getPort());
-        }
-        
-        VStringBuffer ns_ext("xmlns=\"%s\""
-            " xsi:schemaLocation=\"%s %s/%s/%s?xsd\"", 
-            ns, ns, host.str(), svcName, method ? method : "");
+        StringBuffer ns_ext;
+        appendXMLAttr(ns_ext, "xmlns", ns, nullptr, true);
+        appendXMLAttr(ns_ext, "schemaLocation", schemaLocation, "xsi", true);
         return new CSoapFault(me, ns_ext);
     }
 
@@ -164,7 +151,10 @@ int CHttpSoapBinding::onSoapRequest(CHttpRequest* request, CHttpResponse* respon
         catch (IMultiException* mex)
         {
             StringBuffer ns;
-            soapFault.setown(makeSoapFault(request,mex, generateNamespace(*request->queryContext(), request, request->queryServiceName(), request->queryServiceMethod(), ns).str()));
+            StringBuffer schemaLocation;
+            generateNamespace(*request->queryContext(), request, request->queryServiceName(), request->queryServiceMethod(), ns);
+            getSchemaLocation(*ctx, request, ns, schemaLocation);
+            soapFault.setown(makeSoapFault(request,mex, ns.str(), schemaLocation.str()));
             //SetHTTPErrorStatus(mex->errorCode(),response);
             SetHTTPErrorStatus(500,response);
             StringBuffer errMessage;
@@ -176,9 +166,12 @@ int CHttpSoapBinding::onSoapRequest(CHttpRequest* request, CHttpResponse* respon
         catch (IException* e)
         {
             StringBuffer ns;
+            StringBuffer schemaLocation;
+            generateNamespace(*request->queryContext(), request, request->queryServiceName(), request->queryServiceMethod(), ns);
+            getSchemaLocation(*ctx, request, ns, schemaLocation);
             Owned<IMultiException> mex = MakeMultiException("Esp");
             mex->append(*e); // e is owned by mex
-            soapFault.setown(makeSoapFault(request,mex, generateNamespace(*request->queryContext(), request, request->queryServiceName(), request->queryServiceMethod(), ns).str()));
+            soapFault.setown(makeSoapFault(request,mex, ns.str(), schemaLocation.str()));
             SetHTTPErrorStatus(500,response);
             StringBuffer errMessage;
             ctx->addTraceSummaryValue(LogMin, "msg", e->errorMessage(errMessage).str(), TXSUMMARY_GRP_ENTERPRISE);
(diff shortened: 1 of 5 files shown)


timothyklemm (changes requested):
I don't believe our current value for schemaLocation is valid. First, `&` is only valid in an attribute value if part of an entity; we don't use it as an entity. Second, the value is defined as a non-empty list of string pairs, with the first of each pair being a namespace URI and the second being an actual location; the `noNamespaceSchemaLocation` allows a single location to be given, and select elements may allow `schemaLocation` to also accept a single location when paired with a separate `namespace` attribute.

As for the usefulness of the IP address in the URL, it isn't helpful. The binding's WSDL address is preferred, and the IP is only included as a last resort. Might need to consider whether a single static WSDL address configuration makes sense, or if we should be looking at dynamically defined values.

We may want to consider replacing the Boolean attribute with a textual selector allowing for more interpretations of this value. We could have distinct selector values for legacy behavior, value omission, and standard-compliant values. To be clear, I'm only suggesting that we anticipate standard-compliance as an additional option at this time.

asselitx:
As per conversation with Tony and Tim's comments I've pushed an update. The value won't be returned by default because the current value is not compliant and is not correctly XML encoded. Because of that, and because the value when returned does use the wsdlServiceAddress, I'm keeping a simple boolean to toggle.

esp/bindings/SOAP/Platform/soapbind.cpp
```diff
@@ -87,30 +87,19 @@ const char * CHttpSoapBinding::getTransportType()
...
-            " xsi:schemaLocation=\"%s %s/%s/%s?xsd\"", 
-            ns, ns, host.str(), svcName, method ? method : "");
+        StringBuffer ns_ext;
+        appendXMLAttr(ns_ext, "xmlns", ns, nullptr, true);
+        appendXMLAttr(ns_ext, "schemaLocation", schemaLocation, "xsi", true);
         return new CSoapFault(me, ns_ext);
     }
-
-    return new CSoapFault(me);
+    else
```
afishbeck, line 99: You don't need to add the else here because the "if" above returns from the function.

esp/services/esdl_svc_engine/esdl_binding.hpp
```diff
@@ -378,6 +379,8 @@ class EsdlBindingImpl : public CHttpSoapBinding
         return m_isAttached;
     }
 
+
```
afishbeck, line 382: Don't need these extra lines.

esp/bindings/SOAP/Platform/soapbind.cpp
```diff
@@ -164,7 +153,10 @@ int CHttpSoapBinding::onSoapRequest(CHttpRequest* request, CHttpResponse* respon
         catch (IMultiException* mex)
         {
             StringBuffer ns;
-            soapFault.setown(makeSoapFault(request,mex, generateNamespace(*request->queryContext(), request, request->queryServiceName(), request->queryServiceMethod(), ns).str()));
+            StringBuffer schemaLocation;
+            generateNamespace(*request->queryContext(), request, request->queryServiceName(), request->queryServiceMethod(), ns);
+            getSchemaLocation(*ctx, request, ns, schemaLocation);
+            soapFault.setown(makeSoapFault(request,mex, ns.str(), schemaLocation.str()));
```
afishbeck, line 159: @asselitx does scapps always return the schemaLocation in a SoapFault?

afishbeck (commented):
@asselitx some comments.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -2262,7 +2267,9 @@ int EsdlBindingImpl::onGetInstantQuery(IEspContext &context,
                     StringBuffer ns, schemaLocation;
                     generateNamespace(context, request, srvdef->queryName(), mthdef->queryName(), ns);
 
-                    getSchemaLocation(context, request, schemaLocation);
+                    if(m_pESDLService->m_returnSchemaLocation)
+                        getSchemaLocation(context, request, ns, schemaLocation);
```
timothyklemm, line 2271: Unless there's a use case where we need the schema location regardless of the configuration value, I would prefer `getSchemaLocation` to include the configuration check instead of expecting every use of the method to wrap the call with a check.

I suppose this depends on the answer to Tony's question about soap faults unconditionally including the schema.

esp/bindings/SOAP/Platform/soapbind.cpp
asselitx: Yes

esp/services/esdl_svc_engine/esdl_binding.cpp
asselitx: I wasn't sure if we'd really need/want the flexibility to have three states: omit from responses only/always return/omit from responses and faults. I opted to keep the behavior consistent with scapps in that the soap fault always returns the schemaLocation, but if you all think there should be even more flexibility it could be changed.

timothyklemm (approved):
Code is fine. You might want to revisit the commit message. Stating that the value doesn't have value always suggests it has value sometimes. This suggestion calls the solution of providing the value always or never into doubt.

I'm not suggesting that a per-request option is required. I'm just requesting that the commit message not imply that it should be.

asselitx:
Squashed commits and updated final commit message.

esp/services/esdl_svc_engine/esdl_binding.cpp
afishbeck: This a big improvement.  Any more flexibility would be nice but a low priority.  We can always improve later.

esp/services/esdl_svc_engine/esdl_binding.hpp
```diff
@@ -109,6 +109,7 @@ class EsdlServiceImpl : public CInterface, implements IEspService
     StringBuffer                m_serviceNameSpaceBase;
     StringAttr                  m_namespaceScheme;
     bool                        m_usesURLNameSpace;
+    bool                        m_returnSchemaLocationOnOK;
```
afishbeck, line 112: For new values we should really be initializing them here.. our code will improve that way over time.  Can then remove the setting in init() assuming init() is only called once.  If not, you can leave it there as well.

afishbeck (commented):
One more trivial comment, otherwise looks good.

asselitx:
Need to test Tony's recommended change but the Cert roxies are down. Converting to draft to avoid accidental merge until tests complete.

asselitx:
Tony's initialization changes applied, tested and everything squashed and rebased onto candidate-8.4.x.

esp/services/esdl_svc_engine/esdl_binding.hpp
```diff
@@ -109,6 +109,7 @@ class EsdlServiceImpl : public CInterface, implements IEspService
     StringBuffer                m_serviceNameSpaceBase;
     StringAttr                  m_namespaceScheme;
     bool                        m_usesURLNameSpace;
+    bool                        m_returnSchemaLocationOnOK{false};
```
afishbeck, line 112: Interesting I had to look up that syntax to see it was a valid alternative.
We usually use "bool m_returnSchemaLocationOnOK=false;"

afishbeck (commented):
@asselitx one more trivial comment, otherwise looks good.

PR #15337, opened 2021-09-10
HPCC-26500 Ensure data, lz and spill planes mount and checked

jakesmith:
Spill plane needed to be mounted and checked (if forcePermissions).
Also limit which planes are mounted in different contexts,
e.g. dfuserver and eclccserver never need "spill" mounts.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/helm/hpcc/templates/dali.yaml
+++ b/helm/hpcc/templates/dali.yaml
@@ -24,10 +24,28 @@
 {{ range $dali := $.Values.dali -}}
 {{- if not $dali.disabled -}}
 {{- $env := concat ($.Values.global.env | default list) (.env | default list) -}}
-{{- $commonCtx := dict "root" $ "me" $dali "env" $env }}
+{{- $daliPlaneIncludeCategories := list "dali" -}}
+{{- $commonCtx := dict "root" $ "me" $dali "includeCategories" $daliPlaneIncludeCategories "env" $env -}}
 {{- $sashaServices := $dali.services | default dict -}}
-{{- $daliSecretsCategories := list "system" }}
-{{- $tmpDaliScope := dict "aggregateSashaSecretsCategories" list "aggregateSashaAccess" list -}}
+{{- $daliSecretsCategories := list "system" -}}
+{{- $tmpDaliScope := dict "aggregateSashaSecretsCategories" list "aggregatePlaneCategories" $daliPlaneIncludeCategories "aggregateSashaNamedPlanes" list -}}
+{{- if not ($sashaServices.disabled) -}}
+ {{- range $sashaName, $_sasha := $dali.services -}}
+  {{- $sasha := ($_sasha | default dict) -}}
+  {{- if not $sasha.disabled -}}
+   {{- $_ := set $sasha "name" $sashaName -}}
+   {{- $sashaAccess := splitList " " (include "hpcc.getSashaServiceAccess" $sasha) -}}
+   {{- $sashaPlaneCategories := list (include "hpcc.getSashaPlanesFromAccess" (dict "access" $sashaAccess)) -}}
+   {{- $_ := set $tmpDaliScope "aggregatePlaneCategories" ((concat $tmpDaliScope.aggregatePlaneCategories $sashaPlaneCategories) | uniq) -}}
+   {{- $sashaServiceSecretsCategories := concat ((or (has "data" $sashaAccess) (has "dalidata" $sashaAccess)) | ternary (list "storage") list) $daliSecretsCategories -}}
+   {{- $_ := set $tmpDaliScope "aggregateSashaSecretsCategories" ((concat $tmpDaliScope.aggregateSashaSecretsCategories $sashaServiceSecretsCategories) | uniq) -}}
+   {{- if hasKey $sasha "plane" -}}
+    {{- $sashaStoragePlane := $sasha.plane | default (include "hpcc.getFirstPlaneForCategory" (dict "root" $ "category" "sasha")) -}}
+    {{- $_ := set $tmpDaliScope "aggregateSashaNamedPlanes" ((append $tmpDaliScope.aggregateSashaNamedPlanes $sashaStoragePlane) | uniq) -}}
+   {{- end -}}
+  {{- end -}}
+ {{- end -}}
+{{- end }}
 apiVersion: apps/v1
 kind: Deployment
 metadata:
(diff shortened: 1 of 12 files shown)


jakesmith:
@ghalliday - please review.
NB: this has left eclccserver (and eclccscheduler) only mounting "data" (not "lz") .. is there ever a case where eclccserver should have access to "lz" planes?

richardkchapman:
I'm not convinced eclserver should have access to any plane other than dll - and I don't see why eclscheduler needs access to any planes at all.

ghalliday (commented):
@jakesmith the change looks like it will improve things, but you seem to be go ing away form having a single list defined in the component yaml (e.g., removing it from dfuserver).

In general I think we need to go the other way - if there are multiple storage planes using the same pvc (but used for different categories) then we need to only mount/check them once.

E.g., 
```
{{ include "hpcc.addDataLZSpillVolumeMount" . | indent 12 }}
{{ include "hpcc.addDllVolumeMount" . | indent 12 }}

```
could cause problems if the same pvc/prefix was used for both,

jakesmith:
> I'm not convinced eclserver should have access to any plane other than dll - and I don't see why eclscheduler needs access to any planes at all.

I wondered why it did, thought might be something to do (in eclccserver case) with it accessing logical files at compile time, and indirectly accessing their physical parts (e.g. for size??).

@ghalliday - do you know why eclccserver was setup to mount "data" (and "lz") ?

ghalliday:
> > I'm not convinced eclserver should have access to any plane other than dll - and I don't see why eclscheduler needs access to any planes at all.
> 
> I wondered why it did, thought might be something to do (in eclccserver case) with it accessing logical files at compile time, and indirectly accessing their physical parts (e.g. for size??).
> 
> @ghalliday - do you know why eclccserver was setup to mount "data" (and "lz") ?

No I can't think of a reason.  I don't think eclscheduler needs access to any planes - it gets everything it needs from the dali workunit.

jakesmith:
@ghalliday - have rationilized the plane categories/named planes in various places to single lists (and used addVolume*).
Please see new commits.

jakesmith:
> @ghalliday - have rationilized the plane categories/named planes in various places to single lists (and used addVolume*).
> Please see new commits.

Seem I forgot to refresh the github review status on this one too.

helm/hpcc/templates/localroxie.yaml
```diff
@@ -44,7 +44,7 @@ data:
 {{- if not $roxie.disabled  -}}
 {{- $env := concat ($.Values.global.env | default list) (.env | default list) -}}
 {{- $secretsCategories := list "system" "ecl-user" "ecl" "storage" }}
-{{- $commonCtx := dict "root" $ "me" $roxie "secretsCategories" $secretsCategories "env" $env }}
+{{- $commonCtx := dict "root" $ "me" $roxie "includeCategories" (list "lz" "data" "spill") "secretsCategories" $secretsCategories "env" $env }}
```
ghalliday, line 47: Shouldn't this include "dll", and the call to addDllVolumeMount be removed (and all other places that use addDllVolumeMount/addDllVolume)?

helm/hpcc/templates/eclagent.yaml
```diff
@@ -88,7 +88,7 @@ data:
             workingDir: /var/lib/HPCCSystems
             volumeMounts:
 {{ include "hpcc.addConfigMapVolumeMount" .me | indent 12 }}
```
ghalliday, line 90: I am wondering if something this sequence should be included by all components, and a similar one for the volumes:

```
{{ include "hpcc.addConfigMapVolumeMount" .me | indent 12 }}
{{ include "hpcc.addVolumeMounts" . | indent 12 }}
{{ include "hpcc.addSecretVolumeMounts" . | indent 12 }}
{{ include "hpcc.addCertificateVolumeMount" (dict "root" .root "name" .me.name "component" $apptype) | indent 12 }}
```

?

helm/hpcc/templates/dali.yaml
```diff
@@ -64,18 +81,8 @@ spec:
...
-      {{- $sasha := ($_sasha | default dict) -}}
-      {{- if and (not $sasha.disabled) (hasKey $sasha "plane") }}
-        {{- $sashaStoragePlane := $sasha.plane | default (include "hpcc.getFirstPlaneForCategory" (dict "root" $ "category" "sasha")) }}
-        {{- $_ := append $sashaPlanes $sashaStoragePlane }}
-      {{- end }}
-      {{- end }}
-      {{- end }}
       initContainers: 
-        {{- include "hpcc.changePlaneMountPerms" (dict "root" $ "includeCategories" (list "dali") "includeNames" (uniq $sashaPlanes)) | indent 8 }}
+      {{- include "hpcc.changePlaneMountPerms" (dict "root" $ "includeCategories" $tmpDaliScope.aggregatePlaneCategories "includeNames" $tmpDaliScope.aggregateSashaNamedPlanes) | indent 6 }}
```
ghalliday, line 85: Slightly suprised this works with the leading {{- and indent.  Would nindent be better?  Is the change from 8 to 6 correct?

ghalliday (commented):
@jakesmith looks good - and it got smaller.
A couple of questions/suggests for further rationalisation.

helm/hpcc/templates/localroxie.yaml
jakesmith: I thought about it, but deliberately stopped at going that far with changes, and only focussed on "data" planes (lz, data, spill), leaving helpers in place/use.
But I could go further.

helm/hpcc/templates/eclagent.yaml
jakesmith: quite possibly, maybe next iteration of rationalization.

This PR was primarily supposed to be a bug fix for the "spill" which wasn't being handled properly, but it's creeping..

helm/hpcc/templates/dali.yaml
jakesmith: I am slightly surprised to now you point it out, {{- same as before, but.. yes I think needs to be | indent.

8 -> 6 was intentinal, either 'works' but indenting the list at same level is more conventional (and same as 'containers' indendentation below it)

Not new or related to this PR, but I'll change to nindent.

jakesmith: ah I see why it worked, by luck.. it was because the helper func. changeMountPerms begins with a block of comments.
So this (and other places) were wrong and swallowed up the newline, meaning the 1st comment of changeMountPerms anded up on the initContainers: line, which luckily didn't cause parsing problems.

ghalliday (approved):
@jakesmith looks good.  Please squash and then I'll scan once more/merge.

jakesmith:
@ghalliday - squashed and rebased, and that solved the helm linting issues.

ghalliday (approved):

PR #15340, opened 2021-09-13
HPCC-26270 PTREE support for JSON/YAML names with non-xml allowed chars

afishbeck:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jptree.cpp
+++ b/system/jlib/jptree.cpp
@@ -1461,7 +1461,7 @@ void PTree::setProp(const char *xpath, const char *val)
         if (!val)
             removeAttribute(xpath);
         else
-            setAttribute(xpath, val);
+            setAttribute(xpath, val, false);
     }
     else
     {
@@ -1605,7 +1605,7 @@ void PTree::addProp(const char *xpath, const char *val)
     if (!xpath || '\0' == *xpath)
         addLocal((size32_t)strlen(val)+1, val);
     else if (isAttribute(xpath))
-        setAttribute(xpath, val);
+        setAttribute(xpath, val, false);
     else if ('[' == *xpath)
     {
         aindex_t pos = getChildMatchPos(xpath);
@@ -1636,7 +1636,7 @@ void PTree::appendProp(const char *xpath, const char *val)
         StringBuffer newVal;
         getProp(xpath, newVal);
         newVal.append(val);
-        setAttribute(xpath, newVal.str());
+        setAttribute(xpath, newVal.str(), false);
     }
     else if ('[' == *xpath)
     {
@@ -1718,7 +1718,7 @@ void PTree::setPropInt64(const char * xpath, __int64 val)
     {
         char buf[23];
         numtostr(buf, val);
-        setAttribute(xpath, buf);
+        setAttribute(xpath, buf, false);
     }
     else
     {
@@ -1747,7 +1747,7 @@ void PTree::addPropInt64(const char *xpath, __int64 val)
     {
         char buf[23];
         numtostr(buf, val);
-        setAttribute(xpath, buf);
+        setAttribute(xpath, buf, false);
     }
     else if ('[' == *xpath)
     {
@@ -3619,10 +3619,18 @@ bool LocalPTree::removeAttribute(const char *key)
     return true;
 }
 
-void LocalPTree::setAttribute(const char *key, const char *val)
+void LocalPTree::setAttribute(const char *inputkey, const char *val, bool encoded)
 {
-    if (!key)
+    if (!inputkey)
         return;
+    const char *key = inputkey;
+    if (encoded)
+    {
+        if (*key!='~')
+            encoded=false;
+        else
+            key++;
+    }
     if (!validateXMLTag(key+1))
         throw MakeIPTException(-1, "Invalid xml attribute: %s", key);
     if (!val)
@@ -3639,8 +3647,8 @@ void LocalPTree::setAttribute(const char *key, const char *val)
     {
         attrs = (AttrValue *)realloc(attrs, (numAttrs+1)*sizeof(AttrValue));
         v = new(&attrs[numAttrs++]) AttrValue;  // Initialize new AttrValue
-        if (!v->key.set(key))
-            v->key.setPtr(isnocase() ? AttrStr::createNC(key) : AttrStr::create(key));
+        if (!v->key.set(inputkey)) //AttrStr will not return encoding marker when get() is called
+            v->key.setPtr(isnocase() ? AttrStr::createNC(inputkey) : AttrStr::create(inputkey));
     }
     if (arrayOwner)
     {
@@ -3812,7 +3820,7 @@ void CAtomPTree::freeAttrArray(AttrValue *a, unsigned n)
     }
 }
 
-void CAtomPTree::setAttribute(const char *key, const char *val)
+void CAtomPTree::setAttribute(const char *key, const char *val, bool encoded)
 {
     if (!key)
         return;
@@ -3864,6 +3872,9 @@ void CAtomPTree::setAttribute(const char *key, const char *val)
         v = &newattrs[numAttrs];
         if (!v->key.set(key))
             v->key.setPtr(attrHT->addkey(key, isnocase()));
+        //shared via atom table, may want to add this later... escaped and unescaped versions should be considered unique
+        //if (encoded)
+        //    v->key.setEncoded();
         if (!v->value.set(val))
             v->value.setPtr(attrHT->addval(val));
         numAttrs++;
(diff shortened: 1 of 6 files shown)


afishbeck:
@ghalliday @jakesmith @richardkchapman Here's an initial draft.  I have a bit more testing to do, but would like feedback on the general approach ASAP so that I can make any necessary changes.

system/jlib/jptree.cpp
```diff
@@ -9481,3 +9542,341 @@ void copyPropIfMissing(IPropertyTree & target, const char * targetName, IPropert
...
+{
+    if (!tree || isEmptyString(name))
+        return false;
+    const PTree *pt = static_cast<const PTree*>(tree);
+    return pt->isAttributeNameEncoded(name);
+}
+
+static void decodePTreeNameHexEncoding(StringBuffer &s, const char *&input, unsigned &remaining)
+{
+    byte ch = 0;
```
ghalliday, line 9787: I'm not sure how good code optimizations are these days.  But I think you might want to assign input, remaining to local variables at the head of the function, and assign them to the reference parameters as the end. Otherwise all references to input be indirect.

system/jlib/jptree.ipp
```diff
@@ -423,8 +423,7 @@ struct PtrStrUnion
     inline PtrStrUnion<PTR>() : ptr(nullptr) {}
     inline bool isPtr() const
     {
-        return (flag&1) == 0;
-    }
+        return (flag&1) == 0;    }
```
ghalliday, line 426: trivial: deleted newline

system/jlib/jptree.ipp
```diff
@@ -1024,5 +1051,77 @@ public:
...
+    virtual void beginNode(const char *tag, bool arrayitem, offset_t startOffset) override
+    {
+        if (rootProvided)
+        {
+            currentNode = root;
+            rootProvided = false;
+        }
+        else
+        {
+            StringAttr encoded;
```
ghalliday, line 1074: As before, simpler as a StringBuffer?

ghalliday (commented):
@afishbeck some minor comments.  Looks comprehensive.  Will need @jakesmith to carefully review as well.

testing/unittests/jlibtests.cpp
```diff
@@ -2113,20 +2113,129 @@ element1: !el scalarvalue
...
+  "but_elements_support_it_出售耐久": "no encode",
+  "出\n售耐/久@Aa_a": {
+   "@\naa_bb食品并/林": "atvalue",
+   "": "noname",
+   "abc/d @ "e'f": "value"
+  }
+ }
+})!!";
+
+            static constexpr const char * toYAMLOutput = R"!!(Nothing_ToEnodeHere:
```
richardkchapman, line 2202: Was this supposed to say "NothingToEncodeHere" ?

(and in various other places)?

richardkchapman (commented):
One trivial issue

system/jlib/jptree.cpp
```diff
@@ -9311,18 +9333,30 @@ class YAMLEmitter
 
 static void _toYAML(const IPropertyTree *tree, YAMLEmitter &yaml, byte flags, bool root=false, bool isArrayItem=false)
 {
-    const char *name = tree->queryName();
-    if (!root && !isArrayItem)
+    bool hiddenRootArrayObject = false;
+    StringAttr decoded;
```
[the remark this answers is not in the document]
afishbeck: @ghalliday My thinking was that not needing to decode would be the case 99% of the time and that instantiating a StringAttr was much more efficient in those cases.  In the rare case where I need to encode I use an adapter and take on the extra overhead.  I'm all for simplifying the code if you don't think it helps.

system/jlib/jptree.ipp
afishbeck: Same idea here.  I'll happily simplify the code after you comment on my question about what I though was a slight efficiency gain for 99% of cases.

testing/unittests/jlibtests.cpp
richardkchapman: It was the "Enode" rather than "Encode" that I was commenting on

system/jlib/jptree.cpp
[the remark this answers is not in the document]
ghalliday: A good point, I think you are right.  Worth a comment to explain why it is there.

I also experimented with inlining the constructor for StringBuffer.  It did improve the code, but increased the code size by ~16bytes, so I suspect that shouldn't be done.

system/jlib/jptree.cpp
```diff
@@ -9481,3 +9542,341 @@ void copyPropIfMissing(IPropertyTree & target, const char * targetName, IPropert
...
+}
+
+const char *findFirstInvalidPTreeNameChar(const char *ch, unsigned remaining)
+{
+    if (!ch)
+        return nullptr;
+    if (remaining==0 || *ch=='\0') //empty name still needs to trigger a special encoding
+        return ch;
+    bool isatt = ('@'==*ch);
+    if (isatt)
```
jakesmith, line 9580: avoid double condition? e.g.
```
     bool isAtt;
     if ('@' == *ch)
     {
         isAtt = true;
         ch++;
         remaining--;
     }
     else
          isAtt = false;
```

compiler may optimize to avoid double test anyway.

system/jlib/jptree.cpp
```diff
@@ -9481,3 +9542,341 @@ void copyPropIfMissing(IPropertyTree & target, const char * targetName, IPropert
...
+    {
+        remaining -= chlen;
+        ch += chlen;
+        first = false;
+        return s;
+    }
+    switch (*ch)
+    {
+        case '_':
+            appendPTreeNameSpecialEncoding(s, "_", hexsequence);
```
jakesmith, line 9652: may be worth a variety of appendPTreeNameSpecialEncoding that doesn't check isEmptyString(enc) on each call to use here, since it is exlpicitly not empty.

system/jlib/jptree.cpp
```diff
@@ -9481,3 +9542,341 @@ void copyPropIfMissing(IPropertyTree & target, const char * targetName, IPropert
...
+        if (realStart)
+            startEncoding = (const char *) realStart;
+        unsigned skipSize = startEncoding-value;
+        if (skipSize > size)
+            return s;
+        s.append(skipSize, value);
+        value = startEncoding;
+        size -= skipSize;
+    }
+    bool hexsequence = false;
```
jakesmith, line 9720: trivial: camelCase

system/jlib/jptree.cpp
```diff
@@ -9481,3 +9542,341 @@ void copyPropIfMissing(IPropertyTree & target, const char * targetName, IPropert
...
+    while (remaining)
+    {
+        if ('_'!=*input)
+        {
+            s.append(*input);
+            input++;
+            remaining--;
+        }
+        else
+            decodePTreeNameEncoding(s, input, remaining);
```
jakesmith, line 9872: probably similar elsewhere, but wonder on cost of overhead of repeated calls to StringBuffer::append here.
Could you pre-reserve the estiated output size in the StringBuffer?

jakesmith (commented):
@afishbeck - looks good. Some v. minor comments only.
How does this effect performance on a large read/write where no encoding is involved?
And/or what is the hit when there is a significant amount of encoding required?

system/jlib/jptree.cpp
afishbeck: done, see added ensureCapacity() above.

afishbeck:
Updated based on review and expanded on test case.  Please review.

jakesmith (commented):
@afishbeck - difficult to see what's changed because squashed, but changes look good afaics.

Only concern is how it effects performance. Might be worth adding a couple of perf. unit tests and use to test before/after changes.

system/jlib/jptree.cpp
ghalliday: It looks like the whole set of functions are inlined anyway

ghalliday (approved):

ghalliday:
@afishbeck should this be non-draft now?  (I need to create a PR based on it, but I can do that regardless)

ghalliday:
@afishbeck also needs to now target 8.4.x

system/jlib/jptree.cpp
afishbeck: appendPTreeNameSpecialEncoding no longer checks isEmptyString().

afishbeck:
@ghalliday this is no longer a draft, I have rebased and retargeted.

richardkchapman (approved):

jakesmith (approved):
@afishbeck looks good

PR #15349, opened 2021-09-15
HPCC-26416 Use WU's WUResultMaxSizeMB option to read WU Result

wangkx:
WUResultMaxSizeMB is currently definable at the WsWorkunits service
level and effects all WU result requests. By this PR, an individual
workunit could override this limit by defining its own WUResultMaxSizeMB
using a Debug value.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/services/ws_workunits/ws_workunitsService.cpp
+++ b/esp/services/ws_workunits/ws_workunitsService.cpp
@@ -2845,6 +2845,7 @@ void CWsWorkunitsEx::getWsWuResult(IEspContext &context, const char *wuid, const
     if(!cw)
         throw MakeStringException(ECLWATCH_CANNOT_OPEN_WORKUNIT,"Cannot open workunit %s.",wuid);
     Owned<IConstWUResult> result;
+    __uint64 resultMaxSize = getWUResultMaxSize(cw);
 
     if (notEmpty(name))
         result.setown(cw->getResultByName(name));
@@ -2880,11 +2881,11 @@ void CWsWorkunitsEx::getWsWuResult(IEspContext &context, const char *wuid, const
     else
         rs.setown(resultSetFactory->createNewResultSet(result, wuid));
     if (!filterBy || !filterBy->length())
-        appendResultSet(context, mb, rs, name, start, count, total, bin, xsd, context.getResponseFormat(), result->queryResultXmlns(), wuResultMaxSize);
+        appendResultSet(context, mb, rs, name, start, count, total, bin, xsd, context.getResponseFormat(), result->queryResultXmlns(), resultMaxSize);
     else
     {
         Owned<INewResultSet> filteredResult = createFilteredResultSet(rs, filterBy);
-        appendResultSet(context, mb, filteredResult, name, start, count, total, bin, xsd, context.getResponseFormat(), result->queryResultXmlns(), wuResultMaxSize);
+        appendResultSet(context, mb, filteredResult, name, start, count, total, bin, xsd, context.getResponseFormat(), result->queryResultXmlns(), resultMaxSize);
     }
 
     wuState = cw->getState();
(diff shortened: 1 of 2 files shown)


esp/services/ws_workunits/ws_workunitsService.cpp
```diff
@@ -3198,14 +3199,48 @@ void CWsWorkunitsEx::getFileResults(IEspContext &context, const char *logicalNam
     if (!df)
         throw makeStringExceptionV(ECLWATCH_FILE_NOT_EXIST, "Cannot find file %s.", logicalName);
 
+    __uint64 resultMaxSize = getWUResultMaxSize(context, logicalName);
```
jakesmith, line 3202: Still not clear on this.
My comment from the JIRA:
>If they are asking for the content of a logical file without a wuid context, then that is different functionaliy, which by the sound of it, just happens to be implemented under 'workunit' service code,

What I meant by that is, if they looking at the content of a file, via any context than the user looking at a result of a workunit that happens to be a file, then the workunit limit should not be used.

e.g. if they browsing files under the file tab, then just because a file there happens to have been created by a workunit at some point in the past, it shouldn't apply whatever that limit was to the user that's viewing the file.

A limit on result size for viewing files in general outside of the context of specificially looking at results of a workunit should be handled separately as a different configuration option/different JIRA.

So, I think you need to pass in the limit to getFileResults explicitly only when it is being called for a workunit result.

esp/services/ws_workunits/ws_workunitsService.cpp
```diff
@@ -3198,14 +3199,48 @@ void CWsWorkunitsEx::getFileResults(IEspContext &context, const char *logicalNam
...
+        return wuResultMaxSize; //Use the default.
+
+    __int64 wuResultMaxSizeMB = cw->getDebugValueInt64("wuresultmaxsizemb", 0);
+    if (wuResultMaxSizeMB > 0)
+        return wuResultMaxSizeMB * 0x100000;
+    else
+        return wuResultMaxSize;
+}
+
+__int64 CWsWorkunitsEx::getWUResultMaxSize(IEspContext &context, const char *logicalName)
```
jakesmith, line 3226: given other comments, I am not sure either of these function should remain, at least in their current form.

jakesmith (changes requested):
@wangkx - please see comments

esp/services/ws_workunits/ws_workunitsService.cpp
wangkx: The getFileResults may be called when a user does the following:
1. In the WU Details page, click the Outputs tab.
2. In the File Name column, select a file name. A file tab is shown.
3. Inside the file tab, click the Contents tab.
If I understand your comment correctly, in this case, I do need to pass in the limit to getFileResults.
The getFileResults may also be called from outside of ECLWatch. A user does call it using WsWorkunits.WUResults. But, the user may pass in a logical file which is not created by ECL Workunit. That is why the getWUResultMaxSize has the code to check WUID.

jakesmith: It's very confusing that " WsWorkunits.WUResults" is used by clients to access all file content in general, i.e. nothing to do with wuids necessarily.

In discussion it's become apparent that Esp can't tell the difference between:
1. Viewing a wuid -> wuid resutls -> wuidresult1(which is a file)
2. Viewing a loigcal file (e.g. in eclwatch file tab), view contents.

i.e. (1) does not send a wuid.

So currently there's no way to make this feature apply to wuid results (inc. wuid file results) only.

If the client (e.g. eclwatch) is requesting a workunit result via WsWorkunits.WUResults it whould specify the wuid.
Then the esp service code can apply whatever limit was applied to that specific workunit.

wangkx: @GordonSmith if ECLWatch cannot specify WUID when viewing result via logical file name, can ECLWatch pass a flag indicating that the call is from ECLWatch ECL section (not from ECLWatch Files section)?

GordonSmith: @jakesmith / @wangkx 

I am a bit confused - I think the route a user took to ultimately land on one of these two pages:
*  http://localhost:8010/?NodeGroup=data&LogicalName=progguide::exampledata::people&Widget=ResultWidget
*  http://localhost:8010/?Wuid=W20210921-145544&Sequence=0&Widget=ResultWidget

Is immaterial, if ESP needs to know the WUID (Either ECL WU or DFU WU) associated with a logical file, it should either call WsDFU/DFUInfo or pull the info from Dali directly?

GordonSmith: @wangkx - here are the current URLs to view the paged content of a logical file (current ECL Watch and Modern ECL Watch):
* http://localhost:8010/?NodeGroup=data&LogicalName=progguide%3A%3Aexampledata%3A%3Apeople&Widget=ResultWidget
* http://localhost:8010/esp/files/index.html#/files/data/progguide::exampledata::people/Contents

In both cases the user has navigated directly there and not via any other pages.  I could do a DFUInfo call to get the WUID associated with the logical file (if there is one) and pass it to you, but it would be more sensible for that to happen your end?

PR #15362, opened 2021-09-17
HPCC-25964 Fix disk read/write stats

jakesmith:
Disk read and write stats. were wildly incorrect, due
to a bug in the way they were serialized.
Each time a worker collected disk stats, it was rolling them
up on to the previously collected stats.
Fix by ensuring that the total read/written is set (not merged)
into the statistic collection that is serialized back to the
manager.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jstats.cpp
+++ b/system/jlib/jstats.cpp
@@ -2288,6 +2288,23 @@ unsigned __int64 CRuntimeStatisticCollection::getSerialStatisticValue(StatisticK
     return value + convertMeasure(rawKind, kind, rawValue);
 }
 
+void CRuntimeStatisticCollection::set(const CRuntimeStatisticCollection & other, unsigned node)
+{
+    ForEachItemIn(i, other)
+    {
+        StatisticKind kind = other.getKind(i);
+        unsigned __int64 value = other.getStatisticValue(kind);
+        if (value)
+            setStatistic(kind, value, node);
+    }
+
+    CNestedRuntimeStatisticMap *otherNested = other.queryNested();
+    if (otherNested)
+    {
+        ensureNested().set(*otherNested, node);
+    }
+}
+
 void CRuntimeStatisticCollection::merge(const CRuntimeStatisticCollection & other, unsigned node)
 {
     ForEachItemIn(i, other)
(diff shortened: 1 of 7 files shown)


system/jlib/jstats.cpp
```diff
@@ -2288,6 +2288,23 @@ unsigned __int64 CRuntimeStatisticCollection::getSerialStatisticValue(StatisticK
     return value + convertMeasure(rawKind, kind, rawValue);
 }
 
+void CRuntimeStatisticCollection::set(const CRuntimeStatisticCollection & other, unsigned node)
```
ghalliday, line 2291: These changes look good, but suggest I have got the class structure wrong - I will merge yours first and then go back and fix it....

thorlcr/activities/thdiskbaseslave.cpp
```diff
@@ -468,8 +468,8 @@ void CDiskWriteSlaveActivityBase::close()
                 CriticalBlock block(outputCs);
                 // ensure it is released/destroyed after releasing crit, since the IFileIO might involve a final copy and take considerable time.
                 tmpFileIO.setown(outputIO.getClear());
+                mergeStats(closedPartFileStats, tmpFileIO, diskWriteRemoteStatistics);
```
ghalliday, line 471: Why did these move inside the critical section?  Thread safety?  As long as they don't set any nested stats then it is thread safe, and moving it inside prolongs the length the critsec is held.

I guess it might be to make sure that you don't lose any stats in a race condition when the file is closed, but I think that would generally be ok as long as the final stats are correct.

ghalliday (commented):
@jakesmith looks good.  One question about moving code into critical sections, and a bigger question about how I structured the classes.

thorlcr/activities/thdiskbaseslave.cpp
jakesmith: as you say, it could mean that as 1 part closes, and serializeStats enters the crit, it will have lost the stats from the closed part until the next call, meaning numbers go down temporarily.

I wouldn't think the CS held here (only during a close) is consequential timing wise

jakesmith: @ghalliday - can move that line outside, but I think it's more correct inside, if cost isn't overly concerning

ghalliday (approved):

PR #15375, opened 2021-09-20
HPCC-26392 Add support for npm generated package-lock.json files

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/ecl/hql/hqlcollect.cpp
+++ b/ecl/hql/hqlcollect.cpp
@@ -544,13 +544,26 @@ void FileSystemDirectory::processDependencies(IPropertyTree * dependTree, const 
 {
     bool onlyAllowSHA = isPackageLock;
     Owned<IPropertyTreeIterator> depends = dependTree->getElements(path);
+    StringBuffer decodedName;
     ForEach(*depends)
     {
         IPropertyTree & cur = depends->query();
         const char * name = cur.queryName();
-        // The special node __empty__ is added for an emptry string mapping "" =>, which is the entry for this package
-        if (isEmptyString(name) || strieq(name, "__empty__"))
+
+        // An empty string means this package - so ignore it.  It is specially encoded in the ptree.
+        if (isNullPtreeName(name, isPTreeNameEncoded(&cur)))
             continue;
+
+        //npm generates entries in the format "node_modules/package", so skip the leading "node_modules/"
+        //Note because the "tag" contains a / the name will be encoded.
+        const char * encodedNodeModules = "node__modules_f"; // "node_modules/"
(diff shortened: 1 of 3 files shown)


ghalliday:
This requires https://github.com/hpcc-systems/HPCC-Platform/pull/15340 to be merged for the code to work, but can be merged independently.
@afishbeck fyi

ecl/hql/hqlcollect.cpp
```diff
@@ -548,10 +548,16 @@ void FileSystemDirectory::processDependencies(IPropertyTree * dependTree, const
     {
         IPropertyTree & cur = depends->query();
         const char * name = cur.queryName();
+
         // The special node __empty__ is added for an emptry string mapping "" =>, which is the entry for this package
-        if (isEmptyString(name) || strieq(name, "__empty__"))
+        if (isEmptyString(name) || strieq(name, "__empty__") || strieq(name, "_0"))
```
shamser, line 553: This seems like something we'd be using in lots of different places.  I'm wondering if this should be jlib library function:

isXmlValueEmpty(name) or isYamlValueEmpty(name)

ecl/hql/hqlcollect.cpp
```diff
@@ -548,10 +548,16 @@ void FileSystemDirectory::processDependencies(IPropertyTree * dependTree, const
...
             continue;
 
+        //npm generates entries in the format "node_modules/package", so skip the leading "node_modules/"
+        const char * encodedNodeModules = "node__modules_f"; // "node_modules/"
```
shamser, line 557: I'm confused by this.  The comment  refers to skipping leading "node_modules/" but the code is skipping the leading "node__modules_f".  Please can you clarify what's happening here?

ghalliday: Tony's PR encodes expressions that could not previously be encoded as xml.  So a tag cannot contain / so that is expanded to _f.  Because the tag is encoded _ is also expanded to __.  So encodedNodeModules is the prefix it will have been encoded to.  There may be a potential bug if the module name contains an underscore.  I'll check/fix.

ecl/hql/hqlcollect.cpp
ghalliday: Now tony's PR is merged the middle entry can be removed.  I could add a function to hide the details.

shamser (approved):

ghalliday:
@richardkchapman now squashed

PR #15386, opened 2021-09-21
HPCC-26563 Fix sub dir per parts issues

jakesmith:
1) Roxie wasn't locating file parts locatated in their
own sub-directories
2) renamePhysicalParts was ignoring the subdirperpart flag
and placing the parts outside, causing renamed file parts
not to be found on read.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/common/thorhelper/roxiehelper.cpp
+++ b/common/thorhelper/roxiehelper.cpp
@@ -2399,7 +2399,7 @@ void ClusterWriteHandler::getPhysicalName(StringBuffer & name, const char * clus
 {
     Owned<IStoragePlane> plane = getDataStoragePlane(cluster, false);
     const char * prefix = plane ? plane->queryPrefix() : nullptr;
-    makePhysicalPartName(logicalName.get(), 1, 1, name, 0, DFD_OSdefault, prefix);
+    makePhysicalPartName(logicalName.get(), 1, 1, name, 0, DFD_OSdefault, prefix, false);
 }
 
 void ClusterWriteHandler::addCluster(char const * cluster)
--- a/ecl/eclagent/eclagent.cpp
+++ b/ecl/eclagent/eclagent.cpp
@@ -3184,7 +3184,7 @@ char *EclAgent::getFilePart(const char *lfn, bool create)
         expandLogicalName(full_lfn, lfn);
 
         StringBuffer physical;
-        makePhysicalPartName(full_lfn.str(), 1, 1, physical, false);//MORE: What do we do if local ?
+        makePhysicalPartName(full_lfn.str(), 1, 1, physical, 0, DFD_OSdefault, nullptr, false);//MORE: What do we do if local ?
 
         StringBuffer dir,base;
         splitFilename(physical.str(), &dir, &dir, &base, &base);
--- a/roxie/ccd/ccdfile.cpp
+++ b/roxie/ccd/ccdfile.cpp
@@ -1710,10 +1710,12 @@ public:
                 dlfn.clearForeign();
 #ifdef _CONTAINERIZED
             const char *defaultDir = defaultPlaneDirPrefix;
+            bool defaultDirPerPart = defaultPlaneDirPerPart;
 #else
             const char *defaultDir = nullptr;
+            bool defaultDirPerPart = false;
 #endif
-            makePhysicalPartName(dlfn.get(), partNo, numParts, localLocation, replicationLevel, DFD_OSdefault, defaultDir);
+            makePhysicalPartName(dlfn.get(), partNo, numParts, localLocation, replicationLevel, DFD_OSdefault, defaultDir, defaultDirPerPart);
         }
         Owned<ILazyFileIO> ret;
         try
--- a/thorlcr/mfilemanager/thmfilemanager.cpp
+++ b/thorlcr/mfilemanager/thmfilemanager.cpp
@@ -477,6 +477,7 @@ public:
             desc->queryProperties().setProp("@job", jobStr.str());
             desc->queryProperties().setProp("@owner", userStr.str());
 
+#ifndef _CONTAINERIZED
             // if supporting different OS's in CLUSTER this should be checked where addCluster called
             DFD_OS os = DFD_OSdefault;
             EnvMachineOS thisOs = queryOS(groups.item(0).queryNode(0).endpoint());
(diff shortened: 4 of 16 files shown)


richardkchapman:
Looks ok but not a fan of default parameters - in particular this means that all the other calls to this function have not been examined to see if they may have similar issues.

testing/regress/ecl-test.json
```diff
@@ -4,7 +4,7 @@
         "espIp" : "127.0.0.1",
         "espSocket": "8010",
         "dropzoneIp" : "127.0.0.1",
-        "dropzonePath" : "/var/lib/HPCCSystems/mydropzone/",
+        "dropzonePath" : "/home/jsmith/hpcc3Debug/var/lib/HPCCSystems/mydropzone/",
```
richardkchapman, line 7: I don't think this should have been checked in

richardkchapman (dismissed):
Need to remove incorrectly checked in file

testing/regress/ecl-test.json
jakesmith: removed

thorlcr/mfilemanager/thmfilemanager.cpp
```diff
@@ -539,11 +539,11 @@ class CFileManager : public CSimpleInterface, implements IThorFileManager
                     }
                 }
                 // places logical filename directory in 'dir'
-                makePhysicalPartName(logicalName, 0, 0, dir, false, os, planeDir.str());
+                getLFNDirectoryUsingBaseDir(dir, logicalName, planeDir.str());
```
richardkchapman, line 542: Containerized builds now fail with:

#16 139.0 /hpcc-dev/HPCC-Platform/thorlcr/mfilemanager/thmfilemanager.cpp: In member function ‘virtual IFileDescriptor* CFileManager::create(CJobBase&, const char*, StringArray&, IArrayOf<IGroup>&, bool, unsigned int, bool, unsigned int)’:
#16 139.0 /hpcc-dev/HPCC-Platform/thorlcr/mfilemanager/thmfilemanager.cpp:481:20: error: variable ‘os’ set but not used [-Werror=unused-but-set-variable]
#16 139.0   481 |             DFD_OS os = DFD_OSdefault;
#16 139.0       |                    ^~

richardkchapman (dismissed):

richardkchapman (approved):
Please squash

jakesmith:
@richardkchapman - squashed

common/thorhelper/roxiehelper.cpp
```diff
@@ -2399,7 +2399,7 @@ void ClusterWriteHandler::getPhysicalName(StringBuffer & name, const char * clus
 {
     Owned<IStoragePlane> plane = getDataStoragePlane(cluster, false);
     const char * prefix = plane ? plane->queryPrefix() : nullptr;
-    makePhysicalPartName(logicalName.get(), 1, 1, name, 0, DFD_OSdefault, prefix);
+    makePhysicalPartName(logicalName.get(), 1, 1, name, 0, DFD_OSdefault, prefix, false);
```
ghalliday, line 2402: Shouldn't this be passing in dirPerPart from the plane/default, or do roxie/hthor always not create a subdir?  If so it would be worth commenting.

ecl/eclagent/eclagent.cpp
```diff
@@ -3184,7 +3184,7 @@ char *EclAgent::getFilePart(const char *lfn, bool create)
         expandLogicalName(full_lfn, lfn);
 
         StringBuffer physical;
-        makePhysicalPartName(full_lfn.str(), 1, 1, physical, false);//MORE: What do we do if local ?
+        makePhysicalPartName(full_lfn.str(), 1, 1, physical, 0, DFD_OSdefault, nullptr, false);//MORE: What do we do if local ?
```
ghalliday, line 3187: similar question about never creating in a subdir.

roxie/ccd/ccdfile.cpp
```diff
@@ -1710,10 +1710,12 @@ class CRoxieFileCache : implements IRoxieFileCache, implements ICopyFileProgress
                 dlfn.clearForeign();
 #ifdef _CONTAINERIZED
             const char *defaultDir = defaultPlaneDirPrefix;
+            bool defaultDirPerPart = defaultPlaneDirPerPart;
 #else
             const char *defaultDir = nullptr;
+            bool defaultDirPerPart = false;
 #endif
-            makePhysicalPartName(dlfn.get(), partNo, numParts, localLocation, replicationLevel, DFD_OSdefault, defaultDir);
+            makePhysicalPartName(dlfn.get(), partNo, numParts, localLocation, replicationLevel, DFD_OSdefault, defaultDir, defaultDirPerPart);
```
ghalliday, line 1718: Doesn't this need to pass in a flag to indicate if there is a subdir per part for this file?  I think it would only be a problem if copying from a bare metal system.

ghalliday (commented):
@jakesmith a couple of questions.

common/thorhelper/roxiehelper.cpp
jakesmith: The feature is disabled for 1 part files, i.e. a file will only be published with this flag if it has >1 part.

ecl/eclagent/eclagent.cpp
jakesmith: 1 part files will not be created/published with the subdirperpart flag

roxie/ccd/ccdfile.cpp
jakesmith: Discussed with Richard, but I still not sure.

This is building the localLocation to copy to, or if useRemoteResources and copyResources are false (which I believe is the default in cloud), then no copy is involved.

But doesn't that mean that if the default roxie data plane is different (different prefix) from the plane of file it's looking up here, then the 'defaultDir' is also wrong here?

If so, shouldn't the directory used here to make the localLocation also be using the prefix from the file?
In which case, it wasn't going to work before this change if the planes differed, and yes I think the subdirperpart flag should also come from the file.

jakesmith:
@ghalliday @richardkchapman - please see replies

ghalliday (approved):

PR #15393, opened 2021-09-22
HPCC-25750 Update DFUInfo cost calculation to reflect access cost

shamser:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dali/base/dadfs.cpp
+++ b/dali/base/dadfs.cpp
@@ -177,21 +177,27 @@ static IPropertyTree *getEmptyAttr()
     return createPTree("Attr");
 }
 
-static double calcFileCost(const char * cluster, double sizeGB, double fileAgeDays)
+static double calcFileCost(const char * cluster, double sizeGB, double fileAgeDays, __int64 numDiskWrites, __int64 numDiskReads)
 {
     Owned<IPropertyTree> plane = getStoragePlane(cluster);
-    double atRestCost = 0.0;
+    Owned<IPropertyTree> global;
+    IPropertyTree * costPT = nullptr;
+
     if (plane && plane->hasProp("cost/@storageAtRest"))
     {
-        atRestCost = plane->getPropReal("cost/@storageAtRest", 0.0);
+        costPT = plane->queryPropTree("cost");
     }
     else
     {
-        Owned<IPropertyTree> global = getGlobalConfig();
-        atRestCost = global->getPropReal("cost/@storageAtRest", 0.0);
+        global.setown(getGlobalConfig());
+        costPT = global->queryPropTree("cost");
     }
-    double storageCostDaily = atRestCost * 12 / 365;
-    return storageCostDaily * sizeGB * fileAgeDays;
+    constexpr int accessPriceScalingFactor = 10000; // read/write pricing based on 10,000 operations
+    double atRestPrice = costPT->getPropReal("@storageAtRest", 0.0);
+    double readPrice =  costPT->getPropReal("@storageReads", 0.0);
+    double writePrice =  costPT->getPropReal("@storageWrites", 0.0);
+    double storageCostDaily = atRestPrice * 12 / 365;
+    return (storageCostDaily * sizeGB * fileAgeDays) + (readPrice * numDiskReads / accessPriceScalingFactor) + (writePrice * numDiskWrites / accessPriceScalingFactor);
 }
 
 RemoteFilename &constructPartFilename(IGroup *grp,unsigned partno,unsigned partmax,const char *name,const char *partmask,const char *partdir,unsigned copy,ClusterPartDiskMapSpec &mspec,RemoteFilename &rfn)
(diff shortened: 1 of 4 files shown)


dali/base/dadfs.cpp
```diff
@@ -4767,19 +4772,26 @@ protected: friend class CDistributedFilePart;
         getModificationTime(dt);
         double fileAgeDays = difftime(time(nullptr), dt.getSimple())/(24*60*60);
         double sizeGB = getDiskSize(true, false) / ((double)1024 * 1024 * 1024);
+        const IPropertyTree *attrs = root->queryPropTree("Attr");
+        long long numDiskWrites = 0, numDiskReads;
```
ghalliday, line 4776: minor: long long is a bit ununusual.  We normally use __int64 or in this case possibly stats_type.

We should probably go through the entire code and use int64_t instead....

dali/base/dadfs.cpp
```diff
@@ -4767,19 +4772,26 @@ protected: friend class CDistributedFilePart;
...
+        if (attrs && attrs->hasProp("@maxSkew"))
```
ghalliday, line 4777: Why make this conditional?

dali/base/dadfs.cpp
ghalliday: numDiskReads is uninitialised.  Not sure why the compiler didn't catch it.

dali/base/dadfs.cpp
```diff
@@ -177,21 +177,26 @@ static IPropertyTree *getEmptyAttr()
...
-        atRestCost = global->getPropReal("cost/@storageAtRest", 0.0);
+        global.setown(getGlobalConfig());
+        costPT = global->queryPropTree("cost");
     }
+    double atRestCost = costPT->getPropReal("@storageAtRest", 0.0);
+    double readCost =  costPT->getPropReal("@storageReads", 0.0);
+    double writeCost =  costPT->getPropReal("@storageWrites", 0.0);
     double storageCostDaily = atRestCost * 12 / 365;
-    return storageCostDaily * sizeGB * fileAgeDays;
+    return (storageCostDaily * sizeGB * fileAgeDays) + (readCost * numDiskReads) + (writeCost * numDiskWrites);
```
ghalliday, line 199: haven't scaled by 10,000 for read/write costs

ghalliday (commented):
@shamser a few comments

dali/base/dadfs.cpp
shamser: vscode intellisense showed the return type as long long.  Probably, should have remembered this.

dali/base/dadfs.cpp
shamser: copy& paste mistake

dali/base/dadfs.cpp
```diff
@@ -177,21 +177,27 @@ static IPropertyTree *getEmptyAttr()
...
+        costPT = global->queryPropTree("cost");
     }
-    double storageCostDaily = atRestCost * 12 / 365;
-    return storageCostDaily * sizeGB * fileAgeDays;
+    constexpr int accessPriceScalingFactor = 10000; // read/write pricing based on 10,000 operations
+    double atRestPrice = costPT->getPropReal("@storageAtRest", 0.0);
+    double readPrice =  costPT->getPropReal("@storageReads", 0.0);
+    double writePrice =  costPT->getPropReal("@storageWrites", 0.0);
+    double storageCostDaily = atRestPrice * 12 / 365;
+    return (storageCostDaily * sizeGB * fileAgeDays) + (readPrice * numDiskReads / accessPriceScalingFactor) + (writePrice * numDiskWrites / accessPriceScalingFactor);
```
ghalliday, line 200: trivial: needed to check that the multiply would be done 1st, before the divide, otherwise it would lose precision.  Possibly better would be a double, or a value for multiplication.
double accessScalingFactor = 0.0000001;
correct, as it stands though.

helm/hpcc/values.schema.json
```diff
@@ -272,6 +272,14 @@
             "storageAtRest": {
                 "description": "storage cost (GiB per month)",
                 "type": "number"
+            },
+            "storageReads": {
+                "description": "cost per 10,000 read operations",
+                "type": "number"
+            },
+            "storageWrites": {
+                "description": "cost per 10,000 read operations",
```
ghalliday, line 281: copy and paste - "write" operations.

ghalliday (commented):
@shamser a couple of minor comments.  Please squash after any changes.

dali/base/dadfs.cpp
shamser: The order of precedence has multiplication and division same level and from left to right: https://en.cppreference.com/w/cpp/language/operator_precedence.   Compilers shouldn't re-order a multiply and divisions for the very reason you've mentioned.

ghalliday: Yes, that's why I said it was correct - but I had to double check the spec to be sure.

ghalliday (approved):

PR #15404, opened 2021-09-24
HPCC-26598 Allow eclcc --main with no source file supplied

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/ecl/eclcmd/eclcmd_common.cpp
+++ b/ecl/eclcmd/eclcmd_common.cpp
@@ -700,7 +700,12 @@ bool EclCmdWithEclTarget::finalizeOptions(IProperties *globals)
     if (!EclCmdCommon::finalizeOptions(globals))
         return false;
     if (!param.isEmpty())
-        optObj.set(param);
+    {
+        if (optTargetCluster.isEmpty())
+            optTargetCluster.set(param);
+        else
+            optObj.set(param);
+    }
     if (optObj.type == eclObjTypeUnknown)
     {
         if (optAttributePath.length())


ecl/eclcmd/eclcmd_common.cpp
```diff
@@ -700,7 +700,12 @@ bool EclCmdWithEclTarget::finalizeOptions(IProperties *globals)
     if (!EclCmdCommon::finalizeOptions(globals))
         return false;
     if (!param.isEmpty())
-        optObj.set(param);
+    {
+        if (optTargetCluster.isEmpty())
```
afishbeck, line 704: I'm surprised this scenario wasn't found sooner, but a lot of people still seem to use --target rather than a target parameter.

The target parameter is currently automatically used if there are 2+ parameters, so I suspect this scenario is because 
--main causes target to be the only parameter and it wasn't being detected.

This change could cause different the error to change from what it used to be if people forget to add the target, for example if they are used to using --target but forget.  But it might be ok.

So, it's up to you, but one option if you want avoid any potential backward compatibility issues, is to only setTargetCluster() if optAttributePath is set.

Something like:
if (!optAttributePath.isEmpty() && optTargetCluster.isEmpty())

ghalliday: @afishbeck I debated changing it as you suggest and decided the more correct fix was the one I introduced.  For instance if you have a local hthor directory (as I happened to), instead of getting an error about needing to supply a file or query you get an error that hthor is a directory, which is slightly confusing.
The help says:
`ecl run <target> <file> --name=<val> [--input=<file|xml>][--wait=<i>]`
So I think it makes sense for the 1st argument to be used as the target if no target has been supplied.

ghalliday: @afishbeck see comment above.

afishbeck: Yes, agreed.  I will approve.

afishbeck (approved):
@ghalliday looks good.

PR #15406, opened 2021-09-24
HPCC-26600 Use resourced memory in k8s in hthor

jakesmith:
1) Base default query memory on config resource specifications
(replacing unconfigurable 300MB limit).
2) Add a memory section (like Thor's, see HPCC-26443) to allow
specification of how much memory is user for the "query" vs "thirdParty"
3) Allow options to overriden at workunit level

Also fixes a bug introduced by HPCC-26443 which meant
that when globalMemorySize was specified in Thor (bare-metal),
only a percentage of that specific requirement was
allocated to the query/roxiemem.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/ecl/eclagent/eclagent.cpp
+++ b/ecl/eclagent/eclagent.cpp
@@ -53,6 +53,7 @@
 #include "roxiehelper.hpp"
 #include "jlzw.hpp"
 #include "anawu.hpp"
+#include "hpccconfig.hpp"
 
 using roxiemem::OwnedRoxieString;
 
@@ -2112,16 +2113,6 @@ void EclAgent::runProcess(IEclProcess *process)
     assertex(rowManager==NULL);
     allocatorMetaCache.setown(createRowAllocatorCache(this));
 
-    //Get memory limit. Workunit specified value takes precedence over config file
-    int memLimitMB = agentTopology->getPropInt("@defaultMemoryLimitMB", DEFAULT_MEM_LIMIT);
-    memLimitMB = queryWorkUnit()->getDebugValueInt("hthorMemoryLimit", memLimitMB);
-
-    bool allowHugePages = agentTopology->getPropBool("@heapUseHugePages", false);
-
-    bool allowTransparentHugePages = agentTopology->getPropBool("@heapUseTransparentHugePages", true);
-
-    bool retainMemory = agentTopology->getPropBool("@heapRetainMemory", false);
-
     if (agentTopology->hasProp("@httpGlobalIdHeader"))
         updateDummyContextLogger().setHttpIdHeaders(agentTopology->queryProp("@httpGlobalIdHeader"), agentTopology->queryProp("@httpCallerIdHeader"));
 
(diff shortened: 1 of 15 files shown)


ecl/eclagent/eclagent.cpp
```diff
@@ -2150,6 +2141,34 @@ void EclAgent::runProcess(IEclProcess *process)
...
+    if (isEmptyString(jobMemorySectionName))
+        jobMemorySectionName = "jobMemory";
+    //Get memory limit. Workunit specified value takes precedence over config
+    unsigned memLimitMB = queryWorkUnit()->getDebugValueInt("hthorMemoryLimit", 0);
+    if (0 == memLimitMB)
+    {
+        if (isContainerized())
+        {
+            // not being picked up in Thor..
+            // because it's under eclAgentResources ? (but is that generated in config?)
```
ghalliday, line 2156: Should answer this question

helm/hpcc/values.schema.json
```diff
@@ -487,9 +487,13 @@
           "type": "string",
           "description": "The amount of overall resourced memory to dedicate to the query"
         },
-        "thirdParties": {
+        "thirdParty": {
           "type": "string",
           "description": "The amount of overall resource memory to reserve for 3rd party use"
+        },
+        "maxMemPercentage": {
+          "type": "string",
```
ghalliday, line 495: number?

ghalliday (commented):
@jakesmith I think it looks ok, but I found it very hard to review!
I wasn't really sure why jobMemorySectionName was needed.  Is it because the same configuration file is used for eclagent/thormaster/thorworker?

This also needs to be clearly documented - should have a documentation jira with a description of how the memory values are processed.

ecl/eclagent/eclagent.cpp
jakesmith: should have been removed.
This was solved via template foo in this PR, thor.yaml ensures that "resources" is available as a copy of eclAgentResources.

helm/hpcc/values.schema.json
jakesmith: I wanted it as a real, but that is not a json schema supported type.

thorlcr/master/thmastermain.cpp
```diff
@@ -771,8 +773,20 @@ int main( int argc, const char *argv[]  )
...
 #endif
             }
+
+#ifndef _CONTAINERIZED
+            // @localThor mode - 25% is used for manager and 50% is used for workers
+            // overrides recommended max percentage preferences if present
+            if (globals->getPropBool("@localThor") && (0 == mmemSize))
+            {
+                managerMemory->setProp("@maxMemPercentage", "25.0");
+                workerMemory->setProp("@maxMemPercentage", VStringBuffer("%.2f", 50.0 / slavesPerNode));
```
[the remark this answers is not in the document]
jakesmith: agree, thought about introducing one, I may now

jakesmith:
> I wasn't really sure why jobMemorySectionName was needed. Is it because the same configuration file is used for eclagent/thormaster/thorworker?

hthor uses 'jobMemory', I'm not sure that is the best name for it, perhaps section name should just be "memory".
Because thor has the pre-existing eclagent/thormanager/thorworker *Resource sections, I wanted the *Memory section to follow suit, and they did already, except it was not possible to define specification for the Thor eclagent memory, the natural name for that is "eclAgentMemory" , so symetry with "eclAgentResources".

That meant the common code which picks up this section needed to know the name of the eclAgent Memory section, by default "jobMemory", but in Thor overriden to be "eclAgentMemory".

jakesmith:
> This also needs to be clearly documented - should have a documentation jira with a description of how the memory values are processed.

I'll open up a doc. JIRA describing the section(s) and what they're intended for / how they work.

jakesmith (commented):
@ghalliday - please review new commit, ereply comments and doc. JIRA : https://track.hpccsystems.com/browse/HPCC-26614

helm/hpcc/values.schema.json
ghalliday: I think "number" represents a floating point type, while "integer" does not allow a fractional part.

jakesmith: you're right, have changed to "number"

jakesmith:
> I think "number" represents a floating point type, while "integer" does not allow a fractional part.

@ghalliday - I've pushed a commit for tihs change. Please let me know when you're ready for me to squash.

system/jlib/jptree.cpp
```diff
@@ -1772,6 +1772,67 @@ void PTree::addPropInt64(const char *xpath, __int64 val)
     }
 }
 
+void PTree::setPropReal(const char * xpath, double val)
```
ghalliday, line 1775: looks ok, but I expected it to be implemented as something like
```
{
    std::string s = std::to_string(val);
    setProp(xpath, s.c_str());
}
```
would that be better?

system/jlib/jptree.cpp
```diff
@@ -1772,6 +1772,67 @@ void PTree::addPropInt64(const char *xpath, __int64 val)
...
+        {
+            IPropertyTree *propBranch = queryCreateBranch(branch, prop);
+            propBranch->setPropReal(NULL, val);
+        }
+    }
+}
+
+void PTree::addPropReal(const char *xpath, double val)
+{
+    if (!xpath || '\0' == *xpath)
```
ghalliday, line 1804: similar comment here about implementation

ghalliday (commented):
@jake 1 minor comment.  I think it can be squashed.

ghalliday:
@jakesmith that should have been

system/jlib/jptree.cpp
jakesmith: it would have same net effect, but calling setProp would involve a virtual and some unecessary checking and an extra strlen call, ultimately to get to same setLocal call being called directly here - so this is more efficient.

jakesmith:
@ghalliday - rebased/squashed

ghalliday (approved):

PR #15412, opened 2021-09-27
HPCC-26601 Changes for front-end currency formatting

shamser:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/helm/hpcc/values.schema.json
+++ b/helm/hpcc/values.schema.json
@@ -265,6 +265,11 @@
           "description": "resource cost",
           "type": "object",
           "properties": {
+            "currencyCode": {
+                "description": "currency code (ISO 4217)",
+                "type": "string",
+                "maxLength": 3
+            },
             "perCpu": {
               "description": "cost of a single cpu",
               "type": "number"
--- a/helm/hpcc/values.yaml
+++ b/helm/hpcc/values.yaml
@@ -47,7 +47,7 @@ global:
     # kubeApiPort: 7443
 
   cost:
-    moneyLocale: "en_US.UTF-8"
+    currencyCode: USD
     perCpu: 0.126
     storageAtRest: 0.0135
     storageReads: 0.0485
--- a/esp/platform/espcfg.cpp
+++ b/esp/platform/espcfg.cpp
@@ -373,7 +373,6 @@ CEspConfig::CEspConfig(IProperties* inputs, IPropertyTree* envpt, IPropertyTree*
             mergeConfiguration(*cost, *espConfigCost, nullptr, false);
         }
 #endif
-
         initializeStorageGroups(daliClientActive());
 
         const unsigned dafilesrvConnectTimeout = m_cfg->getPropInt("@dafilesrvConnectTimeout", 10)*1000;
--- a/esp/scm/ws_dfu.ecm
+++ b/esp/scm/ws_dfu.ecm
@@ -191,7 +191,7 @@ ESPStruct [nil_remove] DFUFileDetail
     [min_ver("1.39")] ESParray<ESPstruct DFUFileBloom> Blooms;
     [min_ver("1.40")] int ExpireDays;
     [min_ver("1.41")] string KeyType;
-    [min_ver("1.59")] string Cost;
+    [min_ver("1.60")] double Cost;
 };
 
 ESPStruct DFUSpaceItem
@@ -968,8 +968,8 @@ ESPresponse [exceptions_inline, nil_remove] DFUFilePublishResponse
 //  ===========================================================================
 ESPservice [
     auth_feature("DEFERRED"),
-    version("1.59"),
-    default_client_version("1.59"),
+    version("1.60"),
+    default_client_version("1.60"),
     noforms,
     exceptions_inline("./smc_xslt/exceptions.xslt")] WsDfu
 {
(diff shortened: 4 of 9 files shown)


initfiles/componentfiles/configxml/esp.xsl
```diff
@@ -68,6 +68,7 @@
                 <xsl:copy-of select="/Environment/Software/Directories"/>
             </Software>
             <xsl:copy-of select="/Environment/Hardware/cost"/>
+            <xsl:copy-of select="/Environment/Software/locale"/>
```
GordonSmith, line 71: Could this be worked out at runtime?

shamser: esp.xml configuration is generated from the environment.xml when the esp service is started with hpcc-init.  There is a mechanism in ESP that supports the reloading of an updated configuration (into a running esp) and, so the locale can be changed and reloaded at runtime.  Is that what you mean?

GordonSmith: I was thinking more of reading the locale from the machine / OS instance, so the human is not involved?  But given they have to enter the pricing anyway its fine as is.

shamser: I see what you mean.  I'd imagine the locale would need to be installed in the OS to be able to do that and that's a bit cumbersome in the containerized deployment.

helm/hpcc/values.schema.json
```diff
@@ -261,6 +261,24 @@
         "defaultMirrorPath": {
           "type": "string"
         },
+        "locale" : {
+          "decription": "the locale used for internationalization",
+          "type": "object",
+          "properties": {
+            "languageCode": {
```
ghalliday, line 268: what are these going to be used for?
The costCurrencyLocale makes sense, but I would expect everything else to be determined by the user.

helm/hpcc/values.yaml
```diff
@@ -29,7 +29,13 @@ global:
   # If not specified, the first esp component that exposes eclservices application is assumed.
   # Can also be overridden locally in individual components
   ## defaultEsp: eclservices
-  
+
+  # The locale used for internationalization
+  locale:
+    langugeCode: "en" # ISO 639-1 language code (https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes)
+    countryCode: "US" # ISO 3166-2 country code (https://en.wikipedia.org/wiki/ISO_3166-2)
+    currencyCode: "GBP" # ISO 4217 currency code (https://en.wikipedia.org/wiki/ISO_4217)
```
ghalliday, line 37: USD would be a more natural choice.

helm/hpcc/values.yaml
```diff
@@ -29,7 +29,13 @@ global:
   # If not specified, the first esp component that exposes eclservices application is assumed.
   # Can also be overridden locally in individual components
   ## defaultEsp: eclservices
-  
+
+  # The locale used for internationalization
+  locale:
```
ghalliday, line 34: What was the reason for moving to a new section? I can't see the other entries being used.

ghalliday (commented):
@shamser a few comments.  Most are minor.  The locale details might be worth discussing some more, since I'm not sure sure I understand how the other values will ever be used.

helm/hpcc/values.schema.json
shamser: The JavaScript functions that I looked at that handle currency formatting  requires these three components: country, language and currency (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat.  We could of course provide the required information as 2 fields (language locale+ currency).  However, by providing the 3 required items separately, multi-language locale support is possible. i.e. JavaScript needs languageCode+countryCode+currencyCode.  C++ standard library functions require languageCode+countryCode (currency is inferred).

I moved them to the global section because I was thinking as the fields are separate, the language and country attribute could be used for other purposes.  Would we want to have a costCurrencyLocale with these 3 items under the field and quite possibly have a requirement to have another locale attribute for other purposes?

helm/hpcc/values.yaml
shamser: Please see previous comment regarding making locale more generic.

shamser:
@ghalliday I'll fix the typos and decimal formatting.  Please can you see the other comments regarding the new locale structure.

helm/hpcc/values.schema.json
ghalliday: The locale formatting may require 3 options - but I think 2 of those options are actually supplied by the client.  For instance. 
```
console.log(new Intl.NumberFormat('de-DE', { style: 'currency', currency: 'EUR' }).format(123456.789));
> "123.456,79 €"
```
which is what I would expect if I was german.
but 
```
console.log(new Intl.NumberFormat('en-UK', { style: 'currency', currency: 'EUR' }).format(123456.789));
> "€123,456.79"
```
is what I would expect as an english person viewing the same information.

i.e. the currency needs to come from the platform, but the natural way for it to be displayed really comes from the locale being used by the user.  @GordonSmith what are your thoughts?

GordonSmith: I agree - I plan on formatting the number per the local locale and using the short code to resolve the currency symbol and failing that use the "currency" value supplied?

ghalliday (approved):
@shamser please squash

ghalliday (approved):

PR #15431, opened 2021-09-29
HPCC-26616 add default network policies for exposed esp services

xwang2713:
Also user can set additional ingress rules for roxie

Add following in values.schema.json:
1) exposedServices list
2) ingress network policy

Provide default ingress rules in hpcc/values.yaml for exposedServices.
Allow user add additional ingress rules for roxie. Current roxie ingress rule is kept as default.

Add some basic instruction in hpcc/values.yaml

Signed off by Xiaoming.Wang@LexisNexis.com

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

## Testing:

Tested with "helm lint" and "helm template".
I will tested more on Azure

### DIFF
--- a/helm/hpcc/templates/network.yaml
+++ b/helm/hpcc/templates/network.yaml
@@ -157,6 +157,25 @@ spec:
         matchLabels:
           app: eclservices
 ---
+{{ range .Values.esp -}}
+{{- if has .name $.Values.global.visibilities.exposedServices }}
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: {{ .name }}
+spec:
+  podSelector:
+    matchLabels:
+      app: {{ .name }}
+  ingress:
+{{- if .service.ingress }}
+{{ toYaml .service.ingress | indent 2}}
+{{- else }}
+{{ toYaml $.Values.global.visibilities.ingress | indent 2}}
+{{- end }}
+---
+{{- end -}}
+{{- end -}}
 ##
 ## Provides access to pods labeled prometheusMetricsReporter: "yes"
 ## onto port declared in global.metrics.sinks[type=prometheus].configuration.port
--- a/helm/hpcc/values.schema.json
+++ b/helm/hpcc/values.schema.json
@@ -316,6 +316,15 @@
         "visibilities": {
           "type": "object",
           "description" : "Global visibilities configuration",
+          "properties" : {
+            "exposedServices": {
+              "type": "array",
+              "items": { "type": "string" }
+	    },
+            "ingress": {
+              "$ref" : "#/definitions/ingress"
+	    }
+	  },
           "additionalProperties": {
             "type": "object",
             "required": [ "type" ],
@@ -752,6 +761,15 @@
         }
       }
     },
+    "ingress": {
+      "type": "array",
+      "items": {
+        "additionalProperties": {
+          "type": "array",
+          "items": { "type": "object" }
+        }
+      }
+    },
     "egress": {
       "type": "object",
       "properties": {
@@ -791,6 +809,9 @@
         "visibility": {
           "type": "string",
           "description": "Should this service be exposed outside the cluster, locally or to the internet"
+        },
+        "ingress": {
+          "$ref": "#/definitions/ingress"
         },
         "annotations": {
           "type": "object",
@@ -1092,7 +1113,10 @@
         "labels": {
           "type": "object",
           "additionalProperties": { "type": "string" }
-        }
+        },
+	"ingress": {
+          "$ref" : "#/definitions/ingress"
+	}
       }
     },
     "toposerver": {
--- a/helm/hpcc/values.yaml
+++ b/helm/hpcc/values.yaml
@@ -82,8 +82,12 @@ global:
       #labels:
       #  mylabel: "4"
       type: LoadBalancer
-      
-      
+    # Ingress Network Polices will be created for following esp pods. The Network Policies will only take effect
+    # when a Kubernetes supported network polocy provider is ininstalled.
+    exposedServices: ["eclwatch", "eclqueries", "esdl-sandbox", "sql2ecl"]
+    # The following settings can be overwritted by ingress rules defined esp services/ingress
+    ingress:
+    - {}
 
   # For pod placement instruction and examples please reference docs/placements.md
 
@@ -444,6 +448,13 @@ esp:
     ## You can also specify labels on a service
     #labels:
     #  mylabel: "3"
+
+    # Additional ingress rules can be defined here
+    # ingress:
+    # - from:
+    #   - ipBlock:
+    #      cidr: <internal-cidr/>
+
   #resources:
   #  cpu: "1"
   #  memory: "2G"
@@ -523,6 +534,10 @@ roxie:
   topoServer:
     replicas: 1
   #directAccessPlanes: []  #add direct access planes that roxie will read from without copying the data to its default data plane
+
+  # Additional roxie ingress rule can be defined here
+  # ingress:
+  # - {}
 
 thor:
 - name: thor


helm/hpcc/templates/network.yaml
```diff
@@ -157,6 +157,25 @@ spec:
         matchLabels:
           app: eclservices
 ---
+{{ range .Values.esp -}}
+{{- if has .name $.Values.global.visibilities.exposedServices }}
```
jakesmith, line 161: minor: This will cause an error if "visibilities" isn't defined.
Elsewhere we check for it's existence before accessing keys in it, e.g. in "hpcc.isVisibilityPublic".

helm/hpcc/templates/network.yaml
```diff
@@ -157,6 +157,25 @@ spec:
...
+spec:
+  podSelector:
+    matchLabels:
+      app: {{ .name }}
+  ingress:
+{{- if .service.ingress }}
+{{ toYaml .service.ingress | indent 2}}
+{{- else }}
+{{ toYaml $.Values.global.visibilities.ingress | indent 2}}
+{{- end }}
```
jakesmith, line 175: minor: it's worth trying to indent the template language structure for readability.
.e.g the above can become:
```
{{- if .service.ingress }}
 {{- toYaml .service.ingress | nindent 2 }}
{{- else }}
 {{- toYaml $.Values.global.visibilities.ingress | nindent 2 }}
{{- end }}
```

helm/hpcc/values.schema.json
```diff
@@ -316,6 +316,15 @@
         "visibilities": {
           "type": "object",
           "description" : "Global visibilities configuration",
+          "properties" : {
+            "exposedServices": {
+              "type": "array",
+              "items": { "type": "string" }
+	    },
```
jakesmith, line 323: minor: indentation issue, and lines below.

helm/hpcc/values.schema.json
```diff
@@ -1092,7 +1113,10 @@
         "labels": {
           "type": "object",
           "additionalProperties": { "type": "string" }
-        }
+        },
+	"ingress": {
```
jakesmith, line 1117: minor: indentation issue, tabs vs spaces?

jakesmith: Is this the right place for the roxie ingress section, or should it go under roxie.services under each "roxieservice" ?

helm/hpcc/values.yaml
```diff
@@ -82,8 +82,12 @@ global:
       #labels:
       #  mylabel: "4"
       type: LoadBalancer
-      
-      
+    # Ingress Network Polices will be created for following esp pods. The Network Policies will only take effect
```
jakesmith, line 85: >for following 

for the following

helm/hpcc/values.yaml
```diff
@@ -82,8 +82,12 @@ global:
...
+    # when a Kubernetes supported network polocy provider is ininstalled.
```
jakesmith, line 86: >polocy

policy

> ininstalled

installed

helm/hpcc/values.yaml
```diff
@@ -82,8 +82,12 @@ global:
...
+    exposedServices: ["eclwatch", "eclqueries", "esdl-sandbox", "sql2ecl"]
+    # The following settings can be overwritted by ingress rules defined esp services/ingress
```
jakesmith, line 88: >overwritted

overwritten

jakesmith: > defined esp services/ingress

defined by the ingress section under each esp service.

jakesmith: I could apply to other (non-esp) components too?
e.g. if someone wanted to expose the dafilesrv.service.
Perhaps drop the word 'esp' in comment.

helm/hpcc/templates/network.yaml
```diff
@@ -157,6 +157,25 @@ spec:
...
+kind: NetworkPolicy
+metadata:
+  name: {{ .name }}
+spec:
+  podSelector:
+    matchLabels:
+      app: {{ .name }}
+  ingress:
+{{- if .service.ingress }}
+{{ toYaml .service.ingress | indent 2}}
```
jakesmith, line 172: is it correct for there to be an ingress section if the component defines visibility: local ?

helm/hpcc/templates/network.yaml
```diff
@@ -157,6 +157,25 @@ spec:
         matchLabels:
           app: eclservices
 ---
+{{ range .Values.esp -}}
```
jakesmith, line 160: Should this walk all components that have a service and generate a NetworkPolicy if they have ingress?
So all services are covered in same way.

Apart from esp and roxie services, there's also potentially dafilesrv and sasha services we might want ingress on too.

jakesmith (commented):
@xwang2713 - please see comment/questions.

helm/hpcc/templates/network.yaml
richardkchapman: Seems wrong to be specifying exposed services via global.visibilities. I would have thought the settings in the relevant visibility section should be sufficient

helm/hpcc/templates/network.yaml
richardkchapman: Agree with Jake, I don't know why this is just for esp rather than for anything defining a service

helm/hpcc/values.yaml
```diff
@@ -82,8 +82,12 @@ global:
...
+    ingress:
```
richardkchapman, line 89: ingress section should be within each of the user-defined sections above. Any "visibility" that has type "loadbalancer" probably wants to also define its own ingress settings, which are specific to that visibility rule.

helm/hpcc/values.yaml
```diff
@@ -444,6 +448,13 @@ esp:
     ## You can also specify labels on a service
     #labels:
     #  mylabel: "3"
+
```
richardkchapman, line 451: We don't need additional ingress rules at the individual service level. People can define a new visibility if they want different ingress rules for some services than others.

richardkchapman (changes requested):

helm/hpcc/values.schema.json
xwang2713: Current roxie networkpolicy rules are under roxie instead of roixe.services

helm/hpcc/templates/network.yaml
xwang2713: I think it doesn't matter the visibility local or global. The networkpolicies for ingress are needed for exposed services.

helm/hpcc/templates/network.yaml
xwang2713: Each component seems define service differently. Could we defer this for later improvement? I currently only implement the services through "External IP"

helm/hpcc/values.yaml
xwang2713: I thought we wanted a global default ingress rules for the services which has type "loadbalancer"

helm/hpcc/templates/network.yaml
xwang2713: So each service define its own ingress as:

> esp:
> \- name: eclwatch
>     service:
>         ingress:
>         \- {}

helm/hpcc/templates/network.yaml
xwang2713: fixed

helm/hpcc/values.schema.json
xwang2713: fixed

helm/hpcc/values.schema.json
xwang2713: fixed the indentation

helm/hpcc/values.yaml
xwang2713: fixed

helm/hpcc/values.yaml
xwang2713: fixed

helm/hpcc/values.yaml
xwang2713: I only implemented the services which has "loadbalancer" now. Can we defer the other services later since each of them probably need treat differently.

I fixed the phrase but probably will remove "exposedServices" and move "ingress" to each esp service section.

helm/hpcc/values.yaml
xwang2713: little confused here. From previous comment I thought you suggest not define ingress rule in global.visibilities. So I plan to move the ingress definition here.

PR #15434, opened 2021-09-30
HPCC-26616 Add default network policies for exposed services

richardkchapman:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/helm/hpcc/values.yaml
+++ b/helm/hpcc/values.yaml
@@ -78,10 +78,17 @@ global:
         #service.beta.kubernetes.io/aws-load-balancer-internal: "true"
         service.beta.kubernetes.io/azure-load-balancer-internal: "true"
       type: LoadBalancer
+      # If ingress is specified, an ingress Network Policy will be created for any pod implementing a service with this visibility
+      # Default allows ingress from anywhere, but more restrictive rules can be used if preferred.
+      # Ingress rules can also be overriden by individual services
+      ingress:
+      - {}
     global:
       #labels:
       #  mylabel: "4"
       type: LoadBalancer
+      ingress:
+      - {}
       
       
 # For pod placement instruction and examples please reference docs/placements.md
@@ -318,6 +325,7 @@ dafilesrv:
   service:
     servicePort: 7600
     visibility: local
+    
 
 dali:
 - name: mydali
@@ -509,6 +517,8 @@ roxie:
     listenQueue: 200
     numThreads: 30
     visibility: local
+    # Can override ingress rules for each service if desired - for example to add no additional ingress permissions you can use
+    # ingress: []
   ## replicas indicates the number of replicas per channel
   replicas: 2  
   numChannels: 2
--- a/helm/hpcc/templates/_helpers.tpl
+++ b/helm/hpcc/templates/_helpers.tpl
@@ -918,8 +918,9 @@ A template to generate a service
 Pass in dict with .root, .name, .service, .defaultPort, .selector defined
 */}}
 {{- define "hpcc.addService" }}
-{{- $lvars := dict "type" "ClusterIP" "labels" dict "annotations" dict -}}
+{{- $lvars := dict "type" "ClusterIP" "labels" dict "annotations" dict "ingress" list "serviceName" .name -}}
 {{- if hasKey . "service" -}}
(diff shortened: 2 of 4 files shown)


helm/hpcc/values.yaml
```diff
@@ -318,6 +320,8 @@ dafilesrv:
   service:
     servicePort: 7600
     visibility: local
+    ingress: []
```
richardkchapman, line 323: This was for me testing, and should be removed

helm/hpcc/values.yaml
```diff
@@ -509,6 +513,7 @@ roxie:
     listenQueue: 200
     numThreads: 30
     visibility: local
+    ingress: []
```
richardkchapman, line 516: This was for me testing, and should be removed and/or commented out

helm/hpcc/values.yaml
```diff
@@ -78,6 +78,8 @@ global:
         #service.beta.kubernetes.io/aws-load-balancer-internal: "true"
         service.beta.kubernetes.io/azure-load-balancer-internal: "true"
       type: LoadBalancer
+      ingress:
```
richardkchapman, line 81: Needs a comment.

helm/hpcc/values.yaml
```diff
@@ -78,6 +78,8 @@ global:
...
+      - {}
     global:
       #labels:
       #  mylabel: "4"
```
richardkchapman, line 85: Need to put a default ingress rule on global too...

helm/hpcc/values.yaml
```diff
@@ -78,10 +78,17 @@ global:
         #service.beta.kubernetes.io/aws-load-balancer-internal: "true"
         service.beta.kubernetes.io/azure-load-balancer-internal: "true"
       type: LoadBalancer
+      # If ingress is specified, an ingress Network Policy will be created any pod implementing a service with this visibility
```
jakesmith, line 81: >will be created any pod

will be created for any pod

helm/hpcc/values.yaml
```diff
@@ -78,10 +78,17 @@ global:
...
+      # Default allows ingress from anywhere, but more restrictive rules can be used if preferred.
+      # Ingress rules can also be overriden by individual services
+      ingress:
+      - {}
```
jakesmith, line 85: I see this is the correct way from the docs, but wasn't obvious to me - this was ingress rule that meant allow all.
Perhaps add a comment next to it?

helm/hpcc/values.yaml
```diff
@@ -509,6 +517,8 @@ roxie:
     listenQueue: 200
     numThreads: 30
     visibility: local
+    # Can override ingress rules for each service if desired - for example to add no additional ingress permissions you can use
+    # ingress: []
```
jakesmith, line 521: I wonder if there should be an option to merge with the global ingress rules.

jakesmith (commented):
@richardkchapman looks good, v. minor comments only.

jakesmith:
@richardkchapman - didn't past helm checks,  but not sure why, is it kube score and this ?
>[CRITICAL] roxie networking.k8s.io/v1/NetworkPolicy: The NetworkPolicys selector doesn't match any pods

helm/hpcc/values.yaml
richardkchapman: There's a comment on line 82...

helm/hpcc/values.yaml
richardkchapman: Because they are lists rather than dictionaries, it's not clear what merging would mean. In particular I suspect it would mean "allow whatever either location says is allowed" which means you would not be able to override with something MORE restrictive.

richardkchapman:
I will look at the helm test issue

helm/hpcc/values.yaml
jakesmith: it was the syntax itself meaning what line 82 says, that confused me.

jakesmith (approved):
@richardkchapman - looks good to go

xwang2713 (commented):
Just tested it. Looks good

ghalliday (approved):
Looks sensible.

PR #15440, opened 2021-10-01
HPCC-26590 Variable/delimitted spray fails if the first record/row is longer than the default block size (256k).

AttilaVamos:
Fix first record size issue in CSV quick partitioner.

Originally the parameter of the ensureBuffered() function was blockSize  and it assumed the first record is not too big and ignored the record size parameter. It was wrong.
The new parameter is the bufferSize what is calculated from the blockSize and expectedRecordSize

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

## Testing:

Tested manually with the user provided file

### DIFF
--- a/dali/ft/daftformat.cpp
+++ b/dali/ft/daftformat.cpp
@@ -143,8 +143,11 @@ void CPartitioner::commonCalcPartitions()
         //Don't add an empty block on the start of the this chunk to transfer.
         if ((split != firstSplit) || (inputOffset != startInputOffset))
         {
-            results.append(*new PartitionPoint(whichInput, split-1, startInputOffset-thisOffset+thisHeaderSize, inputOffset - startInputOffset - cursor.trimLength, cursor.outputOffset-startOutputOffset));
-            JSON_DBGLOG("commonCalcPartitions: startInputOffset:%lld, inputOffset: %lld, startOutputOffset: %lld, cursor.outputOffset: %lld",startInputOffset ,inputOffset, startOutputOffset, cursor.outputOffset);
+            // To avoid the first partition's whichOutput value becomes -1 (which means invalid)
+            unsigned whichOutput = (split > 0 ? split-1 : split);
+            results.append(*new PartitionPoint(whichInput, whichOutput, startInputOffset-thisOffset+thisHeaderSize, inputOffset - startInputOffset - cursor.trimLength, cursor.outputOffset-startOutputOffset));
+            JSON_DBGLOG("commonCalcPartitions: startInputOffset:%lld, inputOffset: %lld, startOutputOffset: %lld, cursor.outputOffset: %lld, whichInput: %u, whichOutput: %u",
+                                               startInputOffset ,inputOffset, startOutputOffset, cursor.outputOffset, whichInput, whichOutput);
             startInputOffset = inputOffset;
             startOutputOffset = cursor.outputOffset;
         }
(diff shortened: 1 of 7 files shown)


dali/ft/daftformat.cpp
```diff
@@ -1065,11 +1065,10 @@ void CCsvQuickPartitioner::findSplitPoint(offset_t splitOffset, PartitionCursor
     else
     {
         // We are in the first part of the file
-        bool eof;
-        if (format.maxRecordSize + maxElementLength > blockSize)
-            eof = !ensureBuffered(blockSize);
-        else
-            eof = !ensureBuffered(format.maxRecordSize + maxElementLength);
```
jakesmith, line 1072: The above condition looks backward, but correct if flipped ?
i.e. if > blockSize should be using format.maxRecordSize + maxElementLength

jakesmith: also, the same code is present in a few other places, if this is wrong, so are other places I think.

dali/ft/daftformat.cpp
```diff
@@ -1065,11 +1065,10 @@ void CCsvQuickPartitioner::findSplitPoint(offset_t splitOffset, PartitionCursor
...
+        // Originally the parameter of the ensureBuffered() function was blockSize
+        // and it assumed the first record is not too big and ignored the record size parameter. It was wrong.
+        // The new parameter is the bufferSize what is calculated from the blockSize and expectedRecordSize
```
jakesmith, line 1070: Shouldn't really add comments to code referrring to code that no longer exists,
but should be in PR/JIRA description/comments.

jakesmith (commented):
@AttilaVamos - please see comments.

Also there's a typo in commit/PR message : falls->fails, and PR check boxes not checked.

dali/ft/daftformat.cpp
AttilaVamos: Removed and added to PR description

dali/ft/daftformat.cpp
```diff
@@ -143,8 +143,11 @@ void CPartitioner::commonCalcPartitions()
         //Don't add an empty block on the start of the this chunk to transfer.
         if ((split != firstSplit) || (inputOffset != startInputOffset))
         {
-            results.append(*new PartitionPoint(whichInput, split-1, startInputOffset-thisOffset+thisHeaderSize, inputOffset - startInputOffset - cursor.trimLength, cursor.outputOffset-startOutputOffset));
-            JSON_DBGLOG("commonCalcPartitions: startInputOffset:%lld, inputOffset: %lld, startOutputOffset: %lld, cursor.outputOffset: %lld",startInputOffset ,inputOffset, startOutputOffset, cursor.outputOffset);
+            // To avoid the first partition's whichOutput value becomes -1 (which means invalid)
+            unsigned whichOutput = (split > 0 ? split-1 : split);
```
jakesmith, line 147: when does it enter this block, such that split is 0?
I thought the if (split != firstSplit) covered that case?
Perhaps comment need extending to explain.

dali/ft/daftformat.cpp
```diff
@@ -1002,79 +1005,64 @@ void CCsvQuickPartitioner::findSplitPoint(offset_t splitOffset, PartitionCursor
 {
     const byte *buffer = bufferBase();
     numInBuffer = bufferOffset = 0;
-    if (splitOffset != 0)
```
jakesmith, line 1005: git not making it easy to see the change, but I see it's just removal of this if and the else block.
tbh, I'm not familiar with this code, and am not sure what it was doing (or what it thought it was doing) in the case of splitOffset == 0 vs not.
@ghalliday - could you also review this please?

dali/ft/daftformat.cpp
```diff
@@ -1339,81 +1327,67 @@ void CUtfQuickPartitioner::findSplitPoint(offset_t splitOffset, PartitionCursor
...
```
jakesmith, line 1342: git not making it easy to see the change, but I see it's just removal of this if and the else block.
@ghalliday - would you look at this change too please?

dali/ft/filecopy.cpp
```diff
@@ -830,8 +830,12 @@ void FileSprayer::assignPartitionFilenames()
             cur.inputName.set(sources.item(cur.whichInput).filename);
             setCanAccessDirectly(cur.inputName);
         }
-        cur.outputName.set(targets.item(cur.whichOutput).filename);
-        setCanAccessDirectly(cur.outputName);
+        // Prevent to use invalid (-1) whichOutput index.
+        if (cur.whichOutput != (unsigned)-1)
```
jakesmith, line 834: I am not entirely sure at the moment, how a PartitionPoint with a whichOutput of -1 is getting added at the moment,
hence I'm not sure if this change should be necessary. See my other comment in CPartitioner::commonCalcPartitions() in daftformt.cpp

jakesmith (commented):
@AttilaVamos - a couple of questions.
But I'd also like @ghalliday to take a look.

ghalliday (commented):
@AttilaVamos I don't think the fix is right for a few reasons.

Previously are headers included in the sprayed file?  I think this change would also strip them.

AttilaVamos:
@jake @ghalliday 
It is hard to explain how findSplitPoint() function works and why the changes may be good based on  some grabbed out lines of code.
First of all the “if (splitOffset != 0)” condition and the related else block introduced when the initial version of CSV record structure discovery (create a naïve/basic ECL record structure with matching number of String/UTF8 fields)  or load the record structure defined in the first record of CSV file (if recordStructurePresent flag is true) implemented and the aim of that condition and else block was to process first record individually.
Later, when the processFullBuffer introduced to improve performance, to handling the record structure discovery/load is changed and it wasn’t necessary to handle the first record separately anymore. It is shame that “if – else” branch didn’t removed at that time. 
However this separated first record handling didn’t caused any problem until it faced a file with large (>256kB) first record. The problem was it reads the first 256kB (buffer size) of the file and it ignores the hint comes from the recordSize parameter.  So, the large first record caused exception and even though the user specified a larger record size value, it was not used.
My first fix was to modify that else branch to use recordSize parameter to read the first record, but Jake’s first review comments pointed out it wasn’t clear. Then I recheck the whole findSplitPoint() function and find out couple of things:
1. My first changes can cause large buffer allocation for the first record – to avoid it I need implement some loop to process the first record in chunks. But it is already implemented in the true branch.
2. Record structure handling doesn’t require separate first record handle. 
So I removed the “if(splitOffset != 0)” and the related else branch but this  change caused spray error related the output file. After some tracing and debugging I find out when the splitOffset == 0 the partition out file index becomes “-1” what means invalid/not defined file.  For usual files (with multiple records what can distribute among the target parts) the split offset moving from first split to the last one during the partition creation, but in a special occasion e.g. when there is only one large record for a multiple target parts environment, the scan can reach the end of the record (and the input file) and it is still in the splitOffset==0 state. 
A spray_test_csv.ecl regression test created to test several special occasions, like only one small, medium, large record in the input file, one small and one large/huge record in input with alternated sequence and so on.  To keep it quick (to enamel to run in Smoketest) it has been split into two test cases: spray_test_csv.ecl and spray_test_csv_stress.csv (marked with ‘//class=stress’).  The first executed in every environment (Smoketest, OBT), but the latter is only in OBT (based on the longer execution time)

ghalliday (commented):
@AttilaVamos A couple of comments, and a few comments from before that you haven't responded to.
I am a bit concerned that the way the record header code is implemented will lead to additional reads from files - with a significant cost overhead reading on azure.  I have opened a separate jira HPCC-26844 for that.

dali/ft/daftformat.cpp
```diff
@@ -1002,79 +1005,64 @@ void CCsvQuickPartitioner::findSplitPoint(offset_t splitOffset, PartitionCursor
...
                     }
-                    LOG(MCdebugProgress, unknownJob, "Found split after reading %d", ensureSize);
                 }
+                LOG(MCdebugProgress, unknownJob, "Found split after reading %d", ensureSize);
             }
         }
-        else if (splitOffset - thisOffset < thisSize)
-            throwError2(DFTERR_UnexpectedReadFailure, fullPath.get(), splitOffset-thisOffset+thisHeaderSize);
-    }
-    else
```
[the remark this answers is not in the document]
ghalliday: I think it is probably more subtle than that - you may lose the first characters of the header if multi character line terminators are used.  The fix suggest above should avoid the problem.

jakesmith (commented):
@AttilaVamos - I have no additional comments over Gavin and my previous comments, please address/respond to those.

dali/ft/daftformat.cpp
AttilaVamos: After some tracing and debugging I find out when the split == 0 the partition out file index becomes “-1” what means invalid/not defined file. For usual files (with multiple records what can distribute among the target parts) the split moving from first split to the last one during the partition creation, but in a special occasion e.g. when there is only one record for a multiple target parts environment, the scan can reach the end of the record (and the input file) and it is still in the split ==0 state.

dali/ft/filecopy.cpp
AttilaVamos: Changed to throw and error.

dali/ft/daftformat.cpp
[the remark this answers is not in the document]
AttilaVamos: Hmm... I am not 100% sure about what you mean. I need to recheck this.

dali/ft/daftformat.cpp
AttilaVamos: I gave my explanation below why I removeed this condition and the related else branch.

dali/ft/daftformat.cpp
```diff
@@ -1002,79 +1006,64 @@ void CCsvQuickPartitioner::findSplitPoint(offset_t splitOffset, PartitionCursor
...
+        eof = !ensureBuffered(blockSize);
+    else
+        eof = !ensureBuffered(format.maxRecordSize + maxElementLength);
+    bool fullBuffer = false;
+    //Could be end of file - if no elements read.
+    if (numInBuffer != bufferOffset)
     {
-        seekInput(splitOffset-thisOffset+thisHeaderSize);
+        //Throw away the first match, incase we hit the \n of a \r\n or something similar.
+        if (maxElementLength > 1)
```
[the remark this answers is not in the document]
AttilaVamos: At the moment I can't see why 'if splitOffset != 0', but I will trace this part again to check.

ghalliday:
Timing this PR out.  Please reopen if comments are addressed

PR #15472, opened 2021-10-12
HPCC-26679 Throw exception when destination storage plane not specified

wangkx:
When despraying a file to a dropzone in cloud environment, the destination
storage plane has to be specified. The code is added to throw an exception
if the destination storage plane is not specified.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/services/ws_fs/ws_fsService.cpp
+++ b/esp/services/ws_fs/ws_fsService.cpp
@@ -2502,10 +2502,6 @@ bool CFileSprayEx::onDespray(IEspContext &context, IEspDespray &req, IEspDespray
         double version = context.getClientVersion();
         StringBuffer destip(req.getDestIP());
         const char* destPlane = req.getDestPlane();
-#ifdef _CONTAINERIZED
-        if (isEmptyString(destPlane))
-            destPlane = req.getDestGroup();  // allow eclwatch to continue providing storage plane as 'destgroup' field
-#endif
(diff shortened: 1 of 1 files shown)


esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -2514,8 +2510,15 @@ bool CFileSprayEx::onDespray(IEspContext &context, IEspDespray &req, IEspDespray
         MemoryBuffer& dstxml = (MemoryBuffer&)req.getDstxml();
         if(dstxml.length() == 0)
         {
-            if(isEmptyString(destPlane) && destip.isEmpty())
+#ifdef _CONTAINERIZED
+            if (isEmptyString(destPlane))
```
ghalliday, line 2514: I'm not sure that this should be conditional on containerized.  A different issue.

ghalliday (approved):
I debated for a while whether the destPlane should be optional in containerized - since it could be a unique external ip.  On balance though I think we should require it, and prefer it is supplied for non-containerized as well.

wangkx:
@richardkchapman please merge this PR.

PR #15491, opened 2021-10-18
HPCC-26650 Plugin info is failing to load for libnlp

dehilsterlexis:
Signed-off-by: David de Hilster <david.dehilster@lexisnexisrisk.com>

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/plugins/nlp/nlp.hpp
+++ b/plugins/nlp/nlp.hpp
@@ -35,4 +35,10 @@
 #include "nlp_eng.hpp"
 #include <sstream>
 
+extern "C"
+{
+    ECL_NLP_API bool getECLPluginDefinition(ECLPluginDefinitionBlock *pb);
+    ECL_NLP_API void setPluginContext(IPluginContext * _ctx);
+}
+
 #endif
--- a/plugins/nlp/nlp.cpp
+++ b/plugins/nlp/nlp.cpp
@@ -49,8 +49,6 @@ namespace nlp {
     static CriticalSection cs; 
     static NLPEng *nlpEng = NULL;
 
-    ECL_NLP_API void setPluginContext(IPluginContext * _ctx) { parentCtx = _ctx; }
-
     //--------------------------------------------------------------------------------
     //                           ECL SERVICE ENTRYPOINTS
     //--------------------------------------------------------------------------------
@@ -71,3 +69,7 @@ namespace nlp {
         memcpy_iflen(tgt, sso.str().c_str(), tgtLen);
     }
 } // namespace nlp
+
+using namespace nlp;
+
+ECL_NLP_API void setPluginContext(IPluginContext * _ctx) { parentCtx = _ctx; }


GordonSmith:
@dehilsterlexis is this PR still valid, if so can you fill in the checkboxes?

plugins/nlp/nlp.hpp
```diff
@@ -30,9 +35,19 @@
...
+extern "C" {
+
+#ifdef ECL_NLP_EXPORTS
+ECL_NLP_API bool getECLPluginDefinition(ECLPluginDefinitionBlock *pb);
+ECL_NLP_API void setPluginContext(IPluginContext * _ctx);
+#endif
+
+}
+
+extern "C++" {
```
richardkchapman, line 47: extern "c++" is redundant nd confusing.

richardkchapman: Probably should change this line to read 

    namespace "nlp" {

richardkchapman (changes requested):

plugins/nlp/nlp.hpp
dehilsterlexis: Agreed. I'm working on this PR right now to verify it fixes the current crash problem.

richardkchapman: I think the changes in this PR are unrelated to the crash, but will address the unwanted tracing regarding failing to load. The crash is most likely because you are assuming that strings from ECL are null terminated, when they are not.

dehilsterlexis: It was a design decision to make it so the string would not have to have a length given that texts sent to NLP analyzers are variable in length. I don't think that is the problem with the current crash. Supposedly, this fixed the crash back a couple of months. I will talk with Gordon on all this to work out the deatils.

richardkchapman: Your code in nlp.cpp does not pass the length it gets from ECL to the nlp-analyser. Meaning it has no way to know how long it is.

dehilsterlexis: You are right. It assumes a null terminator from that point and thus needs to be addressed.

richardkchapman:
@dehilsterlexis This PR seems to be stalled. As far as I can tell the issue raised is still valid, and so are the comments on this PR. The comments relating to other fixes (null termination) can be ignored as they have been addressed elsewhere

dehilsterlexis:
@richardkchapman you are right. I'm testing right now and will complete the checkboxes.

dehilsterlexis:
@richardkchapman I believe I have checked all the appropriate boxes, although I am not fully familiar with submitting PRs to HPCC-Systems yet.

dehilsterlexis:
@GordonSmith for review

plugins/nlp/nlp.hpp
```diff
@@ -35,4 +35,15 @@
 #include "nlp_eng.hpp"
 #include <sstream>
 
+extern "C"
+{
+
+#ifdef ECL_NLP_EXPORTS
```
richardkchapman, line 41: Not sure why this is conditional...

plugins/nlp/nlp.hpp
```diff
@@ -35,4 +35,15 @@
...
+    ECL_NLP_API bool getECLPluginDefinition(ECLPluginDefinitionBlock *pb);
+    ECL_NLP_API void setPluginContext(IPluginContext * _ctx);
 #endif
+
+ECL_NLP_API void ECL_NLP_CALL AnalyzeText(size32_t & tgtLen, char * & tgt, size32_t anaLen, const char * ana, size32_t txtLen, const char * txt);
```
richardkchapman, line 46: I am not sure that this should be inside the extern "C" given the definition in the ecllib file

richardkchapman (approved):
Looks plausible

plugins/nlp/nlp.hpp
```diff
@@ -35,4 +35,9 @@
 #include "nlp_eng.hpp"
 #include <sstream>
 
+extern "C"
+{
+	ECL_NLP_API bool getECLPluginDefinition(ECLPluginDefinitionBlock *pb);
+	ECL_NLP_API void setPluginContext(IPluginContext * _ctx);
```
dcamper, line 41: setPluginContext in the .cpp file seems to be defined as a C++ function in the nlp namespace.  Should that function be moved outside the namespace?

dcamper (changes requested):

GordonSmith (changes requested):
@dehilsterlexis - relocate the setPluginContext in the cpp file as per Dans comment.

GordonSmith (approved):

dehilsterlexis:
@dcamper for review

dcamper (approved):

PR #15503, opened 2021-10-20
HPCC-26717 hThor file costs (all except index reads)

shamser:
## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/ecl/hthor/hthor.cpp
+++ b/ecl/hthor/hthor.cpp
@@ -446,6 +446,7 @@ void CHThorDiskWriteActivity::stop()
     close();
     if((helper.getFlags() & (TDXtemporary | TDXjobtemp) ) == 0 && !agent.queryResolveFilesLocally())
         publish();
+    io.clear();
     incomplete = false;
     if(clusterHandler)
         clusterHandler->finish(file);
@@ -568,7 +569,6 @@ void CHThorDiskWriteActivity::open()
         encrypted = true;
         blockcompressed = true;
     }
-    Owned<IFileIO> io;
     if(blockcompressed)
         io.setown(createCompressedFileWriter(file, groupedMeta->getFixedSize(), extend, true, ecomp, COMPRESS_METHOD_LZW));
     else
@@ -576,7 +576,7 @@ void CHThorDiskWriteActivity::open()
     if(!io)
         throw MakeStringException(errno, "Failed to create%s file %s for writing", (encrypted ? " encrypted" : (blockcompressed ? " compressed" : "")), filename.get());
     incomplete = true;
-    
+
     diskout.setown(createBufferedIOStream(io));
     if(extend)
         diskout->seek(0, IFSend);
@@ -659,9 +659,9 @@ void CHThorDiskWriteActivity::publish()
     desc->setDefaultDir(dir.str());
 
     Owned<IPropertyTree> attrs;
-    if(clusterHandler) 
+    if(clusterHandler)
         attrs.setown(createPTree("Part")); // clusterHandler is going to set attributes
-    else 
+    else
     {
         // add cluster
         StringBuffer mygroupname;
@@ -740,6 +740,11 @@ void CHThorDiskWriteActivity::publish()
     if (helper.getFlags() & TDWrestricted)
         properties.setPropBool("restricted", true);
 
+    if (io)
+    {
+        numDiskWrites = io->getStatistic(StNumDiskWrites);
+        properties.setPropInt64("@numDiskWrites", numDiskWrites);
+    }
     StringBuffer lfn;
     expandLogicalFilename(lfn, mangledHelperFileName.str(), agent.queryWorkUnit(), agent.queryResolveFilesLocally(), false);
     CDfsLogicalFileName logicalName;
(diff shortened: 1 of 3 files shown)


ecl/hthor/hthor.cpp
```diff
@@ -752,11 +757,24 @@ void CHThorDiskWriteActivity::publish()
         Owned<IDistributedFile> file = queryDistributedFileDirectory().createNew(desc);
         if(file->getModificationTime(modifiedTime))
             file->setAccessedTime(modifiedTime);
+        StringBuffer clusterName;
```
ghalliday, line 760: minor: move clusterName inside the if()

ecl/hthor/hthor.cpp
```diff
@@ -1211,6 +1229,7 @@ void CHThorIndexWriteActivity::execute()
         builder->finish(metadata, &fileCrc);
         duplicateKeyCount = builder->getDuplicateCount();
         cummulativeDuplicateKeyCount += duplicateKeyCount;
+        numDiskWrites = io->getStatistic(StNumDiskWrites);
```
ghalliday, line 1232: Slight question about if this was executed in a child query, but that is so unusual that it doesn't matter...

ecl/hthor/hthor.cpp
```diff
@@ -8414,6 +8437,32 @@ unsigned __int64 CHThorDiskReadBaseActivity::getLocalFilePosition(const void * r
...
+                if (superfile)
+                {
+                    unsigned subfile, lnum;
+                    if (superfile->mapSubPart(previousPartNum, subfile, lnum))
+                    {
+                        IDistributedSuperFile *super = dFile->querySuperFile();
+                        dFile = &(super->querySubFile(subfile, true));
+                    }
+                }
+                dFile->addAttrValue("@numDiskReads", curDiskReads);
```
ghalliday, line 8458: @jakesmith is there any issue with updating these properties?  Is the dali entry locked, and does it happen straight away.  What implications does this have once esp provides the meta information!

ecl/hthor/hthor.cpp
```diff
@@ -8446,8 +8496,7 @@ bool CHThorDiskReadBaseActivity::openNext()
             unsigned numCopies = curPart?curPart->numCopies():ldFile->numPartCopies(partNum);
             //MORE: Order of copies should be optimized at this point....
             StringBuffer file, filelist;
-            closepart();
-            if (dfsParts && superfile && curPart)
+            if (dfsParts && superfile && curPart && !subfileLogicalFilenames.empty())
```
ghalliday, line 8499: Has the closepart() moved to the correct place?  Would it also work to call closepart() after the while loop?  Asking because there is another call to closepart() within the loop code, and not sure exactly why the change was needed (to make sure stats were updated?)

ghalliday (commented):
@shamser a few, fairly minor comments.

common/workunit/workunit.cpp
```diff
@@ -2620,6 +2620,40 @@ cost_type aggregateCost(const IConstWorkUnit * wu, const char *scope, bool exclu
...
+        case SSTfunction:
+        case SSTchannel:
+        case SSTglobal:
+            continue;
+        default:
+            throwUnexpected();
+        }
+    }
+    return totalCost;
+}
```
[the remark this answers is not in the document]
shamser: I'll update to use HPCC-26788

common/workunit/workunit.cpp
```diff
@@ -2620,6 +2620,40 @@ cost_type aggregateCost(const IConstWorkUnit * wu, const char *scope, bool exclu
...
+
```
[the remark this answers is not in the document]
shamser: All the places that call aggregateDiskAccessCost(const IConstWorkUnit * wu, const char *scope) doesn't have access to IStatisticCollection.  e.g. EclAgent::doProcess(), EclAgentWorkflowMachine::noteTiming() and EclGraph::execute().  I'm not sure if I can get the required IStatisticCollection, from a call to a member function of IConstWorkUnit

common/workunit/workunit.hpp
```diff
@@ -1708,6 +1708,7 @@ extern WORKUNIT_API void updateWorkunitTimings(IWorkUnit * wu, ITimeReporter *ti
 extern WORKUNIT_API void updateWorkunitTimings(IWorkUnit * wu, StatisticScopeType scopeType, StatisticKind kind, ITimeReporter *timer);
 extern WORKUNIT_API void aggregateStatistic(StatsAggregation & result, IConstWorkUnit * wu, const WuScopeFilter & filter, StatisticKind search);
 extern WORKUNIT_API cost_type aggregateCost(const IConstWorkUnit * wu, const char *scope=nullptr, bool excludehThor=false);
+extern WORKUNIT_API cost_type aggregateDiskAccessCost(IStatisticCollection * statsCollection);
```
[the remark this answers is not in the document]
shamser: Removed.  No longer needed. Using AggregateStatistic() instead.

common/workunit/workunit.cpp
```diff
@@ -2620,6 +2620,40 @@ cost_type aggregateCost(const IConstWorkUnit * wu, const char *scope, bool exclu
...
+    cost_type totalCost = 0;
+    Owned<IStatisticCollectionIterator> iter = &(statsCollection->getScopes(nullptr, false));
+    ForEach(*iter)
+    {
+        IStatisticCollection & cur = iter->query();
+        switch (cur.queryScopeType())
+        {
+        case SSTactivity:
+            totalCost += cur.queryStatistic(StCostFileAccess);
+            break;
```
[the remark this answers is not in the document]
shamser: The new aggregateStatistic() is including the child activities costs.

ecl/hthor/hthor.cpp
shamser: I realize that it's not actually reading the file within that loop.  I've moved closepart back to the original location.

ghalliday (approved):
I can't see any problems.  All looks good afaics.

ecl/hthor/hthor.cpp
jakesmith: addAttrValue automatically locks the file attr properties, for an atomic update, and the update is pushed immediately.
There will need to be an implemtation of this that does the same atomic update when it's via a esp service.

ghalliday: Yes, I added a subtask/comment to the jira to make sure it was covered.

PR #15530, opened 2021-11-01
HPCC-26750 Fix issue with instanceName not being consumed in daserver startup

Michael-Gardner:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jutil.cpp
+++ b/system/jlib/jutil.cpp
@@ -3282,7 +3282,8 @@ bool checkCreateDaemon(unsigned argc, const char * * argv)
 #ifndef _CONTAINERIZED
     for (unsigned i=0;i<(unsigned)argc;i++) {
         if (streq(argv[i],"--daemon") || streq(argv[i],"-d")) {
-            if (daemon(1,0) || write_pidfile(argv[++i])) {
+            const char * instanceName = argv[i + 1];
+            if (!strncmp(instanceName,"-",1) || daemon(1,0) || write_pidfile(instanceName)) {
                 perror("Failed to daemonize");
                 return false;
             }
--- a/dali/server/daserver.cpp
+++ b/dali/server/daserver.cpp
@@ -427,6 +427,7 @@ int main(int argc, const char* argv[])
 
         for (unsigned i=1;i<(unsigned)argc;i++) {
             if (streq(argv[i],"--daemon") || streq(argv[i],"-d")) {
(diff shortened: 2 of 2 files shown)


Michael-Gardner:
@ghalliday When we consolidated the daemon code back in 8.2 this error snuck in.  Originally (in daserver.cpp) we used ++i to bump up which argv we were referencing in the original for loop.  This consumed the expected instanceName argument after --daemon|-d properly. When we moved the code over into a function in jutil, we were passing the args by value and the original for loop in daserver.cpp had no idea we had used the instanceName already to write a pidfile.

This change bumps up i in the for loop when we find --daemon|-d and we no longer hit an 'unknown' argument and just print out the usage.

I've gone through the other 13 references for checkCreateDaemon and ensured this issue isn't anywhere else in the branch.

system/jlib/jutil.cpp
```diff
@@ -3282,7 +3282,8 @@ bool checkCreateDaemon(unsigned argc, const char * * argv)
 #ifndef _CONTAINERIZED
     for (unsigned i=0;i<(unsigned)argc;i++) {
         if (streq(argv[i],"--daemon") || streq(argv[i],"-d")) {
-            if (daemon(1,0) || write_pidfile(argv[++i])) {
```
ghalliday, line 3285: Why the extra check (seems a reasonable check), but if you are improving the error checking it should also be checking that i+1 < argc

ghalliday (approved):
@Michael-Gardner change looks good.  One comment related to your extra checking, but could merge as is.

system/jlib/jutil.cpp
Michael-Gardner: I'm just checking that the argument after --daemon isn't another flag.  You're absolutely right, I should add a check for i + 1 < argc.

ghalliday: With the extra check you are preventing the filename from starting with a -, but you should also probably have an equivalent check in daserver.cpp.  It seems an unnecessary extra test.

Michael-Gardner:
Closing to use @lpezet PR's instead

Michael-Gardner:
Reopening and removing the jutil changes. Just doing the increment in daserver.

PR #15532, opened 2021-11-01
HPCC-21700 Created optimized pow expr for ints

jpmcmu:
## Type of change:
- [X] This change improves the code (refactor or other change that does not change the functionality)

## Checklist:
  - [ ] The commit is signed.
  - [X] I have added tests to cover my changes.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization
- [X] There are no similar instances of the same problem that should be addressed

## Smoketest:

### DIFF
--- a/ecl/hqlcpp/hqlcpp.cpp
+++ b/ecl/hqlcpp/hqlcpp.cpp
@@ -3281,7 +3281,7 @@ void HqlCppTranslator::buildExpr(BuildCtx & ctx, IHqlExpression * expr, CHqlBoun
         doBuildExprSysFunc(ctx, expr, tgt, log10Id, options.divideByZeroAction);
         return;
     case no_power:
-        doBuildExprSysFunc(ctx, expr, tgt, powerId);
+        doBuildExprPow(ctx, expr, tgt);
         return;
     case no_fail:
         doBuildStmtFail(ctx, expr);
@@ -7484,6 +7484,66 @@ void HqlCppTranslator::doBuildAssignExecuteWhen(BuildCtx & ctx, const CHqlBoundT
     {
         buildExprAssign(ctx, target, value);
         buildStmt(ctx, action);
+    }
+}
+
+void HqlCppTranslator::doBuildExprPow(BuildCtx & ctx, IHqlExpression * expr, CHqlBoundExpr & tgt)
+{
+    assertex(expr->numChildren() == 2);
+
+    IHqlExpression * right = expr->queryChild(1);
+
+    bool notIntegerPower = !right->queryType()->isInteger();
+    if (notIntegerPower)
+    {
+        doBuildExprSysFunc(ctx, expr, tgt, powerId);
+        return;
+    }
+
+    IValue * value = right->queryValue();
+    bool notConstant = (value == nullptr);
+    if (notConstant) 
+    {
+        doBuildExprSysFunc(ctx, expr, tgt, powerId);
+        return;
+    }
+
+    __int64 rightIntVal = value->getIntValue();
+    __int64 absExp = abs(rightIntVal);
+
+    const unsigned int MAX_EXP_INLINE_VALUE = 3;
+    if (absExp > MAX_EXP_INLINE_VALUE)
+    {
+        doBuildExprSysFunc(ctx, expr, tgt, powerId);
+        return;
+    }
+
+    IHqlExpression * left = expr->queryChild(0);
+    if (rightIntVal == 0)
+    {
+        tgt.expr.setown(createConstant(expr->queryType()->castFrom(1.0)));
+    }
+    else
+    {
+        OwnedITypeInfo realType = makeRealType(DEFAULT_REAL_SIZE);
+        OwnedHqlExpr castLeft = ensureExprType(left, realType);
+
+        CHqlBoundExpr leftBound;
+        buildSimpleExpr(ctx, castLeft, leftBound);
+
+        tgt.set(leftBound);
+        for (int i = 1; i < absExp; i++)
+        {
+            tgt.expr.setown(createValue(no_mul, LINK(realType), tgt.expr.getClear(), LINK(leftBound.expr)));
+        }
+
+        if (rightIntVal < 0)
+        {
+            OwnedHqlExpr numerator = createConstant(realType->castFrom(1.0));
+            OwnedHqlExpr eclDivisor = tgt.getTranslatedExpr();
+            OwnedHqlExpr divide = createValue(no_div, LINK(realType), numerator.getClear(), LINK(eclDivisor));
+            buildExpr(ctx, divide, tgt);
+        }
     }
 }
 
--- a/ecl/regress/power3.ecl
+++ b/ecl/regress/power3.ecl
@@ -1,0 +1,27 @@
+person := dataset('person', { unsigned8 person_id, string1 per_sex, unsigned per_ssn, string40 per_first_name, data9 per_cid, unsigned8 xpos }, thor);
+
+zero := 0.0 : stored('zero');
+output(power(zero, -1));
+
+output(person,
+    {
+        POWER(person.person_id,0),
+        POWER(person.person_id,1),
+        POWER(person.person_id,2),
+        POWER(person.person_id,3),
+        POWER(person.person_id,4),
+        POWER(person.person_id,-1),
+        POWER(person.person_id,-2),
+        POWER(person.person_id,-3),
+        POWER(person.person_id,-4),
+
+        POWER(person.person_id,1.1),
+        POWER(person.person_id,2.1),
+        POWER(person.person_id,3.1),
+        POWER(person.person_id,4.1),
+        POWER(person.person_id,-1.1),
+        POWER(person.person_id,-2.1),
+        POWER(person.person_id,-3.1),
+        POWER(person.person_id,-4.1),
+        0
+    },'out.d00');
--- a/testing/regress/ecl/key/power.xml
+++ b/testing/regress/ecl/key/power.xml
@@ -1,0 +1,42 @@
+<Dataset name='pow_0'>
+ <Row><pow_0>1.0</pow_0></Row>
+</Dataset>
+<Dataset name='pow_1'>
+ <Row><pow_1>1.123456</pow_1></Row>
+</Dataset>
+<Dataset name='pow_2'>
+ <Row><pow_2>1.262153383936</pow_2></Row>
+</Dataset>
+<Dataset name='pow_3'>
+ <Row><pow_3>1.417973792103203</pow_3></Row>
+</Dataset>
+<Dataset name='pow_4'>
+ <Row><pow_4>1.593031164581096</pow_4></Row>
+</Dataset>
+<Dataset name='pow_n1'>
+ <Row><pow_n1>0.8901105161216817</pow_n1></Row>
+</Dataset>
+<Dataset name='pow_n2'>
+ <Row><pow_n2>0.7922967309104065</pow_n2></Row>
+</Dataset>
+<Dataset name='pow_n3'>
+ <Row><pow_n3>0.7052316520721831</pow_n3></Row>
+</Dataset>
+<Dataset name='pow_n4'>
+ <Row><pow_n4>0.6277341098113171</pow_n4></Row>
+</Dataset>
+<Dataset name='pow_5'>
+ <Row><pow_5>0.0</pow_5></Row>
+</Dataset>
+<Dataset name='pow_6'>
+ <Row><pow_6>9.999999999999999e+299</pow_6></Row>
+</Dataset>
+<Dataset name='pow_6_2'>
+ <Row><pow_6_2>0.0</pow_6_2></Row>
+</Dataset>
+<Dataset name='pow_7'>
+ <Row><pow_7>0.5</pow_7></Row>
+</Dataset>
+<Dataset name='pow_8'>
+ <Row><pow_8>8.0</pow_8></Row>
+</Dataset>
--- a/testing/regress/ecl/power.ecl
+++ b/testing/regress/ecl/power.ecl
@@ -1,0 +1,40 @@
+/*##############################################################################
+
+    HPCC SYSTEMS software Copyright (C) 2022 HPCC Systems®.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+############################################################################## */
+
+REAL    real_const := 1.123456;
+REAL    rVal := real_const : STORED('real_stored');
+
+OUTPUT(POWER(rVal,0), named('pow_0'));
+OUTPUT(POWER(rVal,1), named('pow_1'));
+OUTPUT(POWER(rVal,2), named('pow_2'));
+OUTPUT(POWER(rVal,3), named('pow_3'));
+OUTPUT(POWER(rVal,4), named('pow_4'));
+OUTPUT(POWER(rVal,-1), named('pow_n1'));
+OUTPUT(POWER(rVal,-2), named('pow_n2'));
+OUTPUT(POWER(rVal,-3), named('pow_n3'));
+OUTPUT(POWER(rVal,-4), named('pow_n4'));
+
+REAL    zero := 0.0 : STORED('zero');
+OUTPUT(POWER(zero, -1), named('pow_5'));
+
+REAL    nearZero := 1.0E-300 : STORED('nearZero');
+OUTPUT(POWER(nearZero, -1), named('pow_6'));
+OUTPUT(POWER(nearZero, -2), named('pow_6_2'));
+
+UNSIGNED1 uint2 := 2 :STORED('uint2');
+OUTPUT(POWER(uint2, -1), named('pow_7'));
+OUTPUT(POWER(uint2, 3), named('pow_8'));
--- a/ecl/hqlcpp/hqlcpp.ipp
+++ b/ecl/hqlcpp/hqlcpp.ipp
@@ -1413,6 +1413,7 @@ public:
     void doBuildExprCountDict(BuildCtx & ctx, IHqlExpression * expr, CHqlBoundExpr & tgt);
     void doBuildExprCount(BuildCtx & ctx, IHqlExpression * expr, CHqlBoundExpr & tgt);
     void doBuildExprCounter(BuildCtx & ctx, IHqlExpression * expr, CHqlBoundExpr & tgt);
+    void doBuildExprPow(BuildCtx & ctx, IHqlExpression * expr, CHqlBoundExpr & tgt);
     void doBuildExprDivide(BuildCtx & ctx, IHqlExpression * expr, CHqlBoundExpr & tgt);
     void doBuildExprEmbedBody(BuildCtx & ctx, IHqlExpression * expr, CHqlBoundExpr * tgt);
     void doBuildExprEvaluate(BuildCtx & ctx, IHqlExpression * expr, CHqlBoundExpr & tgt);


ecl/hqlcpp/hqlcpp.cpp
```diff
@@ -7487,6 +7487,83 @@ void HqlCppTranslator::doBuildAssignExecuteWhen(BuildCtx & ctx, const CHqlBoundT
     }
 }
 
+void HqlCppTranslator::doBuildExprPow(BuildCtx & ctx, IHqlExpression * expr, CHqlBoundExpr & tgt)
+{
+    assertex(expr->numChildren() == 2);
+    
+    IHqlExpression * left = expr->queryChild(0);
```
ghalliday, line 7494: minor: left is not needed  until later, so may as well delay that until later.

ecl/hqlcpp/hqlcpp.cpp
```diff
@@ -7487,6 +7487,83 @@ void HqlCppTranslator::doBuildAssignExecuteWhen(BuildCtx & ctx, const CHqlBoundT
...
+    IValue * value = right->queryValue();
+    bool notConstant = (value == nullptr);
+    if (notConstant) 
+    {
+        doBuildExprSysFunc(ctx, expr, tgt, powerId);
+        return;
+    }
+
+    const uint MAX_EXP_INLINE_VALUE = 3;
+    int64_t rightIntVal = value->getIntValue();
```
ghalliday, line 7513: trivial: We tend to use __int64 rather than int64_t.

jpmcmu:
@ghalliday missing a change, about to push it up. Did have a few questions about the best way to handle div / cast.

ecl/hqlcpp/hqlcpp.cpp
```diff
@@ -7487,6 +7487,83 @@ void HqlCppTranslator::doBuildAssignExecuteWhen(BuildCtx & ctx, const CHqlBoundT
...
+    if (abs(rightIntVal) > MAX_EXP_INLINE_VALUE)
+    {
+        doBuildExprSysFunc(ctx, expr, tgt, powerId);
+        return;
+    }
+
+    IHqlExpression* outExpr = nullptr;
```
ghalliday, line 7520: A couple of issues:
a. Expression link counts.  The case right==0 will create an expression with a link count of 1, the case right=1 is no linking the result.  The expression is not released - so it will leak for right==0 (and right != 1)
You should check changes carefully for memory leaks (can use valgrind, but need to use the --leakcheck option to ensure everything is cleaned up correctly.

b. Different kinds of expressions.  The same structures are used to represent ECL expressions and c++ expressions, but you need to be very careful about not mixing them.  So buildExpr() expects an ECL expression, but on line 7563 you are passing it a c++ expression.  That could cause problems.

I think in this case you should not be calling buildExpr at all, but should be setting tgt.expr.

ecl/hqlcpp/hqlcpp.cpp
```diff
@@ -7487,6 +7487,83 @@ void HqlCppTranslator::doBuildAssignExecuteWhen(BuildCtx & ctx, const CHqlBoundT
...
+        // Is there a better way to get a temp real here?
+        //---------------------------------------------------------------------------
+
+        CHqlBoundExpr leftBound;
+        buildTempExpr(ctx, left, leftBound);
+
+        CHqlBoundExpr tempReal;
+        OwnedITypeInfo realType = makeRealType(DEFAULT_REAL_SIZE);
+        doBuildCastViaTemp(ctx, realType, leftBound, tempReal);
+        outExpr = tempReal.expr;
```
ghalliday, line 7542: Another mixture of not linking here, and linking within the loop;

ecl/hqlcpp/hqlcpp.cpp
```diff
@@ -7487,6 +7487,83 @@ void HqlCppTranslator::doBuildAssignExecuteWhen(BuildCtx & ctx, const CHqlBoundT
...
+
+        for (int i = 1; i < abs(rightIntVal); i++)
+        {
+            outExpr = createValue(no_mul, realType, LINK(outExpr), LINK(tempReal.expr));
```
ghalliday, line 7546: realType should be linked - it would crash if the expression wasn't leaked.

ecl/hqlcpp/hqlcpp.cpp
```diff
@@ -7487,6 +7487,83 @@ void HqlCppTranslator::doBuildAssignExecuteWhen(BuildCtx & ctx, const CHqlBoundT
...
+
+        if (rightIntVal < 0)
+        {
+            CHqlBoundExpr bound;
+            buildTempExpr(ctx, outExpr, bound);
+
+            //---------------------------------------------------------------------------
+            // Question:
+            // This inserts a check for divide by zero. What is the best way to not have
+            // that check generated?
```
ghalliday, line 7557: The value could be zero, so it needs to have that check.

ecl/hqlcpp/hqlcpp.cpp
```diff
@@ -7487,6 +7487,83 @@ void HqlCppTranslator::doBuildAssignExecuteWhen(BuildCtx & ctx, const CHqlBoundT
...
+
+        for (int i = 1; i < abs(rightIntVal); i++)
+        {
+            outExpr = createValue(no_mul, realType, LINK(outExpr), LINK(tempReal.expr));
+        }
+
+        if (rightIntVal < 0)
+        {
+            CHqlBoundExpr bound;
+            buildTempExpr(ctx, outExpr, bound);
```
ghalliday, line 7552: again not needed - it is already a c++ expression.

ecl/hqlcpp/hqlcpp.cpp
```diff
@@ -7487,6 +7487,83 @@ void HqlCppTranslator::doBuildAssignExecuteWhen(BuildCtx & ctx, const CHqlBoundT
...
+
+            //---------------------------------------------------------------------------
+            // Question:
+            // This inserts a check for divide by zero. What is the best way to not have
+            // that check generated?
+            //---------------------------------------------------------------------------
+            outExpr = createValue(no_div, realType, createConstant(1.0), bound.expr);
```
ghalliday, line 7559: tgt.expr.setown(createValue(no_div, LINK(realType), createConstant(1.0), LINK(tgt.expr));

ghalliday (changes requested):
@jpmcmu a few comments, main issues are making sure link counting is correct, and difference between translated(c++), and untranslated expressions (ECL)

ecl/hqlcpp/hqlcpp.cpp
jpmcmu: Right, I was looking to move that check above the entire expansion. I ended up doing this during the code review changes, not sure if the approach I took was the best route though.

jpmcmu:
@ghalliday implemented code review changes, please review

ghalliday:
@jpmcmu I'll take a look.  For future reference you can click refresh on the review to re-request a review.  (I have done it).

ecl/hqlcpp/hqlcpp.cpp
```diff
@@ -7487,6 +7487,74 @@ void HqlCppTranslator::doBuildAssignExecuteWhen(BuildCtx & ctx, const CHqlBoundT
...
+    if (notConstant) 
+    {
+        doBuildExprSysFunc(ctx, expr, tgt, powerId);
+        return;
+    }
+
+    __int64 rightIntVal = value->getIntValue();
+    __int64 absExp = abs(rightIntVal);
+
+    const uint MAX_EXP_INLINE_VALUE = 3;
```
ghalliday, line 7514: should be unsigned not uint (it causes the windows build to fail)

ecl/hqlcpp/hqlcpp.cpp
```diff
@@ -7487,6 +7487,74 @@ void HqlCppTranslator::doBuildAssignExecuteWhen(BuildCtx & ctx, const CHqlBoundT
...
+    {
+        tgt.expr.setown(createConstant(left->queryType()->castFrom(1.0)));
+    }
+    else if (rightIntVal == 1)
+    {
+        tgt.expr.setown(LINK(leftBound.expr));
+    }
+    else
+    {
+        LINK(leftBound.expr);
```
ghalliday, line 7538: This should be combined with the outExpr definition:

        IHqlExpression* outExpr = LINK(leftBound.expr);

But better is to use an OwnedHqlExpr so that if an exception is thrown it will be cleanly released.

ecl/hqlcpp/hqlcpp.cpp
```diff
@@ -7487,6 +7487,74 @@ void HqlCppTranslator::doBuildAssignExecuteWhen(BuildCtx & ctx, const CHqlBoundT
...
+        }
+
+        tgt.expr.setown(outExpr);
+
+        if (rightIntVal < 0)
+        {
+            tgt.expr.setown(createValue(no_div, LINK(realType), createConstant(1.0), LINK(tgt.expr)));
+        }
+
+        OwnedHqlExpr cond = createBoolExpr(no_ne, LINK(leftBound.expr), getZero());
```
ghalliday, line 7553: The types of the left and right hand sizes of this expression do not match - that may trigger internal checks.

ecl/hqlcpp/hqlcpp.cpp
```diff
@@ -7487,6 +7487,74 @@ void HqlCppTranslator::doBuildAssignExecuteWhen(BuildCtx & ctx, const CHqlBoundT
...
+        tgt.expr.setown(createValue(no_if, leftBound.expr->getType(), LINK(cond), LINK(tgt.expr), getZero()));
```
ghalliday, line 7554: The behaviour of division by zero can be configured. (See doBuildAssignDivide).

Also, the types of the the left and right sides of the if() do not match - real and integer.  Better would be to reuse the divide code.

ecl/regress/power3.ecl
```diff
@@ -0,0 +1,24 @@
+person := dataset('person', { unsigned8 person_id, string1 per_sex, unsigned per_ssn, string40 per_first_name, data9 per_cid, unsigned8 xpos }, thor);
+
+output(person,
```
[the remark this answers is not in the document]
ghalliday: Still need tests in testing/regress/ecl

ecl/hqlcpp/hqlcpp.cpp
```diff
@@ -7487,6 +7487,74 @@ void HqlCppTranslator::doBuildAssignExecuteWhen(BuildCtx & ctx, const CHqlBoundT
...
+    OwnedHqlExpr castLeft = ensureExprType(left, realType);
+    buildTempExpr(ctx, castLeft, leftBound);
+
+    if (rightIntVal == 0)
+    {
+        tgt.expr.setown(createConstant(left->queryType()->castFrom(1.0)));
+    }
+    else if (rightIntVal == 1)
+    {
+        tgt.expr.setown(LINK(leftBound.expr));
```
ghalliday, line 7534: tgt.set(leftBound)

is simpler

ghalliday: Even better to avoid the special case since the general case will generate the same.

ecl/hqlcpp/hqlcpp.cpp
ghalliday: This is only needed if rightIntVal < 0.

ghalliday (commented):
@jpmcmu getting closer.   A few more comments.

AttilaVamos (changes requested):

testing/regress/ecl/power.ecl
```diff
@@ -0,0 +1,12 @@
+REAL    real_const := 1.123456;
```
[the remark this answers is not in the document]
jpmcmu: @AttilaVamos Thanks fixed.

jpmcmu:
@ghalliday Should I be rounding values coming from POWER in the tests to avoid floating point comparison issues between different cpu archs / compiler flags?

ecl/hqlcpp/hqlcpp.cpp
```diff
@@ -7487,6 +7487,81 @@ void HqlCppTranslator::doBuildAssignExecuteWhen(BuildCtx & ctx, const CHqlBoundT
     }
 }
 
+void HqlCppTranslator::doBuildExprPow(BuildCtx & ctx, IHqlExpression * expr, CHqlBoundExpr & tgt)
+{
+    assertex(expr->numChildren() == 2);
+    
+    IHqlExpression * right = expr->queryChild(1);
+
```
ghalliday, line 7495: whitespace: wouldn't normally have a blank line here

ecl/hqlcpp/hqlcpp.cpp
```diff
@@ -7487,6 +7487,81 @@ void HqlCppTranslator::doBuildAssignExecuteWhen(BuildCtx & ctx, const CHqlBoundT
...
+        doBuildExprSysFunc(ctx, expr, tgt, powerId);
+        return;
+    }
+    
+    OwnedITypeInfo realType = makeRealType(DEFAULT_REAL_SIZE);
+    CHqlBoundExpr leftBound;
+    
+    IHqlExpression * left = expr->queryChild(0);
+    OwnedHqlExpr castLeft = ensureExprType(left, realType);
+    buildTempExpr(ctx, castLeft, leftBound);
```
ghalliday, line 7526: trivial: This code can move inside the else branch.

ecl/hqlcpp/hqlcpp.cpp
```diff
@@ -7487,6 +7487,81 @@ void HqlCppTranslator::doBuildAssignExecuteWhen(BuildCtx & ctx, const CHqlBoundT
...
+        for (int i = 1; i < absExp; i++)
+        {
+            outExpr.setown(createValue(no_mul, LINK(realType), outExpr.getClear(), LINK(leftBound.expr)));
+        }
+
+        tgt.expr.setown(outExpr.getClear());
+
+        if (rightIntVal < 0)
+        {
+            OwnedHqlExpr numerator = createConstant(tgt.expr->queryType()->castFrom(1.0));
```
ghalliday, line 7544: minor: use realType instead of tgt.expr->queryType()

ecl/hqlcpp/hqlcpp.cpp
```diff
@@ -7487,6 +7487,81 @@ void HqlCppTranslator::doBuildAssignExecuteWhen(BuildCtx & ctx, const CHqlBoundT
...
+            IHqlExpression* divisor = tgt.expr.get();
```
ghalliday, line 7545: minor: "= tgt.expr" would be more normal.  We don't normally call get unless required (e.g. in a printf)

ecl/hqlcpp/hqlcpp.cpp
```diff
@@ -7487,6 +7487,81 @@ void HqlCppTranslator::doBuildAssignExecuteWhen(BuildCtx & ctx, const CHqlBoundT
...
+            OwnedHqlExpr pureExpr = createValue(no_div, left->getType(), numerator.getClear(), LINK(divisor));
```
ghalliday, line 7546: bug: left->getType() is theoretically wrong - this should be realType (not sure if it currently can be non-real).

ecl/hqlcpp/hqlcpp.cpp
```diff
@@ -7487,6 +7487,81 @@ void HqlCppTranslator::doBuildAssignExecuteWhen(BuildCtx & ctx, const CHqlBoundT
...
+
+            CHqlBoundTarget tempTarget;
+            createTempFor(ctx, expr, tempTarget);
+
+            IValue * zero = left->queryType()->castFrom(false, 0);
```
ghalliday, line 7551: bug: left->queryType() is not correct.

ecl/hqlcpp/hqlcpp.cpp
```diff
@@ -7487,6 +7487,81 @@ void HqlCppTranslator::doBuildAssignExecuteWhen(BuildCtx & ctx, const CHqlBoundT
...
+            OwnedHqlExpr eZero = createConstant(zero);
+
+            BuildCtx subctx(ctx);
+            IHqlStmt * stmt = subctx.addFilter(leftBound.expr);
```
ghalliday, line 7555: bug: This test needs to be on pureExpr, (but avoid re-evaluating it).  If left is 1E-1000 and right is -2, then left*left will be 0 (due to real underflow), but this test will not be triggered.

ecl/hqlcpp/hqlcpp.cpp
```diff
@@ -7522,8 +7597,6 @@ void HqlCppTranslator::doBuildExprDivide(BuildCtx & ctx, IHqlExpression * expr,
     }
 }
 
-
```
ghalliday, line 7525: whitespace: unrelated change

ecl/hqlcpp/hqlcpp.cpp
ghalliday, line 7526: minor: better to call buildSimpleExpr() than buildTempExpr().  It will reduce the number of occasions that a temporary variable is actually created.

testing/regress/ecl/power.ecl
```diff
@@ -0,0 +1,29 @@
...
+
+OUTPUT(POWER(rVal,0), named('pow_0'));
+OUTPUT(POWER(rVal,1), named('pow_1'));
+OUTPUT(POWER(rVal,2), named('pow_2'));
+OUTPUT(POWER(rVal,3), named('pow_3'));
+OUTPUT(POWER(rVal,4), named('pow_4'));
+OUTPUT(POWER(rVal,-1), named('pow_n1'));
+OUTPUT(POWER(rVal,-2), named('pow_n2'));
+OUTPUT(POWER(rVal,-3), named('pow_n3'));
+OUTPUT(POWER(rVal,-4), named('pow_n4'));
```
ghalliday, line 29: This should include test cases for the unusual conditions.  POWER(0, -1), POWER(1E-1000,-2)

ghalliday (commented):
@jpmcmu see comments

ecl/hqlcpp/hqlcpp.cpp
jpmcmu: Good point. Is there a way to do an epsilon comparison with addFilter?

ghalliday: I don't think that is the way to do it - it will end up depending on which power you raise the number to, and possibly the length.  You need to be checking that the multiplied value is 0.

jpmcmu:
@ghalliday implemented code review changes, squashed and targeted 8.6.X

testing/regress/ecl/key/power.xml
```diff
@@ -0,0 +1,40 @@
...
+ <Row><pow_n2>0.7922967309104065</pow_n2></Row>
+</Dataset>
+<Dataset name='pow_n3'>
+ <Row><pow_n3>0.7052316520721831</pow_n3></Row>
+</Dataset>
+<Dataset name='pow_n4'>
+ <Row><pow_n4>0.6277341098113171</pow_n4></Row>
+</Dataset>
+<Dataset name='pow_5'>
+ <Row><pow_5>0</pow_5></Row>
```
AttilaVamos, line 29: The generated result is 0.0

testing/regress/ecl/key/power.xml
```diff
@@ -0,0 +1,40 @@
...
+</Dataset>
+<Dataset name='pow_6'>
+ <Row><pow_6>0</pow_6></Row>
```
AttilaVamos, line 32: The generated result is 0.0

AttilaVamos (changes requested):
Some mismatches between the expected (key file) and generated result.

AttilaVamos:
@jpmcmu The "280. genjoin2, 281. genjoin3, 416. joinattr2, 425. key" errors on Thor are not related to this PR.

testing/regress/ecl/power.ecl
```diff
@@ -0,0 +1,39 @@
...
+
+REAL    zero := 0.0 : STORED('zero');
+OUTPUT(POWER(zero, -1), named('pow_5'));
+
+REAL    nearZero := 1.0E-1000 : STORED('nearZero');
```
ghalliday, line 34: My mistake- this is actually unrepresentable (mixing up base 2 and base 10).  Should be 1.0E-300.

ghalliday (commented):
Looks close - need to tidy up key files, add a test case, and check through comments for any missed items (e.g., unrelated newline change).

Note, this does not match code like
x := POWER(value, 2.0).
But there are relatively few instances, so I would prefer to merge as is.

ghalliday:
> @jpmcmu The "280. genjoin2, 281. genjoin3, 416. joinattr2, 425. key" errors on Thor are not related to this PR.

@AttilaVamos I had a look at the zap report - and it looks like the weird issues are problem with the regression suite.  The zap report for one of the workunits contained results.  Is there an open issue for it?

testing/regress/ecl/key/power.xml
```diff
@@ -0,0 +1,43 @@
...
+<Dataset name='pow_6_2'>
+ <Row><pow_6_2>0.0</pow_6_2></Row>
+</Dataset>
+<Dataset name='pow_7'>
+ <Row><pow_7>0.5</pow_7></Row>
+</Dataset>
+<Dataset name='pow_8'>
+ <Row><pow_8>8.0</pow_8></Row>
+</Dataset>
+
```
AttilaVamos, line 43: This extra empty line caused the errors. Please remove it.

AttilaVamos (changes requested):

jpmcmu:
@AttilaVamos @ghalliday Key issues fixed please review

AttilaVamos (approved):
Looks good (in testing POW).

ghalliday (approved):

PR #15559, opened 2021-11-08
HPCC-26795 Reduce number of critical sections in udp transport layer

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [x] Performance
  - [ ] Security
  - [x] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/roxie/udplib/udpsha.cpp
+++ b/roxie/udplib/udpsha.cpp
@@ -91,39 +91,58 @@ queue_t::~queue_t()
     }
 }
 
-unsigned queue_t::available()
-{
-    CriticalBlock b(c_region);
-    if (count < limit)
-        return limit - count;
-    return 0;
-}
-
-int queue_t::free_slots() 
-{
-    int res=0;
-    while (res <= 0)
-    {
-        c_region.enter();
-        res = limit - count;
-        if (res <= 0)
-            signal_free_sl++;
-        c_region.leave();
-        if (res <= 0)
-        {
-            while (!free_sl.wait(3000))
-            {
-                if (udpTraceLevel >= 1)
-                    DBGLOG("queue_t::free_slots blocked for 3 seconds waiting for free_sl semaphore");
-            }
-        }
-    }
-    return res;
-}
-
 void queue_t::interrupt()
 {
     data_avail.interrupt();
+}
+
+void queue_t::doEnqueue(DataBuffer *buf)
+{
+    // Must currently be called within a critical section.  Does not signal - that should be done outside the CS.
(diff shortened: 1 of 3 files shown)


ghalliday:
Currrently untested...

mckellyln (approved):
Looks great.
One question - should we pushOwnWait() on the input queue also ?
```
                    parent.input_queue->pushOwn(b);
```

mckellyln:
And should I do a build and test on the cert cluster ?

ghalliday:
> Looks great. One question - should we pushOwnWait() on the input queue also ?
> 
> ```
>                     parent.input_queue->pushOwn(b);
> ```
No - that code is correct.  Richard changed the receive queue so that it could always receive new entries to prevent it causing problems pulling data off the read socket.  If the flow control goes wrong (permits expire) then more packets could be pushed than the queue length.

roxie/udplib/udpsha.cpp
```diff
@@ -91,39 +91,58 @@ queue_t::~queue_t()
...
-        res = limit - count;
-        if (res <= 0)
-            signal_free_sl++;
-        c_region.leave();
-        if (res <= 0)
         {
-            while (!free_sl.wait(3000))
+            CriticalBlock b(c_region);
+
+            if (count < limit)
```
richardkchapman, line 131: I wonder if you could check this before grabbing the critsec, and then either check again after or accept that just occasionally you would get an incorrect answer if count was being updated by another thread - would it matter?

Not sure it would make a lot of difference, depends on whether it's normal for the queue to be full when first entering.

roxie/udplib/udpsha.hpp
```diff
@@ -88,27 +88,34 @@ class queue_t
     DataBuffer *head = nullptr;      // We add at tail and remove from head
     DataBuffer *tail = nullptr;
 
-    unsigned count = 0;
+    RelaxedAtomic<unsigned> count{0};       // always updated inside a critical section, by atomic to guarantee it can be read from outside.
```
richardkchapman, line 91: What does "by atomic" mean?

ghalliday: @richardkchapman updated the comment and repushed

roxie/udplib/udpsha.cpp
ghalliday: I think if true the doEnqueue() needs to be inside the critical section, and if false he signal_free_sl needs to be inside the cs, so there wouldn't be any benefit.

PR #15574, opened 2021-11-12
HPCC-26819 Ensure ISocket::close() aborts reads from the socket

ghalliday:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jsocket.cpp
+++ b/system/jlib/jsocket.cpp
@@ -497,7 +497,7 @@ private:
     #ifdef _WIN32
             return ::closesocket(s);
     #else
-            ::shutdown(s, SHUT_WR);
+            ::shutdown(s, SHUT_RDWR);
             return ::close(s);
     #endif
         }
@@ -516,7 +516,7 @@ Semaphore CSocket::receiveblocksem(2);
 #ifdef _WIN32
 class win_socket_library 
 {
-    static bool initdone; // to prevent dependancy probs very early on (e.g. jlog)
+    static bool initdone; // to prevent dependency probs very early on (e.g. jlog)
 public:
     win_socket_library() { init(); }
     bool init()


ghalliday:
@mckellyln please review/sanity check - I'm changing code I don't really understand....
Without this change the improved simulated udp tests sometimes waited for 5 seconds - which corresponded to the time in

`                        flow_socket->read(&f, readsize, readsize, res, 5);
`
With this change in place that delay never occurred.  I can't think of any downsides, but you will have a much better idea.  I imagine it may help in other situations as well.

ghalliday:
The failures may be related to this PR - although not sure why.  Possibly items being added to the thor queue but not being noticed as removed?

mckellyln (commented):
What is the specific case this is solving ?  Is it a wait_read() not terminating and then another thread calling shutdown() ?  Feels a bit like setting nolinger.
I am little apprehensive about adding RD in case there is any queued data still ?

mckellyln:
I will test.  the close() should trigger the poll() in wait_read() and then the read's recv() should return 0 indicating a close.

ghalliday:
> What is the specific case this is solving ? Is it a wait_read() not terminating and then another thread calling shutdown() ? Feels a bit like setting nolinger.

Yes that is exactly the case - the flow thread is waiting for another flow control message, but the object is being destroyed, so it calls close on the thread.  That *sometimes* takes the full timeout before aborting the read.

> I am little apprehensive about adding RD in case there is any queued data still ?
It seems to be causing an issue - although I'm not sure why it would.

mckellyln:
I think close() in one thread does not do the same thing to the socket as close() in another process would and so the shutdown(RD) is needed to get the other thread to see an event in poll().

mckellyln:
Can you call sock->shutdown(2) in your code when you know you are finished ?

ghalliday:
Following marks' comments I don't think this is the correct fix - so will close and open a new PR later

PR #15576, opened 2021-11-12
HPCC-26821 Add preStop hook for clean k8s autoscale of Roxie

richardkchapman:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [ ] Performance
  - [ ] Security
  - [x] Thread-safety
  - [x] Cloud-compatibility
  - [x] Premature optimization

## Smoketest:

### DIFF
--- a/helm/hpcc/templates/roxie.yaml
+++ b/helm/hpcc/templates/roxie.yaml
@@ -253,6 +253,7 @@ spec:
       serviceAccountName: "hpcc-default"
       initContainers:
       {{- include "hpcc.changePlaneMountPerms" $commonCtx | indent 6 }}
+      terminationGracePeriodSeconds: {{ add ($roxie.agentQueryReleaseDelaySeconds | default 60) 30 }}
       containers:
       - name: {{ $servername | quote }}
         workingDir: /var/lib/HPCCSystems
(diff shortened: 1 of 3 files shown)


ghalliday (commented):
@richardkchapman Changes look ok, but how does this interact with running queries?  I wasn't convinced it would wait for them to complete, but the || queue.length() change might be the fix.  I assume k8s prevents new queries being sent to it once close-down is initiated.

richardkchapman:
k8s will stop new queries going to the server once it has decided to close it down.

The idea is that once the server gets the indication that it is closing down, it updates the topology so that it should stop receiving agent requests - but that takes a while to be sure there are none in the pipeline/still running, which is what the delayedReleaser is for.

There are two changes in here related to the delayedReleaser - making sure it is employed when deleting the package manager (i.e. on closedown) as well as when deploying a new package, and making sure it waits for all delays to expire even when closing down.

roxie/ccd/ccdmain.cpp
```diff
@@ -297,11 +297,14 @@ class Waiter : public CInterface, implements IAbortHandler
     }
 } waiter;
 
+Semaphore closedDown;
```
jakesmith, line 300: trivial: could be static

helm/hpcc/templates/roxie.yaml
```diff
@@ -242,6 +242,7 @@ spec:
       serviceAccountName: "hpcc-default"
       initContainers:
       {{- include "hpcc.changePlaneMountPerms" $commonCtx | indent 6 }}
+      terminationGracePeriodSeconds: 100
```
jakesmith, line 245: is 100 seconds always enough?
Should it be configurable and propagate through, so that "agentQueryReleaseDelaySeconds" is based on it?

jakesmith (approved):
@richardkchapman - looks good, 1 trivial, 1 question.

ghalliday (approved):
Not new, but at some point should agentQueryReleaseDelaySeconds appear in the schema?

richardkchapman:
There are a LOT of Roxie options that do not appear in the schema. Many of them ones we would prefer that people didn't mess with.

I can open a Jira to look into expanding the schema sometime.

richardkchapman:
Added Jira https://track.hpccsystems.com/browse/HPCC-26911

jakesmith (approved):
@richardkchapman - looks good to me, bar squashing.

PR #15581, opened 2021-11-12
Multi permits

richardkchapman:
## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/roxie/udplib/udpsha.cpp
+++ b/roxie/udplib/udpsha.cpp
@@ -777,6 +777,8 @@ void CSimulatedReadSocket::readtms(void* buf, size32_t min_size, size32_t max_si
 int CSimulatedReadSocket::wait_read(unsigned timeout)
 {
     bool ret = avail.wait(timeout);
+//    if (me.port==9001)
+  //      MilliSleep(1);
     return ret;
 }
 
@@ -800,6 +802,7 @@ class SimulatedUdpStressTest : public CppUnit::TestFixture
             udpResendLostPackets = true;
             udpRequestToSendTimeout = 1000;
             udpRequestToSendAckTimeout = 1000;
+            udpMaxPendingPermits = 10;
             isUdpTestMode = true;
             roxiemem::setTotalMemoryLimit(false, false, false, 20*1024*1024, 0, NULL, NULL);
             dbm.setown(roxiemem::createDataBufferManager(roxiemem::DATA_ALIGNMENT_SIZE));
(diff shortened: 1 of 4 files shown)


roxie/udplib/udptrr.cpp
```diff
@@ -325,7 +326,11 @@ class CReceiveManager : implements IReceiveManager, public CInterface
         Owned<ISocket> flow_socket;
         const unsigned flow_port;
         const unsigned maxSlotsPerSender;
+        const unsigned maxPermits = 0;
         std::atomic<bool> running = { false };
+        unsigned permitsIssued = 0;
+        inline void returnPermits(unsigned &permits) { permitsIssued -= permits; permits = 0; }
```
ghalliday, line 332: setting to zero is a bit confusing - at least one call also clears it afterwards.  I think better for the caller to explicitly clear

roxie/udplib/udptrr.cpp
```diff
@@ -398,11 +410,12 @@ class CReceiveManager : implements IReceiveManager, public CInterface
                 break;
             }
             requester->state = flowType::send_completed;
+            requester->permitsIssued = 0;
```
ghalliday, line 413: cleaner to move following line 401.

roxie/udplib/udptrr.cpp
```diff
@@ -356,17 +363,21 @@ class CReceiveManager : implements IReceiveManager, public CInterface
...
                 // Fall through
             case flowType::send_completed:
                 pendingPermits.append(requester);
+                requester->permitsIssued = allocatePermits(slots);
                 requester->state = flowType::ok_to_send;
                 break;
             case flowType::ok_to_send:
                 // Perhaps the sender never saw our permission? Already on queue...
+                returnPermits(requester->permitsIssued);
+                requester->permitsIssued = allocatePermits(slots);
```
ghalliday, line 380: This means that the number of slots to allocated is decided without including the number that has just completed.  That could cause issues if large numbers have been allocated

roxie/udplib/udptrr.cpp
ghalliday: Could make this an assert that permitsIssued == 0

roxie/udplib/udptrr.cpp
```diff
@@ -339,6 +344,8 @@ class CReceiveManager : implements IReceiveManager, public CInterface
                 // Fall through
             case flowType::send_completed:
                 pendingRequests.append(requester);
+                returnPermits(requester->permitsIssued);
```
ghalliday, line 347: I think this only needs to be in the previous case (ok_to_send) - might make the code clearer.

roxie/udplib/udpsha.cpp
```diff
@@ -777,6 +777,8 @@ void CSimulatedReadSocket::readtms(void* buf, size32_t min_size, size32_t max_si
 int CSimulatedReadSocket::wait_read(unsigned timeout)
 {
     bool ret = avail.wait(timeout);
+//    if (me.port==9001)
```
ghalliday, line 780: remove?

ghalliday (commented):
@richardkchapman a couple of comments and a couple of possible changes/improvements

- You could add code to decrement permitsIssued for a sender, and permitsIssued for the flow when a data packet is received.  A slightly higher overhead might allow other senders to begin earlier.  I'm not sure how big a benefit it will actually provided in addition to the send_start releasing permits early.  You might then need to check if other permits can be granted after packets are sent, and might need a minimum number of slots allocated.  I suspect the extra complicaton isn't worth it, but it would be interesting to see a trace of the permits allocated on a real system and see what is going on.

PR #15586, opened 2021-11-15
Hpcc 24577 remote log fetch

rpastrana:
## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:


rpastrana:
Replacement PR for long running PR: https://github.com/hpcc-systems/HPCC-Platform/pull/15073

ghalliday (commented):
@rpastrana I have got through the PR and commented on all sorts of different things, including some style comments that I would not normally comment on.  My general comments are

- The interface should be clearer. Both easier to find and better documented.
- The interface seems to have intermediate classes that don't really provide any value and complicate understanding it.  They make the processing code seem much more complicated than it is.
- It is useful to think about what needs to be public, and what can be private.  The more extra code there is in the public interfaces the harder it is to understand it, or see what is important.  E.g. Some code is in header files that should be in c++ files.  (This is a problem with jlog in general, so in a sense you were copying the existing style.)

Other comments
- How does this fits in with the existing log search functions?
- Are you planning on implementing something that uses the log files (as the original jira mentions)?

rpastrana (commented):
@ghalliday please review latest 2 commits, and take a look at my responses.

system/jlib/jlog.hpp
```diff
@@ -1343,4 +1345,281 @@ interface IComponentLogFileCreator : extends IInterface
...
+
+typedef enum
+{
+    LOGACCESS_FILTER_jobid,
+    LOGACCESS_FILTER_class,
+    LOGACCESS_FILTER_audience,
+    LOGACCESS_FILTER_component,
+    LOGACCESS_FILTER_or,
+    LOGACCESS_FILTER_and,
+    LOGACCESS_FILTER_wildcard,
```
ghalliday, line 1395: not clear what this means.  I had assumed it was a wildcard match on the text, but it looks like it matches everything.

system/jlib/jlog.hpp
```diff
@@ -1343,4 +1345,345 @@ interface IComponentLogFileCreator : extends IInterface
...
+        return LOGACCESS_LOGFORMAT_xml;
+    else if(strieq(name, "json"))
+        return LOGACCESS_LOGFORMAT_json;
+    else if(strieq(name, "csv"))
+        return LOGACCESS_LOGFORMAT_csv;
+    else
+        throw makeStringExceptionV(-1, "Encountered unknown Log Access Format name: '%s'", name);
+}
+
+interface IRemoteLogAccess : extends IInterface
```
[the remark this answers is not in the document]
ghalliday: My suggestion would be twofold:
- clearly separate the functions relating to log retrieval from log generation.  E.g., using a divider like
```
//------------------------------------------------------------------------------------
// Functions relating to log retrieval
```

Move many of the inline functions out of line.
A couple of reasons.  They make it hard to get an appreciation of the public interface because the inline functions occupy lots of vertical space.  They are also need to be read by every file that #includes jllog.hpp - which is probably every file in the system.  Each function adds a marginal compile time overhead, but if it was a general pattern it would add up.

The question is if someone is looking for something in this file, can they quickly find it?  If they are reading it to get an overview can they clearly see the structure?

system/logaccess/ElasticStack/ElasticStackLogAccess.hpp
```diff
@@ -0,0 +1,118 @@
...
+    static constexpr const char * LOGMAP_COLUMNNAME_PATH = "logmap/global/@searchcolumn";
+    static constexpr const char * LOGMAP_TIMESTAMPCOL_PATH = "logmap/global/@timestampcolumn";
+
+    ElasticStackLogAccess(const std::vector<std::string> &hostUrlList, IPropertyTree & logAccessPluginConfig);
+    virtual ~ElasticStackLogAccess() override = default;
+
+    virtual bool fetchLog(const LogAccessConditions & options, StringBuffer & returnbuf, LogAccessLogFormat format) override;
+    cpr::Response performESQuery(const LogAccessConditions & options);
+
+    const IPropertyTree * getESStatus();
```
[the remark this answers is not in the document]
ghalliday: I guess my question is that if this is only accessed through the IRemoteLogAccess interface, what other classes/functions are you expecting will call these public functions?

system/logaccess/ElasticStack/ElasticStackLogAccess.hpp
```diff
@@ -0,0 +1,122 @@
...
+    virtual ~ElasticStackLogAccess() override = default;
+
+    // IRemoteLogAccess methods
+    virtual bool fetchLog(const LogAccessConditions & options, StringBuffer & returnbuf, LogAccessLogFormat format) override;
+    virtual const char * getRemoteLogAccessType() const override { return type; }
+    virtual IPropertyTree * queryLogMap() const override { return m_pluginCfg->queryPropTree("logmap");}
+    virtual const char * fetchConnectionStr() const override { return m_esConnectionStr.str();}
+
+    // Elastic Stack specific
+    cpr::Response performESQuery(const LogAccessConditions & options);
```
ghalliday, line 115: This looks like a private method - why would anything outside the class call this?

system/logaccess/ElasticStack/ElasticStackLogAccess.cpp
```diff
@@ -0,0 +1,502 @@
...
+    if (retrievedDocument.status_code != 200)
+        throw makeStringExceptionV(-1, "ElasticSearch request failed: %s", retrievedDocument.status_line.c_str());
+
+    DBGLOG("Retrieved ES JSON DOC: %s", retrievedDocument.text.c_str());
+
+    Owned<IPropertyTree> tree = createPTreeFromJSONString(retrievedDocument.text.c_str());
+    if (!tree)
+        throw makeStringExceptionV(-1, "%s: Could not parse ElasticSearch query response", COMPONENT_NAME);
+
+    if (tree->hasProp("timed_out") && tree->getPropBool("timed_out"))
```
ghalliday, line 156: note: the "tree->hasProp("timed_out")" can be removed from this line and line 158 and have the same effect.  Even clearer if a default of false/0 is passed to the getProp function.

system/logaccess/ElasticStack/ElasticStackLogAccess.cpp
```diff
@@ -0,0 +1,502 @@
...
+        LOG(MCuserProgress,"ES Log Access: failed _shards reported");
+
+    DBGLOG("ES Log Access: hit count: '%d'", tree->getPropInt("hits/total/value"));
+
+    Owned<IPropertyTreeIterator> iter = tree->getElements("hits/hits/fields");
+    switch (format)
+    {
+        case LOGACCESS_LOGFORMAT_xml:
+        {
+            returnbuf.set("<lines>");
```
ghalliday, line 168: > Acknowledged, but I explicitly set the buffer. It feels cleaner, more well defined.
The argument against this is that the general pattern for the rest of the system is to append, so anyone calling this function may assume that it will append rather than replace.

system/logaccess/ElasticStack/ElasticStackLogAccess.cpp
```diff
@@ -0,0 +1,502 @@
...
+            StringBuffer hitchildxml;
+
+            ForEach(*iter)
+            {
+                IPropertyTree & cur = iter->query();
+                toXML(&cur,hitchildxml.clear());
```
ghalliday, line 174: note: related to comment above, this could be coded as
```
returnbuf.append("<line>");
toXML(&cur,returnbug);
returnbuf.append("</line>");
```
avoiding the temporary variable.

ghalliday (commented):
@rpastrana A few minor extra comments/responses.  Many of the style issues cleaned up.
I'm not convinced that that the filter classes provide enough benefit, but that isn't enough reason to change it at this point.
My main concern with the PR at the moment is that neither the esp service, nor the elastic search implementation actually address the only use-case in the jira:
> fetch logging for all or some components, for this jobid.
Which suggests more thought should have been put into how this was going to be used.

The other comment in the jira was
> The 1st implementation may be a simple one based on physical files, that have been persisted to a known mounted volume
Is this implementation sufficient to be used for all the existing log file searching (even if it isn't implemented).  How does this change support that functionality on a cloud system?

(Note as part of code review discussion: These should really be resolved before the code review)

system/jlib/jlog.hpp
rpastrana: It's used for queries not targeting a specific column (jobid, component, etc). Considered labeling it "all", "none" but wildcard seemed most appropriate, please provide a suggestion if you think this needs to be clarified. @ghalliday

system/logaccess/ElasticStack/ElasticStackLogAccess.hpp
rpastrana: These functions are useful for anybody using this library independently of the Ilogaccess interface making them private ensures they are not used (sans further changes). How much improvement do you see by forcing the library to be completely coupled to the Ilogaccess interface? @ghalliday

system/logaccess/ElasticStack/ElasticStackLogAccess.cpp
rpastrana: those general patterns need to be explicitly documented and periodically assessed. begrudgingly changing simply for the sake of this PR's prosperity.

system/logaccess/ElasticStack/ElasticStackLogAccess.cpp
rpastrana: but the trade-off is an extra call to the StringBuffer's append per iteration... can this cause extra resizes?

rpastrana:
@ghalliday please see latest commit and responses.

rpastrana:
> @rpastrana A few minor extra comments/responses. Many of the style issues cleaned up. I'm not convinced that that the filter classes provide enough benefit, but that isn't enough reason to change it at this point. My main concern with the PR at the moment is that neither the esp service, nor the elastic search implementation actually address the only use-case in the jira:
> 
> > fetch logging for all or some components, for this jobid.
> > Which suggests more thought should have been put into how this was going to be used.
> 
> The other comment in the jira was
> 
> > The 1st implementation may be a simple one based on physical files, that have been persisted to a known mounted volume
> > Is this implementation sufficient to be used for all the existing log file searching (even if it isn't implemented).  How does this change support that functionality on a cloud system?
> 
> (Note as part of code review discussion: These should really be resolved before the code review)
Regarding the first use case, I thought we discussed with @jakesmith offline and we agreed fetching all entries with associated w/ a jobid was good enough. Added this Jira: https://track.hpccsystems.com/browse/HPCC-26876 change should be elasticstack driver specific, no need to hold up this pr. By the way we should ensure the appropriate components are actually logging jobids. 

There definitely is no longer a need to fetch from log files (based on offline discussion w/ Jake). 
@ghalliday FYI.

system/jlib/jlog.hpp
```diff
@@ -1343,4 +1345,345 @@ interface IComponentLogFileCreator : extends IInterface
...
+                  searchColumn: "hpcc.log.class"
+            )!!";
+            Owned<IPropertyTree> testTree = createPTreeFromYAMLString(simulatedGlobalYaml, ipt_none, ptr_ignoreWhiteSpace, nullptr);
+            logAccessPluginConfig.setown(testTree->getPropTree("global/logAccess"));
+        }
+#endif
+        if (!logAccessPluginConfig)
+            throw makeStringException(-1, "RemoteLogAccessLoader: logaccess configuration not available!");
+
+        static constexpr const char * methodName = "loadLogAccessPlugin";
```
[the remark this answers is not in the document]
ghalliday: note: I've just discovered that __func__ is defined by the c++11 standard as the name of the function.  It would be cleaner to use that.  I may start using it myself...

system/logaccess/ElasticStack/ElasticStackLogAccess.hpp
```diff
@@ -0,0 +1,120 @@
...
+using namespace elasticlient;
+
+class ELASTICSTACKLOGACCESS_API ElasticStackLogAccess : public CInterfaceOf<IRemoteLogAccess>
+{
+private:
+    const char * type = "elasticstack";
+
+    Owned<IPropertyTree> m_pluginCfg;
+
+    StringBuffer m_globalIndexSearchPattern;
```
ghalliday, line 48: style: There are a few style differences from the bulk of the code (m_ as a prefix for member variables), compound identifiers that don't use camelCase (e.g. logmessageprefix, reqloglevel)
Since this is an independent chunk of code, and esp shares some similar style that is ok.  Not using camel case does reduce readability.

system/logaccess/ElasticStack/ElasticStackLogAccess.hpp
```diff
@@ -0,0 +1,120 @@
...
+
+#ifdef _CONTAINERIZED
+    //In containerized world, most likely Elastic Search host is their default k8s hostname
+    static constexpr const char * DEFAULT_ES_HOST = "elasticsearch-master";
+#else
+    //In baremetal, localhost is good guess as any
+    static constexpr const char * DEFAULT_ES_HOST = "localhost";
+#endif
+
+    static constexpr const char * DEFAULT_ES_PROTOCOL = "http";
```
ghalliday, line 86: minor: None of these constants look like part of the public interface, or even that they need to be in this file.  I would vote for constants at the head of the cpp file.  No need to change.

ghalliday (commented):
@rpastrana A few final comments.  My overall comments still stand - that I'm not sure that this solves the real life use cases, and various issues should have been discussed much earlier , before the PR.  The JIRA should be updated if the requirements change.  (It is a general comment, and I will add to the code review doc.)  The design of the esp service not supporting the original use case is concerning.

I don't think you have responded to the question about how this fits in with the existing functions to access the log files.  What we want to avoid is one api for accessing log files in bare-metal and a different one for accessing them in K8s.  What is the future plan to avoid this?

However, if @jakesmith is happy with the functionality I will merge it (with final comments addressed), but with reservations about the esp api.

rpastrana:
> @rpastrana A few final comments. My overall comments still stand - that I'm not sure that this solves the real life use cases, and various issues should have been discussed much earlier , before the PR. The JIRA should be updated if the requirements change. (It is a general comment, and I will add to the code review doc.) The design of the esp service not supporting the original use case is concerning.
> 
> I don't think you have responded to the question about how this fits in with the existing functions to access the log files. What we want to avoid is one api for accessing log files in bare-metal and a different one for accessing them in K8s. What is the future plan to avoid this?
> 
> However, if @jakesmith is happy with the functionality I will merge it (with final comments addressed), but with reservations about the esp api.

@ghalliday - agreed. The scope of the original requirement shifted as far back as before work started without updating the jira (this happens often and the code review doc should cover this situation - likely a formal requirements and design document for new features?).

I mentioned (likely before you were involved) the ESP service was out of scope, and intended for presentation layer consumption. In hind sight, it would have been added as an subsequent PR. Anyway, WUCreateAndDownloadZAPInfo, and/or WU helpers are not expected to utilize the ESP service, new functionality based on the ILogAccess will be required.

This functionality proposed is not intended to supplant the existing log retrieval mechanisms on bare metal and now I see the root of some of your prior questions/concerns. If we decide to move in that direction, a new log-access plugin would need to be created (likely requiring new configuration), and consumers of the legacy log files would need to utilize the IlogAccess framework.

Setting up a meeting to ensure we're all on the same page.

jakesmith:
>Anyway, WUCreateAndDownloadZAPInfo, and/or WU helpers are not expected to utilize the ESP service, new functionality based on the ILogAccess will be required.

I think the key issue is, not that it move over to it now, or that the bare-metal case be implemented,
but that the abstract new log access interface provides the functionality so that the existing logging use cases (zap, helpers) could be moved over to it, without having to change the interfaces.

ghalliday (approved):
@rpastrana please squash.
Note there are still some non-resolved comments (e.g. non use of camel case) which are borderline for needing to be fixed. In the interests of moving forward (and because it is a separate project) I will merge though.

rpastrana:
@ghalliday ready for merge

PR #15587, opened 2021-11-15
HPCC-26485 ecl command support for query-xref and query superfile list

afishbeck:
Expose roxie "<control:getQueryXrefInfo>" as a new ecl command line
feature "ecl roxie xref".

ecl roxie xref <cluster> --queryids=<comma seperated query list>

Also update the "ecl queries list" feature to show superfiles in
addition to logical files in use.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/services/ws_smc/ws_smcService.cpp
+++ b/esp/services/ws_smc/ws_smcService.cpp
@@ -2200,6 +2200,82 @@ inline const char *controlCmdMessage(int cmd)
     return NULL;
 }
 
+bool CWsSMCEx::onRoxieXrefCmd(IEspContext &context, IEspRoxieXrefCmdRequest &req, IEspRoxieXrefCmdResponse &resp)
+{
+    context.ensureFeatureAccess(ROXIE_CONTROL_URL, SecAccess_Full, ECLWATCH_SMC_ACCESS_DENIED, SMC_ACCESS_DENIED);
+
+    StringBuffer controlReq;
+    if (0==req.getQueryIds().length())
+        controlReq.append("<control:getQueryXrefInfo/>");
+    else
+    {
+        controlReq.append("<control:getQueryXrefInfo>");
+        ForEachItemIn(i, req.getQueryIds())
+        {
+            const char *id = req.getQueryIds().item(i);
+            if (!isEmptyString(id))
+                controlReq.appendf("<Query id='%s'/>", id);
+        }
+        controlReq.append("</control:getQueryXrefInfo>");
+    }
+
+#ifndef _CONTAINERIZED
+    const char *process = req.getRoxieCluster();
+    if (isEmptyString(process))
+        throw makeStringException(ECLWATCH_MISSING_PARAMS, "Process cluster not specified.");
+
+    SocketEndpointArray addrs;
+    getRoxieProcessServers(process, addrs);
+    if (!addrs.length())
+        throw makeStringException(ECLWATCH_CANNOT_GET_ENV_INFO, "Process cluster not found.");
+    Owned<IPropertyTree> controlResp;
+    if (req.getCheckAllNodes())
+        controlResp.setown(sendRoxieControlAllNodes(addrs.item(0), controlReq, true, req.getWait()));
+    else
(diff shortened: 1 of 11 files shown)


richardkchapman (approved):

esp/services/ws_smc/ws_smcService.cpp
```diff
@@ -2200,6 +2200,82 @@ inline const char *controlCmdMessage(int cmd)
     return NULL;
 }
 
+bool CWsSMCEx::onRoxieXrefCmd(IEspContext &context, IEspRoxieXrefCmdRequest &req, IEspRoxieXrefCmdResponse &resp)
+{
+    context.ensureFeatureAccess(ROXIE_CONTROL_URL, SecAccess_Full, ECLWATCH_SMC_ACCESS_DENIED, SMC_ACCESS_DENIED);
+
+    StringBuffer controlReq;
```
ghalliday, line 2207: design: @afishbeck does this solve the issue outline in the jira.  

> "In addition can we also have the '<control:getQueryXrefInfo/>'  "testsocket" functionality in ecl command line so we don't have to rely on roxie being up to get the queries/superkeys list"

I thought the intent was to retrieve this information from the data stored in dali, instead of going out to the roxie to retrieve it.

esp/services/ws_smc/ws_smcService.cpp
```diff
@@ -2200,6 +2200,82 @@ inline const char *controlCmdMessage(int cmd)
...
+        {
+            const char *id = req.getQueryIds().item(i);
+            if (!isEmptyString(id))
+                controlReq.appendf("<Query id='%s'/>", id);
+        }
+        controlReq.append("</control:getQueryXrefInfo>");
+    }
+
+#ifndef _CONTAINERIZED
+    const char *process = req.getRoxieCluster();
```
ghalliday, line 2223: minor: This code shares quite a few similarities with other code in this file (onRoxieControlCmd()) and ws_workunitQuerySets.cpp.  I suspect it could be simplified by combining them.

ghalliday (commented):
@afishbeck see comment - is this providing what the user requested?

esp/services/ws_smc/ws_smcService.cpp
afishbeck: Right.  After discussion with him offline I added the xref passthrough for cases where he can't get to the roxie via testsocket.
"ecl queries files" is meant to be the function that parallels ```<control:getQueryXrefInfo/>``` without going to roxie, but there are still a few enhancements that could be made that would make it more directly comparable.  

For example, 
1. right now it only works one query at a time.. but xref supports "all queries".  
2. It's quite a bit more work, but we could start keep track and showing the full nesting of superfiles.  I'm not sure what  ```<control:getQueryXrefInfo/>``` does there.

Perhaps I should open another JIRA for additional enhancements to "ecl queries files"?  I think these changes will help him right away.

ghalliday (approved):
ok.  Seems a reasonable stepping stone.

venkatan-github:
Thank you for working on this, just wanted to make sure that this fix will also return keys that are declared as optional because testsocket currently doesn't return keys that are declared optional

afishbeck:
@venkatan-github the "ecl roxie xref" command works exactly like testsocket "control:getQueryXrefInfo" including only working if roxie is up and the query is not suspended.

We should focus on getting "ecl queries files" to do what you need.  I believe it already includes the optional files... can you try it with the version you have now?  With this change here it will start listing superfiles as well.  Right now it is for a single query, we could enhance it to show all queries if that is something you really need.

"ecl queries files" does not require roxie to be up, and works even if the query is suspended.

venkatan-github:
Thank you Tony,

ecl queries files is extremely slow, each call takes about 30 secs to 1 min to return. If we have an environment that has 600 queries it doesn't look good. I have opened a ticket https://track.hpccsystems.com/browse/HPCC-26485

Anyway we can get this is in the near future releases?

PR #15595, opened 2021-11-19
HPCC-26859 Move udp simulation tests out of unittests framework

richardkchapman:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/roxie/udplib/udptrs.cpp
+++ b/roxie/udplib/udptrs.cpp
@@ -40,10 +40,6 @@ unsigned udpMaxRetryTimedoutReqs = 0; // 0 means off (keep retrying forever)
 unsigned udpRequestToSendTimeout = 0; // value in milliseconds - 0 means calculate from query timeouts
 unsigned udpRequestToSendAckTimeout = 10; // value in milliseconds
 
-#ifdef _DEBUG
-//#define TEST_DROPPED_PACKETS
-#endif
-
 using roxiemem::DataBuffer;
 /*
  *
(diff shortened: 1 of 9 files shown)


roxie/udplib/udptrs.cpp
```diff
@@ -41,7 +41,7 @@ unsigned udpRequestToSendTimeout = 0; // value in milliseconds - 0 means calcula
 unsigned udpRequestToSendAckTimeout = 10; // value in milliseconds
 
 #ifdef _DEBUG
-//#define TEST_DROPPED_PACKETS
+#define TEST_DROPPED_PACKETS
```
ghalliday, line 44: Did you mean to leave this defined?

ghalliday (commented):
@richardkchapman one question

system/jlib/jptree.cpp
```diff
@@ -8638,7 +8638,7 @@ Owned<IPropertyTree> getGlobalConfigSP()
     return getGlobalConfig();
 }
 
-jlib_decl IPropertyTree * loadArgsIntoConfiguration(IPropertyTree *config, const char * * argv, std::initializer_list<const char *> ignoreOptions)
+jlib_decl IPropertyTree * loadArgsIntoConfiguration(IPropertyTree *config, const char * * argv, std::initializer_list<const std::string> ignoreOptions)
```
ghalliday, line 8641: This change doesn't look like it is related - and I think it will mean previously constexpr arrays now need to dynamically initialise the object array.

ghalliday (commented):
@richardkchapman one comment

system/jlib/jptree.cpp
richardkchapman: It's fixing a bug I found while testing, that the suppression of certain flags when using this function does not work as the comparison is done using pointers rather than strcmp. There's only one place that passes in an initializer array, and it compiles on on my machine/branch

ghalliday: It should compile, just slightly less efficient.  Given the context it is called in, and the number of times it is used, that is probably irrelevant.

ghalliday (approved):

PR #15596, opened 2021-11-19
HPCC-26588 Document configuration of containerized systems

g-pan:
## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
+++ b/docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
@@ -1,0 +1,1269 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE section PUBLIC "-//OASIS//DTD DocBook XML V4.5//EN"
+"http://www.oasis-open.org/docbook/xml/4.5/docbookx.dtd">
+<chapter id="ContainerConfigurationcHAPTER">
+  <title>Containerized Configuration</title>
+
+  <para>This chapter describes the configuration of HPCC Systems for a
+  Kubernetes Containerized deployment. The following sections detail how
+  configurations are supplied to helm charts, how to find out what options are
+  available and some details of the configuration file structure. Subsequent
+  sections will also provide a brief walk through of some of the contents of
+  the default <emphasis>values.yaml</emphasis> file used in configuring the
+  HPCC Systems for a containerized deployment.</para>
+
+  <sect1 id="Intro_Containerized_Environments">
+    <title>Introduction to the Container Environment</title>
+
+    <para>One of the ideas behind our move to the cloud was to try and
+    simplify the system configuration while also delivering a solution
+    flexible enough to meet the demands of our community while taking
+    advantage of container features without sacrificing performance.</para>
+
+    <para>The entire HPCC Systems configuration in the container space, is
+    governed by a single file, a <emphasis>values.yaml</emphasis> file, and
+    its associated schema file.</para>
+
+    <sect2 id="WhatIsValues.Yaml">
+      <title>What is <emphasis>values.yaml</emphasis> and how is it
+      used?</title>
+
+      <para>The <emphasis>values.yaml</emphasis> file is the delivered
+      configuration values for a Helm chart. The
+      <emphasis>values.yaml</emphasis> file is used by the Helm chart to
+      control how HPCC Systems is deployed to the cloud. This values file is
+      one file used to configure and get an HPCC Systems instance up and
+      running on Kubernetes. The <emphasis>values.yaml</emphasis> file defines
+      everything that happens to configure and/or define your system for a
+      containerized deployment. You should use the values file provided as a
+      basis for modeling the specific customizations for your deployment
+      specific to your requirements.</para>
+
+      <para>The HPCC Systems<emphasis> values.yaml</emphasis> file can be
+      found in the HPCC Systems github repository. To use the HPCC Systems
+      Helm chart, first add the hpcc chart repository using Helm, then access
+      the Helm chart values from the charts in that repository.</para>
+
+      <para>For example, when you add the "hpcc" repository, as recommended
+      prior to installing the Helm chart with the following command:</para>
+
+      <programlisting>helm repo add hpcc https://hpcc-systems.github.io/helm-chart
+</programlisting>
+
+      <para>You can now view the HPCC Systems delivered charts and see the
+      values there by issuing:</para>
+
+      <programlisting>helm show values hpcc/hpcc</programlisting>
+
+      <para>You can capture the output of this command, look at how the
+      defaults are configured and use it as a basis for your
+      customization.</para>
+    </sect2>
+
+    <sect2 id="Values-SchemaJSONFile">
+      <title>The values-schema.json</title>
+
+      <para>The <emphasis>values-schema.json</emphasis> is a JSON file that
+      declares what is valid and what is not within the sum total of the
+      merged values that are passed into Helm at install time. It defines what
+      values are allowed, and validates the values file against them. All the
+      core items are declared in the schema file, while the default
+      <emphasis>values.yaml</emphasis> file also contains comments on the most
+      important elements. If you wanted to know what options are available for
+      any particular component then the schema is a good place to
+      start.</para>
+
+      <para>The schema file typically contains (for a property) a name and a
+      description. It will often include details of the type, and the items it
+      can contain if it is a list or dictionary. For instance: </para>
+
+      <programlisting>    "roxie": { 
+      "description": "roxie process",
+      "type": "array"
+      "items": { "$ref": "#/definitions/roxie" }
+    },</programlisting>
+
+      <para>Each plane, in the schema file has a list of properties generally
+      containing a prefix (path), a subpath (subpath), and additional
+      properties. For example, for a storage plane the schema file has a list
+      of properties including the prefix. The "planes" in this case are a
+      reference ($ref) to another section of the schema. The schema file
+      should be complete, and contain everything required including
+      descriptions which should be relatively self-explanatory.</para>
+
+      <programlisting>    "storage": {
+      "type": "object",
+      "properties": {
+        "hostGroups": {
+          "$ref": "#/definitions/hostGroups"
+        },
+        "planes": {
+          "$ref": "#/definitions/storagePlanes"
+        }
+      },
+      "additionalProperties": false
+</programlisting>
+
+      <para>Note the <emphasis>additionalProperties</emphasis> value typically
+      at the end of each section in the schema. It specifies whether the
+      values allow for additional properties or not. If that
+      <emphasis>additionalProperties</emphasis> value is present and set to
+      false, then no other properties are allowed and the property list is
+      complete.</para>
+
+      <para>In working with the HPCC Systems <emphasis>values.yaml</emphasis>,
+      the values file must validate against this schema. If there is a value
+      that is not allowed as defined in the schema file it will not start and
+      instead generate an ERROR.</para>
+    </sect2>
+  </sect1>
+
+  <sect1 id="TheValuesYaml_FileAndRelated" role="nobrk">
+    <title>The HPCC Components in the <emphasis>values.yaml</emphasis>
+    File</title>
+
+    <para>The HPCC Systems Helm charts all ship with stock/default values.
+    These Helm charts have a set of default values ideally to be used as a
+    guide in configuring your deployment. Generally, every HPCC Systems
+    component is a list. That list defines the properties for each instance of
+    the component.</para>
+
+    <para>This section will provide additional details and any noteworthy
+    insight for the HPCC Systems components defined in the
+    <emphasis>values.yaml</emphasis> file.</para>
+
+    <sect2 id="YAMLHPCC_Components">
+      <title>The HPCC Components</title>
+
+      <para>One of the key differences between the bare metal and
+      container/cloud is that in bare metal storage is directly tied to the
+      Thor or the Thor worker nodes, and the Roxie worker nodes, or even in
+      the case of the ECLCC Server the DLLs. In containers these are
+      completely separate and anything having to do with files is defined in
+      the <emphasis>values.yaml</emphasis>.</para>
+
+      <para>In containers component instances run dynamically. For instance,
+      if you have configured your system to use a 50-way Thor, then a 50-way
+      Thor will be spawned when a job is queued to it. When that job is
+      finished that Thor instance will disappear. This is the same pattern for
+      the other components as well.</para>
+
+      <para>Every component should have a resources entry, in the delivered
+      <emphasis>values.yaml</emphasis> file the resources are present but
+      commented out as indicated here.</para>
+
+      <para><programlisting>  #resources:
+  #  cpu: "1"
+  #  memory: "4G"
+</programlisting>The stock values file will work and allow you to stand up a
+      functional system, however you should define the component resources in
+      a manner that corresponds best to your operational strategy.</para>
+
+      <sect3 id="DALI_ValueYAML">
+        <title>Dali</title>
+
+        <para>When configuring Dali, which also has a resources section, it is
+        going to want plenty of memory and a good amount of CPU as well. It is
+        very important to define these carefully. Otherwise Kubernetes could
+        assign all the pods to the same virtual machine and components
+        fighting for memory will crush them. Therefore more memory assigned
+        the better. If you define these wrong and some process starts using
+        too much memory, Kubernetes will kill the pod if it is using much more
+        memory than what it was configured for.</para>
+      </sect3>
+
+      <sect3 id="DAFLESRV_DFURVR_YMLSECT">
+        <title>Components: dafilesvrs, dfuserver</title>
+
+        <para>The HPCC Systems components of dafilesvrs, eclccservers,
+        dfuserver, are declared as a lists in the yaml, as is the ECL
+        Agent.</para>
+
+        <para>Consider the dfuserver which is in the delivered HPCC Systems
+        <emphasis>values.yaml</emphasis> as:</para>
+
+        <programlisting>dfuserver:
+- name: dfuserver
+  maxJobs: 1</programlisting>
+
+        <para>If you were to add a mydfuserver as follows</para>
+
+        <para><programlisting>dfuserver:
+- name: dfuserver
+  maxJobs: 1
+- name: mydfuserver
+  maxJobs: 1
+</programlisting>In this scenario you would have another item here named
+        mydfuserver and it would show up in ECLWatch and you can submit items
+        to that.</para>
+
+        <para>If you wanted to add another dfuserver, you can add that to the
+        list similarly. You can likewise instantiate other components by
+        adding them to their respective lists.</para>
+      </sect3>
+
+      <sect3 id="VALYml_ECLCCServer">
+        <title>ECL Agent and ECLCC Server</title>
+
+        <para>Values of note for the ECL Agent and ECLCC Server.</para>
+
+        <para><emphasis role="bold">useChildProcess</emphasis> -- As defined
+        in the schema, launches each workunit compile as a child process
+        rather than in its own container. When you submit a job or query to
+        compile it gets queued and processed, with this option set to true it
+        will spawn a child process utilizing almost no additional overhead in
+        starting. Ideal for sending many small jobs to compile. However,
+        because each compile job is no longer executed as an independent pod
+        with it's own resource specifications, but instead runs as a child
+        process within the ECLCC Server pod itself, the ECLCC Server pod must
+        be defined with adequate resources for itself (minimal for listening
+        to the queue etc.) and all the jobs it may have to run in
+        parallel.</para>
+
+        <para>For example, imagine <emphasis>maxJobs</emphasis> is set to 4,
+        and 4 large queries are queued rapidly, that will mean 4 child
+        processes are launched each consuming cpu and memory within the ECLCC
+        Server pod. With the component configured with
+        <emphasis>useChildProcesses</emphasis> set to true, each job will run
+        in the same pod (up to the value of <emphasis>maxJobs</emphasis> in
+        parallel). Therefore with <emphasis>useChildProcesses</emphasis>
+        enabled, the component resources must be defined such that the pod has
+        enough resources to handle the resource demands of all the those jobs
+        to be able to run in parallel.</para>
+
+        <para>With useChildProcess enabled it could be rather expensive in
+        most cloud pricing models, and rather wasteful if there aren't any
+        jobs running. Instead you can set this
+        <emphasis>useChildprocess</emphasis> to false (the default) to start a
+        pod to compile each query with the the required memory for the job
+        which will be disposed of when done. Now this model also has
+        overheard, perhaps 20 seconds to a minute to spawn the Kubernetes
+        cluster to process the job. Which may not be ideal for an environment
+        which is sending several small jobs, but rather larger jobs which
+        would minimize the effect of the overhead in starting the Kubernetes
+        cluster.</para>
+
+        <para>Setting <emphasis>useChildProcess</emphasis> to false better
+        allows for the possibility of dynamic scaling. For jobs which would
+        take a long while to compile, the extra (start up) overhead is
+        minimal, and that would be the ideal case to have the
+        <emphasis>useChildProcess</emphasis> as false. Setting
+        <emphasis>useChildProcess</emphasis> to false only allows 1 pod per
+        compile, though there is an attribute for putting a time limit on that
+        compilation.</para>
+
+        <para><emphasis role="bold">ChildProcessTimeLimit</emphasis> is the
+        Time limit (in seconds) for child process compilation before aborting
+        and using a separate container, when the
+        <emphasis>useChildProcesses</emphasis> is false.</para>
+
+        <para><emphasis role="bold">maxActive</emphasis>: The maximum number
+        of jobs that can be run in parallel. Again use caution because each
+        job will need enough memory to run. For instance, if
+        <emphasis>maxActive</emphasis> is set to 2000, you could submit a very
+        big job and in that case spawn some 2000 jobs using a considerable
+        amount of resources, which could potentially run up a rather expensive
+        compilation bill, again depending on your cloud provider and your
+        billing plan.</para>
+      </sect3>
+
+      <sect3 id="ValYML_Sasha">
+        <title>Sasha</title>
+
+        <para>The configuration for Sasha is an outlier as it is a dictionary
+        type structure and not a list. You can't have more than one archiver
+        or dfuwu-archiver as that is a value limitation, you can choose to
+        either have the service or not (set the 'disabled' value to
+        true).</para>
+      </sect3>
+
+      <sect3 id="ValYML_Thor">
+        <title>Thor</title>
+
+        <para>Thor instances run dynamically, as do the other components in
+        containers. The configuration for Thor also consists of a list of Thor
+        instances. Each instance dynamically spawns a collection of pods
+        (manager + N workers) when jobs are queued to it. When idle there are
+        no worker (or manager) pods running.</para>
+
+        <para>If you wanted a 50-way Thor you set the number of workers, the
+        <emphasis role="bold">numWorkers</emphasis> value to 50 and you would
+        have a 50-way Thor. As indicated in the following example:</para>
+
+        <para><programlisting>thor:
+- name: thor
+  prefix: thor
+  numWorkers: 50</programlisting></para>
+
+        <para>In doing so, ideally you should rename the resource to something
+        which clearly describes it, such as <emphasis>thor_50</emphasis> as in
+        the following example.</para>
+
+        <para><programlisting>-name: thor_50</programlisting></para>
+
+        <para>Updating the <emphasis>numWorkers</emphasis> value will restart
+        the Thor agent listening to the queue, causing all new jobs to use the
+        new configuration.</para>
+
+        <para><emphasis role="bold">maxJobs</emphasis>: controls the number of
+        jobs, specifically <emphasis>maxJobs</emphasis> sets the maximum
+        number of jobs.</para>
+
+        <para><emphasis role="bold">maxGraphs</emphasis>: limits the maximum
+        amount of graphs. It generally makes sense to keep this value below or
+        at the same number as <emphasis>maxJobs</emphasis>, since not all jobs
+        submit graphs and when they do the Thor jobs are not executing graphs
+        all the time. If there are more than 2 submitted (Thor) graphs, the
+        second would be blocked until the next Thor instance becomes
+        available.</para>
+
+        <para>The idea here is that jobs may spend significant amount of time
+        outside of graphs, such as waiting on a workflow state (outside of the
+        Thor engine itself), blocked on a persist, or updating super files,
+        etc. Then it makes sense for Thor to have a higher limit of concurrent
+        jobs (<emphasis>maxJobs</emphasis>) than graphs
+        (<emphasis>maxGraphs</emphasis> / Thor instances). Since Thor
+        instances (graphs) are relatively expensive (lots of pods/higher
+        resource use), while workflow pods (jobs) are comparatively
+        cheap.</para>
+
+        <para>Thus, the delivered (example) chart values defines
+        <emphasis>maxJobs</emphasis> to be greater than
+        <emphasis>maxGraphs</emphasis>. Jobs queued to a Thor aren't always
+        running graphs. Therefore it can make sense to have more of these
+        jobs, which are not consuming a large Thor and all its resources, but
+        restrict the max number of Thor instances running.</para>
+
+        <para>Thor has 3 components (that correspond to the resource
+        sections).</para>
+
+        <orderedlist>
+          <listitem>
+            <para>Workflow</para>
+          </listitem>
+
+          <listitem>
+            <para>Manager</para>
+          </listitem>
+
+          <listitem>
+            <para>Workers</para>
+          </listitem>
+        </orderedlist>
+
+        <para>The Manager and Workers are launched together and consume quite
+        a bit of resoures (and nodes) typically. While the Workflow is
+        inexpensive and usually doesn't require as many resources. You might
+        expect in a Kubernetes world, many of them would co-exist on the same
+        node (and therefore be inexpensive). So it makes sense for
+        <emphasis>maxJobs</emphasis> to be higher, and
+        <emphasis>maxGraphs</emphasis> to be lower</para>
+
+        <para>In Kubernetes, jobs run independently in their own pods. While
+        in bare metal we can have jobs that could effect other jobs because
+        they are running in the same process space.</para>
+      </sect3>
+    </sect2>
+  </sect1>
+
+  <sect1 id="Delivered_HPCC_ValuesYaml">
+    <title>The HPCC Systems <emphasis>values.yaml</emphasis> file</title>
+
+    <para>The delivered HPCC systems <emphasis>values.yaml</emphasis> file is
+    more of an example providing a basic type configuration which should be
+    customized for your specific needs. One of the main ideas behind the
+    values file is to be able to relatively easily customize it to your
+    specific scenario. The delivered chart is set up to be sensible enough to
+    understand, while also allowing for relatively easy customization to
+    configure a system to your specific requirements. This section will take a
+    closer look at some aspects of the delivered
+    <emphasis>values.yaml</emphasis>.</para>
+
+    <para>The delivered HPCC Systems Values file primarily consists of the
+    following areas:</para>
+
+    <itemizedlist>
+      <listitem>
+        <para>global</para>
+      </listitem>
+
+      <listitem>
+        <para>storage</para>
+      </listitem>
+
+      <listitem>
+        <para>data planes</para>
+      </listitem>
+
+      <listitem>
+        <para>certificates</para>
+      </listitem>
+
+      <listitem>
+        <para>security</para>
+      </listitem>
+
+      <listitem>
+        <para>secrets</para>
+      </listitem>
+
+      <listitem>
+        <para>components</para>
+      </listitem>
+    </itemizedlist>
+
+    <para>The subsequent sections will examine some of these more closely and
+    why each of them is there.</para>
+
+    <sect2 id="ValYAML_STorage">
+      <title>Storage</title>
+
+      <para>Containerized Storage is another key concept that differs from
+      bare metal. There are a few differences between container and bare metal
+      storage. The Storage section is fairly well defined between the schema
+      file, and the <emphasis>values.yaml</emphasis>. A good approach towards
+      storage is to clearly understand your storage needs, and to outline
+      them, and once you have that basic structure in mind the schema can help
+      to fill in the details. The schema should have a decent description for
+      each attribute. All storage should be defined via planes. There is a
+      relevant comment in the <emphasis>values.yaml</emphasis> further
+      describing storage.</para>
+
+      <programlisting>## storage:
+##
+## 1. If an engine component has the dataPlane property set, 
+#       then that plane will be the default data location for that component.
+## 2. If there is a plane definition with a category of "data" 
+#       then the first matching plane will be the default data location
+##
+## If a data plane contains the storageClass property then an implicit pvc 
+#       will be created for that data plane.
+##
+## If plane.pvc is defined, a Persistent Volume Claim must exist with that name, 
+#       storageClass and storageSize are not used.
+##
+## If plane.storageClass is defined, storageClassName: &lt;storageClass&gt;
+## If set to "-", storageClassName: "", which disables dynamic provisioning
+## If set to "", choosing the default provisioner.  
+#       (gp2 on AWS, standard on GKE, AWS &amp; OpenStack)
+##
+## plane.forcePermissions=true is required by some types of provisioned
+## storage, where the mounted filing system has insufficient permissions to be
+## read by the hpcc pods. Examples include using hostpath storage (e.g. on
+## minikube and docker for desktop), or using NFS mounted storage.
+</programlisting>
+
+      <para>There are different categories of storage, for an HPCC Systems
+      deployment you must have at a minimum a dali category, a dll category,
+      and at least 1 data category. These types are generally applicable for
+      every configuration in addition to other optional categories of
+      data.</para>
+
+      <para>All storage should be in a storage plane definition. This is best
+      described in the comment in the storage definition in the values
+      file.</para>
+
+      <programlisting>planes:
+  #   name: &lt;required&gt;
+  #   prefix: &lt;path&gt;                        # Root directory for accessing the plane 
+                                            # (if pvc defined), 
+  #                                         # or url to access plane.
+  #   category: data|dali|lz|dll|spill|temp # What category of data is stored on this plane?
+  #
+  # For dynamic pvc creation:
+  #   storageClass: ''
+  #   storageSize: 1Gi
+  #
+  # For persistent storage:
+  #   pvc: &lt;name&gt;                           # The name of the persistant volume claim
+  #   forcePermissions: false
+  #   hosts: [ &lt;host-list ]                 # Inline list of hosts
+  #   hostGroup: &lt;name&gt;                     # Name of the host group for bare metal 
+  #                                         # must match the name of the storage plane..
+  #
+  # Other options:
+  #   subPath: &lt;relative-path&gt;              # Optional sub directory within &lt;prefix&gt; 
+  #                                         # to use as the root directory
+  #   numDevices: 1                         # number of devices that are part of the plane
+  #   secret: &lt;secret-id&gt;                   # what secret is required to access the files.  
+  #                                         # This could optionally become a list if required 
+                                            # (or add secrets:).
+
+  #   defaultSprayParts: 4                  # The number of partitions created when spraying 
+                                            # (default: 1)
+
+  #   cost:                                 # The storage cost
+  #     storageAtRest: 0.0135               # Storage at rest cost: cost per GiB/month</programlisting>
+
+      <para>Each plane has 3 required fields: The name, the category and the
+      prefix.</para>
+
+      <para>When the system is installed,using the stock supplied values it
+      will create a storage volume which has 1 GB capacity via the following
+      properties.</para>
+
+      <para>For example:</para>
+
+      <programlisting>- name: dali
+  storageClass: ""
+  storageSize: 1Gi
+  prefix: "/var/lib/HPCCSystems/dalistorage"
+  category: dali
+</programlisting>
+
+      <para>Most commonly the prefix: defines the path within the container
+      where the storage is mounted. The prefix can be a URL for blob storage.
+      All pods will use the (prefix: ) path to access the storage.</para>
+
+      <para>For the above example, when you look at the storage list, the
+      <emphasis>storageSize</emphasis> will create a volume with 1 GB
+      capacity. The prefix will be the path, the category is used to limit
+      access to the data, and to minimize the number of volumes accessible
+      from each component.</para>
+
+      <para>The dynamic storage lists in the <emphasis>values.yaml</emphasis>
+      file are characterized by the storageClass: and storageSize:
+      values.</para>
+
+      <para><emphasis role="bold">storageClass</emphasis>: defines which
+      storage provisioner should be used to allocate the storage. A blank
+      storage class indicates it should use the default cloud providers
+      storage class.</para>
+
+      <para><emphasis role="bold">storageSize</emphasis>: As indicated in the
+      example, defines the capacity of the volume.</para>
+
+      <sect3 id="YAML_StorageCategory">
+        <title>Storage Category</title>
+
+        <para>Storage category is used to indicate the kind of data that is
+        being stored in that location. Different planes are used for the
+        different categories to isolate the different types of data from each
+        other, but also because they often require different performance
+        characteristics. A named plane may only store one category of data.
+        The following sections look at the currently supported categories of
+        data used in our containerized deployment.</para>
+
+        <para><programlisting> category: data|dali|lz|dll|spill|temp  # What category of data is stored on this plane?</programlisting></para>
+
+        <para>The system itself can write out to any data plane. This is how
+        the data category can help to improve performance. For example, if you
+        have an index, Roxie would want rapid access to data, versus other
+        components.</para>
+
+        <para>Some components may use only 1 category, some can use several.
+        The values file can contain more than one storage plane definition for
+        each category. The first storage plane in the list for each category
+        is used as the default location to store that category of data. These
+        categories minimize the exposure of plane data to components that
+        don't need them. For example the ECLCC Server component does not need
+        to know about landing zones, or where Dali stores its data, so it only
+        mounts the plane categories it needs.</para>
+      </sect3>
+
+      <sect3 id="YML_EphemeralStorage">
+        <title>Ephemeral Storage</title>
+
+        <para>Ephemeral storage is allocated when the HPCC Systems cluster is
+        installed and deleted when the chart is uninstalled. This is helpful
+        in keeping cloud costs down but may not be appropriate for your
+        data.</para>
+
+        <para>In your system, you would want to override the delivered stock
+        value(s) with storage appropriate for your specific needs. The
+        supplied values create ephemeral or temporary persistent volumes that
+        get automatically deleted when the chart is uninstalled. You probably
+        want the storage to be persistent. You should customize the storage to
+        a more suitable configuration for your needs.</para>
+      </sect3>
+
+      <sect3 id="YAML_Persist_storage">
+        <title>Persistent Storage</title>
+
+        <para>Kubernetes uses persistent volume claims (pvcs) to provide
+        access to data storage. HPCC Systems supports cloud storage through
+        the cloud provider that can be exposed through these persistent volume
+        claims.</para>
+
+        <para>Persistent Volume Claims can be created by overriding the
+        storage values in the delivered Helm chart. The values in the
+        <emphasis
+        role="boldIt">examples/local/values-localfile.yaml</emphasis> provided
+        override the corresponding entries in the original delivered stock
+        HPCC Systems helm chart. The localfile chart creates persistent
+        storage volumes. You can use the
+        <emphasis>values-localfile.yaml</emphasis> directly (as demonstrated
+        in separate docs/tutorials) or you can use it as a basis for creating
+        your own override chart.</para>
+
+        <para>To define a storage plane that utilizes a PVC, you must decide
+        on where that data will reside. You create the storage directories,
+        with the appropriate names and then you can install the localfiles
+        Helm chart to create the volumes to use the local storage option, such
+        as in the following example:</para>
+
+        <programlisting>helm install mycluster hpcc/hpcc -f examples/local/values-localfile.yaml</programlisting>
+
+        <para>Note: The settings for the PVC's must be ReadWriteMany, except
+        for Dali which can be ReadWriteOnce.</para>
+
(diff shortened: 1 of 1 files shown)


g-pan:
Not intended for Merge. Created to facilitate easier tracking review/revisions.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+        <title>Cluster Disambiguation</title>
+
+        <para>In the past we referred to data and storage as a cluster. We
+        overused the term cluster, and created some confusion around that
+        term. In the container space, it doesn't make any sense to say cluster
+        when really talking about a data plane. For data storage with nodes,
+        such as Thor running on those nodes, we will use the term, plane, as
+        in data plane, or storage plane. Aligning more so with industry
+        standards in the container space. When specifying where a workunit
+        would run we will use the tern queue. Doing so differentiates two
```
g-pan, line 941: term

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+        then wipe it.</para>
+      </sect3>
+
+      <sect3 id="YAML_Lists">
+        <title>Lists</title>
+
+        <para>Lists are groups of elements beginning at the same indentation
+        level starting with a - (a dash and a space). Every element of the
+        list is indented and starts with a dash and a space. Lists can also be
+        nested.</para>
```
ghalliday, line 166: This isn't clear - I don't think they can't directly.  Lists can be lists of dictionaries, which may in turn have list properties.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+
+        <para>An example of a nested list looks like:</para>
```
ghalliday, line 168: An example of a list of dictionaries, with placement.tolerations as a nested list.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+
+        <para><programlisting>placements:
+- pods: ["all"]
+  placement:
+    tolerations:
+    - key: "kubernetes.azure.com/scalesetpriority"
+</programlisting></para>
+
+        <para>A key is a minus sign, which is an entry item in the list, which
```
ghalliday, line 177: no sure of the phrase "a key is a minus sign".

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+      <title>Usage</title>
+
+      <para>The HPCC Systems values.yaml file is used by the Helm chart to
+      control how HPCC Systems is deployed. The values file contains
+      dictionaries and lists, and they can be nested to create more complex
+      structures. The stock HPCC Systems Helm chart is intended as a quick
+      start demonstration installation guide which is not appropriate for
+      practical production usage. You may want to customize your deployment to
+      one which is more suited towards your specific needs. To customize your
+      deployment you could override the stock values in the values.yaml file,
```
ghalliday, line 194: remove word "could"

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+      </sect3>
+    </sect2>
+
+    <sect2 id="YAML_Usage">
+      <title>Usage</title>
+
+      <para>The HPCC Systems values.yaml file is used by the Helm chart to
+      control how HPCC Systems is deployed. The values file contains
+      dictionaries and lists, and they can be nested to create more complex
+      structures. The stock HPCC Systems Helm chart is intended as a quick
```
ghalliday, line 190: the stock values.yaml - not the helm chart.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+
+      <para>As we will see later, components are defined as lists, so any
+      definition of a component in a user values file will replace all
+      instances of the component in the default chart. Other options (for
+      instance configuring the costs for cpu or file access) are implemented
+      as a dictionary, so options can be selectively set in a users values
+      file, and the other options will be retained.</para>
+
+      <sect3 id="merging_AND_Overrides">
+        <title>Merging and Overriding</title>
```
ghalliday, line 250: A chunk of the previous section is about merging as well - should be combined/rationalised.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+      such as the <emphasis>myoverrides.yaml</emphasis> used in the above
+      example and merged using helm. Further information about a customized
+      deployments is covered in other sections, as well as the Kubernetes Helm
+      documentation. Consulting the Helm documentation provides complete
+      detail for every aspect of Helm chart usage, and not only for a few
+      select cases described.</para>
+
+      <itemizedlist>
+        <listitem>
+          <para>To (set a value to) disable something in a values file, it
```
ghalliday, line 216: I don't follow the point this sentence is making.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+        non-changing areas are left untouched. The rule here to keep in mind
+        where multiple yaml files are applied, the later ones will always
+        overwrite the values in the earlier ones. They are merged in in
+        sequence.</para>
+
+        <para>Another point to consider, where there is a global dictionary
+        such as root: and its value redefined in the 2nd file (as a
+        dictionary) it would not be overwritten. There is no “nice way” to
+        replace a dictionary, you cant overwrite a dictionary. You can
+        redefine a dictionary and set it to null, which will effectively wipe
```
ghalliday, line 265: this is a repeat of something mentioned earlier.  Should probably only be once (since it is a detail of helm usage, rather than our HPCC charts).

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+        redefined it would wipe out every definition in the list.</para>
+
+        <para>Another means to wipe out every value in a list is to pass in an
+        empty set denoted by a [ ] such as this example:</para>
+
+        <para><programlisting>bundles: []</programlisting>This would wipe out
+        any properties defined for bundles.</para>
+      </sect3>
+
+      <sect3 id="GenerallyApplicable">
```
ghalliday, line 279: Not sure why this is here - what do name/prefix/service refer to?  Should include examples.  And yaml files merged in order is a Helm feature covered above

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+
+      <para>Therefore to add a new component, for instance another Roxie, you
+      can extend the list of Roxies by either redefining and overriding the
+      list of Roxies, or copying the existing list and adding another entry to
+      it.</para>
+
+      <para>Again, the order the values are merged is the same in which they
+      are issued on the command line.</para>
+
+      <sect3 id="using_the_set_flag">
```
ghalliday, line 355: This is a helm specific detail, so should be under the general helm section.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+    embedded: "allow"
+    pipe:  "allow"
+    extern: "allow"
+    datafile: "allow"</programlisting>In the above examples, storage: and
+        security: are global chart values.</para>
+      </sect3>
+    </sect2>
+
+    <sect2>
+      <title>Adding Components</title>
```
ghalliday, line 333: This section doesn't really seem to be describing how to add a component - or even which part of the values file relates to the components.  This could either be in with the description of the shipped values file, or this section could be expanded.  E.g., the components are found in this section, each is a list of instances, they share some characteristics such as...

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+        <para><programlisting>  #resources:
+  #  cpu: "1"
+  #  memory: "4G"
+</programlisting>Every component should have a resources entry, but some
+        components such as Thor has multiple resources. The manager: worker,
+        eclagent components all have different resource requirements.</para>
+      </sect3>
+
+      <sect3 id="TAINTS_TOLERATIONS_PLACEMENTS">
+        <title>Taints, Tolerations, and placements</title>
```
ghalliday, line 406: This is a bit of an obscure topic to address this early in the document.  Ok if it is dealing with an expanded section describing common component properties, but there are probably some others that should be here if so.

ghalliday: I wonder if the document would read better if each of these sections were included in the description of the standard values.yaml.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+        example</para>
+
+        <para><programlisting>  root: "hpccsystems" # change to pull your images somewhere other than DockerHub hpccsystems
+</programlisting>In values.yaml file this uses our HPCC Systems specific
+        repository. It is possible you may want to pull from some other
+        repository, this then is where to set that value.</para>
+
+        <para><emphasis role="bold">root:</emphasis> SomeValue</para>
+
+        <para>The root value is a scalar value, Boolean (T|F), or a number
```
ghalliday, line 597: Comment from previous email:

> “root: SomeValue The root value is a scalar value, Boolean (T|F), or a number such as .0”
> This doesn’t seem to be connected with the previous section.  And root has to be a string (see the schema)!

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+        <sect4>
+          <title>temp</title>
+
+          <para>An optional category where temp files can be written
+          to.</para>
+        </sect4>
+      </sect3>
+
+      <sect3 id="CYML_BareMEtalStorage">
+        <title>Bare Metal Storage</title>
```
ghalliday, line 822: This should be the last specific example of storage since it the least likely to be used.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+      <para>Each plane has 3 required fields: The name, the category and the
+      prefix.</para>
+
+      <para>When the system is installed,using the stock supplied values it
+      will create a storage volume which has 1 GB capacity via the following
+      properties.</para>
+
+      <para>For example:</para>
+
+      <programlisting>  name: dali
```
ghalliday, line 690: best to have a - at the start since this is a list entry

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+
+        <para>There are two aspects to using bare metal storage in the
+        Kubernetes system. The first is the <emphasis>hostGroups</emphasis>
+        entry in the storage section which provides named lists of hosts. The
+        hostGroups entries can take one of two forms. This is the most common
+        form, and directly associates a list of host names with a name:</para>
+
+        <programlisting>storage: 
+  hostGroups: 
+  - name: "The name of the host group process"
```
ghalliday, line 832: delete the word process

ghalliday, line 832: Do you have a convention for showing values that are to be replaced?
E.g. angle brackets?
- name: "<The name of the host group>"

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+</programlisting>
+
+        <para>The second form allows one host group to be derived from
+        another:</para>
+
+        <programlisting>storage: 
+  hostGroups: 
+  - name: "The name of the host group process" 
+    hostGroup: "Name of the hostgroup to create a subset of" 
+    count: "Number of hosts in the subset"
```
ghalliday, line 843: The following 3 are numbers, so should not be in quotes.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
ghalliday: So other examples which would come first are
- ephemeral storage
- persistent storage (pvc)

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+        <para>Note: The settings for the PVC's must be ReadWriteMany, except
+        for Dali which can be ReadWriteOnce.</para>
+
+        <para>There are a number of resources, blogs, tutorials, even
+        developer videos that provide step-by-step detail for creating
+        persistent storage volumes.</para>
+      </sect3>
+
+      <sect3 id="YML_TheDataCategory">
+        <title>Cluster Disambiguation</title>
```
ghalliday, line 932: I'm not quite sure how this relates to the rest of the document.

ghalliday (commented):
@g-pan some comments.  I think some of the comments in my previous email are still relevant - I have picked out a couple of them, but it would be worth checking through again.
Main comment is that in some places it feels like a document that has been pulled together from multiple sources, and there still  ares where the same information is repeated twice, or the document doesn't flow quite naturally.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
g-pan: should I just remove this section <sect3> entirely?

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
g-pan: does this even belong in this document?  
perhaps in the beginning section?

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+
+          <para>This is the location where archived workunits, etc are stored
+          and it is typically less speed critical, requiring lower storage
+          costs.</para>
+        </sect4>
+
+        <sect4>
+          <title>spill</title>
+
+          <para>An optional category where the spill files from Thor are
```
jakesmith, line 808: Not just Thor, from any engine (Thor, hThor, Roxie).

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+        code.</para>
+      </sect3>
+    </sect2>
+  </sect1>
+
+  <sect1 id="TheValuesYaml_FileAndRelated" role="nobrk">
+    <title>The HPCC Components in the values.yaml File</title>
+
+    <para>The HPCC Systems Helm charts all ship with stock/default values.
+    These delivered stock Helm chart has a set of default values ideally to be
```
jakesmith, line 1000: > These delivered stock Helm chart has a set of default values ideally to be used as a guide in configuring your deployment.

'delivered stock helm chart' .. is confusing.

It's just a standard helm chart, with defaults (as all helm charts have).

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+    certificates, secrets, and security. This section will provide additional
+    details and any noteworthy insight for the HPCC Systems components defined
+    in the values.yaml file.</para>
+
+    <sect2 id="YAMLHPCC_Components">
+      <title>the HPCC Components</title>
+
+      <para>Some general concepts regarding the HPCC Systems components, as
+      the components are lists.If you had an [ ] empty value list, it would
+      invalidate that list elsewhere. The previous section detailed that
```
jakesmith, line 1016: I think there were some previous comments above that related to this (deleting contents of list to disable a component).
It would be good to consolidate the related comments together in 1 section.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+      global values can be accessed from any chart or subchart, so long as
+      they are explicitly declared. One of the key differences between the
+      bare metal and container/cloud is that in bare metal storage is directly
+      tied to the Thor or the Thor worker nodes or the Roxie worker nodes, or
+      even in the case of the ECLCC Server (DLLS) In containers these are
```
jakesmith, line 1021: >(DLLS) In

missing fullstop.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+      functional system, however you should define the component resources in
+      a manner that corresponds best to your operational strategy.</para>
+
+      <sect3 id="DALI_ValueYAML">
+        <title>Dali</title>
+
+        <para>When configuring Dali, which also has a resources section, it is
+        going to want heaps of memory and a fair bit of CPU as well. It is
+        very important to define these carefully. Otherwise Kubernetes will
+        assign all the pods to same virtual machine and components fighting
```
jakesmith, line 1042: >will assign all the pods to same virtual

could assign all the pods to the same virtual

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+        adding them to their respective lists.</para>
+      </sect3>
+
+      <sect3 id="VALYml_ECLCCServer">
+        <title>ECLCCserver</title>
+
+        <para>Some values of note for ECLCC Server section.</para>
+
+        <para><emphasis role="bold">useChildProcess</emphasis> -- primarily
+        used by ECL Agent and the ECLCC Server. When you submit a job or query
```
jakesmith, line 1084: it's true it's a property of eclagent too, but the heading says ECLCC Server values, so a bit confusing.
I see below the section goes on to talk about the ECL Agent and what useChildProcesses means to it.

So it sounds like the heading ("Some values of note for ECLCC Server section.") is wrong, given it's talking about the useChildProcesses and it's meaning for not just eclccserver

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+        to compile it gets queued and processed, with this option set to true
+        it will spawn a child process utilizing almost no additional overhead
+        in starting. Ideal for sending many small jobs to compile, however it
+        can work to your disadvantage in that a pod needs enough memory to not
+        only listen to the queue but also for job the component is performing.
+        It may need up to 32 GB of memory to listen for jobs. With an ECL
```
jakesmith, line 1090: > It may need up to 32 GB of memory to listen for jobs

this makes it sound like 32GB relates to some in-built HPCC default, which it does not.

I think this example needs clarifying and generlizing.

> however it can work to your disadvantage in that a pod needs enough memory to not only listen to the queue but also for job the component is performing. It may need up to 32 GB of memory to listen for jobs.

Perhaps:

However, because each compile job is no longer executed as an independent pod with it's own resource specifications, but instead run as a child process within the eclccserver pod itself, the eclccserver pod must have been defined with adequate resources for itself (minimal for listening to the queue etc.) and all the jobs it may run in parallel. For example, imagine maxJobs is set to 4, and 4 large queries are queued rapidly, that will mean 4 child processes are launched each consuming cpu and memory within the eclccserver pod.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+        Agent configured with useChildProcesses set to true, each job will run
+        in the same pod (up to the value of maxJobs in parallel). Therefore
```
jakesmith, line 1092: identical behaviour for eclccserver, so should be generalized description for both.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+        with useChildProcesses enabled, the component resources must be
+        defined such that the pod has enough resources to handle the resource
+        demands of all the those jobs to be able to run in parallel.</para>
+
+        <para>With useChildProcess enabled it could be rather expensive in
+        most cloud pricing models, and rather wasteful if there aren't any
+        jobs running. Instead you can set this
+        <emphasis>useChildprocess</emphasis> to false (the default) to start
+        your pod with only the required memory for the job and then dispose of
```
jakesmith, line 1101: > and then dispose of it when done.

which will be disposed of when done.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+        in starting the Kubernetes cluster. Setting
+        <emphasis>useChildProcess</emphasis> to false better allows for the
+        possibility of dynamic scaling. For jobs which would take a long while
+        to compile, the extra (start up) overhead is minimal, and that would
+        be the ideal case to have the <emphasis>useChildProcess</emphasis> as
+        false. Setting <emphasis>useChildProcess</emphasis> to false only
+        allows 1 pod per compile.</para>
+
+        <para><emphasis role="bold">maxActive</emphasis>: The number of how
+        many jobs it can run at any one time. Again use caution because each
```
jakesmith, line 1115: >The number of how many jobs it can run at any one time. 

The maximum number of jobs that can be run in parallel.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+  numWorkers: 2</programlisting></para>
+
+        <para>Continuing on this example, if you wanted a 50 node Thor, you
+        set <emphasis>numWorkers:</emphasis> value to 50 and that would deploy
+        a 50 node Thor. In doing so, ideally you should rename the resource to
+        something like <emphasis>thor_50</emphasis> as in the above example,
+        and that would then re-read the file and spawn the 50 node Thor.
+        Caution however, if you were to put that in a helm update that would
+        do nothing, if there are no jobs running there are no Thor instances
+        running.</para>
```
jakesmith, line 1164: this paragraph is misleading.

The reason nothing would happen on a helm update if no jobs are running, is the same reason you don't see 50 pods when you deploy the helm chart in the 1st place.

Thor instances run dynamically. i.e. a 50-way Thor will be spawned when a job is queued to it.
When it is finished that Thor instance will disappear.
This is the same pattern for compile jobs or eclagent jobs, in those cases a single pod per job is spawned (with useChildProcess set to false), in Thor's case, a collection of pods is spawned (manager + N workers).

It should be emphasised that when idle there are no worker (or manager) pods running.
Updating numWorkers will restart the Thor agent listening to the queue, causing all new jobs to use the new configuration.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+        ECLCC Server, Thor has some analogous options.</para>
+
+        <para><emphasis role="bold">MaxJobs</emphasis>: controls the number of
+        jobs, specifically <emphasis>MaxJobs</emphasis> sets maximum number of
+        jobs.</para>
+
+        <para><emphasis role="bold">MaxGraphs</emphasis>: likewise
+        <emphasis>MaxGraphs</emphasis> also limits the maximum amount of
+        graphs. It makes sense to keep this value at same number, since not
+        all jobs submit graphs - but some jobs may be contained in ECL Agent
```
jakesmith, line 1176: >It makes sense to keep this value at same number

There were some previous comments (email) re. this.
It makes sense (as per the default values) for maxJobs to be higher than maxGraphs (see previous email comments).

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+        and may not want Thor graphs. In that case it wouldn't make sense to
+        have max of 4 If you need a Thor to generate graphs. If there are more
+        than 2 submitted (Thor) graphs, the second would be blocked until the
+        next Thor instance becomes available. It may simply make sense to set
+        <emphasis>maxGraphs</emphasis> to the same value as
+        <emphasis>MaxJobs</emphasis>.</para>
+
+        <para>The delivered (example) chart values defines maxJobs to be
+        greater than maxGraphs. It does so, because that makes sense. Jobs
```
jakesmith, line 1185: see prev. comment. Looks like this was based on previous feedback, but the previous section which states it makes senss for them to be same needs to be revised or deleted.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+        queued to a Thor, aren't always running graphs. They might be blocked
+        on a persist, or updating super files, or waiting on some other
+        workflow event. Therefore it can make sense to have of these jobs,
```
jakesmith, line 1188: >make sense to have of these jobs

make sense to have **more** of these jobs

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+        <para>The Manager and Workers are launched together and consume quite
+        a bit of resoures (and nodes) typically. While the Workflow is
+        inexpensive and usually doesn't require many resources. You might
+        expect in a Kubernetes world, many of them would co-exist on the same
+        node (and therefore be inexpensive). So it makes sense for maxJobs to
+        be higher, and maxGraphs to be lower</para>
+
+        <para>In containers jobs are run independently on individual pods and
+        have no overlap at all, while in bare metal we can have jobs effect
+        other (unrelated) jobs. There is another trade off with using job
```
jakesmith, line 1218: > while in bare metal we can have jobs effect other (unrelated) jobs

makes it sound deliberate/by design. Jobs *could* effect other jobs because they are running in the same process space.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+        </orderedlist>
+
+        <para>The Manager and Workers are launched together and consume quite
+        a bit of resoures (and nodes) typically. While the Workflow is
+        inexpensive and usually doesn't require many resources. You might
+        expect in a Kubernetes world, many of them would co-exist on the same
+        node (and therefore be inexpensive). So it makes sense for maxJobs to
+        be higher, and maxGraphs to be lower</para>
+
+        <para>In containers jobs are run independently on individual pods and
```
jakesmith, line 1216: >In containers jobs are run independently on individual pods

In Kubernetes, jobs run independently in their own pods.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+        have no overlap at all, while in bare metal we can have jobs effect
+        other (unrelated) jobs. There is another trade off with using job
+        processes. If you have large number of small jobs coming in all the
```
jakesmith, line 1219: >If you have large number 

If you have *a* large number

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+        time then you may opt for allowing the system to sit idle and wait for
+        jobs to come in rather than to incur the overhead of starting up an
+        instance only to process a small job, and then discard it only to
+        restart when the next job comes in.</para>
```
jakesmith, line 1223: The paragraph sounds like it's referencing the topic of useChildProcesses?
If so, it's not clear that's true, not sure why it's placed here.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
...
+
+        <para>Compiling jobs and graphs which will run in a completely
+        independent Thor (node), unlike the bare metal where a job can effect
+        future jobs the containerized deployment does not have that problem.
+        However, there is a trade-off between how much cost to incur for
+        uptime, and how much the overhead in starting up an instance for each
```
jakesmith, line 1229: Feels like this has already been said, and this is slightly confusingly written.
Should consolidate into a single clearer topic section.

jakesmith (commented):
@g-pan - Please see comments.

There are various topics revisited in various places, that could do with consolidation into 1 section for each.

There also seemed to be quite a few non-technical issues, e.g. missing words, gramatical type issues, and/or sentences could do with rewording slightly,
that were quite easy to spot when reading through.

So worth re-reading carefully to flush out all other similar issues for next review.

richardkchapman:
Missing checkboxes

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1249 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE section PUBLIC "-//OASIS//DTD DocBook XML V4.5//EN"
+"http://www.oasis-open.org/docbook/xml/4.5/docbookx.dtd">
+<chapter id="ContainerConfigurationcHAPTER">
+  <title>Containerized Configuration</title>
```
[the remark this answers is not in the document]
g-pan: the justification there follows the style for initial chapter pages in all docs

g-pan:
Made significant revisions based on feedback. Purposely did not squash and rebase - in order to preserve history and comments. (however will do so upon request)  Can provide composed PDF for ease of review - again upon request. 
@ghalliday  @jakesmith  please review if possible. Thank you.

ghalliday (commented):
@g-pan comments so far (I haven't ready any details on the components);

one general comment is that the conversations should be resolved by the reviewer rather than the contributor.  (See https://github.com/hpcc-systems/HPCC-Platform/pull/15603 which it would be useful to have your input on)

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1255 @@
...
+        parallel). Therefore with <emphasis>useChildProcesses</emphasis>
+        enabled, the component resources must be defined such that the pod has
+        enough resources to handle the resource demands of all the those jobs
+        to be able to run in parallel.</para>
+
+        <para>With useChildProcess enabled it could be rather expensive in
+        most cloud pricing models, and rather wasteful if there aren't any
+        jobs running. Instead you can set this
+        <emphasis>useChildprocess</emphasis> to false (the default) to start
+        your pod with only the required memory for the job which will be
```
ghalliday, line 1129: clarify: to start a pod to compile each query with the the required memory...

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1255 @@
...
+        seconds to a minute to spawn the Kubernetes cluster to process the
+        job. Which may not be ideal for an environment which is sending
+        several small jobs, but rather larger jobs which would minimize the
+        effect of the overhead in starting the Kubernetes cluster. Setting
+        <emphasis>useChildProcess</emphasis> to false better allows for the
+        possibility of dynamic scaling. For jobs which would take a long while
+        to compile, the extra (start up) overhead is minimal, and that would
+        be the ideal case to have the <emphasis>useChildProcess</emphasis> as
+        false. Setting <emphasis>useChildProcess</emphasis> to false only
+        allows 1 pod per compile.</para>
```
ghalliday, line 1140: Note: https://github.com/hpcc-systems/HPCC-Platform/pull/15609 changes this slightly.  There is now an option to try compiling the query within the process, but kill it if it takes more than 10 seconds, and launch a job instead.  So small jobs will run quickly, larger jobs will start a K8s job.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1255 @@
...
+        big job and in that case spawn some 2000 jobs using a considerable
+        amount of resources, which could potentially run up a rather expensive
+        compilation bill, again depending on your cloud provider and your
+        billing plan.</para>
+      </sect3>
+
+      <sect3 id="ValYML_Sasha">
+        <title>Sasha</title>
+
+        <para>Sasha is an outlier as it is a dictionary type structure and not
```
ghalliday, line 1155: The configuration for Sasha is an outlier..

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1255 @@
...
+        a list. You can't have more than one archiver or dfuwu-archiver as
+        that is a value limitation, you can choose to either have the service
+        or not (set the 'disabled' value to true). The Sasha component is an
```
ghalliday, line 1158: Not sure what this sentence is conveying.  Is it just a repeat of the 1st in the paragraph.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1255 @@
...
+
+      <sect3 id="ValYML_Thor">
+        <title>Thor</title>
+
+        <para>Thor instances run dynamically. As do the other components in
+        containers. In Thor's case, a collection of pods is spawned (manager +
+        N workers) when jobs are queued to it. When idle there are no worker
+        (or manager) pods running.</para>
+
+        <para>In the case of Thor (components) there are similarities with the
```
ghalliday, line 1170: not sure that referring to eclccserver is helpful/clarifies anything.  rather something like

"The configuration for thor also consists of a list of thor instances.  Each instance dynamically spawns...

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1255 @@
...
+        ECLCC Server. Thor is also a list, and if that list has a value
+        of</para>
+
+        <programlisting>-name: thor </programlisting>
```
ghalliday, line 1174: not sure how these examples relate to the surrounding text.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1255 @@
...
+        something which clearly describes it, such as
+        <emphasis>thor_50</emphasis> as in the above example. Updating the
+        <emphasis>numWorkers</emphasis> value will restart the Thor agent
+        listening to the queue, causing all new jobs to use the new
+        configuration.</para>
+
+        <para>Similarly to the <emphasis>maxActive</emphasis> values in the
+        ECLCC Server, Thor has some analogous options.</para>
+
+        <para><emphasis role="bold">MaxJobs</emphasis>: controls the number of
```
ghalliday, line 1202: careful about casing.  "maxJobs" not "MaxJobs".  Also "maxGraphs".

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1255 @@
...
+        <para>Continuing on this example, if you wanted a 50-way Thor, you set
+        <emphasis>numWorkers:</emphasis> value to 50 and that would deploy a
+        50 node Thor. In doing so, ideally you should rename the resource to
+        something which clearly describes it, such as
+        <emphasis>thor_50</emphasis> as in the above example. Updating the
+        <emphasis>numWorkers</emphasis> value will restart the Thor agent
+        listening to the queue, causing all new jobs to use the new
+        configuration.</para>
+
+        <para>Similarly to the <emphasis>maxActive</emphasis> values in the
```
ghalliday, line 1199: not sure comparison with eclccserver helps.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1255 @@
...
+        ECLCC Server, Thor has some analogous options.</para>
+
+        <para><emphasis role="bold">MaxJobs</emphasis>: controls the number of
+        jobs, specifically <emphasis>MaxJobs</emphasis> sets maximum number of
+        jobs.</para>
+
+        <para><emphasis role="bold">MaxGraphs</emphasis>: likewise
+        <emphasis>MaxGraphs</emphasis> also limits the maximum amount of
```
ghalliday, line 1207: This doesn't help me understand what the difference between the settings is (I don't know, but I can guess). I think this needs to explain that thor queries may only spend some of their time running graphs.  I'm not sure if having two different settings is very useful.  @jakesmith ?

ghalliday (commented):
A  few extra comments on the components.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1255 @@
...
+        in separate docs/tutorials) or you can use it as a basis for creating
+        your own override chart.</para>
+
+        <para>To define a storage plane that utilizes a PVC, you must decide
+        on where that data will reside. You create the storage directories,
+        with the appropriate names and then you can install the localfiles
+        Helm chart to create the volumes to use the local storage option, such
+        as in the following example:</para>
+
+        <programlisting>helm install mycluster hpcc/ -f examples/local/values-localfile.yaml</programlisting>
```
jakesmith, line 681: to be consistent, should use repo. form (not point to directory), i.e.: should say "hpcc/hpcc" not "hpcc/"

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1255 @@
...
+    <sect2>
+      <title>Replicas and Resources</title>
+
+      <para>Other noteworthy values in the charts that have bearing on HPCC
+      Systems set up and configuration.</para>
+
+      <sect3 id="REPLICAS_">
+        <title>Replicas</title>
+
+        <para>Replicas: defines how many replica nodes come up, how many pods
```
jakesmith, line 892: "replicas" not "Replicas"

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
jakesmith: The idea was that jobs may spend significant amount of time outside of graphs, e.g. waiting on workflow state (outside of the Thor engine itself), like on persist points, superfile ops. etc., so it makes sense for Thor's to have a higher limit of concurrent jobs (maxJobs) than graphs (maxGraphs / Thor instances).

I agree, it is a bit confusing, but I also think we want this disctinction, because Thor instances (graphs) are expensive (lots of pods/higher resources), and workflow pods (jobs) are relatively cheap.

I wonder if we should change it, so maxActive means what maxGraphs (Thor instances) means now,
but there is also still a maxJobs, which is not set by default and is automatically set to e.g. 50% more than maxActive.

@ghalliday - thoughts?

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
g-pan: Question: The change was to useChildProcess behavior, or is it adding the ChildProcessTimeLimit, and should I include some additional language for that time limit?

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
g-pan: likely. Just 'notes' I had jotted down felt relevant. If irrelevant or repetitive, should I remove??

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
g-pan: It does seem a bit awkward. 
Would more of the Thor component list in the values.yaml make more sense?    

thor:
- name: thor
  prefix: thor
  numWorkers: 2
  maxJobs: 4
  maxGraphs: 2

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1255 @@
...
+      <title>Global</title>
+
+      <para>The first section of the values.yaml file describes global values.
+      The global.image.root is a string denoting which version to pull. Global
+      applies generally to everything.</para>
+
+      <programlisting># Default values for hpcc.
+
+global:
+  # Settings in the global section apply to all HPCC components in all subcharts
```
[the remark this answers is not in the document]
g-pan: Question: This example code was a "copy - paste" directly from line 4 in the values.yaml file. 
I can remove it from this doc, would I need to modify the values.yaml file (or create a JIRA for someone else to)?

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1255 @@
...
+
+  # logging sets the default logging information for all components. Can be overridden locally
+  logging:
+    detail: 80
+</programlisting>
+
+      <para>In the delivered HPCC Systems values.yaml file excerpt (above)
+      <emphasis>global:</emphasis> is a top level dictionary. As noted in the
+      comments, the settings in the global section apply to all HPCC Systems
+      components in any and all sub-charts. Note from the indentation that the
```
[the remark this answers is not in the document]
g-pan: this is my content, I will make the change, and also remove any other mentions of sub-charts

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1255 @@
...
+        keep in mind where multiple yaml files are applied, the later ones
+        will always overwrite the values in the earlier ones. They are merged
+        in in sequence.</para>
+
+        <para>Another point to consider, where there is a global dictionary
+        such as root: and its value redefined in the 2nd file (as a
+        dictionary) it would not be overwritten. You can't simply overwrite a
+        dictionary. You can redefine a dictionary and set it to null (such as
+        the Thor example in the previous section), which will effectively wipe
+        it out.</para>
```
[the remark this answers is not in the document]
g-pan: Could you please suggest a dictionary I could use as an example, or should I change the preceding text to read list instead of dictionary?

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
ghalliday: I think it is ok if is explained clearly.

"since not all jobs submit graphs - but some jobs may be contained in ECL Agent and may not want Thor graphs."

This isn't really correct - Jake's comment above is better.  The issue is that the thor jobs are not executing graphs all the time, but they will some of the time.  Another example is spraying files.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
ghalliday: I think it is more that the sentence doesn't make sense.

"and if that list has a value of

        -name: thor

or

        -name: thor_50

If you have..."

That isn't really a sentence. and I don't understand what it means..

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
ghalliday: The change is adding childProcessTimeLimit, and should probably have some extra language.

g-pan:
Made some significant changes - per the suggestions of moving the general Helm/Yaml info towards the end and bringing forward the more HPCC Systems specifics - in addition to addressing the many comments. If it would be helpful I could close this PR and reissue a new PR...as this one seems to be a little confusing after this type of change. If having a composed PDF file would make review easier... please let me know I can provide it.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1270 @@
...
+      can contain if it is a list or dictionary. For instance:</para>
+
+      <programlisting>    "roxie": { 
+      "description": "roxie process",
+      "type": "array"
+      "items": { "$ref": "#/definitions/roxie" }
+    },</programlisting>
+
+      <para>The "items" in this case are a reference ($ref) to another section
+      of the schema. Each plane, in the schema file has a list of properties
```
ghalliday, line 84: This seems to be a mixture of an example and a general statement.  I think clearer would be to move this after the next statement, and start it with

"For example, for a storage plane the schema file has a list of properties including a prefix...."

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1270 @@
...
+        },
+        "secret": {
+          "description": "optional name of any secret required to access this storage plane",
+          "type": "string"
+
+
+</programlisting>
+
+      <para>Note the <emphasis>additionalProperties</emphasis> value typically
+      at the end of each section in the schema. It specifies whether the
```
ghalliday, line 114: If this is being noted it should be included in the example above.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1270 @@
...
+          "type": "string"
+        },
+        "subPath": {
+          "description": "optional subdirectory within the mount directory",
+          "type": "string"
+        },
+        "secret": {
+          "description": "optional name of any secret required to access this storage plane",
+          "type": "string"
+
```
ghalliday, line 109: missing closing brackets from the example.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1270 @@
...
+  </sect1>
+
+  <sect1 id="TheValuesYaml_FileAndRelated" role="nobrk">
+    <title>The HPCC Components in the values.yaml File</title>
+
+    <para>The HPCC Systems Helm charts all ship with stock/default values.
+    These Helm charts have a set of default values ideally to be used as a
+    guide in configuring your deployment. Generally, every HPCC Systems
+    component is a list. That list defines the properties for each
+    component.</para>
```
ghalliday, line 133: I think "each instance of the component. " is more correct.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1270 @@
...
+  #  memory: "4G"
+</programlisting>The stock values file will work and allow you to stand up a
+      functional system, however you should define the component resources in
+      a manner that corresponds best to your operational strategy.</para>
+
+      <sect3 id="DALI_ValueYAML">
+        <title>Dali</title>
+
+        <para>When configuring Dali, which also has a resources section, it is
+        going to want heaps of memory and a fair bit of CPU as well. It is
```
ghalliday, line 170: "heaps" "fair bit" a bit colloquial?

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1270 @@
...
+      </sect3>
+    </sect2>
+
+    <sect2 id="StorageItems_HPCC_Systems_Coomponents">
+      <title>Storage Items for HPCC Systems Components</title>
+
+      <sect3 id="YML-DOC_GenData-Storage">
+        <title>General Data Storage</title>
+
+        <para>General data files generated by HPCC are stored stored in data.
```
ghalliday, line 700: should be a \<sect4> for data category for consistency

ghalliday (commented):
@g-pan A few comments.  I suspect it is best to get this merged asap so it is ready for 8.6, and then circle back and edit further.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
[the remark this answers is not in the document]
jakesmith: values.yaml should be modified there are no sub-charts of the hpcc chart.
Yes, if you could open a new JIRA and/or fix that separately that would be great. Thanks.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
g-pan: this one is tricky because it would require adding some content at the sect3 level to be able to move this down to the sect4 level... instead I am going to bump the subsequent sect4 headings up to the sect3 to make this section more uniform.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
g-pan: probably better to just choose an example with less properties than the storagePlane to illustrate the the json file, I can always revert, if necessary however using storage seems the entire section there only has the 2 properties and will fit nicely on a single page - think that would suffice?

g-pan:
Agreed, unless there are any major inaccuracies, we can get this merged, and can make edits as necessary.

ghalliday (approved):
I think good enough.  @g-pan when Jake approves please squash the PR.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1269 @@
...
+      <para>All storage should be in a storage plane definition. This is best
+      described in the comment in the storage definition in the values
+      file.</para>
+
+      <programlisting>planes:
+  #   name: &lt;required&gt;
+  #   prefix: &lt;path&gt;                        # Root directory for accessing the plane 
+                                            # (if pvc defined), 
+  #                                         # or url to access plane.
+  #   category: data|dali|lz|dll|spill|temp # What category of data is stored on this plane?
```
jakesmith, line 471: this should updated for 8.6, where "git" and "remote" will now be plane types too.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1269 @@
...
+    offset: &lt;the first host to include in the subset&gt;  
+    delta: &lt;Cycle offset to apply to the hosts&gt; 
+</programlisting>
+
+        <para>Some typical examples with bare-metal clusters are smaller
+        subsets of the host, or the same hosts, but storing different parts on
+        different nodes, for example:</para>
+
+        <programlisting>storage: 
+  hostGroups:
```
jakesmith, line 648: future: it feels like this (and probably other specialist areas), should be rerefenced, but not in this main doc. @ghalliday what do you think

jakesmith (approved):
@g-pan - a few comments for future revisions.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
ghalliday: Yes.  Need two different documents really.  An introduction - which this is, and a reference guide, which would go through the options in detail.

docs/EN_US/ContainerizedHPCC/ContainerizedMods/ConfigureValues.xml
```diff
@@ -0,0 +1,1269 @@
...
+          </listitem>
+        </orderedlist>
+
+        <para>The Manager and Workers are launched together and consume quite
+        a bit of resoures (and nodes) typically. While the Workflow is
+        inexpensive and usually doesn't require as many resources. You might
+        expect in a Kubernetes world, many of them would co-exist on the same
+        node (and therefore be inexpensive). So it makes sense for
+        <emphasis>maxJobs</emphasis> to be higher, and
+        <emphasis>maxGraphs</emphasis> to be lower</para>
```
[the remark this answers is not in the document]
ghalliday: Yes.  I think it might be better to describe it the other way round even.  E.g.

  maxGraphs is the maximum number of thor clusters that you want to support.  However some thor jobs spend a significant quantity of their time outside of a graph - e.g. spraying a file, or processing super file operations.  So maxJobs can be set to a higher number to allow those low-resource jobs to run at the same time.

stuartort:
the reference guide is being worked on separately

PR #15597, opened 2021-11-23
HPCC-26852 DFUQuery to return cost field

shamser:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dali/base/dadfs.cpp
+++ b/dali/base/dadfs.cpp
@@ -12785,7 +12785,8 @@ IDFProtectedIterator *CDistributedFileDirectory::lookupProtectedFiles(const char
 
 const char* DFUQResultFieldNames[] = { "@name", "@description", "@group", "@kind", "@modified", "@job", "@owner",
     "@DFUSFrecordCount", "@recordCount", "@recordSize", "@DFUSFsize", "@size", "@workunit", "@DFUSFcluster", "@numsubfiles",
-    "@accessed", "@numparts", "@compressedSize", "@directory", "@partmask", "@superowners", "@persistent", "@protect", "@compressed" };
+    "@accessed", "@numparts", "@compressedSize", "@directory", "@partmask", "@superowners", "@persistent", "@protect", "@compressed",
+    "@cost", "@numDiskReads", "@numDiskWrites" };
 
 extern da_decl const char* getDFUQResultFieldName(DFUQResultField feild)
 {
@@ -12854,6 +12855,28 @@ IPropertyTreeIterator *deserializeFileAttrIterator(MemoryBuffer& mb, unsigned nu
                 file->setPropBool(getDFUQResultFieldName(DFUQRFiscompressed), true);
         }
 
+        void setCost(IPropertyTree* file, const char *nodeGroup)
+        {
+            StringBuffer str;
+            double fileAgeDays = 0.0;
+            if (file->getProp(getDFUQResultFieldName(DFUQRFtimemodified), str))
+            {
+                CDateTime dt;
+                dt.setString(str.str());
+                fileAgeDays = difftime(time(nullptr), dt.getSimple())/(24*60*60);
+            }
+            __int64 sizeDiskSize = 0;
+            if (isCompressed(*file))
+                sizeDiskSize = file->getPropInt64(getDFUQResultFieldName(DFUQRFcompressedsize), 0);
+            else
+                sizeDiskSize = file->getPropInt64(getDFUQResultFieldName(DFUQRForigsize), 0);
+            double sizeGB = sizeDiskSize / ((double)1024 * 1024 * 1024);
+            __int64 numDiskWrites = file->getPropInt64(getDFUQResultFieldName(DFUQRFnumDiskReads), 0);
+            __int64 numDiskReads = file->getPropInt64(getDFUQResultFieldName(DFUQRFnumDiskWrites), 0);
+            double totalCost = calcFileCost(nodeGroup, sizeGB, fileAgeDays, numDiskWrites, numDiskReads);
+            file->setPropReal(getDFUQResultFieldName(DFUQRFcost), totalCost);
+        }
+
         IPropertyTree *deserializeFileAttr(MemoryBuffer &mb, StringArray& nodeGroupFilter)
         {
             IPropertyTree *attr = getEmptyAttr();
@@ -12892,6 +12915,9 @@ IPropertyTreeIterator *deserializeFileAttrIterator(MemoryBuffer& mb, unsigned nu
             attr->setPropInt64(getDFUQResultFieldName(DFUQRFsize), attr->getPropInt64(getDFUQResultFieldName(DFUQRForigsize), -1));//Sort the files with empty size to front
             setRecordCount(attr);
             setIsCompressed(attr);
+            const char *firstNodeGroup = attr->queryProp(getDFUQResultFieldName(DFUQRFnodegroup));
+            if (!isEmptyString(firstNodeGroup))
+                setCost(attr, firstNodeGroup);
             return attr;
         }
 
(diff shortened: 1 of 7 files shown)


dali/base/dadfs.cpp
```diff
@@ -12892,6 +12923,7 @@ IPropertyTreeIterator *deserializeFileAttrIterator(MemoryBuffer& mb, unsigned nu
             attr->setPropInt64(getDFUQResultFieldName(DFUQRFsize), attr->getPropInt64(getDFUQResultFieldName(DFUQRForigsize), -1));//Sort the files with empty size to front
             setRecordCount(attr);
             setIsCompressed(attr);
+            setCost(attr);
```
jakesmith, line 12926: function: I think there'll be a problem here if the file is on >1 plane/filegroup.

If it is, each copy is returned/listed individually, and duplicateFileAttrOnOtherNodeGroup is called to clone the deserialized properties (which will include cost).
It looks like as it stands, that will mean each file copy will appear to have the sum total cost, not the individual cost on the plane that the copy resides on.

I think the call here should say something like:
```
const char *firstNodeGroup = attr->queryProp(getDFUQResultFieldName(DFUQRFnodegroup));
if (!isEmptyString(firsNodeGroup))
    setCost(attr, firstNodeGroup);
```
( and setCost changed so it deals with a specific node group ).

Then in duplicateFileAttrOnOtherNodeGroup it can call setCost again for the node group it has moved on to.

esp/services/ws_dfu/ws_dfuHelpers.cpp
```diff
@@ -183,6 +183,8 @@ bool WsDFUHelpers::addToLogicalFileList(IPropertyTree& file, const char* nodeGro
             if (!accessed.isEmpty())
                 lFile->setAccessed(accessed.replace('T', ' '));
         }
+        if (version >= 1.61)
+            lFile->setCost(file.getPropReal(getDFUQResultFieldName(DFUQRFcost)));
```
jakesmith, line 187: question: what happens if the file is a super file? (or a super file with >1 subfile).
I don't think the stats will be available/returned, because that would require the server to have aggregated numDiskReads/numDiskWrites, and there'd be a combination of locations too..
NB: they may be available if it's a super of just 1 sub.

Worth double checking, but if needed attributes not available, it's not worth calling setCost here

jakesmith: future: if above is true, it will be a known caveat that supers (poss. with >1 sub) do not show aggregate costs, but worth opening a JIRA for future enhancement.

jakesmith (commented):
@shamser - looks good in general, spotted a couple of potential issues

dali/base/dadfs.cpp
```diff
@@ -12854,6 +12854,37 @@ IPropertyTreeIterator *deserializeFileAttrIterator(MemoryBuffer& mb, unsigned nu
...
+            }
+            else
+            {
+                ForEachItemIn(i,fileNodeGroups)
+                {
+                    const char * group = fileNodeGroups.item(i);
+                    totalCost += calcFileCost(group, sizeGB, fileAgeDays, numDiskWrites, numDiskReads);
+                }
+            }
+            file->addPropReal(getDFUQResultFieldName(DFUQRFcost), totalCost);
```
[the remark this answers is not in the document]
shamser: This is for the  logical file table page.  They can see a break-down by opening the logical file's detail page.  I'm not sure this extra information in the logical file table page is that useful.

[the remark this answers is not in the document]
jakesmith: >They can see a break-down by opening the logical file's detail page

right, ok, yes that is probably sufficient. Unless it was useful to search and order by files that had cost most to store vs read or vice versa, but even if it is, not for this PR.

esp/services/ws_dfu/ws_dfuHelpers.cpp
shamser: You're absolutely right. Super files aren't supported at the moment.  I'll open a Jira to look into this later.

dali/base/dadfs.cpp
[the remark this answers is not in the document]
shamser: I'll leave as is for now.  It can always be added later if the functionality is requested.

dali/base/dadfs.cpp
```diff
@@ -12854,6 +12854,37 @@ IPropertyTreeIterator *deserializeFileAttrIterator(MemoryBuffer& mb, unsigned nu
...
 
+        void setCost(IPropertyTree* file)
+        {
+            StringBuffer str;
+            double fileAgeDays = 0.0;
+            if (file->getProp("@modified",str))
+            {
+                CDateTime dt;
+                dt.setString(str.str());
+                fileAgeDays = difftime(time(nullptr), dt.getSimple())/(24*60*60);
```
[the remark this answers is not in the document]
shamser: Yes, the calculations would be incorrect.  However, I believe support for multiple time-zones needed to be supported, is outside the scope of this Jira.

dali/base/dadfs.cpp
shamser: @jakesmith if the firstNodeGroup is empty, it isn't going to calculate any cost for the file. Is that ok?  When is this likely to be empty?

jakesmith: I think that's correct. There are 2 ways it could be empty afaics:
1) for a superfile, where either it has 0 subs, or >1 (if >1, then it's the separate issue of handling superfiles in general).
2) the groups have been filtered out by nodeGroupFilter, during deserialization, i.e. resulting in fileNodeGroups container no groups.

dali/base/dadfs.cpp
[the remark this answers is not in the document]
jakesmith: I agree, as it's a pre-existing issue with getCost(), but it's likely to be a common scenario (at least for LN) and quite significant(ly misleading).

Can you open a JIRA and reference it here and/or link it to this PR's JIRA?

esp/services/ws_dfu/ws_dfuHelpers.cpp
jakesmith: for reference, can you reference the new JIRA here and/or link it to this PR's JIRA?

shamser: Superfile costs: https://track.hpccsystems.com/browse/HPCC-26871

dali/base/dadfs.cpp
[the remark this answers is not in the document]
shamser: Jira to examine TZ issue: https://track.hpccsystems.com/browse/HPCC-26872

dali/base/dadfs.cpp
jakesmith: >Then in duplicateFileAttrOnOtherNodeGroup it can call setCost again for the node group it has moved on to.

setCost needs recalling as it walks other filtered groups.

jakesmith (commented):
@shamser - 1 issue remaining I think.

dali/base/dadfs.cpp
```diff
@@ -12854,6 +12855,28 @@ IPropertyTreeIterator *deserializeFileAttrIterator(MemoryBuffer& mb, unsigned nu
...
+            __int64 sizeDiskSize = 0;
+            if (isCompressed(*file))
+                sizeDiskSize = file->getPropInt64(getDFUQResultFieldName(DFUQRFcompressedsize), 0);
+            else
+                sizeDiskSize = file->getPropInt64(getDFUQResultFieldName(DFUQRForigsize), 0);
+            double sizeGB = sizeDiskSize / ((double)1024 * 1024 * 1024);
+            __int64 numDiskWrites = file->getPropInt64(getDFUQResultFieldName(DFUQRFnumDiskReads), 0);
+            __int64 numDiskReads = file->getPropInt64(getDFUQResultFieldName(DFUQRFnumDiskWrites), 0);
+            double totalCost = calcFileCost(nodeGroup, sizeGB, fileAgeDays, numDiskWrites, numDiskReads);
+            file->addPropReal(getDFUQResultFieldName(DFUQRFcost), totalCost);
```
jakesmith, line 12877: trivial: this should really be setPropReal , addProp* implies adding multiple property values of the specified name.

jakesmith (approved):
@shamser - looks good, 1 trivial new comment I'd missed before. Please go ahead and squash.

jakesmith (approved):
@shamser - looks good to me.

jakesmith (approved):

shamser:
@GordonSmith This PR extends DFUQuery to return the cost (CostFileAccess + CostExecute) in the 'cost' field.

shamser:
@ghalliday Please can you merge this.

PR #15601, opened 2021-11-24
HPCC-26868 WUQuery to return cost fields

shamser:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/common/workunit/workunit.cpp
+++ b/common/workunit/workunit.cpp
@@ -3660,20 +3660,8 @@ public:
         totalThorTime = (unsigned)nanoToMilli(extractTimeCollatable(p.queryProp("@totalThorTime"), nullptr));
         _isProtected = p.getPropBool("@protected", false);
 
-        Owned<IPropertyTreeIterator> iterCostExecute = p.getElements("./Statistics/Statistic[@kind='CostExecute'][@scope='']");
-        costExecute = 0;
-        ForEach(*iterCostExecute)
-        {
-            IPropertyTree &stat = iterCostExecute->query();
-            costExecute += stat.getPropInt64("@value");
-        }
-        Owned<IPropertyTreeIterator> iterFileAccess = p.getElements("./Statistics/Statistic[@kind='CostFileAccess'][@scope='']");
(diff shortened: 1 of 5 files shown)


common/workunit/workunit.cpp
```diff
@@ -3659,6 +3659,21 @@ class CLightweightWorkunitInfo : public CInterfaceOf<IConstWorkUnitInfo>
         appvalues.loadBranch(&p,"Application");
         totalThorTime = (unsigned)nanoToMilli(extractTimeCollatable(p.queryProp("@totalThorTime"), nullptr));
         _isProtected = p.getPropBool("@protected", false);
+
+        Owned<IPropertyTreeIterator> iterCostExecute = p.getElements("./Statistics/Statistic[@kind='CostExecute'][@scope='']");
```
ghalliday, line 3663: question: Could this more simply be coded as
`costExecute = p.getPropInt64("./Statistics/Statistic[@kind='CostExecute'][@scope='']/@value");
`
?

ghalliday, line 3663: design: I wonder if it should really be stored as @executedCost in the workunit when it is aggregated.  It would add a minor impact when aggregating, but would save a relatively complex data transfer when getting the workunit information.
The other issue is how does this work with cassandra workunits?  If you stored it in an attribute I think it would be a relatively trivial change.  Without it it may be complex.
You could take it further and create an implicit statistic from the workunit property, but that probably doesn't have a major benefit.

ghalliday (commented):
@shamser the change looks good.  I have a question/suggestion about the efficiency to discuss.

shamser:
@ghalliday The cost values are being stored in the workunit now and accessed from the workunit, so that the entire Statistics subtree is pulled in from dali.

ghalliday (approved):
Looks nice and clean.

PR #15609, opened 2021-12-07
HPCC-26841 Use child processes for small eclcc compiles

richardkchapman:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Cloud-compatibility
  - [x] Premature optimization
  - [ x Existing deployed queries will not be broken

## Smoketest:

## Testing:
Using ECL code that constant-folds a delay...

### DIFF
--- a/ecl/eclccserver/eclccserver.cpp
+++ b/ecl/eclccserver/eclccserver.cpp
@@ -131,26 +131,38 @@ private:
 // Check for aborts of the workunit as it is compiling
 //------------------------------------------------------------------------------------------------------------------
 
+static bool useChildProcesses = false;      // Use k8s jobs for compile tasks
+static unsigned childProcessTimeLimit = 0;  // If using k8s jobs to compile, try a child process first but abort if it takes longer than this time (seconds)
+
 class AbortWaiter : public Thread
 {
 public:
-    AbortWaiter(IPipeProcess *_pipe, IConstWorkUnit *_wu)
-        : Thread("EclccCompileThread::AbortWaiter"), pipe(_pipe), wu(_wu)
+    AbortWaiter(IConstWorkUnit *_wu, unsigned _timeLimit)
+        : Thread("EclccCompileThread::AbortWaiter"), wu(_wu), timeLimit(_timeLimit)
     {
     }
 
     virtual int run()
     {
         wu->subscribe(SubscribeOptionAbort);
+        unsigned start = msTick();
+        timedOut = false;
         try
         {
             for (;;)
             {
-                if (sem.wait(2000))
+                if (sem.wait(1000))
                     break;
-                if (wu->aborting())
+                timedOut = (timeLimit && (msTick() - start >= timeLimit*1000));
+                if (timedOut || wu->aborting())
                 {
-                    pipe->abort();
+                    CriticalBlock b(crit);
+                    DBGLOG("Aborting compilation");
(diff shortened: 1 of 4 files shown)


ecl/eclccserver/eclccserver.cpp
```diff
@@ -301,7 +342,43 @@ class EclccCompileThread : implements IPooledThread, implements IErrorReporter,
...
     }
 
-    unsigned executeLine(const char *line, StringBuffer &output, bool alreadyFailed)
+    unsigned doRunCompileCommand(AbortWaiter &abortWaiter, StringBuffer &output, const char *cmd)
+    {
+        try
+        {
+            Owned<IPipeProcess> pipe = createPipeProcess();
+            AbortPipeWaiter aborter(abortWaiter, pipe);
+            int ret = START_FAILURE;
```
ghalliday, line 351: Should this check abortWaiter.hasTimedout()? otherwise there is a small race condition.  Or better recode IPipeProcess to save that it has been aborted, and fail when run is being called (otherwise there will still be a window).

ecl/eclccserver/eclccserver.cpp
```diff
@@ -560,12 +641,11 @@ class EclccCompileThread : implements IPooledThread, implements IErrorReporter,
             }
             unsigned retcode = pipe->wait();
             errorReader->join();
-            abortWaiter->stop();
```
ghalliday, line 563: Test by adding a sleep here - and ensure that the compile is aborted.

ghalliday (commented):
@richardkchapman broadly looks correct.  Only concern is what happens if it times out after one pipe process has completed, and before another completes.  Either between eclcc and gcc, or between gcc commands.

richardkchapman:
Added fixes for issues raised in review

helm/hpcc/values.yaml
```diff
@@ -428,6 +428,7 @@ eclccserver:
   replicas: 1
   ## Set to false if you want to launch each workunit compile in its own container, true to run as child processes in eclccserver pod.
   useChildProcesses: false
+  childProcessTimeLimit: 10
```
ghalliday, line 431: minor: Should this also go in the schema, with a comment that 0 disables it.

ghalliday (approved):
@richardkchapman one comment, and please squash.

richardkchapman:
@ghalliday Squashed and added to schema

helm/hpcc/values.yaml
```diff
@@ -428,6 +428,9 @@ eclccserver:
   replicas: 1
   ## Set to false if you want to launch each workunit compile in its own container, true to run as child processes in eclccserver pod.
   useChildProcesses: false
+  ## If non-zero, and useChildProcesses is false, try speding up to this number of seconds compiling using a child process before switching to
```
ghalliday, line 431: typo: spe*n*ding

ecl/eclccserver/eclccserver.cpp
richardkchapman: I did... and (in latest version) it was.

helm/hpcc/values.yaml
richardkchapman: Fixed and repushed

PR #15622, opened 2021-12-16
HPCC-26910 LDAP SecMgr should create Admin group/user on startup

RussWhitehead:
Allow HPCCAdmin User to be specified as secret.
Precreate HPCC Admin group, add HPCCAdmin User to that group, set
group permissions to allow ECLWatch

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [x] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/applications/common/ldap/ldap.yaml
+++ b/esp/applications/common/ldap/ldap.yaml
@@ -20,4 +20,6 @@ ldap:
   workunitsBasedn: ou=workunits,ou=ecl
   ldapAdminSecretKey: "ldapadmincredskey"
   ldapAdminVaultId: ""
+  hpccAdminSecretKey: ""
+  hpccAdminVaultId: ""
   checkScopeScans: true
--- a/system/security/LdapSecurity/ldapconnection.cpp
+++ b/system/security/LdapSecurity/ldapconnection.cpp
@@ -277,6 +277,9 @@ private:
     bool                 m_sysuser_specified;
     StringBuffer         m_sysuser_dn;
 
+    StringBuffer         m_HPCCAdminUser_username;
+    StringBuffer         m_HPCCAdminUser_password;
+
     int                  m_maxConnections;
     
     StringBuffer         m_sdfieldname;
@@ -360,19 +363,19 @@ public:
         //  - Hardcoded : systemCommonName, systemPassword (legacy environment.xml)
         //------------------------------------------------
 
-        StringBuffer systemUserSecretKey;
-        cfg->getProp(".//@ldapAdminSecretKey", systemUserSecretKey);//vault/secrets LDAP username key
-        if (!systemUserSecretKey.isEmpty())
+        StringBuffer adminUserSecretKey;
+        cfg->getProp(".//@ldapAdminSecretKey", adminUserSecretKey);//vault/secrets LDAP username key
+        if (!adminUserSecretKey.isEmpty())
         {
             StringBuffer vaultId;
             cfg->getProp(".//@ldapAdminVaultId", vaultId);//optional HashiCorp vault ID
 
-            DBGLOG("Retrieving LDAP Admin username/password from secrets repo: %s %s", !vaultId.isEmpty() ? vaultId.str() : "", systemUserSecretKey.str());
+            DBGLOG("Retrieving LDAP Admin username/password from secrets repo: %s %s", !vaultId.isEmpty() ? vaultId.str() : "", adminUserSecretKey.str());
             Owned<IPropertyTree> secretTree;
-            if (!isEmptyString(vaultId.str()))
-                secretTree.setown(getVaultSecret("authn", vaultId, systemUserSecretKey.str(), nullptr));
+            if (!vaultId.isEmpty())
+                secretTree.setown(getVaultSecret("authn", vaultId, adminUserSecretKey.str(), nullptr));
             else
-                secretTree.setown(getSecret("authn", systemUserSecretKey.str()));
+                secretTree.setown(getSecret("authn", adminUserSecretKey.str()));
             if (!secretTree)
                 throw MakeStringException(-1, "Error retrieving LDAP Admin username/password");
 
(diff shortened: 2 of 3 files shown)


RussWhitehead:
@afishbeck Please review

system/security/LdapSecurity/ldapconnection.cpp
```diff
@@ -1593,8 +1644,58 @@ class CLdapClient : implements ILdapClient, public CInterface
...
+                        DBGLOG("HPCC Admin user not specified in configuration (hpccAdminSecretKey)");
+                }
+            }
             createdOU = true;
         }
+
+        //Set ECLWatch permissions for HPCC admin group
+        if (!adminGroupName.isEmpty())
+        {
+            //SmcAccess OU attribute is not created until later in
```
afishbeck, line 1684: Seems like common code may be assuming we're in EclWatch here?  Is there a way this can be driven by the Authmap config or something else so no assumptions are made.

If I'm running other services and not EclWatch is this going to happen?

afishbeck (commented):
@RussWhitehead a few questions / comments.

system/security/LdapSecurity/ldapconnection.cpp
```diff
@@ -444,6 +447,32 @@ class CLdapConfig : implements ILdapConfig, public CInterface
...
+        //------------------------------------------------
+        cfg->getProp(".//@hpccAdminSecretKey", adminUserSecretKey.clear());//vault/secrets HPCC Admin username key
+        if (!adminUserSecretKey.isEmpty())
+        {
+            StringBuffer vaultId;
+            cfg->getProp(".//@hpccAdminVaultId", vaultId);//optional HashiCorp vault ID
+
+            DBGLOG("Retrieving optional HPCC Admin username/password from secrets repo: %s %s", !vaultId.isEmpty() ? vaultId.str() : "", adminUserSecretKey.str());
+            Owned<IPropertyTree> secretTree;
+            if (!isEmptyString(vaultId.str()))
```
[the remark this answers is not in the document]
RussWhitehead: Fixed in 2 places

system/security/LdapSecurity/ldapconnection.cpp
```diff
@@ -1561,6 +1600,18 @@ class CLdapClient : implements ILdapClient, public CInterface
...
+
+        //Isolate optional HPCC Admin group name
+        StringBuffer adminGroupName;
+        if (!isEmptyString(m_ldapconfig->getAdminGroupDN()))
+        {
+            const char * p = strchr(m_ldapconfig->getAdminGroupDN(), '=');
+            if (p)
+            {
+                adminGroupName.append(++p);
+                adminGroupName.replace(',', (char)0);
```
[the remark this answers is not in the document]
RussWhitehead: StringBuffer replace correctly sets the buffer length if replacing with a NULL character

StringBuffer & StringBuffer::replace(char oldChar, char newChar)
{
    size_t l = curLen;
    for (size_t i = 0; i < l; i++)
    {
        if (buffer[i] == oldChar)
        {
            buffer[i] = newChar;
            if (newChar == '\0')
            {
                curLen = i;
                break;
            }
        }
    }
    return *this;
}

system/security/LdapSecurity/ldapconnection.cpp
RussWhitehead: Currently, this code to create default OUs has always been called by any LDAP client that includes the OU definitions in the config. So, this has always been done by ECLWatch, Dali, ECLCCServer, ECLRepository and various ESP Services.  To fix his would require fixing the OU creation as well. The proper fix is for each client that uses LDAP resource to ensure the LDAP resources they require are created, instead of trying to brute force all of them here.
That being said, this code is safe in that it won't break anything, it's just an unnecessary LDAP call on startup.  I recommend leaving it as is for now

system/security/LdapSecurity/ldapconnection.cpp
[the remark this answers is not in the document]
afishbeck: cool, I'll have to remember that.

system/security/LdapSecurity/ldapconnection.cpp
afishbeck: Ok, since the primary functionality already exists.

esp/applications/common/ldap/ldap.yaml
```diff
@@ -20,4 +20,6 @@ ldap:
   workunitsBasedn: ou=workunits,ou=ecl
   ldapAdminSecretKey: "ldapadmincredskey"
   ldapAdminVaultId: ""
+  hpccAdminSecretKey: ""
```
afishbeck, line 23: Would it work without the tag being defined?  Are these empty strings just here to be self documenting?

afishbeck: If I added the first one it was probably because I wanted to guarantee the automatically generated esp.xml that is generated from this yaml to be identical to what existed in the past.  But it's probably not necessary for anything new. But as I hinted at, these can help the file/format be self documenting.

afishbeck (approved):
@RussWhitehead One question, but the PR looks fine.

esp/applications/common/ldap/ldap.yaml
RussWhitehead: Yes it would work without the tag. The behaviour in that case would be to not create the HPCC Admin user, forcing the Ops team to create one manually on the Domain Controller as they have in the past. The reason I put them in is, as you stated, to document what is available. To me this is way better than forcing the user to look in the manuals, manually type them into the override file with possible spelling errors.  I prefer to cut/paste/modify, but I am not sure if there is a precedent set by other components

RussWhitehead:
@ghalliday @richardkchapman Please merge

richardkchapman:
Please squash

RussWhitehead:
@richardkchapman squashed, please merge

RussWhitehead:
@richardkchapman Please merge

PR #15626, opened 2022-01-04
HPCC-26916 Re-implement gitfile to use gitlib2 instead of child processes

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/common/remote/hooks/git/gitfile.cpp
+++ b/common/remote/hooks/git/gitfile.cpp
@@ -24,6 +24,9 @@
 #include "jfile.hpp"
 #include "jregexp.hpp"
 #include "gitfile.hpp"
+#include "jlog.hpp"
+
+#include "git2.h"
 
 /*
  * Direct access to files in git repositories, by revision, without needing to check them out first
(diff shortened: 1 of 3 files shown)


ghalliday:
Initial push to check how it compiles on windows...

common/remote/hooks/git/libgit2.cmake
```diff
@@ -0,0 +1,48 @@
...
+# Component: azure
+
+#####################################################
+# Description:
+# ------------
+#    Cmake Input File for azure support libraries
+#####################################################
+
+
+if(1)
```
richardkchapman, line 26: Why?

common/remote/hooks/git/gitfile.cpp
```diff
@@ -410,10 +413,329 @@ class GitRepositoryDirectoryIterator : implements IDirectoryIterator, public CIn
...
+                curFile.setown(new NewGitRepositoryFile(gitFileName, size, false, true, commitTree, &oid));
+        }
+        else
+            curFile.clear();
+    }
+};
+
+//--------------------------------------------------------------------------------------------------------------------
+
+static constexpr bool useLibGit = true;
```
richardkchapman, line 730: I assume this will be removed (and the old code removed) once new is proved to work?

common/remote/hooks/git/gitfile.cpp
```diff
@@ -410,10 +413,329 @@ class GitRepositoryDirectoryIterator : implements IDirectoryIterator, public CIn
...
+        return curFile->isDirectory()==fileBool::foundYes;
+    }
+    virtual __int64 getFileSize()
+    {
+        assertex(curFile);
+        return curFile->size();
+    }
+    virtual bool getModifiedTime(CDateTime &ret)
+    {
+        UNIMPLEMENTED;
```
richardkchapman, line 609: Will it be?

richardkchapman (commented):
Some initial comments. Looks like a very good start

common/remote/hooks/git/gitfile.cpp
```diff
@@ -410,10 +413,329 @@ class GitRepositoryDirectoryIterator : implements IDirectoryIterator, public CIn
     }
 };
 
+
+//--------------------------------------------------------------------------------------------------------------------
+// New implementation using libgit2
+//--------------------------------------------------------------------------------------------------------------------
+
+#define GIT_CHECK(x) check(x, #x)
```
richardkchapman, line 421: style: This appears to be used inconsistently below

common/remote/hooks/git/gitfile.cpp
```diff
@@ -410,10 +413,329 @@ class GitRepositoryDirectoryIterator : implements IDirectoryIterator, public CIn
...
+    git_repository * gitRepo = nullptr;
+    git_commit * gitCommit = nullptr;
+    git_tree * gitRoot = nullptr;
+};
+
+
+class NewGitRepositoryFileIO : implements CSimpleInterfaceOf<IFileIO>
+{
+public:
+    NewGitRepositoryFileIO(const char * gitDirectory, const char * revision, const char * relFileName, GitCommitTree * commitTree, const git_oid * oid)
```
richardkchapman, line 490: First three parameters appear to be unused

common/remote/hooks/git/gitfile.cpp
```diff
@@ -410,10 +413,329 @@ class GitRepositoryDirectoryIterator : implements IDirectoryIterator, public CIn
...
+    unsigned __int64 getStatistic(StatisticKind kind)
+    {
+        //This could be implemented, but not likely to be useful so currently return nothing.
+        return 0;
+    }
+protected:
+    MemoryBuffer buf;
+};
+
+class NewGitRepositoryFile : public GitRepositoryFile
```
richardkchapman, line 546: There is code around line 274 that creates a GitRepositoryFile object directly that may need chaging

richardkchapman (changes requested):
Build errors on Windows and OSX need addressing

common/remote/hooks/git/gitfile.cpp
ghalliday: It wasn't implemented in the previous code - so probably not unless it is needed.

common/remote/hooks/git/gitfile.cpp
ghalliday: removed

common/remote/hooks/git/gitfile.cpp
ghalliday: well spotted, I will change

common/remote/hooks/git/gitfile.cpp
```diff
@@ -546,26 +324,27 @@ class NewGitRepositoryFileIO : implements CSimpleInterfaceOf<IFileIO>
 class NewGitRepositoryFile : public GitRepositoryFile
```
richardkchapman, line 324: What's the purpose of having the base class now?

ghalliday: none really.  I'll combine, and rename (to remove the New prefixes)

richardkchapman (approved):
Looks much cleaner now. Please squash

ghalliday:
@richardkchapman I have squashed (and added some default initialisers onto some class members to clean it up rather than fix an issue).
I think this would be suitable for older branches, but have a reservation.  Is the libgit2 being installed correctly?  Initially it was building but not installing.  I think that is now fixed, but there are no regression tests so not really tested.  Can you think of any good ways to add a regression test?  Should we extend the regression suite to allow links that compile from a url?  Should I add a weird ecl format @..... or <link>...</link> to implement this in the compiler?

PR #15628, opened 2022-01-04
HPCC-26922 Support additional ESDL binding yaml configuration

asselitx:
Recognize Setting properties defined inside the authNZ resource_map and treat
them just as the Feature properties are currently.

Generalize the binding configuration to be specific to a service (aka esp
application type like esdl or eclwatch). Service sections under the
bindingInfo element are only copied into matching service binding configs.

Under the esdl service section we can place a desdlBinding section which
holds configuration specific to dynamically bound services.

Patch a leak of an IException thrown in the CEspCfg constructor.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Cloud-compatibility
  - [x] Premature optimization

## Smoketest:

### DIFF
--- a/esp/platform/application_config.cpp
+++ b/esp/platform/application_config.cpp
@@ -195,6 +195,11 @@ void bindAuthResources(IPropertyTree *legacyAuthenticate, IPropertyTree *app, co
     Owned<IPropertyTreeIterator> features = appAuth->getElements(featuresPath);
     ForEach(*features)
         legacyAuthenticate->addPropTree("Feature", LINK(&features->query()));
+
+    VStringBuffer settingsPath("resource_map/%s/Setting", service);
+    Owned<IPropertyTreeIterator> settings = appAuth->getElements(settingsPath);
+    ForEach(*settings)
+        legacyAuthenticate->addPropTree("Setting", LINK(&settings->query()));
 }
 
 void bindService(IPropertyTree *legacyEsp, IPropertyTree *app, const char *service, const char *protocol, const char *netAddress, unsigned port, const char *bindAuth, int seq)
@@ -214,6 +219,14 @@ void bindService(IPropertyTree *legacyEsp, IPropertyTree *app, const char *servi
         if (authenticate)
             bindAuthResources(authenticate, app, service, auth);
     }
+
+    // bindingInfo has sub-elements named for services like 'esdl' or 'eclwatch'.
+    // The elements under each service section are merged into the bindingEntry
+    // for matching services.
+    xpath.setf("bindingInfo/%s", service);
+    IPTree *branch = app->queryBranch(xpath.str());
+    if (branch)
+        mergePTree(bindingEntry, branch);
 }
 
 static void mergeServicePTree(IPropertyTree *target, IPropertyTree *toMerge)
--- a/esp/services/esdl_svc_engine/esdl_monitor.cpp
+++ b/esp/services/esdl_svc_engine/esdl_monitor.cpp
@@ -518,6 +518,31 @@ private:
                     bindingtree->setProp("@port", port);
                     servicetree->setProp("@name", serviceName.str());
                     servicetree->setProp("@type", "DynamicESDL");
+
+                    // Copy resource_map properties from the desdlBinding.resource_map section
+                    // for this dynamic service into the bindingtree
+                    IPropertyTree* desdlBindingConfig = bindingtree->queryPropTree("desdlBinding");
+                    if (desdlBindingConfig)
+                    {
+                        IPropertyTree* bindingAuthenticate = bindingtree->queryPropTree("Authenticate");
+                        xpath.setf("resource_map/%s", serviceName.str());
+                        IPropertyTree* resourceMap = desdlBindingConfig->queryPropTree(xpath.str());
+                        if (resourceMap && bindingAuthenticate)
+                        {
+                            mergePTree(bindingAuthenticate, resourceMap);
+                            desdlBindingConfig->removeProp(xpath.str());
+                        }
+
+                        // Likewise copy over any additional dynamic service-specific binding configuration
+                        xpath.setf("additionalInfo/%s", serviceName.str());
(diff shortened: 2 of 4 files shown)


esp/platform/application_config.cpp
```diff
@@ -214,6 +219,27 @@ void bindService(IPropertyTree *legacyEsp, IPropertyTree *app, const char *servi
...
+    // Add contents of applicationBindingInfo to static EspBinding element by service name
+    xpath.setf("applicationBindingInfo/%s", service);
+    IPTree *branch = app->queryBranch(xpath.str());
+    if (branch)
+    {
+        Owned<IAttributeIterator> aiter = branch->getAttributes();
+        ForEach (*aiter)
+        {
+            const char *name = aiter->queryName();
+            bindingEntry->addProp(name+1, aiter->queryValue());
```
afishbeck, line 232: Note that this means we can't easily define attributes to be added, but because of the way Yaml ptree works if you make attributes easy to add, then elements are more verbose:

So, either:
```
MyYaml:
  notanattribute: 123
  element: abc
```
or
```
MyYaml:
  attribute: 123
  element: [ abc ]
```
Having to do all scalar elements like that would be cumbersome.  And the need for custom top level binding attributes may be rare.

esp/platform/application_config.cpp
```diff
@@ -214,6 +219,27 @@ void bindService(IPropertyTree *legacyEsp, IPropertyTree *app, const char *servi
...
+        }
+
+        xpath.append("/*");
```
afishbeck, line 235: You already have branch, so don't really need to append to the xpath.
branch->getElements("*"); should be fine.

esp/platform/application_config.cpp
afishbeck: Actually, the real problem is that this will be inconsistent with the ptree added below.  That is, at this top level simple scalars will be elements, but when you do the ```bindingEntry->addPropTree(ele.queryName(), LINK(&ele));``` below they will be attributes.

afishbeck: Will take some playing with to make sure it's usable.  But I would think, just replace this whole section that is merging the attributes and elements with a call to mergePTree(target, toMerge);

If users want simple scalar elements they will have to do.
```
MyYaml:
  attribute: 123
  element: [ abc ]
 ```

File: esp/services/esdl_svc_engine/esdl_monitor.cpp [hunk 3]
```diff
@@ -518,6 +518,45 @@ class CEsdlMonitor : implements IEsdlMonitor, public CInterface, implements IEsd
                     bindingtree->setProp("@port", port);
                     servicetree->setProp("@name", serviceName.str());
                     servicetree->setProp("@type", "DynamicESDL");
+
+                    // Copy properties from the procpt into the bindingtree for this dynamic service
+                    // Identify the secmgr for the binding template
+                    StringBuffer authMethod;
+                    bindingtree->getProp("Authenticate/@method", authMethod);
```
afishbeck, line 525: is there a chance this won't exist?

esp/services/esdl_svc_engine/esdl_monitor.cpp
```diff
@@ -518,6 +518,45 @@ class CEsdlMonitor : implements IEsdlMonitor, public CInterface, implements IEsd
...
+                        if (resourceMap && bindingAuthenticate)
+                        {
+                            mergePTree(bindingAuthenticate, resourceMap);
+                        }
+                    }
+
+                    // Copy over any dynamic service-specific binding configuration
+                    xpath.setf("applicationBindingInfo/%s", serviceName.str());
+                    IPropertyTree* appBindingInfo = procpt->queryPropTree(xpath.str());
+                    if (appBindingInfo)
```
afishbeck, line 543: See above about how this makes the top level Yaml inconsistent with the data added via ```bindingtree->addPropTree(ele.queryName(), LINK(&ele));```

Replacing all of it with a simple call to mergePTree() would be most consistent.  But of course we can discuss any issues that might have.

afishbeck (commented):
@asselitx some comments / questions.

richardkchapman:
Should probably be targeting 8.6.x

esp/services/esdl_svc_engine/esdl_monitor.cpp
```diff
@@ -518,6 +518,31 @@ class CEsdlMonitor : implements IEsdlMonitor, public CInterface, implements IEsd
...
+                        if (resourceMap && bindingAuthenticate)
+                        {
+                            mergePTree(bindingAuthenticate, resourceMap);
+                            desdlBindingConfig->removeProp(xpath.str());
+                        }
+                    }
+
+                    // Likewise copy over any additional dynamic service-specific binding configuration
+                    xpath.setf("additionalInfo/%s", serviceName.str());
+                    IPropertyTree* addtlInfo = desdlBindingConfig->queryPropTree(xpath.str());
```
afishbeck, line 539: I think this block should be inside the 
```
if (desdlBindingConfig)
{
```
otherwise desdlBindingConfig might be null here when you access queryPropTree()?

esp/platform/application_config.cpp
```diff
@@ -214,6 +219,16 @@ void bindService(IPropertyTree *legacyEsp, IPropertyTree *app, const char *servi
...
             bindAuthResources(authenticate, app, service, auth);
     }
+
+    // bindingInfo has sub-elements named for services like 'esdl' or 'eclwatch'.
+    // The elements under each service section are merged into the bindingEntry
+    // for matching services.
+    xpath.setf("bindingInfo/%s", service);
+    IPTree *branch = app->queryBranch(xpath.str());
+    if (branch)
+    {
```
afishbeck, line 229: Trivial but I don't use braces around an if block if there is only one line.

afishbeck (approved):
One trivial comment otherwise looks good.

asselitx:
Squashed and ready for merge.

PR #15629, opened 2022-01-05
HPCC-18963 Avoid critical sections in makeVoidType()/makeNullType() etc.

jpmcmu:
## Type of change:
- [X] This change improves the code (refactor or other change that does not change the functionality)

## Checklist:
  - [X] My code does not create any new warnings from compiler, build system, or lint.
- [X] The commit message is properly formatted and free of typos.
  - [X] The commit message title makes sense in a changelog, by itself.
  - [X] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/common/deftype/deftype.cpp
+++ b/common/deftype/deftype.cpp
@@ -47,15 +47,17 @@ static IAtom * emptyAtom;
 static CriticalSection * typeCS;
 static TypeCache * globalTypeCache;
 
-static CBoolTypeInfo *btt = NULL;
-static CBlobTypeInfo *bltt = NULL;
-static CVoidTypeInfo *vtt = NULL;
-static CNullTypeInfo *ntt = NULL;
-static CRecordTypeInfo *rtt = NULL;
-static CAnyTypeInfo *anytt = NULL;
-static CPatternTypeInfo *patt = NULL;
-static CTokenTypeInfo *tokentt = NULL;
-static CFeatureTypeInfo *featurett = NULL;
+static std::atomic<CBoolTypeInfo*> btt(nullptr);
+static std::atomic<CBlobTypeInfo*> bltt(nullptr);
+static std::atomic<CVoidTypeInfo*> vtt(nullptr);
+static std::atomic<CNullTypeInfo*> ntt(nullptr);
+static std::atomic<CRecordTypeInfo*> rtt(nullptr);
+static std::atomic<CPatternTypeInfo*> patt(nullptr);
+static std::atomic<CTokenTypeInfo*> tokentt(nullptr);
+static std::atomic<CFeatureTypeInfo*> featurett(nullptr);
+static std::atomic<CEventTypeInfo*> ett(nullptr);
+static std::atomic<CAnyTypeInfo*> anytt(nullptr);
+
 static CStringTypeToTypeMap * stt;
 static CStringTypeToTypeMap * datatt;
 static CStringTypeToTypeMap * vstt;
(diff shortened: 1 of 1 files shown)


common/deftype/deftype.cpp
```diff
@@ -47,31 +47,39 @@ static IAtom * emptyAtom;
...
 static CStringTypeToTypeMap * datatt;
 static CStringTypeToTypeMap * vstt;
 static CStringTypeToTypeMap * qstt;
 static CUnicodeTypeToTypeMap * utt;
 static CUnicodeTypeToTypeMap * vutt;
 static CUnicodeTypeToTypeMap * u8tt;
-static CIntTypeInfo *itt[2][8];
-static CSwapIntTypeInfo *sitt[2][8];
+
+// Note: C++ standard does not guarantee std::atomic will initialize memory
```
ghalliday, line 69: This comment is not true for static variables - they are default zero-initialised.  The initialiser list is not required.

common/deftype/deftype.cpp
```diff
@@ -1427,9 +1435,29 @@ bool CModifierTypeInfo::equals(const CTypeInfo & _other) const
     return (baseType == other.baseType) && (kind == other.kind) && (extra == other.extra);
 }
 
-
 //===========================================================================
 
+template<typename T>
+T* getOrCreateSingletonAtomic(std::atomic<T*>& singleton)
```
ghalliday, line 1441: style: I don't like the implicit creation of the object based on the type of the parameter - it is very unclear from the caller what is going to be created.  Better would  be to pass in a factory method similar to querySingleton in jmutex.hpp
note: Is there a reason not to use querySingleton()?

ghalliday, line 1441: naming: getLinkedSingleton() would be a better name.  A singleton by definition is either retrieved or created, so no need for orCreate in the function name.  Atomic is also really redundant.

common/deftype/deftype.cpp
```diff
@@ -1623,12 +1651,23 @@ extern DEFTYPE_API ITypeInfo *makeIntType(int len, bool isSigned)
     if (len <= 0 || len > 8)
         return NULL;
 
-    CriticalBlock procedure(*typeCS);
-    CIntTypeInfo *ret = itt[isSigned][len-1];
-    if (ret==NULL)
-        ret = itt[isSigned][len-1] = new CIntTypeInfo(len, isSigned);
-    ::Link(ret);
-    return ret;
+    std::atomic<CIntTypeInfo*>& singleton = itt[isSigned][len-1];
```
ghalliday, line 1654: reuse: simpler if this also used querySingleton(), or reused the function above.  (One reason why passing in the function to create the object is a good idea).
Same comment goes for all the other functions for objects with non-default constructors.

common/deftype/deftype.cpp
```diff
@@ -1687,121 +1748,111 @@ extern DEFTYPE_API ITypeInfo *makeBitfieldType(int len, ITypeInfo * basetype)
...
 
 extern DEFTYPE_API ITypeInfo *makeBoolType()
 {
-    CriticalBlock procedure(*typeCS);
-    if (!btt)
-        btt = new CBoolTypeInfo();
-    ::Link(btt);
-    return btt;
+    ITypeInfo* type = getOrCreateSingletonAtomic(btt);
+    return type;
```
ghalliday, line 1781: style: why assign to a temporary first?

common/deftype/deftype.cpp
```diff
@@ -1998,6 +2049,16 @@ inline void ReleaseAndClear(T * & ptr)
     }
 }
 
+template <class T>
+inline void ReleaseAndClear(std::atomic<T*>& atomicPtr)
```
ghalliday, line 2053: scope: This does not need to be atomic - it is only called single threaded at process termination.

common/deftype/deftype.cpp
```diff
@@ -1623,12 +1651,23 @@ extern DEFTYPE_API ITypeInfo *makeIntType(int len, bool isSigned)
...
-    if (ret==NULL)
-        ret = itt[isSigned][len-1] = new CIntTypeInfo(len, isSigned);
-    ::Link(ret);
-    return ret;
+    std::atomic<CIntTypeInfo*>& singleton = itt[isSigned][len-1];
+    CIntTypeInfo* val = singleton.load();
+    if (val == nullptr)
+    {
+        CIntTypeInfo* newVal = new CIntTypeInfo(len, isSigned);
+        if (!singleton.compare_exchange_strong(val, newVal))
```
ghalliday, line 1659: readability: Although this code is correct I had to read it several times to check that val was correctly set up on all the paths.
Add a comment should to the true branch, or change to always assign val to singleton.load() after the if.

ghalliday (commented):
Various comments.  I would hope the code would be simpler after this change - it currently is more complicated.

common/deftype/deftype.cpp
jpmcmu: I didn't realize a function like querySingleton already existed I will use that instead.

common/deftype/deftype.cpp
jpmcmu: Agreed, will do.

AttilaVamos:
@jpmcmu I don't think those Regression Test Errors are related to this PR (I saw them today in other PR based on master as well).

AttilaVamos:
@jpmcmu Now it is clean (it was a test system problem).

ghalliday (commented):
@jpmcmu looks good - simpler code and improved semantics.  Please squash.

ghalliday (approved):

jpmcmu:
@ghalliday great, squashed.

PR #15641, opened 2022-01-10
HPCC-26858 ecl-queries-files improvements

afishbeck:
Precache the data for ecl-queries-files command to greatly improve
performance.

Also add a new option to show the files for the entire target (queryset).

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/ecl/eclcmd/queries/ecl-queries.cpp
+++ b/ecl/eclcmd/queries/ecl-queries.cpp
@@ -344,14 +344,44 @@ public:
             fputs("Target must be specified.\n", stderr);
             return false;
         }
-        if (optTarget.isEmpty())
-        {
-            fputs("Query must be specified.\n", stderr);
-            return false;
-        }
         if (!EclCmdCommon::finalizeOptions(globals))
             return false;
         return true;
+    }
+
+    void outputQueryFiles(const char *id, IArrayOf<IConstFileUsedByQuery> &files, IArrayOf<IConstQuerySuperFile> &superfiles)
+    {
+        fputs("------------------\n", stdout);
+        if (!isEmptyString(id))
+            fprintf(stdout, "Query: %s\n", id);
+        if (!files.length())
+            fputs("No files used.\n", stdout);
+        else
+            fputs("Files used:\n", stdout);
+        ForEachItemIn(i, files)
+        {
+            IConstFileUsedByQuery &file = files.item(i);
+            StringBuffer line("  ");
+            line.append(file.getFileName()).append(", ");
+            line.append(file.getFileSize()).append(" bytes, ");
+            line.append(file.getNumberOfParts()).append(" part(s)\n");
+            fputs(line, stdout);
+        }
+        fputs("\n", stdout);
+
+        if (superfiles.length())
+        {
+            fputs("SuperFiles used:\n", stdout);
+            ForEachItemIn(sp, superfiles)
+            {
+                IConstQuerySuperFile &superfile = superfiles.item(sp);
+                fprintf(stdout, "    %s\n", superfile.getName());
+                StringArray &subfiles = superfile.getSubFiles();
+                ForEachItemIn(sb, subfiles)
+                    fprintf(stdout, "    > %s\n", subfiles.item(sb));
+            }
+            fputs("\n", stdout);
+        }
     }
 
     virtual int processCMD()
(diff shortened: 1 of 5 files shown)


AttilaVamos:
@afishbeck I think they are not related to your changes. Test rescheduled.

ecl/eclcmd/queries/ecl-queries.cpp
```diff
@@ -368,34 +398,16 @@ class EclCmdQueryFiles : public EclCmdCommon
...
                 {
-                    IConstQuerySuperFile &superfile = superfiles.item(sp);
-                    fprintf(stdout, "  %s\n", superfile.getName());
-                    StringArray &subfiles = superfile.getSubFiles();
-                    ForEachItemIn(sb, subfiles)
-                        fprintf(stdout, "    > %s\n", subfiles.item(sb));
+                    IConstQueryFilesUsed &query = resp->getQueries().item(i);
+                    outputQueryFiles(query.getQueryId(), query.getFiles(), query.getSuperFiles());
                 }
-                fputs("\n", stdout);
```
wangkx, line 398: question: is it possible that the optQuery is not set and the resp->getQueries() returns empty? If yes, should we output a message similar to "No files used.\n" at that situation?

esp/services/ws_workunits/ws_workunitsQuerySets.cpp
```diff
@@ -1895,8 +1875,74 @@ bool CWsWorkunitsEx::onWUQueryFiles(IEspContext &context, IEspWUQueryFilesReques
...
+        queryid.set(registeredQuery->queryProp("@id"));
+    }
+
+    Owned<IPropertyTree> tree = filesInUse.getTree();
+    if (queryid.length())
+    {
+        VStringBuffer xpath("%s/Query[@id='%s']", target, queryid.get());
+        IPropertyTree *queryTree = tree->queryPropTree(xpath);
+        if (!queryTree)
+        throw MakeStringException(ECLWATCH_QUERYID_NOT_FOUND, "Query not found in file cache (%s)", xpath.str());
```
wangkx, line 1927: indent here?

ecl/eclcmd/queries/ecl-queries.cpp
```diff
@@ -368,34 +398,16 @@ class EclCmdQueryFiles : public EclCmdCommon
         if (ret == 0)
         {
             IArrayOf<IConstFileUsedByQuery> &files = resp->getFiles();
-            if (!files.length())
-                fputs("No files used.\n", stdout);
+            IArrayOf<IConstQueryFilesUsed> &queries = resp->getQueries();
```
wangkx, line 401: question: how about moving this line before line 406 and use the queries in line 406 and 408? Or remove this line?

wangkx (commented):
@afishbeck 3 questions

ghalliday (commented):
@afishbeck looks good - a couple of minor comments.

A couple of other comments...  It would be useful to have a comment on getTree() inidcating it is thread safe because it creates a new tree if the tree had changed.  (I double checked because it wasn't obvious).

Also any ideas why the operation is so slow?  @jakesmith is the following line in referencedfilelist.cpp(685) inefficient?:

    Owned<IPropertyTreeIterator> iter = xgmml->getElements("//node[att/@name='_*ileName']");

would it be better to remove the filter (since it is checked in the loop, how efficient is the wildcarding) - although that will significanly increase the side of the target array.  Is it better to iterate each level in turn (I doubt it)?

@afishbeck looking at the code in ReferencedFileList::addFilesFromQuery, it looks like the code will still do lots of work - even if the file has already been resolved from another activity in the same package/  I suspect this is a fairly common occurrence.

jakesmith:
> 
> Also any ideas why the operation is so slow? @jakesmith is the following line in referencedfilelist.cpp(685) inefficient?:
> 
> ```
> Owned<IPropertyTreeIterator> iter = xgmml->getElements("//node[att/@name='_*ileName']");
> ```
> 
> would it be better to remove the filter (since it is checked in the loop, how efficient is the wildcarding) - although that will significanly increase the side of the target array. Is it better to iterate each level in turn (I doubt it)?
> 

not sure how slow it is.
One observation, it looks like getGraphs tries to pre-load/cache the graphs, but doesn't fully load them, meaning they will be fetched from SDS layer by layer on demand. If so, it could help if a queryBranch() was used to ensure the whole graph was loaded in one go.

The wildcards will be preventing the fast cached qualifier maps being used, so linear scans for matches of all nodes at each level, searching for a match will occur (as opposed to single HT lookup at each level).
So I do wonder whether 2 passes, 1 searching for '_fileName' matches, followed by a 2nd searching for '_indexFileName' might be faster.

Is this actually the area of slowness vs the code using the results of this iterator?

ghalliday:
@jakesmith I don't actually know what is causing this code to be slow - I was just reviewing the code to see if anything jumped out.  It would need testing and ideally profiling to find the cause.
I don't think your qualifier maps will help here (will it?) - because the filter is on existence of children.

jakesmith:
> I don't think your qualifier maps will help here (will it?) - because the filter is on existence of children.

you're right, won't apply in this case, only for attribute=value cases.

ecl/eclcmd/queries/ecl-queries.cpp
afishbeck: Sure, I will add a message.

afishbeck:
@richardkchapman @ghalliday updated.  With changes relevant to this PR.  

Any performance improvements can come later.. but I don't think it's that much slower than I would expect.  The perception of slowness was because it rebuilds the entire cache when dirty, not just for a given query.. and odds are this is often called right after dirtying the cache.

The code doesn't do any DALI lookups until it has a deduped list across the entire query.  There may still be some improvements I can make caching some information across multiple queries.

wangkx (approved):
The changes look fine to me.

ghalliday (approved):
Looks good, please squash

afishbeck:
@ghalliday squashed

PR #15664, opened 2022-01-15
HPCC-26809 Validate WU file request using WU XML data

wangkx:
The existing file path check does not work when WU files (log, cpp,
etc) are stored using symlinks and the requested file paths are
source file paths. The source file path is stored in WU's XML. In
this fix, a check to source file path is added using the file path
stored in WU's XML.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/services/ws_workunits/ws_workunitsHelpers.cpp
+++ b/esp/services/ws_workunits/ws_workunitsHelpers.cpp
@@ -2245,6 +2245,46 @@ void WsWuInfo::getWUProcessLogSpecs(const char* processName, const char* logSpec
         logSpecs.sortAscii(false); //Sort the logSpecs from old to new
 }
 
+bool WsWuInfo::validateWUProcessLog(const char* file, bool eclAgent)
+{
+    Owned<IStringIterator> logs = cw->getLogs(eclAgent ? "EclAgent" : "Thor");
+    ForEach (*logs)
+    {
+        SCMStringBuffer logName;
+        logs->str(logName);
+        if (logName.length() < 1)
+            continue;
+
+        if (strieq(file, logName.str()))
+            return true;
+    }
+    return false;
+}
+
+bool WsWuInfo::validateWUAssociatedFile(const char* file, WUFileType type)
+{
+    Owned<IConstWUQuery> query = cw->getQuery();
+    if (!query)
+        return false;
+
+    Owned<IConstWUAssociatedFileIterator> iter = &query->getAssociatedFiles();
+    ForEach(*iter)
+    {
+        IConstWUAssociatedFile & cur = iter->query();
+        if (cur.getType() != type)
+            continue;
+
+        SCMStringBuffer name;
+        cur.getName(name);
+        if (name.length() < 1)
+            continue;
+
+        if (strieq(file, name.str()))
+            return true;
+    }
+    return false;
+}
+
 void WsWuInfo::getWorkunitResTxt(MemoryBuffer& buf)
 {
     Owned<IConstWUQuery> query = cw->getQuery();
@@ -4011,8 +4051,11 @@ IFileIOStream* CWsWuFileHelper::createWUFileIOStream(IEspContext& context, const
     return createIOStreamWithFileName(zipFileNameWithPath.str(), IFOread);
 }
 
-void CWsWuFileHelper::validateFilePath(const char *file, bool UNCFileName, const char *fileType, const char *compType, const char *compName)
+void CWsWuFileHelper::validateFilePath(const char* file, WsWuInfo& winfo, CWUFileType wuFileType, bool UNCFileName, const char* fileType, const char* compType, const char* compName)
 {
+    if (validateWUFile(file, winfo, wuFileType))
+        return;
+
     StringBuffer actualPath;
     if (UNCFileName)
         splitUNCFilename(file, nullptr, &actualPath, nullptr, nullptr);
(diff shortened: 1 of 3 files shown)


jakesmith:
@AttilaVamos - any idea what caused the regression suite failure? - they don't look like they're related to the changes in this PR.

jakesmith:
@wangkx - I haven't been through the changes in-depth, but am wondering if this covers all cases, and if it even can, e.g. validateFilePath is also called from ws_topology.

Would a solution and simpler alternative be to ensure that the 'realpath' is validated against the configuration directories? i.e. both the path being looked up and the configuration directory being checked are converted to their real paths, before validating that the file being referenced is indeed part of the correct configuration directory.

AttilaVamos:
@wangkx @jakesmith I don't think those errors are related to this PR. They are appear in other PRs as well. It seems something not right in that Smoketest environment, but unfortunately I don't know yet, what is it.

wangkx:
@jakesmith 
> I haven't been through the changes in-depth, but am wondering if this covers all cases, and if it even can, e.g. validateFilePath is also called from ws_topology.
Most likely, the calls are from ECLWatch. From the existing ECLWatch, the paths in the ws_topology calls are the same as configuration directory, not 'realpath (ex. /mnt/disk1/var/log/...)'. But, we may need to check the 'realpath' in ws_topology because the calls may be not from ECLWatch and the paths being looked upare 'realpath'.

> Would a solution and simpler alternative be to ensure that the 'realpath' is validated against the configuration directories? i.e. both the path being looked up and the configuration directory being checked are converted to their real paths, before validating that the file being referenced is indeed part of the correct configuration directory.
How to convert the path to their real paths (ex. /mnt/disk1/var/log/...)?

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -4011,8 +4051,11 @@ IFileIOStream* CWsWuFileHelper::createWUFileIOStream(IEspContext& context, const
     return createIOStreamWithFileName(zipFileNameWithPath.str(), IFOread);
 }
 
-void CWsWuFileHelper::validateFilePath(const char *file, bool UNCFileName, const char *fileType, const char *compType, const char *compName)
+void CWsWuFileHelper::validateFilePath(const char* file, WsWuInfo& winfo, CWUFileType wuFileType, bool UNCFileName, const char* fileType, const char* compType, const char* compName)
 {
+    if (validateWUFile(file, winfo, wuFileType))
```
afishbeck, line 4056: Is there ever a valid case where this should returns false and the user is accessing a file they should?  That is, the file they ask for should have been listed in the workunit, correct?

Just want clarification just in case.  I'm wondering if a return of false above should ever be allowed to continue below?  Doesn't that almost imply something is wrong and they may be trying to access files in a way that they shouldn't?  A file outside the scope of this workunit?

If there are some files that need to be accessed that aren't found in validateWUFile() then ok we just need to control where they can access.  but otherwise, I think the narrower the scope of access the better.

afishbeck: @jakesmith @wangkx I think Jake and I had similar thoughts but different conclusions. His suggestion of validating the real path does solve the immediate problem you were having.

But I do think validating onWUFile requests (not other ways of getting files) to make sure you have access to the workunit and the file you are asking for is referenced in the workunit is a stronger form of security checking in the end. 

A user can't fish for files related to another workunit.. or other files.

Perhaps we can fix the realpath checking now and consider restricting further in the future.

wangkx: Yes, we can fix the realpath checking now and consider restricting further in the future. @jakesmith any comment? Could you please let me know how to convert the path (such as /var/log/...) to their real paths (such as /mnt/disk1/var/log/...)?

jakesmith:
> How to convert the path to their real paths (ex. /mnt/disk1/var/log/...)?

I think jlib's makeAbsolutePath is suitable.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
jakesmith: > But I do think validating onWUFile requests (not other ways of getting files) to make sure you have access to the workunit and the file you are asking for is referenced in the workunit is a stronger form of security checking in the end.
> 
> A user can't fish for files related to another workunit.. or other files.

I agree, it makes sense to limit valid requests to files only within the workunit they are being request for.
So I think this change makes sense, even if there are other issues (like potential out of scope files via topology), that might need tackling differently, e.g. with help from realpath.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -2245,6 +2245,46 @@ void WsWuInfo::getWUProcessLogSpecs(const char* processName, const char* logSpec
...
+}
+
+bool WsWuInfo::validateWUAssociatedFile(const char* file, WUFileType type)
+{
+    Owned<IConstWUQuery> query = cw->getQuery();
+    if (!query)
+        return false;
+
+    Owned<IConstWUAssociatedFileIterator> iter = &query->getAssociatedFiles();
+    ForEach(*iter)
```
jakesmith, line 2271: future/efficiency: if there are a lot of associated files, and a lot of calls to validateWUAssociatedFile, this linear scanning could be a bit expensive. It's probably 'ok' because esp will be reopening the workunit in general frequently, so caching/a more efficient mechanism won't be available.

jakesmith (approved):
@wangkx - some minor comments. I can't see any problems with what has been implemented, and as @afishbeck says, it reduces the scope to access files in valid system paths, but that are not part of the wuid.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -2245,6 +2245,46 @@ void WsWuInfo::getWUProcessLogSpecs(const char* processName, const char* logSpec
...
 }
 
+bool WsWuInfo::validateWUProcessLog(const char* file, bool eclAgent)
+{
+    Owned<IStringIterator> logs = cw->getLogs(eclAgent ? "EclAgent" : "Thor");
+    ForEach (*logs)
+    {
+        SCMStringBuffer logName;
+        logs->str(logName);
+        if (logName.length() < 1)
```
ghalliday, line 2255: question: is it worth checking this at all - won't it fail the strieq test anyway?

ghalliday (approved):
Looks ok.  Some minor concerns about efficiency, but I don't think they will be significant.

PR #15683, opened 2022-01-21
HPCC-26845 Fix problems with node_modules directories on windows

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/ecl/hql/hqlcollect.cpp
+++ b/ecl/hql/hqlcollect.cpp
@@ -517,9 +517,12 @@ void FileSystemDirectory::processDependencies(IDirectoryIterator * dir)
         const char * tail = pathTail(filename);
         if (file.isDirectory() == fileBool::foundYes)
         {
-            //If there is a directory within node_module, then treat that directory as a package
-            PackageDependency * depend = new PackageDependency(createIdAtom(tail), filename, false);
-            contents.append(*depend);
+            if (isValidIdentifier(tail))
+            {
+                //If there is a directory within node_module, then treat that directory as a package
+                PackageDependency * depend = new PackageDependency(createIdAtom(tail), filename, false);
+                contents.append(*depend);
+            }
         }
         else if (file.isFile() == fileBool::foundYes)
         {
--- a/ecl/hql/hqlrepository.cpp
+++ b/ecl/hql/hqlrepository.cpp
@@ -36,6 +36,11 @@ static const char * queryExtractFilename(const char * urn)
     case '/':
         return urn;
     }
+#ifdef _WIN32
+    //Check for drive:....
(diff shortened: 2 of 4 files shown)


ghalliday:
This is actually 3 different fixes - I will add comments on each

@GordonSmith FYI

ecl/hql/hqlcollect.cpp
```diff
@@ -517,9 +517,12 @@ void FileSystemDirectory::processDependencies(IDirectoryIterator * dir)
         const char * tail = pathTail(filename);
         if (file.isDirectory() == fileBool::foundYes)
         {
-            //If there is a directory within node_module, then treat that directory as a package
-            PackageDependency * depend = new PackageDependency(createIdAtom(tail), filename, false);
-            contents.append(*depend);
+            if (isValidIdentifier(tail))
```
ghalliday, line 520: This is an optimization - ignore any directories which cannot be modules (e.g. .bin, @xxx)

ecl/hql/hqlrepository.cpp
```diff
@@ -36,6 +36,11 @@ static const char * queryExtractFilename(const char * urn)
     case '/':
         return urn;
     }
+#ifdef _WIN32
```
ghalliday, line 39: This is the fix for the issue.  It wasn't spotting c:\ as an absolute filename.

system/jlib/jdebug.cpp
```diff
@@ -1096,7 +1096,11 @@ void getCpuInfo(unsigned &numCPUs, unsigned &CPUSpeed)
 static unsigned evalAffinityCpus()
 {
     unsigned numCpus = 0;
+#ifdef __64BIT__
+    DWORD64 ProcessAffinityMask, SystemAffinityMask;
```
ghalliday, line 1100: This is causing a memory corruption in 64bit - which was causing a very strange crash on exit (and probably a long pause on release builds).  Should possibly be backported even further.

ghalliday:
@shamser @richardkchapman please take a look.

system/jlib/jdebug.cpp
richardkchapman: If you want it backported further (and probably even if you don't) then it should be in a separate JIRA/PR

richardkchapman (changes requested):
Should be split in two

shamser (commented):
@ghalliday Other than issues reported by Richard, everything else looks good.

richardkchapman (approved):

PR #15692, opened 2022-01-26
HPCC-26759 New ESP Service to create HPCC LDAP environment

RussWhitehead:
…ments

Add new stand alone ESP service to assist standing up new HPCC LDAP environments.
Options allow admin user to request creation of a private OU subtree, share
an OU with existing clusters (file scopes, users, etc), or to enter their own
hardcode OU string. User can select whether the service just generates the OU
strings in HELM format, or if it also calls out to LDAP to create the OU
structure

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [x] Security
  - [ ] Thread-safety
  - [x] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:


RussWhitehead:
@afishbeck Please review

RussWhitehead:
@afishbeck This is the HELM configuration I used to test with

- name: ldapenvironment
  application: ldapenvironment
  hpccRootName: hpcc,dc=russ,dc=myldap,dc=com 
  sharedFilesBaseDN: ou=files,ou=hpcc,dc=russ,dc=myldap,dc=com
  sharedGroupsBaseDN: ou=groups,ou=hpcc,dc=russ,dc=myldap,dc=com
  sharedUsersBaseDN: ou=users,ou=hpcc,dc=russ,dc=myldap,dc=com
  sharedResourcesBaseDN: ou=espservices,ou=hpcc,dc=russ,dc=myldap,dc=com
  sharedWorkunitsBaseDN: ou=workunits,ou=hpcc,dc=russ,dc=myldap,dc=com
  adminGroupName: HPCCAdmins
  auth: ldap
  ldap:
    ldapAddress: "192.168.56.101"
    serverType: 389DirectoryServer
    ldapProtocol: ldap
    ldapAdminSecretKey: myldapcreds
    hpccAdminSecretKey: myhpcccreds
    resourcesBasedn: ou=smc,ou=espservices,ou=ecl

esp/services/ldapenvironment/ldapenvironmentService.cpp
```diff
@@ -0,0 +1,204 @@
...
+#ifdef _DEBUG
+        StringBuffer sb;
+        toXML(cfg, sb);
+        if (!sb.isEmpty())
+            DBGLOG("\n------ldapenvironment config\n%s\n------\n", sb.str());
+#endif
+        VStringBuffer prefix("Software/EspProcess[@name='%s']/@", _process);
+        StringBuffer xpath;
+
+        xpath.clear().appendf("%shpccRootName", prefix.str());
```
afishbeck, line 39: Not a big deal, but I would find it a bit easier to read:
```
xpath.clear().append(prefix).append("hpccRootName");
or
xpath.set(prefix).append("hpccRootName");
```

esp/services/ldapenvironment/ldapenvironmentService.cpp
```diff
@@ -0,0 +1,204 @@
...
+bool CldapenvironmentEx::onLDAPQueryDefaults(IEspContext &context, IEspLDAPQueryDefaultsRequest &req, IEspLDAPQueryDefaultsResponse &resp)
+{
+    resp.setHPCCRootName(ldapRootOU.str());
+    resp.setAdminGroupName(adminGroupName.str());
+    resp.setSharedFilesBaseDN(sharedFilesBaseDN.str());
+    resp.setSharedGroupsBaseDN(sharedGroupsBaseDN.str());
+    resp.setSharedUsersBaseDN(sharedUsersBaseDN.str());
+    resp.setSharedResourcesBaseDN(sharedResourcesBaseDN.str());
+    resp.setSharedWorkunitsBaseDN(sharedWorkunitsBaseDN.str());
+    resp.setRetcode(0);
```
afishbeck, line 90: What is the purpose of the return code - always zero.  Unless you have some intended values, an fatal error can be specified by throwing an exception.

esp/scm/ldapenvironment.ecm
```diff
@@ -0,0 +1,102 @@
...
+    string  sharedUsersBaseDN;
+    string  sharedResourcesBaseDN;
+    string  sharedWorkunitsBaseDN;
+};
+
+/*##############################################################################*/
+
+ESPenum OUMode : int
+{
+    GeneratePrivate(0, "Generate private OU branch, using LDAP Root and Environment name (Recommended)"),
```
afishbeck, line 38: I wonder if "stand alone" or "independent" would be a better term than "private"?  Private just sounds like it could mean something else security wise.

esp/services/ldapenvironment/ldapenvironmentService.cpp
```diff
@@ -0,0 +1,204 @@
...
+    return ou.str();
+}
+
+bool CldapenvironmentEx::onLDAPCreateEnvironment(IEspContext &context, IEspLDAPCreateEnvironmentRequest &req, IEspLDAPCreateEnvironmentResponse &resp)
+{
+    try
+    {
+        if (secmgr == nullptr)
+        {
+            secmgr = dynamic_cast<CLdapSecManager*>(context.querySecManager());
```
afishbeck, line 117: secmgr will need to be protected to make it thread-safe.

esp/services/ldapenvironment/ldapenvironmentService.cpp
```diff
@@ -0,0 +1,204 @@
...
+                               "    hpccAdminSecretKey: TODO!  Create Kubernetes Secret for %s\n"
+                               "    filesBasedn: %s\n"
+                               "    groupsBasedn: %s\n"
+                               "    usersBasedn: %s\n"
+                               "    resourcesBasedn: %s\n"
+                               "    workunitsBasedn: %s\n",
+                               adminGroupName.str(), respAdminUser.str(), respFilesBaseDN.str(), respGroupsBaseDN.str(),
+                               respUsersBaseDN.str(), respResourcesBaseDN.str(), respWorkunitsBaseDN.str());
+        resp.setLDAPHelm(ldapHelm.str());
+        resp.setRetcode(0);
```
afishbeck, line 197: You can always add the retcode field later when you are actually going to set it to something other than 0.

esp/services/ldapenvironment/ldapenvironmentService.cpp
```diff
@@ -0,0 +1,204 @@
...
+    resp.setSharedWorkunitsBaseDN(sharedWorkunitsBaseDN.str());
+    resp.setRetcode(0);
+    return true;
+}
+
+const char * CldapenvironmentEx::formatOUname(StringBuffer &ou, const char * envName, int mode, const char * sharedOU, const char * reqBaseDN, const char * privateOU)
+{
+    switch (mode)
+    {
+        case 0: // Generated Private OU
```
afishbeck, line 98: Use the values generated from the ESDL in the ECM file:

case COUMode_GeneratePrivate:
case COUMode_UseGlobal:
case COUMode_CreateCustom:

The ESDL creates a c++ enum:

`enum COUMode { OUMode_Undefined=-1,COUMode_GeneratePrivate=0, COUMode_UseGlobal=1, COUMode_CreateCustom=2, };
`

esp/services/ldapenvironment/ldapenvironmentService.cpp
```diff
@@ -0,0 +1,204 @@
...
+            throw MakeStringException(-1, "Environment name must be specified");
+            if (strchr(req.getEnvName(), ' '))
+                throw MakeStringException(-1, "Environment name cannot contain spaces or special characters");
+        }
+        if (isEmptyString(req.getEnvOwnerName()))
+            throw MakeStringException(-1, "Environment Owner name must be specified");
+        if (isEmptyString(req.getEnvDescription()))
+            throw MakeStringException(-1, "Environment Description must be specified");
+
+        if (req.getFilesMode()==2  && isEmptyString(req.getCustomFilesBaseDN()))
```
afishbeck, line 135: if (req.getFilesMode()==COUMode_CreateCustom

esp/services/ldapenvironment/ldapenvironmentService.cpp
```diff
@@ -0,0 +1,204 @@
...
+        
+        //----------------------------------
+        // Set response
+        //----------------------------------
+        resp.setHPCCAdminUsername(respAdminUser.str());
+
+        VStringBuffer ldapHelm("\n"
+                               "  ldap:\n"
+                               "    adminGroupName: %s\n"
+                               "    hpccAdminSecretKey: TODO!  Create Kubernetes Secret for %s\n"
```
afishbeck, line 188: Is creating the secret necessary for the respAdminUser to work at all?

esp/services/ldapenvironment/ldapenvironmentService.cpp
```diff
@@ -0,0 +1,204 @@
...
+    resp.setRetcode(0);
+    return true;
+}
+
+const char * CldapenvironmentEx::formatOUname(StringBuffer &ou, const char * envName, int mode, const char * sharedOU, const char * reqBaseDN, const char * privateOU)
+{
+    switch (mode)
+    {
+        case 0: // Generated Private OU
+            ou.appendf("%s,ou=%s,ou=%s", privateOU, envName, ldapRootOU.str());
```
afishbeck, line 99: I'm curious why the ldapRoutOU doesn't already start with "ou="?

helm/hpcc/values.yaml
```diff
@@ -534,6 +534,23 @@ esp:
   #resources:
   #  cpu: "250m"
   #  memory: "1G"
+- name: ldapenvironment
+  #ldapenvironment is a stand alone ESP service used to help stand up new HPCC LDAP Environments
+  application: ldapenvironment
+  auth: ldap
+  #specify the hpccRootName, without leading "ou="
+  hpccRootName: hpcc,dc=myldap,dc=com
```
afishbeck, line 542: Again, curious why hpccRootName doesn't already start with ou=?

esp/services/ldapenvironment/ldapenvironmentService.cpp
```diff
@@ -0,0 +1,204 @@
...
+        formatOUname(respWorkunitsBaseDN, req.getEnvName(), req.getWorkunitsMode(), sharedWorkunitsBaseDN.str(), req.getCustomWorkunitsBaseDN(), "ou=workunits");
+
+        // Create HPCC Admin Username
+
+        VStringBuffer respAdminUser("admin_%s", req.getEnvName());
+
+
+        //Create LDAP resources OU Hierarchy
+
+        if (req.getCreateLDAPEnvironment())
```
afishbeck, line 162: Is setting this to false just a way to get the ldapHelm even after it's been created?

esp/services/ldapenvironment/ldapenvironmentService.cpp
```diff
@@ -0,0 +1,204 @@
...
+        {
+            VStringBuffer description("%s, created by %s", req.getEnvDescription(), req.getEnvOwnerName());
```
afishbeck, line 164: Hopefully these will fail if the environment already exists.  People shouldn't think they were successful in creating an environment if it already existed.

esp/services/ldapenvironment/ldapenvironmentService.cpp
```diff
@@ -0,0 +1,204 @@
...
+        StringBuffer respFilesBaseDN, respGroupsBaseDN, respUsersBaseDN, respResourcesBaseDN, respWorkunitsBaseDN;
+        formatOUname(respFilesBaseDN, req.getEnvName(), req.getFilesMode(), sharedFilesBaseDN.str(), req.getCustomFilesBaseDN(), "ou=files");
+        formatOUname(respGroupsBaseDN, req.getEnvName(), req.getGroupsMode(), sharedGroupsBaseDN.str(), req.getCustomGroupsBaseDN(), "ou=groups");
+        formatOUname(respUsersBaseDN, req.getEnvName(), req.getUsersMode(), sharedUsersBaseDN.str(), req.getCustomUsersBaseDN(), "ou=users");
+        formatOUname(respResourcesBaseDN, req.getEnvName(), req.getResourcesMode(), sharedResourcesBaseDN.str(), req.getCustomResourcesBaseDN(), "ou=smc,ou=espservices");
+        formatOUname(respWorkunitsBaseDN, req.getEnvName(), req.getWorkunitsMode(), sharedWorkunitsBaseDN.str(), req.getCustomWorkunitsBaseDN(), "ou=workunits");
+
+        // Create HPCC Admin Username
+
+        VStringBuffer respAdminUser("admin_%s", req.getEnvName());
```
afishbeck, line 157: Is admin_<env> user created?  or is that a TBD?

When the env admin is created I think it should be automatically be given rights to manager private/stand alone ous, but not global/shared ones.. operations should have to grant that.

afishbeck (commented):
@RussWhitehead some comments and questions, but I like the way it works over all.

esp/services/ldapenvironment/ldapenvironmentService.cpp
RussWhitehead: done

esp/services/ldapenvironment/ldapenvironmentService.cpp
RussWhitehead: removed retcode

esp/scm/ldapenvironment.ecm
RussWhitehead: Changed wording to "stand alone"

esp/services/ldapenvironment/ldapenvironmentService.cpp
RussWhitehead: I'm not sure this is the case.  LDAPEnvironment will have it's own instance of the security manager and is the only process on the pod using that instance.  Please elaborate

esp/services/ldapenvironment/ldapenvironmentService.cpp
```diff
@@ -0,0 +1,204 @@
...
+            if (secmgr == nullptr)
+                throw MakeStringException(ECLWATCH_INVALID_SEC_MANAGER, MSG_SEC_MANAGER_IS_NULL);
+        }
+        
+        //Verify request
+
+        if (isEmptyString(req.getEnvName()))
+        {
+            throw MakeStringException(-1, "Environment name must be specified");
+            if (strchr(req.getEnvName(), ' '))
```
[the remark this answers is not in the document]
RussWhitehead: Done

esp/services/ldapenvironment/ldapenvironmentService.cpp
RussWhitehead: Retcode remove from both methods

esp/services/ldapenvironment/ldapenvironmentService.cpp
RussWhitehead: Cool, added to various places in my code

esp/services/ldapenvironment/ldapenvironmentService.cpp
RussWhitehead: Done

esp/services/ldapenvironment/ldapenvironmentService.cpp
RussWhitehead: I cannot create the Kubernetes secret, but it is required for this user to work. Either way the admin will have to create it and specify it in the Helm chart "Secrets: authn: myhpccadmin" and in the ldapenvironment config. Since this is the case, I defer creation of a password to the admin. This "TODO!" is meant to alert the admin they need to do something, perhaps there is a better way?

esp/services/ldapenvironment/ldapenvironmentService.cpp
RussWhitehead: Good question, and this comes from the legacy mess we have with the current LDAP.  I could allow the user to enter it, and then just strip it off (if it's specified) before passing to LDAP if you think this will avoid headaches in the future. Probably should be allowed, let me know what you  think

helm/hpcc/values.yaml
RussWhitehead: LDAPSecurityManager anomaly, see proposal above to allow admin to optionally specify it

esp/services/ldapenvironment/ldapenvironmentService.cpp
RussWhitehead: I could argue this either way. What happens if the create the environment, then realize that, say, the shared ou=files should be  really be stand alone (not shared).  They need to be able to rerun the creation in order to make this happen. Thoughts?

esp/services/ldapenvironment/ldapenvironmentService.cpp
RussWhitehead: The HPCCAdmins group, and the HPCCAdmin user are created by ESP when it starts, and that user is then granted SMC Access (and nothing more). With SMC Access, they can then add/modify the powers that user/group has.  If we want to change that, I think a separate Jira will be needed,, let me know

RussWhitehead:
@afishbeck Please re-review based on your comments
Note I had a few questions in my response
Also, please advise why my service is connecting to Dali?   I can't seem to make sense of that

esp/services/ldapenvironment/ldapenvironmentService.cpp
afishbeck: I just mean that onLDAPCreateEnvironment() could be called multiple times at once, so you could have two threads in this block of code at the same time so secmgr should be protected.

Another option that would only happen once at startup might be to have your binding class  (CldapenvironmentSoapBindingEx) override the addService() method.  The binding object inherits a pointer to the secmgr so you can tie everything together there. 

Something like:
```

class CldapenvironmentEx : public Cldapenvironment  //base class name built from ECM ESPservice name
{
public:
   void setSecMgr( ISecManager*sm){secmgr = dynamic_cast<CLdapSecManager*>(sm);}
}

class CldapenvironmentSoapBindingEx : public CldapenvironmentSoapBinding  
{
public:
    void addService(const char * name, const char * host, unsigned short port, IEspService & service) override
    {
        CldapenvironmentEx *srv = dynamic_cast<CldapenvironmentEx*>(&service);
        srv->setSecMgr(querySecManager());  //call EspHttpBinding::querySecManager();
        CEspBinding::addService(name, host, port, service);
    }
};
```

esp/services/ldapenvironment/ldapenvironmentService.cpp
afishbeck: ok.

esp/services/ldapenvironment/ldapenvironmentService.cpp
afishbeck: I would prefer to have a separate update function.  I guess I want the create and update to be able to evolve separately if they have to.  And I want to prevent accidental changes.

esp/services/ldapenvironment/ldapenvironmentService.cpp
afishbeck: I thought maybe you could provide an option to create the secret if it doesn't exist.  You should be able to do that by calling the kubectl command line tool.  Just like you do for your secrets testing.  We'll have to make sure your pod has the right permissions.  EclWatch can call kubectl for example.

helm/hpcc/values.yaml
afishbeck: ok, if it's an expected format from the secmgr.

esp/services/ldapenvironment/ldapenvironmentService.cpp
```diff
@@ -0,0 +1,202 @@
...
+#endif
+        VStringBuffer prefix("Software/EspProcess[@name='%s']/@", _process);
+        StringBuffer xpath;
+
+        xpath.set(prefix).append("hpccRootName");
+        cfg->getProp(xpath.str(), ldapRootOU);
+        if (ldapRootOU.isEmpty())
+            throw MakeStringException(-1, "hpccRootName must be specified in configuration (ex. 'hpcc,dc=myLDAPDomain,dc=com')");
+
+        xpath.clear().appendf("%sadminGroupName", prefix.str());
```
afishbeck, line 44: Sorry, I meant you could use the xpath.set(prefix).append("hpccRootName"); form for all of these cases.

afishbeck:
We're working on an option to not always have to deploy DALI for an environment.  The current helm chart automatically sets the --daliServers option.  Part of the no dali change will have to null out the daliServers option.

The value is set here in helm/hpcc/templates/esp.yaml the line:
                {{ include "hpcc.daliArg" $ }}

helm/hpcc/values.yaml
RussWhitehead: I changed the code to allow the more familiar "ou=rootName" syntax

esp/services/ldapenvironment/ldapenvironmentService.cpp
RussWhitehead: Done

esp/services/ldapenvironment/ldapenvironmentService.cpp
RussWhitehead: Done, using the binding method you recommended

RussWhitehead:
@afishbeck Please rereview

esp/services/ldapenvironment/ldapenvironmentService.cpp
```diff
@@ -87,61 +87,77 @@ bool CldapenvironmentEx::onLDAPQueryDefaults(IEspContext &context, IEspLDAPQuery
...
 
+const char * CldapenvironmentEx::generatePassword(StringBuffer &pwd, int len)
+{
+    const char alphaUC[] = "ABCDEFGHIJKLMNOPQRSTUVWXYZ";
+    const char alphaLC[] = "abcdefghijklmnopqrstuvwxyz";
+    const char numeric[] = "0123456789";
+    const char symbol[] = "~`!@#$%^&*()_-+={[}]|:;<,>.?/";
+
+    srand(time(0));
+    for(int i = 0; i < len; i += 4)
```
afishbeck, line 118: You create an interesting pattern of mixed character types.  You could completely randomize it.  Have just one for loop and randomly choose first which character set and then which character from that set.

I found an example that does it that way:  https://x-engineer.org/random-password-generator-algorithm/#:~:text=is%20displayed%20below.-,clear(),-clc()%0A%0A//%20Define%20character

But in any case your password will be plenty hard to guess.

esp/services/ldapenvironment/ldapenvironmentService.cpp
```diff
@@ -87,61 +87,77 @@ bool CldapenvironmentEx::onLDAPQueryDefaults(IEspContext &context, IEspLDAPQuery
...
+        pwd.appendf("%c", alphaUC[rand() %  (sizeof(alphaUC)-1) ]);
+    for(int i = 1; i < len; i += 4)
+        pwd.appendf("%c", alphaLC[rand() %  (sizeof(alphaLC)-1) ]);
```
afishbeck, line 121: Did you try a simple append?  It should use the character version of append.

`pwd.append(alphaLC[rand() %  (sizeof(alphaLC)-1) ]);`

esp/services/ldapenvironment/ldapenvironmentService.cpp
```diff
@@ -152,49 +168,81 @@ bool CldapenvironmentEx::onLDAPCreateEnvironment(IEspContext &context, IEspLDAPC
...
         
+        //----------------------------------
+        // Create HPCC Admin Username/password.
+        // Attempt to create the Kubernetes secret for that user
+        // NOTE!  HPCCAdmin LDAP group and HPCCAdmin LDAP user will be
+        // created by ESP when started in Admin mode
+        //----------------------------------        
+        VStringBuffer respHPCCAdminUser("admin_%s", req.getEnvName());
+        StringBuffer  respHPCCAdminPwd;
+        StringBuffer  hpccAdminK8sSecretName("hpccadminsecret");
```
afishbeck, line 204: I would add the environment name (envName) to the k8s secret name so there aren't collisions as people use the service to create multiple environments over time.

esp/services/ldapenvironment/ldapenvironmentService.cpp
```diff
@@ -152,49 +168,81 @@ bool CldapenvironmentEx::onLDAPCreateEnvironment(IEspContext &context, IEspLDAPC
...
+
+        generatePassword(respHPCCAdminPwd, 10);
+
+        VStringBuffer cmdLine("kubectl create secret generic %s --from-literal=username=%s --from-literal=password=%s", hpccAdminK8sSecretName.str(), respHPCCAdminUser.str(), respHPCCAdminPwd.str());
+        try
+        {
+            DBGLOG("\nExecuting '%s'\n", cmdLine.str());
```
afishbeck, line 211: Need to make sure this doesn't get logged in a real deployment?

afishbeck (commented):
A few questions / comments, but it's looking good.

esp/services/ldapenvironment/ldapenvironmentService.cpp
```diff
@@ -0,0 +1,252 @@
...
+        // NOTE!  HPCCAdmin LDAP group and HPCCAdmin LDAP user will be
+        // created by ESP when started in Admin mode
+        //----------------------------------        
+        VStringBuffer respHPCCAdminUser("admin_%s", req.getEnvName());
+        StringBuffer  respHPCCAdminPwd;
+        StringBuffer  hpccAdminK8sSecretName("hpccadminsecret");
+
+        generatePassword(respHPCCAdminPwd, 10);
+
+        VStringBuffer cmdLine("kubectl create secret generic %s --from-literal=username=%s --from-literal=password=%s", hpccAdminK8sSecretName.str(), respHPCCAdminUser.str(), respHPCCAdminPwd.str());
```
afishbeck, line 208: I'm wondering if the k8s secret should be optional?

esp/services/ldapenvironment/ldapenvironmentService.cpp
RussWhitehead: You are correct

esp/services/ldapenvironment/ldapenvironmentService.cpp
RussWhitehead: Done

esp/services/ldapenvironment/ldapenvironmentService.cpp
RussWhitehead: Good catch, removed password from log and notes

esp/services/ldapenvironment/ldapenvironmentService.cpp
RussWhitehead: Agreed, added a bool checkbox in ECM to chose

RussWhitehead:
@afishbeck Please rereview

afishbeck (approved):
@RussWhitehead looks good.

RussWhitehead:
@afishbeck Squashed into 1 commit
@richardkchapman Please merge

richardkchapman:
This is targeting 8.8 - is that correct?

RussWhitehead:
@richardkchapman Discussed with @afishbeck and we feel that we should release this as soon as possible as a technical preview, since it's stand-alone/POC. We would like to get it into the hands of the stakeholders (TonyK, Fern/etc) so they can comment and request changes as needed
If you/ @stuartort agree, what release shall I rebase to ?

richardkchapman:
8.6.x

afishbeck:
@RussWhitehead @richardkchapman I agree we should get it out there for POC testing.  But this discussion made me take a look at how it get's deployed.  Perhaps the changes to values.yaml should probably be commented out and POC testers can uncomment and install this where they will be testing.  Even in future production we should only deploy these in select locations.

RussWhitehead:
@afishbeck I commented out the lines in values.yaml as you requested
@richardkchapman I rebased to 8.6.x
Please merge

RussWhitehead:
@richardkchapman Please merge

richardkchapman:
@RussWhitehead Needs rebasing for a clean merge

RussWhitehead:
@richardkchapman Rebased, please merge

PR #15747, opened 2022-02-08
HPCC-27152 Helm support for generating remote client certificates for services

afishbeck:
New remote issuer must be enabled (default is not enabled)

```
certificates:
  enabled: true
  issuers:
    remote:
      name: hpcc-remote-issuer
      enabled: true
```

And the issuer requires an additional private key pair
"hpcc-remote-issuer-key-pair", which can be created as follows:

```
openssl req -x509 -newkey rsa:2048 -nodes -keyout remoteca.key -sha256 -days 1825 -out remoteca.crt -config examples/certmanager/ca-req.cfg

kubectl create secret tls hpcc-remote-issuer-key-pair --cert=remoteca.crt --key=remoteca.key

```

Example values config:
---------------------------
```
esp:
- name: eclwatch
  application: eclwatch
  auth: none
  replicas: 1
  remoteClients:
  - name: myclient
```

This will:
1. Make the ESP require client certificates.
2. Create a client certificate for each remoteClient. The certificates and private keys must be set up for the client which varies by client and is outside the scope of this PR.
3. Restricts access to the ESP to listed clients.

Future work may define other ways of adding or using client certificates.

**To use with ecl command line:**

Example ecl.ini entries:

ecl.ini
----------------------
```
eclSSL=true
eclClientCert=/home/jdoe/myclient/tls.crt
eclClientPrivateKey=/home/jdoe/myclient/tls.key
eclCACert=/home/jdoe/myclient/ca.crt
```

example command line parameters:
```
ecl run --ssl --cert=/home/jdoe/myclient/tls.crt --key=/home/jdoe/myclient/tls.key --cacert=/home/jdoe/myclient/ca.crt hthor myquery.ecl
```

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/bindings/SOAP/Platform/soapbind.cpp
+++ b/esp/bindings/SOAP/Platform/soapbind.cpp
@@ -269,6 +269,30 @@ int CHttpSoapBinding::HandleSoapRequest(CHttpRequest* request, CHttpResponse* re
 }
 
 
+static IPropertyTree *createSecClientConfig(const char *clientCertPath, const char *clientPrivateKey, const char *caCertsPath, bool acceptSelfSigned)
+{
+    Owned<IPropertyTree> info = createPTree();
+
+    if (!isEmptyString(clientCertPath))
+    {
+        info->setProp("certificate", clientCertPath);
+        if (!isEmptyString(clientPrivateKey))
+            info->setProp("privatekey", clientPrivateKey);
+    }
+
+    IPropertyTree *verify = ensurePTree(info, "verify");
+    if (!isEmptyString(caCertsPath))
+    {
+        IPropertyTree *ca = ensurePTree(verify, "ca_certificates");
+        ca->setProp("@path", caCertsPath);
+    }
+    verify->setPropBool("@enable", true);
+    verify->setPropBool("@accept_selfsigned", acceptSelfSigned);
+    verify->setProp("trusted_peers", "anyone");
+
+    return info.getClear();
+}
+
 void CSoapRequestBinding::post(const char *proxy, const char* url, IRpcResponseBinding& response, const char *soapaction)
 {
     CRpcCall rpccall;
@@ -286,6 +310,8 @@ void CSoapRequestBinding::post(const char *proxy, const char* url, IRpcResponseB
         soapclient.setReadTimeoutSecs(readTimeoutSecs_);
     if (mtls_secret_.length())
         soapclient.setMtlsSecretName(mtls_secret_);
+    if (client_cert_.length() || ca_certs_.length() || accept_self_signed_)
+        soapclient.setSecureSocketConfig(createSecClientConfig(client_cert_, client_priv_key_, ca_certs_, accept_self_signed_));
 
     soapclient.setUsernameToken(soap_getUserId(), soap_getPassword(), soap_getRealm());
 
--- a/esp/bindings/SOAP/client/soapclient.cpp
+++ b/esp/bindings/SOAP/client/soapclient.cpp
@@ -145,10 +145,14 @@ int CSoapClient::postRequest(const char* contenttype, const char* soapaction, IR
     Owned<CSoapResponse> soap_response;
     soap_response.setown(new CSoapResponse);
 
+    Owned<IHttpClientContext> httpctx;
     //Send request and get response
     if(m_transportclient.get() == NULL)
     {
-        Owned<IHttpClientContext> httpctx = getHttpClientSecretContext(m_mtls_secret);
+        if (m_sec_config)
+            httpctx.setown(createHttpClientContext(LINK(m_sec_config)));
+        else
+            httpctx.setown(getHttpClientSecretContext(m_mtls_secret));
         Owned<IHttpClient> httpclient = httpctx->createHttpClient(call->getProxy(), call->get_url());
         if (m_disableKeepAlive)
             httpclient->disableKeepAlive();
@@ -357,7 +361,11 @@ int CSoapClient::postRequest(IRpcMessage & rpccall, StringBuffer & responsebuf, 
     //Send request and get response
     if(m_transportclient.get() == NULL)
     {
-        Owned<IHttpClientContext> httpctx = getHttpClientContext();
+        Owned<IHttpClientContext> httpctx;
+        if (m_sec_config)
+            httpctx.setown(createHttpClientContext(LINK(m_sec_config)));
+        else
+            httpctx.setown(getHttpClientContext());
         IHttpClient* httpclient = httpctx->createHttpClient(call->getProxy(), call->get_url());
         m_transportclient.setown(httpclient);
         if (m_disableKeepAlive)
--- a/helm/hpcc/templates/_helpers.tpl
+++ b/helm/hpcc/templates/_helpers.tpl
@@ -826,6 +826,10 @@ Generate instance queue names
 {{ end -}}
 {{- end -}}
 
+{{- define "hpcc.usesRemoteClientCertificates" -}}
+  {{- if (hasKey . "remoteClients") -}}{{- if (.remoteClients) -}} true {{- end -}}{{- end -}}
+{{- end -}}
+
 {{/*
 Generate service entries for TLS
 */}}
@@ -838,7 +842,8 @@ Generate service entries for TLS
     {{- if and ($externalService) (hasKey .component "certificate") }}
   tls: true
     {{- else }}
-      {{- $issuerName := ternary "public" "local" $externalService }}
+      {{- $externalIssuerName := ternary "remote" "public" (eq "true" ( include "hpcc.usesRemoteClientCertificates" . )) -}}
+      {{- $issuerName := ternary $externalIssuerName "local" $externalService }}
       {{- $certificates := (.root.Values.certificates | default dict) -}}
       {{- if not $certificates.enabled }}
   tls: false
@@ -883,7 +888,7 @@ Generate list of available services
   class: esp
   type: {{ $esp.application }}
   port: {{ $esp.service.servicePort }}
-  {{- include "hpcc.addTLSServiceEntries" (dict "root" $ "service" $esp "component" $esp "visibility" $esp.service.visibility) }}
+  {{- include "hpcc.addTLSServiceEntries" (dict "root" $ "service" $esp "component" $esp "visibility" $esp.service.visibility "remoteClients" $esp.remoteClients) }}
 {{ end -}}
 {{- range $dali := $.Values.dali -}}
 {{- $sashaServices := $dali.services | default dict -}}
@@ -1341,7 +1346,8 @@ use "public" or "local"
 {{- define "hpcc.addCertificate" }}
 {{- if (.root.Values.certificates | default dict).enabled -}}
 {{- $externalCert := or (and (hasKey . "external") .external) (ne (include "hpcc.isVisibilityPublic" .) "") -}}
-{{- $issuerName := .issuer | default (ternary "public" "local" $externalCert) -}}
+{{- $externalIssuerName := ternary "remote" "public" (eq "true" ( include "hpcc.usesRemoteClientCertificates" . )) -}}
+{{- $issuerName := .issuer | default (ternary $externalIssuerName "local" $externalCert) -}}
 {{- if eq (include "hpcc.isIssuerEnabled" (dict "root" .root "issuer" $issuerName)) "true" -}}
 {{- $issuer := get .root.Values.certificates.issuers $issuerName -}}
 {{- if $issuer -}}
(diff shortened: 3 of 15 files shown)


afishbeck:
@jakesmith @richardkchapman @ghalliday First draft up for discussion.

Makes it easy to create client certs for a particular service.  We can use them at first for simple mTLS, but later I can generate certificate name based ACLists from these same values.  

i.e. Only allow these specific certificates to work:
myremoteclient42@eclwatch101.eclwatch.hpccnamespace.
myremoteclient42@eclwatch101.eclwatch.hpccnamespace.

richardkchapman (approved):
Looks sensible to me

helm/hpcc/templates/_helpers.tpl
```diff
@@ -1337,6 +1337,67 @@ spec:
...
+  remoteClients:
+  - name: myRemoteClient
+Will generate secrets containing certificates that can be deployed to the remote client.
+If ESP or Roxie is configured to require client certificates then these certificates
+can be used to allow those remote clietns to connect.  Unique certificates should be created
+for each client so we can start to create certificate based access control lists in the near future.
+*/}}
+{{- define "hpcc.addClientCertificate" }}
+{{- if (.root.Values.certificates | default dict).enabled -}}
+{{- $externalCert := or (and (hasKey . "external") .external) (ne (include "hpcc.isVisibilityPublic" .) "") -}}
```
ghalliday, line 1353: indent: minor comment, would should the structure more clearly if the ifs were indented - line 1365 is for some reason.

ghalliday (approved):
One minor comment on indentation.

helm/hpcc/templates/dafilesrv.yaml
```diff
@@ -89,5 +89,12 @@ kind: ConfigMap
 {{ include "hpcc.addService" ( dict "root" $ "name" .name "service" $service "selector" .name "defaultPort" 7600 ) }}
 ---
 {{ include "hpcc.addCertificate" (dict "root" $ "name" .name "service" .service "component" "dafilesrv" "visibility" .service.visibility) }}
+
+{{- $instance := .name -}}
+{{- $visibility := .service.visibility -}}
+{{- range $remoteClient := .remoteClients }}
+ {{ include "hpcc.addClientCertificate" (dict "root" $ "client" $remoteClient.name "instance" $instance "component" "dafilesrv" "visibility" $visibility) }}
+{{- end }}
```
jakesmith, line 97: the cert. for foreign access may not be appropriate, given that it isn't possible to mix secure/non-secure client access from a foreign environments, and in near future remote won't use certs, but the existing MTLS and signature verificaion route.

jakesmith (commented):
1 comment.
needs a bit more discussion for specific use case,
but looks good.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -1337,6 +1337,67 @@ spec:
...
 
+{{/*
+Use cert-manager to create a public certificate and private key for use as
+remote client certificates.
+Adding the following to ESP, Roxie, or dafilesrv
+  remoteClients:
+  - name: myRemoteClient
+Will generate secrets containing certificates that can be deployed to the remote client.
+If ESP or Roxie is configured to require client certificates then these certificates
+can be used to allow those remote clietns to connect.  Unique certificates should be created
```
jakesmith, line 1348: typo: "clietns"

helm/hpcc/templates/dafilesrv.yaml
```diff
@@ -89,5 +89,12 @@ kind: ConfigMap
 {{ include "hpcc.addService" ( dict "root" $ "name" .name "service" $service "selector" .name "defaultPort" 7600 ) }}
 ---
 {{ include "hpcc.addCertificate" (dict "root" $ "name" .name "service" .service "component" "dafilesrv" "visibility" .service.visibility) }}
+
+{{- $instance := .name -}}
+{{- $visibility := .service.visibility -}}
+{{- range $remoteClient := .remoteClients }}
+ {{ include "hpcc.addClientCertificate" (dict "root" $ "client" $remoteClient.name "instance" $instance "component" "dafilesrv" "visibility" $visibility) }}
```
jakesmith, line 96: formatting: looks like the helm indentation may be wrong here, would expect perhaps it to be {{- include ..... -}}

afishbeck:
@jakesmith updated draft.  I'll convert to non-draft PR soon, but you can take a look, try it out now if you'd like.

1. ESP only for now (roxie needs a bit of work to turn on client cert support and add a cert based ACL.
2. Now supports new explicit remote cert-manager issuer.
3. Generated clients can now be configured with certificates, key, and cacert.
4. Command line works with client certs (although these certs are better for system to system rather than users which will come later).

afishbeck:
@richardkchapman @jakesmith @ghalliday changed from draft to PR.
Now currently ESP only.  Roxie needs a bit of minor refactoring to turn on client cert and ACL support.
Uses a dedicated "remote" certificate issuer.
This now adds support in the generated client code for setting client cert, key, and cacert.
Adds command line support.
- the current client certs are more appropriate certs for individual users.
- We need to think a bit more about user certs if we want to support them)
Adds ecl.ini support for SSL in general and these certs.

esp/bindings/SOAP/client/soapclient.cpp
```diff
@@ -145,10 +145,14 @@ int CSoapClient::postRequest(const char* contenttype, const char* soapaction, IR
     Owned<CSoapResponse> soap_response;
     soap_response.setown(new CSoapResponse);
 
+    Owned<IHttpClientContext> httpctx;
     //Send request and get response
     if(m_transportclient.get() == NULL)
     {
-        Owned<IHttpClientContext> httpctx = getHttpClientSecretContext(m_mtls_secret);
+        if (m_sec_config)
+          httpctx.setown(createHttpClientContext(LINK(m_sec_config)));
```
richardkchapman, line 153: Indentation looks dodgy here - tabs?

helm/hpcc/templates/_helpers.tpl
```diff
@@ -1370,6 +1376,109 @@ spec:
...
+  name: client-{{ $issuerName }}-{{ $component }}-{{ $instance }}-{{ $client }}-cert
+  namespace: {{ $namespace }}
+spec:
+  # Secret names are always required.
+  secretName: client-{{ $issuerName }}-{{ $component }}-{{ $instance }}-{{ $client }}-tls
+  duration: 2160h # 90d
+  renewBefore: 360h # 15d
+  subject:
+    organizations:
+    - HPCC Systems
```
richardkchapman, line 1461: Is this appropriate, in the OSS helm charts? Not all users will be us...

helm/hpcc/values.yaml
```diff
@@ -483,6 +495,8 @@ esp:
   application: eclwatch
   auth: none
   replicas: 1
+# remoteClients:
```
richardkchapman, line 498: Perhaps a better comment would be helpful

richardkchapman (changes requested):
Some minor comments.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -1370,6 +1376,109 @@ spec:
 {{- end }}
 {{- end }}
 
+{{/*
+Builds the commonName for a client certificate.  Used in creation of both certificate and access control list.
```
jakesmith, line 1380: useful in keeping with [most] other helpers, if comment says what helper template expects to be passed in in dictionary, e.g. "Pass in root, me .. "

helm/hpcc/templates/_helpers.tpl
```diff
@@ -1370,6 +1376,109 @@ spec:
...
+{{- end }}
+
+{{/*
+Use cert-manager to create a public certificate and private key for use as
+remote client certificates.
+Adding the following to ESP (Roxie support to be added later)
+  remoteClients:
+  - name: myRemoteClient
+Will generate certificates that can be deployed to the remote client.
+Will cause ESP require client certificates when a socket connects.
```
jakesmith, line 1424: will cause ESP to require

jakesmith (commented):
@afishbeck - looks good a few minor comments. I will try to test it.

esp/bindings/SOAP/Platform/soapbind.cpp
```diff
@@ -286,6 +310,8 @@ void CSoapRequestBinding::post(const char *proxy, const char* url, IRpcResponseB
         soapclient.setReadTimeoutSecs(readTimeoutSecs_);
     if (mtls_secret_.length())
         soapclient.setMtlsSecretName(mtls_secret_);
+    if (client_cert_.length() || ca_certs_.length() || accept_self_signed_)
+        soapclient.setSecureSocketConfig(LINK(createSecClientConfig(client_cert_, client_priv_key_, ca_certs_, accept_self_signed_)));
```
ghalliday, line 314: Leak: LINK shouldn't be required - because the function returns an object with the link count already incremented.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -1370,6 +1376,109 @@ spec:
 {{- end }}
 {{- end }}
 
+{{/*
+Builds the commonName for a client certificate.  Used in creation of both certificate and access control list.
+*/}}
+{{- define "hpcc.getClientCommonName" -}}
+  {{- if (.root.Values.certificates | default dict).enabled -}}
```
ghalliday, line 1383: indent: minor.  I think we agreed on using a single space for indenting helm files.  @jakesmith ?

helm/hpcc/templates/_helpers.tpl
```diff
@@ -1370,6 +1376,109 @@ spec:
...
+
+{{/*
+Use cert-manager to create a public certificate and private key for use as
+remote client certificates.
+Adding the following to ESP (Roxie support to be added later)
+  remoteClients:
+  - name: myRemoteClient
+Will generate certificates that can be deployed to the remote client.
+Will cause ESP require client certificates when a socket connects.
+Will create a TLS based access contol list which ESP will check to make sure a connections client certificate is enabled.
```
ghalliday, line 1425: typo: control

ghalliday (commented):
A few minor comments

helm/hpcc/templates/_helpers.tpl
jakesmith: yes we did.

esp/bindings/SOAP/Platform/soapbind.cpp
afishbeck: Fixed.

afishbeck:
Updated based on review:

1. Organization can now be set for client certificate
esp:
  - name: eclwatch
    remoteClients:
    - name: myclient
      organization: mycompany
2. Fixed client code leak
3. Cleaned up indents and comments.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -1370,6 +1376,121 @@ spec:
...
+    {{- $service := (.service | default dict) -}}
+    {{- $domain := ( $service.domain | default $issuer.domain | default $namespace | default "default" ) -}}
+    {{- .client }}@{{ .instance }}.{{ .component }}.{{ $domain }}
+   {{- end -}}
+  {{- end -}}
+ {{- end -}}
+{{- end -}}
+
+{{/*
+Turns an array of remoteClients into a | delemited string to be used for the trusted_peers element of SecureSocket settings.
```
jakesmith, line 1404: typo: delemited -> delimited

jakesmith (approved):
I've now tested this, have a PR near ready based on this, where DFS uses generated client certificates.
The only caveat (as discussed offline), is that the mechanism requires key and certificate base files, whereas the secret interface returns binary blobs.
As discussed, the workaround for now, will be fetch the local secrets path and use the files directly (that won't work if secrets are in vault though). Th other alternative is to save the secret into local ephemeral storage.

We should explore whether the secure socket implementation can be changed to support keys/certs coming from binary blob to prevent this workaround. Is there JIRA for that?

@afishbeck - 1 typo,  otherwise looks good to squash.

afishbeck:
Fixed typo, rebased, and created two JIRAs about vault and in memory keys and certificates:

https://track.hpccsystems.com/browse/HPCC-27254
https://track.hpccsystems.com/browse/HPCC-27255

afishbeck:
@richardkchapman this should be ready for merging.

richardkchapman (approved):

PR #15751, opened 2022-02-09
HPCC-27128 Implement #$ and #$.^ expanding macros in another package

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/ecl/hql/hqlexpr.cpp
+++ b/ecl/hql/hqlexpr.cpp
@@ -7347,9 +7347,9 @@ unsigned __int64 CHqlAnnotation::querySequenceExtra()
     return body->querySequenceExtra();
 }
 
-IInterface * CHqlAnnotation::queryUnknownExtra()
+IInterface * CHqlAnnotation::queryUnknownExtra(unsigned idx)
 {
-    return body->queryUnknownExtra();
+    return body->queryUnknownExtra(idx);
 }
 
 IValue * CHqlAnnotation::queryValue() const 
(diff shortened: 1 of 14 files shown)


ecl/hql/hqlexpr.cpp
```diff
@@ -10705,6 +10709,62 @@ StringBuffer &CHqlUnknown::toString(StringBuffer &ret)
 
 //==============================================================================================================
 
+CHqlMacro::CHqlMacro(node_operator _op, ITypeInfo * _type, IEclPackage * _package, IAtom * _name, IFileContents * _ownedContents)
```
ghalliday, line 10712: I could have extended CHqlUnknown, but macros are sufficiently different that I decided it was probably better to have a separate class.  Many of the functions are very similar.  The main change is an extra package parameter, which is retrieved later on in the lexer.

ecl/hql/hqlexpr.cpp
```diff
@@ -10705,6 +10709,62 @@ StringBuffer &CHqlUnknown::toString(StringBuffer &ret)
...
+
+IHqlExpression *CHqlMacro::clone(HqlExprArray &newkids)
+{
+    assertex(newkids.ordinality() == 0);
+    return LINK(this);
+}
+
+StringBuffer &CHqlMacro::toString(StringBuffer &ret)
+{
+    return ret.append(name);
```
shamser, line 10763: Should this return the macro definition rather than just the name?

shamser (changes requested):

ecl/hql/hqlexpr.cpp
ghalliday: I don't think this is ever called, and it is the same as the previous implementation in CHqlUnknown - so I would keep it the same.

shamser (approved):

PR #15767, opened 2022-02-11
HPCC-27177 Enables logging feature for DESDL scripted service

marksr:
Signed-off-by: Marcos Ribeiro <marcos.ribeiro@lexisnexisrisk.com>

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Cloud-compatibility
  - [x] Premature optimization

## Smoketest:

### DIFF
--- a/esp/services/esdl_svc_engine/esdl_binding.cpp
+++ b/esp/services/esdl_svc_engine/esdl_binding.cpp
@@ -1053,7 +1053,7 @@ void EsdlServiceImpl::handleServiceRequest(IEspContext &context,
             context.addTraceSummaryTimeStamp(LogNormal, "end-reqproc", TXSUMMARY_GRP_CORE|TXSUMMARY_GRP_ENTERPRISE);
             context.addTraceSummaryValue(LogNormal, "custom_fields.recCount", recordCount, TXSUMMARY_GRP_ENTERPRISE);
 
-            if(isPublishedQuery(implType))
+            if(isPublishedQuery(implType) || implType==EsdlMethodImplScript)
                 tgtctx.setown(createTargetContext(context, tgtcfg.get(), srvdef, mthdef, req));
 
             reqcontent.set(reqWriter->str());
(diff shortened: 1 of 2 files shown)


esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -1080,6 +1080,7 @@ void EsdlServiceImpl::handleServiceRequest(IEspContext &context,
             {
                 m_pEsdlTransformer->process(context, EsdlResponseMode, srvdef.queryName(), mthdef.queryName(), out, origResp.str(), txResultFlags, ns, schema_location);
                 runPostEsdlScript(context, scriptContext, srvdef, mthdef, out, txResultFlags, ns, schema_location);
+                logdata.set("<LogDatasets/>");
```
afishbeck, line 1083: I think this is good for now.  I do wonder if in the future we will need to have the service script able to populate LogDatasets.  I guess we could already do it another way and we should worry about it in the future.  One scenario I can already imagine is where the service script is calling multiple roxie queries that each might want to log something.

Anyway, just something to keep in mind for the future.

afishbeck (approved):
@marksr looks good.

afishbeck:
@asselitx please review.

afishbeck:
Actually @marksr just one thing.  In your commit can you remove the second "-"?

```
HPCC-27177 - Enable logging feature for DESDL scripted service
           ^
           |
```

marksr:
> Actually @marksr just one thing. In your commit can you remove the second "-"?
> 
> ```
> HPCC-27177 - Enable logging feature for DESDL scripted service
>            ^
>            |
> ```

Changed!

esp/services/esdl_svc_engine/esdl_binding.cpp
asselitx: Even with this change though, the script type methods will still be able to populate LogDatasets in the PreLogging entry point, correct? It does appear that that the handleResultLogging() function is called for those also. If I'm reading correctly, this change just ensures that there is a LogDatasets context section to write into. 

I have seen other cases where a context section doesn't exist unless you define an entry point script that would operate on that section (eg EsdlResponse). So perhaps in a future change we'll move this change into a central location that ensures all context sections exist regardless of the execution path.

In any case this looks ok.

asselitx (approved):
Looks good.

esp/services/esdl_svc_engine/esdl_binding.cpp
marksr: Exactly, going forward without LogDatasets tag causes some errors into handleResultLogging(). It would be good to have this previously populated in a central location.

PR #15770, opened 2022-02-11
HPCC-24691 Convert ESP to use jlog standards

kenrowland:
Converted code for container builds to use log category set
in config yaml files. Legacy bare metal continues to use
existing ESP log level.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/bindings/SOAP/client/soapclient.cpp
+++ b/esp/bindings/SOAP/client/soapclient.cpp
@@ -124,10 +124,7 @@ int CSoapClient::postRequest(const char* contenttype, const char* soapaction, IR
         multipart->serialize(contenttypestr, requeststr);
     }
 
-    if (getEspLogLevel(rpccall.queryContext())>LogNormal)
-    {
-        DBGLOG("Content type: %s", contenttypestr.str());
-    }
+    LOG(getEspLogCategoryForContext(rpccall.queryContext(), LegacyMsgCatMax), "Content type: %s", contenttypestr.str());
 
     Owned<CSoapRequest> soap_request;
     soap_request.setown(new CSoapRequest);
--- a/esp/bindings/http/client/httpclient.cpp
+++ b/esp/bindings/http/client/httpclient.cpp
@@ -416,11 +416,8 @@ HttpClientErrCode CHttpClient::sendRequest(const char* method, const char* conte
         httprequest->addHeader("WWW-Authenticate", authheader.str());
     }
 
-    if (getEspLogLevel()>LogNormal)
-    {
-        DBGLOG("Content type: %s", contenttype);
-        DBGLOG("Request content: %s", request.str());
-    }
+    LOG(LegacyMsgCatMax, "Content type: %s", contenttype);
+    LOG(LegacyMsgCatMax, "Request content: %s", request.str());
 
     httprequest->setContent(request.str());
 
@@ -463,8 +460,7 @@ HttpClientErrCode CHttpClient::sendRequest(const char* method, const char* conte
     m_persistable = httpresponse->getPersistentEligible();
     m_numRequests++;
 
-    if (getEspLogLevel()>LogNormal)
-        DBGLOG("Response content: %s", response.str());
+    LOG(LegacyMsgCatMax,"Response content: %s", response.str());
 
     return HttpClientErrCode::OK;
 }
@@ -596,12 +592,9 @@ HttpClientErrCode CHttpClient::proxyRequest(IHttpMessage *request, IHttpMessage 
     httprequest->setContentType(contentType);
     httprequest->setContent(forwardRequest->queryContent());
 
-    if (getEspLogLevel()>LogNormal)
-    {
-        StringBuffer s;
-        DBGLOG("Content type: %s", forwardRequest->getContentType(s).str());
-        DBGLOG("Request content: %s", forwardRequest->queryContent());
-    }
+    StringBuffer s;
+    LOG(LegacyMsgCatMax, "Content type: %s", forwardRequest->getContentType(s).str());
+    LOG(LegacyMsgCatMax, "Request content: %s", forwardRequest->queryContent());
 
     copyCookies(*httprequest, *forwardRequest, m_host);
 
@@ -629,8 +622,7 @@ HttpClientErrCode CHttpClient::proxyRequest(IHttpMessage *request, IHttpMessage 
     m_persistable = httpresponse->getPersistentEligible();
     m_numRequests++;
 
-    if (getEspLogLevel()>LogNormal)
-        DBGLOG("Response content: %s", httpresponse->queryContent());
+    LOG(LegacyMsgCatMax, "Response content: %s", httpresponse->queryContent());
 
     return HttpClientErrCode::OK;
 }
@@ -708,11 +700,9 @@ IHttpMessage *CHttpClient::sendRequestEx(const char* method, const char* content
         httprequest->addHeader("WWW-Authenticate", authheader.str());
     }
 
-    if (getEspLogLevel()>LogNormal)
-    {
-        DBGLOG("Content type: %s", contenttype);
-        DBGLOG("Request content: %s", content.str());
-    }
+    LOG(LegacyMsgCatMax, "Content type: %s", contenttype);
+    LOG(LegacyMsgCatMax, "Request content: %s", content.str());
+
 
     httprequest->setContent(content.str());
 
@@ -773,8 +763,7 @@ HttpClientErrCode CHttpClient::sendRequest(IProperties *headers, const char* met
     resp->getContent(responseContent);
     resp->getStatus(responseStatus);
 
-    if (getEspLogLevel()>LogNormal)
-        DBGLOG("Response content: %s", responseContent.str());
+    LOG(LegacyMsgCatMax, "Response content: %s", responseContent.str());
 
     return code;
 }
@@ -1025,11 +1014,8 @@ HttpClientErrCode CHttpClient::postRequest(ISoapMessage &req, ISoapMessage& resp
     StringBuffer content;
     httpresponse->getContent(content);
 
-    if (getEspLogLevel()>LogNormal)
-    {
-        if(httpresponse->isTextMessage())
-            DBGLOG("http response content = %s", content.str());
-    }
+    if(httpresponse->isTextMessage())
+        LOG(LegacyMsgCatMax, "http response content = %s", content.str());
 
     response.set_text(content.str());
             
--- a/esp/bindings/http/platform/httpbinding.cpp
+++ b/esp/bindings/http/platform/httpbinding.cpp
@@ -674,14 +674,14 @@ void EspHttpBinding::populateRequest(CHttpRequest *request)
         ISecResourceList* settinglist = m_setting_authmap->getResourceList("*");
         if(settinglist == NULL)
             return ;
-        if (getEspLogLevel()>=LogMax)
+        if (queryLogMsgManager()->rejectsCategory(LegacyMsgCatMax))
         {
             StringBuffer s;
-            DBGLOG("Set security settings: %s", settinglist->toString(s).str());
+            LOG(LegacyMsgCatMax, "Set security settings: %s", settinglist->toString(s).str());
         }
         ctx->setSecuritySettings(settinglist);
     }
-    return ;
+    return;
 }
 
 bool EspHttpBinding::doAuth(IEspContext* ctx)
@@ -799,7 +799,7 @@ bool EspHttpBinding::basicAuth(IEspContext* ctx)
             {
                 const char *desc=curres->getDescription();
                 VStringBuffer msg("Access for user '%s' denied to: %s. Access=%d, Required=%d", user->getName(), desc?desc:"<no-desc>", access, required);
-                ESPLOG(LogMin, "%s", msg.str());
+                LOG(LegacyMsgCatMin, "%s", msg.str());
                 ctx->AuditMessage(AUDIT_TYPE_ACCESS_FAILURE, "Authorization", "Access Denied: Not Authorized", "Resource: %s [%s]", curres->getName(), (desc) ? desc : "");
                 ctx->setAuthError(EspAuthErrorNotAuthorized);
                 ctx->setRespMsg(msg.str());
@@ -864,11 +864,11 @@ bool EspHttpBinding::sendFromESPCache(IEspCache* cacheClient, CHttpRequest* requ
 {
     StringBuffer content, contentType;
     if (!cacheClient->readResponseCache(cacheID, content.clear(), contentType.clear()))
-        ESPLOG(LogMax, "Failed to read from ESP Cache for %s.", request->queryServiceMethod());
+        LOG(LegacyMsgCatMax, "Failed to read from ESP Cache for %s.", request->queryServiceMethod());
     if (content.isEmpty() || contentType.isEmpty())
         return false;
 
-    ESPLOG(LogMax, "Sending from ESP Cache for %s.", request->queryServiceMethod());
+    LOG(LegacyMsgCatMax, "Sending from ESP Cache for %s.", request->queryServiceMethod());
     response->setContentType(contentType.str());
     response->setContent(content.str());
     response->send();
@@ -881,9 +881,9 @@ void EspHttpBinding::addToESPCache(IEspCache* cacheClient, CHttpRequest* request
     response->getContent(content);
     response->getContentType(contentType);
     if (cacheClient->cacheResponse(cacheID, cacheSeconds, content.str(), contentType.str()))
-        ESPLOG(LogMax, "AddTo ESP Cache for %s.", request->queryServiceMethod());
+        LOG(LegacyMsgCatMax, "AddTo ESP Cache for %s.", request->queryServiceMethod());
     else
-        ESPLOG(LogMax, "Failed to add ESP Cache for %s.", request->queryServiceMethod());
+        LOG(LegacyMsgCatMax, "Failed to add ESP Cache for %s.", request->queryServiceMethod());
 }
 
 void EspHttpBinding::clearCacheByGroupID(const char *ids)
@@ -895,7 +895,7 @@ void EspHttpBinding::clearCacheByGroupID(const char *ids)
     if (!espContainer->hasCacheClient())
         return;
 
-    ESPLOG(LogMax, "clearCacheByGroupID %s.", ids);
+    LOG(LegacyMsgCatMax, "clearCacheByGroupID %s.", ids);
     StringArray errorMsgs;
     espContainer->clearCacheByGroupID(ids, errorMsgs);
     if (errorMsgs.length() > 0)
@@ -946,10 +946,8 @@ int EspHttpBinding::onGet(CHttpRequest* request, CHttpResponse* response)
 
     // At this time, the request is already received and fully passed, and
     // the user authenticated
-    LogLevel level = getEspLogLevel(&context);
-    if (level >= LogNormal)
-        DBGLOG("EspHttpBinding::onGet");
-    
+    LOG(getEspLogCategoryForContext(&context, LegacyMsgCatMax), "EspHttpBinding::onGet");
+
     response->setVersion(HTTP_VERSION);
     response->addHeader("Expires", "0");
 
@@ -1555,8 +1553,8 @@ int EspHttpBinding::onGetIframe(IEspContext &context, CHttpRequest* request, CHt
     if (plainText.length() > 0)
         inner.appendf("&PlainText=%s", plainText.str());
             
-    ESPLOG(LogNormal,"Inner: %s", inner.str());
-    ESPLOG(LogNormal,"Param: %s", request->queryParamStr());
+    LOG(LegacyMsgCatNormal,"Inner: %s", inner.str());
+    LOG(LegacyMsgCatNormal,"Param: %s", request->queryParamStr());
     content.appendf("<frame name=\"imain\" target=\"main\" src=\"%s\" scrolling=\"auto\" frameborder=\"0\" noresize=\"noresize\"/>", inner.str());
     content.append("</frameset>");
     content.append("</body></html>");
@@ -1578,7 +1576,7 @@ int EspHttpBinding::onGetConfig(IEspContext &context, CHttpRequest* request, CHt
     {
         if (getESPContainer() && getESPContainer()->queryApplicationConfig())
         {
-            ESPLOG(LogNormal, "Get config generated during application init");
+            LOG(LegacyMsgCatNormal, "Get config generated during application init");
             StringBuffer content("<?xml version=\"1.0\" encoding=\"UTF-8\"?>");
             if (context.queryRequestParameters()->hasProp("display"))
                 content.append("<?xml-stylesheet type=\"text/xsl\" href=\"/esp/xslt/xmlformatter.xsl\"?>");
@@ -1590,7 +1588,7 @@ int EspHttpBinding::onGetConfig(IEspContext &context, CHttpRequest* request, CHt
             return 0;
         }
 
-        ESPLOG(LogNormal, "Get config file: %s", m_configFile.get());
+        LOG(LegacyMsgCatNormal, "Get config file: %s", m_configFile.get());
 
         StringBuffer content;
         xmlContentFromFile(m_configFile, "/esp/xslt/xmlformatter.xsl", content);
@@ -2823,8 +2821,7 @@ void EspHttpBinding::validateResponse(IEspContext& context, CHttpRequest* reques
     getSchema(xsd,context,request,serviceQName,methodQName,true);
             
     // validation
-    if (getEspLogLevel()>LogMax)
-        DBGLOG("[VALIDATE] xml: %s\nxsd: %s\nns: %s",xml.str(), xsd.str(), ns.str());
+    LOG(LegacyMsgCatMax, "[VALIDATE] xml: %s\nxsd: %s\nns: %s",xml.str(), xsd.str(), ns.str());
 
     IXmlValidator* v = getXmlLibXmlValidator();
 
@@ -2859,7 +2856,7 @@ void EspHttpBinding::validateResponse(IEspContext& context, CHttpRequest* reques
 void EspHttpBinding::sortResponse(IEspContext& context, CHttpRequest* request, MemoryBuffer& content, 
                                       const char *serviceName, const char* methodName)
 {
-    ESPLOG(LogNormal,"Sorting Response XML...");
+    LOG(LegacyMsgCatNormal,"Sorting Response XML...");
 
     try {
         StringBuffer result;
@@ -2873,8 +2870,7 @@ void EspHttpBinding::sortResponse(IEspContext& context, CHttpRequest* request, M
 
         xform->setXmlSource(respXML,respXML.length());
         xform->transform(result);
-        if (getEspLogLevel()>LogNormal)
-            DBGLOG("XML sorted: %s", result.str());
+        LOG(LegacyMsgCatMax, "XML sorted: %s", result.str());
         unsigned len = result.length();
         content.setBuffer(len, result.detach(), true);      
     } catch (IException* e) {
--- a/esp/bindings/http/platform/httpservice.cpp
+++ b/esp/bindings/http/platform/httpservice.cpp
@@ -232,7 +232,7 @@ int CEspHttpServer::processRequest()
         StringBuffer serviceName;
         StringBuffer methodName;
         m_request->getEspPathInfo(stype, &pathEx, &serviceName, &methodName, false);
-        ESPLOG(LogNormal,"sub service type: %s. parm: %s", getSubServiceDesc(stype), m_request->queryParamStr());
+        LOG(LegacyMsgCatNormal,"sub service type: %s. parm: %s", getSubServiceDesc(stype), m_request->queryParamStr());
 
         m_request->updateContext();
         ctx->setServiceName(serviceName.str());
@@ -263,9 +263,9 @@ int CEspHttpServer::processRequest()
         StringBuffer peerStr, pathStr;
         const char *userid=ctx->queryUserId();
         if (isEmptyString(userid))
-            ESPLOG(LogMin, "%s %s, from %s", method.str(), m_request->getPath(pathStr).str(), m_request->getPeer(peerStr).str());
+            LOG(LegacyMsgCatMin, "%s %s, from %s", method.str(), m_request->getPath(pathStr).str(), m_request->getPeer(peerStr).str());
         else //user ID is in HTTP header
-            ESPLOG(LogMin, "%s %s, from %s@%s", method.str(), m_request->getPath(pathStr).str(), userid, m_request->getPeer(peerStr).str());
+            LOG(LegacyMsgCatMin, "%s %s, from %s@%s", method.str(), m_request->getPath(pathStr).str(), userid, m_request->getPeer(peerStr).str());
 
         authState = checkUserAuth();
         if ((authState == authTaskDone) || (authState == authFailed))
@@ -969,7 +969,7 @@ EspAuthState CEspHttpServer::checkUserAuth()
     if (!authReq.authBinding)
         throw MakeStringException(-1, "Cannot find ESP HTTP Binding");
 
-    ESPLOG(LogMax, "checkUserAuth: %s %s", m_request->isSoapMessage() ? "SOAP" : "HTTP", authReq.httpMethod.isEmpty() ? "??" : authReq.httpMethod.str());
+    LOG(LegacyMsgCatMax, "checkUserAuth: %s %s", m_request->isSoapMessage() ? "SOAP" : "HTTP", authReq.httpMethod.isEmpty() ? "??" : authReq.httpMethod.str());
 
     //The preCheckAuth() does not return authUnknown when:
     //No authentication is required for the ESP binding;
@@ -1416,12 +1416,12 @@ EspAuthState CEspHttpServer::handleUserNameOnlyMode(EspAuthRequest& authReq)
             strieq(authReq.httpMethod.str(), POST_METHOD);
         if (basicAuthentication)
         {
-            ESPLOG(LogMin, "Authentication failed: send BasicAuthentication.");
+            LOG(LegacyMsgCatMin, "Authentication failed: send BasicAuthentication.");
             m_response->sendBasicChallenge(authReq.authBinding->getChallengeRealm(), true);
         }
         else
         {
-            ESPLOG(LogMin, "Authentication failed: call askUserLogin.");
+            LOG(LegacyMsgCatMin, "Authentication failed: call askUserLogin.");
             //Display a GetUserName (similar to login) page to get a user name.
             askUserLogin(authReq, "Empty username.");
         }
@@ -1462,7 +1462,7 @@ bool CEspHttpServer::isAuthRequiredForBinding(EspAuthRequest& authReq)
 
 EspAuthState CEspHttpServer::checkUserAuthPerSession(EspAuthRequest& authReq, StringBuffer& authorizationHeader)
 {
-    ESPLOG(LogMax, "checkUserAuthPerSession");
+    LOG(LegacyMsgCatMax, "checkUserAuthPerSession");
 
     unsigned sessionID = readCookie(authReq.authBinding->querySessionIDCookieName());
     if (sessionID > 0)
@@ -1515,7 +1515,7 @@ EspAuthState CEspHttpServer::checkUserAuthPerSession(EspAuthRequest& authReq, St
 
 EspAuthState CEspHttpServer::checkUserAuthPerRequest(EspAuthRequest& authReq)
 {
-    ESPLOG(LogMax, "checkUserAuthPerRequest");
+    LOG(LegacyMsgCatMax, "checkUserAuthPerRequest");
 
     authReq.authBinding->populateRequest(m_request.get());
     if (authReq.authBinding->doAuth(authReq.ctx))
@@ -1525,7 +1525,7 @@ EspAuthState CEspHttpServer::checkUserAuthPerRequest(EspAuthRequest& authReq)
         authReq.ctx->setAuthStatus(AUTH_STATUS_OK); //May be changed to AUTH_STATUS_NOACCESS if failed in feature level authorization.
 
         StringBuffer userName, peer;
-        ESPLOG(LogNormal, "Authenticated for %s@%s", authReq.ctx->getUserID(userName).str(), m_request->getPeer(peer).str());
+        LOG(LegacyMsgCatNormal, "Authenticated for %s@%s", authReq.ctx->getUserID(userName).str(), m_request->getPeer(peer).str());
         return authSucceeded;
     }
     if (!authReq.isSoapPost)
@@ -1551,7 +1551,7 @@ EspAuthState CEspHttpServer::authNewSession(EspAuthRequest& authReq, const char*
     StringBuffer peer;
     m_request->getPeer(peer);
 
-    ESPLOG(LogMax, "authNewSession for %s@%s", _userName, peer.str());
+    LOG(LegacyMsgCatMax, "authNewSession for %s@%s", _userName, peer.str());
 
     authReq.ctx->setUserID(_userName);
     authReq.ctx->setPassword(_password);
@@ -1559,7 +1559,7 @@ EspAuthState CEspHttpServer::authNewSession(EspAuthRequest& authReq, const char*
     if (!authReq.authBinding->doAuth(authReq.ctx) && (authReq.ctx->getAuthError() != EspAuthErrorNotAuthorized))
     {
         authReq.ctx->setAuthStatus(AUTH_STATUS_FAIL);
-        ESPLOG(LogMin, "Authentication failed for %s@%s", _userName, peer.str());
+        LOG(LegacyMsgCatMin, "Authentication failed for %s@%s", _userName, peer.str());
         return handleAuthFailed(true, authReq, unlock, "User authentication failed.");
     }
 
@@ -1569,7 +1569,7 @@ EspAuthState CEspHttpServer::authNewSession(EspAuthRequest& authReq, const char*
     unsigned sessionID = createHTTPSession(authReq.authBinding, _userName, sessionStartURL);
     authReq.ctx->setSessionToken(sessionID);
 
-    ESPLOG(LogMax, "Authenticated for %s@%s", _userName, peer.str());
+    LOG(LegacyMsgCatMax, "Authenticated for %s@%s", _userName, peer.str());
 
     VStringBuffer cookieStr("%u", sessionID);
     addCookie(authReq.authBinding->querySessionIDCookieName(), cookieStr.str(), 0, true);
@@ -1785,7 +1785,7 @@ void CEspHttpServer::resetSessionTimeout(EspAuthRequest& authReq, unsigned sessi
         addCookie(authReq.authBinding->querySessionIDCookieName(), sessionIDStr.str(), 0, true);
         addCookie(SESSION_AUTH_OK_COOKIE, "true", 0, false); //client can access this cookie.
 
-        if (getEspLogLevel()>=LogMax)
+        if (!queryLogMsgManager()->rejectsCategory(MCuserProgress))
         {
             CDateTime timeoutAtCDT;
             StringBuffer timeoutAtString, nowString;
@@ -1795,7 +1795,7 @@ void CEspHttpServer::resetSessionTimeout(EspAuthRequest& authReq, unsigned sessi
                 now.getString(nowString).str(), createTime, timeoutAtCDT.getString(timeoutAtString).str(), timeoutAt);
         }
         else
-            ESPLOG(LogMin, "Reset %s for (/%s/%s) : %ld", PropSessionTimeoutAt, authReq.serviceName.isEmpty() ? "" : authReq.serviceName.str(),
+            LOG(LegacyMsgCatMin, "Reset %s for (/%s/%s) : %ld", PropSessionTimeoutAt, authReq.serviceName.isEmpty() ? "" : authReq.serviceName.str(),
                 authReq.methodName.isEmpty() ? "" : authReq.methodName.str(), timeoutAt);
 
         if (format == ESPSerializationJSON)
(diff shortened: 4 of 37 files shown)


esp/platform/espp.cpp
```diff
@@ -298,14 +329,38 @@ void openEspLogFile(IPropertyTree* envpt, IPropertyTree* procpt)
...
+    //only here to manipulate default log level on stderr stream for smoketest purposes
+    Owned<ILogMsgFilter> filter = getCategoryLogMsgFilter(MSGAUD_all, MSGCLS_all, mapLogLevelToThreshold(LogMin), true);
+    queryLogMsgManager()->changeMonitorFilter(queryStderrLogMsgHandler(), filter);
 #else
     setupContainerizedLogMsgHandler();
+
+    //
+    // ESP Log Update
+    // Add custom log monitor to handle elevated context based logging
+    CEspForceLogOutputFilter *filter = new CEspForceLogOutputFilter();
```
rpastrana, line 356: elevated context base logging is not a very common feature as far as I know, if attaching this filter to the logmsghandler adds any kind of overhead, let's delay loading this as late as possible.

esp/platform/espp.cpp
```diff
@@ -544,6 +600,9 @@ int init_main(int argc, const char* argv[])
             config->bindServer(*server.get(), *server.get());
             config->checkESPCache(*server.get());
 
+            if (fileHandler)
+                server->setFileHandler(fileHandler);
```
rpastrana, line 604: this is really a logmsghandler, rename the local var, and the setFileHandler function

rpastrana (changes requested):
@kenrowland left a few comments we need to address. Setup a meeting if you'd like to go over them.
I didn't review the ESPLOG to LOG changes.

esp/services/ws_sql/ws_sqlService.cpp
```diff
@@ -62,10 +62,12 @@ bool CwssqlEx::onGetDBMetaData(IEspContext &context, IEspGetDBMetaDataRequest &r
...
         const char * querysetfilter = req.getQuerySet();
 #ifdef _CONTAINERIZED
-        ESPLOG(LogNormal, "WsSQL: getting containerTargetClusters...");
+//        ESPLOG(LogNormal, "WsSQL: getting containerTargetClusters...");
+        LOG(LegacyMsgCat_Normal, "WsSQL: getting containerTargetClusters...");
         Owned<IStringIterator> targets = getContainerTargetClusters(nullptr, nullptr);
 #else
-        ESPLOG(LogNormal, "WsSQL-legacy: getting targetClusters...");
+//        ESPLOG(LogNormal, "WsSQL-legacy: getting targetClusters...");
+        LOG(LegacyMsgCat_Normal, "WsSQL-legacy: getting targetClusters...");
```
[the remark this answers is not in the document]
kenrowland: Make PROGLOG as well per review

esp/bindings/http/client/httpclient.cpp
```diff
@@ -79,6 +79,7 @@ void CHttpClientContext::initPersistentHandler()
         DBGLOG("Persistent connection won't be enabled for httpclient because maxPersistentIdleTime or maxPersistentRequests is set to 0");
         return;
     }
+    // todo - logging anything here that needs to be done?
     m_persistentHandler.setown(createPersistentHandler(nullptr, maxIdleTime, maxReqs, static_cast<PersistentLogLevel>(getEspLogLevel()), true));
```
rpastrana, line 83: This reference to esploglevel should be removed, if outside the scope of this PR, please open jira and post here.

esp/logging/logginglib/logthread.cpp
```diff
@@ -181,9 +182,14 @@ bool CLogThread::queueLog(IEspUpdateLogRequest* logRequest)
 
 bool CLogThread::queueLog(IEspUpdateLogRequestWrap* logRequest)
 {
-    unsigned startTime = (getEspLogLevel()>=LogNormal) ? msTick() : 0;
+// ESP Log Update
+// Would be good to find a way to delay calculation of delta time in case log message is filtered
+// Look for msTick() - startTime elsewhere
+//    unsigned startTime = (getEspLogLevel()>=LogNormal) ? msTick() : 0;
+    unsigned startTime = msTick();
```
rpastrana, line 189: Likely ok, but is the overhead of the msTick call warranted? @wangkx -thoughts?

esp/logging/logginglib/logthread.cpp
```diff
@@ -255,19 +266,25 @@ void CLogThread::sendLog()
...
+                    unsigned startTime1 = msTick();
                     logRequestReader->addACK(GUID);
                     PROGLOG("%s acked: %dms\n", GUID, msTick() -  startTime1);
                 }
                 else if(ensureFailSafe && logFailSafe.get())
                 {
-                    unsigned startTime1 = (getEspLogLevel()>=LogNormal) ? msTick() : 0;
+// ESP Log Update
+//                    unsigned startTime1 = (getEspLogLevel()>=LogNormal) ? msTick() : 0;
+                    unsigned startTime1 = msTick();
```
rpastrana, line 284: can we make this var name more self documenting?

esp/logging/test/main.cpp
```diff
@@ -24,7 +24,8 @@
 //Copy the code to esdl_binding.cpp.
 
 #ifdef USE_TEST_RESP
-                ESPLOG(LogMin,"origResp: %s", origResp.str());
+//                ESPLOG(LogMin,"origResp: %s", origResp.str());
+                LOG(LegacyMsgCat_Min,"origResp: %s", origResp.str());
```
rpastrana, line 28: This is a good example of a logging entry which the developer chooses to forcefully output the statement, let's add a appropriate threshold which the developer could use 
ie.
LOG(Forced_Threshold, "some important message")

wangkx (commented):
@kenrowland a few comments. I didn't review esdl related code.

esp/logging/logginglib/logthread.cpp
wangkx: The msTick call does add the overhead. I wish we can query the log level here.

esp/services/espcontrol/ws_espcontrolservice.cpp
```diff
@@ -223,7 +222,20 @@ bool CWSESPControlEx::onGetLoggingSettings(IEspContext& context, IEspGetLoggingS
...
+        LogLevel level = m_container->getLogLevel();
+
+#ifdef _CONTAINERIZED
+        switch (level)
+        {
+            case ErrMsgThreshold:   level = LogNone;    break;
+            case WarnMsgThreshold:  level = LogMin;     break;
+            case InfoMsgThreshold:  level = LogNormal;  break;
+            case DebugMsgThreshold: level = LogMax;     break;
+        }
```
timothyklemm, line 235: Is this an inclusive mapping, or could something else (like progress) go unmapped?

timothyklemm (commented):
My main takeaways are:

- DBGLOG should be unchanged unless the category is dependent on context. Same for PROGLOG, and any other jlog functions.
- ESPLOG implementation should be changed to map level to jlog category, then just use jlog if the current threshold is >= the category's level.
- ESPLOG usage should not have to change, in our code or anyone else's. If impractical to work the same in bare metal and containers, the ESPLOG implementation should handle that difference.

esp/logging/logginglib/logthread.cpp
timothyklemm: We can. It's just call a threshold now.

esp/platform/espp.hpp
```diff
@@ -28,6 +28,13 @@ typedef ISocket * isockp;
 typedef MapBetween<int, int, isockp, isockp> SocketPortMap;
 typedef CopyReferenceArrayOf<ISocket> SocketPortArray;
 
+class CEspElevatedLoggingFilter : public CategoryLogMsgFilter
+{
+    public:
+        CEspElevatedLoggingFilter() : CategoryLogMsgFilter(MSGAUD_all, MSGCLS_elevated, DebugMsgThreshold, false) {}
+        bool mayIncludeCategory(const LogMsgCategory & cat) const { return (cat.queryAudience() & audienceMask) && (cat.queryClass() & classMask) && (cat.queryDetail() <= maxDetail); }
```
rpastrana, line 35: I'm still not sure why this doesn't just return true as long as the class matches?

esp/scm/esp.ecm
```diff
@@ -616,3 +617,18 @@ SCMinterface IEspNgServiceBinding(IInterface)
 
 
 };
+
+
+#define LogNone_Threshold   CriticalMsgThreshold
+#define LogMin_Threshold    ErrMsgThreshold
+#define LogNormal_Threshold ProgressMsgThreshold
+#define LogMax_Threshold    DebugMsgThreshold
```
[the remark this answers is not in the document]
rpastrana: Well, I believe LogMax_Threshold is meant to represent the maximum logging threshold not necessarily the jlog threshold named "debug"

rpastrana (changes requested):
@kenrowland current commit reviewed. Other issues still outstanding.

esp/platform/espp.hpp
rpastrana: return cat.queryClass() & classMask;

esp/scm/ws_espcontrol.ecm
```diff
@@ -17,9 +17,17 @@
...
+    None(0, "none"),
+    Min(1, "min"),
+    Normal(5, "normal"),
+    Max(10, "max"),
+};
+
 ESPrequest [nil_remove] SetLoggingRequest
 {
-    int LoggingLevel;
+    ESPenum LogLevel LoggingLevel;
```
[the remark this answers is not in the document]
rpastrana: pseudo code discussed offline
[max_ver(1.34)]int LoggingLevel;
[min_ver(1.35)]int LoggingLevelEnum;

esp/platform/espp.cpp
```diff
@@ -332,37 +331,31 @@ ILogMsgHandler * openEspLogFile(IPropertyTree* envpt, IPropertyTree* procpt)
         lf->setAliasName("esp");
         lf->setMaxLogFileSize(maxLogFileSize);
 
-        //
         // If log level defined in legacy config, honor it
         int loglev = -1;
         if(procpt->hasProp("@logLevel"))
             loglev = procpt->getPropInt("@logLevel", LogMin);
 
+        // If log level is valid, set max detail, otherwise use system default set in jlog
```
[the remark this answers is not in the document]
kenrowland: LogMin looks to be the default for legacy bare metal.

esp/platform/espp.cpp
```diff
@@ -298,14 +329,38 @@ void openEspLogFile(IPropertyTree* envpt, IPropertyTree* procpt)
...
         lf->setMaxLogFileSize(maxLogFileSize);
-        lf->beginLogging();
+
+        //
+        // If log level defined in legacy config, honor it
+        int loglev = -1;
+        if(procpt->hasProp("@logLevel"))
+            loglev = procpt->getPropInt("@logLevel", LogMin);
+
+        if (loglev > -1)
```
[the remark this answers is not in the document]
kenrowland: It does.

esp/platform/espp.hpp
kenrowland: Remove the audience and threshold tests

esp/scm/esp.ecm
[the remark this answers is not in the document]
kenrowland: Changed to extraneous

esp/bindings/http/platform/httpbinding.cpp
```diff
@@ -799,7 +802,9 @@ bool EspHttpBinding::basicAuth(IEspContext* ctx)
             {
                 const char *desc=curres->getDescription();
                 VStringBuffer msg("Access for user '%s' denied to: %s. Access=%d, Required=%d", user->getName(), desc?desc:"<no-desc>", access, required);
-                ESPLOG(LogMin, "%s", msg.str());
+// ESP Log Update
+//                ESPLOG(LogMin, "%s", msg.str());
+                LOG(LegacyMsgCat_Max, "%s", msg.str());
```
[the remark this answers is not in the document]
kenrowland: It shouldn't have.

esp/bindings/http/platform/httpbinding.cpp
```diff
@@ -946,10 +959,12 @@ int EspHttpBinding::onGet(CHttpRequest* request, CHttpResponse* response)
...
     // the user authenticated
-    LogLevel level = getEspLogLevel(&context);
-    if (level >= LogNormal)
-        DBGLOG("EspHttpBinding::onGet");
-    
+// ESP Log Update
+//    LogLevel level = getEspLogLevel(&context);
+//    if (level >= LogNormal)
+//        DBGLOG("EspHttpBinding::onGet");
+    LOG(LegacyMsgCat_Max, "EspHttpBinding::onGet");
```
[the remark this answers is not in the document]
kenrowland: Early change that was not updated. Fixed it to use the category.

esp/platform/espcontext.cpp
```diff
@@ -1020,20 +1028,53 @@ unsigned getSlowProcessingTime()
...
+    return existingCategory;
 }
 
-void ESPLOG(IEspContext* ctx, LogLevel level, const char* fmt, ...)
+
+// #infdef _CONTAINERIZED ????
+void ESPLOG(LogLevel level, const char* fmt, ...)
 {
-    if (getEspLogLevel(ctx)>=level)
+    if (getEspLogLevel(NULL)>=level)
```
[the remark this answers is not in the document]
kenrowland: ESPLOG is also being deprecated.

esp/bindings/http/platform/httpbinding.cpp
```diff
@@ -864,11 +869,15 @@ bool EspHttpBinding::sendFromESPCache(IEspCache* cacheClient, CHttpRequest* requ
 {
     StringBuffer content, contentType;
     if (!cacheClient->readResponseCache(cacheID, content.clear(), contentType.clear()))
-        ESPLOG(LogMax, "Failed to read from ESP Cache for %s.", request->queryServiceMethod());
+        LOG(LegacyMsgCatMax, "Failed to read from ESP Cache for %s.", request->queryServiceMethod());
+// ESP Log Update
+        LOG(LegacyMsgCatMax, "Failed to read from ESP Cache for %s.", request->queryServiceMethod());
```
rpastrana, line 874: this line seems to duplicate line 872

esp/bindings/http/client/httpclient.cpp
rpastrana: Either update https://track.hpccsystems.com/browse/HPCC-27215 to include  Persistenthandler, or open a new jira and post here

esp/bindings/http/platform/httpservice.cpp
```diff
@@ -1785,18 +1797,25 @@ void CEspHttpServer::resetSessionTimeout(EspAuthRequest& authReq, unsigned sessi
         addCookie(authReq.authBinding->querySessionIDCookieName(), sessionIDStr.str(), 0, true);
         addCookie(SESSION_AUTH_OK_COOKIE, "true", 0, false); //client can access this cookie.
 
-        if (getEspLogLevel()>=LogMax)
+// ESP Log Update (in this case Both log entries are being sent out because of removing the log level comparison
+//        if (getEspLogLevel()>=LogMax)
```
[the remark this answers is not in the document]
rpastrana: Tim's issue doesn't seem to have been addressed.

esp/logging/loggingagent/cassandraloggingagent/cassandralogagent.cpp
```diff
@@ -243,7 +243,12 @@ void CCassandraLogAgent::setUpdateLogStatement(const char* dbName, const char* t
 
 void CCassandraLogAgent::executeSimpleStatement(const char* st)
 {
-    CassandraStatement statement(cassSession->prepareStatement(st, getEspLogLevel()>LogNormal));
+#ifdef _DEBUG
+    bool trace = true;
+#else
+    bool trace = false;
+#endif
```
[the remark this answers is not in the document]
rpastrana: As Tim points out, it does seem getEspLogLevel()>LogNormal could be replaced by level check.
also, if we're keeping this trace flag, let's give it a more meaningful name, and consider exporting at the module level (other function seems to perform similar check)

esp/logging/logginglib/LogFailSafe.cpp
```diff
@@ -78,7 +78,8 @@ CLogFailSafe::CLogFailSafe(IPropertyTree* cfg, const char* pszService, const cha
 
 CLogFailSafe::~CLogFailSafe()
 {
-    ESPLOG(LogMax, "CLogFailSafe::~CLogFailSafe()");
+//    ESPLOG(LogMax, "CLogFailSafe::~CLogFailSafe()");
+    LOG(LegacyMsgCatMax, "CLogFailSafe::~CLogFailSafe()");
```
rpastrana, line 82: @wangkx can confirm but this seems like a good candidate for DBGLOG

esp/logging/logginglib/logthread.cpp
```diff
@@ -181,9 +182,15 @@ bool CLogThread::queueLog(IEspUpdateLogRequest* logRequest)
 
 bool CLogThread::queueLog(IEspUpdateLogRequestWrap* logRequest)
 {
-    unsigned startTime = (getEspLogLevel()>=LogNormal) ? msTick() : 0;
+// ESP Log Update
+// Would be good to find a way to delay calculation of delta time in case log message is filtered
+// Look for msTick() - startTime elsewhere
+//    unsigned startTime = (getEspLogLevel()>=LogNormal) ? msTick() : 0;
+    bool isLogging = !queryLogMsgManager()->rejectsCategory(LegacyMsgCatNormal);
```
rpastrana, line 189: This seems reasonable but we should avoid LegacyMsgCat if possible.
Is there a pre-existing jlog category we could use here @wangkx thoughts?

esp/logging/logginglib/logthread.cpp
```diff
@@ -192,7 +199,10 @@ bool CLogThread::enqueue(IEspUpdateLogRequestWrap* logRequest, const char* guid)
     if (!logRequestReader && logFailSafe.get())
     {
         StringBuffer GUID, reqBuf;
-        unsigned startTime = (getEspLogLevel()>=LogNormal) ? msTick() : 0;
+
+//        unsigned startTime = (getEspLogLevel()>=LogNormal) ? msTick() : 0;
+        bool isLogging = !queryLogMsgManager()->rejectsCategory(LegacyMsgCatNormal);
```
rpastrana, line 204: same comment as above

esp/logging/logginglib/logthread.cpp
```diff
@@ -573,11 +602,17 @@ void CLogThread::writeJobQueue(IEspUpdateLogRequestWrap* jobToWrite)
 IEspUpdateLogRequestWrap* CLogThread::readJobQueue()
 {
 #define LOG_LEVEL LogNormal
-    unsigned startTime = (getEspLogLevel()>=LOG_LEVEL) ? msTick() : 0;
+// ESP Log Update
```
rpastrana, line 605: seems like L604 & L617 is no longer used?

esp/logging/logginglib/logthread.cpp
```diff
@@ -612,12 +647,14 @@ void CLogRequestReader::threadmain()
     unsigned waitMillSeconds = 1000*settings->waitSeconds;
     while (!stopping)
     {
-        ESPLOG(LogMax, "#### CLogRequestReader: the loop for reading log requests begins.");
+//        ESPLOG(LogMax, "#### CLogRequestReader: the loop for reading log requests begins.");
+        LOG(LegacyMsgCatMax, "#### CLogRequestReader: the loop for reading log requests begins.");
```
rpastrana, line 651: A lot of these statements seem like they should be using DBGLOG

esp/logging/loggingmanager/loggingmanager.cpp
```diff
@@ -340,7 +340,9 @@ bool CLoggingManager::saveToTankFile(IEspUpdateLogRequestWrap& logRequest, CLogR
         return false;
     }
 
-    unsigned startTime = (getEspLogLevel()>=LogNormal) ? msTick() : 0;
+// ESP Log Update
+//    unsigned startTime = (getEspLogLevel()>=LogNormal) ? msTick() : 0;
+    unsigned startTime = msTick();
```
rpastrana, line 345: queryLogMsgManager()->rejectsCategory?

esp/scm/esp.ecm
```diff
@@ -616,3 +617,18 @@ SCMinterface IEspNgServiceBinding(IInterface)
 
 
 };
+
+
+#define LogNone_Threshold   CriticalMsgThreshold
```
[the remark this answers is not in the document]
rpastrana: fair enough. In that case let's label them LegacyThreshold or something similar to clearly document they're legacy.

esp/platform/espcontext.cpp
```diff
@@ -1020,6 +1020,53 @@ unsigned getSlowProcessingTime()
...
+        return(LogMin_Threshold);
+    }
+    else if (level >= LogNormal && level < LogMax)
+    {
+        return(LogNormal_Threshold);
+    }
+    return(LogMax_Threshold);
+}
+
+LogMsgCategory getEspLogCategoryForContext(IEspContext* ctx, LogMsgCategory defaultLogMsgCategory)
```
rpastrana, line 1040: getEspLogCategory**_For_**Context or **_From_**Context

esp/scm/esp.ecm
```diff
@@ -616,3 +617,18 @@ SCMinterface IEspNgServiceBinding(IInterface)
...
+#define LogMin_Threshold    ErrMsgThreshold
```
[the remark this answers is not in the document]
rpastrana: It seems L623-L626 still have the _

esp/services/espcontrol/ws_espcontrolservice.cpp
```diff
@@ -192,8 +193,27 @@ bool CWSESPControlEx::onSetLogging(IEspContext& context, IEspSetLoggingRequest&
...
-        if (!req.getLoggingLevel_isNull())
-            m_container->setLogLevel(req.getLoggingLevel());
+        if (version >= 1.04)
+        {
+            unsigned requestedLevel = req.getLoggingLevelEnum();
+            unsigned espLevel;
+
+            switch (requestedLevel)
+            {
+                case 0:   espLevel = LogNone;    break;
```
rpastrana, line 203: The standard for the platform code seems to add new lines for all cases in a switch

system/jlib/jlog.hpp
```diff
@@ -59,16 +59,17 @@ typedef enum
...
+    MSGCLS_disaster    = 0x001, // Any unrecoverable or critical system errors
+    MSGCLS_error       = 0x002, // Recoverable/not critical Errors
+    MSGCLS_warning     = 0x004, // Warnings
+    MSGCLS_information = 0x008, // Config, environmental and internal status  info
+    MSGCLS_progress    = 0x010, // Progress of workunits. Status of file operations
+    MSGCLS_metric      = 0x020, // A metric line
+    MSGCLS_addid       = 0x040, // Internal use within log system
+    MSGCLS_removeid    = 0x080, // Internal use within log system
+    MSGCLS_all         = 0x0FF, // Use as a filter to select all messages except those below
+    MSGCLS_elevated    = 0x800, // Use for component elevated logging
```
rpastrana, line 72: it still feels wrong to define a bitmap value past the "ALL" mask, for record keeping sake, why aren't we expanding MSGCLS_all to 0xFFF.
also why isn't MSGCLS_elevated = 0x100,

system/jlib/jlog.hpp
```diff
@@ -249,6 +256,8 @@ typedef unsigned LogMsgDetail;
  * It represents the lowest logging level (detail) required to output
  * messages of the given category.
  */
+
+constexpr LogMsgDetail ForcedMsgThreshold      = 0;  //Use to declare categories that force output
```
rpastrana, line 260: This doesn't seem to be used

system/jlib/jlog.ipp
```diff
@@ -128,6 +128,7 @@ public:
     void                      addToPTree(IPropertyTree * tree) const;
     void                      orWithFilter(const ILogMsgFilter * filter);
     void                      reset();
+
```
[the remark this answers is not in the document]
rpastrana: please revert

rpastrana (changes requested):
@kenrowland looks good, left several comments

esp/logging/logginglib/LogFailSafe.cpp
wangkx: I do agree.

esp/logging/logginglib/logthread.cpp
wangkx: How about InfoMsgThreshold or DebugMsgThreshold?

wangkx (commented):
Looks good. I guess you missed https://github.com/hpcc-systems/HPCC-Platform/pull/15770#discussion_r806931713.

esp/logging/logginglib/logthread.cpp
rpastrana: @wangkx it's up to you, I believe LegacyMsgCatNormal is comprised of All Audience, and All class...
If we use InfoMsgThreshold or DebugMsgThreshold, it would be subject to be rejected based on the jlog configuration for Audience and/or class type.

esp/logging/logginglib/logthread.cpp
rpastrana: At this point we should address them in upcoming PR.

system/jlib/jlog.hpp
rpastrana: @kenrowland let's remove

esp/platform/espp.hpp
```diff
@@ -155,7 +169,7 @@ class CEspServer : public CInterface,
             m_waitForExit.signal();
     }
 
-    void setLogLevel(LogLevel level) { m_logLevel = level; }
+    void setLogLevel(LogLevel level); //{ m_logLevel = level; }
```
wangkx, line 172: Remove the commented code.

esp/scm/ws_espcontrol.ecm
```diff
@@ -17,13 +17,23 @@
...
+    None(0, "none"),
+    Min(1, "min"),
+    Normal(5, "normal"),
+    Max(10, "max"),
+};
+
 ESPrequest [nil_remove] SetLoggingRequest
 {
-    int LoggingLevel;
+    [max_ver("1.03")] int LoggingLevel;
```
wangkx, line 30: Change to [depr_ver("1.04")] int LoggingLevel for accuracy

esp/scm/ws_espcontrol.ecm
```diff
@@ -36,7 +46,8 @@ ESPrequest [nil_remove] GetLoggingSettingsRequest
 
 ESPresponse [exceptions_inline, nil_remove, http_encode(0)] GetLoggingSettingsResponse
 {
-    int LoggingLevel;
+    [max_ver("1.03")] int LoggingLevel;
```
wangkx, line 49: same as the line 30

wangkx (commented):
A few comments.

esp/bindings/http/platform/httpbinding.cpp
```diff
@@ -946,10 +944,8 @@ int EspHttpBinding::onGet(CHttpRequest* request, CHttpResponse* response)
 
     // At this time, the request is already received and fully passed, and
     // the user authenticated
-    LogLevel level = getEspLogLevel(&context);
-    if (level >= LogNormal)
-        DBGLOG("EspHttpBinding::onGet");
-    
+    LOG(getEspLogCategoryForContext(&context, LegacyMsgCatMax), "EspHttpBinding::onGet");
```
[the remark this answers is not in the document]
kenrowland: The previous code retrieved the log level based on the context 

LogLevel level = getEspLogLevel(&context);

Then used it in the comparison. So, the existing DBGLOG statement was changed to use the category returned from the context in order to maintain compatibility.

esp/bindings/http/platform/httpservice.cpp
```diff
@@ -1785,17 +1785,18 @@ void CEspHttpServer::resetSessionTimeout(EspAuthRequest& authReq, unsigned sessi
...
-        if (getEspLogLevel()>=LogMax)
+        if (!queryLogMsgManager()->rejectsCategory(MCuserProgress))
         {
             CDateTime timeoutAtCDT;
             StringBuffer timeoutAtString, nowString;
             timetToIDateTime(&timeoutAtCDT, timeoutAt);
             PROGLOG("Reset %s for (/%s/%s) at <%s><%ld> : expires at <%s><%ld>", PropSessionTimeoutAt,
                 authReq.serviceName.isEmpty() ? "" : authReq.serviceName.str(), authReq.methodName.isEmpty() ? "" : authReq.methodName.str(),
                 now.getString(nowString).str(), createTime, timeoutAtCDT.getString(timeoutAtString).str(), timeoutAt);
+
```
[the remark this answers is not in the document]
kenrowland: Removed

esp/bindings/http/platform/httpservice.cpp
```diff
@@ -2016,8 +2017,9 @@ EspAuthState CEspHttpServer::authExistingSession(EspAuthRequest& authReq, unsign
     {
         time_t timeoutAt = createTime + authReq.authBinding->getServerSessionTimeoutSeconds();
         sessionTree->setPropInt64(PropSessionTimeoutAt, timeoutAt);
-        ESPLOG(LogMin, "Updated %s for (/%s/%s) : %ld", PropSessionTimeoutAt, authReq.serviceName.isEmpty() ? "" : authReq.serviceName.str(),
-            authReq.methodName.isEmpty() ? "" : authReq.methodName.str(), timeoutAt);
+        LOG(LegacyMsgCatMin, "Updated %s for (/%s/%s) : %ld", PropSessionTimeoutAt, authReq.serviceName.isEmpty() ? "" : authReq.serviceName.str(),
+               authReq.methodName.isEmpty() ? "" : authReq.methodName.str(), timeoutAt);
+
```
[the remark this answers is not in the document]
kenrowland: Removed

esp/platform/espp.hpp
kenrowland: Removed

esp/scm/ws_espcontrol.ecm
kenrowland: Changed.

esp/scm/ws_espcontrol.ecm
kenrowland: Changed

esp/platform/espp.hpp
```diff
@@ -20,13 +20,21 @@
...
 
 
 typedef MapBetween<int, int, isockp, isockp> SocketPortMap;
 typedef CopyReferenceArrayOf<ISocket> SocketPortArray;
 
+class CEspElevatedLoggingFilter : public CategoryLogMsgFilter
+{
+public:
+    CEspElevatedLoggingFilter() : CategoryLogMsgFilter(MSGAUD_all, MSGCLS_elevated, DebugMsgThreshold, false) {}
+    bool mayIncludeCategory(const LogMsgCategory & cat) const { return (cat.queryClass() & classMask); }
```
timothyklemm, line 35: Why does the check not consider the message level? Values > DebMsgThreshold should be excluded.

esp/services/ws_sql/ws_sqlService.cpp
```diff
@@ -922,17 +922,17 @@ bool CwssqlEx::onExecuteSQL(IEspContext &context, IEspExecuteSQLRequest &req, IE
...
-        ESPLOG(LogMax, "WsSQL: getWorkUnitFactory...");
+        DBGLOG("WsSQL: getWorkUnitFactory...");
         Owned<IWorkUnitFactory> factory = getWorkUnitFactory(context.querySecManager(), context.queryUser());
 
-        ESPLOG(LogMax, "WsSQL: checking query cache...");
+        DBGLOG("WsSQL: checking query cache...");
         if(cacheeligible && getCachedQuery(normalizedSQL.str(), compiledwuid.s))
         {
-           ESPLOG(LogMax, "WsSQL: cache hit opening wuid %s...", compiledwuid.str());
+            DBGLOG("WsSQL: cache hit opening wuid %s...", compiledwuid.str());
```
timothyklemm, line 931: Nitpicking, but extra leading space.

esp/services/ws_store/ws_storeService.cpp
```diff
@@ -223,7 +223,7 @@ bool CwsstoreEx::onDeleteNamespace(IEspContext &context, IEspDeleteNamespaceRequ
 
     if (!global && !isEmptyString(targetUser))
     {
-        ESPLOG(LogMin, "CwsstoreEx::onDeleteNamespace: '%s' requesting to delete namespace on behalf of '%s'", user, targetUser);
+        AERRLOG("CwsstoreEx::onDeleteNamespace: '%s' requesting to delete namespace on behalf of '%s'", user, targetUser);
```
timothyklemm, line 226: This doesn't seem like an error message, and misclassifying non-errors as errors might not be a good idea.

esp/platform/espp.hpp
```diff
@@ -182,14 +196,13 @@ class CEspServer : public CInterface,
...
     {
-        if (getLogLevel()>=level)
+        if (m_elevatedLoggingFilter == nullptr)
         {
-            va_list args;
-            va_start(args, fmt);
-            VALOG(MCdebugInfo, unknownJob, fmt, args);
-            va_end(args);
+            // Add custom log monitor to handle elevated context based logging
+            m_elevatedLoggingFilter = new CEspElevatedLoggingFilter();
```
timothyklemm, line 204: Does this need to be thread safe?

kenrowland: Unlikely would be a problem. However, there is a small window where two requests could each be the first requesting elevated logging at the same time. @rpastrana Should we make this threadsafe?

esp/platform/espp.hpp
rpastrana: why would we want to exclude any message based on threshold? would that follow existing behavior?
DebMsgThreshold  is 80, DBGLOG() in baremetal I believe logs all entries.

esp/services/ws_store/ws_storeService.cpp
rpastrana: good point, this really should use LOG(MCauditWarning

esp/services/ws_sql/ws_sqlService.cpp
rpastrana: should be fixed here: https://track.hpccsystems.com/browse/HPCC-27246

esp/bindings/http/platform/httpbinding.cpp
```diff
@@ -674,14 +674,11 @@ void EspHttpBinding::populateRequest(CHttpRequest *request)
         ISecResourceList* settinglist = m_setting_authmap->getResourceList("*");
         if(settinglist == NULL)
             return ;
-        if (getEspLogLevel()>=LogMax)
-        {
-            StringBuffer s;
-            DBGLOG("Set security settings: %s", settinglist->toString(s).str());
-        }
+        StringBuffer s;
```
afishbeck, line 677: These kinds of examples are very unfortunate.  We will now be doing the work to build this string, and using up memory, for every request.  In the past it would very rarely be enabled and the work and memory would be avoided.

"settinglist->toString(s).str()" is a complete serialization of the list content.

esp/platform/espp.hpp
timothyklemm: Audience isn't checked because the filter accepts any. Class is checked for reasons obvious to me. I'm unsure why level isn't checked when `cat` may specify a higher value.

I am admittedly unfamiliar with filters, and maybe this does make sense. I'm really just asking for an explanation so I can understand it.

esp/platform/espcontext.cpp
```diff
@@ -1020,6 +1020,53 @@ unsigned getSlowProcessingTime()
...
+            {
+                const char* mode = sec->getValue();
+                if ( mode && (streq(mode,"1") || streq(mode, "true")) )
+                {
+#ifdef _CONTAINERIZED
+
+                    if (getContainer())
+                    {
+                        getContainer()->enableElevatedLogging();
+                        return MsgCatEspElevated;
```
afishbeck, line 1058: Should this result be cached in the ctx and only looked up in the security settings once?

esp/platform/espp.hpp
afishbeck: I would.

esp/services/ws_workunits/ws_workunitsQuerySets.cpp
```diff
@@ -3816,11 +3816,11 @@ bool CWsWorkunitsEx::onWUQueryGetSummaryStats(IEspContext& context, IEspWUQueryG
...
         }
 
-        if (getEspLogLevel() >= LogMax)
         {
             StringBuffer sb;
-            toXML(queryAggregates, sb);
-            DBGLOG("getQueryStats(): '%s' => '%s'", control.str(), sb.str());
+            if (!queryLogMsgManager()->rejectsCategory(LegacyMsgCatMax))
+                toXML(queryAggregates, sb);
+            LOG(LegacyMsgCatMax, "getQueryStats(): '%s' => '%s'", control.str(), sb.str());
```
afishbeck, line 3823: I would put the LOG in the same if with the toXML.  No reason to log the empty string.

esp/bindings/http/platform/httpbinding.cpp
afishbeck: There aren't many like this.  If you replaced the original if with:
```
if (!queryLogMsgManager()->rejectsCategory(LegacyMsgCatMax))
```
I assume that would default to rejecting the category, and the user would have to enable that category to log this expensive content?

esp/platform/espp.hpp
rpastrana: yep, Ken's taking care of this per offline discussion

esp/platform/espp.cpp
```diff
@@ -15,6 +15,8 @@
     limitations under the License.
 ############################################################################## */
 
+//#define _CONTAINERIZED 1
```
rpastrana, line 18: Don't forget to remove this line

esp/platform/espp.hpp
```diff
@@ -20,13 +20,21 @@
...
 typedef ISocket * isockp;
 
 
 typedef MapBetween<int, int, isockp, isockp> SocketPortMap;
 typedef CopyReferenceArrayOf<ISocket> SocketPortArray;
 
+class CEspElevatedLoggingFilter : public CategoryLogMsgFilter
+{
+public:
+    CEspElevatedLoggingFilter() : CategoryLogMsgFilter(MSGAUD_all, MSGCLS_elevated, DebugMsgThreshold, false) {}
```
rpastrana, line 34: @kenrowland discussed @timothyklemm 's comment on line 35, and discovered we were actually discussing the DebugMsgThreshold passed in to the constructor. Ultimately I don't believe the detail is considered in the mayIncludeCategory, but for optics we should use a max detail threshold - I'd prefer a constructor with only class but not sure if that's possible.

esp/platform/espp.hpp
```diff
@@ -79,6 +79,7 @@ class CEspServer : public CInterface,
     Owned<IPropertyTree> applicationConfig;
     ILogMsgHandler * m_pLogMsgHandler;
     CEspElevatedLoggingFilter *m_elevatedLoggingFilter;
+    CriticalSection elevatedLoggingCs;
```
rpastrana, line 82: Even though some do not like to prefix members w/ a m_, it seems to be the standard in this file.

rpastrana (approved):
@rpastrana looks good to me. Left one very minor comment

esp/platform/espp.hpp
afishbeck: If the user is setup for debug logging, this will currently be called a lot.. until caching of the MsgCatEspElevated is done.  You could consider optimizing so you rarely need the crit.

By checking for nullptr twice for example:
```
        if (m_elevatedLoggingFilter == nullptr)
        {
            CriticalBlock b(m_elevatedLoggingCs);
            if (m_elevatedLoggingFilter == nullptr)
            {
                // Add custom log monitor to handle elevated context based logging
                m_elevatedLoggingFilter = new CEspElevatedLoggingFilter();
                queryLogMsgManager()->addMonitorOwn(queryStderrLogMsgHandler(), m_elevatedLoggingFilter);
            }
        }
```

kenrowland:
@richardkchapman  please merge.

wangkx (approved):

esp/platform/espcontext.hpp
```diff
@@ -136,9 +155,16 @@ esp_http_decl const unsigned int getTxSummaryGroup();
 esp_http_decl const unsigned int readTxSummaryGroup(const char* group);
 esp_http_decl bool getTxSummaryResourceReq();
 esp_http_decl unsigned getSlowProcessingTime();
+esp_http_decl LogMsgDetail mapLegacyEspLogLevelToJlogThreshold(LogLevel level);
 
+esp_http_decl LogMsgCategory getEspLogCategoryForContext(IEspContext* ctx, LogMsgCategory defaultLogMsgCategory);
```
ghalliday, line 160: This should probably be a const reference to the category - I'm not sure how gcc will pass otherwise.

esp/bindings/SOAP/client/soapclient.cpp
```diff
@@ -124,10 +124,7 @@ int CSoapClient::postRequest(const char* contenttype, const char* soapaction, IR
         multipart->serialize(contenttypestr, requeststr);
     }
 
-    if (getEspLogLevel(rpccall.queryContext())>LogNormal)
-    {
-        DBGLOG("Content type: %s", contenttypestr.str());
-    }
+    LOG(getEspLogCategoryForContext(rpccall.queryContext(), LegacyMsgCatMax), "Content type: %s", contenttypestr.str());
```
ghalliday, line 127: I'm checking my understanding.

Previously this would only log if the requested logging level was high, and then it would log at DebugMsgThreshold.
Now it will always log - either with category LegacyMsgCatMax, or MsgCatEspElevated;

If so I think this will have the side-effect that all messages above the normal detail level will now get a read lock in CLogMsgManager::rejectsCategory.

esp/platform/espcontext.cpp
```diff
@@ -1020,6 +1020,53 @@ unsigned getSlowProcessingTime()
...
+                    }
+#else
+                    return LegacyMsgCatForce;  // bare metal does not add a custom monitor
+#endif
+                }
+            }
+        }
+    }
```
ghalliday, line 1066: This doesn't do the same as the previous code in getEspLogLevel(), which checked getContainer()->getLogLevel();

esp/platform/espcontext.cpp
```diff
@@ -1020,6 +1020,53 @@ unsigned getSlowProcessingTime()
...
+        return(LogNormalLegacyThreshold);
+    }
+    return(LogMaxLegacyThreshold);
+}
+
+LogMsgCategory getEspLogCategoryForContext(IEspContext* ctx, LogMsgCategory defaultLogMsgCategory)
+{
+    if (ctx)
+    {
+        ISecPropertyList* properties = ctx->querySecuritySettings();
```
ghalliday, line 1044: This would be cleaner encapsulated as a function ctx->includeDebugLogging() or similar.  I'm not clear what the function could be called because I'm not sure what it is logically testing.

ghalliday (changes requested):
Some initial comments.  I am still struggling to understand exactly how the change should work - there isn't any clear documentation in the jira, or the source that I could find.  I'll return again tomorrow.

ghalliday (changes requested):
@kenrowland I think there are a few problems
- It isn't clear what the new model is.  E.g., getEspLogCategoryForContext() would be called.  And without that I'm not convinced it is any cleaner.
- The legacy categories are a backward step, and using illegal values for the tracing.  Each logging line needs to be specific in why it is being logged, and who would be interested in it.
I think there are examples of debug, progress, audit and probably others in this list.

- 
-

esp/bindings/http/platform/httpservice.cpp
```diff
@@ -2093,44 +2093,44 @@ EspAuthState CEspHttpServer::handleAuthFailed(bool sessionAuth, EspAuthRequest&
...
-        ESPLOG(LogMin, "Unlock failed: invalid user name or password.");
+        LOG(LegacyMsgCatMin, "Unlock failed: invalid user name or password.");
         sendLockResponse(false, true, "Invalid user name or password");
         return authTaskDone;
     }
 
     if (!sessionAuth)
     {
-        ESPLOG(LogMin, "Authentication failed: send BasicAuthentication.");
+        LOG(LegacyMsgCatMin, "Authentication failed: send BasicAuthentication.");
```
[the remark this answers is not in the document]
kenrowland: Actual log statement content not being addressed by this PR. A full assessment of logging is necessary.

esp/bindings/SOAP/client/soapclient.cpp
kenrowland: See comment in https://track.hpccsystems.com/browse/HPCC-24691

kenrowland (commented):
Please see answers in line. Also information added to https://track.hpccsystems.com/browse/HPCC-24691

kenrowland:
I will push a commit with the jlog.hpp conflict resolved once approved.

ghalliday (commented):
I added a comment to the jira, pasted below:
>>There are at least 3 ways that logging output is controlled at the moment:

- which logging lines should go to which output (jlog filters)
- enabling logging in specific areas (e.g. udp within roxie)
- controlling logging levels for individual queries

How the different components handle that depend partly on  whether they run single jobs or multiple jobs at the same time.  If they run a single job there is no need for query level control, so the jlog detail levels can be used to control output.

eclcc:
- here the logging level is used to control the amount of logging that is generated.  Lines are logged at different levels, and they can be displayed by adjusting --loglevel.  Conditional logging is mainly done through #defines, but would be much better done by dynamically testing an option.

Thor:
- probably logs too much.  I don't think it controls what is output by level.  I think most conditional logging is via #defines.

Roxie
- Uses all 3.  E.g. increased logging for individual areas - e.g. udp, and you can specify on a query what level of logging you want to generate.

I don't think restricting the code to use the jlog filters to control the output provides the flexibility needed to cover each of the 3 options.
<<

Taking the example line:
`LOG(getEspLogCategoryForContext(rpccall.queryContext(), LegacyMsgCatMax), "Content type: %s", contenttypestr.str());
`
There is no indication why the programmer should be calling getEspLogCategoryForContext() for that line.  Why not call that function for all logging?  The name of the function gives no clues.

I think this is probably an example of feature-controlled debugging e.g.,
```
   if (rpccall.queryContext()->traceDetailedProgress())
   {
        DBGLOG("Content type: %s", contenttypestr.str());
    }
```
Or something similar.  The name of the function should explain why the tracing line was being controlled

@kenrowland please can you close this PR and open a replacement one.  There are now too many comments - and any extra comments are likely to be lost.

kenrowland:
Closing and opening new PR per request from @ghalliday

PR #15829, opened 2022-02-25
HPCC-27250 Remove mutex protecting exception from CAsyncFor code

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jthread.cpp
+++ b/system/jlib/jthread.cpp
@@ -673,8 +673,7 @@ void CAsyncFor::For(unsigned num,unsigned maxatonce,bool abortFollowingException
             Do(0);
         return;
     }
-    Mutex errmutex;
-    IException *e=NULL;
+    AtomicShared<IException> e;
     Owned<IShuffledIterator> shuffler;
     if (shuffled) {
         shuffler.setown(createShuffledIterator(num));
@@ -689,10 +688,7 @@ void CAsyncFor::For(unsigned num,unsigned maxatonce,bool abortFollowingException
             }
             catch (IException * _e)
             {
-                if (e)
-                    _e->Release();  // only return first
-                else
-                    e = _e;
+                e.setownIfNull(_e);
                 if (abortFollowingException) 
                     break;
             }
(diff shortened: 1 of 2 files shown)


ghalliday:
Pushing to discuss the pattern + to provide a concrete example of it.

system/jlib/jthread.cpp
```diff
@@ -725,18 +719,13 @@ void CAsyncFor::For(unsigned num,unsigned maxatonce,bool abortFollowingException
...
-                            erre = _e;
+                        erre.setownIfNull(_e);
                     }
     #ifndef NO_CATCHALL
                     catch (...)
                     {
-                        synchronized block(*errmutex);
-                        if (!erre)
-                            erre = MakeStringException(0, "Unknown exception in Thread %s", getName());
+                        if (!erre.isSet())
```
richardkchapman, line 727: Doesn't the ifNull imply only if set already?

system/jlib/jthread.cpp
```diff
@@ -785,18 +772,13 @@ void CAsyncFor::For(unsigned num,unsigned maxatonce,bool abortFollowingException
...
```
richardkchapman, line 780: Same question here - this test feels superfluous

system/jlib/jthread.cpp
jakesmith: I think it's checking, to avoid the MakeStringException call/creation, which might be thrown away if erre is already.
Might not be worth saving the wasted creating and just call setownIfNull(makeStringException....) for clearer code?

system/jlib/jthread.cpp
```diff
@@ -837,8 +813,8 @@ void CAsyncFor::For(unsigned num,unsigned maxatonce,bool abortFollowingException
             }
         }
     }
-    if (e)
-        throw e;
+    if (e.isSet())
```
jakesmith, line 816: should AtomicShared have an operatr * (like Owned/Shared) so can just say if (e) ?

jakesmith (approved):
looks sensible to me.

system/jlib/jthread.cpp
ghalliday: That's how I had it to start off with.  I'll change it back.

system/jlib/jthread.cpp
ghalliday: I think it does - I'll check

ghalliday: I think there isn't a operator * because it isn't generally thread-safe - only if it was linked before returning, and even that isn't thread sfae (hence the getLinkNonAtomic).

jakesmith (approved):
@ghalliday - looks good, please squash.

richardkchapman (approved):

PR #15833, opened 2022-02-28
HPCC-27257 Add support for DFS client certificates

jakesmith:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dali/daliadmin/daliadmin.cpp
+++ b/dali/daliadmin/daliadmin.cpp
@@ -616,7 +616,7 @@ static void testDFSFile(IDistributedFile *legacyDfsFile, const char *logicalName
     const char *dir = legacyDfsFile->queryDefaultDir();
     PROGLOG("dir = %s", dir);
     const char *mask = legacyDfsFile->queryPartMask();
-    PROGLOG("mask = %s", dir);
+    PROGLOG("mask = %s", mask);
     IPropertyTree &attrs = legacyDfsFile->queryAttributes();
     legacyDfsFile->lockProperties();
     legacyDfsFile->unlockProperties();
--- a/esp/clients/ws_dfsclient/ws_dfsclient.cpp
+++ b/esp/clients/ws_dfsclient/ws_dfsclient.cpp
@@ -26,6 +26,7 @@
 
 #include "eclwatch_errorlist.hpp" // only for ECLWATCH_FILE_NOT_EXIST
 #include "soapmessage.hpp"
+#include "soapbind.hpp"
 
 #include "dafdesc.hpp"
 #include "dadfs.hpp"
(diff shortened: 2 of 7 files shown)


jakesmith:
Documentation JIRA: https://track.hpccsystems.com/browse/HPCC-27274

dali/daliadmin/daliadmin.cpp
```diff
@@ -616,7 +616,7 @@ static void testDFSFile(IDistributedFile *legacyDfsFile, const char *logicalName
     const char *dir = legacyDfsFile->queryDefaultDir();
     PROGLOG("dir = %s", dir);
     const char *mask = legacyDfsFile->queryPartMask();
-    PROGLOG("mask = %s", dir);
+    PROGLOG("mask = %s", mask);
```
jakesmith, line 619: NB: not really related/incidental fix - noticed during testing.

esp/clients/ws_dfsclient/ws_dfsclient.cpp
```diff
@@ -316,70 +317,74 @@ class CServiceDistributedFile : public CServiceDistributedFileBase<IDistributedF
...
-            // prefix. In particular, both a local and remote plane may want to use the same prefix/mount.
-            // So, the local plane will be defined with a unique prefix locally.
-            // Files backed by URL's or hostGroups will be access directly, are not mounted, and do not require
-            // this translation.
-            const char *filePlanePrefix = filePlane->queryProp("@prefix");
-            if (isAbsolutePath(filePlanePrefix) && !filePlane->hasProp("@hosts")) // otherwise assume url
+            Owned<IPropertyTree> remoteStorage = getRemoteStorage(remoteName);
+            if (!remoteStorage)
+                throw makeStringExceptionV(0, "Remote storage '%s' not found", remoteName);
+            if (!remoteStorage->getPropBool("@useDafilesrv"))
```
jakesmith, line 327: NB: the change in this function is mostly block movement. The significant change is the addition of this check.
@useDafileSrv can be defined as a property of the remote definition, meaning it will force a remap to the direct dafilesrv service, and avoid plane remapping.


It is intended for future use where bare-metal systems would connect/pull data off a dafilesrv not via mounts, but used during testing, reading files via daliadmin remotetest.

esp/clients/ws_dfsclient/ws_dfsclient.cpp
```diff
@@ -556,16 +561,64 @@ IClientWsDfs *getDfsClient(const char *serviceUrl, IUserDescriptor *userDesc)
...
 }
 
+static CriticalSection localSecretCrit;
+static void configureClientSSL(IEspClientRpcSettings &rpc, const char *secretName)
+{
+    /*
+     * This is a bit of a kludge, it gets the certificates from secrets, and writes them to local temp strorage.
+     * It does this so that it can pass the filename paths to rpc ssl / secure socket layer, which currently only
+     * accepts filenames, not binary blobs from memory.
+     */
```
jakesmith, line 571: @afishbeck - was a JIRA opened for securesocket for the work to get it to support certs/keys as in-memory blobs?

system/security/securesocket/securesocket.cpp
```diff
@@ -599,21 +599,11 @@ StringBuffer& CSecureSocket::get_cn(X509* cert, StringBuffer& cn)
 
 bool CSecureSocket::verify_cert(X509* cert)
 {
-    DBGLOG ("peer's certificate:\n");
-
```
jakesmith, line 603: NB: incidental change, verbose tracing removed. Noticed during testing.

esp/clients/ws_dfsclient/ws_dfsclient.cpp
afishbeck: Yes, https://track.hpccsystems.com/browse/HPCC-27255.  It's a sub task of the JIRA for getting TLS secrets from a hashicorp vault.

esp/clients/ws_dfsclient/ws_dfsclient.cpp
```diff
@@ -556,16 +561,64 @@ IClientWsDfs *getDfsClient(const char *serviceUrl, IUserDescriptor *userDesc)
     return dfsClient.getClear();
 }
 
+static CriticalSection localSecretCrit;
+static void configureClientSSL(IEspClientRpcSettings &rpc, const char *secretName)
```
afishbeck, line 565: Another possibility would be to add a new method to jsecrets:
```
void setRpcClientSecretInfo(IEspClientRpcSettings &rpc, const char *category, const char *secret);
```
That for now, sets up the client rpc settings with the actual paths to the secrets.  Later once I add binary blob support to IEspClientRpcSettings I can change the routine to set them using a blob.

jakesmith: similarly, my 1st cut off this workaround, added secret methods to get the direct path to the K8s secrets, and then use them directly (to avoid the save to local), but decided against it in the end, because it would mean you would not be able to use Vault secrets.
Ok, so a jsecrets method could detect where the secrets were coming from (K8s or Vault), and only have to save them to local if Vault, but as this is required for vault secrets anyway, and it's temporary/will be removed once HPCC-27255 is done, I think ok as is.

afishbeck: Agreed.  Yes, I'm definitely not a huge fan of calling jsecrets to get the paths... it seems to miss the point somehow and doing so through an interface as I described isn't that much better.

afishbeck (approved):
@jakesmith looks good.

esp/clients/ws_dfsclient/ws_dfsclient.cpp
```diff
@@ -556,16 +561,64 @@ IClientWsDfs *getDfsClient(const char *serviceUrl, IUserDescriptor *userDesc)
...
+    {
+        Owned<IFile> dir = createIFile(tempDirStr.str());
+        dir->createDirectory();
+        StringBuffer secretValue;
+
+        Owned<IFile> file = createIFile(clientCertFilename);
+        Owned<IFileIO> io = file->open(IFOcreate);
+        getSecretValue(secretValue, "storage", secretName, "tls.crt", true);
+        io->write(0, secretValue.length(), secretValue.str());
+
```
ghalliday, line 595: minor:  Should calls to write() be followed by io->close() to ensure flushed, and avoid any issues with disk full throwing exceptions??

ghalliday (approved):
@jakesmith one minor comment

esp/clients/ws_dfsclient/ws_dfsclient.cpp
jakesmith: agree, would be better. I'll add the io->close()'s

ghalliday (approved):
@richardkchapman

PR #15836, opened 2022-03-01
HPCC-27215 Remove dependency on log level and use jlog standards

kenrowland:
Removed use of log level in determining log output. Only using
jlog functions to output log statements.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Cloud-compatibility
  - [x] Premature optimization

## Smoketest:

### DIFF
--- a/system/security/securesocket/securesocket.cpp
+++ b/system/security/securesocket/securesocket.cpp
@@ -133,7 +133,6 @@ private:
     bool        m_verify;
     bool        m_address_match;
     CStringSet* m_peers;
-    int         m_loglevel;
     bool        m_isSecure;
     StringBuffer m_fqdn;
     size32_t    nextblocksize = 0;
@@ -150,8 +149,8 @@ private:
 public:
     IMPLEMENT_IINTERFACE;
 
-    CSecureSocket(ISocket* sock, SSL_CTX* ctx, bool verify = false, bool addres_match = false, CStringSet* m_peers = NULL, int loglevel=SSLogNormal, const char *fqdn = nullptr);
-    CSecureSocket(int sockfd, SSL_CTX* ctx, bool verify = false, bool addres_match = false, CStringSet* m_peers = NULL, int loglevel=SSLogNormal, const char *fqdn = nullptr);
+    CSecureSocket(ISocket* sock, SSL_CTX* ctx, bool verify = false, bool addres_match = false, CStringSet* m_peers = NULL, int loglevel=0, const char *fqdn = nullptr);
+    CSecureSocket(int sockfd, SSL_CTX* ctx, bool verify = false, bool addres_match = false, CStringSet* m_peers = NULL, int loglevel=0, const char *fqdn = nullptr);
     ~CSecureSocket();
 
     virtual int secure_accept(int logLevel);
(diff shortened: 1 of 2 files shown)


system/security/securesocket/securesocket.cpp
```diff
@@ -685,8 +682,7 @@ int CSecureSocket::secure_accept(int logLevel)
         return err;
     }
 
-    if (logLevel > SSLogNormal)
-        DBGLOG("SSL accept ok, using %s", SSL_get_cipher(m_ssl));
+    DBGLOG("SSL accept ok, using %s", SSL_get_cipher(m_ssl));
```
richardkchapman, line 685: Not sure what the brief is here - but calling DBGLOG unconditionally (and relying on the log level setting in jlog to decide what to output) can mean that additional work is done creating strings to output which are then thrown away...

rpastrana (changes requested):
@kenrowland these changes might result in unintended (or unexpected) flood of log output.
Let's meet offline to discuss.

system/security/securesocket/securesocket.cpp
kenrowland: For log statements that have expensive variables, there is a jlog function that can be called that to check if a log statement would be output. This provides a way to avoid building strings when not necessary and allows the component to rely on the logging levels set in jlog. This method was employed in recent changes to ESP in order to remove relying on an internally kept log level.

kenrowland: @richardkchapman  Here are some notes on the logging problem being solved when controlling logging 
with an input log level.

Secure sockets accept a log level during construction. The log level is compared in code prior to
calling a DBGLOG statement. The comparison provided a gate for filtering logging with the expectation
that the DBGLOG statement would be filtered out. This worked because the default threshold
in bare metal is set to debug. Mu understanding is that the threshold was never set reliably
in bare metal. This same comparison model was used in ESP.

This model will not work for cloud since the logging threshold, as well as audience and class masks,
are set from the yaml config files. Components can no longer rely on the default logging threshold set to
debug. Using the comparison model does not guarantee the DBGLOG statement will make it to the log.

Using the threshold set by the yaml config, here are 3 possible solutions:

1. Remove the input log level in favor of only providing log statements based on their intended logging
    threshold. This moves the control of log filtering out of the component to jlog. The component may 
    dynamically change the threshold using jlog as needed, or rely on the setting in the yaml config files 
    for container builds. Bare metal may need to set the jlog threshold at init time.

Keep the log level as a parameter to secure sockets thus allowing the component to separately control
the logging done by secure sockets. As described above, this will not reliably work in container builds
since the filtering is controlled by the level set in jlog through yaml config files. Motivation for using one
of these solutions depends on the desire to retain the ability to selectively control logging for certain
functionality in the component versus relying on the globally set logging threshold. Here are two ways 
to achieve this.

A. Define a new "forced" threshold for logging that is higher priority than critical. The new value,
    probably 0, would always be numerically less than critical and thus would never be filtered. Components can use
    the threshold in a logging statement to force our a log message.

B. Adopt the current ESP solution that uses a filter that looks for the elevated logging class. When
    encountered, the log message is passed through. Components can use this class when they want to force
    logging.

The ability to use a jlog function to determine if a log statement will be output can still be used with the above
solutions to avoid expensive log parameter evaluations.

PR #15837, opened 2022-03-02
HPCC-26035 Enable WU logs in container mode

rpastrana:
- Adds logic to create zap log file
- Adds logstream interface
- Utilizes ES scroll api
- Adds select column modes

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:


wangkx (commented):
@rpastrana looks good. Most of comments are minor.
I saw multiple DBGLOG calls. Please make sure they are needed.

rpastrana:
@wangkx please review latest comit 4c94aa0c23d6b592a72a2136743a54d30089882d

system/jlib/jlog.cpp
```diff
@@ -3482,69 +3482,78 @@ bool fetchLogByClass(StringBuffer & returnbuf, IRemoteLogAccess & logAccess, Log
...
 
+static IRemoteLogAccess * logAccessor = nullptr;
+static CriticalSection logAccessCrit;
+
 IRemoteLogAccess * queryRemoteLogAccessor()
 {
-    Owned<IPropertyTree> logAccessPluginConfig = getGlobalConfigSP()->getPropTree("logAccess");
+    if (!logAccessor)
+    {
+        CriticalBlock block(logAccessCrit);
```
wangkx, line 3492: Should this CriticalBlock be called before line 3490?

system/jlib/jlog.cpp
```diff
@@ -3482,69 +3482,78 @@ bool fetchLogByClass(StringBuffer & returnbuf, IRemoteLogAccess & logAccess, Log
...
-              searchColumn: "hpcc.log.audience"
-            - type: "class"
-              searchColumn: "hpcc.log.class"
-        )!!";
-        Owned<IPropertyTree> testTree = createPTreeFromYAMLString(simulatedGlobalYaml, ipt_none, ptr_ignoreWhiteSpace, nullptr);
-        logAccessPluginConfig.setown(testTree->getPropTree("global/logAccess"));
-    }
+        if (!logAccessPluginConfig)
+        {
+            const char *simulatedGlobalYaml = R"!!(global:
```
wangkx, line 3497: Add a space after the '*'? (see line 3485 and 3488)

system/jlib/jlog.cpp
```diff
@@ -3482,69 +3482,78 @@ bool fetchLogByClass(StringBuffer & returnbuf, IRemoteLogAccess & logAccess, Log
...
+        xproc = (newLogAccessPluginMethod_t_)GetSharedProcedure(logAccessPluginLib, instFactoryName);
+        if (xproc == nullptr)
+            throw makeStringExceptionV(-1, "%s cannot locate procedure %s in library '%s'", methodName, instFactoryName, libName.str());
 
-    //Call logaccessplugin instance factory and return the new instance
-    DBGLOG("Calling '%s' in log access plugin '%s'", instFactoryName, libName.str());
-    IRemoteLogAccess * pLogAccessPlugin = dynamic_cast<IRemoteLogAccess*>(xproc(*logAccessPluginConfig));
+        //Call logaccessplugin instance factory and return the new instance
+        DBGLOG("Calling '%s' in log access plugin '%s'", instFactoryName, libName.str());
+        if (!logAccessor)
```
wangkx, line 3550: If the 'CriticalBlock block(logAccessCrit);' is called before line 3490, you do not need to check the logAccessor here.

wangkx (commented):
@rpastrana 4 comments

rpastrana:
@ghalliday please review.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -205,6 +205,56 @@ WsWUExceptions::WsWUExceptions(IConstWorkUnit& wu): numerr(0), numwrn(0), numinf
...
+    }
+
+    if (cw->getStatistic(timeStamp, "", StWhenFinished))
+    {
+        formatStatistic(end.clear(), (timeStamp+6000000), SMeasureTimestampUs).str();
+        DBGLOG("WUID '%s' ended (adjusted) '%s'", wuid.str(), end.str()); //end + time buffer
+    }
+
+    struct LogAccessTimeRange range;
+    range.setStart(start.str());
```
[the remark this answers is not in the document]
rpastrana: setStart requires a const char *

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -205,6 +205,56 @@ WsWUExceptions::WsWUExceptions(IConstWorkUnit& wu): numerr(0), numwrn(0), numinf
...
+    if (!end.isEmpty())
+        range.setEnd(end.str());
```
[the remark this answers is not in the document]
rpastrana: setEnd requires a const char *

rpastrana:
@AttilaVamos FYI one of VCPKG (u20.04) tests reported this issue:
 Fetched 35.9 MB in 5s (7924 kB/s)
E: Failed to fetch http://security.ubuntu.com/ubuntu/pool/main/o/openjdk-lts/openjdk-11-jre-headless_11.0.13+8-0ubuntu1~20.04_amd64.deb  404  Not Found [IP: 52.252.75.106 80]
E: Failed to fetch http://security.ubuntu.com/ubuntu/pool/main/o/openjdk-lts/openjdk-11-jre_11.0.13+8-0ubuntu1~20.04_amd64.deb  404  Not Found [IP: 52.252.75.106 80]
E: Failed to fetch http://security.ubuntu.com/ubuntu/pool/main/o/openjdk-lts/openjdk-11-jdk-headless_11.0.13+8-0ubuntu1~20.04_amd64.deb  404  Not Found [IP: 52.252.75.106 80]
E: Failed to fetch http://security.ubuntu.com/ubuntu/pool/main/o/openjdk-lts/openjdk-11-jdk_11.0.13+8-0ubuntu1~20.04_amd64.deb  404  Not Found [IP: 52.252.75.106 80]
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?

esp/services/ws_workunits/ws_workunitsHelpers.cpp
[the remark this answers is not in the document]
wangkx: The StringBuffer start will be converted automatically to const char * by StringBuffer's inline operator.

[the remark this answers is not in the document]
rpastrana: Thanks @wangkx other than not being ideal for maintenance, is there an advantage worth making the change?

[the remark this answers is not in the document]
wangkx: It makes the code clear. I got warnings from code reviewers when I left the str() inside my code.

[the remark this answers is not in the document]
rpastrana: @wangkx thanks for clarification. It seems like a personal preference. It might be clear if the developer/maintainer has deep knowledge of The StringBuffer inline operator. The explicit .str() function call makes the code explicit and self documenting, hence without an added advantage in performance, we'll keep the code as is. Please note your code review conclusion

wangkx (approved):

rpastrana:
@AttilaVamos failed tests seem unrelated to changes
E: Failed to fetch http://security.ubuntu.com/ubuntu/pool/main/o/openjdk-lts/openjdk-11-jre-headless_11.0.13+8-0ubuntu1~20.04_amd64.deb  404  Not Found [IP: 40.119.46.219 80]
E: Failed to fetch http://security.ubuntu.com/ubuntu/pool/main/o/openjdk-lts/openjdk-11-jre_11.0.13+8-0ubuntu1~20.04_amd64.deb  404  Not Found [IP: 40.119.46.219 80]
E: Failed to fetch http://security.ubuntu.com/ubuntu/pool/main/o/openjdk-lts/openjdk-11-jdk-headless_11.0.13+8-0ubuntu1~20.04_amd64.deb  404  Not Found [IP: 40.119.46.219 80]
E: Failed to fetch http://security.ubuntu.com/ubuntu/pool/main/o/openjdk-lts/openjdk-11-jdk_11.0.13+8-0ubuntu1~20.04_amd64.deb  404  Not Found [IP: 40.119.46.219 80]
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
Error: Process completed with exit code 100.

ghalliday:
@rpastrana there are a few comments from kevin which were collapsed which I think you missed.

rpastrana:
> @rpastrana there are a few comments from kevin which were collapsed which I think you missed.

Thanks @ghalliday looked around and didn't see them. Can you point out what caught your eye so I can focus on addressing them? Thanks.

system/jlib/jlog.hpp
```diff
@@ -1449,11 +1449,20 @@ interface jlib_decl ILogAccessFilter : public IInterface
     virtual LogAccessFilterType filterType() const = 0;
 };
 
+typedef enum
```
ghalliday, line 1452: style: enum LogAccessReturnColsMode { }; is preferred - this style is old C style.

system/logaccess/ElasticStack/ElasticStackLogAccess.cpp
```diff
@@ -262,6 +266,46 @@ void processESJsonResp(const cpr::Response & retrievedDocument, StringBuffer & r
...
+        throw makeStringExceptionV(-1, "%s: Could not parse ElasticSearch query response", COMPONENT_NAME);
+
+    if (tree->getPropBool("timed_out", 0))
+        LOG(MCuserProgress,"ES Log Access: timeout reported");
+    if (tree->getPropInt("_shards/failed",0) > 0)
+        LOG(MCuserProgress,"ES Log Access: failed _shards reported");
+
+    DBGLOG("ES Log Access: hit count: '%d'", tree->getPropInt("hits/total/value"));
+
+    processHitsJsonResp(tree->getElements("hits/hits/fields"), returnbuf, format);
```
ghalliday, line 293: result of getElements will leak.  Worth running this code under valgrind (there is an option in the helm chart to enable that).

system/logaccess/ElasticStack/ElasticStackLogAccess.cpp
```diff
@@ -331,33 +375,51 @@ void esMatchQueryString(std::string & search, const char *searchval, const char
...
+            ForEachItemIn(idx, selectcols)
+            {
+                sourcecols.appendf("\"%s\"", selectcols.item(idx));
+                if (idx < selectcols.length() -1)
+                    sourcecols.append(",");
+            }
 
-        if (!sourcecols.isEmpty())
-            search += sourcecols.str() ;
+            if (!sourcecols.isEmpty())
```
ghalliday, line 409: redundant: isEmpty will never be true because of the test on line 399.

system/logaccess/ElasticStack/ElasticStackLogAccess.hpp
```diff
@@ -35,9 +35,32 @@
...
+    elasticlient::Scroll m_esSroller;
+
+    std::string m_queryString;
+    LogAccessLogFormat m_outputFormat;
+
+public:
+    static constexpr const char * DEFAULT_SCROLL_TIMEOUT = "1m"; //Elastic Time Units (i.e. 1m = 1 minute).
+    static constexpr std::size_t  DEFAULT_SCROLL_SIZE = 100;
+
+    virtual bool readLogEntries(StringBuffer & record) override;
```
ghalliday, line 56: minor:  Normally a class would have constructors, then destuctors, then virtuals to make it easy to read.

system/logaccess/ElasticStack/ElasticStackLogAccess.hpp
```diff
@@ -35,9 +35,32 @@
...
+    ElasticStackLogStream(std::string & queryString, const char * connstr, const char * indexsearchpattern, LogAccessLogFormat format, std::size_t scrollSize, std::string scrollTO);
```
ghalliday, line 57: Should scrollTO be scrollTo?

system/logaccess/ElasticStack/ElasticStackLogAccess.hpp
```diff
@@ -35,9 +35,32 @@
 
 #include <cpr/response.h>
 #include <elasticlient/client.h>
+#include <elasticlient/scroll.h>
 
 using namespace elasticlient;
 
+class ElasticStackLogAccess;
+
+class ELASTICSTACKLOGACCESS_API ElasticStackLogStream : public CInterfaceOf<IRemoteLogAccessStream>
```
ghalliday, line 44: encapsulation.  This class could be defined within the c++ file instead of the header.

system/logaccess/ElasticStack/ElasticStackLogAccess.cpp
```diff
@@ -520,11 +603,36 @@ cpr::Response ElasticStackLogAccess::performESQuery(const LogAccessConditions &
...
-    processESJsonResp(esresp, returnbuf, format);
+    processESSearchJsonResp(esresp, returnbuf, format);
 
     return true;
 }
 
+bool ElasticStackLogStream::readLogEntries(StringBuffer & record)
+{
+    Json::Value res;
+    Json::StreamWriterBuilder builder;
```
ghalliday, line 614: It is expensive to recreate this each time.  This should be a member, and builder.newStreamWriter() should be called once, and stored as a member, to avoid the overhead of creating that.

ghalliday (changes requested):
@rpastrana some comments, a few leaks and problems.

ghalliday:
@rpastrana I can't find any either - please ignore that comment!

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -205,6 +205,56 @@ WsWUExceptions::WsWUExceptions(IConstWorkUnit& wu): numerr(0), numwrn(0), numinf
...
+    logFetchOptions.setReturnColsMode(retColsMode);
+    logFetchOptions.setTimeRange(range);
+    logFetchOptions.setLimit(maxLogRecords);
+    logFetchOptions.setStartFrom((offset_t)0);
+
+    Owned<IFileIOStream> outIOS;
+    if (!isEmptyString(outFile))
+    {
+        CWsWuFileHelper helper(nullptr);
+        outIOS.setown(helper.createIOStreamWithFileName(outFile, IFOcreate));
```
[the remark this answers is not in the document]
wangkx: I just created https://track.hpccsystems.com/browse/HPCC-27348.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -205,6 +205,56 @@ WsWUExceptions::WsWUExceptions(IConstWorkUnit& wu): numerr(0), numwrn(0), numinf
...
+    }
+
+    if (cw->getStatistic(timeStamp, "", StWhenFinished))
+    {
+        formatStatistic(end, (timeStamp+wuLogSearchTimeBuff), SMeasureTimestampUs).str();
+        DBGLOG("Searching for WUID '%s' log entries ending time: '%s'", wuid.str(), end.str()); //end + time buffer
+    }
+
+    struct LogAccessTimeRange range;
+    range.setStart(start.str());
```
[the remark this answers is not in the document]
rpastrana: The query time frame is required, in which the start time is required, end can default to "now"
I can move both setStart and setEnd into their corresponding if (getstatistic)...
This begs the question, can we rely on getStatistic(StWhenCreated) ? should we develop a fall back?

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -205,6 +205,56 @@ WsWUExceptions::WsWUExceptions(IConstWorkUnit& wu): numerr(0), numwrn(0), numinf
         errors.append(*e.getLink());
     }
 }
+#ifdef _CONTAINERIZED
```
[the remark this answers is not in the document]
rpastrana: It doesn't, however there's no use case for it.
Harmless to include in baremetal, was seemingly following precedent but will remove conditional.

system/logaccess/ElasticStack/ElasticStackLogAccess.cpp
```diff
@@ -331,33 +375,51 @@ void esMatchQueryString(std::string & search, const char *searchval, const char
...
+                search += "\"*\"";
+                //search += "!*.keyword"; //all fields ignoring the .keyword postfixed fields
+        }
         else
-            search += "\"*\"";
-            //search += "!*.keyword"; //all fields ignoring the .keyword postfixed fields
+        {
+            throw makeStringExceptionV(-1, "%s: Custom return columns specified, but no columns provided", COMPONENT_NAME);
+        }
+    }
```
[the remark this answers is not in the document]
rpastrana: 100%
it actually was tested, but somehow fix included in subsequent PR: https://github.com/hpcc-systems/HPCC-Platform/blob/5b248590caa3e5d076a8e3b9b74cdc3fe8ece071/system/logaccess/ElasticStack/ElasticStackLogAccess.cpp#L491

https://github.com/hpcc-systems/HPCC-Platform/pull/15872/files
will ensure it is addressed here.

system/logaccess/ElasticStack/ElasticStackLogAccess.cpp
rpastrana: This sounds like a great feature the entire organization should exploit.
Even though it seems straightforward and simple to use, I did not see any documentation and the helm template wasn't obvious enough for me to utilize it.
It seemed roxie centric by the way. Anyway, can you provide "getting started" type info?
Perhaps startall.sh should provide an explicit valgrind option.

rpastrana (commented):
@ghalliday all your concerns should be accounted for, please take a look.
BTW, I did make a change to the way I calculate the query time frame (WU start - Xminutes) using CDateTime rather than attempting arithmetic operations directly on the epoch value.

system/jlib/jlog.cpp
```diff
@@ -2645,6 +2647,7 @@ MODULE_EXIT()
     thePassNoneFilter = nullptr;
     thePassLocalFilter = nullptr;
     thePassAllFilter = nullptr;
+    logAccessor->Release();
```
ghalliday, line 2650: I suspect this should be
`::Release(logAccessor)
`to remove the crash.

rpastrana: Done. Looked at the implementation, is it simply protecting the Release call from referencing nullptr?

AttilaVamos:
@rpastrana I am sure the reported workflow_contingency_8 failure is not related to this PR.

system/jlib/jlog.cpp
ghalliday: yes

system/logaccess/ElasticStack/ElasticStackLogAccess.cpp
```diff
@@ -214,9 +218,9 @@ void processHitsJsonResp(IPropertyTreeIterator * iter, StringBuffer & returnbuf,
 
             ForEach(*iter)
             {
-                IPropertyTree & cur = iter->query();
+                Owned<IPropertyTree> cur = &iter->get();
```
ghalliday, line 221: These changes seem to be unrelated (and undersirable/less efficient)

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -215,24 +215,36 @@ void WsWuInfo::readWorkunitComponentLogs(MemoryBuffer& buf, const char* outFile,
...
+    struct LogAccessTimeRange range;
     stat_type timeStamp;
     if (cw->getStatistic(timeStamp, "", StWhenCreated))
     {
-        formatStatistic(start, (timeStamp-wuLogSearchTimeBuff), SMeasureTimestampUs); //start - time buffer
-        DBGLOG("Searching for WUID '%s' log entries starting time: '%s'", wuid.str(), start.str());
+
+        CDateTime startt;
+        startt.set(timeStamp / MICROSECS_TO_SEC_DIVISOR);
+        startt.adjustTime(-wuLogSearchTimeBuff);
```
ghalliday, line 225: minor: You could have used adjustTimeSecs and kept the time in seconds

system/jlib/jlog.cpp
```diff
@@ -2645,6 +2647,7 @@ MODULE_EXIT()
     thePassNoneFilter = nullptr;
     thePassLocalFilter = nullptr;
     thePassAllFilter = nullptr;
+    ::Release(logAccessor);
```
ghalliday, line 2650: also set logAccessor to null - just to avoid problems if this is called twice.  (Unlikely but we've been chasing similar bugs elsewhere.)

system/logaccess/ElasticStack/ElasticStackLogAccess.cpp
```diff
@@ -241,9 +248,10 @@ void processESJsonResp(const cpr::Response & retrievedDocument, StringBuffer & r
         {
             ForEach(*iter)
             {
-                IPropertyTree & cur = iter->query();
+                Owned<IPropertyTree> cur = &iter->get();
```
jakesmith, line 251: minor: not sure why linking changed here.

jakesmith (commented):
@rpastrana - looks good, a few questions and mostly minor or trivial comments.

system/logaccess/ElasticStack/ElasticStackLogAccess.cpp
```diff
@@ -608,29 +611,58 @@ bool ElasticStackLogAccess::fetchLog(const LogAccessConditions & options, String
...
+        return false;
     }
 
-    return false;
+    ElasticStackLogStream(std::string & queryString, const char * connstr, const char * indexsearchpattern, LogAccessLogFormat format,  std::size_t scrollSize, std::string scrollTO)
+     : m_esSroller(std::make_shared<elasticlient::Client>(std::vector<std::string>({connstr})), scrollSize, scrollTO)
+    {
+        m_outputFormat = format;
+        m_esSroller.init(indexsearchpattern, DEFAULT_ES_DOC_TYPE, queryString);
+        m_jsonStreamBuilder.newStreamWriter();
```
ghalliday, line 639: The result of this function call should be saved as a StreamWriterPtr, then needs code like:

  std::ostringstream sout;
  writer->write(value, &sout);
  process(sout.str());

**and** writer->write(res, &sout);

system/logaccess/ElasticStack/ElasticStackLogAccess.cpp
```diff
@@ -520,11 +606,65 @@ cpr::Response ElasticStackLogAccess::performESQuery(const LogAccessConditions &
...
+
+    virtual ~ElasticStackLogStream() override
+    {
+        m_esSroller.clear();
+    }
+
+private:
+    elasticlient::Scroll m_esSroller;
+
+    std::string m_queryString;
```
ghalliday, line 650: Is this used for anything?

ghalliday (commented):
@rpastrana a few comments on the recent changes.  I haven't gone through and re-reviewed all.

rpastrana:
@ghalliday your comments/concerns should be addressed in latest commit

system/jlib/jlog.cpp
```diff
@@ -3484,67 +3488,72 @@ typedef IRemoteLogAccess * (*newLogAccessPluginMethod_t_)(IPropertyTree & logAcc
...
+        //Load the DLL/SO
+        HINSTANCE logAccessPluginLib = LoadSharedObject(libName.str(), true, false);
+        if(logAccessPluginLib == nullptr)
+            throw makeStringExceptionV(-1, "%s cannot load library '%s'", methodName, libName.str());
 
-    //Load the DLL/SO
-    HINSTANCE logAccessPluginLib = LoadSharedObject(libName.str(), true, false);
-    if(logAccessPluginLib == nullptr)
-        throw makeStringExceptionV(-1, "%s cannot load library '%s'", methodName, libName.str());
+        newLogAccessPluginMethod_t_ xproc = nullptr;
```
ghalliday, line 3544: trivial: clearer to combine these two lines.

system/jlib/jlog.cpp
```diff
@@ -3484,67 +3488,72 @@ typedef IRemoteLogAccess * (*newLogAccessPluginMethod_t_)(IPropertyTree & logAcc
...
-    xproc = (newLogAccessPluginMethod_t_)GetSharedProcedure(logAccessPluginLib, instFactoryName);
-    if (xproc == nullptr)
-        throw makeStringExceptionV(-1, "%s cannot locate procedure %s in library '%s'", methodName, instFactoryName, libName.str());
+        //Call logaccessplugin instance factory and return the new instance
+        DBGLOG("Calling '%s' in log access plugin '%s'", instFactoryName, libName.str());
 
-    //Call logaccessplugin instance factory and return the new instance
-    DBGLOG("Calling '%s' in log access plugin '%s'", instFactoryName, libName.str());
-    IRemoteLogAccess * pLogAccessPlugin = dynamic_cast<IRemoteLogAccess*>(xproc(*logAccessPluginConfig));
+        logAccessor = dynamic_cast<IRemoteLogAccess*>(xproc(*logAccessPluginConfig));
```
ghalliday, line 3552: readability: The dynamic cast is confusing.  xproc already returns a pointer of the correct type.

system/jlib/jlog.hpp
```diff
@@ -1466,10 +1475,20 @@ struct LogAccessConditions
         timeRange = l.timeRange;
         setFilter(LINK(l.filter));
```
ghalliday, line 1476: style: It is not clear that setFilter takes an owned pointer.  A more recent convention is for setFilter to link the argument, or if there is a good reason to pass an owned pointer name the function setFilterOwn (e.g. see LogMsgMonitor).  I know this is rule is broken a lot....but mainly in constructors rather than set() functions

system/logaccess/ElasticStack/ElasticStackLogAccess.cpp
```diff
@@ -520,11 +605,75 @@ cpr::Response ElasticStackLogAccess::performESQuery(const LogAccessConditions &
...
+            if (!res["total"]["value"].empty())
+            {
+                totalRecs = res["total"]["value"].asUInt();
+            }
+
+            if (!res["hits"].empty())
+            {
+                recsRead = res["hits"].size();
+                std::ostringstream sout;
+                m_jsonWriter->write(res, &sout);
```
ghalliday, line 634: design: This process is a bit convoluted, and was hard for me to follow.

I think the sequence is
* The value comes back in a Json object structure
* Written that value as json
* Read the json in as an IPropertyTree
* iterate elements in the IPropertyTree
* convert to json/csv/xml as required.

That may be the best way to code it, but it would be useful to have a comment clarifying that (I originally thought the json writer here was formatting the final output).  It is probably good for reusing code, but it is potentially quite inefficient - especially if these logs get large.

system/logaccess/ElasticStack/ElasticStackLogAccess.cpp
```diff
@@ -520,11 +605,75 @@ cpr::Response ElasticStackLogAccess::performESQuery(const LogAccessConditions &
...
+
+        return false;
+    }
+
+    ElasticStackLogStream(std::string & queryString, const char * connstr, const char * indexsearchpattern, LogAccessLogFormat format,  std::size_t pageSize, std::string scrollTO)
+     : m_esSroller(std::make_shared<elasticlient::Client>(std::vector<std::string>({connstr})), pageSize, scrollTO)
+    {
+        m_outputFormat = format;
+        m_esSroller.init(indexsearchpattern, DEFAULT_ES_DOC_TYPE, queryString);
+        m_jsonWriter = m_jsonStreamBuilder.newStreamWriter();
```
ghalliday, line 648: This will leak.  Need to use std::unique_ptr.  Worth running all of this code under valgrind to see if it creates any other leaks.

system/logaccess/ElasticStack/ElasticStackLogAccess.cpp
```diff
@@ -520,11 +605,75 @@ cpr::Response ElasticStackLogAccess::performESQuery(const LogAccessConditions &
...
+{
+    return getLogReader(options, format, DEFAULT_MAX_RECORDS_PER_FETCH);
+}
+
+IRemoteLogAccessStream * ElasticStackLogAccess::getLogReader(const LogAccessConditions & options, LogAccessLogFormat format, unsigned int pageSize)
+{
+    std::string queryString;
+    std::string queryIndex;
+    createQueryString(queryString, queryIndex, options);
+    return new ElasticStackLogStream(queryString, m_esConnectionStr.str(), m_globalIndexSearchPattern.str(), format, pageSize, DEFAULT_SCROLL_TIMEOUT);
```
ghalliday, line 674: minor: Better to pass queryIndex for m_globalIndexSearchPattern?  Otherwise it is confusing why it is returned.

ghalliday (changes requested):
@rpastrana I have gone through and reviewed it fully.  A few comments from trivial to more important.

system/logaccess/ElasticStack/ElasticStackLogAccess.hpp
```diff
@@ -73,13 +74,21 @@ class ELASTICSTACKLOGACCESS_API ElasticStackLogAccess : public CInterfaceOf<IRem
...
 
+    void esSearchMetaData(std::string & search, const LogAccessReturnColsMode retcolmode, const  StringArray & selectcols, unsigned size, offset_t from);
+    void getMinReturnColumns(std::string & columns);
+    void getDefaultReturnColumns(std::string & columns);
+    void getAllColumns(std::string & columns);
 public:
     ElasticStackLogAccess(const std::vector<std::string> &hostUrlList, IPropertyTree & logAccessPluginConfig);
     virtual ~ElasticStackLogAccess() override = default;
 
+    void createQueryString(std::string & queryString, std::string & queryIndex, const LogAccessConditions & options);
```
jakesmith, line 85: style/personal: the method name (and any method name createXX) suggests it will return something (the query string in this case), whereas queryIndex is an ancillary return value, so I think it would be better as:
```
    std::string createQueryString(const LogAccessConditions & options, std::string & queryIndex);
```
or even:
```
    std::pair<std::string, std::string> createQueryString(const LogAccessConditions & options);
```

system/logaccess/ElasticStack/ElasticStackLogAccess.hpp
```diff
@@ -73,13 +74,21 @@ class ELASTICSTACKLOGACCESS_API ElasticStackLogAccess : public CInterfaceOf<IRem
...
+
     // IRemoteLogAccess methods
     virtual bool fetchLog(const LogAccessConditions & options, StringBuffer & returnbuf, LogAccessLogFormat format) override;
     virtual const char * getRemoteLogAccessType() const override { return type; }
     virtual IPropertyTree * queryLogMap() const override { return m_pluginCfg->queryPropTree("logmap");}
     virtual const char * fetchConnectionStr() const override { return m_esConnectionStr.str();}
+    virtual IRemoteLogAccessStream * getLogReader(const LogAccessConditions & options, LogAccessLogFormat format) override;
+    virtual IRemoteLogAccessStream * getLogReader(const LogAccessConditions & options, LogAccessLogFormat format, unsigned int scrollSize) override;
```
jakesmith, line 93: trivial: you've changed scrollSize to pageSize elsewhere.

jakesmith (commented):
@rpastrana - a few more comments/questions

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -206,6 +206,71 @@ WsWUExceptions::WsWUExceptions(IConstWorkUnit& wu): numerr(0), numwrn(0), numinf
...
+    Owned<IRemoteLogAccessStream> logreader = m_remoteLogAccessor->getLogReader(logFetchOptions, logFormat);
+
+    StringBuffer logcontent;
+    unsigned totalRecsRead = 0;
+    unsigned recsRead = 0;
+    unsigned totalRecsReported = 0;
+    while (logreader->readLogEntries(logcontent.clear(), recsRead, totalRecsReported))
+    {
+        totalRecsRead += recsRead;
+        outputALine(logcontent.length(), logcontent.str(), buf, outIOS);
```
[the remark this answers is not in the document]
rpastrana: outputALine is a pre-existing function.
The DBGLOG executes once per page of records.
Will remove DBGLOG

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -206,6 +206,71 @@ WsWUExceptions::WsWUExceptions(IConstWorkUnit& wu): numerr(0), numwrn(0), numinf
...
+        outIOS.setown(helper.createIOStreamWithFileName(outFile, IFOcreate));
+    }
+
+    Owned<IRemoteLogAccessStream> logreader = m_remoteLogAccessor->getLogReader(logFetchOptions, logFormat);
+
+    StringBuffer logcontent;
+    unsigned totalRecsRead = 0;
+    unsigned recsRead = 0;
+    unsigned totalRecsReported = 0;
+    while (logreader->readLogEntries(logcontent.clear(), recsRead, totalRecsReported))
```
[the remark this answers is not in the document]
rpastrana: > design: I'm not sure that the interface should present total on every call read logs call. I think I'd expect this interface to look something like
> 
> ```
> enum 
> unsigned recsRead = logReader->readLogEntries(logContent.clear(), hitEnd);
> ```
> 
> so you can detect that it has hit the last [partial] page of results.
> 
> If the implementation supports it [cheaply], it would also be separately useful to know what the total is before you started to get the results in some cases, e.g. where what you ask for is way too big, but I think if possible, that should be a separate interface call, e.g.:
> 
> ```
> unsigned total = logReader->countLogEntries();
> ```

not sure what the enum is.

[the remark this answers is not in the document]
rpastrana: legitimate issue.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -206,6 +206,71 @@ WsWUExceptions::WsWUExceptions(IConstWorkUnit& wu): numerr(0), numwrn(0), numinf
...
+    {
+        totalRecsRead += recsRead;
+        outputALine(logcontent.length(), logcontent.str(), buf, outIOS);
+        DBGLOG("Fetched %u log records of %u total available", totalRecsRead, totalRecsReported);
```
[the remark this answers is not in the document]
rpastrana: one log output per fetch, one fetch per page.
Removing the log output.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -206,6 +206,71 @@ WsWUExceptions::WsWUExceptions(IConstWorkUnit& wu): numerr(0), numwrn(0), numinf
...
+    }
```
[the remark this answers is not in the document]
rpastrana: As far as elastic stack, we do get back the query total, adding a totalCount to the interface is a good idea.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -3775,6 +3853,37 @@ void CWsWuFileHelper::createZAPECLQueryArchiveFiles(IConstWorkUnit* cwu, const c
     }
 }
 
+void CWsWuFileHelper::createWULogFile(IConstWorkUnit *cwu, WsWuInfo &winfo, const char *path, unsigned maxLogRecords, LogAccessReturnColsMode retColsMode, LogAccessLogFormat logFormat, unsigned wuLogSearchTimeBuffSecs)
+{
+    if (cwu->getWuidVersion() == 0)
+        return;
+
+    MemoryBuffer mb;
```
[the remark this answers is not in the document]
rpastrana: The intent is to keep the standards in the file and use pre-existing logic.
Quickly looking over pre-existing logic, it seems they intended to be flexible and stream out to file, or membuff

system/logaccess/ElasticStack/ElasticStackLogAccess.cpp
rpastrana: great catch, fixing.

system/logaccess/ElasticStack/ElasticStackLogAccess.hpp
rpastrana: yeah I see your point, will rename to match its functionality

system/logaccess/ElasticStack/ElasticStackLogAccess.cpp
rpastrana: will fix

system/logaccess/ElasticStack/ElasticStackLogAccess.cpp
rpastrana: optimally we would serialize directly from JSON::Value to json/csv/xml. This change would be somewhat involved. How critical do you think it is to optimize this now?

system/jlib/jlog.cpp
```diff
@@ -3537,19 +3537,18 @@ IRemoteLogAccess * queryRemoteLogAccessor()
         libName.append("lib").append(type.str()).append("logaccess");
 
         //Load the DLL/SO
-        HINSTANCE logAccessPluginLib = LoadSharedObject(libName.str(), true, false);
+        HINSTANCE logAccessPluginLib = LoadSharedObject(libName.str(), false, true);
         if(logAccessPluginLib == nullptr)
             throw makeStringExceptionV(-1, "%s cannot load library '%s'", methodName, libName.str());
```
jakesmith, line 3542: minor: check no longer needed (the last param of LoadSharedObject enforces the check)

system/logaccess/ElasticStack/ElasticStackLogAccess.cpp
```diff
@@ -613,52 +619,58 @@ bool ElasticStackLogAccess::fetchLog(const LogAccessConditions & options, String
...
         {
-
-            if (!res["total"]["value"].empty())
-            {
-                totalRecs = res["total"]["value"].asUInt();
-            }
+            //if (!res["total"]["value"].empty())
+            //{
+            //    m_totalRecs = res["total"]["value"].asUInt();
+            //}
```
jakesmith, line 638: minor: remove commented block?

system/logaccess/ElasticStack/ElasticStackLogAccess.cpp
```diff
@@ -613,52 +619,58 @@ bool ElasticStackLogAccess::fetchLog(const LogAccessConditions & options, String
...
 
-    ElasticStackLogStream(std::string & queryString, const char * connstr, const char * indexsearchpattern, LogAccessLogFormat format,  std::size_t pageSize, std::string scrollTO)
-     : m_esSroller(std::make_shared<elasticlient::Client>(std::vector<std::string>({connstr})), pageSize, scrollTO)
+    ElasticStackLogStream(std::string & queryString, const char * connstr, const char * indexsearchpattern, LogAccessLogFormat format,  std::size_t pageSize, std::string scrollTo)
+     : m_esSroller(std::make_shared<elasticlient::Client>(std::vector<std::string>({connstr})), pageSize, scrollTo)
     {
         m_outputFormat = format;
         m_esSroller.init(indexsearchpattern, DEFAULT_ES_DOC_TYPE, queryString);
-        m_jsonWriter = m_jsonStreamBuilder.newStreamWriter();
+        m_jsonWriter = std::unique_ptr<Json::StreamWriter>(m_jsonStreamBuilder.newStreamWriter());
```
jakesmith, line 658: minor: this can be more concisely expressed as:

jsonWriter.reset(m_jsonStreamBuilder.newStreamWriter());

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -270,19 +270,41 @@ void WsWuInfo::readWorkunitComponentLogs(const char* outFile, unsigned maxLogRec
...
         }
         totalRecsRead += recsRead;
         outIOS->write(logcontent.length(), logcontent.str());
     }
 
-    if (logFormat == LOGACCESS_LOGFORMAT_json)
-        outIOS->write(2, "]}");
-    else if (logFormat == LOGACCESS_LOGFORMAT_xml)
-        outIOS->write(8, "</lines>");
+    if (totalRecsRead == maxLogRecords) //Warn of possible truncation
```
jakesmith, line 281: question: could request maxLogRecords+1, but truncate to maxLogRecords. That way we (and the consumer) would know for sure that it exceeded the limit / that there are more logs (even if it were 1)., but it would be unambiguous that over maxLogRecords at that point.

system/jlib/jlog.cpp
```diff
@@ -3486,11 +3485,10 @@ bool fetchLogByClass(StringBuffer & returnbuf, IRemoteLogAccess & logAccess, Log
...
 typedef IRemoteLogAccess * (*newLogAccessPluginMethod_t_)(IPropertyTree & logAccessPluginConfig);
 
-IRemoteLogAccess * queryRemoteLogAccessor()
+IRemoteLogAccess &queryRemoteLogAccessor()
 {
-    CriticalBlock block(logAccessCrit);
-    if (!logAccessor)
-    {
+    return *logAccessor.query([] 
+     {
```
jakesmith, line 3491: indent: looks like indented by 5?

jakesmith (commented):
@rpastrana - looks good, 1 question re. removing ambiguity re. whether hit end of limit or not + 1 trivial indent issue.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
rpastrana: Yeah interesting proposal, my concern is if the plus one forces an extra fetch (preventable, but then we still wouldn't be able to determine truncation). 
The trade off is one more potential fetch if the maxLogRecords % pagesize == 0 AND we happen to encounter exactly maxlogrecords. Tradeoff is we warn of possible truncation.
It seems to me non-deterministic warning is better.

system/logaccess/ElasticStack/ElasticStackLogAccess.cpp
rpastrana: @ghalliday  Jira to address optimization: https://track.hpccsystems.com/browse/HPCC-27423

jakesmith (commented):
@rpastrana - looks good, 1 missed minor comment: https://github.com/hpcc-systems/HPCC-Platform/pull/15837#discussion_r836216617

Please address and squash.

rpastrana:
> @rpastrana - looks good, 1 missed minor comment: [#15837 (comment)](https://github.com/hpcc-systems/HPCC-Platform/pull/15837#discussion_r836216617)
> 
> Please address and squash.

Yep. Addressed. Awaiting comment from Gavin before squashing all commits.

jakesmith:
https://github.com/hpcc-systems/HPCC-Platform/pull/15837#discussion_r836216617

@rpastrana, last commit removes the wrong code, it removed:
```
            if (xproc == nullptr)
                throw makeStringExceptionV(-1, "%s cannot locate procedure %s in library '%s'", methodName, instFactoryName, libName.str());
```

but comment(s) were to remove the unneeded check on LoadSharedObject failing:
```
            if(logAccessPluginLib == nullptr)
                throw makeStringExceptionV(-1, "%s cannot load library '%s'", methodName, libName.str());
```

jakesmith (approved):
@rpastrana - looks ready to me.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -206,6 +206,107 @@ WsWUExceptions::WsWUExceptions(IConstWorkUnit& wu): numerr(0), numwrn(0), numinf
...
+    }
+
+    Owned<IRemoteLogAccessStream> logreader = m_remoteLogAccessor->getLogReader(logFetchOptions, logFormat);
+
+    StringBuffer logcontent;
+    unsigned totalRecsRead = 0;
+    unsigned recsRead = 0;
+    
+    if (logFormat == LOGACCESS_LOGFORMAT_json)
+        outIOS->write(11, "{\"lines\": [");
```
ghalliday, line 267: This will core if outFile is empty.  Should the code above throw an exception if it is not provided?

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -206,6 +206,107 @@ WsWUExceptions::WsWUExceptions(IConstWorkUnit& wu): numerr(0), numwrn(0), numinf
...
+    else if (logFormat == LOGACCESS_LOGFORMAT_xml)
+        outIOS->write(7, "<lines>");
```
ghalliday, line 269: writeStringToStream would be simpler, and less error prone (and above/below)

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -206,6 +206,107 @@ WsWUExceptions::WsWUExceptions(IConstWorkUnit& wu): numerr(0), numwrn(0), numinf
...
+            writeCharToStream(*outIOS, ',');
+        }
+        totalRecsRead += recsRead;
+        outIOS->write(logcontent.length(), logcontent.str());
+    }
+
+    if (totalRecsRead == maxLogRecords) //Warn of possible truncation
+    {
+        VStringBuffer truncationWarnmessage("Log query reached record limit (%u). Log report could be incomplete.", maxLogRecords);
+        WARNLOG("WsWuInfo: %s", truncationWarnmessage.str());
```
ghalliday, line 284: This information doesn't seem like it is being reported in the correct way.
Better is likely to be a flag that is returned by the function call, and can be returned in the calling service, or processed some other way.
I'm not really convinced it should be a warning - it is more information.
My concern is that it will mess up anything reading the results.  For the moment that is likely to be human (so won't cause problems) or requested as json.
It can be addressed later.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -206,6 +206,107 @@ WsWUExceptions::WsWUExceptions(IConstWorkUnit& wu): numerr(0), numwrn(0), numinf
...
+
+        if (logFormat == LOGACCESS_LOGFORMAT_json)
+        {
+            VStringBuffer jsonMessage("], \"Warning\": \"%s\"}", truncationWarnmessage.str()); //close lines array, append Message object
+            writeStringToStream(*outIOS, jsonMessage.str()); 
+        }
+        else if (logFormat == LOGACCESS_LOGFORMAT_xml)
+        {
+            VStringBuffer xmlMessage("</lines>\n<!-- %s -->", truncationWarnmessage.str()); //close lines element, append comment message
```
ghalliday, line 293: More consistent with json would be
<warning>x</warning></lines>

system/jlib/jlog.cpp
```diff
@@ -2645,6 +2647,7 @@ MODULE_EXIT()
     thePassNoneFilter = nullptr;
     thePassLocalFilter = nullptr;
     thePassAllFilter = nullptr;
+    delete logAccessor.queryExisting();
```
ghalliday, line 2650: I suspect this should logAccessor member should also be cleared - but that is a general issue, and I'll create a jira/pr to address it.

system/jlib/jlog.hpp
```diff
@@ -1466,10 +1475,20 @@ struct LogAccessConditions
...
         startFrom = l.startFrom;
-
+        returnColsMode = l.returnColsMode;
```
ghalliday, line 1478: See comment on setFilter above.

system/logaccess/ElasticStack/ElasticStackLogAccess.cpp
```diff
@@ -520,11 +611,69 @@ cpr::Response ElasticStackLogAccess::performESQuery(const LogAccessConditions &
...
+     : m_esSroller(std::make_shared<elasticlient::Client>(std::vector<std::string>({connstr})), pageSize, scrollTo)
+    {
+        m_outputFormat = format;
+        m_esSroller.init(indexsearchpattern, DEFAULT_ES_DOC_TYPE, queryString);
+        m_jsonWriter.reset(m_jsonStreamBuilder.newStreamWriter());
+    }
+
+    virtual ~ElasticStackLogStream() override
+    {
+        m_esSroller.clear();
```
ghalliday, line 652: trivial: Do these need to be explict?

ghalliday (commented):
@rpastrana a few final comments.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
jakesmith: I'll take the blame for this suggestion.
I wanted a way to know from the ZAP report that the log files had been truncated.
An alternative would be to capture the fact that it has been truncated and store as a piece of meta data in the summary report file within the ZAP report.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
rpastrana: > More consistent with json would be x

@ghalliday not sure if your message was truncated. Anyway, I much rather insert a less intrusive comment in XML/JSON/CSV rather than an inline non-conforming entry with an unexpected message however I didn't find a way to do so in CSV nor JSON

esp/services/ws_workunits/ws_workunitsHelpers.cpp
rpastrana: @ghalliday I want to explore how to report this message to log.
I can see why there's hesitancy reporting it as a warn- If the truncation detection was deterministic, it would definitely be a warning (at least from my point of view). But ultimately It's an event which should be investigated and the false positive is not likely to happen often. Reporting it as info has two issues - it's more likely to be filtered out (perhaps ok), and it implies no further action should be taken.

rpastrana: Regarding the proper approach to handle possible truncation, I'm not crazy about the in-line message but it gets the message across.
Clearly, bubbling up an indication of truncation is the cleanest approach which the UI would process appropriately.
This functionality is a little odd in that we're creating various files and packaging them into a zip file for download.
Perhaps there could be a "status" file within the zip file which reports any issues encountered? Anyway, it seems a bit outside the scope of this particular PR. if nothing else, I can return the number of records fetched and streamed to local file for future processing.
thoughts @ghalliday @jakesmith

system/logaccess/ElasticStack/ElasticStackLogAccess.cpp
rpastrana: It seemed like the correct approach, can you expand your thoughts on this?

system/jlib/jlog.hpp
rpastrana: I see your point, however this is previously established code and the change is (as you point out) more stylistic and perhaps improves maintainability. does that fit the criteria for a separate jira?

system/logaccess/ElasticStack/ElasticStackLogAccess.cpp
ghalliday: Because the default destructor would do this anyway?

esp/services/ws_workunits/ws_workunitsHelpers.cpp
ghalliday: Not truncated, by tags not handled cleanly.  It should have said:

\<warning>x\</warning>\</lines>

esp/services/ws_workunits/ws_workunitsHelpers.cpp
ghalliday: The reason for my comments are that this is a general purpose function that happens to currently be called from the zap report code.  If it was called from another context (e.g. eclwatch to implement a paged filtered log view)  then it isn't really a warning to return to the user.  It is expected behaviour.  Similarly the caller would want to know that there might be more.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
rpastrana: right, that's what I assumed. and as I mentioned, it feels more important to be less intrusive (comment) rather than inserting a message inline w/ the log entries. How strongly do you feel about adding the message inline?

ghalliday: I don't think it needs changing for this PR, but will likely need revisiting if this function is called from another context, or if it is reported differently in the zap report.

ghalliday (approved):
@rpastrana looks good.   Please squash.

rpastrana:

PR #15853, opened 2022-03-03
HPCC-27294 Enable hthor+roxie DFS service access

jakesmith:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:


dali/base/dautils.cpp
```diff
@@ -3233,281 +3233,6 @@ void safeChangeModeWrite(IRemoteConnection *conn,const char *name,bool &reload,
     }
 }
 
-
-class CLocalOrDistributedFile: implements ILocalOrDistributedFile, public CInterface
```
jakesmith, line 3237: NB: impl. moved into ws_dfsclient, because it [conditionally] calls wsdfs::lookup, but didn't want dalibase to be dependent on ws_dfsclient lib.

esp/clients/ws_dfsclient/ws_dfsclient.cpp
```diff
@@ -716,8 +716,8 @@ IDFSFile *lookupDFSFile(const char *logicalName, unsigned timeoutSecs, unsigned
                     throw makeStringException(-1, "Could not find any DFS services in the target HPCC configuration.");
             }
         }
-        serviceUrl.append(dfsServiceUrls[currentDfsServiceUrl++].c_str());
-        logicalName = remoteLogicalFileName;
+        currentDfsServiceUrl = (currentDfsServiceUrl+1 == dfsServiceUrls.size()) ? 0 : currentDfsServiceUrl+1;
```
jakesmith, line 719: incidental bug fix, but only affected local case (which is only enabled by a debug option at moment)

roxie/ccd/ccdfile.cpp
```diff
@@ -1669,31 +1669,9 @@ class CRoxieFileCache : implements IRoxieFileCache, implements ICopyFileProgress
...
-                unsigned numStripedDevices = plane->numDevices();
-                stripeNum = calcStripeNumber(partNo-1, dlfn.get(), numStripedDevices);
-                FileDescriptorFlags fileFlags = static_cast<FileDescriptorFlags>(fileDesc.queryProperties().getPropInt("@flags"));
-                if (FileDescriptorFlags::none != (fileFlags & FileDescriptorFlags::dirperpart))
-                    defaultDirPerPart = true;
-            }
-#endif
-            makePhysicalPartName(dlfn.get(), partNo, numParts, localLocation, replicationLevel, DFD_OSdefault, defaultDir.str(), defaultDirPerPart, stripeNum);
+            RemoteFilename rfn;
+            pdesc->getFilename(replicationLevel, rfn);
```
jakesmith, line 1673: @richardkchapman - can you check this ? I couldn't see why it wasn't already using the IPartDescriptor to get the part filename - and it was needed as part of this change because the plane/defaults/prefix are encoded in the filedescriptor/partdescriptor not in the local plane.

esp/clients/ws_dfsclient/ws_dfsclient.cpp
ghalliday: This isn't equivalent - previously it indexed by the value pre-increment.  Now it is post increment.

ecl/eclcc/CMakeLists.txt
```diff
@@ -61,17 +62,18 @@ endif()
...
-         thorhelper 
+         dllserver
          hql
          hqlcpp
-         dllserver
+         jlib
+         nbcd 
+         thorhelper 
+         workunit 
+         ws_dfsclient
```
ghalliday, line 76: @richardkchapman I'm a bit concerned about this.  It means eclcc now has dependencies on esphttp, xmllib.  I'm not sure what implications it has for the client tools.  Possibly none since the ecl command already pulls them in.

I would be very tempted to get rid of the code that looks up the current record definition in dali, or at least mark it as deprecated.  The record translation is a much better bet.

ghalliday (commented):
Looks generally sensible.  one concern (probably ok), one comment, and @richardkchapman a question for you to answer from Jake.

esp/clients/ws_dfsclient/ws_dfsclient.cpp
jakesmith: will swap line 719 and 720.

ecl/eclcc/CMakeLists.txt
jakesmith: I'm not sure how the dali record layout is used by the compiler...
It generates static transform at compiler time? What controls when it does that currently?

ghalliday (approved):
Looks ok to me.

roxie/ccd/ccdfile.cpp
richardkchapman: No idea. It's fragile code around here though

jakesmith:
@richardkchapman - I think this is ready to merge. I'll go ahead and squash.

jakesmith:
squashed

jakesmith:
@richardkchapman - can this be merged?

richardkchapman (approved):

PR #15877, opened 2022-03-09
HPCC-24691 Convert ESP to use jlog standards

kenrowland:
Converted code for container builds to use log category set
in config yaml files. Legacy bare metal continues to use
existing ESP log level.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Cloud-compatibility
  - [x] Premature optimization

## Smoketest:

### DIFF
--- a/esp/bindings/SOAP/client/soapclient.cpp
+++ b/esp/bindings/SOAP/client/soapclient.cpp
@@ -124,10 +124,8 @@ int CSoapClient::postRequest(const char* contenttype, const char* soapaction, IR
         multipart->serialize(contenttypestr, requeststr);
     }
 
-    if (getEspLogLevel(rpccall.queryContext())>LogNormal)
-    {
-        DBGLOG("Content type: %s", contenttypestr.str());
-    }
+    auto logCat = (rpccall.queryContext() != nullptr) ? (rpccall.queryContext()->debugTraceEnabled() ? MsgCatEspElevated : LegacyMsgCatMax) : LegacyMsgCatMax;
+    LOG(logCat, "Content type: %s", contenttypestr.str());
 
     Owned<CSoapRequest> soap_request;
     soap_request.setown(new CSoapRequest);
--- a/esp/bindings/http/client/httpclient.cpp
+++ b/esp/bindings/http/client/httpclient.cpp
@@ -416,11 +416,8 @@ HttpClientErrCode CHttpClient::sendRequest(const char* method, const char* conte
         httprequest->addHeader("WWW-Authenticate", authheader.str());
     }
 
-    if (getEspLogLevel()>LogNormal)
-    {
-        DBGLOG("Content type: %s", contenttype);
-        DBGLOG("Request content: %s", request.str());
-    }
+    LOG(LegacyMsgCatMax, "Content type: %s", contenttype);
+    LOG(LegacyMsgCatMax, "Request content: %s", request.str());
 
     httprequest->setContent(request.str());
 
@@ -463,8 +460,7 @@ HttpClientErrCode CHttpClient::sendRequest(const char* method, const char* conte
     m_persistable = httpresponse->getPersistentEligible();
     m_numRequests++;
 
-    if (getEspLogLevel()>LogNormal)
-        DBGLOG("Response content: %s", response.str());
+    LOG(LegacyMsgCatMax,"Response content: %s", response.str());
 
     return HttpClientErrCode::OK;
 }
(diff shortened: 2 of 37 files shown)


ghalliday (changes requested):
This PR contains various categories of changes.  I think it is worth looking at each of them in turn.

1. ESPLOG(LogMax, ...) -> LOG(LegacyMsgCatMax, ...)

Previously these would have been output as DBGLOG() if the logging level was set sufficiently high.  Now they generate invalid logging lines because LegacyMsgCatMax does not define a valid category/audience.  (Audience/class unknown).  E.g.

`00000017 UNK 2022-03-11 08:53:29.986 397162 397682 "POST /WsStore/Set.json, from 192.168.86.202"
`
This suggests the logging output wasn't compared, or looked at carefully.

This change makes sense, but it should be using a valid category, and it is worth taking the time to choose the correct category as those changes are done.

See https://github.com/hpcc-systems/HPCC-Platform/pull/12188 for a similar change in the rest of the platform a while back - where each log message was carefully categorized when the function was changed.


2. ESPLOG(LogMax, "CLogFailSafe::~CLogFailSafe()"); -> DBGLOG()

It is now generating a message for normal logging, where it would have been a max level before.  Probably just a bug , there may be more like this (I didn't check).


3. DBGLOG(...) -> LOG(LegacyMsgCatMax, ...)

E.g.
```
if (getEspLogLevel()>LogNormal)
    DBGLOG("Response content: %s", response.str());

```
->
`    LOG(LegacyMsgCatMax,"Response content: %s", response.str());
`
Previously this would have output a DBGLOG entry with a log level of DebugMsgThreshold(80) if the logging level was raised.   Now it will be reported with level of ExtraneousMsgThreshold(90).  It will appear in any logs which have also had their logging detail level increased (esp log and stderr).  This is a reasonable change, (except for the change of category), because it is being more specific about the category of the line being output..

4. Elevated logging

```
if (getEspLogLevel(rpccall.queryContext())>LogNormal)
    DBGLOG("Content type: %s", contenttypestr.str());

```
->
```
auto logCat = (rpccall.queryContext() != nullptr) ? (rpccall.queryContext()->debugTraceEnabled() ? MsgCatEspElevated : LegacyMsgCatMax) : LegacyMsgCatMax;
LOG(logCat, "Content type: %s", contenttypestr.str());
```

The new code is much harder to read/understand.

Previously it would have output the logging as a DebugMsgThreshold if detailed logging as selected, or if the special debug mode was selected.

Now it always reports the message.  It either has a detail level of ExtraneousMsgThreshold, or it has a elevated category.   It has the same effect (other than losing the category) for bare-metal, including if the detail level was changed on the logging.

The way the MsgCatEspElevated has been implemented means that the logging lines will no longer be output as "DBG" - so it has lost details of the expected audience and category.  I'm also not convinced the mechanism is clearly explained - what this category means.

There are two separate questions, which logging lines should be generated, and which logging lines should be output to each of the different targets.  I think this is attempting to solve the first question using the mechanism that solves the second.  It probably requires a different solution.  

NOTE: With this change the log lines are now comming out non-sequentially in esp e.g.

```
00000009 UNK 2022-03-11 08:53:29.771 397162 397364 "GET /esp/files/dist/font-awesome/css/font-awesome.min.css, from 192.168.86.202"
0000000F UNK 2022-03-11 08:53:29.771 397162 397376 "GET /esp/files/node_modules/es6-promise/dist/es6-promise.auto.min.js, from 192.168.86.202"
```

This suggests that logging is not being caught by the pre-filter, or some other issue that needs investigating.


In conclusion, I don't think any of the categories of changes to the logging lines is correct (except for #3) - the problem needs to be understood better.  There are also some implementation issues, but I haven't checked carefully or commented on them.

ghalliday:
A couple of other comments.  It would be better to have the change broken up into smaller chunks.  It looks like something similar happened to https://github.com/hpcc-systems/HPCC-Platform/pull/12188 - it was broken up into smaller changes.
I'll discuss this with Richard/Jake/others to see if we can agree on the best approach.

kenrowland:
The following changes were made. The commit was rebased so that the entirety of the changes can be seen in context with the original code.

The audience and class values for the "legacy" logging categories were set to those used by MCdebugInfo which matches the DBGLOG calls employed by the ESPLOG function. 

The conversion of ESPLOG calls to use JLOG specific targeted log functions (ERRLOG, AUDLOG, etc.) were removed and replaced with LOG statements in line with what was done elsewhere in this PR.

Non sequential log ID values were not seen in captured ESP logs after the above changes. It was unclear if the log output above was created based on elevated logging in an ESP context. Knowing this would be useful.

The checks for context based elevated logging were not changed. Guards against a null context are required as the smoketest will fail if the guards are removed. The ESPLOG overload that accepted a context provided the guard in the existing code. An intermediate version of this code provided a function that returned the log category for the context and guarded against null.

Perhaps a meeting is in order to discuss the motivation for these changes, including the new category filter used for elevated logging. Let me know.

PR #15893, opened 2022-03-11
HPCC-27339 Store read/write stats for spray

shamser:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dali/ft/fttransform.cpp
+++ b/dali/ft/fttransform.cpp
@@ -521,7 +521,8 @@ void TransferServer::appendTransformed(unsigned chunkIndex, ITransformer * input
 
     const offset_t startInputOffset = curPartition.inputOffset;
     const offset_t startOutputOffset = curPartition.outputOffset;
-
+    stat_type prevNumWrites =  out->getStatistic(StNumDiskWrites);
+    stat_type prevNumReads = input->getStatistic(StNumDiskReads);
     for (;;)
     {
         unsigned gotLength = input->getBlock(out);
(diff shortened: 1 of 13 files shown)


ghalliday:
@shamser the failure of supercopy.ecl is likely to be related.

dali/ft/fttransform.hpp
```diff
@@ -31,6 +31,7 @@ interface ITransformer : public IInterface
     virtual void setInputCRC(crc32_t value) = 0;
     virtual bool setPartition(RemoteFilename & remoteInputName, offset_t _startOffset, offset_t _length, bool compressedInput, const char *decryptKey) = 0;
     virtual offset_t tell() = 0;
+    virtual unsigned __int64 getStatistic(StatisticKind kind) = 0;
```
ghalliday, line 34: Is this function ever called?

dali/ft/fttransform.cpp
```diff
@@ -868,7 +871,8 @@ bool TransferServer::pull()
                     LOG(MCdebugProgress, unknownJob, "Extend length of target file to %" I64F "d", lastOffset);
                 }
             }
-
+            if (out)
+                curProgress.numWrites += out->getStatistic(StNumDiskWrites);
```
ghalliday, line 875: I think this should be
prevWrites = out->getStatistic(StNumDiskWrites);
 before the call to out_>write, and then
 curProgress.numWrites = out->getStatistic(StNumDiskWrites) - prevWrites
here.  I would expect it to be 0 a lot of the time because the write will be buffered but not written

ghalliday: However, looking at the code I'm not sure if the output is buffered or not.  I am slightly concerned that some of the options may have almost no buffering.  e.g., if utf8 translation is going on, or other things.
There is also a call to out->flush() which might cause potentially efficiency issues.

ghalliday (commented):
@shamser change looks generally good.  A couple of specific questions.  I also have some concern about how often the output is flushed/the size of the writes when using different formats and translations.
Please can you try and gather some stats for those to see if it is something we need to worry about.

dali/ft/fttransform.hpp
shamser: yes, this is called in TransferServer::appendTransformed by "input->getStatistic(StNumDiskReads)".  (line 545 fttransform.cpp)

ghalliday (approved):
@shamser please squash

PR #15906, opened 2022-03-17
HPCC-27346 Refactor workunit loading in preparation for binary switch

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/ecl/hqlcpp/hqlecl.cpp
+++ b/ecl/hqlcpp/hqlecl.cpp
@@ -603,7 +603,7 @@ bool HqlDllGenerator::generateCode(HqlQueryContext & query)
 
 void HqlDllGenerator::addWorkUnitAsResource()
 {
-    bool defaultBinaryWorkunit = false;
+    bool defaultBinaryWorkunit = true;
     if (wu->getDebugValueInt("saveBinaryWorkunit", defaultBinaryWorkunit))
     {
         MemoryBuffer wuBinary;
--- a/ecl/eclccserver/eclccserver.cpp
+++ b/ecl/eclccserver/eclccserver.cpp
@@ -709,7 +709,7 @@ class EclccCompileThread : implements IPooledThread, implements IErrorReporter, 
                         workunit->setJobName(jobname);
                     if (!workunit->getDebugValueBool("obfuscateOutput", false))
                     {
-                        StringBuffer wuXML;
+                        StringBuffer wuXML; // Not sure this is a good idea.... better to always get it from the dll resource
                         Owned<IWUQuery> query = workunit->updateQuery();
                         if (getArchiveXMLFromFile(realdllfilename, wuXML.clear()))  // MORE - if what was submitted was an archive, this is probably pointless?
                             query->setQueryText(wuXML.str());


ghalliday:
This can be backported to 7.12.x with relatively little pain (clashes in 3 files from reoganisation).

common/dllserver/thorplugin.cpp
```diff
@@ -710,7 +719,7 @@ extern DLLSERVER_API IPropertyTree *getEmbeddedManifestPTree(const ILoadedDllEnt
     return getEmbeddedManifestXML(dll, xml) ? createPTreeFromXMLString(xml.str()) : createPTree();
 }
 
-extern DLLSERVER_API bool checkEmbeddedWorkUnitXML(ILoadedDllEntry *dll)
+extern DLLSERVER_API bool containsEmbeddedWorkUnit(ILoadedDllEntry *dll)
```
richardkchapman, line 722: Given the new name I expected this function to check for either XML or binary workunit to be present?

common/workunit/workunit.cpp
```diff
@@ -11873,7 +11910,7 @@ void exportWorkUnitToXMLFileWithHiddenPasswords(IPropertyTree *p, const char *fi
 
 extern WORKUNIT_API StringBuffer &exportWorkUnitToXML(const IConstWorkUnit *wu, StringBuffer &str, bool unpack, bool includeProgress, bool hidePasswords)
 {
-    // MORE - queryPTree isn't really safe without holding CLocalWorkUnit::crit - really need to move these functions into CLocalWorkunit
+    // MORE - queryPTree isn't really safe without holding CLocalWorkUnit::crit     - really need to move these functions into CLocalWorkunit
```
richardkchapman, line 11913: not sure what changed here or why...

ecl/eclagent/eclagent.cpp
```diff
@@ -3656,14 +3656,7 @@ extern int HTHOR_API eclagent_main(int argc, const char *argv[], StringBuffer *
...
 #endif
-        Owned<ILocalWorkUnit> standAloneWorkUnit;
-        if (wuXML)
-        {
-            //Create workunit from XML
-            standAloneWorkUnit.setown(createLocalWorkUnit(wuXML->str()));
-            wuXML->kill();  // free up text as soon as possible.
-        }
-
+        Owned<ILocalWorkUnit> standAloneWorkUnit(std::move(localWu));
```
richardkchapman, line 3659: Not sure why this copy is needed rather than using the one passed in

ecl/hqlcpp/hqlecl.cpp
```diff
@@ -603,9 +603,19 @@ bool HqlDllGenerator::generateCode(HqlQueryContext & query)
 
 void HqlDllGenerator::addWorkUnitAsResource()
 {
-    StringBuffer wuXML;
-    exportWorkUnitToXML(wu, wuXML, false, false, false);
-    code->addCompressResource("WORKUNIT", wuXML.length(), wuXML.str(), NULL, 1000);
+    bool defaultBinaryWorkunit = true; // MORE: CHANGE before release?
```
richardkchapman, line 606: Is an option to write both (for maximum back-compatibility) worth considering?

roxie/ccd/ccdmain.cpp
```diff
@@ -742,7 +742,7 @@ int CCD_API roxie_main(int argc, const char *argv[], const char * defaultYaml)
         else
         {
             Owned<ILoadedDllEntry> dll = createExeDllEntry(argv[0]);
-            if (checkEmbeddedWorkUnitXML(dll))
+            if (containsEmbeddedWorkUnit(dll))
```
richardkchapman, line 745: This looked like it only picked up old ones to me...

richardkchapman (changes requested):
Looks mostly ok, one possible issue

ecl/eclagent/eclagent.cpp
ghalliday: I can't think of a good reason either - possibly something to do with the lifetime, but I'll change and simplify

ghalliday:
@jakesmith can you think of any reason why serializing the ptree could cause compatibility problems?

richardkchapman (approved):

ecl/hqlcpp/hqlecl.cpp
jakesmith: @ghalliday - were you going to remove the comment?

roxie/ccd/ccdmain.cpp
jakesmith: looks like this has been addressed (containsEmbeddedWorkUnit now checks both )

jakesmith (approved):
@ghalliday - looks good, 1 trivial question, re. whether you're going to remove a comment.

jakesmith:
> can you think of any reason why serializing the ptree could cause compatibility problems?

no. Does assume the format is going to remain static/backward compatible from here on in.
I'm not sure there are any other places that persist serialized ptree's (to potentially be used by other versions).
I wonder if it's worth adding a version # at the head of the binworkunit blob, so if a variant format is introduced, it can upped/old checked ?

ecl/hqlcpp/hqlecl.cpp
ghalliday: @jakesmith needs discussion, but I think the plan is to back port this to 7.12.x so that old systems can read the new format, and then possibly enable it on 8.6.x

ghalliday:
@jakesmith does the dali code serialize ptrees?  In which case that is an example that needs to maintain backwards compatibility.  Adding the version number is a sensible idea though.

jakesmith:
> @jakesmith does the dali code serialize ptrees? In which case that is an example that needs to maintain backwards compatibility. Adding the version number is a sensible idea though.

it does, but dali clients/server know each others versions, so could switch implementations if need be via a version check.

richardkchapman:
Looks good, please squash

ghalliday:
I would wait for the 7.12.x code to be merged and upmerged - assuming that will be done.  This is mainly still open to make it easy to look at the changes when there are problems upmerging.

ghalliday:
(This will then become a single line changing the default for 8.6.x)

PR #15918, opened 2022-03-18
 HPCC-26921 Avoid CQ's re-looking up logical files.

jakesmith:
If a CQ performed a logical file lookup, e.g. via a diskread/kj/
indexread etc., and the logical filename could vary, e.g. it was
computed each time, then each worker would re-initialized the
manager activity, and cause a new Dali lookup per iteration.
e.g. on a 400 way Thor executing a CQ 1000 times, there would be
400,000 Dali lookups.

Change so that if within a CQ, the DFS lookups are cached
(not cleared) per execution. This results in the LFN being looked
up only once if the computed logical file remains constant.

Also, add temporary workaround for subFileStats tracking issues,
related to activities being able to have >1 readfile, some of
which can be superfiles.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/thorlcr/graph/thgraphmaster.cpp
+++ b/thorlcr/graph/thgraphmaster.cpp
@@ -606,7 +606,7 @@ void CMasterActivity::updateFileReadCostStats(std::vector<OwnedPtr<CThorStatsCol
      * NB: also changed when the expansion of readFiles (from supers to subfiles) happens, a new super was being added
      * each CQ iteration and re-expanded, meaing readFiles kept growing.
      * 
-     * Also, superkey1.ecl hits a dbgasserex whilst deserializaing stats (before and after these PR changes),
+     * Also, superkey1.ecl hits a dbgasserex whilst deserializing stats (before and after these PR changes),
      * but is caught/ignored. I haven't investigated further.
      */
 
--- a/thorlcr/graph/thgraphmaster.cpp
+++ b/thorlcr/graph/thgraphmaster.cpp
@@ -388,13 +388,6 @@ CMasterActivity::~CMasterActivity()
     delete [] data;
 }
 
-void CMasterActivity::addReadFile(IDistributedFile *file, bool temp)
-{
-    readFiles.append(*LINK(file));
-    if (!temp) // NB: Temps not listed in workunit
-        queryThorFileManager().noteFileRead(container.queryJob(), file);
-}
-
 IDistributedFile *CMasterActivity::queryReadFile(unsigned f)
 {
     if (f>=readFiles.ordinality())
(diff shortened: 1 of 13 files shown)


jakesmith:
@ghalliday - please review/discuss. This I believe is the main reason for severe slowness in the reported JIRA (and others similar).
There are some incidental changes to the subfile stats stuff, but spotted a few issues in relation to queries like these, but also KJ's (where>1 file read) that need addressing, but in a separate JIRA.

thorlcr/graph/thgraphmaster.cpp
```diff
@@ -542,35 +556,84 @@ void CMasterActivity::getEdgeStats(IStatisticGatherer & stats, unsigned idx)
...
 void CMasterActivity::updateFileReadCostStats(std::vector<OwnedPtr<CThorStatsCollection>> & subFileStats)
 {
-    if (!subFileStats.empty())
-    {
-        unsigned numSubFiles = subFileStats.size();
-        for (unsigned i=0; i<numSubFiles; i++)
-        {
-            IDistributedFile *file = queryReadFile(i);
-            if (file)
+    /* JCSMORE->SHAMSER: (separate JIRA needed)
```
jakesmith, line 584: Opened JIRA: https://track.hpccsystems.com/browse/HPCC-27399

ghalliday (approved):
@jakesmith a few trivial comments.  will defer to @shamser about the stats related changes - although they looked good to me.

ghalliday:
@shamser please note the comments jake has added for you in the PR, and can you double check his related changes

shamser (commented):
I see that subFileStats will be expanded at least to the size of the numsubs.  However, if the numsubs is smaller in a subsequent CQ execution,  it wouldn't get shrunk back.  If that is the case, wouldn't the workers deserialize too few CThorStatsCollection as the manager's array would be larger?

jakesmith:
> I see that subFileStats will be expanded at least to the size of the numsubs. However, if the numsubs is smaller in a subsequent CQ execution, it wouldn't get shrunk back. If that is the case, wouldn't the workers deserialize too few CThorStatsCollection as the manager's array would be larger?

yes, subFileStats will the high water mark, but any worker might be processing a file with less subfiles, meaning it will send less, and there'll be deserialization error as it tries to deserialize more than is available.
The full fix is going to require a map, or table in the manager, and the workers will need to identify which file (subfile) specifically they are dealing with/sending back , so the manager can reference the correct CThorStatsCollection.

I don't think it's worth going any further in this PR, it can still hit deserialization errors if there are variable filenames that have mix of super sizes (or some might not be supers), but it's why I opened up the separate JIRA: https://track.hpccsystems.com/browse/HPCC-27399

shamser (approved):

PR #15922, opened 2022-03-21
HPCC-25904: Add option to DFU copy files on query/packagemap deployment

afishbeck:
Please review the high level design.  I'll add a separate comment with add
command line walk through of useage.

There are still several things that need work:
1. More testing especially of actions other than "ecl publish",
   and doing remote copies.
2. Improvement of how "publisher dfu workunit" status is checked.
3. Fix how the DFU workunint waitForCompletion works - I am temporarily
   polling the workunit rather than subscribing.
4. Test and improve how other DFU actions interact with the
   "publisher dfu workunits".
5. A dependency conflict made me remove the ReferencedFileList.* files
   from workunit.dll.  I need to improve the way I share those files.
6. EclWatch support will have to be implemented after this PR.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/common/pkgfiles/CMakeLists.txt
+++ b/common/pkgfiles/CMakeLists.txt
@@ -1,0 +1,65 @@
+################################################################################
+#    HPCC SYSTEMS software Copyright (C) 2022 HPCC Systems®.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+################################################################################
+
+
+# Component: pkgfiles
+#####################################################
+# Description:
+# ------------
+#    Cmake Input File for pkgfiles
+#####################################################
+
+project( pkgfiles )
+
+set (    SRCS
+         package.cpp
+         referencedfilelist.cpp
+    )
+
+include_directories (
+        ${HPCC_SOURCE_DIR}/system/include
+        ${HPCC_SOURCE_DIR}/system/jlib
+        ${HPCC_SOURCE_DIR}/system/security/shared
+        ${HPCC_SOURCE_DIR}/system/mp
+        ${HPCC_SOURCE_DIR}/dali/base
+        ${HPCC_SOURCE_DIR}/dali/dfu
+        ${HPCC_SOURCE_DIR}/rtl/include
+        ${HPCC_SOURCE_DIR}/rtl/eclrtl
+        ${HPCC_SOURCE_DIR}/common/workunit
+    )
+
+if (NOT CONTAINERIZED)
+    include_directories ( ${HPCC_SOURCE_DIR}/common/environment )
+endif()
+
+HPCC_ADD_LIBRARY( pkgfiles SHARED ${SRCS} )
+
+set_target_properties(pkgfiles PROPERTIES
+    COMPILE_FLAGS -D_USRDLL
+    DEFINE_SYMBOL PKGFILES_EXPORTS )
+if(NOT PLUGIN)
+    install(
+        TARGETS pkgfiles
+        RUNTIME DESTINATION ${EXEC_DIR}
+        LIBRARY DESTINATION ${LIB_DIR})
+    target_link_libraries(
+        pkgfiles
+        dfuwu
+        )
+    if (NOT CONTAINERIZED)
+        target_link_libraries(pkgfiles environment)
+    endif()
+endif()
--- a/common/pkgfiles/referencedfilelist.cpp
+++ b/common/pkgfiles/referencedfilelist.cpp
@@ -23,7 +23,13 @@
 #include "dautils.hpp"
 #include "dasds.hpp"
 #include "dadfs.hpp"
+#include "daqueue.hpp"
 #include "dasess.hpp"
+#include "dfuwu.hpp"
+#ifndef _CONTAINERIZED
+#include "environment.hpp"
+#endif
+
 
 #define WF_LOOKUP_TIMEOUT (1000*15)  // 15 seconds
 
@@ -125,7 +131,7 @@ class ReferencedFile : implements IReferencedFile, public CInterface
 public:
     IMPLEMENT_IINTERFACE;
     ReferencedFile(const char *lfn, const char *sourceIP, const char *srcCluster, const char *prefix, bool isSubFile, unsigned _flags, const char *_pkgid, bool noDfs, bool calcSize)
-    : flags(_flags), pkgid(_pkgid), noDfsResolution(noDfs), calcFileSize(calcSize), fileSize(0), numParts(0), trackSubFiles(false)
+    : pkgid(_pkgid), fileSize(0), numParts(0), flags(_flags), noDfsResolution(noDfs), calcFileSize(calcSize), trackSubFiles(false)
     {
         {
             //Scope ensures strings are assigned
@@ -163,6 +169,8 @@ public:
     void resolve(const StringArray &locations, const char *srcCluster, IUserDescriptor *user, INode *remote, const char *remotePrefix, bool checkLocalFirst, StringArray *subfiles, bool trackSubFiles, bool resolveForeign=false);
     void resolve(const char *dstCluster, const char *srcCluster, IUserDescriptor *user, INode *remote, const char *remotePrefix, bool checkLocalFirst, StringArray *subfiles, bool trackSubFiles, bool resolveForeign=false);
 
+    virtual bool needsCopying(bool cloneForeign) const override;
+
     virtual const char *getLogicalName() const {return logicalName.str();}
     virtual unsigned getFlags() const {return flags;}
     virtual const SocketEndpoint &getForeignIP(SocketEndpoint &ep) const
@@ -173,8 +181,8 @@ public:
             ep.set(NULL);
         return ep;
     }
-    virtual void cloneInfo(unsigned updateFlags, IDFUhelper *helper, IUserDescriptor *user, const char *dstCluster, const char *srcCluster, bool cloneForeign, unsigned redundancy, unsigned channelsPerNode, int replicateOffset, const char *defReplicateFolder);
-    void cloneSuperInfo(unsigned updateFlags, ReferencedFileList *list, IUserDescriptor *user, INode *remote);
+    virtual void cloneInfo(const IPropertyTree *directories, IDFUWorkUnit *publisherWu, unsigned updateFlags, IDFUhelper *helper, IUserDescriptor *user, const char *dstCluster, const char *srcCluster, bool cloneForeign, unsigned redundancy, unsigned channelsPerNode, int replicateOffset, const char *defReplicateFolder, const char *dfu_queue);
+    void cloneSuperInfo(IDFUWorkUnit *publisherWu, unsigned updateFlags, ReferencedFileList *list, IUserDescriptor *user, INode *remote);
     virtual const char *queryPackageId() const {return pkgid.get();}
     virtual __int64 getFileSize()
     {
@@ -209,8 +217,8 @@ class ReferencedFileList : implements IReferencedFileList, public CInterface
 {
 public:
     IMPLEMENT_IINTERFACE;
-    ReferencedFileList(const char *username, const char *pw, bool allowForeignFiles, bool allowFileSizeCalc)
-        : allowForeign(allowForeignFiles), allowSizeCalc(allowFileSizeCalc)
+    ReferencedFileList(const char *username, const char *pw, bool allowForeignFiles, bool allowFileSizeCalc, const char *_jobname)
+        : jobName(_jobname), allowForeign(allowForeignFiles), allowSizeCalc(allowFileSizeCalc)
     {
         if (username && pw)
         {
@@ -219,8 +227,8 @@ public:
         }
     }
 
-    ReferencedFileList(IUserDescriptor *userDesc, bool allowForeignFiles, bool allowFileSizeCalc)
-        : allowForeign(allowForeignFiles), allowSizeCalc(allowFileSizeCalc)
+    ReferencedFileList(IUserDescriptor *userDesc, bool allowForeignFiles, bool allowFileSizeCalc, const char *_jobname)
+        : jobName(_jobname), allowForeign(allowForeignFiles), allowSizeCalc(allowFileSizeCalc)
     {
         if (userDesc)
             user.set(userDesc);
@@ -240,17 +248,23 @@ public:
     void addFilesFromPackage(IPropertyTree &package, const char *_daliip, const char *srcCluster, const char *_remotePrefix);
 
     virtual IReferencedFileIterator *getFiles();
-    virtual void cloneFileInfo(const char *dstCluster, unsigned updateFlags, IDFUhelper *helper, bool cloneSuperInfo, bool cloneForeign, unsigned redundancy, unsigned channelsPerNode, int replicateOffset, const char *defReplicateFolder);
+    virtual void cloneFileInfo(StringBuffer &publisherWuid, const char *dstCluster, unsigned updateFlags, IDFUhelper *helper, bool cloneSuperInfo, bool cloneForeign, unsigned redundancy, unsigned channelsPerNode, int replicateOffset, const char *defReplicateFolder);
 
     virtual void cloneRelationships();
-    virtual void cloneAllInfo(const char *dstCluster, unsigned updateFlags, IDFUhelper *helper, bool cloneSuperInfo, bool cloneForeign, unsigned redundancy, unsigned channelsPerNode, int replicateOffset, const char *defReplicateFolder)
+    virtual void cloneAllInfo(StringBuffer &publisherWuid, const char *dstCluster, unsigned updateFlags, IDFUhelper *helper, bool cloneSuperInfo, bool cloneForeign, unsigned redundancy, unsigned channelsPerNode, int replicateOffset, const char *defReplicateFolder)
     {
-        cloneFileInfo(dstCluster, updateFlags, helper, cloneSuperInfo, cloneForeign, redundancy, channelsPerNode, replicateOffset, defReplicateFolder);
+        cloneFileInfo(publisherWuid, dstCluster, updateFlags, helper, cloneSuperInfo, cloneForeign, redundancy, channelsPerNode, replicateOffset, defReplicateFolder);
         cloneRelationships();
     }
     virtual void resolveFiles(const StringArray &locations, const char *remoteIP, const char *_remotePrefix, const char *srcCluster, bool checkLocalFirst, bool addSubFiles, bool trackSubFiles, bool resolveForeign=false) override;
 
     void resolveSubFiles(StringArray &subfiles, const StringArray &locations, bool checkLocalFirst, bool trackSubFiles, bool resolveForeign);
+    virtual bool filesNeedCopying(bool cloneForeign);
+    virtual void setDfuQueue(const char *queue) override
+    {
+        dfu_queue.set(queue);
+    }
+
 
 public:
     Owned<IUserDescriptor> user;
@@ -258,6 +272,8 @@ public:
     MapStringToMyClass<ReferencedFile> map;
     StringAttr srcCluster;
     StringAttr remotePrefix;
+    StringAttr jobName; //used to populate DFU job name, but could be used elsewhere
+    StringAttr dfu_queue;
     bool allowForeign;
     bool allowSizeCalc;
 };
@@ -474,13 +490,145 @@ void ReferencedFile::resolve(const char *dstCluster, const char *srcCluster, IUs
         locations.append(dstCluster);
 }
 
-void ReferencedFile::cloneInfo(unsigned updateFlags, IDFUhelper *helper, IUserDescriptor *user, const char *dstCluster, const char *srcCluster, bool cloneForeign, unsigned redundancy, unsigned channelsPerNode, int replicateOffset, const char *defReplicateFolder)
+static void setRoxieClusterPartDiskMapping(const char *clusterName, const char *defaultFolder, const char *defaultReplicateFolder, bool supercopy, IDFUfileSpec *wuFSpecDest, IDFUoptions *wuOptions)
+{
+    ClusterPartDiskMapSpec spec;
+    spec.setDefaultBaseDir(defaultFolder);
+
+    if (!supercopy)
+        spec.setRepeatedCopies(CPDMSRP_lastRepeated,false);
+    wuFSpecDest->setClusterPartDiskMapSpec(clusterName,spec);
+}
+
+static void getDefaultDFUName(StringBuffer &dfuQueueName)
+{
+// Using the first queue for now.
+#ifdef _CONTAINERIZED
+    Owned<IPropertyTreeIterator> dfuQueues = getComponentConfigSP()->getElements("dfuQueues");
+    ForEach(*dfuQueues)
+    {
+        const char *dfuName = dfuQueues->query().queryProp("@name");
+        if (!isEmptyString(dfuName))
+        {
+            getDfuQueueName(dfuQueueName, dfuName);
+            break;
+        }
+    }
+#else
+    Owned<IEnvironmentFactory> factory = getEnvironmentFactory(true);
+    Owned<IConstEnvironment> env = factory->openEnvironment();
+
+    StringBuffer xpath ("Software/DfuServerProcess");
+    Owned<IPropertyTree> root = &env->getPTree();
+    Owned<IPropertyTreeIterator> targets = root->getElements(xpath.str());
+    ForEach(*targets)
+    {
+        IPropertyTree &target = targets->query();
+        if (target.hasProp("@queue"))
+            dfuQueueName.set(target.queryProp("@queue"));
+    }
+#endif
+}
+
+static void dfuCopy(const IPropertyTree *directories, IDFUWorkUnit *publisherWu, IUserDescriptor *user, const char *sourceLogicalName, const char *destLogicalName, const char *destPlane, const char *srcDali, bool supercopy, bool overwrite, bool preserveCompression, bool nosplit)
+{
+    if(!publisherWu)
+        throw makeStringException(-1, "Failed to create Publisher DFU Workunit.");
+    if(isEmptyString(sourceLogicalName))
+        throw makeStringException(-1, "Source logical file not specified.");
+    if(isEmptyString(destLogicalName))
+        throw makeStringException(-1, "Destination logical file not specified.");
+    if(isEmptyString(destPlane))
+        throw makeStringException(-1, "Destination node group not specified.");
+
+    PROGLOG("Copy from [%s]  %s to %s", isEmptyString(srcDali) ? "local" : srcDali, sourceLogicalName, destLogicalName);
+
+    StringBuffer destFolder, destTitle, defaultFolder, defaultReplicateFolder;
+    DfuParseLogicalPath(directories, destLogicalName, destPlane, destFolder, destTitle, defaultFolder, defaultReplicateFolder);
+
+    CDfsLogicalFileName logicalName;
+    logicalName.set(sourceLogicalName);
+    if (!isEmptyString(srcDali))
+    {
+        SocketEndpoint ep(srcDali);
+        if (ep.isNull())
+            throw MakeStringException(-1, "ReferencedFile Copy %s: cannot resolve SourceDali network IP from %s.", sourceLogicalName, srcDali);
+
+        logicalName.setForeign(ep,false);
+    }
+
+    Owned<IDistributedFile> file = queryDistributedFileDirectory().lookup(logicalName, user, AccessMode::tbdRead, false, false, nullptr, defaultPrivilegedUser);
+    if (!file)
+        throw MakeStringException(-1, "ReferencedFile failed to find file: %s", logicalName.get());
+
+    if (supercopy)
+    {
+        if (!file->querySuperFile())
+            supercopy = false;
+    }
+    else if (file->querySuperFile() && (file->querySuperFile()->numSubFiles() > 1) && isFileKey(file))
+        supercopy = true;
+
+    Owned<IDFUWorkUnitFactory> factory = getDFUWorkUnitFactory();
+    Owned<IDFUWorkUnit> wu = factory->createPublisherSubTask(publisherWu);
+    wu->setJobName(destLogicalName);
+
+    if (user)
+    {
+        StringBuffer username;
+        wu->setUser(user->getUserName(username).str());
+    }
+
+    wu->setClusterName(destPlane);
+    if (supercopy)
+        wu->setCommand(DFUcmd_supercopy);
+    else
+        wu->setCommand(DFUcmd_copy);
+
+    IDFUfileSpec *wuFSpecSource = wu->queryUpdateSource();
+    IDFUfileSpec *wuFSpecDest = wu->queryUpdateDestination();
+    IDFUoptions *wuOptions = wu->queryUpdateOptions();
+    wuFSpecSource->setLogicalName(sourceLogicalName);
+    if (!isEmptyString(srcDali))
+    {
+        SocketEndpoint ep(srcDali);
+        wuFSpecSource->setForeignDali(ep);
+
+        if (user)
+        {
+            StringBuffer srcUserName;
+            user->getUserName(srcUserName);
+            if(!srcUserName.isEmpty())
+                wuFSpecSource->setForeignUser(srcUserName, "");
+        }
+    }
+    wuFSpecDest->setLogicalName(destLogicalName);
+    wuOptions->setOverwrite(overwrite);
+    wuOptions->setPreserveCompression(preserveCompression);
+    wuOptions->setNoSplit(nosplit);
+
+    setRoxieClusterPartDiskMapping(destPlane, defaultFolder.str(), defaultReplicateFolder.str(), supercopy, wuFSpecDest, wuOptions);
+    wuFSpecDest->setWrap(true); // roxie always wraps
+    if (!supercopy)
+        wuOptions->setSuppressNonKeyRepeats(true); // **** only repeat last part when src kind = key
+
+    submitDFUWorkUnit(wu.getClear());
+}
+
+bool ReferencedFile::needsCopying(bool cloneForeign) const
 {
     if ((flags & RefFileCloned) || (flags & RefFileSuper) || (flags & RefFileInPackage))
-        return;
+        return false;
     if ((flags & RefFileForeign) && !cloneForeign)
-        return;
+        return false;
     if (!(flags & (RefFileRemote | RefFileForeign | RefFileNotOnCluster)))
+        return false;
+    return true;
+}
+
+void ReferencedFile::cloneInfo(const IPropertyTree *directories, IDFUWorkUnit *publisherWu, unsigned updateFlags, IDFUhelper *helper, IUserDescriptor *user, const char *dstCluster, const char *srcCluster, bool cloneForeign, unsigned redundancy, unsigned channelsPerNode, int replicateOffset, const char *defReplicateFolder, const char *dfu_queue)
+{
+    if (!needsCopying(cloneForeign))
         return;
     if (fileSrcCluster.length())
         srcCluster = fileSrcCluster;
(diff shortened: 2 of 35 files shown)


afishbeck:
## Option 2:  "only-copy-files"

Create the DFU publisher workunit, then stop.
Still has to compile the query (if not using a precompiled workunit) in order to know what files are used.
You need to explicitly do the real publish at the end once all files are copied.


### Publish the query:

```
ecl publish roxie2 ReadIndexService.ecl --dfu-copy --only-copy-files
W20220321-150602

DFU publisher file copying Wuid: P20220321-150603 is started

```
### Check the publisher wuid status:

```
> dfuplus server=. action=status wuid=P20220321-150603
P20220321-150603T2 status: failed - (null) [certification::full_test_distributed]
P20220321-150603T1 status: failed - (null) [certification::full_test_distributed_index]

```
### Check the status of an individual child task:

```
> dfuplus server=. action=status wuid=P20220321-150603T1
P20220321-150603T1 status: failed - Failed: OVERWRITE not supplied and physical file //127.0.0.1/var/lib/HPCCSystems/hpcc-roxie-data2/certification/full_test_distributed_index._1_of_3 exists

```
Fix any issues found in the file copies.
In this case we just need to use dfu-overwrite.

Try publish again with overwrite (without recompiling query by using query workunit from above).

```
> ecl publish roxie2 W20220321-150602 --name=readindexservice --dfu-copy --only-copy-files --dfu-overwrite

DFU publisher file copying Wuid: P20220321-150713 is started

```
# Check the publisher wuid status:

```
> dfuplus server=. action=status wuid=P20220321-150713
P20220321-150713T2 Finished [certification::full_test_distributed]
P20220321-150713T1 Finished [certification::full_test_distributed_index]

```
Success! Now publish the query again using the compiled workunit

```
> ecl publish roxie2 W20220321-150602 --name=readindexservice --dfu-copy

roxie2/readindexservice.1

```

afishbeck:
@ghalliday @richardkchapman the three options in the walk through above show slight variations in how a user might interact with this feature.  Mostly depending on how separate they want the file copying to be from the actual query publish step.  As a first go if you could comment on the walk through that would be helpful.  Suggestions for changing names of things would be welcome.

ghalliday:
windows build errors are down to incorrect declspec on some functions.

common/workunit/referencedfilelist.cpp
```diff
@@ -474,13 +488,213 @@ void ReferencedFile::resolve(const char *dstCluster, const char *srcCluster, IUs
...
+        user->getUserName(srcUserName);
+        if(!srcUserName.isEmpty())
+        {
+            StringBuffer srcPassWord;
+            user->getPassword(srcPassWord);
+            wuFSpecSource->setForeignUser(srcUserName, srcPassWord);
+        }
+    }
+    wuFSpecDest->setLogicalName(destLogicalName);
+    wuFSpecDest->setFileMask(fileMask.str());
```
ghalliday, line 671: @jakesmith does code need to set file mask, or only if it differs from the default?

common/workunit/referencedfilelist.cpp
```diff
@@ -474,13 +488,213 @@ void ReferencedFile::resolve(const char *dstCluster, const char *srcCluster, IUs
...
+{
+    ClusterPartDiskMapSpec spec;
+    spec.setDefaultBaseDir(defaultFolder);
+
+    if (!supercopy)
+        spec.setRepeatedCopies(CPDMSRP_lastRepeated,false);
+    wuFSpecDest->setClusterPartDiskMapSpec(clusterName,spec);
+}
+
+//tbd make these parameters to requests using --dfu-copy
```
ghalliday, line 548: I think comment is now out of date because it can be configured

common/workunit/referencedfilelist.cpp
```diff
@@ -474,13 +488,213 @@ void ReferencedFile::resolve(const char *dstCluster, const char *srcCluster, IUs
...
+
+    CDfsLogicalFileName lfn;
+    parseLogicalPath(destLogicalName, destNodeGroup.str(), NULL, destFolder, destTitle, defaultFolder, defaultReplicateFolder);
+
+    StringBuffer fileMask;
+    constructFileMask(destTitle.str(), fileMask);
+
+    CDfsLogicalFileName logicalName;
+    logicalName.set(sourceLogicalName);
+    if (!isEmptyString(srcDali))
```
ghalliday, line 608: @jakesmith Should this now be a srcEsp?

common/workunit/referencedfilelist.cpp
```diff
@@ -474,13 +488,213 @@ void ReferencedFile::resolve(const char *dstCluster, const char *srcCluster, IUs
...
+    {
+        SocketEndpoint ep(srcDali);
+        if (ep.isNull())
+            throw MakeStringException(-1, "ReferencedFile Copy %s: cannot resolve SourceDali network IP from %s.", sourceLogicalName, srcDali);
+
+        logicalName.setForeign(ep,false);
```
ghalliday, line 614: New code needs to be using setRemote() rather than setForeign.

common/workunit/referencedfilelist.cpp
```diff
@@ -474,13 +488,213 @@ void ReferencedFile::resolve(const char *dstCluster, const char *srcCluster, IUs
...
+    }
+
+    Owned<IDistributedFile> file = queryDistributedFileDirectory().lookup(logicalName, user, false, false, false, nullptr, defaultPrivilegedUser);
```
ghalliday, line 617: Should now be calling 

`wsdfs::lookup()`

@jakesmith we should make sure all the platform developers know about the change - I suspect there are other changes like this which should be more widely known.

common/workunit/referencedfilelist.cpp
```diff
@@ -474,13 +488,213 @@ void ReferencedFile::resolve(const char *dstCluster, const char *srcCluster, IUs
...
+    Owned<IDFUWorkUnitFactory> factory = getDFUWorkUnitFactory();
+    Owned<IDFUWorkUnit> wu = factory->createPublisherSubTask(publisherWuid);
+    wu->setJobName(destLogicalName);
+    wu->setQueue(dfuQueueName);
+
+    if (user)
+    {
+        StringBuffer username, password;
+        wu->setUser(user->getUserName(username).str());
+        wu->setPassword(user->getPassword(password).str());
```
ghalliday, line 642: Where is the password coming from?  It shouldn't be available.

common/workunit/referencedfilelist.cpp
```diff
@@ -474,13 +488,213 @@ void ReferencedFile::resolve(const char *dstCluster, const char *srcCluster, IUs
...
+    }
+    wuFSpecDest->setLogicalName(destLogicalName);
+    wuFSpecDest->setFileMask(fileMask.str());
+    wuOptions->setOverwrite(overwrite);
+    wuOptions->setPreserveCompression(preserveCompression);
+    wuOptions->setNoSplit(nosplit);
+
+    setRoxieClusterPartDiskMapping(destNodeGroup.str(), defaultFolder.str(), defaultReplicateFolder.str(), supercopy, wuFSpecDest, wuOptions);
+    wuFSpecDest->setWrap(true); // roxie always wraps
+    if (!supercopy)
```
ghalliday, line 678: I'm not sure what this is doing, but it raises a question.  Is it ever correct to duplicate file parts when using non-attached storage?  Will it cause problems if it does.

common/workunit/referencedfilelist.cpp
```diff
@@ -787,14 +1008,35 @@ void ReferencedFileList::resolveFiles(const StringArray &locations, const char *
...
+{
+    ReferencedFileIterator files(this);
+    ForEach(files)
+    {
+        if (files.queryObject().needsCopying(cloneForeign))
+            return true;
+    }
+    return false;
+}
+void ReferencedFileList::cloneFileInfo(StringBuffer &publisherWuid, const char *dstCluster, unsigned updateFlags, IDFUhelper *helper, bool cloneSuperInfo, bool cloneForeign, unsigned redundancy, unsigned channelsPerNode, int replicateOffset, const char *defReplicateFolder)
```
ghalliday, line 1021: whitespace: add a newline between functions

dali/dfu/dfuwu.cpp
```diff
@@ -2437,7 +2499,7 @@ class CDFUWorkUnit: public CDFUWorkUnitBase
         // called with crit locked
         if (!subscriberid) {
             StringBuffer xpath;
-            getXPath(xpath,queryId()).append("/Progress/Edition");
+            getXPath(xpath,queryId()).append("/Progress");///Edition");
```
ghalliday, line 2502: Why this change?  I think it means that changes to Edition won't be notified.

dali/dfu/dfuwu.cpp
```diff
@@ -2505,8 +2567,10 @@ class CDFUWorkUnit: public CDFUWorkUnitBase
             case DFUstate_finished:
                 return ret;
             }
-            if (!completed.wait(timeout))       // should only go round loop once
-                break;
+            MilliSleep(5000);
+            // TBD figure out why subsctiption + Semaphore isn't working
```
ghalliday, line 2571: In what way isn't it working?

dali/dfu/dfuwu.cpp
```diff
@@ -2962,6 +3026,57 @@ class CDFUWorkUnitFactory : implements IDFUWorkUnitFactory, implements ISDSSubsc
...
+        progress->setProp("@state", "started");
+        conn->commit();
+    }
+
+    unsigned incrementPublisherTaskCount(const char *parent)
+    {
+        StringBuffer wuProgressRoot;
+        getXPath(wuProgressRoot, parent);
+        wuProgressRoot.append("/Progress");
+        Owned<IRemoteConnection> conn = querySDS().connect(wuProgressRoot.str(), myProcessSession(), RTM_LOCK_WRITE, SDS_LOCK_TIMEOUT);
```
ghalliday, line 3058: Slightly concerned about the overhead of this operation.  Better to hold on to the connection (via a CDfuWorkuUnit?) and increment, or set once all tasks are created.  @jakesmith would you be concerned?

esp/services/ws_workunits/ws_workunitsQuerySets.cpp
```diff
@@ -910,19 +915,48 @@ bool CWsWorkunitsEx::onWUPublishWorkunit(IEspContext &context, IEspWUPublishWork
...
+        cpr.dfu_queue.set(req.getDfuQueue());
+        cpr.copy(publisherWuid, cw, updateFlags);
 
         if (req.getIncludeFileErrors())
             cpr.gatherFileErrors(resp.getFileErrors());
+        if (!publisherWuid.isEmpty())
+            resp.setDfuPublisherWuid(publisherWuid);
+    }
+
+    if (req.getOnlyCopyFiles() || (req.getStopIfFilesCopied() && !publisherWuid.isEmpty()))
```
ghalliday, line 941: Code like this appears too many times in this file. Is there a way to common it up?  (I know there is already lots of duplication). It is hard to review, and a pain to change if a problem is found.

ghalliday (changes requested):
@afishbeck I've tried to go through reasonably thoroughly.  A few comments/issues, but the overall plan looks reasonable.  I wonder about the overhead of including all the tasks in dali, e.g. does the detailed progress information get removed once a copy completes successfully?

esp/services/ws_dfu/CMakeLists.txt
```diff
@@ -27,6 +27,8 @@ project( ws_dfu )
 include(${HPCC_SOURCE_DIR}/esp/scm/smcscm.cmake)
 
 set (    SRCS 
+         ${HPCC_SOURCE_DIR}/common/workunit/package.cpp
```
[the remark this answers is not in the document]
afishbeck: Yes, there was a circular dependency.  This is just my temporary work around while I work on the larger PR.  I'll probably create a new shared library because it's dependency on both dfu and workunit.

dali/dfu/dfuwu.cpp
```diff
@@ -51,9 +51,20 @@ static StringBuffer &getXPathBase(StringBuffer &wuRoot)
...
+        return wuRoot.append('/').append(wuid).append("/Tasks/").append(wuid).append("T").append(task);
+    return wuRoot.append('/').append(wuid);
+}
+
 static StringBuffer &getXPath(StringBuffer &wuRoot, const char *wuid)
 {
     getXPathBase(wuRoot);
+    const char *task = strchr(wuid, 'T');
+    if (task)
+        return wuRoot.append('/').append(task-wuid, wuid).appendf("/Tasks/").append(wuid);
```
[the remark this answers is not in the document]
afishbeck: Because the task is actually a fully functioning DFU COPY workunit the child tag has a name with both parent wuid + task id.  It ends up looking like:

```
<P20220425-173239>
  <Tasks>
     <P20220425-173239T1>...</P20220425-173239T1>
     <P20220425-173239T2>...</P20220425-173239T2>
 </Tasks>
</P20220425-173239>

```
It allows for easy discovery of parent and child, and also allows existing code to continue to work.

dali/dfu/dfuwu.cpp
afishbeck: Was just me playing with getting the subscribe to work... will revert.  Every other place I look polls for these workunits to complete so I will use the same technique for now.  Perhaps the function that uses subscribe can be fixed in another PR.

dali/dfu/dfuwu.cpp
afishbeck: When the workunit is changed by dfu server the subscribe isn't being notified.  I've tried a few variations, but this code isn't used anywhere.  All other examples I've found in use poll for workunit completion. I'll do the same, and perhaps this mechanism can be fixed in another PR.

esp/scm/ws_packageprocess.ecm
```diff
@@ -42,13 +42,20 @@ ESPrequest AddPackageRequest
     bool UpdateSuperFiles(false); //usually wouldn't be needed, packagemap referencing superfiles?
     bool UpdateCloneFrom(false); //explicitly wan't to change where roxie will grab from
     bool AppendCluster(true); //file exists on other local cluster, add new one, make optional in case of locking issues, but should be made to work
+    bool DfuCopyFiles(false);
```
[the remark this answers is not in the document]
afishbeck: Technically request interfaces aren't as sensitive to versioning issues.  If a client uses an input it means they were designed for or compiled against that version.

But I will add versioning anyway, it helps clarify when changes were made, is self documenting when trying to hit old servers, and keeps the input and output versioning in sync.

esp/services/ws_workunits/ws_workunitsQuerySets.cpp
afishbeck: It happens because we have different interfaces with the same or similar members.  But at least in cases like this one, I can use a template function to common it up, so will do.

common/workunit/referencedfilelist.cpp
```diff
@@ -474,13 +488,213 @@ void ReferencedFile::resolve(const char *dstCluster, const char *srcCluster, IUs
...
+        throw MakeStringException(-1, "Destination logical file not specified.");
+
+    PROGLOG("Copy from %s to %s", sourceLogicalName, destLogicalName);
+
+    StringBuffer destFolder, destTitle, defaultFolder, defaultReplicateFolder;
+    StringBuffer srcNodeGroup;
+    StringBuffer destNodeGroup(destNodeGroupIn);
+    if(destNodeGroup.isEmpty())
+    {
+        getNodeGroupFromLFN(user, sourceLogicalName, destNodeGroup);
```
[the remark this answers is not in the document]
afishbeck: This was how FileCopy was doing it, but I've decided not to try to guess.  The destination group/plane should have already been provided otherwise I will error.

common/workunit/referencedfilelist.cpp
afishbeck: Unless DFU server supports WsDFS, I'd rather wait to change this to use a WsDFS address and call.  Otherwise I guess I would have to ask the user for both addresses?

common/workunit/referencedfilelist.cpp
afishbeck: Command line can still provide it, but I'll look at removing... and how it's used by DFU server.

common/workunit/referencedfilelist.cpp
afishbeck: Copied from other place doing DFU server based copies.  I'll remove.

common/workunit/referencedfilelist.cpp
afishbeck: @jakesmith @ghalliday I could be wrong, but it doesn't look like DFU server COPY currently supports the WsDFS way of doing this?  Ultimately I am just setting up the copy for DFU server  and if it doesn't support WsDFS I need to provide a DALI.

afishbeck:
@ghalliday no longer a draft PR.  Files have been moved and a new shared library (pkgfiles) has been created.  But the file moves are in the last commit.  Look at the commits individually to see actual changes in the moved files.  I've tried to clean up the code as you suggested while reviewing the draft.

ghalliday (commented):
@afishbeck this looks close to ready.  A few trivial comments, and a couple of questions.

common/pkgfiles/CMakeLists.txt
```diff
@@ -0,0 +1,74 @@
...
+        workunit
+        jlib
+        dalibase
+        dllserver
+        nbcd
+        eclrtl
+        deftype
+        dfuwu
+        )
+#    if (USE_OPENSSL)
```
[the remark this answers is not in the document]
afishbeck: done.

common/pkgfiles/referencedfilelist.cpp
```diff
@@ -163,6 +169,8 @@ class ReferencedFile : implements IReferencedFile, public CInterface
     void resolve(const StringArray &locations, const char *srcCluster, IUserDescriptor *user, INode *remote, const char *remotePrefix, bool checkLocalFirst, StringArray *subfiles, bool trackSubFiles, bool resolveForeign=false);
     void resolve(const char *dstCluster, const char *srcCluster, IUserDescriptor *user, INode *remote, const char *remotePrefix, bool checkLocalFirst, StringArray *subfiles, bool trackSubFiles, bool resolveForeign=false);
 
+    bool needsCopying(bool cloneForeign) const override;
```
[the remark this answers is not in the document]
afishbeck: Adding virtual, it's clearest.

common/pkgfiles/referencedfilelist.cpp
```diff
@@ -240,24 +248,32 @@ class ReferencedFileList : implements IReferencedFileList, public CInterface
...
+    virtual void cloneAllInfo(StringBuffer &publisherWuid, const char *dstCluster, unsigned updateFlags, IDFUhelper *helper, bool cloneSuperInfo, bool cloneForeign, unsigned redundancy, unsigned channelsPerNode, int replicateOffset, const char *defReplicateFolder)
     {
-        cloneFileInfo(dstCluster, updateFlags, helper, cloneSuperInfo, cloneForeign, redundancy, channelsPerNode, replicateOffset, defReplicateFolder);
+        cloneFileInfo(publisherWuid, dstCluster, updateFlags, helper, cloneSuperInfo, cloneForeign, redundancy, channelsPerNode, replicateOffset, defReplicateFolder);
         cloneRelationships();
     }
     virtual void resolveFiles(const StringArray &locations, const char *remoteIP, const char *_remotePrefix, const char *srcCluster, bool checkLocalFirst, bool addSubFiles, bool trackSubFiles, bool resolveForeign=false) override;
 
     void resolveSubFiles(StringArray &subfiles, const StringArray &locations, bool checkLocalFirst, bool trackSubFiles, bool resolveForeign);
+    bool needsCopying(bool cloneForeign);
```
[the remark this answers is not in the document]
afishbeck: Renaming.  This is the container class.  The "needsCopying" here isn't part of an interface.  I'll change the name to be less confusing.

common/pkgfiles/referencedfilelist.cpp
```diff
@@ -474,13 +490,203 @@ void ReferencedFile::resolve(const char *dstCluster, const char *srcCluster, IUs
         locations.append(dstCluster);
 }
 
-void ReferencedFile::cloneInfo(unsigned updateFlags, IDFUhelper *helper, IUserDescriptor *user, const char *dstCluster, const char *srcCluster, bool cloneForeign, unsigned redundancy, unsigned channelsPerNode, int replicateOffset, const char *defReplicateFolder)
+static void getNodeGroupFromLFN(IUserDescriptor *user, const char* lfn, StringBuffer& nodeGroup)
```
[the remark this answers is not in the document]
afishbeck: Removing.

common/pkgfiles/referencedfilelist.cpp
```diff
@@ -474,13 +490,203 @@ void ReferencedFile::resolve(const char *dstCluster, const char *srcCluster, IUs
...
+{
+    Owned<IDistributedFile> df = queryDistributedFileDirectory().lookup(lfn, user, AccessMode::tbdRead, false, false, nullptr, defaultPrivilegedUser, INFINITE);
```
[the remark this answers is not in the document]
afishbeck: Right.  Once the new way is integrated with DFUserver I will make ReferencedFileList work the new way.

common/pkgfiles/referencedfilelist.cpp
```diff
@@ -474,13 +490,203 @@ void ReferencedFile::resolve(const char *dstCluster, const char *srcCluster, IUs
...
+    if (!df)
+        throw makeStringExceptionV(-1, "Failed to find file: %s", lfn);
+    df->getClusterGroupName(0, nodeGroup);
+}
+
+static bool parseLogicalPath(const char * pLogicalPath, const char* groupName, const char* cluster, StringBuffer &folder, StringBuffer &title, StringBuffer &defaultFolder, StringBuffer &defaultReplicateFolder)
```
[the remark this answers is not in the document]
afishbeck: I've written a shared version.  I may not actually use it in CFileSprayEx yet to minimize risk in this PR.

dali/dfu/dfuwu.cpp
```diff
@@ -306,6 +318,55 @@ class CLinkedDFUWUchild : public CInterface
...
+            break;
+        default:
+            return false;
+    }
+    StringBuffer publisherRoot;
+    getSubTaskParentXPath(publisherRoot, subtaskWuid);
+
+    //Only lock the Progress section to avoid parent child lock issues
+    publisherRoot.append("/Progress");
+    Owned<IRemoteConnection> conn = querySDS().connect(publisherRoot.str(), myProcessSession(), RTM_LOCK_WRITE, SDS_LOCK_TIMEOUT);
```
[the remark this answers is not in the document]
afishbeck: Updating to 5 mins.  Not sure what the ideal time would be.

esp/scm/ws_workunits_struct.ecm
```diff
@@ -515,7 +515,7 @@ ESPStruct [nil_remove] ResubmittedWU
 
 ESPStruct WUFileOption
 {
-    ESPenum WUFileType FileType;
+    ESPenum WUFileType FileType("XML");
```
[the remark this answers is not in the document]
afishbeck: Yes.  Include extra commit a while back for build break.  Will rebase now.

afishbeck:
@ghalliday updated.  I've created a common ParseLogicalPath and readded the baremetal stuff, but haven't changed ws_fs to use it.. didn't want to add testing and risk to this PR.

Originally I specifically didn't support baremetal for this feature.. now that I do I still have to find a way to test it.

common/pkgfiles/referencedfilelist.cpp
```diff
@@ -474,13 +490,203 @@ void ReferencedFile::resolve(const char *dstCluster, const char *srcCluster, IUs
...
+    if(isEmptyString(destNodeGroup))
+        throw MakeStringException(-1, "Destination node group not specified.");
+
+    PROGLOG("Copy from [%s]  %s to %s", isEmptyString(srcDali) ? "local" : srcDali, sourceLogicalName, destLogicalName);
+
+    StringBuffer destFolder, destTitle, defaultFolder, defaultReplicateFolder;
+    parseLogicalPath(destLogicalName, destNodeGroup, NULL, destFolder, destTitle, defaultFolder, defaultReplicateFolder);
+
+    StringBuffer fileMask;
+    constructFileMask(destTitle.str(), fileMask);
```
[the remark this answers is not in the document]
afishbeck: @jakesmith can you comment?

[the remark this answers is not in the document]
jakesmith: I'm fairly sure that it is not needed.
If it's not set - dfuserver when creating the IFileDescriptor, will create a mask at that point based on the group/lfn name.

ghalliday (approved):
@afishbeck there are a couple of minor comments not addressed, but it looks good to merge, and I doubt that I will spot any other issues.

common/pkgfiles/referencedfilelist.cpp
```diff
@@ -474,13 +490,203 @@ void ReferencedFile::resolve(const char *dstCluster, const char *srcCluster, IUs
...
+    if (supercopy)
+        wu->setCommand(DFUcmd_supercopy);
+    else
+        wu->setCommand(DFUcmd_copy);
+
+    IDFUfileSpec *wuFSpecSource = wu->queryUpdateSource();
+    IDFUfileSpec *wuFSpecDest = wu->queryUpdateDestination();
+    IDFUoptions *wuOptions = wu->queryUpdateOptions();
+    wuFSpecSource->setLogicalName(sourceLogicalName);
+    if (!isEmptyString(srcDali))
```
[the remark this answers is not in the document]
afishbeck: Opened https://track.hpccsystems.com/browse/HPCC-27842

afishbeck:
@ghalliday I think I took care of the last of the trivial issues.  I think it's ready.  I can squash tomorrow unless there is anything else.

common/pkgfiles/CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
...
+
+project( pkgfiles )
+
+set (    SRCS
+         package.cpp
+         referencedfilelist.cpp
+    )
+
+include_directories (
+        ./../../system/include
```
jakesmith, line 33: personal: I much prefer the explicit fully qualified form for including paths, e.g. ${HPCC_SOURCE_DIR}/system/include

common/pkgfiles/CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
...
+        ./../../system/jlib
+         ./../../system/security/shared
```
jakesmith, line 35: trivial: variable indentation

common/pkgfiles/referencedfilelist.cpp
```diff
@@ -240,24 +248,32 @@ class ReferencedFileList : implements IReferencedFileList, public CInterface
...
     {
-        cloneFileInfo(dstCluster, updateFlags, helper, cloneSuperInfo, cloneForeign, redundancy, channelsPerNode, replicateOffset, defReplicateFolder);
+        cloneFileInfo(publisherWuid, dstCluster, updateFlags, helper, cloneSuperInfo, cloneForeign, redundancy, channelsPerNode, replicateOffset, defReplicateFolder);
         cloneRelationships();
     }
     virtual void resolveFiles(const StringArray &locations, const char *remoteIP, const char *_remotePrefix, const char *srcCluster, bool checkLocalFirst, bool addSubFiles, bool trackSubFiles, bool resolveForeign=false) override;
 
     void resolveSubFiles(StringArray &subfiles, const StringArray &locations, bool checkLocalFirst, bool trackSubFiles, bool resolveForeign);
+    virtual bool filesNeedCopying(bool cloneForeign);
+    void setDfuQueue(const char *queue) override
```
jakesmith, line 263: trivial/style: missing virtual

common/pkgfiles/referencedfilelist.cpp
```diff
@@ -474,13 +490,145 @@ void ReferencedFile::resolve(const char *dstCluster, const char *srcCluster, IUs
...
+        if (target.hasProp("@queue"))
+            dfuQueueName.set(target.queryProp("@queue"));
+    }
+#endif
+}
+
+static void dfuCopy(const IPropertyTree *directories, IDFUWorkUnit *publisherWu, IUserDescriptor *user, const char *sourceLogicalName, const char *destLogicalName, const char *destNodeGroup, const char *srcDali, bool supercopy, bool overwrite, bool preserveCompression, bool nosplit)
+{
+    if(!publisherWu)
+        throw MakeStringException(-1, "Failed to create Publisher DFU Workunit.");
```
jakesmith, line 536: trivial: this may be moved/copied code. Generally MakeStringException is considered deprecated in favour of either makeStringException or makeStringExceptionV depending on whether it has variadic args

dali/dfu/dfuwu.cpp
```diff
@@ -233,6 +245,70 @@ StringBuffer &encodeDFUsortfield(DFUsortfield fmt,StringBuffer &str,bool incmodi
...
+    if(!pLogicalPath || !*pLogicalPath)
+        return false;
+
+    folder.clear();
+    title.clear();
+
+    defaultFolder.clear();
+    defaultReplicateFolder.clear();
+
+    if(groupName != NULL && *groupName != '\0')
```
jakesmith, line 260: trivial: this and line 251 could use isEmptyString

dali/dfu/dfuwu.cpp
```diff
@@ -233,6 +245,70 @@ StringBuffer &encodeDFUsortfield(DFUsortfield fmt,StringBuffer &str,bool incmodi
...
+                StringBuffer &folder, StringBuffer &title, StringBuffer &defaultFolder, StringBuffer &defaultReplicateFolder)
+{
+    if(!pLogicalPath || !*pLogicalPath)
+        return false;
+
+    folder.clear();
+    title.clear();
+
+    defaultFolder.clear();
+    defaultReplicateFolder.clear();
```
jakesmith, line 258: style: unusual to clear a StringBuffer on entry to a function. Normally semantics are that it functions only append to StringBuffers' allowing caller to pre-pend if they want to.
(and DfuParseLogicalPath is only called in 1 place, where always blank at the moment anyway)

common/pkgfiles/referencedfilelist.cpp
```diff
@@ -474,13 +490,145 @@ void ReferencedFile::resolve(const char *dstCluster, const char *srcCluster, IUs
...
+    if(isEmptyString(sourceLogicalName))
+        throw MakeStringException(-1, "Source logical file not specified.");
+    if(isEmptyString(destLogicalName))
+        throw MakeStringException(-1, "Destination logical file not specified.");
+    if(isEmptyString(destNodeGroup))
```
jakesmith, line 541: we should be transitioning toward using 'plane' - we'll have a mixed system which refers to groups (bare-metal) for some time, but eventually all systems should be referring to planes.
(and in this case the caller of dfuCopy actually further overloads meaning by calling it a cluster)

@ghalliday - thoughts? .. all new code use 'plane' vs 'group' (or 'cluster') where possible?

dali/dfu/dfuwu.cpp
```diff
@@ -64,9 +75,9 @@ static void removeTree(IPropertyTree *root,const char *name)
         root->removeTree(t);
 }
 
-static StringBuffer &newWUID(StringBuffer &wuid)
+static StringBuffer &newWUID(StringBuffer &wuid, const char prefix = 'D')
```
jakesmith, line 78: minor: prob. better to avoid defaults like this, can lead to confusion, better for the 1 pre-existing call in createWorkUnit() to explicitly pass 'D'

dali/dfu/dfuwu.cpp
```diff
@@ -233,6 +245,70 @@ StringBuffer &encodeDFUsortfield(DFUsortfield fmt,StringBuffer &str,bool incmodi
...
+    getLFNDirectoryUsingBaseDir(folder, pLogicalPath, defaultFolder.str());
+
+    const char *n = pLogicalPath;
+    const char* p;
+    do {
+        p = strstr(n,"::");
+        if(p)
+            n = p+2;
+    } while(p);
+    title.append(n);
```
jakesmith, line 307: for slightly loss of efficiency may be better to use standard lfn parsing code like:
```
CDfsLogicalFileName lfn(pLogicalPath);
lfn.getTail(title);
```

dali/dfu/dfuwu.cpp
```diff
@@ -306,6 +382,55 @@ class CLinkedDFUWUchild : public CInterface
...
+    }
+    StringBuffer publisherRoot;
+    getSubTaskParentXPath(publisherRoot, subtaskWuid);
+
+    //Only lock the Progress section to avoid parent child lock issues
+    publisherRoot.append("/Progress");
+    Owned<IRemoteConnection> conn = querySDS().connect(publisherRoot.str(), myProcessSession(), RTM_LOCK_WRITE, NOTE_SUBTASK_SDS_LOCK_TIMEOUT);
+    if (!conn)
+        return false;
+    OwnedPTree root = conn->getRoot();
```
jakesmith, line 415: trivial: don't really need to link root here

dali/dfu/dfuwu.cpp
```diff
@@ -2972,6 +3135,34 @@ class CDFUWorkUnitFactory : implements IDFUWorkUnitFactory, implements ISDSSubsc
         // created time stamp? TBD
         return ret;
     }
+
+    IDFUWorkUnit * createPublisherWorkUnit() override
+    {
+        StringBuffer wuid;
+        newWUID(wuid, 'P');
```
jakesmith, line 3142: question: does this new dfu workunit name format pose any issues for esp/eclwatch services?
(it really shouldn't - as they should be using interfaces to get results etc.)

dali/dfu/dfuwu.cpp
```diff
@@ -2972,6 +3135,34 @@ class CDFUWorkUnitFactory : implements IDFUWorkUnitFactory, implements ISDSSubsc
...
+    IDFUWorkUnit * createPublisherSubTask(IDFUWorkUnit *publisherWu) override
+    {
+        IDFUprogress *progress = publisherWu->queryUpdateProgress();
+        unsigned taskId = progress->incPublisherTaskCount();
+
+        StringBuffer wuRoot;
+        getXPath(wuRoot, publisherWu->queryId(), taskId);
+        IRemoteConnection* conn = querySDS().connect(wuRoot.str(), myProcessSession(), RTM_LOCK_WRITE|RTM_CREATE_UNIQUE, SDS_LOCK_TIMEOUT);
+        conn->queryRoot()->setProp("@xmlns:xsi", "http://www.w3.org/1999/XMLSchema-instance");
+        IDFUWorkUnit *ret = new CDFUWorkUnit(this, conn, NULL, true);
```
jakesmith, line 3161: minor: safer to make owned, then return via getClear()

I'd also own 'conn' and getClear() when passing ownership to constructor

esp/services/ws_workunits/ws_workunitsQuerySets.cpp
```diff
@@ -2876,6 +2934,59 @@ class QueryCloner
...
+    {
+        OwnedPTree entry = toBePublished->addPropTree("Publish");
+        entry->setProp("@wuid", wuid);
+        entry->setProp("@name", name);
+        entry->setPropBool("@makeActive", makeActive);
+        entry->setProp("@userid", userid);
+        entry->addPropTree("Info", createPTreeFromIPT(query));
+
+        Owned<IAttributeIterator> aiter = query->getAttributes();
+        OwnedPTree attrs = entry->addPropTree("Attrs");
```
jakesmith, line 2948: doesn't look like correct to own this (or the one on line 2940 above).

addPropTree returns a unlinked ptr to new IPT added, owning here is afaics, going to destroy it, whilst it remains in the parent.

jakesmith (commented):
@afishbeck - please see comments

dali/dfu/dfuwu.cpp
afishbeck: Agreed, will change.

dali/dfu/dfuwu.cpp
afishbeck: Right.  Mostly copied code, wrote this as shared code, but changing the other calls is outside the scope of this PR.  But I'll change as suggested.

common/pkgfiles/referencedfilelist.cpp
afishbeck: agreed.

common/pkgfiles/CMakeLists.txt
afishbeck: agreed.

dali/dfu/dfuwu.cpp
afishbeck: It doesn't break EclWatch for the old workunits.  

I was waiting to see if this way of doing things got approved and then I need to open a jira about improving the way it interacts with the new ones.

In the future opening a publisher workunit should open 2 child tabs:
   1. A list of the child dfu workunits (same control as listing DFU workunits now).
   2. The publisher workunit xml.

common/pkgfiles/referencedfilelist.cpp
afishbeck: I'll change the name in dfuCopy, but would rather not change the names in other pre-existing functions in this PR.

afishbeck:
@jakesmith updated.

jakesmith (approved):

afishbeck:
squashed

PR #15944, opened 2022-03-25
HPCC-27421 Update schema to allow definition of alternative plane access

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/helm/hpcc/templates/_helpers.tpl
+++ b/helm/hpcc/templates/_helpers.tpl
@@ -214,15 +214,25 @@ storage:
 {{- range $plane := $planes }}
  {{- if not $plane.disabled }}
   - name: {{ $plane.name | quote }}
-  {{- $planeYaml := omit $plane "name" "pvc" "storageClass" "storageSize" "subPath" -}}
+  {{- $planeYaml := omit $plane "name" "pvc" "storageClass" "storageSize" "subPath" "numMounts" -}}
   {{- if $plane.subPath -}}
    {{- $_ := set $planeYaml "prefix" (printf "%s/%s" $planeYaml.prefix $plane.subPath) -}}
   {{- end -}}
   {{- if and (eq "data" $plane.category) (not $plane.defaultSprayParts) -}}
    {{- $_ := set $planeYaml "defaultSprayParts" (include "hpcc.getMaxNumWorkers" $ | int) -}}
   {{- end -}}
+
+  {{- /* Remove pvc-related properties from the aliases*/ -}}
+  {{- if $plane.aliases }}
+   {{- $_ := set $planeYaml "aliases" (deepCopy $plane.aliases) -}}
+   {{- range $alias := $planeYaml.aliases -}}
+    {{- $_ := unset $alias "pvc" }}
+    {{- $_ := unset $alias "numMounts" }}
+   {{- end -}}
+  {{- end -}}
   {{- toYaml $planeYaml | nindent 4 }}
  {{- end }}
+
 {{- end }}
 {{- if not (include "hpcc.hasPlaneForCategory" (dict "root" $ "category" "spill")) }}
   - name: hpcc-spill-plane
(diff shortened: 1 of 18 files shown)


ghalliday:
@richardkchapman @jakesmith  this is my suggested schema changes.  Essentially the plane properties have been split in two - those that only apply to the storage plane as a logical item, and those that refer to the information needed to get at the data (access options).  The idea is that you would have something like the following:

```
  - name: data
    pvc: data-blob-nfs-pvc
    prefix: "/var/lib/HPCCSystems/hpcc-data"
    category: data
    access:
    - reason: [ random ]
      pvc: data-cache-pvc
      prefix: "/var/lib/HPCCSystems/hpcc-data-cache"
    - reason: [ api ]
      prefix: azure://ghallidaystorage  # not really sure what this syntax should be - likely to change...
      secret: azure-ghallidaystorage
```
When a file is looked up the reason (write, sequential, random, api, ...) is provided.  If there is an alias for that reason, then a the associated information in the alias is used instead of the default plane definition.  (It replaces all access-specific properties, not just the properties that are defined.)

I think the easiest way for the aliases to be processed would be for them to be expanded as complete plane definitions - i.e. have all the information from the plane cloned into them, so that the alias object can be returned and used in the same way that planes currently are.  That could be done in the helm generation, but probably easier in the configuration processing when the config file is read.

(I toyed with expanding them as planes with a name of <plane>:<number>, but that makes subsequent processing more complex.)

Need to be careful in the helm processing to check aliases do not violate the rules about duplicate pvcs that planes are checked against.

Jake, there will need to be some code to map the physical filenames.  Does it need to store it in dali using the default path?  Not sure what complications there are - I don't think they directly affect this change.

Some questions
- Is "access" a good name?  Is "aliases" better? (probably)
- Should it disallow storage options that make no sense - e.g., storageClass/Size/Mode since they can't be used for an alias.
- An alternative to a list would be to have a dictionary mapped by reason.  That would not allow the multiple reasons to be used for the same alias, which seems like too much of a restriction.  Do you agree?
- Should "write" be allowed as a reason, or is it a requirement that the default plane definition is used for writing?
- Any other reasons?
- We could add additional flags e.g. supports seek on write that could be used by the engines.

ghalliday:
I have thought of one complication.  I suspect hpc-cache is mounted as a single device - but we still want the multiple device directories.  The simplest fix is to add a numMounts property - which can be set on each alias (or the default) to override the number of associated pvcs.  The rest of the code should use numDevices, so the changes are minimal.

richardkchapman:
I think I would replace access: with alias: and reason: with access:

Other than that looks sensible

jakesmith:
> Is "access" a good name? Is "aliases" better? (probably)

Agree "alias" would be better as replacement for "access".

> Should it disallow storage options that make no sense - e.g., storageClass/Size/Mode since they can't be used for an alias.

ideally yes, to avoid attempts/confusion from naive users trying to use inappropriate options.

> An alternative to a list would be to have a dictionary mapped by reason. That would not allow the multiple reasons to be used for the same alias, which seems like too much of a restriction. Do you agree?

agree, reasons should remain a list.

> Should "write" be allowed as a reason, or is it a requirement that the default plane definition is used for writing?

I can envisage a plane without any aliases, being read-only, but that's a slightly different topic.
It does (again) make me wonder whether the default should just be another reason alias/access in the list, so that even for the simple case where only 1 way to access, you can provide ways it can be accessed - access: [read-only]
But appreciate that's a complication for backward compatibility in the schema/helper code.

> Any other reasons?

not reasons (aka access type), but a cost per route might factor in at some point, e.g. some engines might need fast and expensive, others might not care so much, and/or might even be a per query setting to reduce cost.

jakesmith:
>Jake, there will need to be some code to map the physical filenames. Does it need to store it in dali using the default path? Not sure what complications there are - I don't think they directly affect this change.

we already have that problem or inconsistency with multiple "Cluster"'s  currently when an lfn is on multiple planes.
The original path is stored at the copy-level (@dir) of the logical file, and then from what I remember, there's some horrible code that remaps it if an alternative cluster is chosen.
I will have to look at how/where best to remap as file is looked up.

jakesmith:
see comments, but looks sensible overall to me.

ghalliday:
HPCC-27427 For the documentation jira

helm/hpcc/templates/_helpers.tpl
```diff
@@ -214,15 +214,25 @@ storage:
...
   {{- if $plane.subPath -}}
    {{- $_ := set $planeYaml "prefix" (printf "%s/%s" $planeYaml.prefix $plane.subPath) -}}
   {{- end -}}
   {{- if and (eq "data" $plane.category) (not $plane.defaultSprayParts) -}}
    {{- $_ := set $planeYaml "defaultSprayParts" (include "hpcc.getMaxNumWorkers" $ | int) -}}
   {{- end -}}
+
+  {{- /* Remove pvc-related properties from the aliases*/ -}}
+  {{- if $plane.aliases }}
+   {{- $_ := set $planeYaml "aliases" (deepCopy $plane.aliases) -}}
```
jakesmith, line 227: not sure this is necessary? "aliases" will already be a key of $planeYaml afaics.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -300,6 +310,8 @@ Add ConfigMap volume for a component
     name: {{ .name }}-configmap
 {{- end -}}
 
+
+
```
jakesmith, line 314: trivial: unnecessary newlines

helm/hpcc/templates/_helpers.tpl
```diff
@@ -351,26 +388,51 @@ The plane will generate a volume if it matches either an includeLabel or an incl
...
+     {{- if not (hasKey $previousMounts $alias.prefix) }}
+      {{- $pvc := $alias.pvc }}
+      {{- $numMounts := int ( $alias.numMounts | default $plane.numDevices | default 1 ) }}
+      {{- if le $numMounts 1 }}
+- name: {{ lower $plane.name }}-pv-alias-{{ $curAlias.num }}
+  persistentVolumeClaim:
+    claimName: {{ $pvc }}
+      {{- else }}
+       {{- range $elem := until $numMounts }}
+- name: {{ lower $plane.name }}-pv-alias-{{ $curAlias.num }}-many-{{- add $elem 1 }}
```
jakesmith, line 425: trivial: {{- add $elem 1 }} , I don't think the remove leading white space syntax needed here, harmless, but made me look.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -351,26 +388,51 @@ The plane will generate a volume if it matches either an includeLabel or an incl
...
+    claimName: {{ $pvc }}-{{- add $elem 1 }}
+       {{- end }}
+      {{- end }}
+      {{- $_ := set $previousMounts $alias.prefix true }}
+      {{- $_ := set $curAlias "num" (add $curAlias.num 1) }}
+     {{- end }}
     {{- end }}
-    {{- $_ := set $previousMounts $plane.prefix true -}}
    {{- end }}
+
```
jakesmith, line 435: trivial/formatting: newline could be removed I think

helm/hpcc/values.schema.json
```diff
@@ -525,15 +560,37 @@
...
+      }
+    },
+    "storagePlaneAlias": {
+      "$comment": "The following properties are required for providing an alternative path to a data plane",
+      "allOf": [
+        { "$ref": "#/definitions/storagePlaneCommon" },
+        {
+          "properties": {
+            "mode": {
+              "description": "A list of access modes the engines would use select this alias",
```
jakesmith, line 573: trivial: "would use to select"

testing/helm/tests/manypvc.yaml
```diff
@@ -0,0 +1,43 @@
...
+    numDevices: 5
+    aliases:
+    - pvc: cachedata-localmany-hpcc-localfile-pvc
+      prefix: "/var/lib/HPCCSystems/manydatacache"
+      mode: [ random ]
+      numMounts: 1
+    - pvc: cachedata-localmany-hpcc-localfile-pvc
+      prefix: "/var/lib/HPCCSystems/manydatacache2"
+      mode: [ sequential ]
+      numMounts: 3
```
jakesmith, line 32: question: this could make sense, but we can't conceive of a scenario at the moment, where there is a 5 PVC plane, with a 3 mount alias can we?

it would still stripe (on read access) according to numDevices, but then need to manipulate the leading prefix according to numMounts... but which mount would have which stripe?

jakesmith (commented):
@ghalliday - a few comments and questions

helm/hpcc/templates/_helpers.tpl
jakesmith: $planeYaml is already a copy of the original dictionary though, because 'omit' which created it, creates a new dictionary:
"The omit function is similar to pick, except it returns a new dict with all the keys that do not match the given keys."

.. which is desirable, because of the other preceding set $planeXML ... calls, that alter the copy and not the original.

I assumed that te copy (via omit) also included a copy of it's child dictionaries (like "aliases") too ?

ghalliday: $planeYaml is a copy of the dictionary - but I don't think the items in the dictionary are copies.  So if you modify them they will modify both.

ghalliday: This code was actually wrong, and revealed exactly this problem when I added the consistency check for the number of devices!

ghalliday:
@jakesmith helm tests are also improved to cause template expansion - since various non schema errors were not being caught in valid tests

jakesmith (approved):
@ghalliday - looks good

richardkchapman (approved):

PR #15952, opened 2022-03-30
HPCC-27437 Fix index read activity blob leaks.

jakesmith:
A few places, where filtering was possible via helper calls,
were not ensuring that any loaded blob memory was cleared.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/ecl/hthor/hthorkey.cpp
+++ b/ecl/hthor/hthorkey.cpp
@@ -1226,7 +1226,10 @@ const void *CHThorIndexNormalizeActivity::nextRow()
                 {
                     expanding = helper.next();
                     if (!expanding)
+                    {
+                        callback.finishedRow(); // next could filter
                         break;
+                    }
 
                     const void * ret = createNextRow();
                     if (ret)
@@ -1234,7 +1237,6 @@ const void *CHThorIndexNormalizeActivity::nextRow()
                 }
             }
 
-            callback.finishedRow();
             while (!klManager->lookup(true))
             {
                 keyedProcessed++;
@@ -1253,6 +1255,8 @@ const void *CHThorIndexNormalizeActivity::nextRow()
                 if (ret)
                     return ret;
             }
+            else
+                callback.finishedRow(); // first could filter
         }
     }
 }
@@ -1263,6 +1267,7 @@ const void * CHThorIndexNormalizeActivity::createNextRow()
     {
         outBuilder.ensureRow();
         size32_t thisSize = helper.transform(outBuilder);
+        callback.finishedRow();
         if (thisSize == 0)
         {
             return NULL;
--- a/roxie/ccd/ccdserver.cpp
+++ b/roxie/ccd/ccdserver.cpp
@@ -22499,6 +22499,7 @@ public:
             }
             transformedSize = normalizeHelper->transform(rowBuilder);
             firstPending = !normalizeHelper->next();
+            reader->finishedRow();
             if (transformedSize)
                 break;
         }
--- a/thorlcr/activities/indexread/thindexreadslave.cpp
+++ b/thorlcr/activities/indexread/thindexreadslave.cpp
@@ -104,6 +104,7 @@ protected:
     protected:
         CIndexReadSlaveBase &activity;
         IKeyManager *keyManager = nullptr;
+        bool needsBlobCleaning = false;
     public:
         TransformCallback(CIndexReadSlaveBase &_activity) : activity(_activity) { };
         IMPLEMENT_IINTERFACE_O_USING(CSimpleInterface)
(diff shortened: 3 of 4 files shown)


thorlcr/activities/indexread/thindexreadslave.cpp
```diff
@@ -763,11 +768,9 @@ class CIndexReadSlaveActivity : public CIndexReadSlaveBase
                 if (needTransform)
                 {
                     size32_t sz = helper->transform(ret, r);
+                    callback.finishedRow();
```
jakesmith, line 771: logically I would like to put this call in an exit scope (COnScopeExit) just before the canMatch, as anything after it can use it, and would make the code safer and cleaner.
But I think it (and it's use of std::function's) does add a fair amount of overhead, including heap allocation from what I've read), so I avoid it.

ecl/hthor/hthorkey.cpp
```diff
@@ -1263,6 +1267,7 @@ const void * CHThorIndexNormalizeActivity::createNextRow()
     {
         outBuilder.ensureRow();
         size32_t thisSize = helper.transform(outBuilder);
+        callback.finishedRow();
```
ghalliday, line 1270: Is this needed?

thorlcr/activities/indexread/thindexreadslave.cpp
```diff
@@ -1471,7 +1473,10 @@ class CIndexNormalizeSlaveActivity : public CIndexReadSlaveBase
                 {
                     expanding = helper->next();
                     if (!expanding)
+                    {
```
ghalliday, line 1476: Like the changes in hthor, I think this code is identical to the code that was there before.

ghalliday (commented):
@jakesmith I think the code would work, and the index read cases fix some potential leaks.  I'm not sure the normalize changes have any effect.

ecl/hthor/hthorkey.cpp
```diff
@@ -1253,6 +1255,8 @@ const void *CHThorIndexNormalizeActivity::nextRow()
                 if (ret)
                     return ret;
             }
+            else
+                callback.finishedRow(); // first could filter
```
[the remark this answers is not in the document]
jakesmith: >I think the result of this change is identical. Is there a case when it is called at a different time?

it is mainly that it wasn't called after next()  (and I had assumed blob could be loaded by first() + next()), and with the added call within createNextRow it would have called finishedRow twice needlessly, so the removal of the finishedRow call on line 1237, and the explicit call to it on the !expanding is to replace that.

Now that I know it is not needed after next() then I can change it.

ecl/hthor/hthorkey.cpp
jakesmith: needs a call to finishedRow in the nextRow loop somewhere, used to have one in the loop at line 1237, now explicit here or if !expanding

richardkchapman (approved):
Not perfect and may want to revisit, but won't make things worse

roxie/ccd/ccdserver.cpp
```diff
@@ -22499,6 +22499,7 @@ class CRoxieServerDiskNormalizeActivity : public CRoxieServerDiskReadBaseActivit
             }
             transformedSize = normalizeHelper->transform(rowBuilder);
             firstPending = !normalizeHelper->next();
+            reader->finishedRow();
```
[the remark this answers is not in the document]
jakesmith: hm, yes. I'll remove

thorlcr/activities/indexread/thindexreadslave.cpp
```diff
@@ -1388,6 +1389,7 @@ class CIndexNormalizeSlaveActivity : public CIndexReadSlaveBase
     {
         RtlDynamicRowBuilder row(allocator);
         size32_t sz = helper->transform(row);
+        callback.finishedRow();
```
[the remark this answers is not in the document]
jakesmith: it's needed afaics, because, like the hthorkey changes, it used to call finishedRow in the loop, it is now called here, and if !expanding

thorlcr/activities/indexread/thindexreadslave.cpp
jakesmith: now I know next() doesn't need a called to finishedRow, yes I think so.

richardkchapman: Why doesn't next call lookupBlob?

PR #15958, opened 2022-03-31
HPCC-27293 Enable plugins for Ubuntu vcpkg build

GordonSmith:
## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:


plugins/kafka/CMakeLists.txt
```diff
@@ -27,58 +27,74 @@ project(kafka)
...
-        if(NOT EXISTS "${PROJECT_SOURCE_DIR}/librdkafka/configure")
-            message(FATAL_ERROR
+        if (VCPKG_APPLOCAL_DEPS)
+            find_package(RdKafka REQUIRED)
+
+            set ( LOCAL_LIBRDKAFKA_LIBS 
+                RdKafka::rdkafka 
+                RdKafka::rdkafka++
+            )
+        else ()
```
GordonSmith, line 37: This "if/else" block could be reworked so it is self contained in commonSetup.cmake (see blas support).

system/security/plugins/jwtSecurity/CMakeLists.txt
```diff
@@ -24,7 +24,7 @@ option(BUILD_TESTS "Configure CMake to build tests (or not)" ON) # Copy from jwt
 set(BUILD_TESTS OFF)
 option(BUILD_EXAMPLES "Configure CMake to build examples (or not)" ON) # Copy from jwt-cpp/CMakeLists.txt
 set(BUILD_EXAMPLES OFF)
-add_subdirectory(jwt-cpp EXCLUDE_FROM_ALL)
+# add_subdirectory(jwt-cpp EXCLUDE_FROM_ALL)
```
GordonSmith, line 27: @richardkchapman I commented this out, but it looks like it could be deleted?

plugins/kafka/CMakeLists.txt
GordonSmith: ...actually it can't as its a "plugin" but it could be cleaned up a little.

GordonSmith: ...also not true.

system/security/plugins/jwtSecurity/CMakeLists.txt
richardkchapman: @dcamper Do you know?

dcamper: https://cmake.org/cmake/help/latest/prop_tgt/EXCLUDE_FROM_ALL.html

IIRC this was part of trying to tightly control the caching/recompilation phase of the underlying code.  Perhaps too tightly.

I would suggest keeping the add_subdirectory() command but deleting the EXCLUDE_FROM_ALL option and see if the results are right.

GordonSmith: Changed as per Dans suggestion

dcamper (approved):

plugins/mysql/CMakeLists.txt
```diff
@@ -58,7 +60,9 @@ if(MYSQLEMBED)
             ${MYSQL_LIBRARIES}
             eclrtl
             roxiemem
-            jlib)
+            jlib
```
richardkchapman, line 63: This looks like it would affect non-vcpkg builds (for example)

GordonSmith: Not really - its just moving the find_package logic out of the macro (which is bit buggy).  The addition of the  ZLIB is simply being more specific due to USE_OPTIONAL "magic".

GordonSmith: IOW in non vcpkg build with USE_PLUGINS=1 and USE_OPTIONAL=1 and ZLIB is not installed, the build will fail.

richardkchapman (approved):

PR #15970, opened 2022-04-05
HPCC-22092 Report Expiration Date in DFUInfoResponse

wangkx:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dali/base/dadfs.cpp
+++ b/dali/base/dadfs.cpp
@@ -3295,9 +3295,21 @@ public:
     virtual void setAccessedTime(const CDateTime &dt) = 0;                      // set date and time last accessed
     virtual bool isExternal() const { return external; }
 
-    virtual int getExpire()
+    virtual int getExpire(StringBuffer *expirationDate)
     {
-        return queryAttributes().getPropInt("@expireDays", -1);
+        int expireDays = queryAttributes().getPropInt("@expireDays", -1);
+        if (!expirationDate || (expireDays == -1))
+            return expireDays;
+
+        const char *lastAccessed = queryAttributes().queryProp("@accessed");
+        if (isEmptyString(lastAccessed))
+            return expireDays;
+
+        CDateTime expires;
+        expires.setString(lastAccessed);
+        expires.adjustTime(60*24*expireDays);
+        expires.getString(*expirationDate);
+        return expireDays;
     }
 
     virtual void setExpire(int expireDays)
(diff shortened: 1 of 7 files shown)


dali/base/dadfs.cpp
```diff
@@ -3295,6 +3295,22 @@ class CDistributedFileBase : implements INTERFACE, public CInterface
     virtual void setAccessedTime(const CDateTime &dt) = 0;                      // set date and time last accessed
     virtual bool isExternal() const { return external; }
 
+    virtual void getExpirationInfo(int &expireDays, StringBuffer &expirationDate)
```
jakesmith, line 3298: It would be better to have a single expiry related method (this new one duplicates the existing functionality of getExpiry()).
I suggest adding StringBuffer *expirationData as an optional parameter to the existing getExpiry()

jakesmith (changes requested):
@wangkx - look ok, but better to avoid having 2 methods in interface that can both get expiryDays. See comment.

wangkx:
@jakesmith the expirationDate has been added as an optional parameter to the existing getExpiry(). Please review.

jakesmith (approved):
@wangkx - looks good.

wangkx:

dali/base/dadfs.cpp
```diff
@@ -3295,9 +3295,21 @@ class CDistributedFileBase : implements INTERFACE, public CInterface
...
+    virtual int getExpire(StringBuffer *expirationDate)
     {
-        return queryAttributes().getPropInt("@expireDays", -1);
+        int expireDays = queryAttributes().getPropInt("@expireDays", -1);
+        if (!expirationDate || (expireDays == -1))
+            return expireDays;
+
+        const char *lastAccessed = queryAttributes().queryProp("@accessed");
+        if (isEmptyString(lastAccessed))
+            return expireDays;
```
richardkchapman, line 3306: Shouldn't this set expirationDate to something?

wangkx: To my understand, since the @accessed is not set, the file will not be expired until the @accessed is set. I am not sure we should set expirationDate to "x days after the file is acccessed".

richardkchapman: @jake, do you know?

jakesmith: I think it's okay as it is. The "accessed" attribute is set by the engines only, so could be missing. If it is, this function can't calculate the expiration date and leaves that param unfilled.

It does beg the question of whether "accessed" should ever be missing if expireDays is set.
Perhaps if unset and setExpire is called, it should also set \@accessed at that point - but that's a separate issue.

NB: files with \@expireDays but without \@accessed will not be deleted (by Sasha).

PR #15979, opened 2022-04-07
HPCC-27339 Store read/write stats for spray (updated)

shamser:
Updated to make compatibility with older ftslave version

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dali/ft/filecopy.cpp
+++ b/dali/ft/filecopy.cpp
@@ -110,7 +110,6 @@ inline void setCanAccessDirectly(RemoteFilename & file)
 #define FAsize              "@size"
 #define FAcompressedSize    "@compressedSize"
 
-
 const unsigned operatorUpdateFrequency = 5000;      // time between updates in ms
 const unsigned abortCheckFrequency = 20000;         // time between updates in ms
 const unsigned sdsUpdateFrequency = 20000;          // time between updates in ms
@@ -388,12 +387,13 @@ bool FileTransferThread::performTransfer()
         sprayer.tgtFormat.serializeExtra(msg, 1);
 
         ForEachItemIn(i2, progress)
-            progress.item(i2).serializeExtra(msg, 1);
+            progress.item(i2).serializeExtra(msg, 1); //must use version==1 to be compatible with old clients
 
         //NB: Any extra data must be appended at the end...
 
         msg.append(sprayer.fileUmask);
-
+        unsigned version = SUPPORTED_MSG_VERSION;
+        msg.append(version);
         if (!catchWriteBuffer(socket, msg))
             throwError1(RFSERR_TimeoutWaitConnect, url.str());
 
@@ -411,7 +411,7 @@ bool FileTransferThread::performTransfer()
 
             OutputProgress newProgress;
             newProgress.deserializeCore(msg);
-            newProgress.deserializeExtra(msg, 1);
+            newProgress.deserializeExtra(msg, version);
             sprayer.updateProgress(newProgress);
 
             LOG(MCdebugProgress(10000), job, "Update %s: %d %" I64F "d->%" I64F "d", url.str(), newProgress.whichPartition, newProgress.inputLength, newProgress.outputLength);
@@ -615,6 +615,8 @@ FileSprayer::FileSprayer(IPropertyTree * _options, IPropertyTree * _progress, IR
     calcedInputCRC = false;
     aborting = false;
     totalLengthRead = 0;
+    totalNumReads = 0;
+    totalNumWrites = 0;
     throttleNicSpeed = 0;
     compressedInput = false;
     compressOutput = options->getPropBool(ANcompress);
@@ -2828,7 +2830,10 @@ void FileSprayer::updateProgress(const OutputProgress & newProgress)
     OutputProgress & curProgress = progress.item(newProgress.whichPartition);
 
     totalLengthRead += (newProgress.inputLength - curProgress.inputLength);
+    totalNumReads += (newProgress.numReads - curProgress.numReads);
+    totalNumWrites += (newProgress.numWrites - curProgress.numWrites);
     curProgress.set(newProgress);
+
     if (curProgress.tree)
         curProgress.save(curProgress.tree);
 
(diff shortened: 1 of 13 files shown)


shamser:
This is an updated PR for HPCC-27339 that is backwardly compatible.

dali/ft/filecopy.cpp
```diff
@@ -388,12 +387,13 @@ bool FileTransferThread::performTransfer()
...
         ForEachItemIn(i2, progress)
-            progress.item(i2).serializeExtra(msg, 1);
+            progress.item(i2).serializeExtra(msg, 1); //must use version==1 to be compatible with old clients
 
         //NB: Any extra data must be appended at the end...
 
         msg.append(sprayer.fileUmask);
-
+        unsigned version = SUPPORTED_MSG_VERSION;
+        msg.append(version);
```
ghalliday, line 396: I would expect a loop here calling 
serializeExtra(msg, 2);
for all the progress elements.  If not needed then there should be a comment why.

dali/ft/ftbase.cpp
```diff
@@ -556,18 +558,21 @@ MemoryBuffer & OutputProgress::deserializeCore(MemoryBuffer & in)
     return in;
 }
 
-MemoryBuffer & OutputProgress::deserializeExtra(MemoryBuffer & in, unsigned version)
+MemoryBuffer & OutputProgress::deserializeExtra(MemoryBuffer & in, unsigned & version)
```
ghalliday, line 561: I'm not sure you gain much from making the version a reference parameter.  The saving is minimal, and it makes the code harder to understand.

dali/ft/ftbase.cpp
```diff
@@ -556,18 +558,21 @@ MemoryBuffer & OutputProgress::deserializeCore(MemoryBuffer & in)
...
-        }
+        in.read(hasCompressed);
+        if (hasCompressed && in.remaining())
+            in.read(compressedPartSize);
+    }
+    if (version>=2)
+    {
+        constexpr unsigned bytesRequired=sizeof(numWrites)+sizeof(numReads);
+        if (in.remaining()>=bytesRequired)
+            in.read(numWrites).read(numReads);
```
ghalliday, line 573: optimization: Could use readPacked (and appendPacked) to shrink the serialization.

dali/ft/ftbase.cpp
```diff
@@ -556,18 +558,21 @@ MemoryBuffer & OutputProgress::deserializeCore(MemoryBuffer & in)
...
-            in.read(hasCompressed);
-            if (hasCompressed)
-                in.read(compressedPartSize);
-            break;
-        }
+        in.read(hasCompressed);
+        if (hasCompressed && in.remaining())
+            in.read(compressedPartSize);
+    }
+    if (version>=2)
```
ghalliday, line 569: This is a bit inconsistent with serializeExtra.  SerializeExtra serialises either version 1 or version 2.  This serialises either version 1 or version 2.  I suspect clearer to match the serialize function, and require an extra call (and remove the verison modifying).

ghalliday (commented):
@shamser looks broadly right.  Some comments about the structure of the functions..

jakesmith:
@shamser - I think this should have a different associate JIRA #/branch name/comment header, or I think it will be confusing when looking at the history JIRA's of the original change, then the JIRA to revert it, then..

dali/ft/fttransform.cpp
```diff
@@ -640,15 +648,24 @@ void TransferServer::deserializeAction(MemoryBuffer & msg, unsigned action)
         tgtFormat.deserializeExtra(msg, 1);
     }
 
+    unsigned ver = 1; //Using ver=1 to avoid breaking old clients
```
jakesmith, line 651: minor: no need for temporary 'ver' here, could initialize new 'version', and read into if msg.remaining() , doing away with need for else branch.

dali/ft/ftbase.cpp
jakesmith: agree, don't think it is worth the reference/changing it.
If there's anything remaining after version 1 deserialization, then it has to be >=2 and it is guaranteed to be compatible with current version. So this code can unconditionally deserialize numWrites/numReads if more remaining() (as guaranteed to be at least ver. 2).
If anything else is wanted in future, it can conditionally deserialize based on version.

jakesmith (commented):

dali/ft/ftbase.cpp
shamser: I can't assume that any remaining data will be the new numReads and numWrites because in the last message at the end it expects to read a boolean - in line 424 of filecopy.cpp there is: msg.read(ok).  The problems are
1) new dfuserver with old ftslave: the boolean "ok" field  will be mistaken for the numreads.  I believe dfuserver will crash as there would be enough data to read both numreads+numwrites+ok.   2) old dfuserver with new ftslave: ftslave deserializes numreads which is read as the "ok" field.

The current code attempts to negotiate the versions so that mismatched ftlave and dfuserver can work together.

Please let me know if there is a simpler way.

shamser:
@jakesmith @ghalliday I've created a new jira for this issue https://track.hpccsystems.com/browse/HPCC-27554.  I'll move over this PR to the new Jira.  However, I thought it would help to complete the discussion here to avoid you having repeat the discussion in the new ticket.  Please can you consider my response regarding making use of versions: https://github.com/hpcc-systems/HPCC-Platform/pull/15979#discussion_r854347374.

dali/ft/ftbase.cpp
jakesmith: >I can't assume that any remaining data will be the new numReads and numWrites because in the last message at the end it expects to read a boolean - in line 424 of filecopy.cpp there is: msg.read(ok).

ok, not if anything remaining, but if > than your bytesRequired... then it has to be >= version 2, and is guaranteed to be compatible. Any future version (>2) can conditionally deserialize based on version.

So I think reference should be removed, version left in for future compatibility (but not actually needed yet):

```
MemoryBuffer & OutputProgress::deserializeExtra(MemoryBuffer & in, unsigned version)
{
    if (in.remaining())
    {
        // version 1
        in.read(hasCompressed);
        if (hasCompressed)
            in.read(compressedPartSize);

        constexpr unsigned bytesRequired=sizeof(numWrites)+sizeof(numReads);
        if (in.remaining()>=bytesRequired) // must be version >=2
        {
            // NB: server has sent clients it's version and guaranteed this meta info to be compatible.
            // version 2 
            in.read(numWrites).read(numReads);
            // if version >2 can conditionally deserialize more
        }
    }
    return in;
}
```

jakesmith (commented):
@shamser - please see reply

shamser:
Re-opening issue as a new jira: https://track.hpccsystems.com/browse/HPCC-27554

PR #16005, opened 2022-04-14
HPCC-27451 LDAPEnvironment issues with Active Directory

RussWhitehead:
Fixed several AD issues
1) OU built incorrectly, changed resource type to RT_FILE_SCOPE
2) Don't include double quotes in kubectl password creation
3) Don't catch init errors when parsing environment
4) Don't display empty keys in HELM response (Vault ID)

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [x] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/services/ldapenvironment/ldapenvironmentService.cpp
+++ b/esp/services/ldapenvironment/ldapenvironmentService.cpp
@@ -26,7 +26,7 @@ limitations under the License.
 void CldapenvironmentEx::init(IPropertyTree *_cfg, const char *_process, const char *_service)
 {
     cfg = _cfg;
-    try
+
     {
 #ifdef _DEBUG
         StringBuffer sb;
@@ -71,11 +71,6 @@ void CldapenvironmentEx::init(IPropertyTree *_cfg, const char *_process, const c
         cfg->getProp(xpath.str(), sharedWorkunitsBaseDN);
         if (sharedWorkunitsBaseDN.isEmpty())
             throw MakeStringException(-1, "sharedWorkunitsBaseDN must be specified in configuration (ex. 'ou=workunits,ou=shared,ou=hpcc,dc=myldap,dc=com')");
-
-    }
-    catch (...)
-    {
-        throw MakeStringException(-1, "ldapenvironment: Error querying environment settings");
     }
 }
 
@@ -125,7 +120,7 @@ bool CldapenvironmentEx::changePermissions(const char * ou, const char * userFQD
     action.m_action = "update";
     action.m_basedn = baseDN;
     action.m_rname =  rName;
-    action.m_rtype = RT_DEFAULT;
+    action.m_rtype = RT_FILE_SCOPE;
     action.m_account_name = userFQDN;//fully qualified user DN
     action.m_account_type = USER_ACT;
     action.m_allows = allows;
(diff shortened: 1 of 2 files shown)


RussWhitehead:
@afishbeck This fixes several issues with AD . Please review

esp/services/ldapenvironment/ldapenvironmentService.cpp
```diff
@@ -155,7 +150,7 @@ bool CldapenvironmentEx::createSecret(SecretType type, const char * secretName,
             break;
     }
 
-    VStringBuffer cmdLine("%s\"%s\"", cmdLineSafe.str(), pwd);
+    VStringBuffer cmdLine("%s%s", cmdLineSafe.str(), pwd);
```
afishbeck, line 153: Is this reversing your prior fix for extended characters in the password?

esp/services/ldapenvironment/ldapenvironmentService.cpp
```diff
@@ -125,7 +120,7 @@ bool CldapenvironmentEx::changePermissions(const char * ou, const char * userFQD
     action.m_action = "update";
     action.m_basedn = baseDN;
     action.m_rname =  rName;
-    action.m_rtype = RT_DEFAULT;
+    action.m_rtype = RT_FILE_SCOPE;
```
afishbeck, line 123: Is this method always for file scope resource type?

afishbeck (commented):
@RussWhitehead a couple questions.

esp/services/ldapenvironment/ldapenvironmentService.cpp
RussWhitehead: Yes it is reversing previous fix.  As it turns out, the double quotes became part of the password. I was not aware of this because I did a copy/paste of the generated helm. But when I  had to manually type in the password if became painfully obvious. Also obvious when dumping the password with
kubectl get secret adminLDAP-INSURANCE -o jsonpath="{.data.password}" | base64 --decode

esp/services/ldapenvironment/ldapenvironmentService.cpp
RussWhitehead: Yes, and this method is only called from one place (to set perms for the LDAP Admin user). Previous to this fix, because of the RT_DEFAULT type, the code was building the fully qualified name with the incorrect prefix (CN= instead of OU=), so the call to change perms failed in AD

esp/services/ldapenvironment/ldapenvironmentService.cpp
afishbeck: Ok, does the other JIRA need to be reopened to provide a different fix?

afishbeck (approved):

RussWhitehead:
@richardkchapman Please merge

esp/services/ldapenvironment/ldapenvironmentService.cpp
RussWhitehead: @afishbeck I would think not, since this new service is not in use yet

PR #16008, opened 2022-04-15
HPCC-25085 Exclude Indexes from file list

kenrowland:
Added option to invert search results to retrieve only
non index file list

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Cloud-compatibility
  - [x] Premature optimization

## Smoketest:

### DIFF
--- a/esp/services/ws_dfu/ws_dfuService.cpp
+++ b/esp/services/ws_dfu/ws_dfuService.cpp
@@ -3612,7 +3612,15 @@ void CWsDfuEx::setDFUQueryFilters(IEspDFUQueryRequest& req, StringBuffer& filter
     setFileNameFilter(req.getLogicalName(), req.getPrefix(), filterBuf);
     setFileTypeFilter(req.getFileType(), filterBuf);
     WsDFUHelpers::appendDFUQueryFilter(getDFUQFilterFieldName(DFUQFFattrowner), DFUQFTwildcardMatch, req.getOwner(), filterBuf);
-    WsDFUHelpers::appendDFUQueryFilter(getDFUQFilterFieldName(DFUQFFkind), DFUQFTwildcardMatch, req.getContentType(), filterBuf);
+
+    if (req.getInvertContent_isNull() || !req.getInvertContent())
+    {
+        WsDFUHelpers::appendDFUQueryFilter(getDFUQFilterFieldName(DFUQFFkind), DFUQFTwildcardMatch, req.getContentType(), filterBuf);
+    }
+    else
+    {
+        WsDFUHelpers::appendDFUQueryFilter(getDFUQFilterFieldName(DFUQFFkind), DFUQFTinverseWildcardMatch, req.getContentType(), filterBuf);
+    }
     WsDFUHelpers::appendDFUQueryFilter(getDFUQFilterFieldName(DFUQFFgroup), DFUQFTcontainString, req.getNodeGroup(), ",", filterBuf);
     if (!req.getIncludeSuperOwner_isNull() && req.getIncludeSuperOwner())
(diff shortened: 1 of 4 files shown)


wangkx (approved):
Looks good to me.

kenrowland:
@richardkchapman Please merge

esp/services/ws_dfu/ws_dfuService.cpp
```diff
@@ -3612,7 +3612,15 @@ void CWsDfuEx::setDFUQueryFilters(IEspDFUQueryRequest& req, StringBuffer& filter
     setFileNameFilter(req.getLogicalName(), req.getPrefix(), filterBuf);
     setFileTypeFilter(req.getFileType(), filterBuf);
     WsDFUHelpers::appendDFUQueryFilter(getDFUQFilterFieldName(DFUQFFattrowner), DFUQFTwildcardMatch, req.getOwner(), filterBuf);
-    WsDFUHelpers::appendDFUQueryFilter(getDFUQFilterFieldName(DFUQFFkind), DFUQFTwildcardMatch, req.getContentType(), filterBuf);
+
+    if (req.getInvertContent_isNull() || !req.getInvertContent())
+    {
+        WsDFUHelpers::appendDFUQueryFilter(getDFUQFilterFieldName(DFUQFFkind), DFUQFTwildcardMatch, req.getContentType(), filterBuf);
```
jakesmith, line 3618: question/not new: if ContentType is not set - why is building/adding it as a filter ?

jakesmith:
Not convinced this feature should be implemented this way - i.e. by request having to set ContentType="key" and InvertContent=true.

(from JIRA description) - "Currently the filter for a file list allows you to include only indexes, but I can't see a way to exclude indexes."

I would have expected this to be implemented by being able to select the file type via a drop-down and therefore choose e.g. "flat" files only. NB: it may be that some files are of unknown type if sprayed.

@ghalliday - may want to comment on this.

esp/services/ws_dfu/ws_dfuService.cpp
kenrowland: @jakesmith the current logic was not changed except for honoring the invert content flag. The appendDFUQueryFilter method simply returns if there is no filter to add.  IMO if that logic should be changed, a separate Jira should be opened.

jakesmith: >The appendDFUQueryFilter method simply returns if there is no filter to add. 

ok, that's fine.

jakesmith:
> The UI sets the ContentType. It can expand to use a dropdown (probably a separate Jira). The new InvertContent flag returns files that don't match the ContentType value. So, to get non-index files, set ContentType to 'key' and the InvertContent flag to true.
> 
> Think of setting the InvertContent flag as making the compare a !=.


I understood the logic/implementation. I was questioning whether it was the right implementation.
I think it is a bit overcomplicated, and only allows the inverse of 1 specific field.

There may be something I'm missing, but why does the DFUQueryRequest need extending?
i.e. why isn't ContentType sufficient already? 
NB: I know the UI only has an Index checkbox, but the ecm request definition looks correct,
it's the UI that is limiting functionality as far as I can see.

kenrowland:
@ghalliday - Jake asked if you were interested in commenting.

jakesmith (approved):
@kenrowland - approved. A rather specific approach (inverse of 1 field type only), rather than a general mechanism, or a way to specifically select the types you want to see, but it provides the functionality requested, can be revisited at another time if we want to generalize/improve.

kenrowland:
@richardkchapman Please merge (not critical)

PR #16033, opened 2022-04-22
HPCC-27554 Store read/write stats for spray (updated)

shamser:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dali/ft/filecopy.cpp
+++ b/dali/ft/filecopy.cpp
@@ -394,6 +394,9 @@ bool FileTransferThread::performTransfer()
 
         msg.append(sprayer.fileUmask);
 
+        ForEachItemIn(i3, progress)
+            progress.item(i3).serializeExtra(msg, 2);
+
         if (!catchWriteBuffer(socket, msg))
             throwError1(RFSERR_TimeoutWaitConnect, url.str());
 
@@ -412,10 +415,10 @@ bool FileTransferThread::performTransfer()
             OutputProgress newProgress;
             newProgress.deserializeCore(msg);
             newProgress.deserializeExtra(msg, 1);
+            newProgress.deserializeExtra(msg, 2);
             sprayer.updateProgress(newProgress);
 
             LOG(MCdebugProgress(10000), job, "Update %s: %d %" I64F "d->%" I64F "d", url.str(), newProgress.whichPartition, newProgress.inputLength, newProgress.outputLength);
-
             if (isAborting())
             {
                 if (!sendRemoteAbort(socket))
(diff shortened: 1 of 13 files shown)


dali/ft/filecopy.cpp
```diff
@@ -412,10 +415,10 @@ bool FileTransferThread::performTransfer()
             OutputProgress newProgress;
             newProgress.deserializeCore(msg);
             newProgress.deserializeExtra(msg, 1);
+            newProgress.deserializeExtra(msg, 2);
```
jakesmith, line 418: question: isn't this now unconditionally trying to deserialize the extra info (which won't be there if version 1) ?

https://github.com/hpcc-systems/HPCC-Platform/pull/15979#discussion_r855017663
conclusion from that thread in previous PR I thought was that must always check for remaining (>bytesRequired) to handle versions < 2.
If so, I'd expect this to be a single call with current version:
```
    newProgress.deserializeExtra(msg, currentProgressVersion);
```
and for deserializeExtra to unconditionally deserialize for v1, and check remaining>v2required to deserialize v2 meta, if above v2, then it knows it's forced compatibility by all version >=2 sending version down and forcing min compact. to be sent back.

shamser: Yes, Gavin pointed out that unconditionally deserializing v2 would work fine.  The reason is that if the 'done' is read in (line 411) with a value true, it breaks out of the loop.  And that means that when 'ok' is read (line 428),  will not read the v2 message.  Basically, there is no messages after the previous v1 message, so adding v2 message after v1 should work fine.

jakesmith (approved):
@shamser - I think it's good. Please squash

dali/ft/fttransform.ipp
```diff
@@ -192,6 +194,7 @@ public:
     virtual bool setPartition(RemoteFilename & remoteInputName, offset_t _startOffset, offset_t _length, bool compressedInput, const char *decryptKey);
     virtual void setInputCRC(crc32_t value);
     virtual offset_t tell();
+    virtual stat_type getStatistic(StatisticKind kind) override { UNIMPLEMENTED; }
```
ghalliday, line 197: more: This does not look right.  For instance, if you are spraying a file and converting from utf16 to utf8 this will cause the spray to fail.  Should either be ignored, or call processor->getStatistic

ghalliday (commented):
@shamser one issue - generally looks good.

ghalliday (approved):
Good enough.... I don't think this code is used very often.

ghalliday:
@shamser please squash (since Jake approved the previous set of changes)

ghalliday (approved):

jakesmith (approved):
Good to merge.

shamser:
@ghalliday Please can you merge

PR #16041, opened 2022-04-25
HPCC-26876 Add LogAccess Compound Filter support

rpastrana:
- Adds support for binary AND/OR operations
- Changes Elastic Search query to query_string
- Adds support for user defined column based filter

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/scm/ws_logaccess.ecm
+++ b/esp/scm/ws_logaccess.ecm
@@ -23,7 +23,8 @@ ESPenum LogAccessType : int
     ByLogType(3, "ByLogType"),
     ByTargetAudience(4, "ByTargetAudience"),
     BySourceInstance(5, "BySourceInstance"),
-    BySourceNode(6, "BySourceNode")
+    BySourceNode(6, "BySourceNode"),
+    ByFieldName(7, "ByFieldName")
 };
 
 ESPenum LogAccessLogFormat : int
@@ -56,45 +57,78 @@ ESPResponse GetLogAccessInfoResponse
     string RemoteLogManagerType;
     string RemoteLogManagerConnectionString;
 };
+
+ESPenum LogAccessFilterOperator : int
+{
+    NONE(0, "NONE"),
+    AND(1, "AND"),
+    OR(2, "OR")
+};
+
 /*
 * Provides mechanism to query log entries
 *
-* Caller can query by JobId, component, log event type, or target audience by providing the appropriate
+* Caller can query by JobId, component, log event type, target audience, source instance, or host by providing the appropriate
 * enumerated value in the LogCategory field, as well as the targeted value in the SearchByValue field.
+* Caller can also query by field name directly by choosing the 'ByFieldName' option and providing the field name, and desired filter value.
 *
 * SearchbyValue is optional if LogCategory == ALL
 *
+* SearchbyValue contains value used to identify target log entries.
+* Limited to values associated with the the LogCategory choice.
+*
+* If searching by "ByJobIdID", the SearchByValue should contain the jobid of interest
+* If searching by "ByComponent", the SearchByValue should contain the component of interest
+* If searching by "ByLogType", the SearchByValue should contain the 3 letter code associated with the log type of interest.
+* valid values at time of writing are:
+*    DIS - Disaster
+*    ERR - Error
+*    WRN - Warning
+*    INF - Information
+*    PRO - Progress
+*    MET - Metric
+*
+*If searching by "ByTargetAudience", the SearchByValue should contain the 3 letter code associated with the target audience of interest.
+* valid values at time of writing are:
+*    OPR - Operator
+*    USR - User
+*    PRO - Programmer
+*    ADT - Audit
+*If searching by "BySourceInstance", the SearchByValue should contain the instance of interest
+*If searching by "BySourceNode", the SearchByValue should contain the node of interest
+*If searching by "ByFieldName", the SearchByValue should contain the field value of interest, and the SearchField 
+* should be populated.
 * Caller should restrict the query to target a specific time range specified in the Range field.
-* By default, the first 100 log entries encountered are returned. Caller to pagenize using the
+* By default, the first 100 log entries encountered are returned. Caller to paginate using the
 * LogLineStartFrom field (specifies as zero-indexed start index) and the LogLineLimit (specifies the maximum
 * number of log entries to be returned)
 * Caller can specify which log columns should be reported via the Columns field, all available columns returned by default.
 *
 * The report format can be specified via the Format field: JSON|XML|CSV
 */
-ESPRequest GetLogsRequest
+ESPStruct LogFilter
 {
     ESPenum LogAccessType LogCategory;
-    string SearchByValue; //Value used to identify target log entries.
-                          //Limited to values associated with the the LogCategory choice.
-                          //
-                          //If searching by "ByJobIdID", the SearchByValue should contain the jobid of interest
-                          //If searching by "ByComponent", the SearchByValue should contain the component of interest
-                          //If searching by "ByLogType", the SearchByValue should contain the 3 letter code associated with the log type of interest.
-                          // valid values at time of writing are:
-                          //    DIS - Disaster
-                          //    ERR - Error
-                          //    WRN - Warning
-                          //    INF - Information
-                          //    PRO - Progress
-                          //    MET - Metric
-                          //
-                          //If searching by "ByTargetAudience", the SearchByValue should contain the 3 letter code associated with the target audience of interest.
-                          // valid values at time of writing are:
-                          //    OPR - Operator
-                          //    USR - User
-                          //    PRO - Programmer
-                          //    ADT - Audit
+    string SearchByValue; 
+    string SearchField;
+};
+
+ESPStruct BinaryLogFilter
+{
+    ESPStruct LogFilter leftFilter;
+    ESParray<ESPstruct BinaryLogFilter> leftBinaryFilter;
+
+    ESPenum LogAccessFilterOperator Operator;
+
+    ESPStruct LogFilter rightFilter;
+    ESParray<ESPstruct BinaryLogFilter> rightBinaryFilter;
+};
+
+ESPRequest GetLogsRequest
+{
+    [min_ver("1.02")] ESPStruct BinaryLogFilter Filter;
+    [depr_ver("1.02")] ESPenum LogAccessType LogCategory;
+    [depr_ver("1.02")] string SearchByValue; //Value used to identify target log entries.
     ESPStruct TimeRange Range;
     unsigned LogLineLimit(100);
     int64 LogLineStartFrom(0);
@@ -108,7 +142,7 @@ ESPResponse GetLogsResponse
     string LogLines;
 };
 
-ESPservice [auth_feature("WsLogAccess:READ"), version("1.01"), default_client_version("1.01"), exceptions_inline("xslt/exceptions.xslt")] ws_logaccess
+ESPservice [auth_feature("WsLogAccess:READ"), version("1.02"), default_client_version("1.02"), exceptions_inline("xslt/exceptions.xslt")] ws_logaccess
 {
     ESPmethod GetLogAccessInfo(GetLogAccessInfoRequest, GetLogAccessInfoResponse);
     ESPmethod GetLogs(GetLogsRequest, GetLogsResponse);
--- a/esp/services/ws_logaccess/WsLogAccessService.cpp
+++ b/esp/services/ws_logaccess/WsLogAccessService.cpp
@@ -52,36 +52,42 @@ LogAccessTimeRange requestedRangeToLARange(IConstTimeRange & reqrange)
     return range;
 }
 
-bool Cws_logaccessEx::onGetLogs(IEspContext &context, IEspGetLogsRequest &req, IEspGetLogsResponse & resp)
+LogAccessFilterType cLogAccessFilterOperator2LogAccessFilterType(CLogAccessFilterOperator cLogAccessFilterOperator)
 {
-    if (!m_remoteLogAccessor)
-        throw makeStringException(-1, "WsLogAccess: Remote Log Access plug-in not available!");
+    switch (cLogAccessFilterOperator)
+    {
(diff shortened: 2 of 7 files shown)


rpastrana:
@afishbeck please take a look. As mentioned offline, this is meant to add nested/compound filter support to wslogaccess.
The current structure doesn't seem right (it should be binary tree-like). 
If you look at the service code, start at Cws_logaccessEx::onGetLogs. thanks.

esp/scm/ws_logaccess.ecm
```diff
@@ -56,24 +57,18 @@ ESPResponse GetLogAccessInfoResponse
...
-* The report format can be specified via the Format field: JSON|XML|CSV
-*/
-ESPRequest GetLogsRequest
+
+ESPenum LogAccessFilterOperator : int
 {
+    NONE(0, "NONE"),
+    AND(1, "AND"),
+    OR(2, "OR"),
+    NOT(3, "NOT")
```
afishbeck, line 66: Does NOT always imply "AND NOT"?  "OR NOT" would be less common but could occasionally be useful.  You could consider a separate operator / boolean for NOT, or explicit "AND NOT", "OR NOT".

rpastrana: no, it's only there for completness. Not currently supported anyway, but not meant to imply binary operation either.

esp/scm/ws_logaccess.ecm
```diff
@@ -95,6 +90,39 @@ ESPRequest GetLogsRequest
...
+* Provides mechanism to query log entries
+*
+* Caller can query by JobId, component, log event type, target audience, source instance, or host by providing the appropriate
+* enumerated value in the LogCategory field, as well as the targeted value in the SearchByValue field.
+* Caller can also query by field name directly by choosing the 'ByFieldName' option and providing the field name, and desired filter value.
+*
+* SearchbyValue is optional if LogCategory == ALL
+*
+* Caller should restrict the query to target a specific time range specified in the Range field.
+* By default, the first 100 log entries encountered are returned. Caller to pagenize using the
```
afishbeck, line 116: I think it would be "paginate".

esp/services/ws_logaccess/WsLogAccessService.cpp
```diff
@@ -52,60 +52,163 @@ LogAccessTimeRange requestedRangeToLARange(IConstTimeRange & reqrange)
...
+        }
+    }
+}
+
+ILogAccessFilter * logAccessFilterFromPreviousAndClause(ILogAccessFilter* previousAccessfilterFromClause, ILogAccessFilter* previousAccessfilterFromOperation, CLogAccessFilterOperator clauseOperator)
+{
+    if (previousAccessfilterFromClause == nullptr )
+        return previousAccessfilterFromOperation;
+
+    if (previousAccessfilterFromClause != nullptr && clauseOperator == CLogAccessFilterOperator_NONE)
```
afishbeck, line 151: At this point we already know previousAccessfilterFromClause != nullptr.

afishbeck:
@rpastrana I think the basic functionality looks like it works (although the function names get a bit confusing for me).  But it also depends on the level of structure you want to support in your tree.

I guess you currently support the root and one level of depth:

((A and B or C and D) and (E or F) or (X and Z))

Which I think the way you build the tree logically ends up:
((((A and B) or C) and D) and (E or F) or (X and Z))

If you want to have more control over additional nesting you would have to:
1. Use truly recursive nesting:  I think that's broken for ECM ESDL and would need fixing (structures would need to be self referencing).  We could add a new ESDL syntax like "ArrayOfSelf" or similar that we new meant we had to make sure everything was predeclared, etc.
2. Hard code in some more layers of depth.  Say 4 or 5 levels, where you currently have 1 level of nesting.
3. keep it flat but let the user specify where the parenthesis go.  You'd basically need "OPEN" and "CLOSE" operations to build your levels.
4. Use a separate syntax for the logic rather than defining it directly in the ECM.

jakesmith:
@rpastrana - typo' in commit/PR message (CompondFilter -> CompoundFilter)

rpastrana:
Thanks @afishbeck.
> I guess you currently support the root and one level of depth

Right, that's the problem I wanted your help with. The intent was to expose a binary tree structure, but the ecm structure I came up with didn't quite work.

> Use truly recursive nesting: I think that's broken for ECM ESDL and would need fixing (structures would need to be self referencing). We could add a new ESDL syntax like "ArrayOfSelf" or similar that we new meant we had to make sure everything was predeclared, etc.

Truly recursive nesting isn't necessary for this use-case, I only attempted to support it because I assumed we could express the structure in ECM trivially, no need to extend ECM syntax from my point of view.

esp/scm/ws_logaccess.ecm
```diff
@@ -95,6 +90,39 @@ ESPRequest GetLogsRequest
...
+* LogLineStartFrom field (specifies as zero-indexed start index) and the LogLineLimit (specifies the maximum
+* number of log entries to be returned)
+* Caller can specify which log columns should be reported via the Columns field, all available columns returned by default.
+*
+* The report format can be specified via the Format field: JSON|XML|CSV
+*/
+ESPRequest GetLogsRequest
+{
+    ESParray<ESPStruct LogFilterClause, LogFilter> LogFilters;
```
GordonSmith, line 125: Should there be version bumps for this change to the API?

esp/scm/ws_logaccess.ecm
```diff
@@ -95,6 +90,39 @@ ESPRequest GetLogsRequest
...
     ESPStruct TimeRange Range;
     unsigned LogLineLimit(100);
     int64 LogLineStartFrom(0);
```
GordonSmith, line 128: @jeclrsg - Any comments on these changes, compared to your existing implementation?

esp/scm/ws_logaccess.ecm
rpastrana: Definitely, I was hoping nobody had started work based on the earlier versions of this interface

esp/scm/ws_logaccess.ecm
rpastrana: @jeclrsg the intent here is to provide support for compound queries such as WUDI == W2022XYZ AND (logtype == ERR OR logtype == WRN). 
Fair warning this interface isn't final yet, but it's a good time to provide feedback

rpastrana:
Sample req for filter = wuid == w2022XYZ AND ( logtype == ERR OR logtype == WRN):

      <urn:GetLogsRequest>
         <urn:Filter>
            <urn:leftFilter>
               <urn:LogCategory>1</urn:LogCategory>
               <urn:SearchByValue>W20220427-200649</urn:SearchByValue>
            </urn:leftFilter>
            <urn:Operator>1</urn:Operator>
            <urn:rightBinaryFilter>
               <urn:BinaryLogFilter>
               <urn:leftFilter>
               <urn:LogCategory>3</urn:LogCategory>
               <urn:SearchByValue>err</urn:SearchByValue>
            </urn:leftFilter>
            	<urn:Operator>2</urn:Operator>
			<urn:rightFilter>
               <urn:LogCategory>3</urn:LogCategory>
               <urn:SearchByValue>wrn</urn:SearchByValue>
            </urn:rightFilter>
               </urn:BinaryLogFilter>
            </urn:rightBinaryFilter>
         </urn:Filter>
         <urn:Range>
            <urn:StartDate>2022-03-02T01:26:27</urn:StartDate>
         </urn:Range>
         <urn:LogLineLimit>10</urn:LogLineLimit>
         <urn:LogLineStartFrom>0</urn:LogLineStartFrom>
         <urn:SelectColumnMode>1</urn:SelectColumnMode>
         <urn:Format>XML</urn:Format>
      </urn:GetLogsRequest>

esp/services/ws_logaccess/WsLogAccessService.cpp
```diff
@@ -52,63 +52,145 @@ LogAccessTimeRange requestedRangeToLARange(IConstTimeRange & reqrange)
...
+    else
+        return nullptr;
+}
+
+ILogAccessFilter * processBinaryLogFilter(IConstBinaryLogFilter * binaryfilter)
+{
+    if (!binaryfilter)
+        return nullptr;
+    
+    ILogAccessFilter * leftFilter = nullptr;
```
afishbeck, line 132: Should you give an error if binaryfilter->getLeftFilter() is filled in and binaryfilter->getLeftBinaryFilter().ordinality() != 0?

Since you don't support both together, checking would prevent user mistakes and confusion, and in some scenarios would let you change the behavior in the future without worrying about breaking existing users.

afishbeck: If so, same would apply to the right hand side.

esp/services/ws_logaccess/WsLogAccessService.cpp
```diff
@@ -52,63 +52,145 @@ LogAccessTimeRange requestedRangeToLARange(IConstTimeRange & reqrange)
...
+        }
+        else
+        {
+            rightFilter = processBinaryLogFilter(&binaryfilter->getRightBinaryFilter().item(0));
+            if (binaryfilter->getRightBinaryFilter().ordinality() > 1)
+            {
+                WARNLOG("TOO MANY right FILTERS ENCOUNTERED");
+            }
+        }
+        return getBinaryLogAccessFilterOwn(leftFilter, rightFilter, cLogAccessFilterOperator2LogAccessFilterType(binaryfilter->getOperator()));
```
afishbeck, line 172: I would have named this function something else besides "get".  More like "createBinaryLogAccessFilterOwn()".  But I see there are quite a few of these functions named similarly.

esp/services/ws_logaccess/WsLogAccessService.cpp
```diff
@@ -52,63 +52,145 @@ LogAccessTimeRange requestedRangeToLARange(IConstTimeRange & reqrange)
...
+
+ILogAccessFilter * processLogFilter(IConstLogFilter * logFilter)
+{
+    if (logFilter)
+        return processLogFilterByFields(logFilter->getLogCategory(), logFilter->getSearchByValue(), logFilter->getSearchField());
+    else
+        return nullptr;
+}
+
+ILogAccessFilter * processBinaryLogFilter(IConstBinaryLogFilter * binaryfilter)
```
afishbeck, line 127: You could argue that "processing" happens inside 

`m_remoteLogAccessor->fetchLog(logFetchOptions, logcontent, logAccessFormatFromName(req.getFormat()));
`
And that the name of this function would be 
buildLogAccessFilter(IConstBinaryLogFilter * binaryfilter);
and above
`buildLogAccessFilter(IConstLogFilter * logFilter);`
Named after what they were building rather than the input.

The main point being that it becomes intuitive that this is the preparation stage, and processing happens further in.

esp/services/ws_logaccess/WsLogAccessService.cpp
rpastrana: Yep, good point.

esp/services/ws_logaccess/WsLogAccessService.cpp
rpastrana: will leave pre-existing function name as is for now.
I'm looking into an automated pre-code review scan for all PRs where we can enforce a lot of our unwritten styles/protocols (not sure if this one could be articulated).

esp/services/ws_logaccess/WsLogAccessService.cpp
afishbeck: I think L146 tells you if no left filter was provided.  I meant the case where you have to many left filters because the nested one was filled in, and the array also had an entry.  Your functionality works and ignores the extra infor, but this would really be a check to make sure the caller knows what they are doing so if you make changes in the future the client won't break.

rpastrana:
@afishbeck addressed the bigger issues you pointed out. Please re-review.

afishbeck (approved):

esp/services/ws_logaccess/WsLogAccessService.cpp
rpastrana: Test case for scenario exposed by Tony, explicit error message now returned to user when left/rightFilter and left/rightBinaryFilter both submitted:

<urn:GetLogsRequest>
         <urn:Filter>
            <urn:leftFilter>
               <urn:LogCategory>1</urn:LogCategory>
               <urn:SearchByValue>W20220504-173939</urn:SearchByValue>
            </urn:leftFilter>
           
            <urn:Operator>1</urn:Operator>
                        <urn:rightFilter>
               <urn:LogCategory>1</urn:LogCategory>
               <urn:SearchByValue>W20220504-173939</urn:SearchByValue>
            </urn:rightFilter>
            <urn:rightBinaryFilter>
               <urn:BinaryLogFilter>
                  <urn:leftFilter>
                     <urn:LogCategory>3</urn:LogCategory>
                     <urn:SearchByValue>err</urn:SearchByValue>
                  </urn:leftFilter>
                  <urn:Operator>2</urn:Operator>
                  <urn:rightFilter>
                     <urn:LogCategory>3</urn:LogCategory>
                     <urn:SearchByValue>wrn</urn:SearchByValue>
                  </urn:rightFilter>
               </urn:BinaryLogFilter>
            </urn:rightBinaryFilter>
         </urn:Filter>
         <urn:Range>
            <urn:StartDate>2022-03-02T01:26:27</urn:StartDate>
         </urn:Range>
         <urn:LogLineLimit>10</urn:LogLineLimit>
         <urn:LogLineStartFrom>0</urn:LogLineStartFrom>
         <urn:SelectColumnMode>1</urn:SelectColumnMode>
         <urn:Format>XML</urn:Format>
      </urn:GetLogsRequest>

esp/services/ws_logaccess/WsLogAccessService.cpp
rpastrana: @GordonSmith / @jeclrsg any comments before we finalize this?

GordonSmith (approved):

PR #16048, opened 2022-04-27
HPCC-27576 Create a blog summarising the storage plane support

ghalliday:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/docs/blogs/FileStorage.md
+++ b/docs/blogs/FileStorage.md
@@ -9,7 +9,7 @@ For those who are used to the bare-metal version of the platform, one of the mai
 Whenever the HPCC system needs to access storage, it uses a 'storage plane' to provide all the information needed to use that storage.  These storage planes are defined within the storage section of the values.yaml file.  Many of the options for a storage plane are optional and will be covered later, but first what are the basics?  There are two main options for storage:
 
 ## Ephemeral Storage
-Ephemeral storage is storage that has the same lifetime as the instance of the HPCC cluster.  When the cluster is installed the storage is created, when the cluster is uninstalled the storage is destroyed.  This is very useful for testing, or experimenting with a system, but not so helpful for a production system!  This type of storage is used in the default values.yaml file to make it easy to deploy the chart and experiment.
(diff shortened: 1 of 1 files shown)


docs/blogs/FileStorage.md
```diff
@@ -0,0 +1,305 @@
...
+      pvc: data-cache-pvc
+      prefix: "/var/lib/HPCCSystems/hpcc-data-cache"
+```
+The alias defines an alternative pvc which is used to access the data, and an alternative path to acccess the data via.  It also contains a list of modes which describe the situations where that access method should be used.   In this case if the file access requires random access (e.g. roxie index operations or thor keyed-join), then the cache should be used.  Otherwise for sequential reading and writing the default definition will be used.
+
+Currently "random" is the only mode that the system treats differently, but it is likely others will be supported in the future.
+
+## Bare metal and Kubernetes co-existing
+
+There have been some situations where a bare-metal roxie cluster has been used to synchronize data files, with the intent of a K8s cluster reusing them it is created (because the files are already copied to the correct place).  Needless to say this is not recommended...
```
jakesmith, line 289: >with the intent of a K8s cluster reusing them it is created

with the intent of a K8s cluster reusing them **if** it is created ?

docs/blogs/FileStorage.md
```diff
@@ -0,0 +1,305 @@
...
+
+The problem with this approach is that by default the bare-metal system puts the files in a roxie subdirectory within the storage location (hpcc-data/roxie), and the k8s system does not (hcc-data).  This means that when the k8s cluster checks to see if the file exists the expected path does not match, the file does not exist so it copies the file again.  This can be worked-around by setting the subPath property of the associated storage plane to "roxie".  Then both systems expect the files in the same place, allowing the k8s cluster to reuse it.
```
jakesmith, line 291: > and the k8s system does not (hcc-data).

and the k8s system does not (the root of its storage plane is hpcc-data/).

jakesmith (commented):
@ghalliday - please see comments. Looks good, mostly typo's and similar.

jakesmith:
@ghalliday - it would highlighting/describing what needs to be setup for mapping a remote plane, if that plane is on numDevices>1, i.e. need same on client 'local' plane, and the same number of pvc's 1-N need to be created.

jakesmith:
@ghalliday - this is probably also the place it should detail how bare-metal system can access DFS via remote (HPCC-27688),
and I think there needs to be some info. re. enabling dafilersrv (direction) in the k8s values configuration

PR #16049, opened 2022-04-27
HPCC-27572 Remove unsupported characters from generated password

RussWhitehead:
Certain password characters are not accepted on the kubectl create secret command
line. This PR removes those characters from the set available to be generated

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [x] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jutil.cpp
+++ b/system/jlib/jutil.cpp
@@ -56,6 +56,8 @@
 
 #include "portlist.h"
 
+#include <random>
+
 static NonReentrantSpinLock * cvtLock;
 
 #ifdef _WIN32
@@ -3349,33 +3351,36 @@ const char * generatePassword(StringBuffer &pwd, int pwdLen)
     const char alphaUC[] = "ABCDEFGHIJKLMNOPQRSTUVWXYZ";
     const char alphaLC[] = "abcdefghijklmnopqrstuvwxyz";
     const char numeric[] = "0123456789";
-    const char symbol[] = "~@#$%^*()_-+={[}]:,.?";
+    const char symbol[] = "~@#%^*_-+{[}]:,.?";
+
+    std::random_device seedGen;//uniformly-distributed integer random number generator that produces non-deterministic random numbers
+    std::mt19937 mtEngine(seedGen());//generates 32-bit pseudo-random numbers using Mersenne twister algorithm
 
     //Ensures each character group used at least once
-    srand(time(0));
-    pwd.append(alphaUC[rand() % (sizeof(alphaUC) - 1)]);
-    pwd.append(alphaLC[rand() % (sizeof(alphaLC) - 1)]);
-    pwd.append(numeric[rand() % (sizeof(numeric) - 1)]);
-    pwd.append(symbol[rand() % (sizeof(symbol) - 1)]);
+    pwd.append(alphaUC[mtEngine() % (sizeof(alphaUC) - 1)]);
+    pwd.append(alphaLC[mtEngine() % (sizeof(alphaLC) - 1)]);
+    pwd.append(numeric[mtEngine() % (sizeof(numeric) - 1)]);
+    pwd.append(symbol[mtEngine() % (sizeof(symbol) - 1)]);
 
     for (int i = 4; i < pwdLen; i++)
     {
-        switch(rand() % NUM_GROUPS)//select a random character group
+        switch(mtEngine() % NUM_GROUPS)//select a random character group
         {
         case 0:
-            pwd.append(alphaUC[rand() % (sizeof(alphaUC) - 1)]);  break;
+            pwd.append(alphaUC[mtEngine() % (sizeof(alphaUC) - 1)]);
+            break;
         case 1:
-            pwd.append(alphaLC[rand() % (sizeof(alphaLC) - 1)]);  break;
+            pwd.append(alphaLC[mtEngine() % (sizeof(alphaLC) - 1)]);
+            break;
         case 2:
-            pwd.append(numeric[rand() % (sizeof(numeric) - 1)]);  break;
+            pwd.append(numeric[mtEngine() % (sizeof(numeric) - 1)]);
+            break;
         case 3:
-            pwd.append(symbol[rand() % (sizeof(symbol) - 1)]);  break;
+            pwd.append(symbol[mtEngine() % (sizeof(symbol) - 1)]);
+            break;
         }
     }
-#ifdef _DEBUG
-    if (pwd.length() != (size32_t)pwdLen)
-        throw makeStringException(0, "Generated Passwords wrong length!");
-#endif
+
     return pwd.str();
 }
 


RussWhitehead:
@afishbeck Please review

afishbeck (approved):

RussWhitehead:
@richardkchapman Please merge

system/jlib/jutil.cpp
```diff
@@ -3349,7 +3349,7 @@ const char * generatePassword(StringBuffer &pwd, int pwdLen)
     const char alphaUC[] = "ABCDEFGHIJKLMNOPQRSTUVWXYZ";
     const char alphaLC[] = "abcdefghijklmnopqrstuvwxyz";
     const char numeric[] = "0123456789";
-    const char symbol[] = "~@#$%^*()_-+={[}]:,.?";
+    const char symbol[] = "~@#%^*_-+{[}]:,.?";
 
     //Ensures each character group used at least once
     srand(time(0));
```
richardkchapman, line 3355: Not new, but this doesn't feel like a secure way to generate passwords to me... If you know the approximate date that an account was created you can brute force through the passwords that would have been generated on that date very easily...

richardkchapman (changes requested):

RussWhitehead:
@richardkchapman Please rereview

system/jlib/jutil.cpp
```diff
@@ -3349,10 +3349,10 @@ const char * generatePassword(StringBuffer &pwd, int pwdLen)
     const char alphaUC[] = "ABCDEFGHIJKLMNOPQRSTUVWXYZ";
     const char alphaLC[] = "abcdefghijklmnopqrstuvwxyz";
     const char numeric[] = "0123456789";
-    const char symbol[] = "~@#$%^*()_-+={[}]:,.?";
+    const char symbol[] = "~@#%^*_-+{[}]:,.?";
 
     //Ensures each character group used at least once
-    srand(time(0));
+    srand(unsigned(GetCurrentProcessId() + GetCurrentThreadId() + time(0)));
```
afishbeck, line 3355: @RussWhitehead You could use urandom which uses system device entropy to generate random numbers.

https://linux.die.net/man/4/urandom

We use it for the global transaction ids here https://github.com/hpcc-systems/HPCC-Platform/blob/870834b08a0550db458f06d0ef3dfe062c018e2c/system/globalid/lnuid.cpp#L91

But there are probably a few ways to call it.

RussWhitehead: @afishbeck Wouldn't spawning a process to do this be extremely slow and somewhat  overkill?
Will
GetCurrentProcessId() + GetCurrentThreadId() + time(0)
will be both fast and achieve the randomness needed?

afishbeck: @RussWhitehead Not sure what you mean about spawning a process.  Our createUniqueId() function doesn't spawn anything.
/dev/random (better for strong cryptography) can block which is why we use /dev/urandom which doesn't.

RussWhitehead: My bad, I thought it was spawned. Still, it's file IO which seems very expensive?

RussWhitehead:
@afishbeck Added a new JLib function (generateRandom) to initialize a given buffer with random values from /dev/urandom , and modified generatePassword to use that function to seed the random number generator. Please review

system/jlib/jutil.cpp
```diff
@@ -3349,10 +3367,13 @@ const char * generatePassword(StringBuffer &pwd, int pwdLen)
     const char alphaUC[] = "ABCDEFGHIJKLMNOPQRSTUVWXYZ";
     const char alphaLC[] = "abcdefghijklmnopqrstuvwxyz";
     const char numeric[] = "0123456789";
-    const char symbol[] = "~@#$%^*()_-+={[}]:,.?";
+    const char symbol[] = "~@#%^*_-+{[}]:,.?";
+
+    unsigned randomNumber;
+    generateRandom((const void *)&randomNumber, sizeof(randomNumber));
+    srand(randomNumber);
```
ghalliday, line 3374: I think this probably needs a bit more discussion (possibly including the code in lnuid).  A couple of items to discuss:

* windows needs to be supported.
* should /dev/urandom only be used for seeding a random number generator (e.g., create an instance of std::mt19937 and seed it with values read from /dev/urandom. (see quote below:)
* rand isn't thread safe

"While some safety margin above that minimum is reasonable, as a guard against flaws in the CPRNG algorithm, no cryptographic primitive available today can hope to promise more than 256 bits of security, so if any program reads more than 256 bits (32 bytes) from the kernel random pool per invocation, or per reasonable reseed interval (not less than one minute), that should be taken as a sign that its cryptography is not skillfully implemented"

RussWhitehead:
@ghalliday @afishbeck The Jira and this PR was meant to remove unsupported characters from the generated password. Should it be merged as it was, and a separate Jira be used to address random numbers and such?

system/jlib/jutil.cpp
afishbeck: There are two mechanisms "/dev/random" and "/dev/urandom".

urandom actually does close to what you suggest already.  It uses the entropy to seed a "cryptographically secure" pseudorandom number generator (CSPRNG).  It then reseeds at a certain rate.  I think the default is 60 sec.  I think the quote above is talking about limits in trusting certain key sizes to be truly random.  But I've seen arguments declaring that urandom is pretty secure.

Using mt19937 would mostly be switching out the repeated use of the internal urandom CSPRNG.  Delivering similar results although perhaps it is a better algorithm?  The mt19937 algorithm should only repeat every (219937 – 1) times.  Would need to investigate if it were really better than whats used in urandom.

But anyway, in this case using mt19937 (and as Russ is trying to do with rand()) as a seed should be fine, but I don't think it fixes what the quote is talking about above, because it just replaces the CPRNG with another generator but I could be wrong if the algorithm is better.

Whether we use urandom or not I guess Russ should not be using rand if it's not thread safe.  So either using urandom to seed mt19937 or just using urandom should be good enough?

As for lnguid, they (Charles, David and Co.) did some pretty extensive analysis trying to determine how frequently the same id would be produced by their algorithm using urandom, but it wouldn't hurt to think it over again.

afishbeck: Bottom line, replace rand since it's not threadsafe.

Replacement could either be just urandom, or mt19937 seeded with urandom.

RussWhitehead:
@afishbeck @ghalliday I would like to get this invalid password character fix merged, and as a separate issue I have opened
https://track.hpccsystems.com/browse/HPCC-27600
to discuss and address usage of rand() throughout the platform
Please merge

richardkchapman:
Fixing this to seed with something other than time seems well within the scope of this ticket, which is about generating passwords...

Generating predictable passwords is almost worse than generating passwords that are not acceptable!

richardkchapman:
Not on containerized systems it doesn't...

system/jlib/jutil.cpp
afishbeck: https://en.wikipedia.org/wiki/Cryptographically-secure_pseudorandom_number_generator#:~:text=Every-,CSPRNG,-should%20satisfy%20the

richardkchapman (approved):

richardkchapman:
@RussWhitehead Thanks - that looks better. Please squash.

system/jlib/jutil.cpp
```diff
@@ -56,6 +56,9 @@
 
 #include "portlist.h"
 
+using namespace std;
```
ghalliday, line 59: picky: this isn't needed needed since your're fully qualifying the classes below.

ghalliday (approved):

RussWhitehead:
@richardkchapman Squashed
@ghalliday  Removed "using namespace std"
Please merge

afishbeck (approved):

PR #16055, opened 2022-04-28
HPCC-27587 Remove initialiser for thread_local defaultJobInfo

ghalliday:
## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jlog.cpp
+++ b/system/jlib/jlog.cpp
@@ -259,7 +259,7 @@ void LogMsgJobInfo::deserialize(MemoryBuffer & in)
 }
 
 static LogMsgJobInfo globalDefaultJobInfo(UnknownJob, UnknownUser);
-static thread_local LogMsgJobInfo defaultJobInfo = globalDefaultJobInfo;
+static thread_local LogMsgJobInfo defaultJobInfo;
 
 const LogMsgJobInfo unknownJob(UnknownJob, UnknownUser);
 
--- a/system/jlib/jthread.cpp
+++ b/system/jlib/jthread.cpp
@@ -134,7 +134,7 @@ unsigned WINAPI Thread::_threadmain(LPVOID v)
 void *Thread::_threadmain(void *v)
 #endif
 {
-    resetThreadLogging();   // Not strictly needed if everything handled via thread_local variable's constructors...
+    resetThreadLogging();
     Thread * t = (Thread *)v;
 #ifdef _WIN32
     if (SEHHandling) 
@@ -607,6 +607,7 @@ void CThreadedPersistent::threadmain()
             break;
         try
         {
+            resetThreadLogging();
             owner->threadmain();
             // Note we do NOT call the thread reset hook here - these threads are expected to be able to preserve state, I think
         }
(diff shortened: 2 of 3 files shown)


ghalliday:
@richardkchapman pushed for discussion.  Two separate commits.  I think the first is probably worthwhile.  The second definitely needs discussion.

system/jlib/jlog.cpp
```diff
@@ -195,11 +195,10 @@ unsigned getPositionOfField(unsigned logfields, unsigned positionoffield)
 
 // LogMsg
 
-LogMsgJobInfo::~LogMsgJobInfo()
-{
-    if (isDeserialized)
-        free((void *) jobIDStr);
-}
+
+//Save all external job strings in a hash table to avoid a non-trivial destructor for LogMsgJobInfo
```
richardkchapman, line 199: How are destructors for thread-local variables handled? Do they have the same performance penalty issues as non-default constructors?

system/jlib/jthread.cpp
```diff
@@ -607,6 +607,7 @@ void CThreadedPersistent::threadmain()
             break;
         try
         {
+            resetThreadLogging();
```
richardkchapman, line 610: Are you sure this is wanted? I can't remember how the thread-local logging stuff works, but I would think that we want in some situations at least to be inheriting the logging settings from the thread that is creating me.

system/jlib/jlog.cpp
```diff
@@ -252,16 +249,16 @@ void LogMsgJobInfo::deserialize(MemoryBuffer & in)
         // >= 8.0 client
         StringBuffer idStr;
         in.read(idStr);
-        jobIDStr = idStr.detach();
+
+        CriticalBlock block(externalJobCS);
```
richardkchapman, line 253: Nervous about adding this atom table without understanding (a) what it is gaining us - is a non-trivial destructor really that painful? and (b) how often this code gets called and will the criticalblock have a worse impact than the original problem

system/jlib/jlog.cpp
ghalliday: I don't know how they are implemented, there must be a thread hook mechanism that is called on thread destruction.  I don't think there is a big performance penalty.
The reason for the change was more that you couldn't have a constexpr object if the destructor is non-trivial

system/jlib/jthread.cpp
ghalliday: Despite the name resetThreadLogging() actually set the thread job id to match the current one.  This means it will be done when the thread starts, previously if it wasn't called it would be on first-use.

system/jlib/jlog.cpp
ghalliday: I agree - included for discussion

system/jlib/jlog.cpp
richardkchapman: Do we need a constexpr object?

ghalliday: Only to avoid a destructor.

ghalliday:
@richardkchapman my vote would be to take the first commit.  This will reduce the cost of accessing the current job, and ensure it is initialised when the thread is started - but will mean it isn't initialised on threads started by other libraries.  I suspect any extra cost on thread creation or destruction is negligible.

richardkchapman:
@ghalliday I agree - just the first commit seems smartest.

richardkchapman (approved):

PR #16060, opened 2022-05-03
HPCC-27584 Ensure thormanager can be communicated to for linger graphs

jakesmith:
The network policy of the thormanager must allow traffic from the
thor-eclagent pod, so that it can be reused for multiple graphs
(via linger mechanism).

Install a separate network policy for thorworkers.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/helm/hpcc/templates/thor.yaml
+++ b/helm/hpcc/templates/thor.yaml
@@ -38,7 +38,8 @@ Pass in dict with root and me
 {{- $eclAgentJobName := printf "%s-_HPCC_JOBNAME_" $eclAgentType }}
 {{- $thorManagerJobName := printf "thormanager-_HPCC_JOBNAME_" }}
 {{- $thorWorkerJobName := printf "thorworker-_HPCC_JOBNAME_" }}
-{{- $thorNetworkPolicyName := printf "thormanager-networkpolicy-_HPCC_JOBNAME_" }}
+{{- $thorManagerNetworkPolicyName := printf "thormanager-networkpolicy-_HPCC_JOBNAME_" }}
+{{- $thorWorkerNetworkPolicyName := printf "thorworker-networkpolicy-_HPCC_JOBNAME_" }}
 {{- $numWorkersPerPod := .me.numWorkersPerPod | default 1 }}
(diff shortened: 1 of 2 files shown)


jakesmith:

This introduces a separate network policy for workers, and ensures the manager policy ingress applies to thor-eclagent and the job name.

I had initially wanted to roll both network policies into a single config. yaml, to be applied at the same time as existing one is.
However, having 2 inline network policies under thormanager-networkspec.yaml isn't possible because of the way config is represented as dictionaries for manipulation/generation (because multiple network policies have same keys and overwrite each other).

ghalliday (approved):
Looks sensible to me.

ghalliday:
@richardkchapman I am going to merge this and tag an rc, but please review it carefully when you have time to see if you can spot anything.

helm/hpcc/templates/thor.yaml
```diff
@@ -270,24 +271,43 @@ data:
...
     kind: NetworkPolicy
     metadata:
-      name: {{ $thorNetworkPolicyName }}
+      name: {{ $thorManagerNetworkPolicyName }}
     spec:
       podSelector:
         matchLabels:
-          app: thor
-          job: "_HPCC_JOBNAME_"
+          app: "_HPCC_JOBNAME_"
```
richardkchapman, line 278: This may work, but it's confusing. I don't know why you would not call the label that indicates the job name "job" rather than renaming it to "app"

jakesmith: 'or' pod label matching isn't support.
To ensure that the network policy only matches this job, I either had to change/add a consistent label name into the thor manager/worker job pods, or change/add the label in the thor-eclagent pod.

Alternatively, could leave 'job' as it was, and introduce a new 'instance' label or similar in both, which in thormanager/thorworker was same as _HPCC_JOBNAME_

PR #16064, opened 2022-05-04
HPCC-27584 NP selector should be on dynamic thor-eclagent label value

jakesmith:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/helm/hpcc/templates/thor.yaml
+++ b/helm/hpcc/templates/thor.yaml
@@ -144,11 +144,13 @@ data:
       template:
         metadata:
           labels:
-            app: "_HPCC_JOBNAME_"
-            component: "thormanager"
             accessDali: "yes"
             accessEsp: "yes"
+            app: "thor"
+            component: "thormanager"
             helmVersion: 8.6.25-closedown0
+            instance: "_HPCC_JOBNAME_"
+            job: "_HPCC_JOBNAME_"
 {{- if hasKey $thorScope "labels" }}
 {{ toYaml $thorScope.labels | indent 12 }}
 {{- end }}
@@ -210,10 +212,12 @@ data:
       template:
         metadata:
           labels:
-            app: "_HPCC_JOBNAME_"
-            component: "thorworker"
             accessEsp: "true"
+            app: "thor"
+            component: "thorworker"
             helmVersion: 8.6.25-closedown0
+            instance: "_HPCC_JOBNAME_"
+            job: "_HPCC_JOBNAME_"
 {{- if hasKey $thorScope "labels" }}
 {{ toYaml $thorScope.labels | indent 12 }}
 {{- end }}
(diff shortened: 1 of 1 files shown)


jakesmith:
@richardkchapman - noticed that I'd left a static app label in for the thor-eclagent - it should have been based on the constructed name (otherwise all agents from other thor's could connect).

Please review.
I could reinstate the 'job' label in this PR, and introduce another label (e.g. 'instance')..?
But it would need to be based on the same value (_HPCC_JOBNAME_), and would need to add one an 'instance' to thor-eclagent as well.

richardkchapman:
It feels correct to me to have app as it was, job with jobname, and a new label to use for selection of which pods get the thormaster network policy

jakesmith:
@richardkchapman - reverted app and job label to what they were, introduced new instance label, which in Thor pods is same as job, and in agents is same as app, such that the manager NP ingress selector can select on both the thor job and the thor-eclagent.

helm/hpcc/templates/thor.yaml
```diff
@@ -331,16 +340,15 @@ spec:
...
     matchLabels:
-      run: {{ $commonCtx.eclAgentName | quote }}
-      app: thor-eclagent
+      app: {{ $commonCtx.eclAgentName | quote }}
   template:
     metadata:
       labels:
-        run: {{ $commonCtx.eclAgentName | quote }}
-        app: thor-eclagent
+        app: {{ $commonCtx.eclAgentName | quote }}
```
richardkchapman, line 347: I would think app should be a constant

helm/hpcc/templates/thor.yaml
```diff
@@ -392,14 +400,15 @@ spec:
   replicas: {{ $commonCtx.thorAgentReplicas }}
   selector:
     matchLabels:
-      run: {{ $commonCtx.thorAgentName | quote }}
+      app: {{ $commonCtx.thorAgentName | quote }}
```
richardkchapman, line 403: And here

helm/hpcc/templates/thor.yaml
```diff
@@ -392,14 +400,15 @@ spec:
...
   template:
     metadata:
       labels:
-        run: {{ $commonCtx.thorAgentName | quote }}
+        app: {{ $commonCtx.thorAgentName | quote }}
```
richardkchapman, line 407: And here

richardkchapman (changes requested):
Not sure what app is now used for but it feels like it should be constant.

richardkchapman (approved):

ghalliday (approved):
Looks sensible to me

richardkchapman:
@jakesmith Please squash

jakesmith:
@richardkchapman - squashed

PR #16081, opened 2022-05-11
HPCC-27581 Remove ThreadDestroyListLock

ghalliday:
## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jthread.cpp
+++ b/system/jlib/jthread.cpp
@@ -121,10 +121,40 @@ void disableThreadSEH() { SEHHandling=false; } // only prevents new threads from
 
 static std::atomic<unsigned> threadCount;
 static size32_t defaultThreadStackSize=0;
-static ICopyArrayOf<Thread> ThreadDestroyList;
-static CriticalSection ThreadDestroyListLock;
 
+/*
+The following variables are used to avoid a race condition between thread termination and join().  There are two situations:
+i) thread is free running. join will never be called.
+ii) thread will be joined
 
+The code in the thread termination code signals a semaphore to indicate it is complete, and then releases the thread
+object.  This ensures the semaphore has not been destroyed before it is signalled, and that threads that are mot joined
+are cleaned up correctly.
+
+The code in join waits for the semaphore to be signalled, and then continues.  If the thread object is a contained
+member the thread object may be destroyed before the release call in the thread termination code - which would lead to
+memory corruption.
+
+Options to prevent this:
+
+* Added the thread to a list before the signal(), and removed after the Release().
+  The joining thread can wake up immediately, but then needs to loop until the thread is not on the list.  (This
+  involved getting a critical section).
+  [This is the previous implementation.]
+
+* The termination code holds the critical section for the { signal(), Release() } duration.
+  The joining thread checks it can get that critical section before continuing.  The joining thread may wake up, try
+  and get the critical section, and then block because the other thread hasn't released it yet, but this is better than
+  the previous situation which would spin-loop.  The join will be blocked by all terminating threads - in a signal,
+  rather than purely user code.
+  [This is the new implementation]
+
+* Terminating thread conditionally signals.
+  The terminating thread could call Release(), and only signal() the semaphore if the object was not destroyed.  This would cope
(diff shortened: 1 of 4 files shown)


ghalliday:
Pushed to testing and for discussion.

richardkchapman:
The core needs investigating. And I'm not sure this should target a point release

ghalliday:
I agree with both comments.  I was lazy about rebasing before pushing.

ghalliday:
I understand the problem - and possibly why the code was there in the first place.  If nothing else I can now improve the comment. to make it clear.

ghalliday:
Actually, I think it is revealing a potential bug....

system/jlib/jthread.cpp
```diff
@@ -119,10 +119,40 @@ void disableThreadSEH() { SEHHandling=false; } // only prevents new threads from
...
+  The joining thread checks it can get that critical section before continuing.  The joining thread may wake up, try
+  and get the critical section, and then block because the other thread hasn't released it yet, but this is better than
+  the previous situation which would spin-loop.  The join will be blocked by all terminating threads - in a signal,
+  rather than purely user code.
+  [This is the new implementation]
 
+* Terminating thread conditionally signals.
+  The terminating thread could call Release(), and only signal() the semaphore if the object was not destroyed.  This would cope
+  with the common joining pattern, but would fail if another object held onto a reference to the thread but didn't call join().
+  Potentially more efficient, but more scope for accidental bugs.
```
jakesmith, line 153: Existing code was very confusing (double Release) etc, and wasteful list, looks a lot better.

I've eventually convinced myself of the need for the CS to be global, but it's a bit of a pity that still a lock involved in thread handling as they finished up/joined.

If startRelease's usages could be removed, and join() mandated, then we could avoid _threadmain ever having the possibility of destroying the thread (get rid of start linking and _threadmain releasing), and then we wouldn't need the CS.

jakesmith (approved):
@ghalliday - looks good, just some musings re. whether we can get rid of global CS, startRelease and mandate join(), but separate JIRA if so.

system/jlib/jthread.cpp
```diff
@@ -475,22 +492,10 @@ bool Thread::join(unsigned timeout)
...
-        if (st==10)     
-            PROGLOG("Thread::join race");   
-#endif
-        Sleep(st);      // switch back to exiting thread (not very elegant!)
-        st++;
-        if (st>10)
-            st = 10; // note must be non-zero for high priority threads
+
+    {
+        CriticalBlock block(ThreadDestroyLock);
```
mckellyln, line 497: Do we need this CB here ?

mckellyln (commented):
Looks better, more direct and simple.
Just have one question.

system/jlib/jthread.cpp
richardkchapman: Perhaps normal threads should mandate join() before the are destroyed (and can check for same and report if missed), and a different class (that does not have a join method) provided for "fire and forget" threads. Though I'm not really convinced there is really any such thing - at the very least they need to be "joined" at process termination.

system/jlib/jthread.cpp
ghalliday: yes - the comment should probably precede the critical block.  It is preventing the other thread stalling between the signal and the release, and this thread continuing, and disposing of the object that contains the threading object.  (Which does happen...)

system/jlib/jthread.cpp
ghalliday: I think threads should always have to be joined, but implement that as a future separate change.

richardkchapman:
Can you rebase and retarget master please

system/jlib/jthread.cpp
ghalliday: moved the comment, and clarified it.

mckellyln (approved):

ghalliday:
@richardkchapman squashed and rebased/target 8.8.x.  Careful review recommended.

PR #16084, opened 2022-05-11
HPCC-27597 Fix JSON output file with nested records has extra newlines

afishbeck:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/rtl/eclrtl/rtlformat.cpp
+++ b/rtl/eclrtl/rtlformat.cpp
@@ -766,7 +766,9 @@ void CommonJsonWriter::outputBeginNested(const char *fieldname, bool nestChildre
         appendJSONName(out, fieldname);
     }
     out.append("{");
-    if (!nestChildren && !nestLimit)
+    if ((flags & XWFonlyindentroot) && !(flags & XWFnoindent))
+        nestLimit++;
+    else if (!nestChildren && !nestLimit)
         nestLimit = indent;
 }
 
@@ -789,7 +791,9 @@ void CommonJsonWriter::outputEndNested(const char *fieldname)
         }
     }
     out.append("}");
-    if (indent==nestLimit)
+    if ((flags & XWFonlyindentroot) && !(flags & XWFnoindent))
+        nestLimit--;
+    else if (indent==nestLimit)
         nestLimit = 0;
 }
 
--- a/ecl/hthor/hthor.cpp
+++ b/ecl/hthor/hthor.cpp
@@ -976,7 +976,10 @@ void CHThorXmlWriteActivity::execute()
     headerLength = header.length();
     diskout->write(headerLength, header.str());
 
-    Owned<IXmlWriterExt> writer = createIXmlWriterExt(helper.getXmlFlags(), 0, NULL, (kind==TAKjsonwrite) ? WTJSONRootless : WTStandard);
+    unsigned xwflags = helper.getXmlFlags();
+    if (kind==TAKjsonwrite)
+        xwflags |= XWFonlyindentroot;
+    Owned<IXmlWriterExt> writer = createIXmlWriterExt(xwflags, 0, NULL, (kind==TAKjsonwrite) ? WTJSONRootless : WTStandard);
     writer->outputBeginArray(rowTag); //need to set up the array
     writer->clear(); //but not output it
 
--- a/roxie/ccd/ccdserver.cpp
+++ b/roxie/ccd/ccdserver.cpp
@@ -12143,7 +12143,10 @@ public:
         headerLength = header.length();
         diskout->write(header.length(), header.str());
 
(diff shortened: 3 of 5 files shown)


afishbeck:
@ghalliday this is mean to be a lower risk change because the CommonJsonWriter is used many places.  This allows the write activity to disable the feature causing the problem and shouldn't affect anything not setting the disable flag.

A fix for the nestChildren feature would be need to be stateful since otherwise there is no way to keep track of the mode/state when outputEndNesting() is called.  A change like that would require much more extensive testing.

afishbeck:
@ghalliday @AttilaVamos this change would completely flatten out the json rows.  I'm not sure if people find value in the way the json is formatted now.  It isn't all that readable.  But I could work on a more specific fix if there is value in the current formatting.

afishbeck:
Btw, the current formatting came about by my trying to approximate what xml did.

ghalliday (commented):
@afishbeck I've spent hjalf an hour trying to understand the code, and can't!  Probably being dense, but it needs a comment on the new code explaining what it is doing and why.
I suspect nestLimit may be poorly named since it is more to do with indentation.

ghalliday:
@afishbeck I'm also not sure what the bug is - it isn't clear from the jira.  Is it that it does not appear correctly in eclwatch, or are newline characters significant within json files?

afishbeck:
@ghalliday updated.

Commented on Jira, and the commit:

"Ultimately the problem isn't that we are outputting invalid JSON.  The current CommonJsonWriter output is valid JSON, but scalars can end up with ugly formatting after a nested record. The real problem is that this confuses despray and we end up losing data. I think despray relies on specific newline placement not just valid JSON.

A proper fix for the nestChildren feature (which is a flag parameter of outputBeginNested()) would be a bit more complicated.

So would a fix for despray which currently relies on how the newlines are placed.

The fix in this PR is meant to be a quick, simple, low risk fix for the ECL output scenario only so that despray works as expected."

ghalliday:
@afishbeck I haven't included it in today's RC.  It needs a comment in the code explaining what the flag is doing (and why it works).  It could also do with an example added to the regression test to demonstrate the issue.
Need to be very confident because it is is targeting such an old version.

rtl/eclrtl/rtlformat.cpp
```diff
@@ -766,7 +766,9 @@ void CommonJsonWriter::outputBeginNested(const char *fieldname, bool nestChildre
         appendJSONName(out, fieldname);
     }
     out.append("{");
-    if (!nestChildren && !nestLimit)
+    if ((flags & XWFnonestchildren) && !(flags & XWFnoindent))
+        nestLimit++;
```
jakesmith, line 770: I haven't really understood the original meaning of nestLimit - that might be because scanning through what they did (before this PR) was not what I expected.

The variable declarations could do with some comments on them, briefly describing their semantics.

e.g. it's not clear to why when outputEndNested hit indent==nestLimit, it reset nestLimit = 0 (which means off?).
Should it support arbitrary depth nesting?

jakesmith: I think as you've eluded to in comments in meeting, current behaviour is buggy, returning nestIndent to 0 after inner most indentation is close (?)

With this change and with XWFnonestchildren on, instead of setting nestLimit to current indent level, it will increment and decrement it - which will prevent any nesting of child rows/datasets ? because won't return to 0 until out of it.

(also not sure what effect CommonJsonWriter::cutFrom has, when it sets nestLimit)

jakesmith (commented):
@afishbeck - a couple of questions

rtl/eclrtl/rtlformat.cpp
afishbeck: nestLimit is really badly named.  The original nestLimit was an attempt at emulating what the XmlWriter did, but not very well.  To fix the feature it I think we need more state keeping track of the stack of how outputBeginNested()  and outputEndNested() are called throughout the json structure.

A future change could fix outputBeginNested(..., nestChildren=true); Yes, "prevent any nesting of child rows/datasets". .. and prevent the effect that happens to nested records even if nestChildren wasn't set.  The current change is meant to be a very low risk way of working in a mode that will support despray.  I think despray can only handle specific placement of newlines.  This will fix ECL JSON file output in a way that is despray compatible.

I'm going to change the name of the flag to be something that more reflects the desired outcome, rather than reflecting "turn off nestChildren".

afishbeck:
Renamed flag, and opened a JIRA about the "nestLimit" issues.  https://track.hpccsystems.com/browse/HPCC-27654

ghalliday (approved):
New name is much better.  I'm still not sure which version this should target though.  @richardkchapman?

richardkchapman (approved):

jakesmith (approved):
@afishbeck - it's clearer now, looks good.

PR #16088, opened 2022-05-12
HPCC-27636 Prevent an unecessary/wasteful 2nd call to activity reset()

jakesmith:
also remove a redundant/unused executeChild method

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/thorlcr/graph/thgraph.cpp
+++ b/thorlcr/graph/thgraph.cpp
@@ -1271,6 +1271,9 @@ void CGraphBase::deserializeCreateContexts(MemoryBuffer &mb)
 
 void CGraphBase::reset()
 {
+    if (!started)
+        return;
+
     setCompleteEx(false);
     clearProgressUpdated();
     graphCancelHandler.reset();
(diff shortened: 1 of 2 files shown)


ghalliday (approved):
I think it is probably correct, but hard to follow for several reasons:
- jira doesn't contain details of the problem
- I think CGraphBase is used for both graphs and subgraphs, but it isn't clear from the function names which they apply to (or both?)
- Some comments/tracing within functions suggest it is processing a graph when it is actually processing a subgraph (e.g. GraphPrintLog("Processing graph");)

jakesmith:
> * jira doesn't contain details of the problem

Agreed, have updated the JIRA with a description.

> * I think CGraphBase is used for both graphs and subgraphs, but it isn't clear from the function names which they apply to (or both?)

all CGraphBase (and derivatives) are subgraphs or childgraphs (inc. loop graphs)

Thor sees a [outer workflow] 'graph' as a job CJobBase and derivatives, though agree there's some overlap confusion when Thor first sees the job it is referenced by wuid+'graph name' - but it equated to a CJob* in Thor.

> * Some comments/tracing within functions suggest it is processing a graph when it is actually processing a subgraph (e.g. GraphPrintLog("Processing graph");)

They're all subgraphs (or child graphs), and I'm sure it's not consistent, however, changing it to 'subgraph', when it's a child might be misleading and not sure it is worth making it conditional.

PR #16094, opened 2022-05-13
HPCC-27638 Read AuthDomain settings for cloud ESP

wangkx:
Default settings are used by the existing ESP application.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/applications/sql2ecl/esp.yaml
+++ b/esp/applications/sql2ecl/esp.yaml
@@ -4,6 +4,15 @@ esp:
    daliServers: dali
    loadDaliBindings: false
    auth: ldap
+   authDomain:
+     domainName: "default"
+     authType : "AuthTypeMixed"
+     clientSessionTimeoutMinutes : 120
+     serverSessionTimeoutMinutes : 240
+     unrestrictedResources : "/favicon.ico,/esp/files/*,/esp/xslt/*"
+     invalidURLsAfterAuth : "/esp/login"
+     loginLogoURL : "/esp/files/eclwatch/img/Loginlogo.png"
+     logonURL : "/esp/files/Login.html"
    tls: true
    service:
      port: 8880
--- a/esp/platform/application_config.cpp
+++ b/esp/platform/application_config.cpp
@@ -89,18 +89,24 @@ static void copyDirectories(IPropertyTree *target, IPropertyTree *src)
     }
 }
 
-void addAuthDomains(const char *authType, IPropertyTree *legacyEsp)
+void addAuthDomains(IPropertyTree *appEsp, IPropertyTree *legacyEsp)
 {
-    if (isEmptyString(authType)) //backward compatible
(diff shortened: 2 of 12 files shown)


wangkx:
@afishbeck please review

afishbeck (approved):

wangkx:

esp/applications/sql2ecl/esp.yaml
```diff
@@ -4,6 +4,16 @@ esp:
    daliServers: dali
    loadDaliBindings: false
    auth: ldap
+   authDomains:
+     authDomain:
```
ghalliday, line 8: @afishbeck the code below is treating authDomain as a list, but the yaml are defining it as an object.  Is it purely a backward compatibility feature, or should the yaml reflect that it is a list?

afishbeck: Good catch.  @wangkx can correct me if I'm wrong, but believe having multiple authDomains in baremetal came down to the fact that we ran multiple services on multiple ports in one process.  Each could have a different authDomain configuration.

In containerized mode we support only one service per esp instance and I think we can simplify this by getting rid of the outer tag "authDomains" and keeping "authDomain" as a single object.

We can then change application_config.cpp to treat it as a single object rather than multiple.

esp/platform/application_config.cpp
```diff
@@ -89,18 +89,29 @@ static void copyDirectories(IPropertyTree *target, IPropertyTree *src)
     }
 }
 
-void addAuthDomains(const char *authType, IPropertyTree *legacyEsp)
+void addAuthDomains(IPropertyTree *appEsp, IPropertyTree *legacyEsp)
 {
-    if (isEmptyString(authType)) //backward compatible
-        authType = "AuthPerRequestOnly";
+    IPropertyTree *authDomainsTree = appEsp->queryPropTree("authDomains");
+    if (authDomainsTree)
```
afishbeck, line 95: @wangkx if you agree, we can get rid of the "authDomains" tag here and skip to accessing one authDomain below.  Changing the Yaml input but keeping the output legacy xml the same.

helm/hpcc/values.schema.json
```diff
@@ -793,6 +793,10 @@
       "type": "object",
       "additionalProperties": { "type": ["object"] }
     },
+    "authDomains": {
```
afishbeck, line 796: I think schema needs updating.

wangkx: Updated. Thanks.

afishbeck (approved):

ghalliday (approved):

wangkx:

PR #16098, opened 2022-05-17
HPCC-27600 Revisit usage of unsafe rand() in platform

RussWhitehead:
Implement new pseudo random number generator helper that is threadsafe, fast,
cross platform, and easy to use. Utilize this new helper to create passwords
in the JLib generatePassword interface

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [x] Performance
  - [ ] Security
  - [x] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jutil.cpp
+++ b/system/jlib/jutil.cpp
@@ -992,6 +992,29 @@ jlib_decl int rand_r(unsigned int *seed)
 #endif
 #endif
 
+//------------------------------------------------------------------------
+//static Mersenne Twister 32 bit Pseudo Random Number Generator (PRNG) wrapper
+static class CMTRandomNumberGenerator32 : implements IMTRandomNumberGenerator32
+{
+    std::mt19937 mtPRNG;//instance of Mersenne Twister Random Number Generator
+public:
+    CMTRandomNumberGenerator32()
+    {
+        std::random_device seedGen;//seed with uniformly-distributed integer random number generator that produces non-deterministic random numbers
+        mtPRNG.seed(seedGen());
+    }
+
+    inline virtual size32_t rand32()    override    { return (size32_t)mtPRNG(); }
+    inline virtual size32_t rand32Min() override    { return mtPRNG.min(); }
+    inline virtual size32_t rand32Max() override    { return mtPRNG.max(); }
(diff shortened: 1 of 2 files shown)


system/jlib/jutil.hpp
```diff
@@ -166,6 +166,23 @@ inline int fastRand()
...
+    virtual size32_t rand32() = 0;//return next 32 bit pseudo random number
+    virtual size32_t rand32Min() = 0;//minimum possible random value (0)
+    virtual size32_t rand32Max() = 0;//maximum possible random value, typically 4,294,967,295 on 32 bit implementation
+};
+
+//Query static instance of IMTRandomNumberGenerator32 object, always available and seeded
+extern jlib_decl IMTRandomNumberGenerator32 *queryMTRandomGen32();
+
+//Handy macro to get a single pseudo random number
+#define GET_MTRAND32() queryMTRandomGen32().rand32()
```
richardkchapman, line 184: Either the function should return a reference or this should say -> , I think?

system/jlib/jutil.cpp
```diff
@@ -992,6 +992,29 @@ jlib_decl int rand_r(unsigned int *seed)
...
+{
+    std::mt19937 mtPRNG;//instance of Mersenne Twister Random Number Generator
+public:
+    CMTRandomNumberGenerator32()
+    {
+        std::random_device seedGen;//seed with uniformly-distributed integer random number generator that produces non-deterministic random numbers
+        mtPRNG.seed(seedGen());
+    }
+
+    inline virtual size32_t rand32()    override    { return (size32_t)mtPRNG(); }
```
ghalliday, line 1007: I don't think this is thread safe.  There is no guarantee the object can be accessed from multiple threads at the same time.

ghalliday (changes requested):
I don't think the code is correct, and I'm not sure this change servers any purpose.  Most calls to rand() were clarified as calls to fastRand() as part of HPCC-18585.

The only examples I found were crypthead in crypt.h and generateNonce() in jwtSecurity.cpp - both of which should probably be fixed.

PR #16107, opened 2022-05-19
HPCC-25041 Fold ftslave functionality into dafilesrv

jakesmith:
Make dfuserver/filesprayer send commands to dafilesrv for ftslave
functionality, instead of spinning up ftslave processes.

Default on in containerized mode, for now default off in bare-metal.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dali/ft/daftformat.hpp
+++ b/dali/ft/daftformat.hpp
@@ -99,6 +99,7 @@ typedef IArrayOf<IFormatPartitioner> FormatPartitionerArray;
 extern DALIFT_API IFormatProcessor * createFormatProcessor(const FileFormat & srcFormat, const FileFormat & tgtFormat, bool calcOutput);
 extern DALIFT_API IOutputProcessor * createOutputProcessor(const FileFormat & format);
 
-extern DALIFT_API IFormatPartitioner * createFormatPartitioner(const SocketEndpoint & ep, const FileFormat & srcFormat, const FileFormat & tgtFormat, bool calcOutput, const char * slave, const char *wuid);
+class FileSprayer;
+extern DALIFT_API IFormatPartitioner * createFormatPartitioner(FileSprayer &sprayer, const SocketEndpoint & ep, const FileFormat & srcFormat, const FileFormat & tgtFormat, bool calcOutput, const char * slave, const char *wuid);
 
 #endif
--- a/dali/ft/filecopy.cpp
+++ b/dali/ft/filecopy.cpp
@@ -345,37 +345,39 @@ void FileTransferThread::prepareCmd(MemoryBuffer &msg, unsigned version)
 bool FileTransferThread::launchFtSlaveCmd(const SocketEndpoint &ep)
 {
     SocketEndpoint connectEP(ep);
-#ifdef _CONTAINERIZED
-    if (sprayer.useFtSlave)
-    {
-        //In containerized world all processes are executed locally, so make sure we try and connect to a local instance
-        connectEP.set("localhost");
-    }
-    else
-    {
-        Owned<IPropertyTree> serviceTree;
-        if (!isEmptyString(sprayer.sprayServiceName))
-        {
-            VStringBuffer serviceQualifier("services[@name='%s']", sprayer.sprayServiceName.get());
-            serviceTree.setown(getGlobalConfigSP()->getPropTree(serviceQualifier));
-            if (!serviceTree)
-                throw makeStringExceptionV(0, "launchFtSlaveCmd: failed to find dafilesrv service named: '%s'", sprayer.sprayServiceName.get());
-            const char *serviceAppType = serviceTree->queryProp("@application");
-            if (!strsame("spray", serviceAppType))
-                throw makeStringExceptionV(0, "launchFtSlaveCmd: configured service '%s' is of application type '%s' ('spray' type required)", sprayer.sprayServiceName.get(), nullIfEmptyString(serviceAppType));
-        }
-        else // find 1st of type 'spray'
-        {
-            Owned<IPropertyTreeIterator> directIOServices = getGlobalConfigSP()->getElements("services[@type='spray']");
-            if (directIOServices->first())
-                serviceTree.set(&directIOServices->query());
-            else
-                throw makeStringException(0, "launchFtSlaveCmd: no 'spray' dafilesrv services found");
-        }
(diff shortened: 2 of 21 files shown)


dali/ft/daftformat.hpp
```diff
@@ -99,6 +99,7 @@ typedef IArrayOf<IFormatPartitioner> FormatPartitionerArray;
 extern DALIFT_API IFormatProcessor * createFormatProcessor(const FileFormat & srcFormat, const FileFormat & tgtFormat, bool calcOutput);
 extern DALIFT_API IOutputProcessor * createOutputProcessor(const FileFormat & format);
 
-extern DALIFT_API IFormatPartitioner * createFormatPartitioner(const SocketEndpoint & ep, const FileFormat & srcFormat, const FileFormat & tgtFormat, bool calcOutput, const char * slave, const char *wuid);
+class FileSprayer;
+extern DALIFT_API IFormatPartitioner * createFormatPartitioner(FileSprayer &sprayer, const SocketEndpoint & ep, const FileFormat & srcFormat, const FileFormat & tgtFormat, bool calcOutput, const char * slave, const char *wuid);
```
ghalliday, line 103: minor: could use a useFtSlave boolean instead of passing in the class and creating a backward dependency

ghalliday, line 103: minor: could use a useFtSlave boolean instead of passing in the class and creating a backward dependency

dali/ft/filecopy.cpp
```diff
@@ -293,10 +293,161 @@ bool FileTransferThread::catchReadBuffer(ISocket * socket, MemoryBuffer & msg, u
...
+    SocketEndpoint connectEP(ep);
+#ifdef _CONTAINERIZED
+    if (sprayer.useFtSlave)
+    {
+        //In containerized world all processes are executed locally, so make sure we try and connect to a local instance
+        connectEP.set("localhost");
+    }
+    else
+    {
+        Owned<IPropertyTree> serviceTree;
```
ghalliday, line 356: Can this code be in the non containerized branch - to check it compiles, and because it would probably be a good way to do it for bare-metal as well.

ghalliday, line 356: Can this code be in the non containerized branch - to check it compiles, and because it would probably be a good way to do it for bare-metal as well.

dali/ft/filecopy.cpp
```diff
@@ -293,10 +293,161 @@ bool FileTransferThread::catchReadBuffer(ISocket * socket, MemoryBuffer & msg, u
...
+
+    LOG(MCdebugProgressDetail, job, "Finished generating part %s [%p] ok(%d) error(%d)", url.str(), this, (int)ok, (int)(error!=NULL));
+
+    // if communicating with ftslave, the process has a final ack wait
+    if (sprayer.useFtSlave)
+    {
+        // ftslave sends back a final ack.
+        msg.clear().append(true);
+        catchWriteBuffer(socket, msg);
+        if (sprayer.options->getPropInt("@fail", 0)) // JCSMORE - legacy? not set anywhere afaics
```
ghalliday, line 443: I suspect it is a debug option which was used to test recovery or something similar.

ghalliday, line 443: I suspect it is a debug option which was used to test recovery or something similar.

dali/ft/ftslavelib.cpp
```diff
@@ -0,0 +1,135 @@
...
+#include "fttransform.ipp"
+#include "daftformat.hpp"
+#include "daftcfg.hpp"
+#include "mptag.hpp"
+#include "ftslavelib.hpp"
+
+
+bool processPullCommand(ISocket * masterSocket, MemoryBuffer & msg)
+{
+    srand((int)get_cycles_now());
```
ghalliday, line 40: not new code, but I think this should be ftslave only, or deleted.  There is another call in deserializeAction().

I think it is is only used to make sure each slave does not send their progress updates at the same time.  Better would probably be to directly use get_cycles_now since they will almost certainly not match:

    int adjust = get_cycles_now() % updateFrequency - (updateFrequency/2);

ghalliday, line 40: not new code, but I think this should be ftslave only, or deleted.  There is another call in deserializeAction().

I think it is is only used to make sure each slave does not send their progress updates at the same time.  Better would probably be to directly use get_cycles_now since they will almost certainly not match:

    int adjust = get_cycles_now() % updateFrequency - (updateFrequency/2);

dali/ft/ftslavelib.cpp
```diff
@@ -0,0 +1,135 @@
...
+    TransferServer server(masterSocket);
+    server.deserializeAction(msg, FTactionpull);
+    return server.pull();
+}
+
+
+bool processPushCommand(ISocket * masterSocket, MemoryBuffer & msg)
+{
+    srand((int)get_cycles_now());
```
ghalliday, line 49: another srand (not thread safe)

ghalliday, line 49: another srand (not thread safe)

ghalliday (commented):
@jakesmith a few comments, but generally looks much less painful than I was expecting.

dali/ft/daftformat.hpp
jakesmith: I decided to pass in the sprayer object as a context as the format partitioner is owned by it.
But yes could reduce dependency by passing in just the config option needed.

I don't have strong feeling about whether it should be change/stay same ?

dali/ft/filecopy.cpp
jakesmith: it coud, with a minor change. I'll remove the compiler time conditional.

dali/ft/ftslavelib.cpp
jakesmith: agree, odd code, get_cycles_now should suffice in both cases.
I'll change.

jakesmith:
@ghalliday - please see new commit and let me know re. https://github.com/hpcc-systems/HPCC-Platform/pull/16107#discussion_r879384015

ghalliday (approved):

ghalliday:
@richardkchapman which version should this target? (see comments in the jira)

PR #16109, opened 2022-05-19
HPCC-27471 Add data masking plug-in framework with sample plug-in

timothyklemm:
- Declare interface to abstract trace logging requirements on a per request
  basis.
- Define an interface implementation that output switching between jlog
  functions, used by default, and printf, used in console applications.
- Declare interfaces to standardize masking requirements.
- Define basic interface implementations that manage relationships between
  instances of the various interface implementations.
- Define implementation extensions demonstrating the incorporation of custom
  logic.
- Create a sample plug-in using the demonstration classes.
- Provide code to manage plug-ins.
- Integrate masking into the ESDL integration script operation "trace".
- Integrate masking into the ESDL ESP.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:


timothyklemm:
This PR impacts the LN project. LN pull request 1007 (https://github.com/hpcc-systems/LN/pull/1007) is required to fix the issue.

system/masking/include/readme.md
```diff
@@ -0,0 +1,439 @@
...
+
+Obfuscation is always applied based on a single snapshot of a domain.
+
+### Profile
+
+A *profile* is a collection of obfuscation requirement snapshots for a single *domain*. Each instance declares a minimum version, a maximum version, and a default version. When defined by a single *plugin*, a decision to separate versions into distinct *profiles* is a question of preference. If defined by multiple *plugins*, each version transition between plugins requires a new *profile*.
+
+If a *domain* is defined by 10 versions, the implementation could be one *profile* defining all 10 versions, 10 *profiles* each defining one version, or any combination in between, if all 10 versions are defined by the same *plugin*. If a second plugin was introduced at version six, then one to five profiles are required for versions one to five and a separate one to five profiles are required for versions sixe to 10.
```
ghalliday, line 28: typo: sixe

system/masking/include/readme.md
```diff
@@ -0,0 +1,439 @@
...
+
+The `canMaskContent` method can be used to establish whether a snapshot supports this operation. A result of *true* indicates that the plugin is capable of performing obfuscation in response to a request. Whether a combination of input parameters exists that results in obfuscation depends on the definition of the snapshot.
+
+The `hasRule` method indicates if at least one *rule* is selected in the snapshot for a given content type. *Rules* not associated with any content type may be interrogated using an empty string.
+
+> It is not possible to discern any information about the selected rules, such as their associated *value types* or the cues used to apply them in a buffer.
+
+#### Implementations
+
+The groundwork exists for two types of *rules*, serial and parallel. Serial rules, as the name implies, are evaluated sequentially. Parallel *rules*, on the other hand, are evaluated concurrently. Concurrent evaluation is expected to be more efficient than sequentual, but no concurrent implementation is provided.
```
ghalliday, line 341: typo: sequential

ghalliday (commented):
I don't understand the problem or the context it is being used in well enough to provide any meaningful review.

timothyklemm:
@afishbeck and @asselitx, I've rebased to 8.8.x and resolved merge conflicts with the recent script changes.

system/xmllib/libxml_xpathprocessor.cpp
```diff
@@ -978,6 +978,35 @@ class CLibXpathContext : public CInterfaceOf<IXpathContext>, implements IVariabl
         return xml;
     }
 
+    virtual bool prepareTrace(ICompiledXpath* select, StringBuffer& content, bool& isValue)
```
afishbeck, line 981: This shares an awful lot of code with the trace function below.  Can trace call prepareTrace?  I haven't looked yet, but I'm wondering if/where the trace function is still being used.

esp/esdllib/esdl_script.cpp
```diff
@@ -1402,7 +1489,30 @@ class CEsdlTransformOperationTrace : public CEsdlTransformOperationWithoutChildr
...
+                {
+                    if (isValue)
+                    {
+                        // TODO: need value type and optional mask style to mask a value
+                    }
+                    else
+                    {
+                        char* buffer = const_cast<char*>(content.str());
+                        if (!masker->maskContent("xml", buffer, 0, content.length()))
+                            tracer.log(MCuserInfo, "no embedded values were masked");
```
afishbeck, line 1508: It seems a little verbose to trace that we didn't have to mask anything in the following trace content.

timothyklemm: Agreed. It's a sanity check during development and debugging. Will remove before completion.

system/xmllib/libxml_xpathprocessor.cpp
timothyklemm: Trace should not, in my opinion, be called. I would like to remove it, but wasn't sure how much backward compatibility would be a concern. If it remains, it should be changed to use prepareTrace.

timothyklemm: If trace remains, it will not mask so keeping it is a risk.

I'm also unsure about the name prepareTrace. It makes sense in the sense that it prepares the content that will be traced. I just don't know if trace output is the only reason we will ever need this functionality, or if a more general name would be better.

system/masking/include/readme.md
```diff
@@ -0,0 +1,439 @@
...
+
+*Profiles* define all obfuscation operations:
+
+- `maskValue` applies obfuscation to a given character range based on *value type* and *mask style*.
+- `maskContent` applies all relevant *rules* to a given character range, obfuscating matching substrings using the default obfuscation format of the matching *rule's* *value type*. *Rule* relevance is based on the requested version and custom *context* properties given with the request.
+- `maskMarkupValue` applies obfuscation to a given character range when a relevant *rule* applies to a given a given document location. Given an element or attribute value, the value is obfuscated if a relevant *rule* applies to the element or attribute. Such a determination may require additional document context, so a callback must be supplied to satisfy requests for context.
```
asselitx, line 34: "a given" repeated twice.

asselitx (commented):
I'm not entirely finished yet combing through datamaskingplugin.hpp, but wanted to give you visibility to my thoughts so far.

system/masking/include/datamaskingplugin.hpp
```diff
@@ -0,0 +1,2072 @@
...
+                    start = nullptr;
+            }
+            return applied;
+        }
+    private:
+        inline static char* strstr_sensitive(const char* haystack, const char* needle)
+        {
+            return const_cast<char*>(strstr(haystack, needle));
+        }
+        inline static char* strstr_insensitive(const char* haystack, const char* needle)
```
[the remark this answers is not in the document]
afishbeck: There is an existing implementation named "stristr" which I think also only currently works for ascii.  Could common the functions up with whatever implementation is most efficient.  When going beyond ascii I agree, I would probably focus on utf8.  There is a local aware toupper although the wrapper seems to do a dynamic cast and I would probably unwind the wrapper a bit and make sure I avoided that.  Would still want to look at the efficiency of toupper compared to a homegrown table like the one in stristr.

system/xmllib/libxml_xpathprocessor.cpp
afishbeck: As we discussed offline, I agree we should get rid of the old function, and name/design prepareTrace more generically since it could be useful for serializing content beyond tracing.

esp/services/esdl_svc_engine/esdl_binding.hpp
```diff
@@ -209,6 +214,27 @@ class EsdlServiceImpl : public CInterface, implements IEspService
     virtual bool subscribeServiceToDali() override {return false;}
     virtual bool attachServiceToDali() override {return false;}
     virtual bool detachServiceFromDali() override {return false;}
+
+private:
+    bool initMaskingEngineDirectory(const char* dir);
+    template <typename file_loader_t>
+    bool initMaskingEngineDirectory(const char* dir, const char* mask, file_loader_t loader)
```
afishbeck, line 221: Can the implementation of this be moved to the cpp file with the other related functions?

system/xmllib/xpathprocessor.hpp
```diff
@@ -152,11 +153,11 @@ extern "C" XMLLIB_API IXpathContext*  getXpathContext(const char * xmldoc, bool
 #define ESDLScriptCtxSection_ScriptRequest "script_request"
 #define ESDLScriptCtxSection_ScriptResponse "script_response"
 
-interface IEsdlScriptContext : extends IInterface
+interface IXmlDocContext : extends IInterface
```
afishbeck, line 156: I'm not sure what the right name is, but I think IXmlDocContext is a bit too generic.  We're dealing with sections, cleaning up between scripts, directing tracing.

testing/unittests/datamaskingtests.cpp
```diff
@@ -0,0 +1,1649 @@
...
+      assertQueryProfile(engine, "", 0, true);
+    }
+
+    void testLegacyDomains()
+    {
+      static const char* cfgTxt = R"!!!(
+        MaskingPlugin:
+          library: datamasker
+          entry-point: newPartialMaskSerialToken
+          Profile:
```
afishbeck, line 114: Quite trivial, but I think we should be standardizing our yaml style.  Take a look at some of the hpcc YAML files for indenting, case, etc.  Case can be tricky if you are trying to support both xml and yaml.  What is the primary format?

testing/unittests/datamaskingtests.cpp
```diff
@@ -0,0 +1,1649 @@
...
+
+    void assertTextIterator(ITextIterator* it, const std::set<std::string>& expectedValues, const char* label)
+    {
+      bool failed = false;
+      Owned<ITextIterator> rawValues(it);
+      std::set<std::string> actualValues;
+      ForEach(*it)
+      {
+        std::string value(it->query());
+        auto result = actualValues.insert(value);
```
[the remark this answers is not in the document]
afishbeck: Personally I don't think auto should be used for basic types that won't change.  Having the type there becomes self documenting.  Auto is extremely helpful with long templatized types though.  And types that may change based on templates, etc.

system/masking/include/traceproducer.h
```diff
@@ -0,0 +1,120 @@
...
+ * to manage context. It is intended to be used by code that creates trace output without regard for
+ * contextual requirements. Included functionality includes:
+ *
+ * - Message creation based on various combinations of audience and class.
+ * - Determination of whether or not an audience and class combination would produce a message.
+ *
+ * As long as a detail level is included in a message category, levels cannot be ignored. The
+ * interface does deemphasize them based on an expectation that they wull be removed in the future.
+ */
+interface ITraceProducer : extends IInterface
```
afishbeck, line 40: I'm not sure the name producer is right for this interface.

testing/unittests/datamaskingtests.cpp
timothyklemm: Can you identify a file that illustrates what you have in mind? I see combinations of title case, camel case and snake case, and found a file that couldn't decide whether to indent one, two, or three spaces. I may be unique with hyphenated names, but I can't tell how to standardize it.

esp/services/esdl_svc_engine/esdl_binding.hpp
timothyklemm: I can try. Since it's private, maybe the definition doesn't have to be in the header like most template definitions.

system/masking/include/datamaskingplugin.hpp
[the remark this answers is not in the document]
timothyklemm: I saw that, but wanted to avoid requiring a system utility to link to an ESP library. I tried strcasestr, which worked in Linux but broke the Windows builds.

testing/unittests/datamaskingtests.cpp
timothyklemm: The format can be either yaml or xml. I would expect yaml to be preferred if standalone, but accept XML so a single definition can either be deployed as a standalone file or included in the ESDL binding.

afishbeck: The prime example especially from a user's configuration perspective is https://github.com/hpcc-systems/HPCC-Platform/blob/master/helm/hpcc/values.yaml.
Notice indent is 2 spaces.
The dash of an array entry lines up with the parent tag.
Names start with lower case and are camel case after that where appropriate.
Names with dashes could be up for discussion I wasn't as concerned about that but we should discuss it.

system/masking/include/datamaskingengine.hpp
```diff
@@ -0,0 +1,611 @@
...
+            size_t addedProfiles = 0;
+            size_t addedDomains = 0;
+            if (streq(configuration.queryName(), "MaskingPlugin"))
+            {
+                if (!load(configuration, addedDomains, addedProfiles))
+                    failed = true;
+            }
+            else
+            {
+                Owned<IPTreeIterator> pluginConfigs(configuration.getElements("//MaskingPlugin"));
```
afishbeck, line 453: Can MaskingPlugin really exist anywhere in the tree?

system/masking/include/datamaskingengine.hpp
```diff
@@ -0,0 +1,611 @@
...
+            TPluginLoader<DataMaskingPluginEntryPoint> loader;
+        public:
+            Loader(ITraceProducer& _trace)
+                : loader(nullptr, nullptr, "@library", "@entry-point", _trace)
+            {
+                trace.set(&_trace);
+            }
+            IDataMaskingProfileIterator* create(const IPTree& configuration)
+            {
+                Owned<IDataMaskingProfileIterator> plugin(loader.create<IDataMaskingProfileIterator>(configuration, [&](DataMaskingPluginEntryPoint entryPoint) { return entryPoint(configuration, *trace); }));
```
afishbeck, line 197: I would usually name the instance of an iterator as plural.

esp/esdllib/esdl_script.cpp
```diff
@@ -2548,7 +2658,7 @@ void processServiceAndMethodTransforms(IEsdlScriptContext * scriptCtx, std::init
     if (isEmptyString(reqtype))
         throw MakeStringException(ESDL_SCRIPT_Error, "ESDL script request name not set");
 
-    IEspContext *context = reinterpret_cast<IEspContext*>(scriptCtx->queryEspContext());
+    IEspContext *context = reinterpret_cast<IEsdlScriptContext*>(scriptCtx->queryUsageContext())->espContext();
```
afishbeck, line 2661: Why is reinterpret_cast<IEsdlScriptContext*>(scriptCtx->queryUsageContext()) necessary?  You already have an IEsdlScriptContext, get the pointer back as a void *, and then recast back to a IEsdlScriptContext?

system/masking/include/datamaskingengine.hpp
timothyklemm: As written, it apparently can. I must have been feeling generous when I wrote it, allowing some leeway in acceptable document structure. I might have been thinking a file which could be used with the manifest tool for inclusion in a binding might wrap the elements.

esp/esdllib/esdl_script.cpp
timothyklemm: It isn't.

timothyklemm:
This PR is now dependent upon https://github.com/hpcc-systems/HPCC-Platform/pull/16436, which refactors IEsdlScriptContext.

This PR continues to impact the LN project. A new PR will be created after https://github.com/hpcc-systems/LN/pull/1007 is accepted and merged.

system/masking/include/datamaskingengine.hpp
timothyklemm: Singular is correct, because a plugin is by definition a profile iterator. Will change to `profiles` to address confusion.

esp/esdllib/esdl_script.cpp
```diff
@@ -53,26 +53,65 @@ class CEsdlScriptContext : public CInterfaceOf<IEsdlScriptContext>
...
+    virtual bool enableMasking(const char* domainId, uint8_t version) override
+    {
+        if (!maskingEngine)
+        {
+            trace().uwarnlog("enable masking request not completed - no masking engine");
+            return false;
+        }
+        if (maskingEnabled())
+        {
+            trace().uwarnlog("enable masking request not completed - already enabled");
```
afishbeck, line 75: Can you output the current and requested domain and version?  If they match may not even need to warn?

esp/esdllib/esdl_script.cpp
```diff
@@ -53,26 +53,65 @@ class CEsdlScriptContext : public CInterfaceOf<IEsdlScriptContext>
...
     {
         if (_espCtx)
             espCtx.set(_espCtx);
         else
             espCtx.setown(createEspContext(nullptr));
         docModel.setown(createSectionalXmlDocModel(this));
+        tracer.setown(new CModularTracer());
+        jlogSink.setown(tracer->getSink());
+        consoleSink.setown(new CConsoleTraceMsgSink());
+        tracer->setSink(jlogSink);
```
afishbeck, line 113: Why get and then set the default sink to the same value without any change?

esp/esdllib/esdl_script.cpp
```diff
@@ -53,26 +53,65 @@ class CEsdlScriptContext : public CInterfaceOf<IEsdlScriptContext>
...
         : functionRegister(_functionRegister)
     {
         if (_espCtx)
             espCtx.set(_espCtx);
         else
             espCtx.setown(createEspContext(nullptr));
         docModel.setown(createSectionalXmlDocModel(this));
+        tracer.setown(new CModularTracer());
+        jlogSink.setown(tracer->getSink());
+        consoleSink.setown(new CConsoleTraceMsgSink());
```
afishbeck, line 112: If consoleSink is less likely to be used, should it be created lazily.  Perhaps an ensureConsoleSink() method which creates if it doesn't exist, and is called by setTraceToStdout()?

esp/esdllib/esdl_script.cpp
afishbeck: If it is necessary, could move this up a line to group the two calls together with a comment explaining why necessary.

esp/esdllib/esdl_script.cpp
```diff
@@ -1451,22 +1490,27 @@ class CEsdlTransformOperationTrace : public CEsdlTransformOperationWithoutChildr
...
             {
-                if (scriptContext->getTraceToStdout())
+                IDataMaskingProfileContext* masker = scriptContext->masker();
+                if (masker)
                 {
-                    if (m_label.isEmpty())
-                        printf("%s\n", content.str());
+                    if (isValue)
+                    {
+                        // TODO: need value type and optional mask style to mask a value
```
afishbeck, line 1501: may want to trace something saying that they selected a value, not content and it couldn't be traced because...

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -208,6 +208,64 @@ bool EsdlServiceImpl::loadLoggingManager(Owned<ILoggingManager>& manager, IPTree
...
+        return true;
+    }
+}
+
+bool EsdlServiceImpl::initMaskingEngineDirectory(const char* dir)
+{
+    bool failed = false;
+    if (!isEmptyString(dir))
+    {
+        if (!initMaskingEngineDirectory(dir, "*.yaml", createPTreeFromYAMLFile))
```
afishbeck, line 261: Don't necessarily need to change anything, but just a note that you should consider... most of the places I've used yaml files so far hasn't just been an alternative mark up to xml.  The format was either designed around our use of helm, kubernetes, configmaps (also kubernetes), etc.

Just consider the fact that loading them as if they were the same as your xml may mean you have to recognize another yaml format in the future coming from a different configuration/management pathway.

esp/esdllib/esdl_script.cpp
timothyklemm: If I had to guess, `tracer` did not originally create the default sink. Both sinks were created here, and one was applied as the default. The creation of the default was updated, but setting the default wasn't.

esp/esdllib/esdl_script.cpp
timothyklemm: Are you asking for value tracing to be blocked until the operation is updated to specify the value type? I was trying to not take away existing functionality before extending the syntax, but I can if that's what you're expecting.

esp/services/esdl_svc_engine/esdl_binding.cpp
timothyklemm: One of my assumptions is that we will (eventually) support service-specific profiles. This strongly suggests that profiles will be written using xml, and may even be generated by the manifest tool. My support for yaml was intentionally as an alternative to xml.

I would like to discuss this offline, even if not changing things now.

system/masking/plugins/datamasker/readme.md
```diff
@@ -0,0 +1,182 @@
...
+> No content type knowledge is implied by this class. An instance with **@contentType** of *xml* does not inherently know how to find values in XML markup. The defined tokens must include characters such as `<` and `>` to match an element name and quotes to match attribute values.
+
+#### TPlugin
+
+An implementation of `IDataMaskingProfileIterator` used for transforming a configuration property tree into a collection of profiles to be returned by a library entry point function. Created profiles are of the same class, identified by the template parameter **profile_t**.
+
+Configuration includes:
+
+- **profile** An optional and repeating element defining a profile. The content of this element depends on **profile_t**.
+- **...** If and only if **profile** is not included as a child of the configuration, the configuration node itself will be viewed as a configursation for **profile_t**. The name of the configuration node is not relevant in this case; it is assumed to be a node recognized by the host platform's `IDataMaskingEngine` implementation as a plugin configuration node.
```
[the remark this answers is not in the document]
timothyklemm: No. The given property tree may contain a list of `profile` elements, which is the first bullet point. If the given property tree contains no `profile` elements, it is assumed to be equivalent to a single `profile` element (equivalent because the name is not relevant). I understand the confusion, but I'm not how to better express this.

asselitx (commented):
IM or call if you want to discuss any questions in depth.

system/masking/include/datamaskingengine.hpp
```diff
@@ -0,0 +1,621 @@
...
+            ReadLockBlock block(domainLock);
+            const Domain* domain = selectDomain(identifier);
+            return (domain ? domain->queryProfile(version ? version : domain->queryDefaultVersion()) : nullptr);
+        }
+        Domain* selectDomain(const char* identifier) const
+        {
+            if (domains.empty())
+                return nullptr;
+            if (isEmptyString(identifier))
+                return domains.front();
```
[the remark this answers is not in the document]
timothyklemm: Yes.

system/masking/include/datamaskingplugin.hpp
```diff
@@ -1929,7 +1929,7 @@ namespace DataMasking
             return false;
         }
     protected:
-        using Tester = std::function<int(int)>;
+        using Tester = int (*)(int); //std::function<int(int)>;
```
timothyklemm, line 1932: I'll remove the commented code. Wasn't sure change would fix the Windows build break.

timothyklemm:
@afishbeck commit https://github.com/hpcc-systems/HPCC-Platform/pull/16109/commits/cbeb94ec7c1ccc4dc1470811cbec8b4377f9727f moves the description of optional and trace to a separate file, as we discussed. Optional is described in the new file, hopefully referring to it as legacy, discouraging its use, and recommending accepting the exceptions makes it sufficiently clear we don't think it should be used.

esp/esdllib/esdl_script.cpp
```diff
@@ -29,6 +29,92 @@
...
+        {
+            traceToStdout = val;
+            tracer->setSink(val ? consoleSink.get() : jlogSink.get());
+        }
+    }
+    virtual bool getTraceToStdout() const override { return traceToStdout; }
+    virtual void setTestMode(bool val) override { testMode = val; }
+    virtual bool getTestMode() const override { return testMode; }
+    virtual ITracer& trace() const override { return *tracer; }
+    virtual bool enableMasking(const char* domainId, uint8_t version) override
```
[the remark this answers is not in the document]
asselitx: I think you're right. I don't recall what prompted me to think that would be desired.

esp/services/esdl_svc_engine/esdl_binding.cpp
```diff
@@ -232,6 +290,14 @@ void EsdlServiceImpl::init(const IPropertyTree *cfg,
 
         loadLoggingManager(m_oStaticLoggingManager, srvcfg->queryPropTree("LoggingManager"));
 
+        bool initMaskingFailed = false;
+        if (!initMaskingEngineDirectory(espcfg->queryProp("@maskingProfileDir")))
+            initMaskingFailed = true;
+        if (!initMaskingEngineEmbedded(m_oStaticMaskingEngine, srvcfg, false))
```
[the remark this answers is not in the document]
asselitx: No I think your explanation makes sense.

system/masking/include/datamaskingplugin.hpp
```diff
@@ -0,0 +1,2343 @@
...
+            if (!profile)
+            {
+                tracer->ierrlog("%s: profile creation failed", configuration.queryName());
+            }
+            else if (!profile->configure(configuration))
+            {
+                const char* tmp = configuration.queryProp("@name");
+                if (isEmptyString(tmp))
+                    tmp = "<unnamed>";
+                tracer->ierrlog("%s: profile configuration failed", profile->tracePrefix());
```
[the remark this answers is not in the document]
asselitx: OK

system/masking/include/readme.md
```diff
@@ -0,0 +1,483 @@
...
+Requirements can change over time. To support this, a *domain* can be seen as a collection of requirement snapshots where each snapshot defines the complete set of requirements for the *domain* at a point in time. Snapshots are referenced by unique version numbers, which should be sequential starting at *1*.
+
+Obfuscation is always applied based on a single snapshot of a domain's requirements.
+
+*Domains* are represented in the framework interface as text identifiers. Each distinct *domain* is identified by at least one unique identifier.
+
+### Masker
+
+A *masker* is a provider of obfuscation for a single snapshot of a *domain's* requirements. There are three masking operations defined in this framework. Each instance decides which of the three it will support, and how it will support them. The three operations are:
+
```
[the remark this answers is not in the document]
asselitx: I think the change reads well

system/masking/include/readme.md
```diff
@@ -284,13 +288,16 @@ bool maskValueUnconditionally(const char* valueType, const char* maskStyle, char
 
 #### Overview
 
-Given a single *domain* datum, obfuscate the content as needed. The framework anticipates two interpretations of "as needed", either conditional or unconditional. A conditional need assumes that a snapshot includes a complete description of which data require obfuscation, and only values corresponding to defined types are affected. An unconditional need suggests that a caller's request for obfuscation establishes the need, and values associated with unrecognized types are obfuscated.
+Given a single *domain* datum, obfuscate the content "as needed". The framework anticipates two interpretations of "as needed", either conditional or unconditional:
+
```
asselitx, line 292: I really like this updated explanation of conditional vs unconditional.

system/masking/include/readme.md
```diff
@@ -0,0 +1,483 @@
...
+The `hasValueType` method of `IDataMaskerInspector` can be used to check if a given name is selected in the snapshot. The reserved name "\*" may be used to detect unconditional obfuscation support when available in the snapshot. Names defined yet unselected in the snapshot are not directly detectable, but can be detected by comparing results of using multiple *context* configurations.
+
+> The name "\*" is reserved by the abstraction for compatibility checks.
+
+To confirm the availability of a specific *mask style* first requires confirmation of the *value type* to which the *mask style* is related. Use `queryValueType` to obtain the defined instance and, from the result, call `queryMaskStyle` to confirm the existence of the *mask stye*. For each of these, corresponding `get...` methods are defined to obtain new references to the objects. The methods are declared by `IDataMaskerInspector`.
+
+#### Implementations
+
+The included implementations are inherently conditional. Obfuscation depends on matching `valueType` with a selected *value type* instance in the snapshot, where an instance is selected if it belongs to the currently selected *value type* set.
```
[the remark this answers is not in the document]
asselitx: looks good

system/masking/include/readme.md
```diff
@@ -0,0 +1,483 @@
...
+
+### maskMarkupValue
+
+```
+bool maskMarkupValue(const char* element, const char* attribute, char* buffer, size_t offset, size_t length, IDataMaskingDependencyCallback& callback) const;
+
+```
+
+#### Overview
+
```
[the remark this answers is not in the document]
asselitx: looks good

asselitx (approved):
Changes look good and read clearly.

system/masking/include/datamaskingplugin.hpp
```diff
@@ -0,0 +1,2343 @@
...
+            Owned<IPTreeIterator> it(configuration.getElements("profile"));
+            ForEach(*it)
+            {
+                configureProfile(it->query());
+            }
+            if (profiles.empty())
+            {
+                configureProfile(configuration);
+            }
+            cur = profiles.end();
```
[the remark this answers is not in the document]
asselitx: OK makes sense.

system/masking/include/datamaskingplugin.hpp
```diff
@@ -0,0 +1,2343 @@
...
+                end = finder(start, endToken.c_str());
+                if (end)
+                {
+                    if (valueType.applyMask(&context, nullptr, start, 0, end - start))
+                        applied = true;
+                    start = finder(end + endToken.length(), startToken.c_str());
+                }
+                else
+                    start = nullptr;
+            }
```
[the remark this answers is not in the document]
asselitx: OK that makes sense.

afishbeck (approved):

timothyklemm:
@ghalliday this is ready to merge.

PR #16126, opened 2022-05-25
HPCC-27694 EsdlScript synchronize tag to synchronize background tasks

afishbeck:
Example, the following script calls out to two external services and creates some additional response content at the same time:

```
 <!-- the following element contains child tasks that will be syncronized -->
 <es:synchronize>

  <!-- script to run while the http calls are blocked calling out to the external services -->
  <es:script>
    <es:element name="SyncBackgroundStuff">
      <es:for-each select="es:tokenize('aaa,bbb,ccc,yyy', ',')">
        <es:add-value target="value" value="."/>
      </es:for-each>
    </es:element>
  </es:script>

  <!-- first external service call -->
  <es:http-post-xml url="'http://127.0.0.1:9876'" section="logging" name="roxiestuff1">
    <es:content>
      <es:element name="Envelope">
        <es:namespace prefix="soap" uri="http://schemas.xmlsoap.org/soap/envelope/" current="true" />
        <es:element name="Body">
          <es:element name="roxieechopersoninfoRequest">
            <es:namespace uri="urn:hpccsystems:ecl:roxieechopersoninfo" current="true" />
            <es:element name="roxieechopersoninforequest">
              <es:element name="Row">
                <es:element name="Name">
                  <es:set-value target="First" value="'aaa'"/>
                  <es:set-value target="Last" value="'bbb'"/>
                  <es:element name="Aliases">
                    <es:set-value target="Alias" value="'ccc'"/>
                    <es:add-value target="Alias" value="'ddd'"/>
                  </es:element>
                </es:element>
              </es:element>
            </es:element>
          </es:element>
        </es:element>
      </es:element>
    </es:content>
  </es:http-post-xml>

  <!-- second external service call -->
  <es:http-post-xml url="'http://127.0.0.1:9876'" section="logging" name="roxiestuff2">
    <es:content>
      <es:element name="Envelope">
        <es:namespace prefix="soap" uri="http://schemas.xmlsoap.org/soap/envelope/" current="true" />
        <es:element name="Body">
          <es:element name="roxieechopersoninfoRequest">
            <es:namespace uri="urn:hpccsystems:ecl:roxieechopersoninfo" current="true" />
            <es:element name="roxieechopersoninforequest">
              <es:element name="Row">
                <es:element name="Name">
                  <es:set-value target="First" value="'lll'"/>
                  <es:set-value target="Last" value="'mmm'"/>
                  <es:element name="Aliases">
                    <es:set-value target="Alias" value="'nnn'"/>
                    <es:add-value target="Alias" value="'ooo'"/>
                  </es:element>
                </es:element>
              </es:element>
            </es:element>
          </es:element>
        </es:element>
      </es:element>
    </es:content>
  </es:http-post-xml>
</es:synchronize>

<!-- now that the synchronize is done, use the results of the calls to external services -->
<es:element name="HttpPostStuff">
  <es:copy-of select="$roxiestuff1"/>
  <es:copy-of select="$roxiestuff2"/>
</es:element>
```

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/esdllib/esdl_script.cpp
+++ b/esp/esdllib/esdl_script.cpp
@@ -29,6 +29,21 @@
 #include <fxpp/FragmentedXmlPullParser.hpp>
 using namespace xpp;
 
+class OptionalCriticalBlock
+{
+    CriticalSection *crit = nullptr;
+public:
+    inline OptionalCriticalBlock(CriticalSection *c) : crit(c)
+    {
+        if (crit)
+            crit->enter();
+    }
+    inline ~OptionalCriticalBlock()
+    {
+        if (crit)
+            crit->leave();
+    }
+};
 
 IEsdlTransformOperation *createEsdlTransformOperation(IXmlPullParser &xpp, const StringBuffer &prefix, bool withVariables, bool ignoreCodingErrors, IEsdlFunctionRegister *functionRegister, bool canCreateFunctions);
 void createEsdlTransformOperations(IArrayOf<IEsdlTransformOperation> &operations, IXmlPullParser &xpp, const StringBuffer &prefix, bool withVariables, bool ignoreCodingErrors, IEsdlFunctionRegister *functionRegister);
@@ -109,6 +124,14 @@ public:
         m_tagname.set(stag.getLocalName());
         m_traceName.set(stag.getValue("trace"));
         m_ignoreCodingErrors = getStartTagValueBool(stag, "optional", false);
+    }
+    virtual bool process(IEsdlScriptContext * scriptContext, IXpathContext * targetContext, IXpathContext * sourceContext) override
+    {
+        return exec(nullptr, nullptr, scriptContext, targetContext, sourceContext);
+    }
+    virtual IInterface *prepareForAsync(IEsdlScriptContext * scriptContext, IXpathContext * targetContext, IXpathContext * sourceContext) override
+    {
+        return nullptr;
     }
 };
 
@@ -191,8 +214,10 @@ public:
     {
     }
 
-    virtual bool process(IEsdlScriptContext * scriptContext, IXpathContext * targetContext, IXpathContext * sourceContext) override
+    virtual bool exec(CriticalSection *crit, IInterface *preparedForAsync, IEsdlScriptContext * scriptContext, IXpathContext * targetContext, IXpathContext * sourceContext) override
     {
+        OptionalCriticalBlock block(crit);
+
         if (m_select)
             return sourceContext->addCompiledVariable(m_name, m_select);
         else if (m_children.length())
@@ -228,8 +253,10 @@ public:
 
     virtual ~CEsdlTransformOperationHttpContentXml(){}
 
-    bool process(IEsdlScriptContext * scriptContext, IXpathContext * targetContext, IXpathContext * sourceContext) override
+    virtual bool exec(CriticalSection *crit, IInterface *preparedForAsync, IEsdlScriptContext * scriptContext, IXpathContext * targetContext, IXpathContext * sourceContext) override
     {
+        OptionalCriticalBlock block(crit);
+
         CXpathContextLocation location(targetContext);
         targetContext->addElementToLocation("content");
         return processChildren(scriptContext, targetContext, sourceContext);
@@ -298,7 +325,7 @@ public:
 
     virtual ~CEsdlTransformOperationMySqlBindParmeter(){}
 
-    bool process(IEsdlScriptContext * scriptContext, IXpathContext * targetContext, IXpathContext * sourceContext) override
+    virtual bool exec(CriticalSection *crit, IInterface *preparedForAsync, IEsdlScriptContext * scriptContext, IXpathContext * targetContext, IXpathContext * sourceContext) override
     {
         return false;
     }
@@ -601,8 +628,14 @@ public:
         if (defaultValue && s.isEmpty())
             s.set(defaultValue);
     }
-    virtual bool process(IEsdlScriptContext * scriptContext, IXpathContext * targetContext, IXpathContext * sourceContext) override
+    virtual bool exec(CriticalSection *crit, IInterface *preparedForAsync, IEsdlScriptContext * scriptContext, IXpathContext * targetContext, IXpathContext * sourceContext) override
     {
+        //OperationMySqlCall is optimized to write directly into the document object
+        //  In future we can create a different version that is optimized to work more asynchronously
+        //  If we did, then inside a synchronous tag we might to have it optional which mode is used. For example efficiently streaming in data while
+        //  an HTTP call is being made to an external service may still work best in the current mode
+        OptionalCriticalBlock block(crit);
+
         StringBuffer section;
         getXpathStringValue(section, sourceContext, m_section, "temporaries");
 
@@ -682,8 +715,9 @@ public:
 
     virtual ~CEsdlTransformOperationHttpHeader(){}
 
-    bool process(IEsdlScriptContext * scriptContext, IXpathContext * targetContext, IXpathContext * sourceContext) override
+    virtual bool exec(CriticalSection *crit, IInterface *preparedForAsync, IEsdlScriptContext * scriptContext, IXpathContext * targetContext, IXpathContext * sourceContext) override
     {
+        OptionalCriticalBlock block(crit);
         return processHeader(scriptContext, targetContext, sourceContext, nullptr);
     }
 
@@ -718,6 +752,16 @@ public:
     }
 };
 
+class OperationStateHttpPostXml : public CInterfaceOf<IInterface> //plain CInterface doesn't actually give us our opaque IInterface pointer
+{
+public:
+    Owned<IProperties> headers = createProperties();
+    StringBuffer url;
+    StringBuffer content;
+    unsigned testDelay = 0;
+};
+
+
 class CEsdlTransformOperationHttpPostXml : public CEsdlTransformOperationBase
 {
 protected:
@@ -726,6 +770,7 @@ protected:
     Owned<ICompiledXpath> m_url;
     IArrayOf<IEsdlTransformOperationHttpHeader> m_headers;
     Owned<IEsdlTransformOperation> m_content;
+    Owned<ICompiledXpath> m_testDelay;
 
 public:
     CEsdlTransformOperationHttpPostXml(IXmlPullParser &xpp, StartTag &stag, const StringBuffer &prefix, IEsdlFunctionRegister *functionRegister) : CEsdlTransformOperationBase(xpp, stag, prefix)
@@ -742,6 +787,9 @@ public:
         if (isEmptyString(url))
             esdlOperationError(ESDL_SCRIPT_MissingOperationAttr, m_tagname, "without url", m_traceName, !m_ignoreCodingErrors);
         m_url.setown(compileXpath(url));
+        const char *msTestDelayStr = stag.getValue("test-delay");
+        if (!isEmptyString(msTestDelayStr))
+            m_testDelay.setown(compileXpath(msTestDelayStr));
 
         int type = 0;
         while((type = xpp.next()) != XmlPullParser::END_DOCUMENT)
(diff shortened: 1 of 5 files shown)


afishbeck:
@asselitx please take a look.  And if possible do a bit of regression testing to make sure I haven't broken any existing operations.

esp/esdllib/esdl_script.cpp
```diff
@@ -1562,6 +1664,59 @@ class CEsdlTransformOperationForEach : public CEsdlTransformOperationWithChildre
...
+                : preps(_preps), children(_children), scriptContext(_scriptContext), targetContext(_targetContext), sourceContext(_sourceContext)
+            {
+            }
+            void Do(unsigned i)
+            {
+                children.item(i).exec(&synchronizeCrit, preps.item(i), scriptContext, targetContext, sourceContext);
+            }
+        } afor(preps, m_children, scriptContext, targetContext, sourceContext);
+
+        afor.For(m_children.ordinality(), 10, false, false);
```
ghalliday, line 1705: max at once - 10 - should probably be configurable.

esp/esdllib/esdl_script.cpp
```diff
@@ -32,9 +32,27 @@ using namespace xpp;
 interface IEsdlTransformOperation : public IInterface
 {
     virtual bool process(IEsdlScriptContext * scriptContext, IXpathContext * targetContext, IXpathContext * sourceContext) = 0;
+    virtual IInterface *prepare(IEsdlScriptContext * scriptContext, IXpathContext * targetContext, IXpathContext * sourceContext) = 0;
+    virtual bool exec(CriticalSection *crit, IInterface *stateIn, IEsdlScriptContext * scriptContext, IXpathContext * targetContext, IXpathContext * sourceContext) = 0;
```
ghalliday, line 36: Is it possible to have something more descriptive/prescriptive for the second parameter, or is it really no restrictions?

esp/esdllib/esdl_script.cpp
```diff
@@ -771,111 +829,129 @@ class CEsdlTransformOperationHttpPostXml : public CEsdlTransformOperationBase
...
-    {
-        VStringBuffer xpath("/esdl_script_context/%s/%s", m_section.str(), m_name.str());
+        if (preparedState.testDelay)
+            MilliSleep(preparedState.testDelay);
+
+        //No need to synchronize until we hit this point and start updating the scriptContext / xml document
+        OptionalCriticalBlock block(crit);
+
         CXpathContextLocation location(targetContext);
+        VStringBuffer xpath("/esdl_script_context/%s/%s", m_section.str(), m_name.str());
```
ghalliday, line 927: trivial: Could be before the critical block

esp/esdllib/esdl_script.cpp
```diff
@@ -771,111 +829,129 @@ class CEsdlTransformOperationHttpPostXml : public CEsdlTransformOperationBase
...
+                targetContext->addElementToLocation("content");
+                targetContext->addXmlContent(response.str());
+            }
+            else
+            {
+                targetContext->ensureSetValue("text", response.str(), true);
+            }
+        }
 
         sourceContext->addXpathVariable(m_name, xpath);
```
ghalliday, line 956: Unrelated, but it seems strange that the source context is updated.

ghalliday (commented):
@afishbeck a few minor comments.

I think the code will do what you want.  I can think of a couple of other ways of structuring it with advantages/disadvantages to each.
One general comment is it isn't completely clear what the critical section is protecting.  I assume it is the source and target contexts.  It would be worth clarifying what is protected.

esp/esdllib/esdl_script.cpp
```diff
@@ -771,111 +829,129 @@ class CEsdlTransformOperationHttpPostXml : public CEsdlTransformOperationBase
...
-        if (!content)
-            return;
+        Owned<OperationStateHttpPostXml> preparedState = new OperationStateHttpPostXml;
         CXpathContextLocation location(targetContext);
-        targetContext->addElementToLocation("response");
+        VStringBuffer xpath("/esdl_script_context/%s/%s", m_section.str(), m_name.str());
+        targetContext->ensureLocation(xpath, true);
+        if (m_url)
+            sourceContext->evaluateAsString(m_url, preparedState->url);
+        if (m_testDelay)
```
timothyklemm, line 856: Would it make sense to make this conditional on being in a test mode? The script context knows to use printf output, which is currently a good indicator of being used for testing.

testing/unittests/esdltests.cpp
```diff
@@ -329,9 +330,12 @@ class ESDLTests : public CppUnit::TestFixture
 
     void runTransform(IEsdlScriptContext *scriptContext, const char *scriptXml, const char *srcSection, const char *tgtSection, const char *testname, int code)
     {
+        unsigned startTime = msTick();
         Owned<IEsdlCustomTransform> tf = createEsdlCustomTransform(scriptXml, nullptr);
 
         tf->processTransform(scriptContext, srcSection, tgtSection);
+        fprintf(stdout, "\nTest (%s) time ms(%d)\n", testname, msTick() - startTime);
```
timothyklemm, line 337: %d should be %u

esp/esdllib/esdl_script.cpp
```diff
@@ -1562,6 +1664,59 @@ class CEsdlTransformOperationForEach : public CEsdlTransformOperationWithChildre
     }
 };
 
+class CEsdlTransformOperationSynchronize : public CEsdlTransformOperationWithChildren
+{
+public:
+    CEsdlTransformOperationSynchronize(IXmlPullParser &xpp, StartTag &stag, const StringBuffer &prefix) : CEsdlTransformOperationWithChildren(xpp, stag, prefix, false, nullptr)
```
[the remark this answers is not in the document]
asselitx: If we were to be more restrictive I think using a function like `loadChooseChildren` or failing to load if unsupported children are present are what I'd prefer. But I'm concerned that being restrictive would limit the expressiveness or power of the operation in detrimental ways- we can't always know which ops may take a long time (a big for-each or calling another function) and requiring a single operation to be wrapped in a script tag doesn't seem right.

Perhaps a warning (even during binding creation?) and good documentation about the operation are the best approaches to limit what would be a poor scripting choice.

esp/esdllib/esdl_script.cpp
asselitx: Do we have a test mode, or are you saying to add one? I think there could be danger in leaving an attribute like this in code that goes to production. What would be the plumbing and work needed for a test mode? Is that a compile flag or something else?

asselitx (commented):
Regression tests show no difference between a deployment with and without this PR.

esp/esdllib/esdl_script.cpp
timothyklemm: We have an implied test mode. By default, trace output is routed through jlog functions but a script context property is set by the unit tests causing the output to be generated by printf. This property, which cannot be changed through scripting or configuration, implies that we are in a test mode when set.

My concern with the `testDelay` script property is that it can be used in production environments. If we're OK with that, it's probably misnamed. If we don't want it used outside of tests, we already have a reliable way to limit it.

esp/esdllib/esdl_script.cpp
afishbeck: Yes currently you could say sourceContext is badly named.  It is currently "sourceAndVariableContext"  or maybe "sourceAndCodeContext".

Variables, scopes, etc. are managed through that object, as well as the proper read only source context stuff.

I may refactor at some point to separate them, but not in this PR.

testing/unittests/esdltests.cpp
```diff
@@ -1857,6 +1862,235 @@ constexpr const char * result = R"!!(<soap:Envelope xmlns:soap="http://schemas.x
...
+        </es:CustomRequestTransform>
+        )!!";
+
+        constexpr const char *config1 = R"!!(<config>
+          <Transform>
+            <Param name='testcase' value="new features"/>
+          </Transform>
+        </config>)!!";
+
+            constexpr const char * result = R"!!(<root>
```
timothyklemm, line 1951: Indentation...

esp/esdllib/esdl_script.cpp
```diff
@@ -1562,6 +1667,64 @@ class CEsdlTransformOperationForEach : public CEsdlTransformOperationWithChildre
...
+                : preps(_preps), children(_children), scriptContext(_scriptContext), targetContext(_targetContext), sourceContext(_sourceContext)
+            {
+            }
+            void Do(unsigned i)
+            {
+                children.item(i).exec(&synchronizeCrit, preps.item(i), scriptContext, targetContext, sourceContext);
+            }
+        } afor(preps, m_children, scriptContext, targetContext, sourceContext);
+
+        unsigned maxAtOnce = m_maxAtOnce.get() ? (unsigned) sourceContext->evaluateAsNumber(m_maxAtOnce) : 5;
```
timothyklemm, line 1712: Why 5 instead of the original 10? I'm not suggesting 10 is better than 5; just wondering what prompted the change.

afishbeck: Once it was configurable I started thinking about how many threads a single transaction should take for parallel processing by default.  This keeps it simple.  When used properly it would probably be rare to exceed this anyway, but users can increase the value if they want.

testing/unittests/esdltests.cpp
afishbeck: Updated.

testing/unittests/esdltests.cpp
```diff
@@ -1857,6 +1862,235 @@ constexpr const char * result = R"!!(<soap:Envelope xmlns:soap="http://schemas.x
...
+                  <ID>GI101</ID>
+                  <ID>GI102</ID>
+               </FullName>
+            </Person>
+          </root>
+        )!!";
+
+        static constexpr const char * script = R"!!(<es:CustomRequestTransform xmlns:es="urn:hpcc:esdl:script" target="Person">
+            <es:variable name='value' select="'abc'"/>
+            <es:synchronize max-at-once="3">
```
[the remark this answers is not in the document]
afishbeck: It's normal behavior but does need documenting so people code accordingly.  This all sits on an AsyncFor.  The max-at-once means that of the X number of tasks max-at-once will run simultaneously.  When one completes another will start.  5 tasks, 4+1, 4+1, 4, 3, 2, 1, done.  The thing that needs to be documented is that things that take most advantage of synchronize while waiting for some resource like xml-http should PROBABLY be at the top of the list of tasks.  Best to suggest they always write their scripts that way.

timothyklemm (approved):

esp/esdllib/esdl_script.cpp
```diff
@@ -718,6 +752,23 @@ class CEsdlTransformOperationHttpHeader : public CEsdlTransformOperationWithoutC
     }
 };
 
+class OperationStateHttpPostXml : public CInterfaceOf<IInterface> //plain CInterface doesn't actually give us our opaque IInterface pointer
+{
+public:
+    OperationStateHttpPostXml()
```
ghalliday, line 758: trivial: constructors/destructors are redundant (and harmless)

ghalliday (approved):
I didn't spot any problems.

asselitx (approved):
Changes look good, no further questions or issues.

esp/esdllib/esdl_script.cpp
```diff
@@ -771,111 +829,129 @@ class CEsdlTransformOperationHttpPostXml : public CEsdlTransformOperationBase
...
+            sourceContext->evaluateAsString(m_url, preparedState->url);
+        if (m_testDelay)
+            preparedState->testDelay = (unsigned) sourceContext->evaluateAsNumber(m_testDelay);
 
-        Owned<IHttpClientContext> httpCtx = getHttpClientContext();
-        Owned<IHttpClient> httpclient = httpCtx->createHttpClient(NULL, url);
-        if (!httpclient)
-            return;
+        buildRequest(*preparedState, scriptContext, targetContext, sourceContext);
+        return (IInterface *)preparedState.getClear();
```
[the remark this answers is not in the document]
afishbeck: Fixed

esp/esdllib/esdl_script.cpp
```diff
@@ -771,111 +829,129 @@ class CEsdlTransformOperationHttpPostXml : public CEsdlTransformOperationBase
...
+    {
+        Owned<IInterface> state = prepare(scriptContext, targetContext, sourceContext);
+        return exec(nullptr, state, scriptContext, targetContext, sourceContext);
+    }
 
-            if (headers && !headers->hasProp("Accept"))
-                headers->setProp("Accept", "text/html, application/xml");
+    virtual bool exec(CriticalSection *crit, IInterface *stateIn, IEsdlScriptContext * scriptContext, IXpathContext * targetContext, IXpathContext * sourceContext) override
+    {
+        OperationStateHttpPostXml &preparedState = *(OperationStateHttpPostXml *)stateIn;
```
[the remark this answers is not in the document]
afishbeck: done.

esp/esdllib/esdl_script.cpp
afishbeck: done.

esp/esdllib/esdl_script.cpp
afishbeck: done.

esp/esdllib/esdl_script.cpp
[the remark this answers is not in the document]
afishbeck: done.

esp/esdllib/esdl_script.cpp
afishbeck: done.

afishbeck:
Squashed.

@ghalliday this PR should be ready to merge.

PR #16167, opened 2022-06-07
HPCC-27714 LDAPEnv generated admin usernames too long

RussWhitehead:
LDAP restricts usernames to 20 characters or less. Because the environment name
is part of the generated username, these names can easily be too long. Therefore
first try to create the username as is, then try truncating to 20 characters and
retrying.  This can fail if the same username exists elsewhere (where we cannot
search), so generate a numeric suffix (1-99) and add that to the generated username
So, we try HPCC_LongEnvironmentName, HPCC_LongEnvironment, HPCC_LongEnvironmen1,
HPCC_LongEnvironmen2, and so forth

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [x] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/services/ldapenvironment/ldapenvironmentService.cpp
+++ b/esp/services/ldapenvironment/ldapenvironmentService.cpp
@@ -178,6 +178,48 @@ void CldapenvironmentEx::createLDAPBaseDN(const char * baseDN, SecPermissionType
     }
 }
 
+bool CldapenvironmentEx::createUser(StringBuffer &userName, const char * prefix, const char * envName, const char * baseDN, const char * pwd, StringBuffer &notes)
+{
+#define MAX_USERNAME_LDAP_LEN 20
+    StringBuffer tmpUserName;
+    tmpUserName.appendf("%s%s", prefix, envName);
+    userName.set(tmpUserName);
+
+    Owned<ISecUser> user = secmgr->createUser(tmpUserName.str());
+    user->credentials().setPassword(pwd);
+
+    int idx = 0;
+    while (true)
+    {
+        try
+        {
+            DBGLOG("Creating user '%s' in '%s', len=%d", tmpUserName.str(), baseDN, tmpUserName.length());
+            secmgr->addUser(*user.get(), baseDN);
+            userName.set(tmpUserName.str());//success!
+            return true;
+        }
+        catch(...)
+        {
+            if (idx == 99)
+            {
+                notes.appendf("\nNon Fatal Error creating '%s'. Username too long, or already exists in target AD. Refer to LDAPEnvironment logfile for more information.", userName.str());
+                return false;
+            }
+            if (idx == 0 && tmpUserName.length() > MAX_USERNAME_LDAP_LEN)
+                tmpUserName.setLength(MAX_USERNAME_LDAP_LEN);//LDAP limits account names to 20 characters
+            else
+            {
+                char szIndex[3];//'1' through '99'
+                itoa(++idx, szIndex, 10);
+                tmpUserName.setLength(tmpUserName.length() - strlen(szIndex));
+                tmpUserName.append(szIndex);
+            }
+            user->setName(tmpUserName.str());
+        }
+    }
+    return false;
+}
+
 bool CldapenvironmentEx::onLDAPCreateEnvironment(IEspContext &context, IEspLDAPCreateEnvironmentRequest &req, IEspLDAPCreateEnvironmentResponse &resp)
 {
     try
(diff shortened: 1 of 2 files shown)


esp/services/ldapenvironment/ldapenvironmentService.cpp
```diff
@@ -178,6 +178,47 @@ void CldapenvironmentEx::createLDAPBaseDN(const char * baseDN, SecPermissionType
...
+                notes.appendf("\nNon Fatal Error creating '%s'", userName.str());
+                return false;
+            }
+            if (idx == 0 && tmpUserName.length() > MAX_USERNAME_LDAP_LEN)
+                tmpUserName.setLength(MAX_USERNAME_LDAP_LEN);//LDAP limits account names to 20 characters
+            else
+            {
+                char szIndex[3];//'1' through '99'
+                itoa(++idx, szIndex, 10);
+                strcpy((char *)(tmpUserName.str() + tmpUserName.length() - strlen(szIndex)), szIndex);
```
afishbeck, line 214: You are making big assumptions about StringBuffer internals here.  Especially not getting out of sync with the internal length.

I would:
```
   char szIndex[3];//'1' through '99'
   itoa(++idx, szIndex, 10);
   tmpUserName.setLength(tmpUserName.length() - strlen(szIndex));
   tmpUserName.append(szIndex);
```
or
```
   tmpUserName.setLength(tmpUserName.length() - ((idx > 9) ? 2 : 1));
   tmpUserName.append(idx);
```

esp/services/ldapenvironment/ldapenvironmentService.cpp
```diff
@@ -178,6 +178,47 @@ void CldapenvironmentEx::createLDAPBaseDN(const char * baseDN, SecPermissionType
...
+        }
+        catch(...)
+        {
+            if (idx == 99)
+            {
+                notes.appendf("\nNon Fatal Error creating '%s'", userName.str());
+                return false;
+            }
+            if (idx == 0 && tmpUserName.length() > MAX_USERNAME_LDAP_LEN)
+                tmpUserName.setLength(MAX_USERNAME_LDAP_LEN);//LDAP limits account names to 20 characters
```
afishbeck, line 209: Is the advantage of doing this here, rather than from the very beginning, in case the size limit gets removed at the server at some point?  That is, longer names just might start working so respond to the error rather than shorten pro-actively?

esp/services/ldapenvironment/ldapenvironmentService.cpp
RussWhitehead: I am never changing the length of the StringBuffer .
I copy to "tmpUserName.str() + tmpUserName.length() - strlen(szIndex))" ,  for a length of "strlen(szIndex)",
so it's always (and only) the last 1 or 2 characters that get overwritten.

esp/services/ldapenvironment/ldapenvironmentService.cpp
```diff
@@ -178,6 +178,47 @@ void CldapenvironmentEx::createLDAPBaseDN(const char * baseDN, SecPermissionType
...
+            DBGLOG("Creating user '%s' in '%s', len=%d", tmpUserName.str(), baseDN, tmpUserName.length());
+            secmgr->addUser(*user.get(), baseDN);
+            userName.set(tmpUserName.str());//success!
+            return true;
+        }
+        catch(...)
+        {
+            if (idx == 99)
+            {
+                notes.appendf("\nNon Fatal Error creating '%s'", userName.str());
```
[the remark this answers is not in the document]
RussWhitehead: Added some more verbage

esp/services/ldapenvironment/ldapenvironmentService.cpp
RussWhitehead: From what I read, this limit could be a domain policy (which is likely the default domain policy), and they imply that this restriction is only on older versions of AD.  So I figure, why not at least give it a try...

esp/services/ldapenvironment/ldapenvironmentService.cpp
```diff
@@ -178,6 +178,47 @@ void CldapenvironmentEx::createLDAPBaseDN(const char * baseDN, SecPermissionType
...
+    while (true)
+    {
+        try
+        {
+            DBGLOG("Creating user '%s' in '%s', len=%d", tmpUserName.str(), baseDN, tmpUserName.length());
+            secmgr->addUser(*user.get(), baseDN);
+            userName.set(tmpUserName.str());//success!
+            return true;
+        }
+        catch(...)
```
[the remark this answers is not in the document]
RussWhitehead: Unfortunately the LDAP error is a generic "constraint violation", which could be anything. If the user can look in the logfile, they will get more information from the security manager log message.

esp/services/ldapenvironment/ldapenvironmentService.cpp
afishbeck: The fact that you have to cast away the const says something about going against the intent. But hmmm.  It will work.  And is preserving the length as is.  I'd be curious to see what @ghalliday thought of this.

RussWhitehead: @afishbeck @ghalliday   I went ahead and made the change

ghalliday: I agree the new code is much better, and old code was a bit dubious..

afishbeck (approved):

RussWhitehead:
@richardkchapman Please merge

PR #16185, opened 2022-06-09
HPCC-27752 Avoid adding duplicate permits to the send queue

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/roxie/udplib/udpsha.cpp
+++ b/roxie/udplib/udpsha.cpp
@@ -75,6 +75,7 @@ bool udpAssumeSequential = false;       // If a data packet with a later sequenc
 bool udpAdjustThreadPriorities = false; // Adjust the priorities for the UDP receiving and sending threads so they have priority.
                                         // Enabling tends to cause a big rise in context switches from other threads, so disabled by default
 bool udpAllowAsyncPermits = false;      // Allow requests to send more data to overtake the data packets that are being sent.
+bool udpRemoveDuplicatePermits = false;
 
 unsigned multicastTTL = 1;
 bool traceIBYTIfails = false;
--- a/roxie/udplib/udpsha.cpp
+++ b/roxie/udplib/udpsha.cpp
@@ -75,7 +75,7 @@ bool udpAssumeSequential = false;       // If a data packet with a later sequenc
 bool udpAdjustThreadPriorities = false; // Adjust the priorities for the UDP receiving and sending threads so they have priority.
                                         // Enabling tends to cause a big rise in context switches from other threads, so disabled by default
 bool udpAllowAsyncPermits = false;      // Allow requests to send more data to overtake the data packets that are being sent.
-bool udpRemoveDuplicatePermits = false;
+bool udpRemoveDuplicatePermits = true;
 
 unsigned multicastTTL = 1;
 bool traceIBYTIfails = false;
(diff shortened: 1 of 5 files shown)


ghalliday:
I checked the generated assembler and the function is cleanly inline.  I might try using auto and see if that also generates neat code - if so I'll push since it will be easier to read.

ghalliday:
Double checked and the code was the same - so force pushed an update to simplify the callback definition.

roxie/udplib/udptrs.cpp
```diff
@@ -1025,6 +1025,25 @@ class CSendManager : implements ISendManager, public CInterface
...
+            auto updateTrackingIfMatches = [&msg](UdpPermitToSendMsg &m)
+            {
+                if (!m.matches(msg))
+                    return false;
+                m.seen = msg.seen; // Update the queue entry with the most recent tracking information
+                return true;
+            };
+
+            //First check to see if there is a matching permit in the queue.  If so update the tracking information and return.
+            if (send_queue.walk(updateTrackingIfMatches))
```
richardkchapman, line 1037: Only concern would be if this is adding a lot of overhead. How big is the queue typically? This is called on a high priority thread...

richardkchapman: Also blocks the queue (with a critsec), so it's not just the size of the queue that might be a concern...

richardkchapman (commented):
Some things to think about

roxie/udplib/udptrs.cpp
ghalliday: I think it would be worth combining the update with the append - to avoid regaining the critical section.
I don't know how long the queue is normally.  It has a max capacity of 100, which is being filled up, but I think by duplicates.
I could add the new code on a switch so we could try with and without to compare (and see if my suspicion about duplicate items is true.)

roxie/udplib/udpsha.cpp
```diff
@@ -75,6 +75,7 @@ bool udpAssumeSequential = false;       // If a data packet with a later sequenc
 bool udpAdjustThreadPriorities = false; // Adjust the priorities for the UDP receiving and sending threads so they have priority.
                                         // Enabling tends to cause a big rise in context switches from other threads, so disabled by default
 bool udpAllowAsyncPermits = false;      // Allow requests to send more data to overtake the data packets that are being sent.
+bool udpRemoveDuplicatePermits = false;
```
richardkchapman, line 78: Should this default to true?

roxie/udplib/udpsha.hpp
```diff
@@ -126,65 +126,78 @@ class simple_queue
...
+
         free_space.signal();
     }
     
     unsigned in_queue() {
-        c_region.enter();
-        unsigned res = active_buffers;
-        c_region.leave();
-        return res;
+        CriticalBlock block(c_region);
```
richardkchapman, line 184: This (and the one below) feel like rather pointless critical blocks - would making active_buffers atomic be better?

richardkchapman (approved):

ghalliday:
@mckellyln I would value your opinion on merging this into 8.6.x/8.8.0.

mckellyln (approved):
I approve, looks good.

PR #16188, opened 2022-06-10
HPCC-27760 Avoid critical section in RecordPullerThread

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/roxie/ccd/ccdserver.cpp
+++ b/roxie/ccd/ccdserver.cpp
@@ -2299,7 +2299,6 @@ protected:
     Semaphore started;                      // MORE: GH->RKC I'm pretty sure this can be deleted, since handled by RestartableThread
     bool groupAtOnce, eog;
     std::atomic<bool> eof;
-    CriticalSection crit;
 
 public:
     RecordPullerThread(bool _groupAtOnce) 
@@ -2387,13 +2386,14 @@ public:
     {
         if (traceStartStop)
             DBGLOG("RecordPullerThread::stop");
-        {
-            CriticalBlock c(crit); // stop is called on our consumer's thread. We need to take care calling stop for our input to make sure it is not in mid-nextRow etc etc.
-            if (inputStream)
-                inputStream->stop();
-            eof = true;
-        }
+
+        //Force the reading thread to terminate
+        eof = true;
         RestartableThread::join();
+
+        //Reader thread is now guaranteed to have stopped, so it is safe to call stop()
+        if (inputStream)
+            inputStream->stop();
     }
 
     void reset()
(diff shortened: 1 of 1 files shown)


roxie/ccd/ccdserver.cpp
```diff
@@ -2394,13 +2393,14 @@ class RecordPullerThread : public RestartableThread
...
             DBGLOG("RecordPullerThread::stop");
-        {
-            CriticalBlock c(crit); // stop is called on our consumer's thread. We need to take care calling stop for our input to make sure it is not in mid-nextRow etc etc.
-            if (inputStream)
-                inputStream->stop();
-            eof = true;
-        }
+
+        //Force the reading thread to terminate
+        eof = true;
```
richardkchapman, line 2398: Not sure this will stop as quickly as calling stop() would - for example it's still going to wait for upstream activities that may have long timeouts...

richardkchapman (changes requested):
I'm not convinced that this won't have negative side effects if, for example, the puller thread is pulling from a SOAPCALL with a long timeout. You need to interrupt the upstream, not just set eof to true.

roxie/ccd/ccdserver.cpp
ghalliday: I think the behaviour is very similar - but a fix would make it better.  There are two cases:
i) other thread is not inside nextRow().
This will set eof to true, which means next time nextRow() is called it will immediately return.  The join will then succeed, and stop will then be called.
Previously it would have grabbed the critical section, called stop, and set the flag.  The next time nextRow() is called it will immediately return.  The join will then succeed.   It may be slightly worse performance because the stop() will not overlap with any processing in the other thread.
ii) other thread is inside nextRow()
Now it will set eof, wait for the thread to terminate then call stop().  Previously it would have waited for the row to be processed, then set eof, called stop() and waitied for the other thread to join.  Again stop() processing will not be overlapped with nextRow() processing.

I think the gain from not requiring a critical section every row (whether or not it is asynchronously closed) is likely to outweigh the cost.

The (minor) improvement that would be worth doing (for old and new code) would be to recheck eof after nextRow() is called - and avoid calling processGroup() etc..  (The append is probably fine.)

richardkchapman: If you are right, then the behaviour in case ii is suboptimal today - things may block for a while.

There may be other mechanisms (eg checkAbort) that make it ok in most cases

ghalliday: On the other hand if the reading thread has finished the new code will be faster (since there will be no need for the critical section).

mckellyln (approved):

richardkchapman (approved):

PR #16195, opened 2022-06-10
HPCC-27723 Fix compiler warning/errors in ubuntu 22.04 builds

RussWhitehead:
Fix various compile errors introduced with the new gcc compiler (11.2.0)
that comes with Ubuntu 22.04.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:


dali/base/dafdesc.cpp
```diff
@@ -672,7 +672,7 @@ protected: friend class CFileDescriptor;
     {
         partIndex = idx;
         ismulti = false;
-        if (!isEmptyPTree(pt)) {
```
richardkchapman, line 675: Not sure what the error this was fixing is, but this appears to make the code worse. Why was it needed?

ecl/hql/hqlgram2.cpp
```diff
@@ -12045,7 +12045,7 @@ IHqlExpression * HqlGram::createCheckMatchAttr(attribute & attr, type_t tc)
         type.setown(makeUtf8Type(UNKNOWN_LENGTH, NULL));
         break;
     default:
-        throwUnexpectedType(type);
```
richardkchapman, line 12048: This is not a desirable change. What was the error that provoked it?

ecl/hthor/hthor.cpp
```diff
@@ -6283,6 +6283,7 @@ const void * CHThorAllJoinActivity::nextRow()
                     return ret;
                 }
             }
+            break;
```
richardkchapman, line 6286: Interesting! Looks like this missing break actually (by luck) has no ill effect. But worth fixing!

esp/services/ws_workunits/ws_workunitsQuerySets.cpp
```diff
@@ -1172,7 +1172,7 @@ void retrieveQuerysetDetailsFromAlias(IEspContext &context, IPropertyTree *regis
         IPropertyTree *query = registry->queryPropTree(xpath);
         if (!query)
         {
-            UWARNLOG("No matching Query %s found for Alias %s", a->getId(), name);
```
richardkchapman, line 1175: This is fixing the symptom, not the bug. I assume the compiler warned that a may be (actually is always, I think) null. But the fix should be to take heed of the warning and trace the id properly, not just add code that makes the warning go away.

system/jlib/jptree.cpp
```diff
@@ -7442,6 +7442,7 @@ class CJSONReader : public CJSONReaderBase<X>, implements IPTreeReader
                 {
                     if ('\0' == *(name+6)) // #value
                     {
+                        assertex(retValue);
```
richardkchapman, line 7445: Again, this feels like the warning is highlighting a bug, and adding code to make it go away is a bad idea. The bug in this case being (as far as I can tell from a very quick glance at the code) that the error checking when parsing malformed JSON is inadequate.

richardkchapman (changes requested):

dali/base/dafdesc.cpp
jakesmith: Russ asked me about this one, the error was:
```
[ 61%] Building CXX object dali/base/CMakeFiles/dalibase.dir/danqs.cpp.o
In constructor ‘CPartDescriptor::CPartDescriptor(CFileDescriptorBase&, unsigned int, IPropertyTree*)’,
    inlined from ‘virtual void CFileDescriptor::setNumParts(unsigned int)’ at /mnt/e/HPCC/HPCC-Platform/dali/base/dafdesc.cpp:1977:75:
/mnt/e/HPCC/HPCC-Platform/dali/base/dafdesc.cpp:676:31: error: ‘this’ pointer is null [-Werror=nonnull]
  676 |             if (pt->getPropInt("@num",idx+1)-1!=idx)
      |                 ~~~~~~~~~~~~~~^~~~~~~~~~~~~~
/mnt/e/HPCC/HPCC-Platform/dali/base/dafdesc.cpp:678:29: error: ‘this’ pointer is null [-Werror=nonnull]
  678 |             overridename.set(pt->queryProp("@name"));
      |             ~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~
```

I suggested making it if (pt && isEmptyPTree(pt)) as a test.

But it appears it's a false positive from the compiler, because the !isEmptyTree should already checks pt for null

system/jlib/jptree.cpp
jakesmith: this was the other one Russ spoke about:
```
Consolidate compiler generated dependencies of target jlib
[  0%] Building CXX object system/jlib/CMakeFiles/jlib.dir/jptree.cpp.o
In member function ‘void CJSONReader<X>::readValueNotify(const char*, bool, StringBuffer*, bool*) [with X = CInstBufferReader]’,
inlined from ‘void CJSONReader<X>::readArray(const char*) [with X = CInstBufferReader]’ at /mnt/e/HPCC/HPCC-Platform/system/jlib/jptree.cpp:7481:32:
/mnt/e/HPCC/HPCC-Platform/system/jlib/jptree.cpp:7445:43: error: ‘this’ pointer is null [-Werror=nonnull]
7445 |                         retValue->swapWith(value);
|                         ~~~~~~^
In file included from /mnt/e/HPCC/HPCC-Platform/system/jlib/jlib.hpp:276,
from /mnt/e/HPCC/HPCC-Platform/system/jlib/jstats.h:22,
```

There is already a dbgassertex(retValue && isValueBinary); a few lines above, but _DEBUG only.
This function is called with retValue is null.

I suspect it's a code path that's never expect and the original assert is just a guard against the unexpected, but @afishbeck may know more.

dali/base/dafdesc.cpp
richardkchapman: Compiler can't know that isEmptyPTree() checks for null. I guess the issue is that the compiler DOES see some calls that pass in pt=null.

It is a false positive, but adding the check may be a reasonable approach in this case.

system/jlib/jptree.cpp
richardkchapman: There is a call that passes null for this parameter. Unless you can be sure that the json being parsed at the point that call is made CANNOT have #value in it , then this warning is highlighting inadequate error checking, and suppressing the warning is wrong.

dali/base/dafdesc.cpp
richardkchapman: Alternatively, adding two constructors - one for the case where NULL is passed and one fo the other case, might be a cleaner fix

ecl/hql/hqlgram2.cpp
RussWhitehead: @richardkchapman 
[ 56%] Built target hpccmetrics_prometheussink
In file included from /mnt/e/HPCC/HPCC-Platform/ecl/hql/hqlgram2.cpp:24:
/mnt/e/HPCC/HPCC-Platform/ecl/hql/hqlgram2.cpp: In member function ‘IHqlExpression* HqlGram::createCheckMatchAttr(attribute&, type_t)’:
/mnt/e/HPCC/HPCC-Platform/ecl/hql/hqlerrors.hpp:621:62: error: ‘this’ pointer is null [-Werror=nonnull]
  621 | #define throwUnexpectedType(type)   throw MakeStringException(ECODETEXT(HQLERR_UnexpectedType), type->queryTypeName(), sanitizeSourceFile(__FILE__), __LINE__)
/mnt/e/HPCC/HPCC-Platform/ecl/hql/hqlgram2.cpp:12048:9: note: in expansion of macro ‘throwUnexpectedType’
12048 |         throwUnexpectedType(type);
      |         ^~~~~~~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors
make[2]: *** [ecl/hql/CMakeFiles/hql.dir/build.make:256: ecl/hql/CMakeFiles/hql.dir/hqlgram2.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:11105: ecl/hql/CMakeFiles/hql.dir/all] Error 2
make: *** [Makefile:166: all] Error 2
russ@DESKTOP-58E0PSA:/mnt/e/HPCC/build$

ghalliday: I'm happy enough with that change - I don't think it can ever occur - so throwUnexpected() is appropriate.

ghalliday (commented):
I'm happy with the codegen change.

RussWhitehead:
Please review. Note that there are a few smoketest build failures, because of the AWS issue that @Michael-Gardner is working.
Also note, I had to add a new CMAKE flag, USE_OPENSSLV3 , because you have to #define the compatibility level before #including any openssl header file (including the one that sets the version).
https://manpages.debian.org/testing/openssl/openssl_user_macros.7ssl.en.html
NOTE: to be effective, a user defined macro must be defined before including any header file that depends on it, either in the compilation command ("cc -DMACRO=value") or by defining the macro in source before including any headers.

esp/services/ws_workunits/ws_workunitsQuerySets.cpp
```diff
@@ -1172,7 +1172,7 @@ void retrieveQuerysetDetailsFromAlias(IEspContext &context, IPropertyTree *regis
...
+            UWARNLOG("No matching Query %s found for Alias %s", a ? a->getId() : "unknown", name);
```
jakesmith, line 1175: 'a' is always going to be null the way it's currently coded.
It clears it on line 1170.

The warning should log the id. 
The code should either get the id before it's cleared,
or, a simpler/better fix IMO, is to change the getClear to a getLink, so it stays valid until end of scope.

system/httplib/httplib.h
```diff
@@ -194,6 +194,10 @@ using socket_t = int;
 #include <thread>
 
 #ifdef CPPHTTPLIB_OPENSSL_SUPPORT
+
+#if defined(_USE_OPENSSLV3)
+    #define OPENSSL_API_COMPAT 0x10100000L
+#endif
```
jakesmith, line 200: It doesn't look right, to redefine this (in these PR changes) 9 times, including in 2  @headers.
I'd expect it to be defined once and be included from a common header.
Also not sure where/what uses it, why it's defined (needs comment).

system/jlib/jptree.cpp
afishbeck: First a little background.  This #value is meant for specific ptree serialization scenarios, not as a syntax for users.  Where we do XML -> ptree -> JSON -> ptree. 

XML supports "mixed content" which is formatted like this:

`<a>extra content1<b>1</b>extra content2</a>`
Notice `<a>` has both a child and content of it's own.

Inside ptree those bits of extra content are merged together, so it turns into:

`<a><b>1</b>extra content1extra content2</a>`

JSON doesn't support any extra content.  There is nowhere to put it:

`{"a": {"b": 1}}`

So to preserve all the ptree data in different forms I added the #value concept... 

`{"a": {"b": 1, "#value": "extra content1extra content2"}}`

It's only meant to support what ptree would have written out.  Of course someone could try to hand code these values into their JSON.

The scenarios where retValue could be null here would likely to have been typed or generated some other way:

An array of #value (PTREE would have only output one entry:

`{"a": {"b": 1, "#value": ["extra", "content1", "extra", "content2"], "c": 2 }}
`
Or a rootless xml document that has #value at the root.

afishbeck: Bottom line, I think we can just add a check before using retValue.

`
if (retValue)
  retValue->swapWith(value);
`
Even if we wanted to support 
`{"a": {"b": 1, "#value": ["extra", "content1", "extra", "content2"], "c": 2 }}`
which shouldn't be output by toJSON(), it would be outside the scope of this PR.

esp/services/ws_workunits/ws_workunitsQuerySets.cpp
RussWhitehead: Opened https://track.hpccsystems.com/browse/HPCC-27812 for closer scrutiny of my "fix"

RussWhitehead:
@jakesmith @richardkchapman @ghalliday @afishbeck @Michael-Gardner Please (re)review

ghalliday (commented):
I suspect many/all of these fixes should target 8.6.x (or earlier?) otherwise once developers upgrade to 22.04 they won't be able to build old versions of the system.

system/httplib/httplib.h
```diff
@@ -3620,6 +3624,7 @@ template <typename... Args>
 inline ssize_t Stream::write_format(const char *fmt, const Args &... args) {
   const auto bufsiz = 2048;
   std::array<char, bufsiz> buf;
+  buf[0]=0;
```
jakesmith, line 3627: question: what was the error on this one?

as far as I can see, it should not be necessary, the sprintf's will null terminate.

If it's a false positive and this is a workaround to keep the compiler happy only,
needs a comment to say what the issue is.

system/httplib/httplib.h
```diff
@@ -4601,6 +4606,7 @@ inline void ClientImpl::close_socket(Socket &socket,
 
 inline bool ClientImpl::read_response_line(Stream &strm, Response &res) {
   std::array<char, 2048> buf;
+  buf[0] = 0;
```
jakesmith, line 4609: question: similar to above, what was the error here / necessity to add this?

system/httplib/httplib.h
RussWhitehead: @jakesmith In both cases the error was this. I will add comments

/mnt/e/HPCC/HPCC-Platform/system/httplib/httplib.h:3629:42: error: ‘buf’ may be used uninitialized [-Werror=maybe-uninitialized]
3629 |   auto sn = snprintf(buf.data(), buf.size() - 1, fmt, args...);

jakesmith: @ghalliday @richardkchapman  - is this a case of the compiler being overzealous?

In cases like this, passing an uninitialized variable/array to a function, which is supposed to fill it, it may not be able to tell that sprintf is a function designed to fill it,
but to have explicit initialization ahead of a function call that initializes, seems ott.

jakesmith: If we have to initialize, I think:
```
std::array<char, bufsiz> buf{""}
```

would be more elegant.

richardkchapman (approved):

jakesmith (approved):

PR #16225, opened 2022-06-20
HPCC-27803 fix release symbol issues

jakesmith:
## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/testing/unittests/CMakeLists.txt
+++ b/testing/unittests/CMakeLists.txt
@@ -30,6 +30,8 @@ endif()
 
 project( unittests )
 
+include(${HPCC_SOURCE_DIR}/esp/scm/espscm.cmake)
+
 set (    SRCS
          unittests.cpp
          remotetests.cpp
@@ -42,8 +44,9 @@ set (    SRCS
          txSummarytests.cpp
          accessmaptests.cpp
          fxpptests.cpp
-         ${HPCC_SOURCE_DIR}/esp/bindings/SOAP/xpp/fxpp/FragmentedXmlPullParser.cpp;
-         ${HPCC_SOURCE_DIR}/esp/bindings/SOAP/xpp/fxpp/FragmentedXmlAssistant.cpp;
+         ${HPCC_SOURCE_DIR}/esp/bindings/SOAP/xpp/fxpp/FragmentedXmlPullParser.cpp
+         ${HPCC_SOURCE_DIR}/esp/bindings/SOAP/xpp/fxpp/FragmentedXmlAssistant.cpp
+         ${CMAKE_BINARY_DIR}/generated/ws_loggingservice_esp.cpp
     )
 
 if (NOT CONTAINERIZED)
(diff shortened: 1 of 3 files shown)


testing/unittests/CMakeLists.txt
```diff
@@ -42,8 +44,9 @@ set (    SRCS
          txSummarytests.cpp
          accessmaptests.cpp
          fxpptests.cpp
-         ${HPCC_SOURCE_DIR}/esp/bindings/SOAP/xpp/fxpp/FragmentedXmlPullParser.cpp;
-         ${HPCC_SOURCE_DIR}/esp/bindings/SOAP/xpp/fxpp/FragmentedXmlAssistant.cpp;
+         ${HPCC_SOURCE_DIR}/esp/bindings/SOAP/xpp/fxpp/FragmentedXmlPullParser.cpp
+         ${HPCC_SOURCE_DIR}/esp/bindings/SOAP/xpp/fxpp/FragmentedXmlAssistant.cpp
+         ${CMAKE_BINARY_DIR}/generated/ws_loggingservice_esp.cpp
```
jakesmith, line 49: @afishbeck - please review + see question re. this and generated export definition in JIRA.

ghalliday:
@afishbeck I am also getting a problem with:

ghalliday:
@afishbeck I am also getting a problem with:

/usr/bin/ld: /home/gavin/dev/hpcc/testing/unittests/loggingtests.cpp:166: undefined reference to `createGetTransactionSeedResponse'

which is missing a DECL_EXPORT (I'm not sure why)

afishbeck:
@ghalliday using those functions externally is rare.  In loggingtests.cpp Tim is calling the service binary interfaces directly.  Normal client code would first create the client interface and use that to make an HTTP call.  So the missing DECL_EXPORT used by extremely rare code like this was probably never noticed.

jakesmith:
> @ghalliday using those functions externally is rare. In loggingtests.cpp Tim is calling the service binary interfaces directly. Normal client code would first create the client interface and use that to make an HTTP call. So the missing DECL_EXPORT used by extremely rare code like this was probably never noticed.



> createGetTransactionSeedResponse

This is the same issue I hit (loggingtests.cpp missing symbol), but different symbol createUpdateLogResponse.
It was fixed by one of the changes in this PR, but adding ${CMAKE_BINARY_DIR}/generated/ws_loggingservice_esp.cpp to the project.
These symbols are in generated files (from .ecm's) , as per question in JIRA,
should all generate factory methods like this be generated with DECL_EXPORT ?
But,

testing/unittests/CMakeLists.txt
afishbeck: Regarding question in JIRA: "yes I think those factories should all have the visibility attributes added. Some of those functions are rarely used externally so probably slipped through the cracks. The most common external usage is to go through the client interface. But some of these testing and configuration uses go directly to some of the less common objects."

afishbeck:
As stated inline, yes, I think DECL_EXPORT should be used for all of those factory methods.  Otherwise changes look good.

afishbeck (approved):

afishbeck:
See https://github.com/hpcc-systems/HPCC-Platform/pull/16244

jakesmith:
@ghalliday - squashed

richardkchapman (approved):

ghalliday (approved):

PR #16265, opened 2022-06-24
HPCC-27856 Make Generated DECLSPECs explicit for EXPORT/IMPORT/LOCAL code

afishbeck:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:


afishbeck:
@ghalliday @jakesmith 

Questions:  
  1. Do the smoketest window builds build release?  I've actually struggled a bit to create some expected failures. 
  2. Do those builds build client only or some larger set of source?

This draft tries to make the 3 types of generated "<NAME>_API_<MODE>" DECLSPECS explicit.
1. LOCAL - generated source is directly included and doesn't need the factories to be exported
3. EXPORTS - generated source directly included, export the factories
4. IMPORTS - for users (clients) of the libraries, this is the default to make using libraries easiest

The hardest part is currently testing.  I'm slowly trying to get my own windows builds working in the background but struggling a bit.

ghalliday:
The windows problems should be independent of debug/release/relwithdebinfo.

I have merged some changes of Gordon's to 8.6.x which should make the windows build simpler

ghalliday (approved):
Looks like a good fix.

esp/logging/loggingmanager/CMakeLists.txt
```diff
@@ -52,7 +52,7 @@ include_directories (
     ./..
 )
 
-ADD_DEFINITIONS( -D_USRDLL -DLOGGINGCOMMON_EXPORTS -DLOGGINGMANAGER_EXPORTS )
+ADD_DEFINITIONS( -D_USRDLL -DLOGGINGCOMMON_EXPORTS -DLOGGINGMANAGER_EXPORTS -DWSLOGGING_API_LOCAL )
```
jakesmith, line 55: loggingmanager links go logginglib which incldues ws_loggingservice_esp.cpp (and is now exporting the symbols),
I think     ${ESPSCM_GENERATED_DIR}/ws_loggingservice_esp.cpp can now be removed from loggingmanager (and this -DWSLOGGING_API_LOCAL)

esp/services/ws_loggingservice/CMakeLists.txt
```diff
@@ -42,7 +42,7 @@ include_directories (
      ./..
 )
 
-ADD_DEFINITIONS( -D_USRDLL -DLOGGINGCOMMON_EXPORTS )
+ADD_DEFINITIONS( -D_USRDLL -DLOGGINGCOMMON_EXPORTS -DWSLOGGING_API_LOCAL )
```
jakesmith, line 45: maybe for another PR, but this could now link logginglib instead of this and compiling in ws_loggingservice_esp.cpp

jakesmith (commented):
@afishbeck 1 half issue I think - loggingmanager now explicitly imports ws_loggingservice_esp.cpp from logginglib and includes it too

esp/logging/loggingmanager/CMakeLists.txt
afishbeck: @jakesmith These various logging dependencies need weeding through.  In the case you mention loggingmanager seems to be using the constructor directly, rather than the factory.  And I think some of the logging stuff in general may actually do things like derive new classes from the generated ones (not sure where just needs to be worked out).  Anyway it should definitely all be cleaned up, but not in this PR... which Gavin just merged.

jakesmith: fair enough, main concern was if the exported symbols from logginglib, and the local syms included in same module within loggingmanager would clash in some way, but build was clean.

PR #16274, opened 2022-06-28
HPCC-27854 Fix missing Query Associated Files for imported WU

wangkx:
When importing ECL WU from a ZAP report in a cloud environment,
1. unzip the WU files into the import folder under the HPCC query folder.
The existing code uses the temp folder which will be reset after HPCC
is restarted in a cloud environment.
2. keep the ip attribute for WU Query Associated Files as localhost.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/common/workunit/workunit.cpp
+++ b/common/workunit/workunit.cpp
@@ -5403,7 +5403,7 @@ void CWorkUnitFactory::importWorkUnit(const char *zapReportFileName, const char 
             f->remove();
             return tree.getClear();
         }
-        void updateWUQueryAssociatedFilesAttrs(const char *localIP)
+        void updateWUQueryAssociatedFilesAttrs()
         {
             ICopyArrayOf<IPropertyTree> fileTreesToRemove;
             IPropertyTree *queryAssociatedTreeRoot = wuTree->queryPropTree("Query/Associated");
@@ -5417,7 +5417,11 @@ void CWorkUnitFactory::importWorkUnit(const char *zapReportFileName, const char 
                 fileNameWithPath.append(PATHSEPSTR).append(pathTail(fileName));
                 if (checkFileExists(fileNameWithPath)) //Check if the ZAP report contains this file.
                 {
-                    fileTree.setProp("@ip", localIP);
+#ifdef _CONTAINERIZED
+                    fileTree.setProp("@ip", "localhost");
+#else
+                    fileTree.setProp("@ip", GetCachedHostName());
+#endif
                     fileTree.setProp("@filename", fileNameWithPath);
                 }
                 else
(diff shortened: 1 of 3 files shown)


ghalliday (approved):
Change look reasonable

common/workunit/workunit.cpp
```diff
@@ -5417,7 +5417,9 @@ void CWorkUnitFactory::importWorkUnit(const char *zapReportFileName, const char
                 fileNameWithPath.append(PATHSEPSTR).append(pathTail(fileName));
                 if (checkFileExists(fileNameWithPath)) //Check if the ZAP report contains this file.
                 {
-                    fileTree.setProp("@ip", localIP);
+#ifndef _CONTAINERIZED
+                    fileTree.setProp("@ip", GetCachedHostName());
```
jakesmith, line 5421: question: perhaps this should set to "localhost" unconditionally in containerized mode?

It doesn't need setting if the zap report was exported from the cloud, where these will already be localhost, but if the zap report is from bare metal, than they will need resetting to localhost to be accessible.

Does that work, importing a zap report from bare-metal to cloud ? I can see it's usefulness

esp/services/ws_workunits/ws_workunitsService.cpp
```diff
@@ -4552,8 +4552,14 @@ int CWsWorkunitsSoapBindingEx::onStartUpload(IEspContext &ctx, CHttpRequest* req
                 throw MakeStringException(ECLWATCH_INVALID_INPUT, "Only one WU ZAP report is allowed.");
 
             VStringBuffer fileName("%s%s", zipFolder, fileNames.item(0));
+#ifdef _CONTAINERIZED
+            VStringBuffer importDir("%s%simport", wswService->getQueryDirectory(), PATHSEPSTR);
```
jakesmith, line 4556: future: all dir. creation of this sort (inside importWorkUnit) should ensure it creates a truly unique dir, by using something like mkdtemp. In theory I could import >1 workunit with same name (but were different, e.g. from diff. envs) that would clash at the moment.

jakesmith (approved):
@wangkx - see comments, but can be merged as is.

common/workunit/workunit.cpp
wangkx: Now, the ip is set to "localhost" unconditionally in containerized mode. With this change, I did import a zap report from bare-metal to cloud. I can see the graph, results, Query Associated Files, etc. But, I need to fix the process log (in https://track.hpccsystems.com/browse/HPCC-27855?).

esp/services/ws_workunits/ws_workunitsService.cpp
wangkx: OK. 
In this PR, a subfolder under the import folder will be created. The subfolder is named using the WUID. The workunit is created when the ZAP report is imported and the WUID is unique.

jakesmith: ah ok, if wuid is new (and hence guaranteed to be unique), this is a non-issue

PR #16317, opened 2022-07-07
HPCC-27927 Basic copying with API

shamser:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dali/base/dadfs.cpp
+++ b/dali/base/dadfs.cpp
@@ -310,6 +310,10 @@ public:
             return str.append(": Failed to delete file: ").append(errstr);
         case DFSERR_RestrictedFileAccessDenied:
             return str.append(": Access to restricted file denied: ").append(errstr);
+        case DFSERR_EmptyStoragePlane:
+            return str.append(": Cluster does not have storage plane: ").append(errstr);
+        case DFSERR_MissingStoragePlane:
+            return str.append(": Storage plane missing: ").append(errstr);
         }
         return str.append("Unknown DFS Exception");
     }
@@ -2432,6 +2436,7 @@ public:
         dirty = false;
         return ret;
     }
+    virtual StringBuffer &getStorageFilePath(StringBuffer & path, unsigned copy) override;
 };
 
 // --------------------------------------------------------
@@ -4885,6 +4890,36 @@ public:
         }
     }
 };
+
+StringBuffer &CDistributedFilePart::getStorageFilePath(StringBuffer & path, unsigned copy)
+{
+    unsigned nc = copyClusterNum(copy, nullptr);
+    IClusterInfo &cluster = parent.clusters.item(nc);
+    const char *planeName = cluster.queryGroupName();
+    if (isEmptyString(planeName))
+    {
+        StringBuffer lname;
+        parent.getLogicalName(lname);
+        throw new CDFS_Exception(DFSERR_EmptyStoragePlane, lname.str());
+    }
+    // Need storage path(prefix) to work out path on storage plane
+    // (After removing prefix, the remaining path is the path on storage plane)
+    Owned<IStoragePlane> storagePlane = getDataStoragePlane(planeName, false);
+    if (!storagePlane)
+        throw new CDFS_Exception(DFSERR_MissingStoragePlane, planeName);
+
+    // Note: striping to be implemented as a separate jira as additional information
+    // will be needed in the configuration
+    path.append(parent.directory);
+    if (parent.hasDirPerPart())
+        addPathSepChar(path).append(partIndex+1); // part subdir 1 based
+    addPathSepChar(path);
+    getPartName(path);
+
+    unsigned prefixLength = strlen(storagePlane->queryPrefix());
+    path.remove(0, prefixLength);
+    return path;
+}
 
 static unsigned findSubFileOrd(const char *name)
 {
--- a/dali/base/dafdesc.cpp
+++ b/dali/base/dafdesc.cpp
@@ -32,6 +32,7 @@
 #include "dafdesc.hpp"
 #include "dadfs.hpp"
 #include "dameta.hpp"
+#include "jsecrets.hpp"
 
 #define INCLUDE_1_OF_1    // whether to use 1_of_1 for single part files
 
@@ -3685,6 +3686,34 @@ public:
     }
     virtual unsigned numDefaultSprayParts() const override { return xml->getPropInt("@defaultSprayParts", 1); }
     virtual bool queryDirPerPart() const override { return xml->getPropBool("@subDirPerFilePart", isContainerized()); } // default to dir. per part in containerized mode
+    virtual StorageType getStorageType() const override
+    {
+        const char * storageTypeStr = xml->queryProp("storageapi/@type");
+        if (isEmptyString(storageTypeStr))
+            return StorageType::StorageTypeUnknown;
+        if (streq(storageTypeStr, "azurefile"))
+            return StorageType::StorageTypeAzureFile;
+        if (streq(storageTypeStr, "azureblob"))
+            return StorageType::StorageTypeAzureBlob;
+        return StorageType::StorageTypeUnknown;
+    }
+    virtual const char * queryStorageApiAccount() const override
+    {
+        return xml->queryProp("storageapi/@account");
+    }
+    virtual const char * queryStorageContainer() const override
+    {
+        return xml->queryProp("storageapi/@container");
+    }
+    virtual StringBuffer & getSASToken(StringBuffer & token) const override
+    {
+        const char * secretName = xml->queryProp("storageapi/@secret");
+        if (isEmptyString(secretName))
+            return token.clear();  // return empty string if no secret name is specified
+        getSecretValue(token, "storage", secretName, "token", false);
+        return token.trimRight();
+    }
+
     virtual IStoragePlaneAlias *getAliasMatch(AccessMode desiredModes) const override
     {
         if (AccessMode::none == desiredModes)
--- a/dali/base/dafdesc.hpp
+++ b/dali/base/dafdesc.hpp
@@ -331,6 +331,13 @@ interface IStoragePlaneAlias: extends IInterface
     virtual const char *queryPrefix() const = 0 ;
 };
 
+enum StorageType
+{
+    StorageTypeUnknown,
+    StorageTypeAzureFile,
+    StorageTypeAzureBlob
+};
+
 //I'm not sure if this should be used in place of an IGroup, probably as system gradually changes
 interface IStoragePlane: extends IInterface
 {
@@ -340,6 +347,10 @@ interface IStoragePlane: extends IInterface
     virtual unsigned numDefaultSprayParts() const = 0 ;
     virtual bool queryDirPerPart() const = 0;
     virtual IStoragePlaneAlias *getAliasMatch(AccessMode desiredModes) const = 0;
+    virtual StorageType getStorageType() const = 0;
+    virtual const char * queryStorageApiAccount() const = 0;
+    virtual const char * queryStorageContainer() const = 0;
+    virtual StringBuffer & getSASToken(StringBuffer & token) const = 0;
 };
 
 IClusterInfo *createClusterInfo(const char *grpname,                  // NULL if roxie label set
--- a/dali/ft/dalift.cmake
+++ b/dali/ft/dalift.cmake
@@ -23,6 +23,11 @@
 
 project( dalift ) 
 
+set(CMAKE_CXX_STANDARD 14)
+find_package(azure-storage-blobs-cpp CONFIG REQUIRED)
+find_package(azure-storage-files-shares-cpp CONFIG REQUIRED)
+find_package(azure-storage-common-cpp CONFIG REQUIRED)
+
 set (    SRCS 
          daft.cpp 
          daftformat.cpp 
@@ -56,6 +61,9 @@ target_link_libraries ( dalift
          hrpc 
          dafsclient
          dalibase 
+         Azure::azure-storage-common
+         Azure::azure-storage-blobs
+         Azure::azure-storage-files-shares
     )
 
 if (NOT CONTAINERIZED)
--- a/dali/ft/filecopy.cpp
+++ b/dali/ft/filecopy.cpp
@@ -22,7 +22,9 @@
 #include "jlib.hpp"
 #include "jio.hpp"
 #include <math.h>
-
+#include <azure/core.hpp>
+#include <azure/storage/blobs.hpp>
+#include <azure/storage/files/shares.hpp>
 #include "jmutex.hpp"
 #include "jfile.hpp"
 #include "jsocket.hpp"
@@ -125,13 +127,13 @@ bool TargetLocation::canPull()
 
 //----------------------------------------------------------------------------
 
-FilePartInfo::FilePartInfo(const RemoteFilename & _filename)
+FilePartInfo::FilePartInfo(const RemoteFilename & _filename, unsigned _partNum) : partNum(_partNum)
 {
     filename.set(_filename);
     init();
 }
 
-FilePartInfo::FilePartInfo()
+FilePartInfo::FilePartInfo(unsigned _partNum) : partNum(_partNum)
 {
     init();
 }
@@ -443,7 +445,7 @@ bool FileTransferThread::launchFtSlaveCmd(const SocketEndpoint &ep)
         msg.clear().append(true);
         catchWriteBuffer(socket, msg);
         if (sprayer.options->getPropInt("@fail", 0)) // probably used as a debug testing option
-            throwError(DFTERR_CopyFailed);
+            throwError1(DFTERR_CopyFailed, "unknown reason");
     }
     return ok;
 }
@@ -1172,7 +1174,7 @@ void FileSprayer::calculateSplitPrefixPartition(const char * splitPrefix)
             remoteFilename.append(curBlob.filename);
             blobTarget.clear();
             blobTarget.setRemotePath(remoteFilename);
-            targets.append(* new TargetLocation(blobTarget));
+            targets.append(* new TargetLocation(blobTarget, target->partNum));
             partition.append(*new PartitionPoint(idx, targets.ordinality()-1, curBlob.offset, curBlob.length, curBlob.length));
         }
     }
@@ -1511,6 +1513,46 @@ void FileSprayer::cleanupRecovery()
 bool FileSprayer::usePushWholeOperation() const
 {
     return targets.item(0).filename.isUrl();
+}
+
+bool FileSprayer::useAPICopy()
+{
+    if (isContainerized()) // Future: support in bare-metal?
+    {
+        bool needCalcCRC = calcCRC();
+        if (needCalcCRC && !sources.item(0).hasCRC)
+            return false;
+
+        StringBuffer targetClusterName;
+        distributedTarget->getClusterName(0, targetClusterName); // change this to support logical files in > 1 clusters
+        Owned<IStoragePlane> targetStoragePlane = getDataStoragePlane(targetClusterName.str(), false);
+        if (targetStoragePlane==nullptr)
+            return false;
+        if (isEmptyString(targetStoragePlane->queryStorageApiAccount()))
+            return false;
+        if (isEmptyString(targetStoragePlane->queryStorageContainer()))
+            return false;
+
+        StringBuffer sourceClusterName;
+        distributedSource->getClusterName(0, sourceClusterName);
+        Owned<IStoragePlane> sourceStoragePlane = getDataStoragePlane(sourceClusterName.str(), false);
+        if (sourceStoragePlane==nullptr)
+            return false;
+        if (isEmptyString(sourceStoragePlane->queryStorageApiAccount()))
+            return false;
+        if (isEmptyString(sourceStoragePlane->queryStorageContainer()))
+            return false;
+
+        StorageType stTarget = targetStoragePlane->getStorageType();
+        StorageType stSource = sourceStoragePlane->getStorageType();
+        if ((stTarget==StorageType::StorageTypeAzureFile || stTarget==StorageType::StorageTypeAzureBlob)
+            && (stSource==StorageType::StorageTypeAzureFile || stSource==StorageType::StorageTypeAzureBlob))
+        {
+            LOG(MCdebugInfo, job, "Using Azure API to copy files");
+            return true;
+        }
+    }
+    return false;
 }
 
 bool FileSprayer::canRenameOutput() const
@@ -2008,7 +2050,7 @@ void FileSprayer::gatherMissingSourceTarget(IFileDescriptor * source)
             {
                 for (unsigned copy=0; copy < numCopies; copy++)
                 {
-                    FilePartInfo & next = * new FilePartInfo;
+                    FilePartInfo & next = * new FilePartInfo(idx1);
                     source->getFilename(idx1, copy, next.filename);
                     if (copy==0)
                         primparts.append(next);
@@ -2074,9 +2116,18 @@ void FileSprayer::gatherMissingSourceTarget(IFileDescriptor * source)
                     break;
                 }
             }
-            if (!done) {
-                sources.append(* new FilePartInfo(*((sourceCopy == 0)? &primary.filename : &secondary.filename)));
-                targets.append(* new TargetLocation(*dst));
+            if (!done)
+            {
+                if (sourceCopy == 0)
+                {
+                    sources.append(* new FilePartInfo(primary.filename, primary.partNum));
+                    targets.append(* new TargetLocation(*dst, primary.partNum));
+                }
+                else
+                {
+                    sources.append(* new FilePartInfo(secondary.filename, secondary.partNum));
+                    targets.append(* new TargetLocation(*dst, secondary.partNum));
+                }
                 sources.tos().size = (sourceCopy == 0) ? primarySize : secondarySize;
             }
         }
@@ -2517,7 +2568,7 @@ void FileSprayer::performTransfer()
         if (isAborting())
             throwError(DFTERR_CopyAborted);
         else
-            throwError(DFTERR_CopyFailed);
+            throwError1(DFTERR_CopyFailed, "unknown reason");
     }
 }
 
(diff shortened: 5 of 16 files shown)


system/jlib/jutil.cpp
```diff
@@ -2405,6 +2405,35 @@ jlib_decl StringBuffer& decodeUrlUseridPassword(StringBuffer& out, const char* i
...
+        {
+            case ' ': out.append("%20"); break;
+            case '+': out.append("%2B"); break;
+            case '\r': out.append("%0D"); break;
+            case '\n': out.append("%0A"); break;
+            case '\'': out.append("%27"); break;
+            case ',': out.append("%2C"); break;
+            case ':': out.append("%3A"); break;
+            case ';': out.append("%3B"); break;
+            default:
```
GordonSmith, line 2422: This list looks too short, see:  https://en.wikipedia.org/wiki/Percent-encoding#References

dockerfiles/platform-build/Dockerfile
```diff
@@ -71,7 +71,7 @@ WORKDIR /hpcc-dev/build
 
 ARG BUILD_TYPE=RelWithDebInfo
 ARG USE_CPPUNIT=1
-RUN cmake /hpcc-dev/HPCC-Platform -Wno-dev -DCONTAINERIZED=1 -DINCLUDE_PLUGINS=1 -DCMAKE_BUILD_TYPE=${BUILD_TYPE} -DUSE_PYTHON2=0 -DSUPPRESS_SPARK=1 -DUSE_CPPUNIT=${USE_CPPUNIT}
+RUN cmake /hpcc-dev/HPCC-Platform -Wno-dev -DCONTAINERIZED=1 -DINCLUDE_PLUGINS=1 -DCMAKE_BUILD_TYPE=${BUILD_TYPE} -DUSE_PYTHON2=0 -DSUPPRESS_SPARK=1 -DUSE_CPPUNIT=${USE_CPPUNIT} -DCMAKE_TOOLCHAIN_FILE=/hpcc-dev/HPCC-Platform/vcpkg/scripts/buildsystems/vcpkg.cmake -DVCPKG_OVERLAY_PORTS=../HPCC-Platform/vcpkg-overlays
```
GordonSmith, line 74: These changes are not needed as they are now included inside the main CMakeLists.txt

dali/ft/filecopy.cpp
```diff
@@ -2596,6 +2625,245 @@ void FileSprayer::pushParts()
...
+                        BlobClient::AbortCopyFromUri(props.CopyId.Value());
+                }
+                catch(const Azure::Core::RequestFailedException& e)
+                {
+                    IERRLOG("Error aborting copy: %s (code %d)", e.ReasonPhrase.c_str(), static_cast<int>(e.StatusCode));
+                }
+                curprogress.status = OutputProgress::StatusBegin;
+            }
+        }
+        */
```
GordonSmith, line 2862: Commented code?

cmake_modules/commonSetup.cmake
```diff
@@ -150,7 +150,7 @@ IF ("${COMMONSETUP_DONE}" STREQUAL "")
         set(PLATFORM OFF)
         set(INCLUDE_PLUGINS OFF)
         set(SIGN_MODULES OFF)
-        set(USE_OPTIONAL OFF) # Force failure if we can't find the plugin dependencies
+	set(USE_OPTIONAL OFF) # Force failure if we can't find the plugin dependencies
```
GordonSmith, line 153: Whitespace tab

GordonSmith (changes requested):

shamser:
@ghalliday I believe I can query the container name rather than have it provided in the yaml - I will be pushing some changes with this later today.

GordonSmith (approved):

dali/ft/filecopy.cpp
```diff
@@ -2596,6 +2625,190 @@ void FileSprayer::pushParts()
     performTransfer();
 }
 
+static StringBuffer & buildAzureStorageURIBase(IStoragePlane * storagePlane, StringBuffer & uri)
+{
+    const char *accountName = storagePlane->queryStorageApiAccount();
+    uri.appendf("https://%s", accountName);
```
ghalliday, line 2631: I think the storage can be striped over multiple accounts.  @jakesmith can you confirm?  (Something to revisit for the striping PR).

dali/ft/filecopy.cpp
```diff
@@ -2596,6 +2625,190 @@ void FileSprayer::pushParts()
...
+    uri.appendf(".core.windows.net/%s", storagePlane->queryStorageContainer());
+
+    return uri;
+}
+
+using namespace Azure::Storage;
+using namespace Azure::Storage::Files;
+using namespace Azure::Storage::Blobs;
+
+bool FileSprayer::transferUsingAPI()
```
ghalliday, line 2646: minor: Azure specific aspects of this should be encapsulated in some azure specific classes.  Ok to do as a separate PR, as long as the jira is created.

dali/ft/filecopy.cpp
```diff
@@ -2596,6 +2625,190 @@ void FileSprayer::pushParts()
...
+            OutputProgress & curprogress= progress.item(idx);
+
+            if (curprogress.status==OutputProgress::StatusCopied)
+                continue;
+
+            StringBuffer sourceURI(sourceBaseURI);
+            StringBuffer sourcePath;
+            cur.inputName.getPathInStoragePlane(sourcePath);
+            StringBuffer tmp;
+            sourceURI.appendf("/%s/%s",  encode_url(tmp, sourcePath.str()).str(), sourceSAStoken.str());
```
ghalliday, line 2699: Does the SAStoken need encoding?

dali/ft/filecopy.cpp
```diff
@@ -2596,6 +2625,190 @@ void FileSprayer::pushParts()
...
+    // 7) authenticate using managed identity
+
+    // Source
+    StringBuffer sourceClusterName, sourceBaseURI, sourceSAStoken;
+    distributedSource->getClusterName(0, sourceClusterName);
+    Owned<IStoragePlane> sourcePlane = getDataStoragePlane(sourceClusterName.str(), false);
+    buildAzureStorageURIBase(sourcePlane, sourceBaseURI);
+    sourcePlane->querySASToken(sourceSAStoken);
+    const char * sourcePrefix = sourcePlane->queryPrefix();
+    unsigned sourcePrefixLength = isEmptyString(sourcePrefix)?0:strlen(sourcePrefix);
```
ghalliday, line 2665: unused

dali/ft/filecopy.cpp
```diff
@@ -2596,6 +2625,190 @@ void FileSprayer::pushParts()
...
+    // 6) handle overwrite flag
+    // 7) authenticate using managed identity
+
+    // Source
+    StringBuffer sourceClusterName, sourceBaseURI, sourceSAStoken;
+    distributedSource->getClusterName(0, sourceClusterName);
+    Owned<IStoragePlane> sourcePlane = getDataStoragePlane(sourceClusterName.str(), false);
+    buildAzureStorageURIBase(sourcePlane, sourceBaseURI);
+    sourcePlane->querySASToken(sourceSAStoken);
+    const char * sourcePrefix = sourcePlane->queryPrefix();
```
ghalliday, line 2664: unused

dali/ft/filecopy.cpp
```diff
@@ -2596,6 +2625,190 @@ void FileSprayer::pushParts()
...
+
+    // Target
+    StringBuffer targetClusterName, targetBaseURI, targetSAStoken;
+    distributedTarget->getClusterName(0, targetClusterName);
+    Owned<IStoragePlane> targetPlane = getDataStoragePlane(targetClusterName.str(), false);
+    buildAzureStorageURIBase(targetPlane, targetBaseURI);
+    targetPlane->querySASToken(targetSAStoken);
+    const char * targetPrefix = targetPlane->queryPrefix();
+    unsigned targetPrefixLength = isEmptyString(targetPrefix)?0:strlen(targetPrefix);
+    auto targetType = targetPlane->getStorageType();
```
ghalliday, line 2675: style: restrict use of auto to cases where the type is awful/impossible to determine.

dali/ft/filecopy.cpp
```diff
@@ -2596,6 +2625,190 @@ void FileSprayer::pushParts()
...
+    const char * sourcePrefix = sourcePlane->queryPrefix();
+    unsigned sourcePrefixLength = isEmptyString(sourcePrefix)?0:strlen(sourcePrefix);
+
+    // Target
+    StringBuffer targetClusterName, targetBaseURI, targetSAStoken;
+    distributedTarget->getClusterName(0, targetClusterName);
+    Owned<IStoragePlane> targetPlane = getDataStoragePlane(targetClusterName.str(), false);
+    buildAzureStorageURIBase(targetPlane, targetBaseURI);
+    targetPlane->querySASToken(targetSAStoken);
+    const char * targetPrefix = targetPlane->queryPrefix();
```
ghalliday, line 2673: Also unused.

dali/ft/filecopy.cpp
```diff
@@ -2596,6 +2625,190 @@ void FileSprayer::pushParts()
...
+                    IERRLOG("transferUsingAPI .poll() failed: %s (code %d)", e.ReasonPhrase.c_str(), static_cast<int>(e.StatusCode));
+                    return false;
+                }
+                if (isAborting())
+                    break;
+            }
+        }
+        else
+        {
+            for(auto & b: blobCopyOp)
```
ghalliday, line 2772: The bulk of this is similar to the code above - should be shared if at all possible.
Probably cleaner as an array of copy objects with derived classes for the different cases.

dali/ft/filecopy.cpp
```diff
@@ -2596,6 +2625,190 @@ void FileSprayer::pushParts()
...
+            StringBuffer sourceURI(sourceBaseURI);
+            StringBuffer sourcePath;
+            cur.inputName.getPathInStoragePlane(sourcePath);
+            StringBuffer tmp;
+            sourceURI.appendf("/%s/%s",  encode_url(tmp, sourcePath.str()).str(), sourceSAStoken.str());
+
+            StringBuffer targetURI(targetBaseURI);
+            StringBuffer targetPath;
+            cur.outputName.getPathInStoragePlane(targetPath);
+            targetURI.appendf("/%s/%s", encode_url(tmp.clear(), targetPath.str()).str(), targetSAStoken.str());
```
ghalliday, line 2704: the source and target code is very similar, and looks like it should be commoned up.

dali/ft/filecopy.cpp
```diff
@@ -2596,6 +2625,190 @@ void FileSprayer::pushParts()
...
+            curprogress.status =  OutputProgress::StatusBegin;
+            curprogress.numReads++;
+            curprogress.numWrites++;
+            if (isAborting())
+                break;
+         }
+    }
+    catch (const Azure::Core::RequestFailedException& e)
+    {
+        IERRLOG("transferUsingAPI failed: %s (code %d)", e.ReasonPhrase.c_str(), static_cast<int>(e.StatusCode));
```
ghalliday, line 2725: What about copies that have been started?  Should it wait for completion and mark them as copied.  If not will it cause problems trying to overwrite files that already exist.

dali/ft/filecopy.cpp
```diff
@@ -2596,6 +2625,190 @@ void FileSprayer::pushParts()
...
+                    {
+                        curprogress.status =  OutputProgress::StatusActive;
+                        anyRemaining=true;
+                    }
+                    else
+                        return false; //Aborted or Failed -> return false to use older copy method
+                }
+                catch(const Azure::Core::RequestFailedException& e)
+                {
+                    IERRLOG("transferUsingAPI .poll() failed: %s (code %d)", e.ReasonPhrase.c_str(), static_cast<int>(e.StatusCode));
```
ghalliday, line 2800: Not sure this should abort before the other copies have completed - see similar comment above.

dali/ft/filecopy.cpp
```diff
@@ -2596,6 +2625,190 @@ void FileSprayer::pushParts()
...
+    Owned<IStoragePlane> targetPlane = getDataStoragePlane(targetClusterName.str(), false);
+    buildAzureStorageURIBase(targetPlane, targetBaseURI);
+    targetPlane->querySASToken(targetSAStoken);
+    const char * targetPrefix = targetPlane->queryPrefix();
+    unsigned targetPrefixLength = isEmptyString(targetPrefix)?0:strlen(targetPrefix);
+    auto targetType = targetPlane->getStorageType();
+
+    std::vector<Shares::StartFileCopyOperation> fileCopyOp;
+    std::vector<StartBlobCopyOperation> blobCopyOp;
+    std::vector<unsigned> copyingInProgress;
```
ghalliday, line 2679: I found this a bit confusing, and not a very helpful name.  I think it provides a mapping from the copy operation to the partition.  As commented elsewhere, cleaner would be to have a vector of copy objects, which could container a pointer to the partition object.

dali/ft/filecopy.cpp
```diff
@@ -2596,6 +2625,190 @@ void FileSprayer::pushParts()
...
+                        curprogress.status = OutputProgress::StatusCopied;
+                        curprogress.outputCRC = curprogress.inputCRC;
+                    }
+                    else if (status==Shares::Models::CopyStatus::Pending)
+                    {
+                        curprogress.status =  OutputProgress::StatusActive;
+                        anyRemaining=true;
+                    }
+                    else
+                        return false; //Aborted or Failed -> return false to use older copy method
```
ghalliday, line 2759: only return when status of all copies known

dali/ft/filecopy.cpp
```diff
@@ -2596,6 +2625,190 @@ void FileSprayer::pushParts()
...
+                {
+                    IERRLOG("transferUsingAPI .poll() failed: %s (code %d)", e.ReasonPhrase.c_str(), static_cast<int>(e.StatusCode));
+                    return false;
+                }
+                if (isAborting())
+                    break;
+            }
+        }
+        if (!anyRemaining||isAborting()) break;
+        sleep(1); //does the the sleep period need to be smarter?
```
ghalliday, line 2808: is 1 ms too short?

ghalliday (commented):
@shamser a few comments.  Some of the suggestions can be delayed...

jakesmith (commented):
@shamser - please see comments.

Looks good in general. My main gripe are the changes to RemoteFilename.
I don't think the mount prefix length / getPathInStoragePlane etc. should be pushed down to that generate endpoint/path container (and doesn't make sense for remote paths)

The prefix/mount knowledge would be better handled at the spray level (in transferUsingAPI) and/or via assistance from DFS where it would be logical for it to know about the planes/prefixes/storage/striping etc. (and also multiple clusters if/when we support it)

See comment here: https://github.com/hpcc-systems/HPCC-Platform/pull/16317/files#r921461644

dali/base/dafdesc.cpp
```diff
@@ -3653,6 +3661,47 @@ class CStoragePlaneInfo : public CInterfaceOf<IStoragePlane>
...
+            return getenv("AZURE_ACCOUNT_NAME");
+    }
+    virtual const char * queryStorageContainer() const override
+    {
+        return xml->hasProp("storageapi/@container") ? xml->queryProp("storageapi/@container"):"";
+    }
+    virtual const StringBuffer & querySASToken (StringBuffer & token) const override
+    {
+        const char *encodedToken = xml->queryProp("storageapi/@sastoken");
+        if (!isEmptyString(encodedToken))
```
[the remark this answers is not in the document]
shamser: Yes, I added for development.  However, I'm removing the ability for providing the token in the config as it's not secure and there is no longer any reason to keep this functionality.

dali/base/dafdesc.cpp
```diff
@@ -3653,6 +3661,47 @@ class CStoragePlaneInfo : public CInterfaceOf<IStoragePlane>
...
+        {
+            JBASE64_Decode(encodedToken, token);
+        }
+        else
+        {
+            const char * accountName = queryStorageApiAccount();
```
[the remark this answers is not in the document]
shamser: I don't think that providing the storage account for every storage plane should be mandatory, at least at this time.  May be revisit this question later.  I'll add a check for nullptr and empty string.

dali/base/dafdesc.cpp
```diff
@@ -3653,6 +3661,47 @@ class CStoragePlaneInfo : public CInterfaceOf<IStoragePlane>
...
+            StringBuffer secretName;
+            secretName.appendf("azure-%s",accountName);
+            StringBuffer encToken;
+            getSecretValue(encToken, "storage", secretName, "key", false);
+            encToken.trimRight();
+            JBASE64_Decode(encToken.str(), token);
```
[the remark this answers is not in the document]
shamser: The tokens are not purely alphanumeric.  Originally, I encoded because it was not being processed correctly by the templates.  However, it not needed for secrets, so I'll remove the code requiring base64 encoding.

dali/base/dafdesc.cpp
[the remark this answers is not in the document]
jakesmith: agree, but does it make sense for a plane to have a "storageapi" section defined, but no storageapi.account set?

dali/ft/filecopy.cpp
shamser: There is a separate jira for adding support for striped storage.  Please see the subtasks under https://track.hpccsystems.com/browse/HPCC-27456.

jakesmith (commented):
@shamser - not sure if you meant to refresh/request for review? 
There are quite a few comments/issues from Gavin + me remaining that aren't addressed/replied to.

shamser:
> @shamser - not sure if you meant to refresh/request for review? There are quite a few comments/issues from Gavin + me remaining that aren't addressed/replied to.

I've made quite a few changes.  I must have missed some comments.  I'm going through all the comments again.

helm/hpcc/values.schema.json
```diff
@@ -529,7 +529,8 @@
             "hosts": {},
             "umask": {},
             "numMounts": {},
-            "cost": {}
+            "cost": {},
+            "storageapi": {}
```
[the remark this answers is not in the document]
shamser: I've defined it in storagePlaneCommon now.  Does it need to be defined here too?

dali/base/dafdesc.cpp
```diff
@@ -3653,6 +3661,47 @@ class CStoragePlaneInfo : public CInterfaceOf<IStoragePlane>
...
+    virtual const char * queryStorageContainer() const override
+    {
+        return xml->hasProp("storageapi/@container") ? xml->queryProp("storageapi/@container"):"";
+    }
+    virtual const StringBuffer & querySASToken (StringBuffer & token) const override
+    {
+        const char *encodedToken = xml->queryProp("storageapi/@sastoken");
+        if (!isEmptyString(encodedToken))
+        {
+            JBASE64_Decode(encodedToken, token);
```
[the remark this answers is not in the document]
shamser: The token consist of characters such as question mark, semicolon and colon.  The token was cause the generated yaml to be rejected by k8s.  It seems the quotes around the token are removed by the use of toYaml().  Anyway, I'm removing the ability to provide tokens in the values as it's not really secure or a good practice to do so.

system/jlib/jfile.cpp
```diff
@@ -4561,6 +4562,16 @@ StringBuffer & RemoteFilename::getPath(StringBuffer & name) const
     else
         return getRemotePath(name);
 }
+StringBuffer  & RemoteFilename::getPathInStoragePlane(StringBuffer & name) const
+{
+    if (isLocal())
+        getLocalPath(name);
+    else
+        getRemotePath(name);
+    if (mountPathLength)
```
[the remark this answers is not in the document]
shamser: local paths are going to have mountPathLength set to 0.   It will only be set to a value for a remote mount. It would be clearer to rename this to mountPathLengthForRemoteMounts?  Any suggestions on this for shorter more descriptive name? @ghalliday @jakesmith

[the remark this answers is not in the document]
jakesmith: well, I think these changes to RemoteFilename should be removed, see main review comment(s) here : https://github.com/hpcc-systems/HPCC-Platform/pull/16317#pullrequestreview-1039425771 (and the one it points to)

dali/ft/dalift.cmake
```diff
@@ -23,6 +23,11 @@
 
 project( dalift ) 
 
+set(CMAKE_CXX_STANDARD 14)
```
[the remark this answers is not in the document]
shamser: yes, this is required otherwise the use of the Azure headers causes build errors.

dali/ft/filecopy.cpp
```diff
@@ -1513,6 +1515,33 @@ bool FileSprayer::usePushWholeOperation() const
     return targets.item(0).filename.isUrl();
 }
 
+bool FileSprayer::useAPICopy()
+{
+#ifdef _CONTAINERIZED
```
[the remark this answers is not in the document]
shamser: yes, done.

dali/dfu/dfuwu.cmake
```diff
@@ -26,6 +26,9 @@
 
 project( dfuwu ) 
 
+find_package(azure-storage-blobs-cpp CONFIG REQUIRED)
```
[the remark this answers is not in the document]
shamser: The copy functionality allows copying from blob storage.

dali/base/dadfs.cpp
```diff
@@ -4884,7 +4884,48 @@ protected: friend class CDistributedFilePart;
...
+        assertex(startsWith(path, planePrefix));
+        unsigned numStripedDevices = storagePlane->numDevices();
+        unsigned mountPathLength = planePrefix.length();
+        assertex(isPathSepChar(path.charAt(mountPathLength)));
+        ++mountPathLength;
+        if (numStripedDevices > 1)
+        {
+            assertex(path.charAt(mountPathLength)=='d');
+            ++mountPathLength;
+            unsigned stripeNum = calcStripeNumber(partNum, lfnHash, numStripedDevices);
```
ghalliday, line 4915: Simpler for this code to just skip digits - and not worry about how many there were.

dali/base/dafdesc.cpp
```diff
@@ -3653,6 +3654,36 @@ class CStoragePlaneInfo : public CInterfaceOf<IStoragePlane>
...
+            return StorageType::StorageTypeAzureFile;
+        else if (streq(storageTypeStr, "azureblob"))
+            return StorageType::StorageTypeAzureBlob;
+        return StorageType::StorageTypeUnknown;
+    }
+    virtual const char * queryStorageApiAccount() const override
+    {
+        const char * storageApiAccount = xml->queryProp("storageapi/@account");
+        if (isEmptyString(storageApiAccount))
+            storageApiAccount = getenv("AZURE_ACCOUNT_NAME");
```
ghalliday, line 3671: Minor: Environment variable shouldn't really have azure in if this is equally applicable to aws.  Is it just for testing?  Should it be required.

dali/base/dafdesc.hpp
```diff
@@ -327,6 +327,13 @@ interface IStoragePlaneAlias: extends IInterface
     virtual const char *queryPrefix() const = 0 ;
 };
 
+enum StorageType
```
ghalliday, line 330: Design: Not 100% sure this is a good idea (rather than returning the api name) because it does not lend itself to  supporting pluggable new APIs.  However that is unlikely to be a short term problem, so ok as it stands.

dali/ft/filecopy.cpp
```diff
@@ -2596,6 +2639,325 @@ void FileSprayer::pushParts()
     performTransfer();
 }
 
+using namespace Azure::Storage;
+using namespace Azure::Storage::Files;
+using namespace Azure::Storage::Blobs;
+
+enum ApiCopyStatus { Pending, Success, Failed, Aborted };
+interface IAPICopyClient : implements IInterface
```
ghalliday, line 2647: refactor: Not for this PR, but as part of the next stage can any azure api-related functionality be split out into a separate source file.  (Common api related functionality, like this interface, can stay here or in a common header.)

dali/ft/filecopy.cpp
```diff
@@ -2596,6 +2639,325 @@ void FileSprayer::pushParts()
...
+        }
+        uri.appendf(".core.windows.net/%s", container);
+
+        return true;
+    }
+    static void buildAzureStorageURI(IDistributedFile *distFile, unsigned partNum, const char * baseURI, const char *sasToken, StringBuffer & uri)
+    {
+        StringBuffer path;
+        bool ret = distFile->getUnmountedPath(partNum, path);
+        assertex(ret);
```
ghalliday, line 2698: possibly better to fallback to normal copying, or report a clearer error.

dali/ft/filecopy.cpp
```diff
@@ -2596,6 +2639,325 @@ void FileSprayer::pushParts()
...
+    {
+        try
+        {
+            fileClient.reset(new Shares::ShareFileClient(target));
+            fileCopyOp = std::make_unique<Shares::StartFileCopyOperation>(std::move(fileClient->StartCopy(source)));
+        }
+        catch (const Azure::Core::RequestFailedException& e)
+        {
+            IERRLOG("AzureFileClient startCopy failed: %s (code %d)", e.ReasonPhrase.c_str(), static_cast<int>(e.StatusCode));
+            status = ApiCopyStatus::Failed;
```
ghalliday, line 2719: the status will be overwritten by the following code to set it to pending.  Missing a return?

dali/ft/filecopy.cpp
```diff
@@ -2596,6 +2639,325 @@ void FileSprayer::pushParts()
...
+    ApiCopyStatus abortCopy()
+    {
+        try
+        {
+            fileCopyOp->Poll();
+            Shares::Models::FileProperties props = fileCopyOp->Value();
+            if (props.CopyStatus.HasValue() && props.CopyStatus.Value()!=Shares::Models::CopyStatus::Success)
+            {
+                if (fileCopyOp->HasValue() && fileCopyOp->Value().CopyId.HasValue())
+                    fileClient->AbortCopy(fileCopyOp->Value().CopyId.Value().c_str());
```
ghalliday, line 2758: Should you also Delete the blob?
Should you check the return (and log if the abort fails for some reason)

dali/ft/filecopy.cpp
```diff
@@ -2596,6 +2639,325 @@ void FileSprayer::pushParts()
...
+                continue;
+            OutputProgress & curProgress = progress.item(idx2);
+            CDateTime dateTime;
+            int64_t outputLength;
+            ApiCopyStatus status = apiClient->getProgress(dateTime, outputLength);
+            switch(status)
+            {
+            case ApiCopyStatus::Success:
+                curProgress.resultTime = dateTime;
+                curProgress.outputLength = curProgress.outputLength;
```
ghalliday, line 2921: Does nothing.  I think this should probably say

```
curProgress.outputLength = outputLength;
````

File: helm/hpcc/templates/dfuserver.yaml [hunk 36]
```diff
@@ -90,10 +90,12 @@ spec:
 {{- include "hpcc.addVolumeMounts" $commonCtx | indent 8 }}
 {{ include "hpcc.addConfigMapVolumeMount" . | indent 8 }}
 {{ include "hpcc.addCertificateVolumeMount" (dict "root" $ "name" .name "component" "dfuserver" "external" false) | indent 8 }}
+{{- include "hpcc.addSecretVolumeMounts" $commonCtx | nindent 8 }}
       volumes:
 {{ include "hpcc.addConfigMapVolume" . | indent 6 }}
 {{- include "hpcc.addVolumes" $commonCtx | indent 6 }}
 {{ include "hpcc.addCertificateVolume" (dict "root" $ "name" .name "component" "dfuserver" "external" false) | indent 6 }}
+{{- include "hpcc.addSecretVolumes" (dict "root" $ "secretsCategories" $secretsCategories) | indent 6 }}
```
ghalliday, line 98: simplify: I thin you can pass $commonCtx instead of constructing the dict

dali/base/dadfs.cpp
```diff
@@ -4884,7 +4884,48 @@ protected: friend class CDistributedFilePart;
...
+
+        StringBuffer planePrefix(storagePlane->queryPrefix());
+        Owned<IStoragePlaneAlias> alias = storagePlane->getAliasMatch(accessMode);
+        if (alias)
+        {
+            StringBuffer newPlanePrefix(alias->queryPrefix());
+            if (startsWith(path, newPlanePrefix))
+                planePrefix.swapWith(newPlanePrefix);
+        }
+        assertex(startsWith(path, planePrefix));
```
ghalliday, line 4906: question: Is it better to throw an exception or return false?   I suspect this should never happen so an exception is fine.

dali/base/dadfs.cpp
```diff
@@ -6875,6 +6916,10 @@ class CDistributedSuperFile: public CDistributedFileBase<IDistributedSuperFile>
             accessCost += tmpAccessCost;
         }
     }
+    virtual bool getUnmountedPath(unsigned partNum, StringBuffer & path) override
+    {
+        UNIMPLEMENTED;
```
ghalliday, line 6921: throwUnexpected() is probably more appropriate since it should never be called.

dali/ft/filecopy.cpp
```diff
@@ -1513,6 +1517,37 @@ bool FileSprayer::usePushWholeOperation() const
...
+        bool needCalcCRC = calcCRC();
+        if (needCalcCRC && !sources.item(0).hasCRC)
+            return false;
+
+        StringBuffer targetClusterName, sourceClusterName;
+        distributedTarget->getClusterName(0, targetClusterName); // change this to support logical files in > 1 clusters
+        Owned<IStoragePlane> targetStoragePlane = getDataStoragePlane(targetClusterName.str(), false);
+        if (targetStoragePlane==nullptr)
+            return false;
+        distributedSource->getClusterName(0, sourceClusterName);
```
ghalliday, line 1533: picky style:  A few extra newlines make it easier to read.  E.g. after each separate test.  (sourceClusterName can also be declared later on, closer to where it is used.)

dali/ft/filecopy.cpp
```diff
@@ -2596,6 +2639,325 @@ void FileSprayer::pushParts()
...
+
+    OwnedPointerArrayOf<IAPICopyClient> apiClients;
+    unsigned failCount = 0;
+    ForEachItemIn(idx1, partition)
+    {
+        PartitionPoint & cur = partition.item(idx1);
+        OutputProgress & curProgress = progress.item(idx1);
+        if (curProgress.status==OutputProgress::StatusCopied)
+        {
+            apiClients.append(nullptr);
```
ghalliday, line 2873: Worth adding a comment to indicate that null values are used to represent parts that have already been copied.

dali/ft/ftbase.cpp
```diff
@@ -113,9 +115,11 @@ void PartitionPoint::deserialize(MemoryBuffer & in)
     in.read(whichInput);
     in.read(inputOffset);
     in.read(inputLength);
+    in.read(inputPartNum); // TODO: check if this format change will cause backward compatibility problems
```
ghalliday, line 118: I suspect this will cause backward compatibility problems....

helm/hpcc/values.schema.json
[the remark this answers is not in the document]
ghalliday: It needs to be defined as an empty dict as you have here - because of the way allOf is implemented by helm.

ghalliday (commented):
@shamser looks cleaner.  A couple of issues, a few minor commments, and a likely compatibility issue.  Looks close though.

dali/ft/filecopy.cpp
shamser: I'll change to delete blob if abort called and report the return status of the .AbortCopy call.

dali/base/dadfs.cpp
```diff
@@ -4884,7 +4884,48 @@ protected: friend class CDistributedFilePart;
...
+                stripeNum /= 10;
+            }
+            assertex(isPathSepChar(path.charAt(mountPathLength)));
+            ++mountPathLength;
+        }
+        path.remove(0, mountPathLength);
+        return true;
+    }
+}
+;
```
jakesmith, line 4928: trivia: ; should be on line before

dali/base/dadfs.cpp
jakesmith: if functionality moved into IDistributedFilePart (CDistributedFilePart) this will be a non-issue (see prev. comment)

dali/base/dafdesc.cpp
```diff
@@ -3653,6 +3654,36 @@ class CStoragePlaneInfo : public CInterfaceOf<IStoragePlane>
...
+    {
+        return xml->queryProp("storageapi/@container");
+    }
+    virtual StringBuffer & getSASToken(StringBuffer & token) const override
+    {
+        const char * secretName = xml->queryProp("storageapi/@secret");
+        if (isEmptyString(secretName))
+            return token;  // return empty string if no secret name is specified
+        getSecretValue(token, "storage", secretName, "token", false);
+        return token.trimRight();
```
jakesmith, line 3684: question: is trimRight necessary?

dali/ft/filecopy.cpp
```diff
@@ -22,7 +22,9 @@
 #include "jlib.hpp"
 #include "jio.hpp"
 #include <math.h>
-
+#include <azure/core.hpp>
+#include <azure/storage/blobs.hpp>
+#include <azure/storage/files/shares.hpp>
```
jakesmith, line 27: related to @ghalliday 's question above - should this be on a conditional compile switch (USE_AZURE) ?

dali/ft/filecopy.cpp
```diff
@@ -1513,6 +1517,37 @@ bool FileSprayer::usePushWholeOperation() const
...
+            return false;
+        if (!targetStoragePlane->hasStorageApi()||!sourceStoragePlane->hasStorageApi())
+            return false;
+        StorageType stTarget = targetStoragePlane->getStorageType();
+        StorageType stSource = sourceStoragePlane->getStorageType();
+        if ((stTarget==StorageType::StorageTypeAzureFile || stTarget==StorageType::StorageTypeAzureBlob)
+             && (stSource==StorageType::StorageTypeAzureFile || stSource==StorageType::StorageTypeAzureBlob))
+        {
+                LOG(MCdebugInfo, job, "Using Azure API to copy files");
+                return true;
```
jakesmith, line 1545: indentation: looks like 8 spaces?

dali/ft/filecopy.cpp
```diff
@@ -2596,6 +2639,327 @@ void FileSprayer::pushParts()
...
+    virtual ApiCopyStatus getStatus() const
+    {
+        return status;
+    }
+    static bool buildAzureStorageURIBase(IStoragePlane * storagePlane, StringBuffer & uri)
+    {
+        const char *accountName = storagePlane->queryStorageApiAccount();
+        if (isEmptyString(accountName))
+        {
+            UWARNLOG("Unable to API copy: no account name provided for storage plane");
```
jakesmith, line 2670: should useAPICopy() have already have verified that it does have an a/c name by this stage? And if so, this should prob. be an exception.

dali/ft/filecopy.cpp
```diff
@@ -2596,6 +2639,327 @@ void FileSprayer::pushParts()
...
+        }
+        else
+        {
+            dbgassertex(storagePlane->getStorageType()==StorageType::StorageTypeAzureBlob);
+            uri.append(".blob");
+        }
+        const char *container = storagePlane->queryStorageContainer();
+        if (isEmptyString(container))
+        {
+            UWARNLOG("Unable to API copy: no container name provided for storage plane");
```
jakesmith, line 2687: see above comment re. useAPICopy() and exception

jakesmith (commented):
@shamser - looks good, a few comments/questions,
but the getUnmountedPath function would benefit from being further simplified - see https://github.com/hpcc-systems/HPCC-Platform/pull/16317/commits/2702046eca185f1ffbeca4a97527be9a3532d4d9#r928808542

dali/base/dafdesc.cpp
shamser: I've removed the use of AZURE_ACCOUNT_NAME env var .

dali/base/dafdesc.cpp
shamser: I added this after trailing spaces was causing problems.

dali/base/dafdesc.hpp
shamser: It's returning the storage type not the api name.  It needs to know the storage type to work out which API to use... will look into a better way of doing this later.

dali/ft/filecopy.cpp
shamser: On an offline discussion, Gavin mentioned that it is not necessary to handle optional components for vcpkg packages as should always be available. @ghalliday @jakesmith

dali/ft/filecopy.cpp
shamser: Will do: https://track.hpccsystems.com/browse/HPCC-28147

dali/ft/filecopy.cpp
shamser: I've moved the checks to useApiCopy() and removed the test here.  I don't think it's necessary to assertex(!isEmptyString(accountName)) here.

dali/ft/filecopy.cpp
shamser: I've moved this check to useApiCopy() and removed the test here.  I don't think it's necessary to assertex(!isEmptyString(accountName)) here.

system/jlib/jfile.hpp
```diff
@@ -412,7 +412,7 @@ class jlib_decl RemoteFilename
         // note that this is for a *local* file
 
 
-    void deserialize(MemoryBuffer & in);
+     void deserialize(MemoryBuffer & in);
```
jakesmith, line 415: trivial: extra indent space.

jakesmith (commented):
@shamser - please see comments and questions. Looks good and close, most comments are minor or trivial. I think the only potential major one is the issue re. serialization compatibility.

dali/ft/filecopy.cpp
```diff
@@ -2596,6 +2651,345 @@ void FileSprayer::pushParts()
...
+    virtual ApiCopyStatus startCopy(const char *target, const char * source) = 0;
+    virtual ApiCopyStatus getProgress(CDateTime & dateTime, int64_t & outputLength) = 0;
+    virtual ApiCopyStatus abortCopy() = 0;
+    virtual ApiCopyStatus getStatus() const = 0;
+};
+
+class CApiCopyClient : implements IAPICopyClient, public CInterface
+{
+protected:
+    ApiCopyStatus status;
```
ghalliday, line 2670: status should have a default initialiser.

dali/ft/filecopy.cpp
```diff
@@ -2596,6 +2651,345 @@ void FileSprayer::pushParts()
     performTransfer();
 }
 
+using namespace Azure::Storage;
+using namespace Azure::Storage::Files;
+using namespace Azure::Storage::Blobs;
+
+enum ApiCopyStatus { Pending, Success, Failed, Aborted };
```
ghalliday, line 2658: could use enum class (especially since always prefixing with ApiCopyStatus).

dali/ft/filecopy.cpp
```diff
@@ -2596,6 +2651,345 @@ void FileSprayer::pushParts()
...
+        status = getProgress(dateTime, outputLength);
+        switch (status)
+        {
+            case ApiCopyStatus::Pending:
+                try
+                {
+                    if (fileCopyOp->HasValue() && fileCopyOp->Value().CopyId.HasValue())
+                        fileClient->AbortCopy(fileCopyOp->Value().CopyId.Value().c_str());
+                    else // not sure that this can ever happen
+                        IERRLOG("Transfer using API .AbortCopy() failed: unexpected response from API");
```
ghalliday, line 2766: The error message does not really correspond to the test - more like a missing copyId.

dali/ft/filecopy.cpp
```diff
@@ -2596,6 +2651,345 @@ void FileSprayer::pushParts()
...
+                    status = ApiCopyStatus::Aborted;
```
ghalliday, line 2767: Does the file still need deleting?  The documentation suggests it leaves a zero length file.

dali/ft/filecopy.cpp
```diff
@@ -2596,6 +2651,345 @@ void FileSprayer::pushParts()
...
+                }
+                catch(const Azure::Core::RequestFailedException& e)
+                {
+                    IERRLOG("Abort copy operation failed: %s (code %d)", e.ReasonPhrase.c_str(), static_cast<int>(e.StatusCode));
+                    status = ApiCopyStatus::Failed;
+                }
+                break;
```
ghalliday, line 2863: Same comment about needing to delete if aborted.  (Or if the abort was too late?)

dali/ft/filecopy.cpp
```diff
@@ -2596,6 +2651,345 @@ void FileSprayer::pushParts()
...
+            StringBuffer sourceURI, targetURI;
+            CApiCopyClient::buildAzureStorageURI(distributedSource, cur.inputPartNum, sourceBaseURI.str(), sourceSAStoken.str(), sourceURI);
+            CApiCopyClient::buildAzureStorageURI(distributedTarget, cur.outputPartNum, targetBaseURI.str(), targetSAStoken.str(), targetURI);
+
+            Owned<IAPICopyClient> apiClient;
+            if (targetType==StorageTypeAzureFile)
+                apiClient.setown(new AzureFileClient());
+            else
+                apiClient.setown(new AzureBlobClient());
+            ApiCopyStatus status = apiClient->startCopy(targetURI.str(), sourceURI.str());
```
ghalliday, line 2923: minor: status is not used.  Should it be?

dali/ft/filecopy.cpp
```diff
@@ -2596,6 +2651,345 @@ void FileSprayer::pushParts()
...
+        {
+            IERRLOG("AzureFileClient startCopy failed: %s (code %d)", e.ReasonPhrase.c_str(), static_cast<int>(e.StatusCode));
+            status = ApiCopyStatus::Failed;
+            throw MakeStringException(DFTERR_CopyFailed, "%s", e.ReasonPhrase.c_str());
+        }
+        return status;
+    }
+    ApiCopyStatus getProgress(CDateTime & dateTime, int64_t & outputLength)
+    {
+        if (status!=ApiCopyStatus::Pending)
```
ghalliday, line 2726: This does not initialise dateTime or outputLength - but the call on line 2950 passes uninitialised values, and assigns them unconditionally.  If nothing else would be better to clear them here...

dali/ft/filecopy.cpp
```diff
@@ -2915,23 +2948,28 @@ bool FileSprayer::transferUsingAPI()
             CDateTime dateTime;
             int64_t outputLength;
             ApiCopyStatus status = apiClient->getProgress(dateTime, outputLength);
```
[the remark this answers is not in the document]
ghalliday: If status is failure this would then override the progress values with uninitialised data.

dali/ft/filecopy.cpp
```diff
@@ -3161,12 +3555,19 @@ void FileSprayer::spray()
...
     beforeTransfer();
-    if (usePushWholeOperation())
-        pushWholeParts();
-    else if (usePullOperation())
-        pullParts();
+    if (useAPICopy() && (replicate || copySource))
+    {
+        transferUsingAPI();
+    }
     else
```
ghalliday, line 3562: Should this fallback to using a normal copy?  As an option?

dali/ft/filecopy.cpp
```diff
@@ -2596,6 +2651,345 @@ void FileSprayer::pushParts()
...
+        ForEachItemIn(idx3, apiClients)
+        {
+            IAPICopyClient * apiClient = apiClients.item(idx3);
+            if (apiClient==nullptr || apiClient->getStatus()==ApiCopyStatus::Success)
+                continue;
+            apiClient->abortCopy();
+            OutputProgress & curProgress = progress.item(idx3);
+            curProgress.status = OutputProgress::StatusBegin;
+        }
+        throw MakeStringException(DFTERR_CopyFailed, DFTERR_CopyFailed_Text, failedReason.str());
```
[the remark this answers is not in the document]
ghalliday: If isAborting() I don't think failedReason will be set to anything sensible - you may want a different error message.

system/jlib/jutil.cpp
```diff
@@ -2405,6 +2405,22 @@ jlib_decl StringBuffer& decodeUrlUseridPassword(StringBuffer& out, const char* i
     return out;
 }
 
+jlib_decl StringBuffer& encode_url(StringBuffer& out, const char* in)
```
ghalliday, line 2408: trivial: functions normally use camelCase.

dali/ft/ftbase.cpp
```diff
@@ -119,15 +121,20 @@ void PartitionPoint::deserialize(MemoryBuffer & in)
     in.read(whichSlave);
     modifiedTime.deserialize(in);
     ::deserialize(in, fixedText);
+    if (in.remaining()>0)
+    {
```
ghalliday, line 125: I don't think these fields are actually needed on the slave - so it may be better to not serialize them, and have them default initialised to an invalid number.  Solves the serialization problem....

... see further comments above you can remove them completely from this class instead.

dali/ft/filecopy.cpp
```diff
@@ -2596,6 +2651,345 @@ void FileSprayer::pushParts()
...
+            PartitionPoint & cur = partition.item(idx1);
+            OutputProgress & curProgress = progress.item(idx1);
+            if (curProgress.status==OutputProgress::StatusCopied)
+            {
+                apiClients.append(nullptr); // represents parts already copied
+                continue;
+            }
+
+            StringBuffer sourceURI, targetURI;
+            CApiCopyClient::buildAzureStorageURI(distributedSource, cur.inputPartNum, sourceBaseURI.str(), sourceSAStoken.str(), sourceURI);
```
ghalliday, line 2915: You could have

`
inputPartNum = sources.item(cur.whichInput).partNum;
`
and avoid storing inputPartNum in the partition.

If it is used more than once you could add a

```
FileSprayer::getInputPartNum(unsigned whichInput)
{
   return sources.item(whichInput).partNum;
}

```

ghalliday (changes requested):
@shamser see comments.

dali/ft/filecopy.cpp
shamser: startCopy throws an exception if startCopy fails, so it's always returing PENDING.  I'm changing to remove the return value from startCopy.

dali/ft/filecopy.cpp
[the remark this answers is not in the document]
shamser: Parallel progress tracking: https://track.hpccsystems.com/browse/HPCC-28188

dali/ft/filecopy.cpp
```diff
@@ -2915,23 +2948,28 @@ bool FileSprayer::transferUsingAPI()
...
+                failedReason.append("unknown");
                 failCount++;
-                curProgress.status = OutputProgress::StatusBegin;
+                newProgress.status = OutputProgress::StatusBegin;
                 break;
             }
+            updateProgress(newProgress);
         }
         if (isAborting() || !anyRemaining || failCount)
             break;
```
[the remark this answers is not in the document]
shamser: Polling interval: https://track.hpccsystems.com/browse/HPCC-28188

dali/ft/filecopy.cpp
shamser: status has been moved within the try block so it is not .

dali/ft/filecopy.cpp
shamser: I've initialized dateTime and outputLength now.

dali/ft/filecopy.cpp
shamser: break removed so it will drop through to attempt file delete (see below)

dali/base/dadfs.cpp
```diff
@@ -4886,6 +4887,29 @@ protected: friend class CDistributedFilePart;
...
+    IClusterInfo &cluster = parent.clusters.item(nc);
+    const char *planeName = cluster.queryGroupName();
+    if (isEmptyString(planeName))
+        throw new CDFS_Exception(DFSERR_EmptyStoragePlane, ".....");
+    // if there's an associated storage plane, use its storage path (prefix)
+    Owned<IStoragePlane> storagePlane = getDataStoragePlane(planeName, false);
+    if (!storagePlane)
+        throw new CDFS_Exception(DFSERR_MissingStoragePlane, planeName);
+
+    path.append(parent.directory);
```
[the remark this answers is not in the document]
shamser: commented.

dali/ft/filecopy.cpp
```diff
@@ -2655,22 +2650,21 @@ using namespace Azure::Storage;
...
     virtual ApiCopyStatus abortCopy() = 0;
     virtual ApiCopyStatus getStatus() const = 0;
 };
 
-class CApiCopyClient : implements IAPICopyClient, public CInterface
+class CApiCopyClient : public CInterfaceOf<IAPICopyClient>
 {
 protected:
-    ApiCopyStatus status;
+    ApiCopyStatus status = ApiCopyStatus::Pending;
```
ghalliday, line 2665: This doesn't look right, because at construction the file copy hasn't started.  It should really have an extra state for not started.

dali/ft/filecopy.cpp
```diff
@@ -2763,20 +2759,21 @@ class AzureFileClient : public CApiCopyClient
...
+                        IERRLOG("AzureFileClient::AbortCopy() failed: CopyId is empty");
                     status = ApiCopyStatus::Aborted;
                 }
                 catch(const Azure::Core::RequestFailedException& e)
                 {
                     IERRLOG("Abort copy operation failed: %s (code %d)", e.ReasonPhrase.c_str(), static_cast<int>(e.StatusCode));
                     status = ApiCopyStatus::Failed;
                 }
-                break;
+                // drop-through to delete any file remnants
```
ghalliday, line 2770: Use // fallthrough or // fall through so that coverity doesn't report a warning.

dali/ft/filecopy.cpp
```diff
@@ -2950,9 +2954,12 @@ void FileSprayer::transferUsingAPI()
             ApiCopyStatus status = apiClient->getProgress(dateTime, outputLength);
             OutputProgress newProgress;
             newProgress.set(curProgress);
-            newProgress.resultTime = dateTime;
-            newProgress.outputLength = outputLength;
-            newProgress.inputLength = outputLength; // bytes read==bytes outputted
+            if (status != ApiCopyStatus::Failed && status!= ApiCopyStatus::Failed)
```
ghalliday, line 2957: so good you tested it twice?

ghalliday (commented):
@shamser a few comments from a scan through.  It looks like it is close to ready.  It may need a new PR because it is getting close to the comment limit that github can support.

dali/ft/filecopy.cpp
shamser: Had to make sure it was really sure! (I've changed one the test to check ::Aborted)

dali/base/dafdesc.cpp
jakesmith: right, because trimRight/clip only trims spaces, not ther whitespace chars..
Not a big deal, will work as is, but it begged the questions of why it was here, and won't acheive much by the sounds of it in case it might be useful.

dali/ft/filecopy.cpp
```diff
@@ -3161,12 +3565,31 @@ void FileSprayer::spray()
...
-    if (usePushWholeOperation())
-        pushWholeParts();
-    else if (usePullOperation())
-        pullParts();
-    else
-        pushParts();
+    bool copiedAlready = false;
+    if (useAPICopy() && (replicate || copySource))
+    {
+        try {
```
jakesmith, line 3571: trivial/style: curly brace on newline elsewhere

helm/hpcc/templates/dfuserver.yaml
```diff
@@ -91,10 +91,12 @@ spec:
 {{- include "hpcc.addVolumeMounts" $commonCtx | indent 8 }}
 {{ include "hpcc.addConfigMapVolumeMount" . | indent 8 }}
 {{ include "hpcc.addCertificateVolumeMount" (dict "root" $ "name" .name "component" "dfuserver" "external" false) | indent 8 }}
+{{- include "hpcc.addSecretVolumeMounts" $commonCtx | nindent 8 }}
       volumes:
 {{ include "hpcc.addConfigMapVolume" . | indent 6 }}
 {{- include "hpcc.addVolumes" $commonCtx | indent 6 }}
 {{ include "hpcc.addCertificateVolume" (dict "root" $ "name" .name "component" "dfuserver" "external" false) | indent 6 }}
+{{- include "hpcc.addSecretVolumes" $commonCtx | indent 6 }}
```
jakesmith, line 99: is {{- to chomp whitespace ok here? (e.g. if there are no secrets).
I may be, but notice elsewhere that addSecretVolumes is included there is no leading white space chomping.

jakesmith (commented):
@shamser - looks good afaics, I had 1 question: https://github.com/hpcc-systems/HPCC-Platform/pull/16317#pullrequestreview-1098890045

but it may well be fine as is.

helm/hpcc/templates/dfuserver.yaml
shamser: It was ok to remove the {{-.  There is an extra blank line but that's caused by hpcc.addCertificateVolume.

shamser:
@jakesmith Shall I squash?

ghalliday (approved):
@shamser a couple of questions, but assuming they don't change anything this looks ready to merge

dali/ft/filecopy.cpp
```diff
@@ -2596,6 +2647,360 @@ void FileSprayer::pushParts()
...
+{
+    std::unique_ptr<Shares::ShareFileClient> fileClient;
+    std::unique_ptr<Shares::StartFileCopyOperation> fileCopyOp;
+public:
+    virtual void startCopy(const char *target, const char * source) override
+    {
+        try
+        {
+            fileClient.reset(new Shares::ShareFileClient(target));
+            fileCopyOp = std::make_unique<Shares::StartFileCopyOperation>(std::move(fileClient->StartCopy(source)));
```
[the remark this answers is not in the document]
shamser: With fileClient, it is possible to use either reset or the = operator.

However, it is a little difference as "fileClient->StartCopy()" returns a reference to an object (not a pointer).  So, the assignment to the unique_pointer is a little different.   The make_unique<Shares::StartFileCopyOperation> constructor with the reference object together with std::move, allows the reference object to be moved & assigned to the unique_ptr.

PR #16322, opened 2022-07-07
HPCC-25241 Add support for running some services without DALI

afishbeck:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/platform/espcfg.ipp
+++ b/esp/platform/espcfg.ipp
@@ -267,8 +267,10 @@ public:
             throw(ie);
         }
         loadBindings();
-        if(useDali)
-            startEsdlMonitor();
+
+        //in spite of the "monitor" name, the ESDLMonitor currently loads both DALI and non-dali bindings and should always be run
+        //  in the future it would be nice to separate the two, and phase out DALI bindings in time, as we migrate to the cloud
+        startEsdlMonitor();
     }
 
     bool reSubscribeESPToDali();
@@ -299,8 +301,10 @@ public:
         unloadBindings();
         unloadServices();
         unloadProtocols();
-        if(useDali)
-           stopEsdlMonitor();
+
+        //in spite of the "monitor" name, the ESDLMonitor currently loads both DALI and non-dali bindings and should always be run
+        //  in the future it would be nice to separate the two, and phase out DALI bindings in time, as we migrate to the cloud
+        stopEsdlMonitor();
 
         serverstatus=NULL;
         
--- a/esp/services/espcontrol/ws_espcontrolservice.cpp
+++ b/esp/services/espcontrol/ws_espcontrolservice.cpp
@@ -112,18 +112,26 @@ bool CWSESPControlEx::handleDaliAttachmentRequest(bool attach, bool force, Strin
 
 bool CWSESPControlEx::onDetachFromDali(IEspContext& context, IEspDetachFromDaliRequest& req, IEspDetachFromDaliResponse& resp)
 {
+    bool status = false;
     StringBuffer message;
-    bool status = handleDaliAttachmentRequest(false, req.getForce_isNull() ? false : req.getForce(), message);
-    resp.setMessage(message.str());
+    if (isContainerized())
+        message.append("Detaching ESP from DALI at runtime is a baremetal only feature");
+    else
+        status = handleDaliAttachmentRequest(false, req.getForce_isNull() ? false : req.getForce(), message);
+    resp.setMessage(message);
     resp.setStatus(status ? 0 : -1);
     return status;
 }
 
 bool CWSESPControlEx::onAttachToDali(IEspContext& context, IEspAttachToDaliRequest& req, IEspAttachToDaliResponse& resp)
 {
+    bool status = false;
     StringBuffer message;
-    bool status = handleDaliAttachmentRequest(true, false, message);
-    resp.setMessage(message.str());
+    if (isContainerized())
+        message.append("Attaching ESP to DALI at runtime is a baremetal only feature");
+    else
+        status = handleDaliAttachmentRequest(true, false, message);
+    resp.setMessage(message);
     resp.setStatus(status ? 0 : -1);
     return status;
 }
--- a/helm/hpcc/templates/_helpers.tpl
+++ b/helm/hpcc/templates/_helpers.tpl
@@ -659,12 +659,19 @@ Add config arg for a component
 
 {{/*
 Add dali arg for a component
+Pass in dict with root, component (in case of error), optional (true if daliArg is optional)
 */}}
 {{- define "hpcc.daliArg" -}}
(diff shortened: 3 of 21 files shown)


afishbeck:
@asselitx here's the draft for running without DALI that I mentioned.  Can you try it?

asselitx (commented):
This works with test ESPs run both as an esdl application and in the legacy ESP style.

afishbeck:
@lsfonseca @asselitx please review.

esp/platform/espcfg.ipp
```diff
@@ -268,8 +268,10 @@ public:
             throw(ie);
         }
         loadBindings();
-        if(useDali)
-            startEsdlMonitor();
+
+        //in spite of the "monitor" name, the ESDLMonitor currently loads both DALI and non-dali binddings and should always be run
```
ghalliday, line 272: typo: bind**d**ings

esp/platform/espcfg.ipp
```diff
@@ -300,8 +302,10 @@ public:
         unloadBindings();
         unloadServices();
         unloadProtocols();
-        if(useDali)
-           stopEsdlMonitor();
+
+        //in spite of the "monitor" name, the ESDLMonitor currently loads both DALI and non-dali binddings and should always be run
```
ghalliday, line 306: ditto

esp/services/espcontrol/ws_espcontrolservice.cpp
```diff
@@ -112,18 +112,28 @@ bool CWSESPControlEx::handleDaliAttachmentRequest(bool attach, bool force, Strin
 
 bool CWSESPControlEx::onDetachFromDali(IEspContext& context, IEspDetachFromDaliRequest& req, IEspDetachFromDaliResponse& resp)
 {
+    bool status = false;
+#ifdef _CONTAINERIZED
+    const char *message = "Dynamically detaching ESP from DALI is a baremetal only feature";
```
ghalliday, line 117: I am curious why it is bare-metal only.  Also, could code this to remove the #ifdef

    StringBuffer message;
    if (isContainerized())
        message.append("Dynamically detaching ESP from DALI is a baremetal only feature");
    else
        status = handleDaliAttachmentRequest(false, req.getForce_isNull() ? false : req.getForce(), message);

helm/hpcc/templates/thor.yaml
```diff
@@ -319,6 +319,12 @@ data:
               job: "_HPCC_JOBNAME_"
 {{- end -}}
 
+{{- if empty $.Values.dali -}}
```
ghalliday, line 322: observation: Repeated lots of times.  I wonder if it is worth commoning up.  Probably not given the complication of helm functions.  Would it be better as a global consistency check?

ghalliday (commented):
@afishbeck looks sensible.  A couple of questions and tpos

helm/hpcc/templates/thor.yaml
afishbeck: I will look at both options.  Do you know if there is a way to guarantee that a global consistency check is run before any of the individual components templates?  I suppose I could create a global conistency check function that is called at the top of each template, but would check everything and fail if something is wrong... I THINK I can cache the fact that it was already run and skip after the first time by writing a value.  Would need to play with it a bit.

ghalliday: I wouldn't spend too long on it... the code is ok as it stands.

esp/services/espcontrol/ws_espcontrolservice.cpp
afishbeck: It's really by choice.  I want to deprecate the feature if I can.  With this change (especially in the cloud) there are better ways of managing these services without DALI and therefore no need to attach/detatch.

I'll change it not to use the #ifdef though.

afishbeck:
Updated based on comments.

asselitx (approved):
looks good

ghalliday (approved):

ghalliday:
I am happy to merge, but will ask @jakesmith to check if he can see any problems with the helm changes.

afishbeck:
Rebased onto latest 8.8.x to hopefully simplify @asselitx building and testing.

helm/hpcc/templates/eclagent.yaml
```diff
@@ -109,6 +109,12 @@ data:
 {{- end }}
 {{- end -}}
 
+{{- if empty $.Values.dali -}}
+  {{- if $.Values.eclagent -}}
+     {{- $_ := fail (printf "ECL agent requires missing DALI" ) -}}
```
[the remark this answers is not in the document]
jakesmith: and similar comment re. wording of "missing"

helm/hpcc/templates/eclscheduler.yaml
```diff
@@ -42,6 +42,11 @@ data:
 {{ include "hpcc.generateGlobalConfigMap" .root | indent 6 }}
 {{- end -}}
 
+{{- if empty $.Values.dali -}}
+  {{- if $.Values.eclscheduler -}}
+     {{- $_ := fail (printf "eclscheduler requires missing DALI" ) -}}
```
jakesmith, line 47: instead of performing this check for each case,
could you pass context (for the key name and name) and check inside/when daliArg is called?

i.e. I assume if a template gets as far as calling daliArg, it needs it.

helm/hpcc/templates/thor.yaml
jakesmith: see my comment/question above, re. could check go in daliArg helper with some passed in context (.)

jakesmith (commented):
@afishbeck - some minor comments and a question re. check in daliArg

afishbeck:
@jakesmith updated.  Moved the check into daliArg as suggested.  It is much cleaner.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -1081,7 +1086,7 @@ Pass in dict with root, me and dali if container in dali pod
           {{ include "hpcc.configArg" . }},
 {{- end }}
           "--service={{ .me.name }}",
-{{ include "hpcc.daliArg" .root | indent 10 }}
+{{ include "hpcc.daliArg" (dict "root" .root "component" "Sasha" "optional" false) | indent 10 }}
```
jakesmith, line 1089: personal: I might have abbreviated this and only passed in optional if true, and default to false, since all bar 1 case is false

jakesmith (approved):
@afishbeck - looks good, please squash

afishbeck:
squashed
@richardkchapman please merge

PR #16323, opened 2022-07-07
HPCC-27855 Report Component Log for imported WU from ZAP report

wangkx:
1. Set component log when importing a WU from a ZAP report.
2. Report imported component log in WU Info response.
3. Read/send imported component log in WU File response.
4. Change the code to use the createUniqueTempDirectory() as
the ZIP working folder.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/services/ws_workunits/ws_workunitsHelpers.cpp
+++ b/esp/services/ws_workunits/ws_workunitsHelpers.cpp
@@ -314,7 +314,15 @@ void WsWuInfo::setLogTimeRange(LogAccessConditions& logFetchOptions, unsigned wu
     logFetchOptions.setTimeRange(range);
 }
 
-#ifdef _CONTAINERIZED
+void WsWuInfo::sendImportedWorkunitComponentLog(const char* logFile, CHttpResponse* response)
+{
+    IFileIOStream* stream = createIOStreamFromFile(logFile, IFOread);
+    if (stream == nullptr)
+        throw makeStringExceptionV(ECLWATCH_INTERNAL_ERROR, "Cannot open stream for file %s", logFile);
+    response->setContent(stream);
+    response->send();
+}
+
 void WsWuInfo::sendWorkunitComponentLogs(IEspContext* context, CHttpResponse* response, WUComponentLogOptions& options)
 {
     if (!m_remoteLogAccessor)
@@ -411,7 +419,6 @@ unsigned WsWuInfo::sendComponentLogContent(IEspContext* context, IRemoteLogAcces
     }
     return totalRecsRead;
 }
-#endif
 
 void WsWuInfo::getSourceFiles(IEspECLWorkunit &info, unsigned long flags)
 {
@@ -763,7 +770,17 @@ void WsWuInfo::getHelpers(IEspECLWorkunit &info, unsigned long flags)
         if ((flags & WUINFO_IncludeHelpers))
         {
             Owned<IEspECLHelpFile> h = createECLHelpFile();
-            h->setName(File_ComponentLog);
+            StringBuffer componentLog;
+            getImportedComponentLog(componentLog);
+            if (componentLog.length() > 0) //from imported WU
+            {
+                h->setName(componentLog.str());
+                offset_t fileSize;
+                if (getFileSize(componentLog.str(), nullptr, fileSize))
+                    h->setFileSize(fileSize);
+            }
+            else
+                h->setName(File_ComponentLog);
             h->setType(File_ComponentLog);
             helpers.append(*h.getLink());
         }
@@ -843,6 +860,27 @@ void WsWuInfo::getHelpers(IEspECLWorkunit &info, unsigned long flags)
         info.setHelpersDesc(eMsg.str());
         e->Release();
     }
+}
+
+const char *WsWuInfo::getImportedComponentLog(StringBuffer &log)
+{
+    SCMStringBuffer s;
+    cw->getDebugValue("imported", s);
+    if (s.length() == 0)
+        return log.str();
+
+    StringArray list;
+    list.appendList(s.str(), ",");
+    ForEachItemIn(i, list)
+    {
+        const char *item = list.item(i);
+        if (hasPrefix(item, "ComponentLog=", false))
+        {
+            log.append(item + strlen("ComponentLog="));
+            break;
+        }
+    }
+    return log.str();
 }
 
 void WsWuInfo::getApplicationValues(IEspECLWorkunit &info, unsigned long flags)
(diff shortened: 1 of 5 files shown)


esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -845,6 +864,29 @@ void WsWuInfo::getHelpers(IEspECLWorkunit &info, unsigned long flags)
     }
 }
 
+#ifdef _CONTAINERIZED
+const char *WsWuInfo::getImportedComponentLog(StringBuffer &log)
+{
+    SCMStringBuffer s;
+    cw->getDebugValue("imported", s);
```
ghalliday, line 871: minor: getDebugValueBool() would be more usual.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -845,6 +864,29 @@ void WsWuInfo::getHelpers(IEspECLWorkunit &info, unsigned long flags)
     }
 }
 
+#ifdef _CONTAINERIZED
+const char *WsWuInfo::getImportedComponentLog(StringBuffer &log)
```
ghalliday, line 868: minor: From the places it is used returning bool seems more useful.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -845,6 +864,29 @@ void WsWuInfo::getHelpers(IEspECLWorkunit &info, unsigned long flags)
...
+    cw->getDebugValue("imported", s);
+    if (s.length() == 0)
+        return log.str();
+
+    StringArray list;
+    list.appendList(s.str(), ",");
+    ForEachItemIn(i, list)
+    {
+        const char *item = list.item(i);
+        if (item && hasPrefix(item, "ComponentLog=", false))
```
ghalliday, line 880: trivial: can item ever be null?

ghalliday (approved):
Code seems reasonable - I haven't checked the logic is correct.  A couple of minor comments.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -315,6 +315,15 @@ void WsWuInfo::setLogTimeRange(LogAccessConditions& logFetchOptions, unsigned wu
 }
 
 #ifdef _CONTAINERIZED
+void WsWuInfo::sendImportedWorkunitComponentLog(const char* logFile, CHttpResponse* response)
```
jakesmith, line 318: is there a reason this can't compile in non-containerized too?

In general we should compile in containerized code, so that it gets (compile) tested at least,
but also because things like remote logging, is something we that could equally apply to bare-metal too in the future (given a suitable configuration etc.)

jakesmith (commented):
@wangkx - a couple of comments/questions

esp/services/ws_workunits/ws_workunitsHelpers.cpp
wangkx: Should not. Will remove the check.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
wangkx: It is inside the _CONTAINERIZED because it is used for ComponentLog and, for now, the ComponentLog is only for containerized environment. I will move it out.

wangkx:

esp/services/ws_workunits/ws_workunitsHelpers.cpp
jakesmith: not new: but there is other code which is to do with remote logging, which really shouldn't be in a _CONTAINIZED only (for same reasons as above).
Unless there's some reason that it only compiles in containized, this may be a good time to remove other #ifdef _CONTAINERIZED ?

jakesmith (commented):
@wangkx - looks ok, but smoketest is failing with a linking error..

jakesmith (commented):
@wangkx - looks good I think.

I'd like to check with @afishbeck though, re. last commit, is there another way? (there were changes in this area - exporting generated functions etc)

Prior to this last commit it ws_sql was failing to link with:
```
/usr/bin/ld: CMakeFiles/ws_sql.dir/__/ws_workunits/ws_workunitsHelpers.cpp.o: in function `ws_workunits::CWsWuFileHelper::sendWUComponentLogStreaming(CHttpRequest*, CHttpResponse*)':
/home/jsmith/git3/HPCC-Platform/esp/services/ws_workunits/ws_workunitsHelpers.cpp:4162: undefined reference to `ws_workunits_req_resp::CWULogFileRequest::CWULogFileRequest(IEspContext*, char const*, IProperties*, MapStringTo<StringBuffer, StringBuffer&, MappingStringTo<StringBuffer, StringBuffer&> >*)'
```

ws_sql is linking ws_workunits which CWULogFileRequest is part of..

afishbeck:
@jakesmith Regarding using the new service shared object factory exports.. yes we could probably get it to work, but it would involve changing the helper code to use the factories instead of instantiating the classes directly, so perhaps that should be done in another PR?

jakesmith:
> @jakesmith Regarding using the new service shared object factory exports.. yes we could probably get it to work, but it would involve changing the helper code to use the factories instead of instantiating the classes directly, so perhaps that should be done in another PR?

Right, yes, I only looked quickly, but it looked like the generated objects were exportable.
Agree not something to do in this PR, I suspect it would clean/clarify boundaries a bit though if these helpers objects could remain in the library and be exported (or via factory methods).

jakesmith:
@wangkx - looks like this now needs rebasing, can you rebase and squash?

wangkx:
Will do it now. May take some time to rebuild and test...

wangkx:
@jakesmith I did rebase and squash.

jakesmith (approved):

wangkx:

richardkchapman:
It does not appear that @ghalliday comments have been addressed. Are you happy to merge, Gavin?

esp/services/ws_workunits/ws_workunitsHelpers.cpp
wangkx: The 'imported' is a comma separated string, not a Boolean.

ghalliday:
@richardkchapman I think it is fine to merge as it is.

PR #16343, opened 2022-07-12
HPCC-27965 Remove extra log message from initWULogReader()

wangkx:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/services/ws_logaccess/WsLogAccessService.cpp
+++ b/esp/services/ws_logaccess/WsLogAccessService.cpp
@@ -11,35 +11,15 @@ Cws_logaccessEx::~Cws_logaccessEx()
 bool Cws_logaccessEx::onGetLogAccessInfo(IEspContext &context, IEspGetLogAccessInfoRequest &req, IEspGetLogAccessInfoResponse &resp)
 {
     bool success = true;
-    if (m_remoteLogAccessor)
+    if (queryRemoteLogAccessor())
     {
-        resp.setRemoteLogManagerType(m_remoteLogAccessor->getRemoteLogAccessType());
-        resp.setRemoteLogManagerConnectionString(m_remoteLogAccessor->fetchConnectionStr());
+        resp.setRemoteLogManagerType(queryRemoteLogAccessor()->getRemoteLogAccessType());
+        resp.setRemoteLogManagerConnectionString(queryRemoteLogAccessor()->fetchConnectionStr());
     }
     else
         success = false;
 
     return success;
-}
-
-void Cws_logaccessEx::init(const IPropertyTree *cfg, const char *process, const char *service)
-{
-    LOG(MCdebugProgress,"WsLogAccessService loading remote log access plug-in...");
-
-    try
-    {
-        m_remoteLogAccessor.set(&queryRemoteLogAccessor());
-
-        if (m_remoteLogAccessor == nullptr)
-            LOG(MCerror,"WsLogAccessService could not load remote log access plugin!");
-    }
-    catch (IException * e)
-    {
-        StringBuffer msg;
-        e->errorMessage(msg);
-        LOG(MCoperatorWarning,"WsLogAccessService could not load remote log access plug-in: %s", msg.str());
-        e->Release();
-    }
 }
 
 LogAccessTimeRange requestedRangeToLARange(IConstTimeRange & reqrange)
@@ -195,7 +175,7 @@ ILogAccessFilter * buildBinaryLogFilter(IConstBinaryLogFilter * binaryfilter)
 
 bool Cws_logaccessEx::onGetLogs(IEspContext &context, IEspGetLogsRequest &req, IEspGetLogsResponse & resp)
 {
-    if (!m_remoteLogAccessor)
+    if (!queryRemoteLogAccessor())
         throw makeStringException(-1, "WsLogAccess: Remote Log Access plug-in not available!");
 
     double version = context.getClientVersion();
@@ -296,7 +276,7 @@ bool Cws_logaccessEx::onGetLogs(IEspContext &context, IEspGetLogsRequest &req, I
     }
     StringBuffer logcontent;
     LogQueryResultDetails LogQueryResultDetails;
-    m_remoteLogAccessor->fetchLog(LogQueryResultDetails, logFetchOptions, logcontent, logAccessFormatFromName(req.getFormat()));
+    queryRemoteLogAccessor()->fetchLog(LogQueryResultDetails, logFetchOptions, logcontent, logAccessFormatFromName(req.getFormat()));
 
     if (version >= 1.02)
     {
--- a/esp/services/ws_workunits/ws_workunitsHelpers.cpp
+++ b/esp/services/ws_workunits/ws_workunitsHelpers.cpp
@@ -209,7 +209,7 @@ WsWUExceptions::WsWUExceptions(IConstWorkUnit& wu): numerr(0), numwrn(0), numinf
 void WsWuInfo::readWorkunitComponentLogs(const char* outFile, unsigned maxLogRecords, const LogAccessReturnColsMode retColsMode,
                                          const LogAccessLogFormat logFormat, unsigned wuLogSearchTimeBuffSecs)
 {
-    if (!m_remoteLogAccessor)
+    if (!queryRemoteLogAccessor())
         throw makeStringException(ECLWATCH_LOGACCESS_UNAVAILABLE, "WsWuInfo: Remote Log Access plug-in not available!");
 
     if (isEmptyString(outFile))
@@ -228,7 +228,7 @@ void WsWuInfo::readWorkunitComponentLogs(const char* outFile, unsigned maxLogRec
     if (!outIOS)
         throw makeStringException(ECLWATCH_CANNOT_OPEN_FILE, "WsWuInfo: Could not create target log file!");
 
-    Owned<IRemoteLogAccessStream> logReader = m_remoteLogAccessor->getLogReader(logFetchOptions, logFormat);
+    Owned<IRemoteLogAccessStream> logReader = queryRemoteLogAccessor()->getLogReader(logFetchOptions, logFormat);
 
     StringBuffer logcontent;
     unsigned totalRecsRead = 0;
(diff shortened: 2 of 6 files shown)


esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -2124,18 +2124,22 @@ unsigned WsWuInfo::getResourceURLCount()
     return 0;
 }
 
+static bool logInitWULogReader = true;
 void WsWuInfo::initWULogReader()
 {
-    LOG(MCdebugProgress,"WsWuInfo loading remote log access plug-in...");
+    if (logInitWULogReader)
+        LOG(MCdebugProgress,"WsWuInfo loading remote log access plug-in...");
```
jakesmith, line 2131: This means that if the configuration changes (e.g. on a helm update) it may log that it failed, but subsequently succeed (silently) or the reverse..

I think it would be better to either remove the tracing altogether or add the logic inside the singleton method in jlog, but also spot when the config changes and update via a ConfigUpdateHook (example use case in eclccserver.cpp, line 1099)

rpastrana (commented):
@wangkx couple of comments

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -2124,18 +2124,22 @@ unsigned WsWuInfo::getResourceURLCount()
     return 0;
 }
 
+static bool logInitWULogReader = true;
 void WsWuInfo::initWULogReader()
 {
-    LOG(MCdebugProgress,"WsWuInfo loading remote log access plug-in...");
+    if (logInitWULogReader)
```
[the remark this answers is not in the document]
jakesmith: Yes, that was what I saying in my comment too : https://github.com/hpcc-systems/HPCC-Platform/pull/16343#discussion_r919168991

esp/services/ws_workunits/ws_workunitsHelpers.cpp
rpastrana: @jakesmith ConfigUpdateHook looks to be very useful, i think I've mentioned this before but these types of features need to be publicized loudly to the entire development team!

wangkx: It seems that the reloadConfigHook.installOnce(std::bind(&CWsWorkunitsEx::initWULogReader, this), false); does not run the  initWULogReader() itself.

rpastrana: @wangkx I see no issue keeping the logging in the init method if only called at load/reconfig time

rpastrana (approved):
@wangkx looks reasonable. 
BTW, I'll handle the related issue discussed offline as part of this change: https://track.hpccsystems.com/browse/HPCC-27983

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -206,6 +206,7 @@ WsWUExceptions::WsWUExceptions(IConstWorkUnit& wu): numerr(0), numwrn(0), numinf
     }
 }
 
+#ifdef _CONTAINERIZED
```
jakesmith, line 209: we should avoid not compiling container centric code, where possible.
Keeping it general and compiling in all builds, helps spot issues/has more coverage.

But also, in this case remote logs is not necessarily just a containerized feature, it could equally apply to bare-metal (if given configuration support etc.), so we should not exclude it upfront.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -4080,6 +4066,7 @@ void CWsWuFileHelper::createZAPECLQueryArchiveFiles(IConstWorkUnit* cwu, const c
...
```
jakesmith, line 4069: see previous comment, better not to exclude unless there's a particular reason

esp/services/ws_workunits/ws_workunitsHelpers.hpp
```diff
@@ -172,18 +172,18 @@ class WsWuInfo
...
     void copyContentFromRemoteFile(const char* sourceFileName, const char* sourceIPAddress,
         const char* sourceAlias, const char *outFileName);
-    void initWULogReader();
-    void setLogTimeRange(LogAccessConditions& logFetchOptions, unsigned wuLogSearchTimeBuffSecs);
     void getPostMortemFiles(IFile* file, unsigned& helpersCount, StringArray& postMortemFiles);
     void addPostMortemFiles(StringArray& postMortemFiles, IArrayOf<IEspECLHelpFile>& helpers);
     void validatePostMortemFile(IFile* file, const char* fileToBeValidated, bool& validated);
 
 public:
+#ifdef _CONTAINERIZED
```
jakesmith, line 186: see prev. comments re. not excluding

jakesmith (commented):
@wangkx - please see comments.

esp/services/ws_workunits/ws_workunitsAuditLogs.cpp
```diff
@@ -1444,6 +1444,9 @@ void CWsWorkunitsSoapBindingEx::createAndDownloadWUZAPFile(IEspContext& context,
 
     //CWsWuFileHelper may need ESP's <Directories> settings to locate log files. 
     CWsWuFileHelper helper(directories);
+#ifdef _CONTAINERIZED
+    helper.setRemoteLogAccess(wswService->queryRemoteLogAccessorPtr());
```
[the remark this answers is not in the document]
wangkx: @jakesmith @rpastrana If I understand the comment correctly, I should roll back all of the changes in https://github.com/hpcc-systems/HPCC-Platform/pull/16343/commits/64e9ea37773bd6b9946802b739544e57ee07c4bd. And remove the LOG line and the OWARNLOG line from initWULogReader() for now. When we revise the Singleton (https://track.hpccsystems.com/browse/HPCC-27983), we may handle the config changes, add the tracing, return a pointer, etc. within the Singleton.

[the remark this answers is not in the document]
jakesmith: Requirements for HPCC-27983 go beyond the requirements for handling a single initialization and ptr return from queryRemoteLogAccessor(). HPCC-27983 is to auto-detect configuration changes and to adjust (if necessary) the underlying object that queryRemoteLogAccessor() returns.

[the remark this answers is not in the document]
jakesmith: I've pushed a pending change to allow the Singleton helper to cope with being initialized (once) to null: https://track.hpccsystems.com/browse/HPCC-27988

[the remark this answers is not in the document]
jakesmith: > And remove the LOG line and the OWARNLOG line from initWULogReader() for now.

I would be happy with that in the short term if @rpastrana - not sure the logging is _that_ useful that we'd miss it, but it's also fine/better to be issued once only (as long as it doesn't complicate all the code that uses it)

Once HPCC-27988 is merged, I'd suggest queryRemoteLogAccessor is changed to return a ptr (and it is valid for it return null), and it can log 1st time initialization and/or exceptions before returning.

That change can be made ahead of work for HPCC-27983 - which is going to be a bit more work.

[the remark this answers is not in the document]
jakesmith: then let's wait for HPCC-27988 to be merged, and have the factory method (which will then only be called once) issue the logging so it's universal.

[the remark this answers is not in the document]
jakesmith: @wangkx https://github.com/hpcc-systems/HPCC-Platform/pull/16368 (HPCC-27988) is now merged.

wangkx:
@jakesmith I made the changes. Please take a look.

jakesmith (dismissed):
@wangkx - the idea was after HPCC-27988 was merged (related to the conversation at https://github.com/hpcc-systems/HPCC-Platform/pull/16343#discussion_r922096792),
for the logInitWULogReader to be redundant,
instead for the initialization logging, or failure logging, to go directly into the factor method (the lambda func. in jlog.cpp - line 3246).

Then callers can cheaply call queryRemoteLogAccessor() and if necessary check if it returns nullptr. No separate storage of a secondary pointer to a IRemoteLogAccess

wangkx:
@jakesmith please review

esp/services/ws_logaccess/WsLogAccessService.cpp
```diff
@@ -11,37 +11,17 @@ Cws_logaccessEx::~Cws_logaccessEx()
 bool Cws_logaccessEx::onGetLogAccessInfo(IEspContext &context, IEspGetLogAccessInfoRequest &req, IEspGetLogAccessInfoResponse &resp)
 {
     bool success = true;
-    if (m_remoteLogAccessor)
+    if (queryRemoteLogAccessor())
```
jakesmith, line 14: future: fine as it is now, but when or if an auto-updating mechanism is added (HPCC-27983), this will method will likely want to turn into getRemoteLogAccessor, so that it links the object, and store it in a temporary Owned, for use below. That way the update can change the object, whilst the existing object is still valid.

system/jlib/jlog.cpp
```diff
@@ -3239,15 +3239,18 @@ bool fetchLogByClass(LogQueryResultDetails & resultDetails, StringBuffer & retur
...
-IRemoteLogAccess &queryRemoteLogAccessor()
+IRemoteLogAccess *queryRemoteLogAccessor()
 {
-    return *logAccessor.query([]
+    return logAccessor.query([]
         {
-            Owned<IPropertyTree> logAccessPluginConfig = getGlobalConfigSP()->getPropTree("logAccess");
-#ifdef LOGACCESSDEBUG
-            if (!logAccessPluginConfig)
+            IRemoteLogAccess *remoteLogAccessor = nullptr;
```
jakesmith, line 3246: I think, as before, should keep a 'loading..' type message (that previously was in initWULogReader and Cws_logaccessEx::init), at the start of this function.
  PROGLOG("Loading remote log access plug-in.");

jakesmith (commented):
@wangkx - looks much cleaner, 1 minor comment

jakesmith (approved):

wangkx:

PR #16358, opened 2022-07-14
HPCC-27928 Copy files with storage striped across containers/file shares

shamser:
## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dali/base/dadfs.cpp
+++ b/dali/base/dadfs.cpp
@@ -2354,7 +2354,7 @@ class CDistributedFilePart: public CInterface, implements IDistributedFilePart
     CriticalSection sect;
     StringAttr overridename;    // may or not be relative to directory
     bool            dirty;      // whether needs updating in tree
-
+    std::vector<unsigned> stripeNumber;
     offset_t getSize(bool checkCompressed);
 
 public:
@@ -2437,6 +2437,7 @@ public:
         return ret;
     }
     virtual StringBuffer &getStorageFilePath(StringBuffer & path, unsigned copy) override;
+    virtual unsigned getStripeNum(unsigned copy) override;
 };
 
 // --------------------------------------------------------
@@ -4908,8 +4909,6 @@ StringBuffer &CDistributedFilePart::getStorageFilePath(StringBuffer & path, unsi
     if (!storagePlane)
         throw new CDFS_Exception(DFSERR_MissingStoragePlane, planeName);
 
-    // Note: striping to be implemented as a separate jira as additional information
-    // will be needed in the configuration
     path.append(parent.directory);
     if (parent.hasDirPerPart())
         addPathSepChar(path).append(partIndex+1); // part subdir 1 based
@@ -4919,6 +4918,29 @@ StringBuffer &CDistributedFilePart::getStorageFilePath(StringBuffer & path, unsi
     unsigned prefixLength = strlen(storagePlane->queryPrefix());
     path.remove(0, prefixLength);
     return path;
+}
+
+unsigned CDistributedFilePart::getStripeNum(unsigned copy)
+{
+    if (copy >= stripeNumber.size() || stripeNumber[copy]==UINT_MAX)
+    {
+        unsigned nc = copyClusterNum(copy, nullptr);
+        IClusterInfo &cluster = parent.clusters.item(nc);
+        const char *planeName = cluster.queryGroupName();
+        if (isEmptyString(planeName))
+        {
+            StringBuffer lname;
+            parent.getLogicalName(lname);
+            throw new CDFS_Exception(DFSERR_EmptyStoragePlane, lname.str());
+        }
+        Owned<IStoragePlane> storagePlane = getDataStoragePlane(planeName, false);
+        if (!storagePlane)
+            throw new CDFS_Exception(DFSERR_MissingStoragePlane, planeName);
+        if (copy >= stripeNumber.size())
+            stripeNumber.insert(stripeNumber.end(), copy - stripeNumber.size() + 1 , UINT_MAX); // create empty place holders
+        stripeNumber[copy] = calcStripeNumber(partIndex, parent.lfnHash, storagePlane->numDevices());
+    }
+    return stripeNumber[copy];
 }
 
 static unsigned findSubFileOrd(const char *name)
--- a/dali/base/dafdesc.cpp
+++ b/dali/base/dafdesc.cpp
@@ -3701,9 +3701,19 @@ public:
     {
         return xml->queryProp("storageapi/@account");
     }
-    virtual const char * queryStorageContainer() const override
+    virtual const char * queryStorageContainer(unsigned stripeNumber) const override
     {
-        return xml->queryProp("storageapi/@container");
+        if (stripeNumber==0) // stripeNumber==0 when not striped -> use first item in 'containers' list
+            stripeNumber++;
+        StringBuffer path;
+        path.appendf("storageapi/containers[%d]", stripeNumber);
+        const char * container = xml->queryProp(path.str());
+        if (isEmptyString(container))
+        {
+            const char * name = xml->queryProp("@name");
+            throw makeStringExceptionV(-1, "No container provided for %s: path %s", name, path.str());
+        }
+        return container;
     }
     virtual StringBuffer & getSASToken(StringBuffer & token) const override
     {
--- a/dali/ft/filecopy.cpp
+++ b/dali/ft/filecopy.cpp
@@ -1534,8 +1534,6 @@ bool FileSprayer::useAPICopy()
             return false;
         if (isEmptyString(targetStoragePlane->queryStorageApiAccount()))
             return false;
-        if (isEmptyString(targetStoragePlane->queryStorageContainer()))
-            return false;
 
         StringBuffer sourceClusterName;
         distributedSource->getClusterName(0, sourceClusterName);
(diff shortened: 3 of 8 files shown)


dali/base/dafdesc.cpp
```diff
@@ -3701,9 +3701,14 @@ class CStoragePlaneInfo : public CInterfaceOf<IStoragePlane>
...
     }
-    virtual const char * queryStorageContainer() const override
+    virtual StringBuffer & getStorageContainer(StringBuffer & container, unsigned stripeNumber) const override
     {
-        return xml->queryProp("storageapi/@container");
+        StringBuffer path("storageapi/");
+        if (stripeNumber==0)
+            path.append("@container");
+        else
+            path.appendf("containers/@stripe%d", stripeNumber);
```
ghalliday, line 3710: There should be a corresponding change in the schema/value.yaml comments.

ghalliday, line 3710: I was expecting containers[%d]/ so that the helm would be

```
containers: [ a, b, c, d ]
```
or
```
containers:
- a
- b
- c
```

```
containers:
  stripe1: a
  stripe2: b
  stripe3: c
```

doesn't seem so natural.  @jakesmith thoughts?

dali/ft/filecopy.cpp
```diff
@@ -2687,15 +2683,19 @@ class CApiCopyClient : public CInterfaceOf<IAPICopyClient>
...
-        uri.appendf(".core.windows.net/%s", container);
+        uri.appendf(".core.windows.net/");
     }
-    static void buildAzureStorageURI(IDistributedFile *distFile, unsigned partNum, const char * baseURI, const char *sasToken, StringBuffer & uri)
+    static void buildAzureStorageURI(IDistributedFile *distFile, IStoragePlane * storagePlane, unsigned partNum, const char * baseURI, const char *sasToken, StringBuffer & uri)
     {
+        IDistributedFilePart & distFilePart = distFile->queryPart(partNum);
         StringBuffer path;
-        distFile->queryPart(partNum).getStorageFilePath(path, 0);
+        distFilePart.getStorageFilePath(path, 0);
```
ghalliday, line 2692: minor:  Clearer to have a local variable for copy to avoid asking the question why is 0 being passed to this function - 

unsigned copy = 0; // no platform replication
distFilePart.getStorageFilePath(path, copy);
...
unsigned stripeNum = distFilePart.getStripeNum(copy);

dali/base/dafdesc.cpp
```diff
@@ -3701,9 +3701,14 @@ class CStoragePlaneInfo : public CInterfaceOf<IStoragePlane>
...
+        return container.append(xml->queryProp(path.str()));
```
ghalliday, line 3711: Should this throw an error if the container is not defined in the correct way.  Can the helm chart check for some consistency - e.g., number of containers matches the number of devices?

ghalliday (commented):
@shamser a few comments - generally looks good.

dali/base/dafdesc.cpp
shamser: @jakesmith @ghalliday What form should I implement?  The first form seems cleaner/neater.  The second form is self-documenting - easier to understand.

dali/base/dafdesc.cpp
shamser: I'll get helm chart to check the configuration and add a check here too.

dali/base/dadfs.cpp
```diff
@@ -4921,6 +4920,24 @@ StringBuffer &CDistributedFilePart::getStorageFilePath(StringBuffer & path, unsi
...
+    {
+        StringBuffer lname;
+        parent.getLogicalName(lname);
+        throw new CDFS_Exception(DFSERR_EmptyStoragePlane, lname.str());
+    }
+    Owned<IStoragePlane> storagePlane = getDataStoragePlane(planeName, false);
+    if (!storagePlane)
+        throw new CDFS_Exception(DFSERR_MissingStoragePlane, planeName);
+
+    return calcStripeNumber(partIndex, parent.lfnHash, storagePlane->numDevices());
```
jakesmith, line 4938: minor: it is probably not a big efficiency concern, but getStripeNum will probably be called repeatedly (e.g. when a process is dealing with multiple parts), it might be better to precalculate the {part#, copy#}->stripe# mapping in the parent once, to avoid each call looking up the data plane etc.

dali/base/dafdesc.cpp
jakesmith: I agree, when looking at the code above, I was expecting it would be dealing with a list of containers.

dali/base/dafdesc.cpp
```diff
@@ -3701,9 +3701,14 @@ class CStoragePlaneInfo : public CInterfaceOf<IStoragePlane>
     {
         return xml->queryProp("storageapi/@account");
     }
-    virtual const char * queryStorageContainer() const override
+    virtual StringBuffer & getStorageContainer(StringBuffer & container, unsigned stripeNumber) const override
```
jakesmith, line 3704: minor: curious why changed to fill in a StringBuffer, vs previous returning a pointer to the property value?

dali/ft/filecopy.cpp
jakesmith: should buildAzureStorageURI perhaps take a copyNum too, even if always 0 for now..

ghalliday (approved):
Changes look correct to me.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -221,7 +221,20 @@ storage:
   {{- if and (eq "data" $plane.category) (not $plane.defaultSprayParts) -}}
    {{- $_ := set $planeYaml "defaultSprayParts" (include "hpcc.getMaxNumWorkers" $ | int) -}}
   {{- end -}}
-
+  {{- /* Make sure there is enough containers provider if storageapi used*/ -}}
```
jakesmith, line 224: trivial: "containers provider" , containers provided?

helm/hpcc/templates/_helpers.tpl
```diff
@@ -221,7 +221,20 @@ storage:
...
+  {{- if $plane.storageapi -}}
+   {{- $numDevices := int ( $plane.numDevices | default $plane.numDevices | default 1 ) }}
+   {{- if gt $numDevices 1 }}
+    {{- $numContainers := len ($plane.storageapi.containers| default list) -}}
```
jakesmith, line 228: super trivial/pedantic: add space after 'containers'

dali/base/dafdesc.cpp
```diff
@@ -3701,9 +3701,20 @@ class CStoragePlaneInfo : public CInterfaceOf<IStoragePlane>
     {
         return xml->queryProp("storageapi/@account");
     }
-    virtual const char * queryStorageContainer() const override
+    virtual const char * queryStorageContainer(unsigned stripeNumber) const override
     {
-        return xml->queryProp("storageapi/@container");
+        StringBuffer path("storageapi/");
+        if (stripeNumber==0)
```
jakesmith, line 3707: question: it doesn't feel right quite right to me, that there's 1 specification for listing the storage container if no striping ("@container") and a different way if striping is on (list under "containers").

We pobably need to support the old way (@container) without striping for backward compatibility,
but going forward, I think you should be able to specify the single container under the "containers" list, in the case of a plane witout striping, so it's consistent.
@ghalliday - what's your opinion on this?

helm/hpcc/templates/_helpers.tpl
```diff
@@ -221,7 +221,20 @@ storage:
   {{- if and (eq "data" $plane.category) (not $plane.defaultSprayParts) -}}
    {{- $_ := set $planeYaml "defaultSprayParts" (include "hpcc.getMaxNumWorkers" $ | int) -}}
   {{- end -}}
-
+  {{- /* Make sure there is enough containers provider if storageapi used*/ -}}
+  {{- if $plane.storageapi -}}
+   {{- $numDevices := int ( $plane.numDevices | default $plane.numDevices | default 1 ) }}
+   {{- if gt $numDevices 1 }}
```
jakesmith, line 227: see comment/question in queryStorageContainer re. supporting single container under list too (but maintaining legacy @container for backward compatibility).

jakesmith (commented):
a couple of trivials, but looks good,
but 1 question: https://github.com/hpcc-systems/HPCC-Platform/pull/16358#discussion_r978429162

dali/base/dafdesc.cpp
ghalliday: That is a reasonable idea, but that can be a follow on change - we need to get this merged into 8.10.x asap.

jakesmith: ok. @shamser please open a JIRA for this change.

jakesmith (approved):
None of the other comments need stop this merging.
@shamser - please squash

dali/base/dafdesc.cpp
jakesmith: I see you've changed it.

It does look cleaner with change, but backward compatibility has been lost.
@shamser @ghalliday - Maybe that isn't important, if no one is actively using it so far?

jakesmith (approved):
Approved, but suggested backward compatibility has been lost. Was that intended? Is it important?

ghalliday:
I think it is ok to lose backward compatibility because no one has had a build that includes that feature yet,.

ghalliday (approved):

PR #16364, opened 2022-07-14
HPCC-27979 Add unicode version of AnalyzeText to NLP Plugin

dehilsterlexis:
Signed-off-by: David de Hilster <david.dehilster@lexisnexisrisk.com>

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:


dehilsterlexis:
@GordonSmith for review

plugins/nlp/lib_nlp.ecllib
```diff
@@ -17,4 +17,5 @@
 
 EXPORT nlp := SERVICE : plugin('nlp'), namespace('nlp'), library('nlp'), CPP, PURE
   string AnalyzeText(const string analyzer, const string txt) : cpp,pure,entrypoint='AnalyzeText';
-END;
+  unicode AnalyzeTextU(const string analyzer, const unicode txt) : cpp,pure,entrypoint='AnalyzeTextU';
```
GordonSmith, line 20: `UnicodeAnalyzeText` would be more inline with other ECL Unicode functions (no need to change entrypoint)

GordonSmith (changes requested):

plugins/nlp/nlp.cpp
```diff
@@ -68,6 +71,24 @@ namespace nlp {
...
+
+    ECL_NLP_API void ECL_NLP_CALL AnalyzeTextU(unsigned & tgtLen, UChar * & tgt, size32_t anaLen, const char * ana, unsigned txtLen, UChar const * txt)
+    {
+        {
+            CriticalBlock block(cs);
+            if (nlpEng == NULL) {
+                nlpEng = new NLPEng();
+            }
+        }
+        UnicodeString const in(txt, txtLen);
```
ghalliday, line 83: minor: The following is more efficient since it avoids cloning the string:

    UnicodeString const in(false, txt, txtLen);

plugins/nlp/nlp.cpp
```diff
@@ -68,6 +71,24 @@ namespace nlp {
...
+        std::string str;
+        in.toUTF8String(str);
+        StringBuffer anaBuff(anaLen,ana);
+        ostringstream sso;
+        tgtLen = nlpEng->nlpEngAnalyze(anaBuff,str.c_str(),sso);
+        tgt = (UChar *) CTXMALLOC(parentCtx, tgtLen);
+        memcpy_iflen(tgt, sso.str().c_str(), tgtLen);
```
ghalliday, line 90: This looks wrong - a string (utf8?) is being copied to a UChar string - which uses double byte characters.  So the length will not be right.

There should be some test cases calling the new function (which I suspect would generate invalid output...)

ghalliday (changes requested):

plugins/nlp/nlp.cpp
dehilsterlexis: Gavin, you are right. I have been aware of a length problem but have not yet had to time sluth it out. This must be the problem.

GordonSmith (approved):

dehilsterlexis:
@ghalliday still working on correcting the "length" problem.

ghalliday (approved):

ghalliday:
@dehilsterlexis did you want these changes in the 8.8.x series?

dehilsterlexis:
@ghalliday I'm not sure this passed the new unicode ECL regression test I added. I will check. Also, I believe the length problem may not yet be fixed.

plugins/nlp/nlp.cpp
```diff
@@ -68,6 +71,26 @@ namespace nlp {
...
+                nlpEng = new NLPEng();
+            }
+        }
+        UnicodeString const in(false, txt, txtLen);
+        std::string str;
+        in.toUTF8String(str);
+        StringBuffer anaBuff(anaLen,ana);
+        ostringstream sso;
+        nlpEng->nlpEngAnalyze(anaBuff,str.c_str(),sso);
+        icu::UnicodeString unicode = icu::UnicodeString::fromUTF8(icu::StringPiece(sso.str()));
```
GordonSmith, line 89: No need for `icu::` prefix (or remove `using icu::UnicodeString;` above);

GordonSmith (changes requested):

dehilsterlexis:
@GordonSmith @ghalliday I believe that the length problem is in fact solved. There are numerous errors but I'm not finding any that seem to be caused by the NLP plugin.

dehilsterlexis:
@GordonSmith it passed the ECL regression test. Looks like we are good to merge.

GordonSmith (approved):

PR #16374, opened 2022-07-18
HPCC-27907 Fix multiJobLinger long stall (until timeout) design flaw

jakesmith:
A multi linger setup would keep Thor's running waiting for new
jobs to reuse them.
However, if max instances were already running and busy (maxGraphs),
then new jobs were queued, and never picked up by the idle Thor's
(who did not listen to the queue).

This meant that if more jobs ran in parallel than there were
Thor's available, some would get queued and not run until one of
the Thor instances span down (after lingerPeriod).

This PR changes the design, so that all multi-linger jobs post
to the Q, instead of scanning and finding a lingering Thor.
Thor instances now pend on the queue for the linger period.

This means either the thoragent or a Thor will get a queued job.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/thorlcr/master/thgraphmanager.cpp
+++ b/thorlcr/master/thgraphmanager.cpp
@@ -15,6 +15,9 @@
     limitations under the License.
 ############################################################################## */
 
+#include <future>
+#include <chrono>
+
 #include "platform.h"
 #include <math.h>
 #include "jarray.hpp"
@@ -27,6 +30,7 @@
 #include "portlist.h"
 #include "wujobq.hpp"
 #include "daclient.hpp"
+#include "daqueue.hpp"
 
 #include "thgraphmaster.ipp"
 #include "thorport.hpp"
@@ -1113,6 +1117,7 @@ static int exitCode = -1;
 void setExitCode(int code) { exitCode = code; }
 int queryExitCode() { return exitCode; }
 
+static Owned<IJobQueue> thorQueue; // used if multiJobLinger is in use, and here so abortThor can cancel
 static unsigned aborting = 99;
 void abortThor(IException *e, unsigned errCode, bool abortCurrentJob)
 {
@@ -1134,6 +1139,11 @@ void abortThor(IException *e, unsigned errCode, bool abortCurrentJob)
         LOG(MCdebugProgress, thorJob, "abortThor called");
         if (jM)
             jM->stop();
+        if (thorQueue)
+        {
+            Owned<IJobQueue> queue = thorQueue.getLink();
+            queue->cancelAcceptConversation();
+        }
     }
     if (2 > aborting && abortCurrentJob)
     {
@@ -1206,7 +1216,6 @@ void closeThorServerStatus()
 }
 
 
-#ifdef _CONTAINERIZED
 /*
  * Waits on recv for another wuid/graph to run.
  * Return values:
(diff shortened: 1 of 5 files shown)


thorlcr/master/thgraphmanager.cpp
```diff
@@ -1215,35 +1218,42 @@ void closeThorServerStatus()
...
-    PROGLOG("Lingering time left: %.2f", ((float)timeoutMs)/1000);
-    CMessageBuffer msg;
-    if (!queryWorldCommunicator().recv(msg, NULL, MPTAG_THOR, nullptr, timeoutMs))
-        return -1;
     StringBuffer next;
-    msg.read(next);
+    CMessageBuffer msg;
+    if (thorQueue)
+    {
+        Owned<IJobQueueItem> item = thorQueue->dequeue(timeoutMs);
```
ghalliday, line 1227: Question: Is there any way to give the lingering thors a higher priority than waiting eclagents?

thorlcr/master/thgraphmanager.cpp
```diff
@@ -1215,35 +1218,42 @@ void closeThorServerStatus()
...
+                return 0; // mismatch/ignore
+            msg.clear().append(true);
+            if (!queryWorldCommunicator().reply(msg, 60*1000)) // should be quick!
+                return -2; // failed to reply to client
         }
-        else
-            return -2;
     }
-    return 0;
+    retWuid.set(sArray.item(0));
```
ghalliday, line 1253: no longer returns 0 if `2 != sArray.ordinality()`.  Does it matter?

thorlcr/master/thgraphmanager.cpp
ghalliday, line 1227: question: If thor is closed down does this cleanly exit?

ghalliday (commented):
@jakesmith looks much simpler.  A couple of questions.

thorlcr/master/thgraphmanager.cpp
jakesmith: I know that's a caveat - that it would be better if the lingering Thor instances took Q items, vs an agent (I thought I added a comment somewhere), but I ignored it.
There isn't a way to do that in the IJoqQueue implementation, and didn't think it was worth pursuing.

The net upshot is, that sometimes when the max instances aren't running, there will be times when the agent spins up, up to max when an idle Thor could have taken it instead.

I see this feature mainly being of benefit though to keep the max instances busy, in which case it will not be an issue.
Could open as a separate JIRA if worth looking into further

thorlcr/master/thgraphmanager.cpp
jakesmith: change not intentional. It should return 0, meaning unrecognised format, ignore and read next.
I'll change.

thorlcr/master/thgraphmanager.cpp
jakesmith: good question, nothing explicitly cancel/aborts it, and probably needs to.
I'll verify.

thorlcr/master/thgraphmanager.cpp
jakesmith: changed to return 0 (as before), meaning ignore unrecognised format

thorlcr/master/thgraphmanager.cpp
jakesmith: Have changed so queue will be cancelled on an abort condition.

thorlcr/master/thgraphmanager.cpp
```diff
@@ -1138,6 +1139,11 @@ void abortThor(IException *e, unsigned errCode, bool abortCurrentJob)
         LOG(MCdebugProgress, thorJob, "abortThor called");
         if (jM)
             jM->stop();
+        if (thorQueue)
+        {
+            Owned<IJobQueue> queue = thorQueue.getClear();
```
ghalliday, line 1144: Why the getClear()?  I think that may create a potential race condition, getLink() is probably safer.

ghalliday (commented):
@jakesmith one comment.
It would also probably be worth adding a jira about prioritised pulling form a queue - since it will potentially reduce the cost and latency using the system.

thorlcr/master/thgraphmanager.cpp
jakesmith: agree, could cause problems, I'll change to getLink

jakesmith:
> @jakesmith one comment. It would also probably be worth adding a jira about prioritised pulling form a queue - since it will potentially reduce the cost and latency using the system.

Have opened a JIRA for improvement : https://track.hpccsystems.com/browse/HPCC-28034

jakesmith:
@ghalliday - please re-review

ghalliday (approved):

richardkchapman:
@jakesmith Please squash

jakesmith:
@richardkchapman - squashed

PR #16385, opened 2022-07-20
HPCC-27632 Provide standard k8s labeling template

rpastrana:
- Provides global template which populates standard k8s labels
- Standardizes ESP/Roxie/PVC labels

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/helm/hpcc/templates/_helpers.tpl
+++ b/helm/hpcc/templates/_helpers.tpl
@@ -1169,6 +1169,7 @@ metadata:
   name: {{ $lvars.serviceName | quote }}
   labels:
     helmVersion: 8.8.0-closedown0
+    {{- include "hpcc.addStandardLabels" (dict "root" $.root "instance" $lvars.serviceName ) | indent 4 }}
 {{- if $lvars.labels }}
 {{ toYaml $lvars.labels | indent 4 }}
 {{- end }}
@@ -1245,6 +1246,50 @@ dali data
 {{- end -}}
 
 {{/*
+A template to generate the standard app.kubernetes.io labels
+
+root name(k8s application name) component(component within the application, can be same as app) instance 
+
+https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/
+----------------------------+-------------------------------------------------------------+----------
+| Label                     | Description                                                 | Example |
+----------------------------+-------------------------------------------------------------+----------
+app.kubernetes.io/name       The name of the application                                   mysql
+app.kubernetes.io/component  The component within the architecture                         database
+app.kubernetes.io/instance   A unique name identifying the instance of an application      mysql-abcxzy
+app.kubernetes.io/version    The current version of the application                        5.7.21
+app.kubernetes.io/part-of    The name of a higher level application this one is part of    wordpress
+app.kubernetes.io/managed-by The tool being used to manage the operation of an application helm
+app.kubernetes.io/created-by The controller/user who created this resource                 controller-manager
+helm.sh/chart                This should be the chart name and version
+*/}}
+{{- define "hpcc.addStandardLabels" }}
+app.kubernetes.io/part-of: HPCC-Platform
+{{- if .name }}
+app.kubernetes.io/name: {{ .name }}
+{{- end }}
+{{- if .component }}
+app.kubernetes.io/component: {{ .component }}
+{{- end }}
+{{- if .instance }}
+app.kubernetes.io/instance: {{ .instance }}
+{{- end }}
+{{- if .root }}
+ {{- if hasKey .root "Release" }}
+app.kubernetes.io/managed-by: {{ .root.Release.Service }}
+ {{- end }}
+ {{- if hasKey .root "Chart" }}
+  {{- if .root.Chart.Version }}
+app.kubernetes.io/version: {{ .root.Chart.Version }}
+   {{- if .root.Chart.Name }}
+helm.sh/chart: {{ .root.Chart.Name }}-{{ .root.Chart.Version | replace "+" "_" }}
+   {{- end }}
+  {{- end }}
+ {{- end }}
+{{- end }}
+{{- end }}
+
+{{/*
 A template to generate a PVC
 Pass in dict with root, me, name, and optional path
 */}}
@@ -1254,10 +1299,7 @@ kind: PersistentVolumeClaim
 metadata:
   name: {{ printf "%s-%s" (include "hpcc.fullname" .) .name }}
   labels:
-    app.kubernetes.io/name: {{ printf "%s-%s" (include "hpcc.fullname" .) .name }}
-    app.kubernetes.io/instance: {{ .root.Release.Name }}
-    app.kubernetes.io/managed-by: {{ .root.Release.Service }}
-    helm.sh/chart: {{ include "hpcc.chart" . }}
+    {{- include "hpcc.addStandardLabels" (dict "root" $.root "instance" .name "component" "storage") | indent 4 }}
 spec:
   accessModes:
     - {{ .mode | default .me.storageMode | default "ReadWriteMany" }}
--- a/helm/hpcc/templates/eclagent.yaml
+++ b/helm/hpcc/templates/eclagent.yaml
@@ -55,6 +55,7 @@ data:
       template:
         metadata:
           labels:
+            {{- include "hpcc.addStandardLabels" (dict "root" $ "component" $apptype "name" "eclagent" "instance" $appJobName) | indent 12 }}
             accessDali: "yes"
             accessEsp: "yes"
             helmVersion: 8.8.0-closedown0
@@ -128,6 +129,7 @@ spec:
   template:
     metadata:
       labels:
+        {{- include "hpcc.addStandardLabels" (dict "root" $ "component" $apptype "name" "eclagent" "instance" .name) | indent 8 }}
         run: {{ .name | quote }}
         accessDali: "yes"
         accessEsp: {{ .useChildProcesses | default false | ternary "yes" "no" | quote }}
--- a/helm/hpcc/templates/eclccserver.yaml
+++ b/helm/hpcc/templates/eclccserver.yaml
@@ -54,6 +54,7 @@ data:
       template:
         metadata:
           labels:
+            {{- include "hpcc.addStandardLabels" (dict "root" $ "component" "eclccserver" "name" "eclccserver" "instance" $compileJobName ) | indent 12 }}
             accessDali: "yes"
             accessEsp: "yes"
             helmVersion: 8.8.0-closedown0
@@ -137,6 +138,7 @@ spec:
   template:
     metadata:
       labels:
+        {{- include "hpcc.addStandardLabels" (dict "root" $ "component" "eclccserver" "name" "eclccserver" "instance" .name) | indent 8 }}
         run: {{ .name | quote }}
         accessDali: "yes"
         accessEsp: {{ .useChildProcesses | default false | ternary "yes" "no" | quote }}
--- a/helm/hpcc/templates/localroxie.yaml
+++ b/helm/hpcc/templates/localroxie.yaml
@@ -69,6 +69,7 @@ spec:
         accessDali: "yes"
         accessEsp: "yes"
         helmVersion: 8.8.0-closedown0
+        {{- include "hpcc.addStandardLabels" (dict "root" $ "component" "roxie-server" "name" "roxie" "instance" $roxie.name) | indent 8 }}
 {{- if hasKey . "labels" }}
 {{ toYaml .labels | indent 8 }}
 {{- end }}
--- a/helm/hpcc/templates/roxie.yaml
+++ b/helm/hpcc/templates/roxie.yaml
@@ -113,6 +113,7 @@ spec:
   template:
     metadata:
       labels:
+        {{- include "hpcc.addStandardLabels" (dict "root" $ "component" "topology-server" "name" "roxie" "instance" $commonCtx.toponame) | indent 8 }}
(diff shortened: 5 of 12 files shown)


richardkchapman:
The referenced jira does not look related?

rpastrana:
> The referenced jira does not look related?

Corrected: https://track.hpccsystems.com/browse/HPCC-27632?filter=-2

rpastrana:
Current mapping needs work. Here's my attempt at visualizing current mapping:

image

helm/hpcc/templates/eclagent.yaml
```diff
@@ -128,10 +128,10 @@ spec:
   template:
     metadata:
       labels:
+{{ include "hpcc.addStandardLabels" (dict "root" $ "component" $apptype "name" "eclagent") | indent 8 }}
```
jakesmith, line 131: missing instance, "instance" .name

helm/hpcc/templates/eclccserver.yaml
```diff
@@ -54,9 +54,9 @@ data:
       template:
         metadata:
           labels:
+{{ include "hpcc.addStandardLabels" (dict "root" $ "component" .me.name "name" "eclccserver" "instance"  $compileJobName ) | indent 12 }}
```
jakesmith, line 57: trivial: double space after "instance"
(also indentation - as per prev. comments - {{- .... | nindent ..}}

helm/hpcc/templates/roxie.yaml
```diff
@@ -234,7 +234,7 @@ spec:
         roxie-cluster: {{ $roxie.name | quote }}
         accessDali: "yes"
         accessEsp: "yes"
-        helmVersion: 8.8.0-closedown0
+{{ include "hpcc.addStandardLabels" (dict "root" $ "component" "server" "name" $roxie.name "instance" $servername) | indent 8 }}
```
jakesmith, line 237: component name "server" is too generic, elsewhere (in roxie.yaml) it is called "roxie-server"

helm/hpcc/templates/roxie.yaml
```diff
@@ -113,9 +113,9 @@ spec:
   template:
     metadata:
       labels:
+{{ include "hpcc.addStandardLabels" (dict "root" $ "component" "roxie" "name" $roxie.name "instance" $commonCtx.toponame) | indent 8 }}
```
jakesmith, line 116: I was expecting this to be component=topology-server, name=roxie, instance=<toponame> ?

helm/hpcc/templates/roxie.yaml
```diff
@@ -175,7 +175,7 @@ kind: Service
 metadata:
   name: {{ $commonCtx.toponame | quote }}
   labels:
-    helmVersion: 8.8.0-closedown0
+{{ include "hpcc.addStandardLabels" (dict "root" $ "component" "topology-server" "name" $roxie.name "instance" $commonCtx.toponame) | indent 4 }}
```
jakesmith, line 178: I'm not clear about "name" here, I thought it should be a fixed name of the application, with instance being the instance of that application, so here name would be "topology-server" (although as we discussed offline, it would be better if the naming convention made it distinct from the deployment name),

And, as with comments on/off line, re. component on things like PVC's and services, I think component be something like "service" here.

helm/hpcc/templates/roxie.yaml
```diff
@@ -328,14 +328,14 @@ spec:
   template:
     metadata:
       labels:
+{{ include "hpcc.addStandardLabels" (dict "root" $ "component" ( printf "%s-agent" $roxie.name ) "name" $roxie.name "instance" $name) | indent 8 }}
```
jakesmith, line 331: I don't think only "instance" should be dynamic - should always be component=roxie-agent (as used elsewhere), and name should be "roxie" I think.

helm/hpcc/templates/thor.yaml
```diff
@@ -82,7 +82,7 @@ data:
           labels:
             accessDali: "yes"
             accessEsp: "yes"
-            helmVersion: 8.8.0-closedown0
+{{ include "hpcc.addStandardLabels" (dict "root" $ "component" "thor-noteclagentusechildprocess" "name" "thor" "instance"  $eclAgentJobName) | indent 8 }}
```
jakesmith, line 85: component should be "eclagent"

double space after "instance"

helm/hpcc/templates/thor.yaml
```diff
@@ -148,9 +148,9 @@ data:
             accessEsp: "yes"
             app: "thor"
             component: "thormanager"
-            helmVersion: 8.8.0-closedown0
             instance: "_HPCC_JOBNAME_"
             job: "_HPCC_JOBNAME_"
+{{ include "hpcc.addStandardLabels" (dict "root" $ "component" "thormanager" "name" "thor" "instance"  $thorManagerJobName ) | indent 12 }}
```
jakesmith, line 153: double space after "instance"

helm/hpcc/templates/thor.yaml
```diff
@@ -348,8 +348,8 @@ spec:
         accessEsp: {{ $commonCtx.eclAgentUseChildProcesses | ternary "yes" "no" | quote }}
         app: "thor"
         component: "thor-eclagent"
-        helmVersion: 8.8.0-closedown0
         instance: {{ $commonCtx.eclAgentName | quote }}
+{{ include "hpcc.addStandardLabels" (dict "root" $ "component" "thor-eclagent" "name" "thor" "instance" $commonCtx.eclAgentName ) | indent 8 }}
```
jakesmith, line 352: component="eclagent"

helm/hpcc/templates/thor.yaml
```diff
@@ -409,8 +409,8 @@ spec:
         accessEsp: "no"
         app: "thor"
         component: "thor-thoragent"
-        helmVersion: 8.8.0-closedown0
         instance: {{ $commonCtx.thorAgentName | quote }}
+{{ include "hpcc.addStandardLabels" (dict "root" $ "component" "thor-thoragent" "name" "thor" "instance" $commonCtx.thorAgentName ) | indent 8 }}
```
jakesmith, line 413: component="eclagent"

jakesmith (commented):
@rpastrana - please see comments

helm/hpcc/templates/roxie.yaml
rpastrana: roxie-server for app roxie is a bit redundant, but that's fine we'll shoot for consistency

helm/hpcc/templates/roxie.yaml
rpastrana: components like PVCs, which are not necessarily dedicated to a single app, therefore not part of a particular component can be categorized into generic "storage", etc component.
This particular service seems to be dedicated to a particular app/component roxie/topology-server

helm/hpcc/templates/thor.yaml
rpastrana: ok but this seems to contradict an earlier change you suggested where the app/component was roxie/server changed to roxie/roxie-server

in this case thor/thor-eclagent changed to thor/eclagent...

helm/hpcc/templates/_helpers.tpl
```diff
@@ -1244,6 +1244,46 @@ dali data
...
+app.kubernetes.io/name       The name of the application                                   mysql
+app.kubernetes.io/component  The component within the architecture                         database
+app.kubernetes.io/instance   A unique name identifying the instance of an application      mysql-abcxzy
+app.kubernetes.io/version    The current version of the application                        5.7.21
+app.kubernetes.io/part-of    The name of a higher level application this one is part of    wordpress
+app.kubernetes.io/managed-by The tool being used to manage the operation of an application helm
+app.kubernetes.io/created-by The controller/user who created this resource 	               controller-manager
+*/}}
+{{ define "hpcc.addStandardLabels" }}
+app.kubernetes.io/part-of: {{ "HPCC-Platform" }}
```
jakesmith, line 1265: app.kubernetes.io/part-of: HPCC-Platform

( no need for {{ }} )

helm/hpcc/templates/eclccserver.yaml
```diff
@@ -54,9 +54,9 @@ data:
       template:
         metadata:
           labels:
+            {{- include "hpcc.addStandardLabels" (dict "root" $ "component" .me.name "name" "eclccserver" "instance" $compileJobName ) | indent 12 }}
```
jakesmith, line 57: > "component" .me.name

I think this should be "eclccserver" (NB: it's really what would be 'eclcc' in bare-metal - but it is rolled into 1 in k8s)
.me.name is going to become whatever name the eclccserver in this list has been given (which happens to be 'eclccserver' in the example values.yaml)

helm/hpcc/templates/localroxie.yaml
```diff
@@ -68,7 +68,7 @@ spec:
         server: {{ $servername | quote }}
         accessDali: "yes"
         accessEsp: "yes"
-        helmVersion: 8.8.0-closedown0
+        {{- include "hpcc.addStandardLabels" (dict "root" $ "component" $servername "name" "roxie" "instance" $roxie.name) | indent 8 }}
```
jakesmith, line 71: I don't think "component" should be user defined (or rather values.yaml defined), it should a known constant.
I think "roxie-server" is appropriate here  (same as in roxie.yaml)

helm/hpcc/templates/_helpers.tpl
```diff
@@ -1244,6 +1244,46 @@ dali data
...
+----------------------------+-------------------------------------------------------------+----------
+app.kubernetes.io/name       The name of the application                                   mysql
+app.kubernetes.io/component  The component within the architecture                         database
+app.kubernetes.io/instance   A unique name identifying the instance of an application      mysql-abcxzy
+app.kubernetes.io/version    The current version of the application                        5.7.21
+app.kubernetes.io/part-of    The name of a higher level application this one is part of    wordpress
+app.kubernetes.io/managed-by The tool being used to manage the operation of an application helm
+app.kubernetes.io/created-by The controller/user who created this resource 	               controller-manager
+*/}}
+{{ define "hpcc.addStandardLabels" }}
```
jakesmith, line 1264: should chomp space at start:
```
{{- define "hpcc.addStandardLabels" }}
```

it's debatable whether should at end too I suppose, but I think correct to leave template adding newline at start by not chomping whitespace at end ( -}} )
Which means when this template is used, you want | indent, not | nindent.

jakesmith:
@rpastrana - please see new comments

helm/hpcc/templates/localroxie.yaml
jakesmith: corrected: should be user defined -> shouldn't be user defined

richardkchapman:
The reason that the helm version was explicitly set in each template was to make it easier to spot when people had messed things up and used the templates from a different version to the chart. Feels like this change undoes that?

jakesmith:
> The reason that the helm version was explicitly set in each template was to make it easier to spot when people had messed things up and used the templates from a different version to the chart. Feels like this change undoes that?

agree, so we want to keep this/a hard coded version of the git tag the template belongs to
(helmVersion is automatically altered/pushed when a gold or RC build is tagged)

rpastrana:
> The reason that the helm version was explicitly set in each template was to make it easier to spot when people had messed things up and used the templates from a different version to the chart. Feels like this change undoes that?

Interesting... that scenario was not considered. 
It seems the new centralized template could set a "helmtemplatever" label with the hardcoded ver? 
Unless we're concerned (doubt it) about users mixing various templates.
Anyway, what process is currently setting the helmversion?

jakesmith:
>Unless we're concerned (doubt it) about users mixing various templates.

I think that's exactly the concern.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -1244,6 +1244,50 @@ dali data
...
+{{- if .root }}
+ {{- if hasKey .root "Release" }}
+app.kubernetes.io/managed-by: {{ .root.Release.Service }}
+ {{- end }}
+ {{- if hasKey .root "Chart" }}
+  {{- if .root.Chart.Version }}
+app.kubernetes.io/version: {{ .root.Chart.Version }}
+   {{- if .root.Chart.Name }}
+helm.sh/chart: {{ .root.Chart.Name }}-{{ .root.Chart.Version | replace "+" "_" }}
+   {{- end }}
```
jakesmith, line 1285: based on the other commens/conv., I suspect better to revert helm.sh/chart to way it was in each template file.

jakesmith (commented):
@rpastrana - except for the helmVersion per template issue, looks good to me.

rpastrana:
> > Unless we're concerned (doubt it) about users mixing various templates.
> 
> I think that's exactly the concern.

The mixing of HPCC-Platform/helm/hpcc/templates/* seems outrageous, but ok, will revert the "helmVersion" removal (which I assume is inserted by some outside process anyway).
My standing suggestion there would be to update the label name to "helmTemplateVersion", but that's outside the scope of this PR anyway.

helm/hpcc/templates/_helpers.tpl
jakesmith: you're right, it was about "helmVersion" - my comment is misplaced.

jakesmith (approved):
@rpastrana - looks good. Please squash.

rpastrana:
@richardkchapman ready for merge

richardkchapman (approved):

PR #16415, opened 2022-07-27
HPCC-27692 Add histogram metric type

kenrowland:
Added histogram metric

Signed-Off-By: Ken Rowland kenneth.rowland@lexisnexisrisk.com

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Cloud-compatibility
  - [x] Premature optimization

## Smoketest:

### DIFF
--- a/devdoc/Metrics.rst
+++ b/devdoc/Metrics.rst
@@ -290,6 +290,19 @@ metric value as the custom metric class provides no update methods. The componen
 for ensuring atomic access to the value if necessary.
 
 
+Histogram Metric
+-----------------
+Records counts of measurements according to defined bucket limits. When created, the caller defines
+as set of bucket limits. During event recording, the component records measurements. The metric
+separates each recorded measurement into its bucket by testing the measurement value against each
+bucket limit using a less than or equal test. Each bucket contains a count of measurements meeting
+that criteria. Additionally, the metric maintains a default bucket for measurements outside of the
+maximum bucket limit. This is sometimes known as the "inf" bucket.
+
+Some storage systems, such as Prometheus, require each bucket to accumulate its measurements with
+the previous bucket(s). It is the responsibility of the sink to accumulate values as needed.
+
+
 *************
 Configuration
 *************
--- a/system/jlib/jmetrics.hpp
+++ b/system/jlib/jmetrics.hpp
@@ -42,7 +42,8 @@ MetricsManager jlib_decl &queryMetricsManager();
 enum MetricType
 {
     METRICS_COUNTER,
-    METRICS_GAUGE
+    METRICS_GAUGE,
+    METRICS_HISTOGRAM
 };
 
 
@@ -86,7 +87,6 @@ interface IMetric
      */
     virtual __uint64 queryValue() const = 0;
 
-
     /*
      * Query the meta data for the metric
      */
@@ -96,6 +96,16 @@ interface IMetric
      * Get the units for the metric
      */
     virtual StatisticMeasure queryUnits() const = 0;
+
+    /*
+     * Query histogram values (for histogram metrics)
+     */
+    virtual std::vector<__uint64> queryHistogramValues() const = 0;
+
+    /*
+     * Query histogram labels
+     */
+    virtual std::vector<std::string> queryHistogramLabels() const = 0;
 };
 
 
@@ -110,8 +120,10 @@ public:
     virtual const std::string &queryName() const override { return name; }
     virtual const std::string &queryDescription() const override { return description; }
     virtual MetricType queryMetricType() const override { return metricType; }
-    const MetricMetaData &queryMetaData() const { return metaData; }
+    const MetricMetaData &queryMetaData() const override { return metaData; }
     StatisticMeasure queryUnits() const override { return units; }
+    virtual std::vector<__uint64> queryHistogramValues() const override { return {}; }
+    virtual std::vector<std::string> queryHistogramLabels() const override { return {}; }
 
 
 protected:
@@ -234,6 +246,48 @@ public:
 
 protected:
     T &value;
+};
+
+
+struct BucketDef
+{
+    std::string label;
+    __uint64 limit;
+};
+
+
+struct Bucket
+{
+    Bucket(const std::string &_label, __uint64 _limit) :
+        label{_label}, limit{_limit}, count{0} {}
+    std::string label;
+    __uint64 limit;
+    __uint64 count;
+};
+
+
+class jlib_decl HistogramMetric : public MetricBase
+{
+public:
+    HistogramMetric(const char *name, const char *desc, StatisticMeasure _units, const std::vector<BucketDef> &bucketDefs, const MetricMetaData &_metaData = MetricMetaData());
+
+    virtual __uint64 queryValue() const override
+    {
+        return sum;
+    }
+
+    void recordMeasurement(__uint64 measurement);
+    virtual std::vector<__uint64> queryHistogramValues() const override;
+    virtual std::vector<std::string> queryHistogramLabels() const override;
+
+protected:
+    Bucket &findBucket(__uint64 measurement);
+
+protected:
+    std::vector<Bucket> buckets;
+    mutable CriticalSection cs;
+    Bucket inf{"inf", 0};
+    __uint64 sum{0};
 };
 
 
--- a/testing/unittests/metrics/MetricFrameworkTests.cpp
+++ b/testing/unittests/metrics/MetricFrameworkTests.cpp
@@ -74,6 +74,7 @@ public:
         CPPUNIT_TEST(Test_scoped_updater_classes);
         CPPUNIT_TEST(Test_metric_meta_data);
         CPPUNIT_TEST(Test_gauge_by_counters_metric);
+        CPPUNIT_TEST(Test_histogram_metric);
     CPPUNIT_TEST_SUITE_END();
 
 protected:
@@ -306,6 +307,59 @@ protected:
 
         pCounterStarted->inc(2);
         CPPUNIT_ASSERT_EQUAL(0, static_cast<int>(pCounterGauge->queryValue()));
+    }
+
+
+    void Test_histogram_metric()
+    {
+        std::vector<BucketDef> bucketDefs = {{"le 2", 2}, {"le 4", 4}, {"le 8", 8} };
+        std::shared_ptr<HistogramMetric> pHistogram = std::make_shared<HistogramMetric>("requests.dist", "description", SMeasureCount, bucketDefs);
+
+        std::vector<__uint64> values;
+
+        //
+        // Verify that the correct number of buckets is returned
+        values = pHistogram->queryHistogramValues();
+        CPPUNIT_ASSERT_EQUAL(bucketDefs.size()+1, values.size());
+
+        //
+        // Add a measurement and make sure it goes into the correct bucket
+        int sumMeasurements = 0;
+
+        pHistogram->recordMeasurement(4);
+        sumMeasurements += 4;
+        checkHistogramBucketResult(pHistogram, sumMeasurements, {0,1,0,0});
+
+        //
+        // Add same measurement and make sure the bucket increments
+        pHistogram->recordMeasurement(4);
+        sumMeasurements += 4;
+        checkHistogramBucketResult(pHistogram, sumMeasurements, {0,2,0,0});
+
+        //
+        // Add a measurement that is beyond the highest bucket and make sure it goes in the
+        // last "inf" bucket
+        pHistogram->recordMeasurement(20);
+        sumMeasurements += 20;
+        checkHistogramBucketResult(pHistogram, sumMeasurements, {0,2,0,1});
+
+        //
+        // Just to make sure, increment another bucket
+        pHistogram->recordMeasurement(8);
+        sumMeasurements += 8;
+        checkHistogramBucketResult(pHistogram, sumMeasurements, {0,2,1,1});
+    }
+
+
+    void checkHistogramBucketResult(std::shared_ptr<IMetric> pHistogram, int expectedSum, const std::vector<int> &expectedValues)
+    {
+        std::vector<__uint64> values = pHistogram->queryHistogramValues();
+        CPPUNIT_ASSERT_EQUAL(static_cast<int>(pHistogram->queryValue()), expectedSum);
+        auto size = expectedValues.size();
+        for (unsigned i=0; i<size; ++i)
+        {
+            CPPUNIT_ASSERT_EQUAL(static_cast<int>(values[i]), expectedValues[i]);
+        }
     }
 
 protected:
--- a/system/jlib/jmetrics.cpp
+++ b/system/jlib/jmetrics.cpp
@@ -28,6 +28,66 @@ MODULE_EXIT()
 }
 
 
+HistogramMetric::HistogramMetric(const char *name, const char *desc, StatisticMeasure _units, const std::vector<BucketDef> &bucketDefs, const MetricMetaData &_metaData) :
+        MetricBase(name, desc, MetricType::METRICS_HISTOGRAM, _units, _metaData)
+{
+    buckets.reserve(bucketDefs.size());
+    for (auto &b: bucketDefs)
+    {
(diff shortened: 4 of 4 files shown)


kenrowland:
@ghalliday First pass at a histogram metric. It follows the requirements for a Prometheus histogram with the exception that it does not accumulate measurements as the bucket limits increase. That can be done by the sink as a metrics storage dependency. 

The class is templated to allow for different types for bucket limits.

system/jlib/jmetrics.hpp
```diff
@@ -237,6 +236,101 @@ class CustomMetric : public MetricBase
 };
 
 
+template <typename T>
```
ghalliday, line 239: I don't think the template parameter is a good idea for a few reasons
* It complicates all the types, and I'm not sure it adds much
* The sinks need to be able to get at the results - they need a virtual function in IMetric that they can call to get them, and that is not possible if the type is a template.

system/jlib/jmetrics.hpp
```diff
@@ -237,6 +236,101 @@ class CustomMetric : public MetricBase
...
+template<typename T>
+class jlib_decl HistogramMetric : public MetricBase
+{
+public:
+    HistogramMetric(const char *name, const char *desc, StatisticMeasure _units, const std::vector<BucketInfo<T>> &bucketDefs, const MetricMetaData &_metaData = MetricMetaData()) :
+        MetricBase(name, desc, MetricType::METRICS_HISTOGRAM, _units, _metaData)
+        {
+            for (auto &b: bucketDefs)
+            {
+                buckets.emplace_back(Bucket<T>{b.label, b.limit});
```
ghalliday, line 284: If you're using emplace_back() I think you should be able to say

        buckets.emplace_back(b.label, b.limit);

system/jlib/jmetrics.hpp
```diff
@@ -237,6 +236,101 @@ class CustomMetric : public MetricBase
...
+    }
+
+    HistogramMetricValues<T> queryValues()
+    {
+        HistogramMetricValues<T> values;
+        values.count = count;
+        values.sum = sum;
+        for (auto const &bucket: buckets)
+        {
+            values.bucketValues.emplace_back(BucketValue{bucket.label, bucket.count});
```
ghalliday, line 307: minor:    emplace_back(bucket.label, bucket.count);

system/jlib/jmetrics.hpp
```diff
@@ -237,6 +236,101 @@ class CustomMetric : public MetricBase
...
+        }
+        values.bucketValues.emplace_back(inf.label, inf.count);
+        return values;
+    }
+
+protected:
+    Bucket<T> &getBucket(T measurement)
+    {
+        for (auto b = buckets.begin(); b != buckets.end(); b++)
```
ghalliday, line 316: minor: use `for (auto &bucket: buckets)` as above

testing/unittests/metrics/MetricFrameworkTests.cpp
```diff
@@ -308,6 +309,62 @@ class MetricFrameworkTests : public CppUnit::TestFixture
...
+    void Test_historgram_metric()
+    {
+        std::vector<BucketInfo<unsigned>> buckets = {{"le 2", 2}, {"le 4", 4}, {"le 8", 8} };
+        std::shared_ptr<HistogramMetric<unsigned>> pHistogram = std::make_shared<HistogramMetric<unsigned>>("requests.dist", "description", SMeasureCount, buckets);
+
+        HistogramMetricValues<unsigned> values;
+
+        //
+        // Verify that the correct number of buckets is returned
+        values = pHistogram->queryValues();
```
ghalliday, line 321: To mimic use from a sink this test needs to use an IMetric to retrieve the values.

system/jlib/jmetrics.hpp
ghalliday, line 307: As far as I can see the Prometheus standard is for the counts to accumulate.  i.e. if the counts are {1 ,2, 2, 1} they would be reported as { 1, 3, 5, 6 }.  Is this the responsibility of the sink or the histogram?  I don't think it matters either way, but it needs to be clearly documented.

system/jlib/jmetrics.hpp
```diff
@@ -237,6 +236,101 @@ class CustomMetric : public MetricBase
...
+    {
+        count.fetch_add(1);           // total cumulative count of measurements
+        sum.fetch_add(measurement);   // sum of all measurements
+        getBucket(measurement).count.fetch_add(1);   // count by buckets
+    }
+
+    HistogramMetricValues<T> queryValues()
+    {
+        HistogramMetricValues<T> values;
+        values.count = count;
```
ghalliday, line 303: This could be inconsistent, does it matter? It is possible the count could be higher than the sum of the bucket counts if other calls to recordMeasurement() come in while it is executing.
It may be worth considering if a critical section is better than atomics.  On my machine uncontended critical sections have an overhead of ~16ns, atomic increments have an overhead of ~7ns. (Run `unittests AtomicTimingTest` to see full details.)

system/jlib/jmetrics.hpp
```diff
@@ -237,6 +236,101 @@ class CustomMetric : public MetricBase
...
+    void recordMeasurement(T measurement)
+    {
+        count.fetch_add(1);           // total cumulative count of measurements
+        sum.fetch_add(measurement);   // sum of all measurements
+        getBucket(measurement).count.fetch_add(1);   // count by buckets
+    }
+
+    HistogramMetricValues<T> queryValues()
+    {
+        HistogramMetricValues<T> values;
```
ghalliday, line 302: worth calling

    values.bucketsValues.reserve(buckets.size()+1);

to avoid reallocation of the result.

system/jlib/jmetrics.hpp
ghalliday, line 307: Does the result need to include the label?  This should be available to the sink from the meta.

testing/unittests/metrics/MetricFrameworkTests.cpp
```diff
@@ -308,6 +309,62 @@ class MetricFrameworkTests : public CppUnit::TestFixture
...
+        countMeasurements++;
+        values = pHistogram->queryValues();
+        CPPUNIT_ASSERT_EQUAL(static_cast<int>(values.bucketValues[0].count), 0);
+        CPPUNIT_ASSERT_EQUAL(static_cast<int>(values.bucketValues[1].count), 2);
+        CPPUNIT_ASSERT_EQUAL(static_cast<int>(values.bucketValues[2].count), 0);
+        CPPUNIT_ASSERT_EQUAL(static_cast<int>(values.bucketValues[3].count), 0);
+        CPPUNIT_ASSERT_EQUAL(static_cast<int>(values.sum), sumMeasurements);
+        CPPUNIT_ASSERT_EQUAL(static_cast<int>(values.count), countMeasurements);
+
+        //
```
ghalliday, line 353: I doubt there is a problem, but the unit test should also test increment other buckets.

Potentially worth adding a 

checkResults(values, sum, count, { 0, 2, 0, 0});

 to reduce the code checking the test cases.

system/jlib/jmetrics.hpp
```diff
@@ -237,6 +236,101 @@ class CustomMetric : public MetricBase
...
+    std::vector<BucketValue> bucketValues;
+    __uint64 count;
+    T sum;
+};
+
+template<typename T>
+class jlib_decl HistogramMetric : public MetricBase
+{
+public:
+    HistogramMetric(const char *name, const char *desc, StatisticMeasure _units, const std::vector<BucketInfo<T>> &bucketDefs, const MetricMetaData &_metaData = MetricMetaData()) :
```
ghalliday, line 279: structure: bucket defs should be part of the metric meta data (and no default parameter)

ghalliday (commented):
@kenrowland it looks broadly right.  A few different comments.

system/jlib/jmetrics.hpp
kenrowland: I opted for each bucket maintained by the metric to be the count of measurements for that bucket only. So, yes it's the responsibility of the sink to convert to the representation required. Additionally, the accumulation only needs to be done at report time reducing the overhead of recording a measurement. As far as documenting, I added a simple explanation in the metrics design doc as part of this PR.

testing/unittests/metrics/MetricFrameworkTests.cpp
kenrowland: Added a method to simplify testing the entire results.

system/jlib/jmetrics.hpp
kenrowland: There is a comment later that explains why I made the definitions separate from meta data.

devdoc/Metrics.rst
```diff
@@ -290,6 +290,19 @@ metric value as the custom metric class provides no update methods. The componen
...
+-----------------
+Records counts of measurements according to defined bucket limits. When created, the caller defines
+as set of bucket limits. During event recording, the component records measurements. The metric
+separates each recorded measurement into its bucket by testing the measurement value against each
+bucket limit using a less than or equal test. Each bucket contains a count of measurements meeting
+that criteria. Additionally, the metric maintains a default bucket for measurements outside of the
+maximum bucket limit. This is sometimes known as the "inf" bucket.
+
+Some storage systems, such as Prometheus, require each bucket to accumulate its measurements with
+the previous bucket(s). It is the repsonsbility of the sink to accumulate values as needed.
```
ghalliday, line 303: typo responsibility

system/jlib/jmetrics.hpp
ghalliday: Have you tried again now it is no longer a template - it compiles for me.

testing/unittests/metrics/MetricFrameworkTests.cpp
ghalliday: Yes, the sink will need to know it is a histogram.  I don't have a strong opinion on the interface, but the minimum would be something like

__uint64 queryValue();// returning the sum of the values and
std::vector<__uint64> getHistogramValues(); // returning the counts.  (The total count could be returned by a separate function, or caulcated by summing all these values.)

system/jlib/jmetrics.hpp
ghalliday: Sorry for the confusion, I had forgotten what the meta data was used for.

Reporting the metrics isn't called very often, so the efficiency is not that important, but cloning the string labels each time is less efficient.

kenrowland:
@ghalliday 
Addressed comments. I opted for three methods to return values from the metric:
queryValue - returns the total of all measurements which is inline with what other metrics do
querySum - Total sum of counts (all bucket counts)
queryHistogramValues - individual bucket counts (including inf)

The caller could sum the counts if needed instead of calling querySum

I also separated labels from the histogram values. There is a queryLabels method now. The intent is the sink would call this only  once to prefill a structure used to report values. This way the labels are not constantly recreated.

Version of my compiler still doesn't like not specifying the structure type in the emplace_back. In the end it's more readable anyway.

testing/unittests/metrics/MetricFrameworkTests.cpp
```diff
@@ -308,6 +309,65 @@ class MetricFrameworkTests : public CppUnit::TestFixture
...
+        // Just to make sure, increment another bucket
+        pHistogram->recordMeasurement(8);
+        sumMeasurements += 8;
+        countMeasurements++;
+        checkHistogramBucketResult(pHistogram, sumMeasurements, countMeasurements, {0,2,1,1});
+    }
+
+
+    void checkHistogramBucketResult(std::shared_ptr<HistogramMetric> pHistogram, int expectedSum, int expectedCount, const std::vector<int> &expectedValues)
```
ghalliday, line 359: This function needs to take an IMetric as a parameter - so it can match the logic that the sinks need to execute.  (The functions are not currently accessible to the sinks...)

ghalliday (changes requested):

testing/unittests/metrics/MetricFrameworkTests.cpp
kenrowland: Converted parameter to IMetric as requested. Add dynamic pointer cast to histogram metric type to gain access to histogram specific methods. 

It is expected that a sink will check the type of each metric in it's report list and for any encountered histograms, execute a dynamic pointer cast to convert the pointer type and gain access to the necessary methods.

ghalliday (changes requested):
This isn't what I expected - it is not well encapsulated.

The sinks should only call functions in the IMetric interface - they should know nothing about the implementation classes.  (Ideally the implementations would be in separate header files that are not even included by the sink code.)

Also, please do not squash commits until the PR has been approved/someone requests you squash it.

kenrowland:
Added histogram methods to metric interface. Base class implementation returns empty values.

system/jlib/jmetrics.hpp
```diff
@@ -237,6 +249,56 @@ class CustomMetric : public MetricBase
...
+{
+    Bucket(const std::string &_label, __uint64 _limit) :
+        label{_label}, limit{_limit}, count{0} {}
+    std::string label;
+    __uint64 limit;
+    __uint64 count;
+};
+
+
+struct BucketValue
```
ghalliday, line 269: This can now be removed.

system/jlib/jmetrics.hpp
```diff
@@ -110,8 +120,10 @@ class jlib_decl MetricBase : public IMetric
     virtual const std::string &queryName() const override { return name; }
     virtual const std::string &queryDescription() const override { return description; }
     virtual MetricType queryMetricType() const override { return metricType; }
-    const MetricMetaData &queryMetaData() const { return metaData; }
+    const MetricMetaData &queryMetaData() const override { return metaData; }
     StatisticMeasure queryUnits() const override { return units; }
+    virtual std::vector<__uint64> queryHistogramValues() const { return {}; }
```
ghalliday, line 125: trivial: can also be override

system/jlib/jmetrics.hpp
```diff
@@ -237,6 +249,56 @@ class CustomMetric : public MetricBase
...
+public:
+    HistogramMetric(const char *name, const char *desc, StatisticMeasure _units, const std::vector<BucketDef> &bucketDefs, const MetricMetaData &_metaData = MetricMetaData());
+
+    virtual __uint64 queryValue() const override
+    {
+        return sum;
+    }
+
+    void recordMeasurement(__uint64 measurement);
+    std::vector<__uint64> queryHistogramValues() const;
```
ghalliday, line 288: minor: virtual and override for clarity (and next function)

ghalliday (commented):
@kenrowland looks good.  A few final minor comments.

kenrowland:
Updated based on recent comments.

testing/unittests/metrics/MetricFrameworkTests.cpp
```diff
@@ -308,6 +309,62 @@ class MetricFrameworkTests : public CppUnit::TestFixture
...
+        // Just to make sure, increment another bucket
+        pHistogram->recordMeasurement(8);
+        sumMeasurements += 8;
+        checkHistogramBucketResult(pHistogram, sumMeasurements, {0,2,1,1});
+    }
+
+
+    void checkHistogramBucketResult(std::shared_ptr<IMetric> pMetric, int expectedSum, const std::vector<int> &expectedValues)
+    {
+        std::shared_ptr<HistogramMetric> pHistogram;
```
ghalliday, line 356: minor: These lines should be deleted, and the pMetric used - so this code will mirror the calls the sink will perform.

ghalliday (commented):
@kenrowland looks good.  One last bit of tidying up on the test case to ensure the interface is correct.

testing/unittests/metrics/MetricFrameworkTests.cpp
kenrowland: Removed cast to histogram pointer type in favor of IMetric interface use only.

ghalliday (approved):

ghalliday:
@kenrowland merged, but then noticed use had the wrong issue number in your commit message.  Should have been HPCC-27962.  Will have to stand as-is.

PR #16426, opened 2022-07-29
HPCC-28069 Default logging to non-dropping mode in K8s.

jakesmith:
To match bare-metal behaviour, logs should not by default drop
logs when queue hits the limit (default behaviour is to block).
In bare-metal - there are couple of exceptions, roxie and
the topology server configure logging to drop logs when queue
is full rather than block.
Preserve that behaviour via default configuration.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/helm/hpcc/values.schema.json
+++ b/helm/hpcc/values.schema.json
@@ -928,6 +928,16 @@
         "detail": {
           "type": "integer",
           "description": "Log output verbosity"
+        },
+        "queueLength": {
+          "type": "integer",
+          "description": "Maximum number of log entries to buffer in the log queue. Set to 0 for synchronous logging",
+          "default": 512
+        },
+        "queueDrop": {
+          "type": "integer",
+          "description": "Number of log entries to drop from the log queue when it is full. Disabled if 0, or if queueLength is 0",
+          "default": 0
         }
       },
       "additionalProperties": { "type": ["integer", "string", "boolean"] }
--- a/roxie/roxie/roxie.cpp
+++ b/roxie/roxie/roxie.cpp
@@ -52,7 +52,6 @@ roxie:
   numChannels: 1
   queueNames: roxie.roxie
   logging:
-    detail: 50
     queueDrop: 32
 )!!";
 
--- a/roxie/roxie/roxie.cpp
+++ b/roxie/roxie/roxie.cpp
@@ -51,6 +51,9 @@ roxie:
   localSlave: true
   numChannels: 1
   queueNames: roxie.roxie
+  logging:
+    detail: 50
+    queueDrop: 32
 )!!";
 
 int main(int argc, const char *argv[])
--- a/system/jlib/jlog.cpp
+++ b/system/jlib/jlog.cpp
@@ -2398,21 +2398,14 @@ static constexpr const char * logFieldsAtt = "@fields";
 static constexpr const char * logMsgDetailAtt = "@detail";
 static constexpr const char * logMsgAudiencesAtt = "@audiences";
 static constexpr const char * logMsgClassesAtt = "@classes";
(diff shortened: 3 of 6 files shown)


helm/hpcc/values.schema.json
```diff
@@ -928,6 +928,16 @@
         "detail": {
           "type": "integer",
           "description": "Log output verbosity"
+        },
```
rpastrana, line 931: Might be helpful to document these new options in the README, and PDF
[queueDropDefault ](https://github.com/hpcc-systems/HPCC-Platform/tree/master/helm/examples/logging#hpcc-systems-application-level-logging-details)

system/jlib/jlog.cpp
```diff
@@ -2453,14 +2446,21 @@ void setupContainerizedLogMsgHandler()
...
-            unsigned queueLen = logConfig->getPropInt(logQueueLenAtt, queueLenDefault);
-            unsigned queueDrop = logConfig->getPropInt(logQueueDropAtt, queueDropDefault);
-
             queryLogMsgManager()->enterQueueingMode();
-            queryLogMsgManager()->setQueueDroppingLimit(queueLen, queueDrop);
+            unsigned queueDrop = logConfig->getPropInt(logQueueDropAtt, queueDropDefault);
+            if (queueDrop)
+            {
+                queryLogMsgManager()->setQueueDroppingLimit(queueLen, queueDrop);
+                PROGLOG("JLog: queuing enabled with dropping: queueLen=%u, queueDrop=%u", queueLen, queueDrop);
```
rpastrana, line 2457: somewhat subjective and likely to change, but this is how PROLOG and INFOLOG have been defined:
    MSGCLS_information = 0x08, // Config, environmental and internal status  info
    MSGCLS_progress    = 0x10, // Progress of workunits. Status of file operations
https://github.com/hpcc-systems/HPCC-Platform/blob/a1240fda33e1110602607e5ab876ab4700b17238/system/jlib/jlog.hpp#L66

rpastrana (commented):
@jakesmith see comments

roxie/roxie/roxie.cpp
```diff
@@ -51,6 +51,9 @@ version: "1.0"
   localSlave: true
   numChannels: 1
   queueNames: roxie.roxie
+  logging:
+    detail: 50
```
ghalliday, line 55: Why 50?

ghalliday (commented):
One question about detail level for roxie.

system/jlib/jlog.cpp
jakesmith: Fair enough, however, PROGLOG is used widely for this purpose, I'd say it is used more for 'internal status'/progress than it is for "Progress of workunits. Status of file operations", but as you say, all a bit subjective.

In fact, it looks like they're being used the opposite way around from what the comments pescribe.
MCuserInfo is being used in 22 places at the moment, and they're mostly related to workunit/files.
PROGLOG is used in 1,624 places, mostly for internal status info purposes.

Perhaps we should flip the comments in jlog.hpp at this stage? (not this PR)

helm/hpcc/values.schema.json
jakesmith: Will add to the readme along with the others, but they're currently in helm/examples/logging/README.md.
examples/logging makes sense for setting up the logging stack setup, but the "logging:" settings would be better placed in helm/hpcc/README.md I think, as they are pertinent with or without a logging stack configuration.

ghalliday (approved):

PR #16442, opened 2022-08-04
HPCC-28059 Validate requests before searching WUs and queries

wangkx:
The code is added to make sure that search requests do not have
invalid character as xpath values. The changes in this PR focus
on the following ESP services:
1. WsWorkunits.WUQuery
2. WsWorkunits.WULightWeightQuery
3. WsWorkunits.WUListQueries
4. WsWorkunits.WUQueryDetails
5. WsWorkunits.WUQueryDetailsLightWeight
6. FileSpray.GetDFUWorkunits

The code is also added to validate the workunit ID in
FileSpray.GetDFUWorkunit request.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/services/ws_fs/ws_fsService.cpp
+++ b/esp/services/ws_fs/ws_fsService.cpp
@@ -939,6 +939,7 @@ bool CFileSprayEx::onGetDFUWorkunits(IEspContext &context, IEspGetDFUWorkunits &
     try
     {
         context.ensureFeatureAccess(DFU_WU_URL, SecAccess_Read, ECLWATCH_DFU_WU_ACCESS_DENIED, "Access to DFU workunit is denied.");
+        CFileSprayValidateHelper::validateGetDFUWorkunitsRequest(req);
 
         StringBuffer wuidStr(req.getWuid());
         const char* wuid = wuidStr.trim().str();
@@ -1392,6 +1393,8 @@ bool CFileSprayEx::onGetDFUWorkunit(IEspContext &context, IEspGetDFUWorkunit &re
         const char* wuid = req.getWuid();
         if (!wuid || !*wuid)
             throw MakeStringException(ECLWATCH_INVALID_INPUT, "Dfu workunit ID not specified.");
+        if (!looksLikeAWuid(wuid, 'D'))
+            throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "GetDFUWorkunit: Invalid Workunit ID: %s", wuid);
 
         bool found = false;
         double version = context.getClientVersion();
@@ -3676,4 +3679,20 @@ bool CFileSprayEx::onGetDFUServerQueues(IEspContext &context, IEspGetDFUServerQu
     }
 
     return true;
+}
+
+void CFileSprayValidateHelper::validateGetDFUWorkunitsRequest(IEspGetDFUWorkunits& req)
+{
+    if (!isEmptyString(req.getWuid()) && !looksLikeAWuid(req.getWuid(), 'D'))
+        throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "GetDFUWorkunits: Invalid Workunit ID: %s", req.getWuid());
+    if (!isEmptyString(req.getOwner()) && !isValidXPathValue(req.getOwner()))
+        throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "GetDFUWorkunits: Invalid Owner %s.", req.getOwner());
+    if (!isEmptyString(req.getCluster()) && !isValidXPathValue(req.getCluster()))
+        throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "GetDFUWorkunits: Invalid Cluster %s.", req.getCluster());
+    if (!isEmptyString(req.getStateReq()) && !isValidXPathValue(req.getStateReq()))
+        throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "GetDFUWorkunits: Invalid StateReq %s.", req.getStateReq());
+    if (!isEmptyString(req.getJobname()) && !isValidXPathValue(req.getJobname()))
+        throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "GetDFUWorkunits: Invalid Jobname %s.", req.getJobname());
+    if (!isEmptyString(req.getPublisherWuid()) && !isValidXPathValue(req.getPublisherWuid()))
+        throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "GetDFUWorkunits: Invalid PublisherWuid %s.", req.getPublisherWuid());
 }
--- a/esp/services/ws_fs/ws_fsService.cpp
+++ b/esp/services/ws_fs/ws_fsService.cpp
@@ -3693,6 +3693,6 @@ void CFileSprayValidateHelper::validateGetDFUWorkunitsRequest(IEspGetDFUWorkunit
         throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "GetDFUWorkunits: Invalid StateReq %s.", req.getStateReq());
     if (!isEmptyString(req.getJobname()) && !isValidXPathValue(req.getJobname()))
         throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "GetDFUWorkunits: Invalid Jobname %s.", req.getJobname());
-    if (!isEmptyString(req.getPublisherWuid()) && !isValidXPathValue(req.getPublisherWuid()))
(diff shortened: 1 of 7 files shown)


wangkx:
@afishbeck @ghalliday please review, especially on: 1. the contents of the validXPathValueChrs; 2. any other validatation should be added (in this PR or in future's PR).

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -3677,3 +3680,19 @@ bool CFileSprayEx::onGetDFUServerQueues(IEspContext &context, IEspGetDFUServerQu
...
+        throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "GetDFUWorkunits: Invalid Workunit ID: %s", req.getWuid());
+    if (!isEmptyString(req.getOwner()) && !isValidXPathValue(req.getOwner()))
+        throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "GetDFUWorkunits: Invalid Owner %s.", req.getOwner());
+    if (!isEmptyString(req.getCluster()) && !isValidXPathValue(req.getCluster()))
+        throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "GetDFUWorkunits: Invalid Cluster %s.", req.getCluster());
+    if (!isEmptyString(req.getStateReq()) && !isValidXPathValue(req.getStateReq()))
+        throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "GetDFUWorkunits: Invalid StateReq %s.", req.getStateReq());
+    if (!isEmptyString(req.getJobname()) && !isValidXPathValue(req.getJobname()))
+        throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "GetDFUWorkunits: Invalid Jobname %s.", req.getJobname());
+    if (!isEmptyString(req.getPublisherWuid()) && !isValidXPathValue(req.getPublisherWuid()))
```
afishbeck, line 3696: I think for PublisherWuid you can do  !looksLikeAWuid(req.getPublisherWuid(), 'P').

wangkx: I just made the change.

system/jlib/jptree.hpp
```diff
@@ -313,6 +313,22 @@ inline static bool isValidXPathChr(char c)
     return ('\0' != c && (isalnum(c) || strchr(validChrs, c)));
 }
 
+static const char *validXPathValueChrs = "!#$%^&()+,-.:;?@_{|}~"; //Do I miss any?
```
afishbeck, line 316: Hmmm.  The level of validation could and probably should be a lot stricter.  For each request parameter we should first ask a few questions:
1. Is it ever stored in DALI/PTree as an element?  If so the allowed characters are a lot less.
2. Is it stored in Ptree as a string?
3. Can it be treated as a restricted name type... with a limited range of characters.
4. Is the parameter allowed to have wild cards?

For checking xpath elements rather than value strings there are existing methods in ptree.cpp
inline static bool isValidXPathStartChr(char c)
{
    return ('\0' != c && (isalpha(c) || strchr(validStartChrs, c)));
}

static const char *validChrs = ":_.-";
inline static bool isValidXPathChr(char c)
{
    return ('\0' != c && (isalnum(c) || strchr(validChrs, c)));
}

For a lot of parameters that are names... cluster names, etc.  We can be very strict.  Perhaps only allowing alphanum and '.' and '-' etc (what others?)?

afishbeck: If it is stored as an element we might not even allow ':' which would act like a namespace... although ptree could still handle it so it may be ok depending on what it is.

afishbeck: You could have at least 3 functions:

isValidXMLElementName(char *s);
isValidXPathValue(char *s);
isValidName(char *s, const char *extraAllowedCharacters);

example:
isValidName("my-secret.101", "-.");  //returns true
isValidName("my-secret.101[1]", "-.");  //returns false

system/jlib/jptree.hpp
```diff
@@ -313,6 +313,22 @@ inline static bool isValidXPathChr(char c)
...
+inline static bool isValidXPathValueChr(char c)
+{
+    return ('\0' != c && (isalnum(c) || strchr(validXPathValueChrs, c)));
+}
+inline static bool isValidXPathValue(const char *v)
+{
+    const char *p = v;
```
afishbeck, line 323: I don't think you actually need p.  The following should work:
while (*v)
{
     if (!isValidXPathValueChr(*v++)) return false;
}

system/jlib/jptree.hpp
wangkx: Good comments. I thought about to create this PR as a draft to collect comments.
I name it as validXPathValueChrs and, in this PR, it is only used to verify an xpath value: xpath='**value**'. I haven't thought about validating elements and names yet.

wangkx: All of the functions can have the default allowed characters and extraAllowedCharacters. The challenge is which characters should be in the default allowed.

system/jlib/jptree.hpp
wangkx: Yes, I do not need p.
The do...while is used to check an empty v.

wangkx:
@ghalliday please review.

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -4908,4 +4908,56 @@ void CWsWuEmailHelper::send(const char* body, const void* attachment, size32_t l
...
+        if (!isEmptyString(item.getName()) && !isValidName(item.getName(), nullptr))
+            throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "%s: Application %s: Invalid Application Name %s.", method, item.getApplication(), item.getName());
+        if (!isEmptyString(item.getValue()) && !isValidXPathValue(item.getValue()))
+            throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "%s: Application %s: Invalid Application Value %s.", method, item.getApplication(), item.getValue());
+    }
+}
+
+void CWsWuValidateHelper::validateWUListQueriesRequest(IEspWUListQueriesRequest& req)
+{
+    if (!isEmptyString(req.getQuerySetName()) && !isValidXPathValue(req.getQuerySetName()))
```
ghalliday, line 4949: One further follow on comment.  It would simplify the code (and make it easier to read) if each isvalidX() function had a bool required.  E.g.

```
bool isValidQuerySetName(const char * name, bool required)
{
   if (!name || *name == '\0')
      return !required;
...
}

```
Then the call will be 

`if (isValidQuerySetName(req.querySetName(), false))
`
which is much easier to read and check.

ghalliday (commented):
@wangkx A more significant comment than I expected to make.  Feel free to discuss.

wangkx:
Thanks for the comments which suggest the checks from another perspective: logical entities. Different logical entities may have different restrictions. Creating an isValid...() for each logical entity may be better for checking those restrictions. Of course, it may take time to design/create those isValid...().

esp/services/ws_workunits/ws_workunitsHelpers.cpp
wangkx: Should we add another bool allowWild?

esp/services/ws_workunits/ws_workunitsHelpers.cpp
```diff
@@ -4908,4 +4908,56 @@ void CWsWuEmailHelper::send(const char* body, const void* attachment, size32_t l
             attachmentName.get(), mailServer.get(), port, sender.get(), &warnings);
 }
 
+void CWsWuValidateHelper::validateWUQueryRequest(IEspWUQueryRequest& req)
+{
+    validateWUQueryRequestCommon("WUQuery", req);
+    if (!isEmptyString(req.getJobname()) && !isValidXPathValue(req.getJobname()))
+        throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "WUQuery: Invalid Jobname %s.", req.getJobname());
+    if (!isEmptyString(req.getECL()) && !isValidXPathValue(req.getECL()))
```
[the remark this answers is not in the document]
wangkx: ```searching for the text "\ should produce a filter on \"\\```

Is the issue reported in the JIRA caused by incorrect encoding? One exception there is: 
XPath Exception: Qualifier brace unclosed\nin xpath = [@jobName=?\"*\"AAA*\"]
Could the extra `\"` cause the exception (the ] is not inside quotes)?

wangkx:
Redesign is needed for this PR. I am closing this PR for now.

PR #16455, opened 2022-08-10
HPCC-28085 enable/disable scope scans is broke

RussWhitehead:
ESP to serialize user's session token in Enable/Disable ScopeScans calls to
Dali. Dali ensures session token is present if UN/PWd and digital signature
is not present

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [x] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dali/server/daldap.cpp
+++ b/dali/server/daldap.cpp
@@ -33,6 +33,7 @@ using namespace cryptohelper;
 #ifndef _NO_LDAP
 #include "seclib.hpp"
 #include "secloader.hpp"
+#include "ldapsecurity.ipp"
 #include "ldapsecurity.hpp"
 
 static void ignoreSigPipe()
@@ -225,9 +226,10 @@ public:
 
         if (!authenticated)
         {
-            user->credentials().setPassword(password);
-            if (!ldapsecurity->authenticateUser(*user, &superUser) || !superUser)
+            CLdapSecManager* ldapSecMgr = dynamic_cast<CLdapSecManager*>(ldapsecurity.get());
+            if (!ldapSecMgr || !ldapSecMgr->isSuperUser(user))
             {
+                DBGLOG("LDAP: EnableScopeScans caller %s must be an LDAP HPCC Admin", username.str());
                 *err = -1;
                 return false;
             }
--- a/system/security/LdapSecurity/ldapsecurity.cpp
+++ b/system/security/LdapSecurity/ldapsecurity.cpp
@@ -1556,6 +1556,7 @@ bool CLdapSecManager::clearPermissionsCache(ISecUser& user, IEspSecureContext* s
             return false;
         }
         m_permissionsCache->flush();
+        PROGLOG("Permissions cache cleared by admin user %s", user.getName());
     }
     return true;
 }


wangkx (approved):
@RussWhitehead looks fine.

RussWhitehead:
@richardkchapman @ghalliday Please merge

system/security/LdapSecurity/ldapsecurity.cpp
```diff
@@ -1556,6 +1556,7 @@ bool CLdapSecManager::clearPermissionsCache(ISecUser& user, IEspSecureContext* s
             return false;
         }
         m_permissionsCache->flush();
+        PROGLOG("Permissions cache cleared by admin user %s", user.getName());
```
jakesmith, line 1559: minor: not really related to the purpose of this PR I assume?

jakesmith (commented):
@RussWhitehead - I'm unclear why enable scope scans needs to authenticate like this, when as far as I understand it, elsewhere e.g. when getPermissionsLDAP->CDaliLdapConnection::getPermissions is called, it only has a username and asserts authenticated already ?

@afishbeck @ghalliday

system/security/LdapSecurity/ldapsecurity.cpp
RussWhitehead: In the Jira,Fern  mentioned that Clear Permissions Cache didn't log anything, so I added this. If you prefer I could create a separate JIra/PR

RussWhitehead:
@jakesmith I was not part of the effort that removed the request signature and timestamp from the permission calls a while back,  so I am also not clear on what the trust-level assumption is in place between the caller and Dali in that call.
As for this call and this Jira, I wanted to ensure the caller has been authenticated, and since ESP no longer caches passwords the session token seems to be the best way

jakesmith:
> @RussWhitehead - I'm unclear why enable scope scans needs to authenticate like this, when as far as I understand it, elsewhere e.g. when getPermissionsLDAP->CDaliLdapConnection::getPermissions is called, it only has a username and asserts authenticated already ?
> 
> @afishbeck @ghalliday

Added Gavin and Tony as reviewers to follow up for comment.

ghalliday:
I am also not clear why dali is doing any authentication - esp will already have performed it and dali should be locked down so that nothing outside the cluster can access it.

It makes some sense to check that the user has the rights to perform the action (or is a super user), but that should be independent of authentication.  My impression of scanning the code is that setting the session token will force it to be marked as authenticated (and not actually check the session token) so it is effectively removing the authentication anyway.

There should be some way of documenting this as part of a threat modelling overview.
@afishbeck thoughts?

afishbeck:
@ghalliday @jakesmith Agreed that Dali shouldn't be authenticating the user.  DALI should be able to trust ESP and ESP would have authenticated the user.  If DALI isn't authenticating ESP well enough that's a separate issue, which is vastly improved in the cloud when using MTLS and because of DALI isolation.

dali/server/daldap.cpp
```diff
@@ -226,6 +226,7 @@ class CDaliLdapConnection: implements IDaliLdapConnection, public CInterface
         if (!authenticated)
         {
             user->credentials().setPassword(password);
+            user->credentials().setSessionToken(udesc->querySessionToken());
```
ghalliday, line 229: @RussWhitehead I think the suggested fix is that this is changed to

    if (!user->isSuperUser()) ....

with no authentication.  (I think it is equivalent to the changes in this PR.)

jakesmith (commented):
@RussWhitehead - no further comments beyond what has been said above.

dali/server/daldap.cpp
```diff
@@ -225,9 +226,10 @@ class CDaliLdapConnection: implements IDaliLdapConnection, public CInterface
 
         if (!authenticated)
         {
-            user->credentials().setPassword(password);
-            if (!ldapsecurity->authenticateUser(*user, &superUser) || !superUser)
+            CLdapSecManager* secmgr = dynamic_cast<CLdapSecManager*>(ldapsecurity.get());
```
ghalliday, line 229: Rather than a dynamic cast, it would be cleaner if isSuperUser() was added to the interface - it is logically a generally useful function.

ghalliday (commented):
Looks good.  One comment to improve the encapsulation of the code.

jakesmith (approved):
Nothing further to add, looks good, but you may want to address Gavin's last comment

dali/server/daldap.cpp
RussWhitehead: @ghalliday I agree that adding this to ISecManager would be better in some ways, it's bad in others. Specifically, this would require implementation in all of the HPCC Security managers (not too big of a deal), but more problematic is that external security managers like the SCAPPS team would be broken until they add this method as well.  Every time I modify this interface I get an earful from @timothyklemm .

timothyklemm: @RussWhitehead I'm not opposed to changes. I just like to know about them as they're happening so I can be prepared for them.

As for this change, I agree with the general usefulness. My concern is an interface change to point releases. At least with this interface, we've generally tried to limit interface changes to future major or minor releases. That this is targeting a prior minor version adds to the complexity of managing external builds.

If it has to happen, it has to happen. It just creates headaches for external builds.

ghalliday (approved):
That is a good point.  I agree this change should go in as it is.  We should consider adding the new virtual in 8.10.0.

dali/server/daldap.cpp
```diff
@@ -225,9 +226,10 @@ class CDaliLdapConnection: implements IDaliLdapConnection, public CInterface
...
+            if (!secmgr->isSuperUser(user))
```
afishbeck, line 230: @RussWhitehead the dynamic cast may have failed, in which case, for now at least I think access should be denied.

So perhaps "if (!secmgr || !secmgr->isSuperUser(user))"

Also, personally I would call the variable ldapmgr since you've casted to an ldap specific interface/class.

RussWhitehead: @afishbeck Done

afishbeck (approved):

dali/server/daldap.cpp
RussWhitehead: @ghalliday @timothyklemm Opened https://track.hpccsystems.com/browse/HPCC-28283 for adding to ISecManager

RussWhitehead:
@richardkchapman @ghalliday Please merge

PR #16458, opened 2022-08-14
HPCC-28125 MP accept from client should use peer IP and not IP sent from client

mckellyln:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [x] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/mp/mpcomm.cpp
+++ b/system/mp/mpcomm.cpp
@@ -2237,6 +2237,37 @@ int CMPConnectThread::run()
                 connectHdr.id[0].get(_remoteep);
                 connectHdr.id[1].get(hostep);
 
+                // HPCC-28125 - if remoteep IP != peerEp IP then log warning - do we change remoteep ?
+                //              if hostep != listen ep then log warning - do we change hostep ?
+                // **VERY CONCERNED** about pods and local IPs and Network Address Translation
+
+                if (!_remoteep.ipequals(peerEp))
+                {
+                    StringBuffer rIp, pIp;
+                    _remoteep.getIpText(rIp);
+                    peerEp.getIpText(pIp);
+                    PROGLOG("MP: WARNING: remote client address (%s:%u) != peer address (%s:%u)", rIp.str(), _remoteep.port, pIp.str(), peerEp.port);
+                    _remoteep.set(peerEp);
+                    connectHdr.id[0].set(_remoteep);
+                    // should we sock->close() and continue ?
+                }
+
+                SocketEndpoint listenEp;
+                listensock->getEndpoint(listenEp);
+                unsigned short lPort = listenEp.port;
+                IpAddress myLocalIp;
+                GetHostIp(myLocalIp);
+                if (!hostep.ipequals(myLocalIp))
+                {
+                    StringBuffer hIp, lIp;
+                    hostep.getIpText(hIp);
+                    myLocalIp.getIpText(lIp);
+                    PROGLOG("MP: WARNING: remote client trying to reach (%s:%u) != listen address (%s:%u)", hIp.str(), hostep.port, lIp.str(), lPort);
+                    hostep.set(lPort, myLocalIp);
+                    connectHdr.id[1].set(hostep);
+                    // should we sock->close() and continue ?
+                }
+
                 unsigned __int64 addrval = DIGIT1*connectHdr.id[0].ip[0] + DIGIT2*connectHdr.id[0].ip[1] + DIGIT3*connectHdr.id[0].ip[2] + DIGIT4*connectHdr.id[0].ip[3] + connectHdr.id[0].port;
 #ifdef _TRACE
                 PROGLOG("MP: Connect Thread: addrval = %" I64F "u", addrval);
--- a/system/mp/mpcomm.cpp
+++ b/system/mp/mpcomm.cpp
@@ -2118,9 +2118,16 @@ void CMPConnectThread::startPort(unsigned short port)
 
 int CMPConnectThread::run()
 {
+    SocketEndpoint listenEp;
+    listensock->getEndpoint(listenEp);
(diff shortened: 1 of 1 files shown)


mckellyln:
I am VERY concerned about this change.
But let me know your thoughts.

mckellyln:
This all came about because Dali would log:
```
    void onClose(SocketEndpoint &ep)
    {
        StringBuffer clientStr;
        PROGLOG("Client closed (%s)", ep.getUrlStr(clientStr).str());
```
And the client IP was not correct, suspicious and causing alarm when a non-MP client (ssh) tried to connect.

mckellyln:
The unittest crash is not related to this change, but it is from the change in https://track.hpccsystems.com/browse/HPCC-28102
I will submit a PR fix for that.

system/mp/mpcomm.cpp
```diff
@@ -2237,6 +2237,37 @@ int CMPConnectThread::run()
...
+                    peerEp.getIpText(pIp);
+                    PROGLOG("MP: WARNING: remote client address (%s:%u) != peer address (%s:%u)", rIp.str(), _remoteep.port, pIp.str(), peerEp.port);
+                    _remoteep.set(peerEp);
+                    connectHdr.id[0].set(_remoteep);
+                    // should we sock->close() and continue ?
+                }
+
+                SocketEndpoint listenEp;
+                listensock->getEndpoint(listenEp);
+                unsigned short lPort = listenEp.port;
```
jakesmith, line 2257: this is in order to get thte port of the listening socket..
rather than call getEndpoint which entails a system call etc., it would probably be better to store the port in the constructor as a member and reuse it here.

system/mp/mpcomm.cpp
```diff
@@ -2237,6 +2237,37 @@ int CMPConnectThread::run()
...
+                IpAddress myLocalIp;
+                GetHostIp(myLocalIp);
```
jakesmith, line 2259: this is going to be done per accept, can we move it to do once in CMPServer ? CMPConnectThread::run can access it via 'parent'

system/mp/mpcomm.cpp
```diff
@@ -2237,6 +2237,37 @@ int CMPConnectThread::run()
...
+
+                if (!_remoteep.ipequals(peerEp))
+                {
+                    StringBuffer rIp, pIp;
+                    _remoteep.getIpText(rIp);
+                    peerEp.getIpText(pIp);
+                    PROGLOG("MP: WARNING: remote client address (%s:%u) != peer address (%s:%u)", rIp.str(), _remoteep.port, pIp.str(), peerEp.port);
+                    _remoteep.set(peerEp);
+                    connectHdr.id[0].set(_remoteep);
+                    // should we sock->close() and continue ?
```
jakesmith, line 2252: Need to test/confirm/try to think of edge cases, but this change looks like the correct fix to me.
It is the peerEp that Dali should be using, not the what the client's notion of what its IP is

system/mp/mpcomm.cpp
```diff
@@ -2237,6 +2237,37 @@ int CMPConnectThread::run()
...
+                }
+
+                SocketEndpoint listenEp;
+                listensock->getEndpoint(listenEp);
+                unsigned short lPort = listenEp.port;
+                IpAddress myLocalIp;
+                GetHostIp(myLocalIp);
+                if (!hostep.ipequals(myLocalIp))
```
jakesmith, line 2260: this is going to be the norm. in the cloud I think (haven't confirmed), and not that unexpected in general.

I can see why the remoteep vs peerEp is a problem that will cause clashes etc.
That needs fixing and I think that fix looks correct.
But I'm not sure what problems the client's notion of what the Dali IP is, vs what Dali itself knows its IP is, is problematic?

system/mp/mpcomm.cpp
```diff
@@ -2237,6 +2237,37 @@ int CMPConnectThread::run()
...
+                {
+                    StringBuffer hIp, lIp;
+                    hostep.getIpText(hIp);
+                    myLocalIp.getIpText(lIp);
+                    PROGLOG("MP: WARNING: remote client trying to reach (%s:%u) != listen address (%s:%u)", hIp.str(), hostep.port, lIp.str(), lPort);
+                    hostep.set(lPort, myLocalIp);
```
jakesmith, line 2266: see comment above, I'm not sure this (client notion of Dali IP being different from Dali's notion of its IP) actually causes any problems...
but I am wondering if changing this (to Dali's notion of itself) may cause problems for the clients that will no longer see a match in packets coming back.

Probably pretty obvious if it breaks things on a cloud system where they will mismatch..

jakesmith (commented):
@mckellyln - added some comments/questions

system/mp/mpcomm.cpp
mckellyln: The concern I have is the port of the remoteep in a valid MP connection is really the other side's server port (how to connect to it), not the ephemeral port in the peerEp and so should we change it here ?

system/mp/mpcomm.cpp
mckellyln: Yes, I can look into that.

system/mp/mpcomm.cpp
mckellyln: Yes, good idea.

system/mp/mpcomm.cpp
mckellyln: I agree, I am not sure changing hostep matters or causes problems.  Lets test and decide what is best.

mckellyln:
Closing for now until we have some more information about MP clients and NAT

PR #16461, opened 2022-08-15
HPCC-28128 First set of improvements to cse code

ghalliday:
## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/ecl/hqlcpp/hqlcse.cpp
+++ b/ecl/hqlcpp/hqlcse.cpp
@@ -60,6 +60,7 @@ bool canCreateTemporary(IHqlExpression * expr)
     switch (expr->getOperator())
     {
     case no_range:
+    case no_rowsetrange:
     case no_rangefrom:
     case no_rangeto:
     case no_rangecommon:
@@ -263,6 +264,9 @@ void CseSpotter::analyseExpr(IHqlExpression * expr)
     case no_newtransform:
     case no_range:
     case no_rangefrom:
+    case no_createrow:
+    case no_list:
+    case no_createdictionary:
         if (expr->isConstant())
             return;
         break;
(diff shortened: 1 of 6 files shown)


ghalliday:
This is submitted as 4 separate commits, because each one is independent and separately tested.
All but the last have no effect on the generated code - they are fixing issues that appeared when other changes were made.  The last change improves the generated code for a set of queries.  One (tkirk29) gets slightly worse - but not enough to warrant any concerns.

ghalliday:
(Compare ignoring whitespace)

ghalliday:
Actually I remembered incorrectly, commit 1 allowed a few if activities to be combined into a case activity.  Commit 3 is the one that caused some significant improvements.

ecl/hqlcpp/hqlcse.cpp
```diff
@@ -292,26 +296,40 @@ void CseSpotter::analyseExpr(IHqlExpression * expr)
         extra->canAlias = true;
 
     bool savedCanAlias = canAlias;
-    if (expr->isDataset() && (op != no_select) && (!spotCseInIfDatasetConditions || (op != no_if)))
+    bool processed = false;
+    if (canAlias)
```
ghalliday, line 300: note: This is primarily restructuring the code.  The only actual difference is for no_createset - which prevents expressions from the underlying dataset from being aliased (which caused problems in combination of some other changes)

shamser (approved):

richardkchapman (approved):

richardkchapman:
Unit tests are reporting a core - not sure if related but should be checked.

ghalliday:
That was a path issue - the test crashes if it cannot find the libloggingmanager.so (which happened when I ran it from the build directory)

ghalliday:
@richardkchapman unit test now passing (I wonder if it was the roxie file problem).  The extra commit to improve the unit test should be harmless/beneficial.

PR #16463, opened 2022-08-15
HPCC-28103 Add convenience class to metrics to measure histogram exec…

kenrowland:
…ution times

Added scaled time from cycles histogram to improve performance of elapsed
time measurements
Added convenience class to simplify measuring and recording elapsed time using
scaled time from cycles histogram

Signed-Off-By: Ken Rowland kenneth.rowland@lexisnexisrisk.com

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Cloud-compatibility
  - [x] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jmetrics.cpp
+++ b/system/jlib/jmetrics.cpp
@@ -29,7 +29,7 @@ MODULE_EXIT()
 
 
 HistogramMetric::HistogramMetric(const char *name, const char *desc, StatisticMeasure _units, const std::vector<BucketDef> &bucketDefs, const MetricMetaData &_metaData) :
-        MetricBase(name, desc, MetricType::METRICS_HISTOGRAM, _units, _metaData)
+    MetricBase(name, desc, MetricType::METRICS_HISTOGRAM, _units, _metaData)
 {
     buckets.reserve(bucketDefs.size());
     for (auto &b: bucketDefs)
@@ -74,8 +74,7 @@ std::vector<std::string> HistogramMetric::queryHistogramLabels() const
 }
 
 
-
-Bucket &HistogramMetric::findBucket(__uint64 measurement)
+HistogramMetric::Bucket &HistogramMetric::findBucket(__uint64 measurement)
 {
     for (auto &b: buckets)
     {
@@ -86,6 +85,18 @@ Bucket &HistogramMetric::findBucket(__uint64 measurement)
     }
     return inf;
 }
+
+
+ScaledHistogramMetric::ScaledHistogramMetric(const char *name, const char *desc, StatisticMeasure units, const std::vector<BucketDef> &bucketDefs, double limitsToMeasurementUnitsScaleFactor, const MetricMetaData &_metaData) :
+    HistogramMetric{name, desc, units, bucketDefs, metaData}
+{
+    for (auto &b: buckets)
+    {
+        b.limit = (__uint64)((double)(b.limit) * limitsToMeasurementUnitsScaleFactor);
+    }
+    outputScaleFactor = 1.0 / limitsToMeasurementUnitsScaleFactor;
+}
+
 
 
 MetricsManager &hpccMetrics::queryMetricsManager()
@@ -387,6 +398,23 @@ std::shared_ptr<GaugeMetricFromCounters> hpccMetrics::registerGaugeFromCountersM
     // the vmt for the control structure does not live in memory allocated by a shared object (dll) that could be unloaded. If make_shared is used and a shared
     // object is unloaded, the vmt is also deleted which causes unpredictable results during weak pointer access.
     std::shared_ptr<GaugeMetricFromCounters> pMetric = std::shared_ptr<GaugeMetricFromCounters>(new GaugeMetricFromCounters(name, desc, units, pBeginCounter, pEndCounter, metaData));
+    queryMetricsManager().addMetric(pMetric);
+    return pMetric;
+}
+
+
+std::shared_ptr<HistogramMetric> registerHistogramMetric(const char *name, const char* desc, StatisticMeasure units, const std::vector<BucketDef> &bucketDefs, const MetricMetaData &metaData)
+{
+    std::shared_ptr<HistogramMetric> pMetric = std::shared_ptr<HistogramMetric>(new HistogramMetric(name, desc, units, bucketDefs, metaData));
+    queryMetricsManager().addMetric(pMetric);
+    return pMetric;
+}
+
+
+std::shared_ptr<ScaledHistogramMetric> registerScaledHistogramMetric(const char *name, const char* desc, StatisticMeasure units, const std::vector<BucketDef> &bucketDefs,
+                                                                               double limitsToMeasurementUnitsScaleFactor, const MetricMetaData &metaData)
+{
+    std::shared_ptr<ScaledHistogramMetric> pMetric = std::shared_ptr<ScaledHistogramMetric>(new ScaledHistogramMetric(name, desc, units, bucketDefs, limitsToMeasurementUnitsScaleFactor, metaData));
     queryMetricsManager().addMetric(pMetric);
     return pMetric;
 }
--- a/system/jlib/jmetrics.hpp
+++ b/system/jlib/jmetrics.hpp
@@ -256,16 +256,6 @@ struct BucketDef
 };
 
 
-struct Bucket
-{
-    Bucket(const std::string &_label, __uint64 _limit) :
-        label{_label}, limit{_limit}, count{0} {}
-    std::string label;
-    __uint64 limit;
-    __uint64 count;
-};
-
-
 class jlib_decl HistogramMetric : public MetricBase
 {
 public:
@@ -281,6 +271,14 @@ public:
     virtual std::vector<std::string> queryHistogramLabels() const override;
 
 protected:
+    struct Bucket
+    {
+        Bucket(const std::string &_label, __uint64 _limit) :
+                label{_label}, limit{_limit}, count{0} {}
+        std::string label;
+        __uint64 limit;
+        __uint64 count;
+    };
     Bucket &findBucket(__uint64 measurement);
 
 protected:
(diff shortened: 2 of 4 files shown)


system/jlib/jmetrics.cpp
```diff
@@ -88,6 +87,90 @@ Bucket &HistogramMetric::findBucket(__uint64 measurement)
 }
 
 
+ScaledTimeFromCyclesHistogram::ScaledTimeFromCyclesHistogram(const char *name, const char *desc, TimeUnits _timeUnits, const std::vector<BucketDef> &bucketDefs, const MetricMetaData &_metaData) :
```
ghalliday, line 90: The code is broadly good, but it should share almost all of the implementation with the normal histogram - not duplicate it.  

I expected a derived class with a scaling factor say `double scaleMeasureToReport` which multiplies the sum when reporting, and the limits are divided by it.  That would then modify two lines of the original code, and I would be tempted to merge it into a single class (debatable).

system/jlib/jmetrics.hpp
```diff
@@ -43,7 +43,8 @@ enum MetricType
 {
     METRICS_COUNTER,
     METRICS_GAUGE,
-    METRICS_HISTOGRAM
+    METRICS_HISTOGRAM,
+    METRICS_SCALED_TIME_HISTOGRAM
```
ghalliday, line 47: Why a new type?  The sink does not need to know it is a different implementation.

system/jlib/jmetrics.hpp
```diff
@@ -291,6 +290,45 @@ class jlib_decl HistogramMetric : public MetricBase
...
+ * Enumerates the scaled time units
+ */
+enum TimeUnits
+{
+    HISTOGRAM_TIMEUNITS_NS,
+    HISTOGRAM_TIMEUNITS_US,
+    HISTOGRAM_TIMEUNITS_MS
+};
+
+class jlib_decl ScaledTimeFromCyclesHistogram : public MetricBase
```
ghalliday, line 303: No need to restrict this to times.  It might be useful to aggregate memory as bytes but report as KB/MB.

testing/unittests/metrics/MetricFrameworkTests.cpp
```diff
@@ -348,18 +348,68 @@ class MetricFrameworkTests : public CppUnit::TestFixture
...
         checkHistogramBucketResult(pHistogram, sumMeasurements, {0,2,1,1});
+
+
+        //
+        // Test the scaled time histogram. Buckets for 500ms, 1000ms and 2000ms
+        std::vector<BucketDef> timeBuckets = {{"le 500", 500}, {"le 1000", 1000}, {"le 2000", 2000} };
+        std::shared_ptr<ScaledTimeFromCyclesHistogram> pTimeHistogram = std::make_shared<ScaledTimeFromCyclesHistogram>("requests.time", "description", TimeUnits::HISTOGRAM_TIMEUNITS_MS, timeBuckets);
+        int totalDelay = 0;
+
+        //
```
ghalliday, line 359: There should be at least 1 test that checks that the scaled value reported back is correct.

testing/unittests/metrics/MetricFrameworkTests.cpp
```diff
@@ -348,18 +348,68 @@ class MetricFrameworkTests : public CppUnit::TestFixture
...
     {
         std::vector<__uint64> values = pHistogram->queryHistogramValues();
-        CPPUNIT_ASSERT_EQUAL(static_cast<int>(pHistogram->queryValue()), expectedSum);
         auto size = expectedValues.size();
         for (unsigned i=0; i<size; ++i)
         {
             CPPUNIT_ASSERT_EQUAL(static_cast<int>(values[i]), expectedValues[i]);
         }
+
+        if (expectedSum > 0)
```
ghalliday, line 411: trivial: >= 0 would be better - I scanned the test cases for 0s and didn't see any.

ghalliday (changes requested):
@kenrowland The general approach is good, but the solution should reuse the existing code more, and be more general.

system/jlib/jmetrics.cpp
kenrowland: @ghalliday 
I debated creating a separate class vs a sub class of the current histogram. What drove me to a separate class was the type, cycle_t in this case, used to record measurements. Using the histogram record measurement method required a conversion from essentially int to uint since cycle_t is signed.

If you are OK with using  cycle_t for recording measurement through the existing record measurement method (with the implicit conversion), then I can simplify the classes into a single class and add a _setScaleFactor_ method that converts the limits and scales the accumulated sum when requested.

Please let me know your thoughts.

ghalliday: Hmm it is actually a bug that cycle_t is signed (I just created HPCC-28135).  The histogram should use a _unit64 for all measurements.

system/jlib/jmetrics.hpp
kenrowland: New type was because it was a new class. I have removed it since the cycle_t is supposed to be unsigned.

system/jlib/jmetrics.hpp
kenrowland: Times were added because it was specific to cycle_t. Since uint64 is the strict measurement type, the time units are no longer needed and have been removed.

testing/unittests/metrics/MetricFrameworkTests.cpp
kenrowland: Agreed. Changed just in case a 0 test was added.

testing/unittests/metrics/MetricFrameworkTests.cpp
kenrowland: There are tests that do that, see the asserts on totalDelay following each block. I changed each test to check a range +/- 10 to take into account any rounding. Local testing has shown the delay to always be right on, but I felt it best to test the range since execution time measurements is not an exact science.

system/jlib/jmetrics.cpp
kenrowland: Since cycle_t is to be unsigned. I refactored into a single class. Note that while cycle_t may be used now, it will be converted to unsigned with a warning in code. Once HPCC-28135 is merged, those issues will disappear.

ghalliday (commented):
@kenrowland a couple of trivial comments and a couple more significant (location of scaling and not using a reference)

system/jlib/jmetrics.hpp
```diff
@@ -376,6 +394,11 @@ jlib_decl std::shared_ptr<GaugeMetricFromCounters> registerGaugeFromCountersMetr
                                                                          const std::shared_ptr<CounterMetric> &pBeginCounter, const std::shared_ptr<CounterMetric> &pEndCounter,
                                                                          const MetricMetaData &metaData = MetricMetaData());
 
+jlib_decl std::shared_ptr<HistogramMetric> registerHistogramMetric(const char *name, const char* desc, StatisticMeasure units, const std::vector<BucketDef> &bucketDefs, const MetricMetaData &metaData = MetricMetaData());
+jlib_decl std::shared_ptr<ScaledHistogramMetric> registerScaledHistogramMetric(const char *name, const char* desc, StatisticMeasure units, const std::vector<BucketDef> &bucketDefs,
```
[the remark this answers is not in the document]
ghalliday: I don't see a comment below.

system/jlib/jmetrics.cpp
```diff
@@ -390,3 +401,20 @@ std::shared_ptr<GaugeMetricFromCounters> hpccMetrics::registerGaugeFromCountersM
...
+std::shared_ptr<HistogramMetric> registerHistogramMetric(const char *name, const char* desc, StatisticMeasure units, const std::vector<BucketDef> &bucketDefs, const MetricMetaData &metaData)
+{
+    std::shared_ptr<HistogramMetric> pMetric = std::shared_ptr<HistogramMetric>(new HistogramMetric(name, desc, units, bucketDefs, metaData));
+    queryMetricsManager().addMetric(pMetric);
+    return pMetric;
+}
+
+
+std::shared_ptr<ScaledHistogramMetric> registerScaledHistogramMetric(const char *name, const char* desc, StatisticMeasure units, const std::vector<BucketDef> &bucketDefs,
+                                                                               double scaleFactorToReport, const MetricMetaData &metaData)
```
ghalliday, line 415: The name of the parameter scaleFactorToReport should be made consistent with limitsToMeasurementUnitsScaleFactor, especially because they have the opposite meaning.

system/jlib/jmetrics.hpp
```diff
@@ -291,6 +289,24 @@ class jlib_decl HistogramMetric : public MetricBase
...
+class jlib_decl ScaledHistogramMetric : public HistogramMetric
+{
+public:
+    ScaledHistogramMetric(const char *name, const char *desc, StatisticMeasure units, const std::vector<BucketDef> &bucketDefs, double _limitsToMeasurementUnitsScaleFactor, const MetricMetaData &_metaData = MetricMetaData());
+
+    virtual __uint64 queryValue() const override
+    {
+        CriticalBlock block(cs);
+        {
+            return (__uint64)((double)sum / limitsToMeasurementUnitsScaleFactor);
```
ghalliday, line 301: trivial efficiency:  It doesn't really matter here because not called very often, but more efficient to multiply than divide.  (Could still have the same parameter, but save the inverse in the class).

ghalliday (commented):
@kenrowland a few (hopefully) final comments

system/jlib/jmetrics.hpp
kenrowland: Change the saved scale factor to outputScaleFactor and saved the inverse of the input. Converted to multiplication.

system/jlib/jmetrics.hpp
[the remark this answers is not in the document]
kenrowland: Ended up not adding the comment to which I referred. Sorry for the confusion. I think once the enum situation is resolved, we can revisit adding a specific scaled histogram for milliseconds. Perhaps a set of getNnnnScaleFactor convenience functions could be added.

ghalliday (approved):
I think this is good enough to merge.  Any subsequent PR needs to call a function that does not pass in a scale for the timing histogram, but that can be delayed to a subsequent PR.  Please squash.

kenrowland:
@ghalliday or @richardkchapman Please merge

PR #16472, opened 2022-08-18
HPCC-28115 Expose k8s connections for predefined k8s services

wangkx:
Expose service type, IPs and ports for a k8s service if the
annotation hpcc.eclwatch.io/enabled for the service is configured
to be true.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/scm/ws_resources.ecm
+++ b/esp/scm/ws_resources.ecm
@@ -25,11 +25,39 @@ ESPStruct HPCCService
     bool TLSSecure;
 };
 
+ESPStruct ServicePorts
+{
+    string Name;
+    string Protocol;
+    unsigned Port;
+};
+
+ESPenum ExternalIPStatus : string
+{
+    Unknown("Unknown"),
+    None("None"),
+    Pending("Pending")
+};
+
+//The ExternalIPsEx or ClusterIPsEx is used when ExternalIPs or ClusterIPs is empty.
+//The value is based on the existing https://github.com/kubernetes/kubernetes.
+ESPStruct ServiceConnection
+{
+    string Type;
+    ESParray<string> ExternalIPs;
+    ESPenum ExternalIPStatus ExternalIPStatus; //Use this field when ExternalIPs is empty. 
+                                               //Set to 'none' for spec type 'ClusterIP' and spec type 'NodePort';
+                                               //set to 'pending' for spec type 'LoadBalancer'; leave it empty for spec type 'ExternalName';
+                                               //set to 'unknown' for unknown spec types.
+    ESParray<ESPstruct ServicePorts> Ports;
+};
+
 ESPStruct DiscoveredWebLink
 {
     string ServiceName;
     string NameSpace;
     ESParray<ESPstruct NamedValue> Annotations;
+    [min_ver("1.02")] ESPstruct ServiceConnection Connection;
 };
 
 ESPStruct ConfiguredWebLink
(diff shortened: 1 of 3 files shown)


wangkx:
@GordonSmith please review, especially on the interface defined in the ecm file.

esp/scm/ws_resources.ecm
```diff
@@ -25,11 +25,35 @@ ESPStruct HPCCService
...
 };
 
+ESPStruct ServicePorts
+{
+    string Name;
+    string Protocol;
+    unsigned Port;
+    unsigned NodePort;
+    unsigned TargetPort;
+};
```
GordonSmith, line 35: What are the 3 ports?  I don't think there is any need to know any of the "internal" ports info?

afishbeck: I think you can remove the internal TargetPort at least until someone asks for it.  Whether anyone will use the NodePort if the type is loadbalancer I'm not sure.

esp/scm/ws_resources.ecm
```diff
@@ -25,11 +25,35 @@ ESPStruct HPCCService
...
+//The value is based on the existing https://github.com/kubernetes/kubernetes.
+ESPStruct ServiceConnection
+{
+    string Type;
+    ESParray<string> ExternalIPs;
+    string ExternalIPsEx; //Use this field when ExternalIPs is empty. Set to '<none>' for spec type 'ClusterIP' and spec type 'NodePort';
+                          //set to '<pending>' for spec type 'LoadBalancer'; leave it empty for spec type 'ExternalName';
+                          //set to '<unknown>' for unknown spec types.
+    ESParray<string> ClusterIPs;
+    string ClusterIPsEx; //when ClusterIPs is empty, set to '<none>'.
```
afishbeck, line 47: The Ex on the name is a little confusing.  I tend to think of Ex as an extended version of something.
If the only purpose is to say "<none>" when the IP list is empty, is it really needed?

esp/scm/ws_resources.ecm
```diff
@@ -25,11 +25,35 @@ ESPStruct HPCCService
...
+    unsigned TargetPort;
+};
+
+//The ExternalIPsEx or ClusterIPsEx is used when ExternalIPs or ClusterIPs is empty.
+//The value is based on the existing https://github.com/kubernetes/kubernetes.
+ESPStruct ServiceConnection
+{
+    string Type;
+    ESParray<string> ExternalIPs;
+    string ExternalIPsEx; //Use this field when ExternalIPs is empty. Set to '<none>' for spec type 'ClusterIP' and spec type 'NodePort';
```
afishbeck, line 43: Also not sure an Ex ending is right here.  Is this field really helpful if it just tells us that the ips are empty?

esp/scm/ws_resources.ecm
wangkx: I may remove it. We may add a field if the value could not be '<none>'.

esp/scm/ws_resources.ecm
wangkx: I could be wrong, but I think the NodePort should be used if the type is "NodePort" (not "LoadBalancer"). See:https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/pkg/apis/core/types.go#L3475

esp/scm/ws_resources.ecm
afishbeck: Maybe call it something like ExternalIPStatus?  You could make it an enum.  We also don't need to follow the kubectl formatting exactly... for example we could remove the '<','>'.

wangkx: Will do

afishbeck (approved):

esp/scm/ws_resources.ecm
GordonSmith: @wangkx from the clients point of view, there isn't any need to report on any IPs or Ports that are "inside the cluster" - however ops may want it in the topology info?

wangkx:
@GordonSmith I just removed the NodePort and ClusterIPs based on our discussion. Please review.

GordonSmith (approved):

PR #16489, opened 2022-08-31
HPCC-28007 Allow dafilesrv limits/throttling to be configured in K8s

jakesmith:
And increase default pool and parallel request limits.

Also:
+ Change spray-service so using cluster visibility (was local).
The service is only needed by the in-cluster dfuserver.
+ Add retry loop (with logging) when spray service is used.
This is to catch the case when the service is overwhelmed dealing
with a lot of concurrent requests already (as configured).
+ Allow spray service to use TLS.
Previously dafilesrv was hardcoded so only the "stream" service
used TLS in containerized mode.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dali/ft/daftformat.cpp
+++ b/dali/ft/daftformat.cpp
@@ -1900,6 +1900,7 @@ void CRemotePartitioner::callRemote()
 
         if (sprayer.useFtSlave)
         {
+            // NB: In containerized mode, spawnRemoteChild will launch ftslave's locally and set connectEP to localhost
             StringBuffer tmp;
             socket.setown(spawnRemoteChild(SPAWNdfu, slave, connectEP, DAFT_VERSION, queryFtSlaveLogDir(), nullptr, wuid));
             if (!socket)
@@ -1911,10 +1912,19 @@ void CRemotePartitioner::callRemote()
         }
         else
         {
-            setDafsEndpointPort(connectEP);
-            if (connectEP.isNull())
-                throwError1(DFTERR_FailedStartSlave, url.str());
-            socket.setown(connectDafs(connectEP, 5000));
+            Owned<IPropertyTree> serviceTree;
+            if (isContainerized())
+            {
+                serviceTree.setown(sprayer.getSprayService());
+                connectEP.set(serviceTree->queryProp("@name"), serviceTree->getPropInt("@port"));
+            }
+            else
+            {
+                setDafsEndpointPort(connectEP);
+                if (connectEP.isNull())
+                    throwError1(DFTERR_FailedStartSlave, url.str());
+            }
+            socket.setown(connectDafs(connectEP, 60000, serviceTree));
             if (!socket)
                 throwError1(DFTERR_FailedStartSlave, url.str());
             prepareCmd(msg);
--- a/dali/ft/filecopy.cpp
+++ b/dali/ft/filecopy.cpp
@@ -347,6 +347,7 @@ bool FileTransferThread::launchFtSlaveCmd(const SocketEndpoint &ep)
     SocketEndpoint connectEP(ep);
 
     // in containerized mode, redirect to dafilesrv service or local (if useFtSlave=true)
+    Owned<IPropertyTree> serviceTree;
     if (isContainerized())
     {
         if (sprayer.useFtSlave)
@@ -356,25 +357,7 @@ bool FileTransferThread::launchFtSlaveCmd(const SocketEndpoint &ep)
         }
         else
         {
-            Owned<IPropertyTree> serviceTree;
-            if (!isEmptyString(sprayer.sprayServiceName))
-            {
-                VStringBuffer serviceQualifier("services[@name='%s']", sprayer.sprayServiceName.get());
-                serviceTree.setown(getGlobalConfigSP()->getPropTree(serviceQualifier));
-                if (!serviceTree)
-                    throw makeStringExceptionV(0, "launchFtSlaveCmd: failed to find dafilesrv service named: '%s'", sprayer.sprayServiceName.get());
-                const char *serviceAppType = serviceTree->queryProp("@application");
-                if (!strsame("spray", serviceAppType))
-                    throw makeStringExceptionV(0, "launchFtSlaveCmd: configured service '%s' is of application type '%s' ('spray' type required)", sprayer.sprayServiceName.get(), nullIfEmptyString(serviceAppType));
-            }
-            else // find 1st of type 'spray'
-            {
-                Owned<IPropertyTreeIterator> directIOServices = getGlobalConfigSP()->getElements("services[@type='spray']");
-                if (directIOServices->first())
-                    serviceTree.set(&directIOServices->query());
-                else
-                    throw makeStringException(0, "launchFtSlaveCmd: no 'spray' dafilesrv services found");
-            }
+            serviceTree.setown(sprayer.getSprayService());
             connectEP.set(serviceTree->queryProp("@name"), serviceTree->getPropInt("@port"));
         }
     }
@@ -397,10 +380,31 @@ bool FileTransferThread::launchFtSlaveCmd(const SocketEndpoint &ep)
     }
     else
     {
-        setDafsEndpointPort(connectEP);
-        if (connectEP.isNull())
-            return false;
-        socket.setown(connectDafs(connectEP, 5000));
+        if (!isContainerized())
+        {
+            setDafsEndpointPort(connectEP);
+            if (connectEP.isNull())
+                return false;
+        }
+        // allow a relatively long time in case service is overloaded (unable to cope with # parallel requests)
+        constexpr unsigned timeoutMs = 1000*60*30; // 30 mins
+        CTimeMon tm(timeoutMs);
+        unsigned remaining;
+        while (true)
+        {
+            try
+            {
+                socket.setown(connectDafs(connectEP, 60000, serviceTree));
+                if (socket)
+                    break;
+            }
+            catch (IException *e)
+            {
+                if (e->errorCode() != JSOCKERR_connection_failed || tm.timedout(&remaining))
+                    throw;
+                PROGLOG("Dafilesrv spray service not responding, retrying..");
+            }
+        }
         if (!socket)
             throwError1(DFTERR_FailedStartSlave, url.str());
         prepareCmd(msg, daFileSrvCommandVersion);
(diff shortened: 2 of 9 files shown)


dali/ft/filecopy.cpp
```diff
@@ -397,10 +397,31 @@ bool FileTransferThread::launchFtSlaveCmd(const SocketEndpoint &ep)
     }
     else
     {
-        setDafsEndpointPort(connectEP);
-        if (connectEP.isNull())
-            return false;
-        socket.setown(connectDafs(connectEP, 5000));
+        if (!isContainerized())
+        {
+            setDafsEndpointPort(connectEP);
```
jakesmith, line 402: NB: this is reall a cosmetic change, previous code wouldn't actually have set EP in containerized mode, because in practice setDafsEndpointPort only set port if not already set.

dali/ft/filecopy.cpp
```diff
@@ -397,10 +397,31 @@ bool FileTransferThread::launchFtSlaveCmd(const SocketEndpoint &ep)
...
+            if (connectEP.isNull())
+                return false;
+        }
+        // allow a relatively long time in case service is overloaded (unable to cope with # parallel requests)
+        constexpr unsigned timeoutMs = 1000*60*30; // 30 mins
+        CTimeMon tm(timeoutMs);
+        unsigned remaining;
+        while (true)
```
jakesmith, line 410: introduces a retry and log loop. The previous short 5 second connect attempt could timeout if K8s service was throttling connections.

fs/dafilesrv/dafilesrv.cpp
```diff
@@ -580,23 +583,6 @@ int main(int argc, const char* argv[])
...
-                throttleQueueLimit = dafileSrvInstance->getPropInt("@throttleQueueLimit", throttleQueueLimit);
-
-                parallelSlowRequestLimit = dafileSrvInstance->getPropInt("@parallelSlowRequestLimit", parallelSlowRequestLimit);
-                throttleSlowDelayMs = dafileSrvInstance->getPropInt("@throttleSlowDelayMs", throttleSlowDelayMs);
-                throttleSlowCPULimit = dafileSrvInstance->getPropInt("@throttleSlowCPULimit", throttleSlowCPULimit);
-                throttleSlowQueueLimit = dafileSrvInstance->getPropInt("@throttleSlowQueueLimit", throttleSlowQueueLimit);
-
-                dedicatedRowServicePort = dafileSrvInstance->getPropInt("@rowServicePort", dedicatedRowServicePort);
-                dedicatedRowServiceSSL = dafileSrvInstance->getPropBool("@rowServiceSSL", dedicatedRowServiceSSL);
-                rowServiceOnStdPort = dafileSrvInstance->getPropBool("@rowServiceOnStdPort", rowServiceOnStdPort);
```
jakesmith, line 599: NB: code moved, so that configuration option code is common to bare-metal/containerized

fs/dafsclient/rmtclient.cpp
```diff
@@ -1085,81 +1097,122 @@ ISocket *checkSocketSecure(ISocket *socket)
     return LINK(socket);
 }
 
-ISocket *connectDafs(SocketEndpoint &ep, unsigned timeoutms)
+ISocket *connectDafs(SocketEndpoint &ep, unsigned timeoutms, const IPropertyTree *service)
 {
     Owned<ISocket> socket;
 
-    if ( (securitySettings.queryConnectMethod() == SSLNone) || (securitySettings.queryConnectMethod() == SSLOnly) )
+    if (isContainerized())
```
jakesmith, line 1104: refactored, because containerized route is significantly simpler than convuluted bare-metal strategy of trying to connect to secure, then insecure etc.

helm/hpcc/values.yaml
```diff
@@ -385,7 +385,7 @@ dafilesrv:
   application: spray
   service:
     servicePort: 7300
-    visibility: local
+    visibility: cluster
```
jakesmith, line 388: NB: this should always have been 'cluster'. We do not want the spray service (which serves the local dfuserver) to have any external access.

jakesmith:
@afishbeck - please review. I've added a few small in-line comments to assist the review

fs/dafsclient/rmtclient.cpp
```diff
@@ -1085,81 +1097,122 @@ ISocket *checkSocketSecure(ISocket *socket)
...
-    if (newtimeout > 5000)
-        newtimeout = 5000;
 
-    int conAttempts = 2;
-    while (conAttempts > 0)
-    {
-        conAttempts--;
-        bool connected = false;
-        try
+        if (service->getPropBool("@tls"))
```
afishbeck, line 1108: Can service be null when containerized?  I see several places passing nullptr.. perhaps should throw an exception if null?

fs/dafilesrv/dafilesrv.cpp
```diff
@@ -642,6 +628,51 @@ int main(int argc, const char* argv[])
...
+    throttleDelayMs = 1000;
+    throttleCPULimit = 85;
+    throttleQueueLimit = 1000;
+    parallelSlowRequestLimit = parallelRequestLimit;
+    throttleSlowDelayMs = throttleDelayMs;
+    throttleSlowCPULimit = throttleCPULimit;
+    throttleSlowQueueLimit = throttleQueueLimit;
+#endif
+
+    maxThreads = dafileSrvInstance->getPropInt("@maxThreads", maxThreads);
```
afishbeck, line 657: Is there a chance in baremetal where we get here with dafileSrvInstance==nullptr?  Perhaps a misconfiguration with no Instance?

fs/dafsclient/rmtclient.cpp
```diff
@@ -126,32 +126,44 @@ static class _securitySettings
 
 static CriticalSection              secureContextCrit;
 static Owned<ISecureSocketContext>  secureContextClient;
+static std::unordered_map<std::string, Owned<ISecureSocketContext>> secureCtxClientIssuerMap;
```
afishbeck, line 129: This caching of issuer based SecureSocketContexts is something to consider commoning up and using elsewhere.  But if we did that, we'd have to insure consistency.  For example don't want to mingle disableClientCertVerification contexts with non disableClientCertVerification contexts.

afishbeck (commented):
A few minor comments, otherwise looks good.

fs/dafsclient/rmtclient.cpp
jakesmith: This is only on the client side at the moment, whereas the disableClientCertVerification option is on the server side.
Not sure there's much need to cache on the server side typically, as it will have 1 context, whereas clients are potentially connecting to disparate servers with different certificates.

But as you say, there are prob. other contexts where it would be good to cache.
If more generalized, it should also cope with configuration changes, e.g. on a helm update, which would potentially change tls info etc., and therefore should invalidate the cache.

In fact, I'll look at adding that change, to clear the map if a configuration chagnes, as it may be a gotcha.

fs/dafilesrv/dafilesrv.cpp
jakesmith: as you perhaps if there's a slipped through misconfiguration.
I'll add a guard.

fs/dafsclient/rmtclient.cpp
jakesmith: That question highlighted one place where it should potentially also have passed the service (in RemotePartitioner).
However, I'm not entirely sure how or even if, that route can be taken in containerized mode, and it would likely not have worked before anyway, because the legacy ftslave route was also incomplete.

Nevertheless for completeness, I'll modify that code, which will mean commoning up the code that gets the service.

The other places that call connectDaFs that pass null, are bare-metal only and cases that I don't think will ever be used in directly containerized mode (mostly dafscontrol). Rather than throw an exception.
I will change the if to be conditional on 'service' and let the code continue to try without certificates.

jakesmith (commented):
@afishbeck - please see replies and new commit

afishbeck (approved):

jakesmith:
Rebased to master to target 8.10

dali/ft/daftformat.cpp
```diff
@@ -1900,6 +1900,11 @@ void CRemotePartitioner::callRemote()
 
         if (sprayer.useFtSlave)
         {
+            if (isContainerized())
```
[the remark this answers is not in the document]
ghalliday: It is called when you need to read the entire file to find the partition points (e.g. csv with quote defined)

spawnRemoteChild() sets the connect ip to "localhost" (rmtspawn.cpp(152)) for containerized - so this shouldn't be needed.

fs/dafilesrv/dafilesrv.cpp
```diff
@@ -391,8 +391,12 @@ int main(int argc, const char* argv[])
     unsigned short  sslport;
     unsigned dedicatedRowServicePort = DEFAULT_ROWSERVICE_PORT;
 #ifdef _CONTAINERIZED
-    bool stream = strsame(config->queryProp("@application"), "stream");
-    connectMethod = stream ? SSLOnly : SSLNone;
+    // Use the "public" certificate issuer, unless it's visibility is "cluster" (meaning internal only)
+    const char *certScope = strsame("cluster", getComponentConfigSP()->queryProp("service/@visibility")) ? "local" : "public";
```
ghalliday, line 395: very minor:  The following would be easier to read:

```
const char * visibility = getComponentConfigSP()->queryProp("service/@visibility");
const char *certScope = strsame("cluster", visibility) ? "local" : "public";
```

ghalliday (approved):
@jakesmith one comment, but can disregard.  Please squash.
Also, should this target 8.8 rather than master?

dali/ft/daftformat.cpp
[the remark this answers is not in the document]
jakesmith: ok, I will remove the code and leave a comment instead

fs/dafilesrv/dafilesrv.cpp
jakesmith: agree, will change.

jakesmith:
> @jakesmith one comment, but can disregard. Please squash. Also, should this target 8.8 rather than master?

it was retargeted to master after a discussion last week, we decided to let this go into 8.10.

ghalliday (approved):

PR #16501, opened 2022-09-06
HPCC-28109 If IP only supplied, use to lookup despray plane

jakesmith:
The fileservice despray command should be supplied with
a target plane (destPlane), but for greater portability with
bare metal, continue to allow a destIP in containerized mode,
to check to see if a plane exists that maps to storage on
'hosts' with that IP.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/services/ws_fs/ws_fsService.cpp
+++ b/esp/services/ws_fs/ws_fsService.cpp
@@ -2491,7 +2491,28 @@ bool CFileSprayEx::onDespray(IEspContext &context, IEspDespray &req, IEspDespray
             if (isEmptyString(destPlane))
                 destPlane = req.getDestGroup();  // allow eclwatch to continue providing storage plane as 'destgroup' field
             if (isEmptyString(destPlane))
-                throw makeStringException(ECLWATCH_INVALID_INPUT, "Destination storage plane not specified.");
+            {
+                if (destip.isEmpty())
+                    throw MakeStringException(ECLWATCH_INVALID_INPUT, "Neither destination storage plane or destination IP specified.");
+                SocketEndpoint destEp(destip);
+                Owned<IPropertyTreeIterator> planesIter = getDropZonePlanesIterator();
+                ForEach(*planesIter)
+                {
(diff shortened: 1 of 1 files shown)


jakesmith:
NB: spray (vs despray) already allows srcIP in fileservice commands.

@ghalliday - please review.

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -2491,7 +2491,28 @@ bool CFileSprayEx::onDespray(IEspContext &context, IEspDespray &req, IEspDespray
...
+                SocketEndpoint destEp(destip);
+                Owned<IPropertyTreeIterator> planesIter = getDropZonePlanesIterator();
+                ForEach(*planesIter)
+                {
+                    IPropertyTree &lzPlane = planesIter->query();
+                    Owned<IPropertyTreeIterator> hostsIter = lzPlane.getElements("hosts");
+                    ForEach(*hostsIter)
+                    {
+                        SocketEndpoint planeEp(hostsIter->query().queryProp(nullptr));
+                        if (planeEp.ipequals(destEp))
```
ghalliday, line 2506: Questions.  Should this be using a string compare (not sure)  and I think it should complain if it is not unique

jakesmith: Hestitated when resolving 1st to IP before comparing, but figured it may be possible for there to be multiple DNS names (e.g. with and without domain) that would resolve to the same physical IP.

> I think it should complain if it is not unique

It is assumed by this code that is true. At least I am assuming it is invalid to define multiple 'lz' planes with the same or overlapping hosts (?).
Feels like that checking should live elsewhere though, not here?

ghalliday (approved):

ghalliday:
@wangkx please can you sanity check since this is code that you normally maintain.

PR #16509, opened 2022-09-07
HPCC-28208 Standardize units and bucket labels for scaled histogram m…

kenrowland:
…etrics

Standardized scaled histogram metrics to NS for time based measurements Removed label strings from bucket definitions as the sink will need to generate these anyway.
Added convenience function to create scaled cycles to ns scaled histogram

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Cloud-compatibility
  - [x] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jmetrics.hpp
+++ b/system/jlib/jmetrics.hpp
@@ -103,9 +103,9 @@ interface IMetric
     virtual std::vector<__uint64> queryHistogramValues() const = 0;
 
     /*
-     * Query histogram labels
+     * Query histogram bucket limits
      */
-    virtual std::vector<std::string> queryHistogramLabels() const = 0;
+    virtual std::vector<__uint64> queryHistogramBucketLimits() const = 0;
 };
 
 
@@ -123,7 +123,7 @@ public:
     const MetricMetaData &queryMetaData() const override { return metaData; }
     StatisticMeasure queryUnits() const override { return units; }
     virtual std::vector<__uint64> queryHistogramValues() const override { return {}; }
-    virtual std::vector<std::string> queryHistogramLabels() const override { return {}; }
+    virtual std::vector<__uint64> queryHistogramBucketLimits() const override { return {}; }
 
 
 protected:
@@ -249,17 +249,11 @@ protected:
 };
 
 
-struct BucketDef
-{
-    std::string label;
-    __uint64 limit;
-};
-
 
 class jlib_decl HistogramMetric : public MetricBase
 {
 public:
-    HistogramMetric(const char *name, const char *desc, StatisticMeasure _units, const std::vector<BucketDef> &bucketDefs, const MetricMetaData &_metaData = MetricMetaData());
+    HistogramMetric(const char *name, const char *desc, StatisticMeasure _units, const std::vector<__uint64> &bucketLimits, const MetricMetaData &_metaData = MetricMetaData());
 
     virtual __uint64 queryValue() const override
     {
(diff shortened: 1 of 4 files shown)


testing/unittests/metrics/MetricFrameworkTests.cpp
```diff
@@ -418,6 +429,15 @@ class MetricFrameworkTests : public CppUnit::TestFixture
             CPPUNIT_ASSERT_EQUAL(static_cast<int>(pHistogram->queryValue()), expectedSum);
     }
 
+
+    void checkHistogramValue(std::shared_ptr<IMetric> pHistogram, __uint64 expectedValue, __uint64 error)
```
ghalliday, line 433: trivial: Slightly better if 1st parameter is a reference (but doesn't really matter since test code).

testing/unittests/metrics/MetricFrameworkTests.cpp
```diff
@@ -312,16 +312,22 @@ class MetricFrameworkTests : public CppUnit::TestFixture
...
     void Test_histogram_metric()
     {
-        std::vector<BucketDef> bucketDefs = {{"le 2", 2}, {"le 4", 4}, {"le 8", 8} };
+        //std::vector<BucketDef> bucketDefs = {{"le 2", 2}, {"le 4", 4}, {"le 8", 8} };
+        std::vector<__uint64> bucketDefs = {2, 4, 8};
         std::shared_ptr<HistogramMetric> pHistogram = std::make_shared<HistogramMetric>("requests.dist", "description", SMeasureCount, bucketDefs);
 
         std::vector<__uint64> values;
 
+        printf("Validating number of returned buckets\n");
```
ghalliday, line 321: Possibly better is to use DBGLOG.  If you run with the -v option the logging is traced to stdout

system/jlib/jmetrics.hpp
```diff
@@ -249,17 +249,11 @@ class CustomMetric : public MetricBase
...
-    std::string label;
-    __uint64 limit;
-};
-
 
 class jlib_decl HistogramMetric : public MetricBase
 {
 public:
-    HistogramMetric(const char *name, const char *desc, StatisticMeasure _units, const std::vector<BucketDef> &bucketDefs, const MetricMetaData &_metaData = MetricMetaData());
+    HistogramMetric(const char *name, const char *desc, StatisticMeasure _units, const std::vector<__uint64> &bucketLimitss, const MetricMetaData &_metaData = MetricMetaData());
```
ghalliday, line 256: typo: bucketLimitss (I had to find something....).  Second occurrence below

ghalliday (commented):
Looks good, always nice when the code gets simpler.
A couple of minor comments.

jakesmith:
> 1) test: MetricFrameworkTests::Test_histogram_metric (E) 
> uncaught exception of type std::exception (or derived).
> - std::bad_alloc

Looks like this is not as a result of this PR, because other PR's are also failing with same, e.g.:
https://github.com/hpcc-systems/HPCC-Platform/pull/16512

@kenrowland - was there a recent related PR that was merged that is causing this?

kenrowland:
@jakesmith @richardkchapman Looking into where it is failing, but cannot see what the new histogram tests could be doing to cause this. They don't allocate any memory other than some small vectors. 

PR https://github.com/hpcc-systems/HPCC-Platform/pull/16463 did not cause any unittest failures

kenrowland:
@ghalliday I removed all of the tracing code I added to the unit tests to see if something was going wrong and causing the bad alloc error being seen. Nothing was found to be incorrect. 

I did get one stack dump from Atilla and it seemed it was failing in a base metric class constructor when it was copying the default meta data object. That object defaults to an empty vector and is passed as a const & to parent class constructors. I don't see how it could cause any problems (and I added some output statements to confirm it was a 0 length vector).

I am as confident as I can be that the bad alloc error happening during smoketest/unit tests is not being caused by any errors in the metric framework tests.

ghalliday (approved):
Looks good - please squash.

kenrowland:

PR #16525, opened 2022-09-13
HPCC-28175 Vault transformations for envmod

Michael-Gardner:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/initfiles/componentfiles/configschema/templates/AddVaults.json
+++ b/initfiles/componentfiles/configschema/templates/AddVaults.json
@@ -1,0 +1,102 @@
+{
+    "name": "Add Vault Feature",
+    "description": "Add necessary keys with empty values for vault feature",
+    "type": "modification",
+    "operations": [
+        {
+            "action": "delete",
+            "target_path": "/Environment/Software/vaults",
+            "data": {
+                "error_if_not_found": false
+            }
+        },
+        {
+            "action": "create",
+            "target_path": "/Environment/Software",
+            "data": {
+                "node_type": "vaults",
+                "save": {
+                    "name" : "vaults_node"
+                }
+            }
+        },
+        {
+            "action": "create",
+            "description": "Create the git element in Environment/Software/vaults",
+            "target_nodeid": "{{vaults_node}}",
+            "data": {
+                "node_type": "git",
+                "attributes": [
+                    {
+                        "name": "name",
+                        "value": "git_vault"
+                    },
+                    {
+                        "name": "url",
+                        "value": "insert_your_url_here"
+                    },
+                    {
+                        "name": "kind",
+                        "value": "kv_v2"
+                    },
+                    {
+                        "name": "client-secret",
+                        "value": "remove_this_if_unused"
+                    }
+                ]
+            }
+        },
+        {
+            "action": "create",
+            "description": "Create the ecl element in Environment/Software/vaults",
+            "target_nodeid": "{{vaults_node}}",
+            "data": {
+                "node_type": "ecl",
+                "attributes": [
+                    {
+                        "name": "name",
+                        "value": "ecl_vault"
+                    },
+                    {
+                        "name": "url",
+                        "value": "insert_your_url_here"
+                    },
+                    {
+                        "name": "kind",
+                        "value": "kv_v2"
+                    },
+                    {
+                        "name": "client-secret",
+                        "value": "remove_this_if_unused"
+                    }
+                ]
+            }
+        },
+        {
+            "action": "create",
+            "description": "Create the eclUser element in Environment/Software/vaults",
+            "target_nodeid": "{{vaults_node}}",
+            "data": {
+                "node_type": "eclUser",
+                "attributes": [
+                    {
+                        "name": "name",
+                        "value": "eclUser_vault"
+                    },
+                    {
+                        "name": "url",
+                        "value": "insert_your_url_here"
+                    },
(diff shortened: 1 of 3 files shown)


Michael-Gardner:
Adding @afishbeck for viewing

@kenrowland I had to populate the attributes with a non zero string (chose "default") or else I was hitting segmentation faults from envmod.  

Transformation adds a new node <vaults/> to Environment/Software and then populates the node with 3 elements and containing a few attributes.

kenrowland:
@Michael-Gardner I am guessing that no XSD was created for these new keys. If so, the XSD could have specified a default value for the attributes where you need to set a non-empty string. The default value can be the empty string. This intended behavior when the tool and supporting libs were created was that XSDs would serve as validation templates for any added keys to the environment.xml. Since we are moving away from that, it does not seep useful to add XSDs for these keys, so adding the values with a non-empty string is the workaround.

initfiles/componentfiles/configschema/templates/AddVaults.json
```diff
@@ -0,0 +1,102 @@
...
+            "data": {
+                "node_type": "git",
+                "attributes": [
+                    {
+                        "name": "name",
+                        "value": "git_vault"
+                    },
+                    {
+                        "name": "url",
+                        "value": "default"
```
afishbeck, line 36: Having it not empty as a workaround is fine, but I wouldn't use the term default because to me that makes it sound like the back end will just use some default value.  Something saying they need to fill in the value would be better.

initfiles/componentfiles/configschema/templates/AddVaults.json
```diff
@@ -0,0 +1,102 @@
...
+            "target_nodeid": "{{vaults_node}}",
+            "data": {
+                "node_type": "git",
+                "attributes": [
+                    {
+                        "name": "name",
+                        "value": "git_vault"
+                    },
+                    {
+                        "name": "url",
```
kenrowland, line 35: Is it necessary that the url, and other keys that follow, have an entry? What would happen if the attributes were left out of the template since you are only using a (workaround) default value.

Or, based on a comment Tony just made, if the intent is for the user to fill it in, use a string indicating such.

If the attribute is not needed, maybe use "fill me in or remove if not needed"

initfiles/componentfiles/configschema/templates/AddVaults.json
```diff
@@ -0,0 +1,102 @@
...
+                        "value": "default"
+                    },
+                    {
+                        "name": "kind",
+                        "value": "kv_v2"
+                    },
+                    {
+                        "name": "client-secret",
+                        "value": "default"
```
afishbeck, line 44: client-secret is actually optional.  I guess it needs to be filled in here so it shows up... but for this one it should tell them to either fill in a value or delete the tag.

Michael-Gardner: This has been modified to 

client-secret : optional_delete_if_unused

initfiles/componentfiles/configschema/templates/AddVaults.json
Michael-Gardner: This has been modified to 

url : required_insert_url

afishbeck (approved):
@Michael-Gardner @kenrowland I'm not an expert on envmod but looks good AFAICT

ghalliday:
@kenrowland please check you are happy with the second commit

initfiles/componentfiles/configschema/xsd/vaults.xsd
```diff
@@ -0,0 +1,56 @@
...
+
+<xs:schema
+        xmlns:xs="http://www.w3.org/2001/XMLSchema" elementFormDefault="qualified" attributeFormDefault="unqualified"
+        xmlns:hpcc="someuri">
+    <xs:include schemaLocation="types.xsd"/>
+    <hpcc:insert hpcc:schemaPath="/Environment/Software">
+        <xs:element name="vaults" hpcc:displayName="vaults section" hpcc:itemType="vaults" maxOccurs="1" use="required" hpcc:class="component">
+            <xs:complexType>
+                <xs:sequence>
+                    <xs:element name="git" hpcc:displayName="Vault Git Group" hpcc:class="elementSet" maxOccurs="1" use="required">
```
kenrowland, line 28: Were you able to test w/o the "required" value to make sure that the empty elements weren't added?

Michael-Gardner: I attempted to remove the use=required and there was no change. Then I attempted to remove the maxOccurs=1. Still no change. Only when I completely removed the xsd did I no longer get the duplicate <ecl \> entries.  Should I just remove the xsd and keep it as the simple template only in the PR?

Michael-Gardner: So for the hpcc:class elementSet, I should have been using minOccurs=0 so it wouldn't force an empty populated element on envmod's first pass through (before it processes the template).  Also removing the use=required from the individual attributes, as that was also causing issues with populating empty elements. With both of those changes its now working how we expect.

kenrowland (approved):
Looks good

PR #16562, opened 2022-09-23
HPCC-28302 Avoid MP clashing by using peer endpoints

jakesmith:
MP clients were self-identifying their endpoints, by serializing them to the MP server. This works fine if both client and server are on the same network, but if clients or servers are on their own private network (clients connecting to a k8s system is a good example), then those private client endpoints are not unique. i.e. 2 clients from different private networks may have identical endpoints and will cause MP clashing and random disconnections.

To avoid this, use the peer endpoint of the connected socket. This means that the MP client connections are now identified and mapped by the peer endpoint of the connected socket.

This also necessitates a few other required changes:

1) Each MP packet also has an endpoint in it, which are encoded in their header. On receipt (via socket notification), change their header to match the peer endpoint.
2) Subscriptions also self-identified, and relied on matching the incoming packets for notification.
Change to ignore the self-identification endpoint, and use the peer endpoint encoded in the message header (sender). 3) The Thor worker nodes rely on being able to identify themselves in the thor process group (formed at registration time), they were relying on the peer sender, which due to the above changes could now be different.
Change to serialize the slaves endpoint to master at registration time and use to build process group.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jsocket.cpp
+++ b/system/jlib/jsocket.cpp
@@ -425,7 +425,7 @@ public:
     void        shutdown(unsigned mode=SHUTDOWN_READWRITE);
     void        shutdownNoThrow(unsigned mode);
 
-    ISocket*    accept(bool allowcancel, SocketEndpoint *peerEp=nullptr);
+    ISocket*    accept(bool allowcancel, SocketEndpoint *retPeerEp=nullptr);
     int         wait_read(unsigned timeout);
     int         logPollError(unsigned revents, const char *rwstr);
     int         wait_write(unsigned timeout);
@@ -1565,10 +1565,9 @@ void CSocket::setTraceName()
 void CSocket::setPeerEndpoint(const SocketEndpoint &ep)
 {
     assertex(!peerEpSet);
-    peerEpCrit.enter();
-    peerEpSet = true;
+    CriticalBlock b(peerEpCrit);
     peerEp = ep;
-    peerEpCrit.leave();
+    peerEpSet = true;
 }
 
 ISocket*  ISocket::connect_wait( const SocketEndpoint &ep, unsigned timems)
--- a/system/jlib/jsocket.cpp
+++ b/system/jlib/jsocket.cpp
@@ -384,6 +384,9 @@ protected:
     SOCKETMODE      sockmode;
     IpAddress       targetip;
     SocketEndpoint  returnep;   // set by set_return_addr
+    SocketEndpoint  peerEp;
+    std::atomic_bool peerEpSet{false};
+    CriticalSection peerEpCrit;
 
     MCASTREQ    *   mcastreq;
     size32_t        nextblocksize;
@@ -468,6 +471,7 @@ public:
 
     void        setTraceName(const char * prefix, const char * name);
     void        setTraceName();
+    void        setPeerEndpoint(const SocketEndpoint &ep);
 
     CSocket(const SocketEndpoint &_ep,SOCKETMODE smode,const char *name);
     CSocket(T_SOCKET new_sock,SOCKETMODE smode,bool _owned);
@@ -483,7 +487,6 @@ public:
     {
         return sock != INVALID_SOCKET;
     }
-
 private:
 
     int closesock()
@@ -1025,7 +1028,7 @@ ErrPortInUse:
 
 
 
-ISocket* CSocket::accept(bool allowcancel, SocketEndpoint *peerEp)
+ISocket* CSocket::accept(bool allowcancel, SocketEndpoint *retPeerEp)
 {
     if ((accept_cancel_state!=accept_not_cancelled) && allowcancel) {
         accept_cancel_state=accept_cancelled;
@@ -1081,10 +1084,13 @@ ISocket* CSocket::accept(bool allowcancel, SocketEndpoint *peerEp)
         THROWJSOCKEXCEPTION(JSOCKERR_cancel_accept);
     }
 
-    if (peerEp)
-        getSockAddrEndpoint(peerSockAddr, peerSockAddrLen, *peerEp);
+    SocketEndpoint _peerEp;
+    getSockAddrEndpoint(peerSockAddr, peerSockAddrLen, _peerEp);
+    if (retPeerEp)
+        *retPeerEp = _peerEp;
 
     CSocket *ret = new CSocket(newsock,sm_tcp,true);
+    ret->setPeerEndpoint(_peerEp);
     ret->set_inherit(false);
     return ret;
 
@@ -1161,6 +1167,11 @@ int CSocket::peer_name(char *retname,size32_t namemax)
 
 SocketEndpoint &CSocket::getPeerEndpoint(SocketEndpoint &ep)
 {
+    if (peerEpSet)
+    {
+        ep = peerEp;
+        return ep;
+    }
     if (state != ss_open) {
         THROWJSOCKEXCEPTION(JSOCKERR_not_opened);
     }
@@ -1175,8 +1186,12 @@ SocketEndpoint &CSocket::getPeerEndpoint(SocketEndpoint &ep)
             DBGLOG("getpeername failed %d",ERRNO());
             ep.set(NULL, 0);
         }
-        else
+        else {
+            CriticalBlock b(peerEpCrit);
             getSockAddrEndpoint(u,ul,ep);
+            peerEp = ep;
+            peerEpSet = true;
+        }
     }
     return ep;
 }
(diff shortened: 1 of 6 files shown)


jakesmith:
@mckellyln - for discussion

system/jlib/jsocket.cpp
```diff
@@ -1547,6 +1562,14 @@ void CSocket::setTraceName()
 #endif
 }
 
+void CSocket::setPeerEndpoint(const SocketEndpoint &ep)
+{
+    assertex(!peerEpSet);
+    peerEpCrit.enter();
+    peerEpSet = true;
```
ghalliday, line 1569: This should be set after peerEP, otherwise theoretically peerEp can be read from getPeerEndpoint() before it is set.

system/jlib/jsocket.cpp
```diff
@@ -1547,6 +1562,14 @@ void CSocket::setTraceName()
 #endif
 }
 
+void CSocket::setPeerEndpoint(const SocketEndpoint &ep)
+{
+    assertex(!peerEpSet);
+    peerEpCrit.enter();
```
ghalliday, line 1568: question: curious why you didn't use criticalBlock

system/mp/mpcomm.cpp
```diff
@@ -2271,27 +2278,24 @@ int CMPConnectThread::run()
...
+                // See HPCC-28125/HPCC-28302
+                // A new client connecting here, creates a new MP channel, which is later looked up based on an
+                // endpoint. The client itself transmits its endpoint in the ConnectHdr and also in the PacketHeader's
+                // sent with each MP packet (received by CMPPacketReader::notifySelected).
+                // The latter is used to enqueue and trigger mp tag matching that delivers the packets to the correct
+                // MP tag listeners.
+                // However, the client-side endpoint cannot be relied upon to be unique, since it may be running on another
+                // private network and therefore multiple clients on different networks connecting to Dali could have the same
+                // client side endpoints. 
+                // To avoid this, always use the server side peer IP and oprt as the mapping from client to MP channel.
```
ghalliday, line 2290: typo: oprt->port

system/mp/mpcomm.cpp
```diff
@@ -2271,27 +2278,24 @@ int CMPConnectThread::run()
...
+                // The peer port must be used as well, since the client side port from multiple private networks can be the same.
+                // See corresponding change in CMPPacketReader::notifySelected
+                // NB: if all clients are part of the same network, then this step is strictly unecessary.
```
ghalliday, line 2293: typo: unnecessary

ghalliday (commented):
one minor issue, a couple of typos.  I'll defer to Mark to check the logic of which ips/ports are used.

system/jlib/jsocket.cpp
jakesmith: yes, changed.

system/mp/mpcomm.cpp
jakesmith: fixed

system/mp/mpcomm.cpp
jakesmith: fixed

jakesmith:
@ghalliday - addressed comments.

jakesmith:
> https://github.com/hpcc-systems/HPCC-Platform/actions/runs/3127212160

@GordonSmith - semi-interesting error:
```
Run mono `./vcpkg fetch nuget | tail -n 1` \
Cannot open assembly 'No': No such file or directory.
Error: Process completed with exit code 2.
```

I suspect it is related to the fact that I changed the target branch changed, and perhaps it was mid-flight at the time, not sure.
might be worth catching whatever underlying error this hit?

jakesmith:
> I suspect it is related to the fact that I changed the target branch changed, and perhaps it was mid-flight at the time

or perhaps not. I've rerun and closed/reopened the PR, but fails with same issue: https://github.com/hpcc-systems/HPCC-Platform/actions/runs/3127494455

@GordonSmith - any ideas?

GordonSmith:
@jakesmith - I have a fix for this, but didn't release to 8.6.x, I will now...

system/mp/mpcomm.cpp
```diff
@@ -1639,6 +1639,13 @@ class CMPPacketReader: public ISocketSelectNotify, public CInterface
                     remaining = hdr.size-sizeof(hdr);
                     activemsg = new CMessageBuffer(remaining); // will get from low level IO at some stage
                     activeptr = (byte *)activemsg->reserveTruncate(remaining);
+
+                    // Ensure MP packet use peer endpoint
+                    SocketEndpoint peerEp;
+                    sock->getPeerEndpoint(peerEp);
+                    peerEp.port = hdr.sender.port;
```
jakesmith, line 1646: this and the similar line in the corresponding line in CMPConnectThread::run is preserving the original port,
but it shouldn't be here, I may have inadvertently reintroduced it.
At one point, subscriptions were failing because they relied on port matching the original (Client) endpoint.

I'll need to remove, retest

mckellyln (commented):
It looks good - after the two fixes for resetting the port are made and tested.

system/jlib/jsocket.cpp
```diff
@@ -1175,8 +1186,12 @@ SocketEndpoint &CSocket::getPeerEndpoint(SocketEndpoint &ep)
             DBGLOG("getpeername failed %d",ERRNO());
             ep.set(NULL, 0);
         }
-        else
+        else {
+            CriticalBlock b(peerEpCrit);
             getSockAddrEndpoint(u,ul,ep);
```
ghalliday, line 1191: tirivial:  Unlikely to have any effect, but this could be before the critical section to reduce the window it is held for.

system/jlib/jsocket.cpp
```diff
@@ -1161,6 +1167,11 @@ int CSocket::peer_name(char *retname,size32_t namemax)
 
 SocketEndpoint &CSocket::getPeerEndpoint(SocketEndpoint &ep)
 {
+    if (peerEpSet)
+    {
+        ep = peerEp;
```
ghalliday, line 1172: technically I suspect this isn't well defined because peerEP isn't atomic - so the compiler is theoretically allowed to read peerEp before testing peerEpSet.  It would also be legal for the compiler to reorder the store to peerEpSet before setting peerEp (and some processors may reorder them).  In practice I suspect it is highly unlikely to cause any problems.
I think the cleanest "correct" solution is to read the value of peerEp within a critical section, but the test for peerEpSet probably needs to be inside the crit sec as well.

ghalliday (commented):
@mckellyln do you think the theoretical race conditions can be ignored?

system/jlib/jsocket.cpp
mckellyln: I agree, move getSocketAddrEndpoint() above CB

system/jlib/jsocket.cpp
mckellyln: I agree, good find.  Put this inside a CS

system/mp/mpcomm.cpp
```diff
@@ -2271,27 +2278,24 @@ int CMPConnectThread::run()
...
+                // private network and therefore multiple clients on different networks connecting to Dali could have the same
+                // client side endpoints. 
+                // To avoid this, always use the server side peer IP and port as the mapping from client to MP channel.
+                // The peer port must be used as well, since the client side port from multiple private networks can be the same.
+                // See corresponding change in CMPPacketReader::notifySelected
+                // NB: if all clients are part of the same network, then this step is strictly unnecessary.
+
+                // Ensure MP connection uses peer endpoint
+                _remoteep.set(peerEp);
+                _remoteep.port = connectHdr.id[0].port;
```
mckellyln, line 2297: And no port reset here and above.

jakesmith:
@mckellyln @ghalliday - I have been working on solving an issue once the override port lines were removed (see https://github.com/hpcc-systems/HPCC-Platform/pull/16562#discussion_r980149235).

I had thought I had already got rid of those lines and everything was working, alas, without those port override lines, there are some problems with groups (IGroup) / ranks (rank(&sender)) etc. - which break Thor - and I think will potentially cause problems elsewhere.

There is no easy fix I can see, there really wants to be a differentiation between a ICommunicator and a IInterCommunicator, where the former is iternal and shouldn't crossing NAT. (and can involve a group of >1 node) and the latter which is point-to-point.
Unfortunately, it shares common global obects server, MyNode, etc. and common list of channels
There might be a way, but not trivially.

For now I'm going to close this PR, and I will issue a new PR to rollback the change introduced by HPCC-28125.
[ which is causes inter cluster communication failure where peer is different from source, e.g. from AKS to external ]

jakesmith:
Revert PR: https://github.com/hpcc-systems/HPCC-Platform/pull/16574

PR #16587, opened 2022-09-29
HPCC-28331 ES Translations for 8.10.X

kunalaswani:
Translations added for Spanish (ES).

Signed-off-by: Kunal Aswani <kunal.aswani@lexisnexis.com>

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:


esp/src/src/nls/es/hpcc.ts
```diff
@@ -338,8 +343,15 @@ export = {
...
     Generate: "Generar",
+    GetDFSCSV: "CSV de DFS",
+    GetDFSMap: "Mapa DFS",
+    GetDFSParents: "Padres de DFS",
+    GetLogicalFile: "Archivo lógico",
+    GetLogicalFilePart: "Parte del archivo lógico",
     GetPart: "Obtener parte",
+    GetProtectedList: "Lista Protegida",
     GetSoftwareInformation: "Obtener informacion del software",
+    GetValue: "Valor",
```
rpastrana, line 354: As is, it translates to Value, if English version is Get Value, this should be Obtener Valor

esp/src/src/nls/es/hpcc.ts
```diff
@@ -338,8 +343,15 @@ export = {
     FromTime: "Tiempo de Comenzar",
     FullName: "Nombre Entero",
     Generate: "Generar",
+    GetDFSCSV: "CSV de DFS",
+    GetDFSMap: "Mapa DFS",
+    GetDFSParents: "Padres de DFS",
+    GetLogicalFile: "Archivo lógico",
+    GetLogicalFilePart: "Parte del archivo lógico",
     GetPart: "Obtener parte",
+    GetProtectedList: "Lista Protegida",
```
rpastrana, line 352: As is, it translates to Protected List, if English version is Get Protected List, this should be Obtener Lista Protegida

esp/src/src/nls/es/hpcc.ts
```diff
@@ -338,8 +343,15 @@ export = {
     FromTime: "Tiempo de Comenzar",
     FullName: "Nombre Entero",
     Generate: "Generar",
+    GetDFSCSV: "CSV de DFS",
```
rpastrana, line 346: L346-350 seen to be missing "Obtener"

esp/src/src/nls/es/hpcc.ts
```diff
@@ -621,6 +639,7 @@ export = {
     PleaseSelectAUserOrGroup: "Por favor escoja un usario o grupo y nombre de archivo",
     PleaseSelectAUserToAdd: "Por favor escoja el usario para agregar",
     Plugins: "Complementos",
+    Pods: "Vainas",
```
rpastrana, line 642: I think we're better off leaving this as pods, vaina is a negative word we should avoid. Also, docker seems to keep the term pod in their ES doc
https://kubernetes.io/es/docs/concepts/workloads/pods/

rpastrana (changes requested):
@kunalaswani a few comments

esp/src/src/nls/es/hpcc.ts
kunalaswani: The actual phrases for the get calls in the english file do not include get as it was not required, but the tag for it retained the get. The English version does not include Get, in the tab names on ECL Watch for reference.

rpastrana: fair enough, that's why I keep calling for these translation to provide the actual English text for context. 
With that said, the phrases on their own don't make much sense, but do not seem to be incorrect.

kunalaswani: Retained translations as is without including get to conform with the english version.

rpastrana (approved):
@kunalaswani good to go

PR #16589, opened 2022-09-30
HPCC-28341 Fix EspBinding to allow no dali with no AuthDomains

asselitx:
Do not rely on readAuthDomainCfg() to determine the AuthType for checking when no dali is active. Because it defaults to AuthTypeMixed when no AuthDomains are configured it must be called after handling bindings with no dali active. The configmgr is one ESP without AuthDomains configured that cannot be assumed to be using AuthTypeMixed.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Cloud-compatibility
  - [x] Premature optimization

## Smoketest:

## Testing:

Tested ECLWatch, configmgr and a DESDL service with dali, and without where possible.

### DIFF
--- a/esp/bindings/http/platform/httpbinding.cpp
+++ b/esp/bindings/http/platform/httpbinding.cpp
@@ -461,15 +461,20 @@ EspHttpBinding::EspHttpBinding(IPropertyTree* tree, const char *bindname, const 
     else
         domainName.set("default");
 
-    readAuthDomainCfg(proc_cfg);
-
     //Even for non-session based environment, the sessionIDCookieName may be used to
     //remove session related cookies cached in some browser page.
     sessionIDCookieName.setf("%s%d", SESSION_ID_COOKIE, m_port);
     if (!m_secmgr.get() || !daliClientActive())
     {
-        if (!daliClientActive() && domainAuthType != AuthPerRequestOnly)
-            throw MakeStringException(-1, "ESP binding %s cannot use AuthDomain named '%s'. Since no dali is active only AuthPerRequestOnly domains are allowed.", bindname, domainName.str());
+        VStringBuffer xpath("AuthDomains/AuthDomain[@domainName=\"%s\"]", domainName.get());
+        IPropertyTree* authDomainTree = proc_cfg->queryPropTree(xpath);
+        if (!daliClientActive() && authDomainTree)
+        {
+            const char* authType = authDomainTree->queryProp("@authType");
+            // Missing authType is assumed to be AuthTypeMixed - see readAuthDomainCfg()
+            if (isEmptyString(authType) || !strieq(authType, "AuthPerRequestOnly"))
+                throw MakeStringException(-1, "ESP binding %s cannot use AuthDomain named '%s'. Since no dali is active only AuthPerRequestOnly domains are allowed.", bindname, domainName.str());
+        }
 
         domainAuthType = AuthPerRequestOnly;
         Owned<IPropertyTree> authcfg = proc_cfg->getPropTree("Authentication");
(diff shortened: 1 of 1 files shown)


esp/bindings/http/platform/httpbinding.cpp
```diff
@@ -461,15 +461,19 @@ EspHttpBinding::EspHttpBinding(IPropertyTree* tree, const char *bindname, const
...
     if (!m_secmgr.get() || !daliClientActive())
     {
-        if (!daliClientActive() && domainAuthType != AuthPerRequestOnly)
-            throw MakeStringException(-1, "ESP binding %s cannot use AuthDomain named '%s'. Since no dali is active only AuthPerRequestOnly domains are allowed.", bindname, domainName.str());
+        VStringBuffer xpath("AuthDomains/AuthDomain[@domainName=\"%s\"]", domainName.get());
+        IPropertyTree* authDomainTree = proc_cfg->queryPropTree(xpath);
+        if (!daliClientActive() && authDomainTree)
+        {
+            const char* authType = authDomainTree->queryProp("@authType");
+            if(!isEmptyString(authType) && !strieq(authType, "AuthPerRequestOnly"))
```
wangkx, line 474: Need a space after the if.
In readAuthDomainCfg(), if isEmptyString(authType), the domainAuthType will be set as AuthTypeMixed. So, should this line be changed as: if (isEmptyString(authType) || !strieq(authType, "AuthPerRequestOnly")) ?

wangkx (commented):
@asselitx one comment

esp/bindings/http/platform/httpbinding.cpp
asselitx: I also considered that, but I _think_ that this way matches the assumptions of the original behavior before I made any changes. Previously we'd only assume domainAuthType = AuthTypeMixed for configurations that didn't fall into the `if (!m_secmgr.get() || !daliClientActive())` clause when domainAuthType was missing _or_ if the whole authDomainTree was missing. Likewise, _inside_ the `if (!m_secmgr.get() || !daliClientActive())` clause, when there were no AuthDomains defined at all we assumed that domainAuthType = AuthPerRequestOnly. It seemed reasonable to assume that inside that clause that a missing authType _also_ implies AuthPerRequestOnly just as a missing AuthDomain had previously implied AuthPerRequestOnly.

Before I made any changes I believe that a configuration with `!m_secmgr.get() || !daliClientActive()` and a missing authType inside an AuthDomain would have thrown an exception- so never would have been accepted as valid. So the question is should a missing authType _always_ imply AuthTypeMixed, or only in cases where it previously did.

However, I defer to your and Tony's judgement if there is a possibility I haven't considered.

wangkx: If a missing authType, the domainAuthType is set to AuthTypeMixed. That is the original design (please see the readAuthDomainCfg() in https://github.com/hpcc-systems/HPCC-Platform/commit/4c62fd501d9496840bdb24f2117124463e9818bb#diff-0e174708d3b6ebe19c80dca5bd73d300646b159c9730101e56596d9d31d04623R317). If I remember correctly, we never changed that. @afishbeck any comment?

afishbeck (approved):

esp/bindings/http/platform/httpbinding.cpp
afishbeck: Yes we defaulted to authTypeMixed because it gave us features of both worlds.  If someone wanted one world or the other then they had to choose.

wangkx (approved):
@asselitx looks fine

PR #16596, opened 2022-09-30
HPCC-28353 Std.File.GetNoCommonDefault() doesn't pick up the value of noCommon from environment.xml.

AttilaVamos:
Add missing xpath

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

## Testing:

Tested manually

### DIFF
--- a/plugins/fileservices/fileservices.cpp
+++ b/plugins/fileservices/fileservices.cpp
@@ -3159,7 +3159,7 @@ FILESERVICES_API bool FILESERVICES_CALL fsGetNoCommonDefault()
         Owned<IPropertyTree> env = getEnvironmentTree(daliEnv);
 
         if (env.get())
-            noCommonDef = env->getPropBool("@noCommon", noCommonDef);
+            noCommonDef = env->getPropBool("Software/Globals/@noCommon", noCommonDef);
     }
 #endif
     return (bool)noCommonDef;


plugins/fileservices/fileservices.cpp
```diff
@@ -3159,7 +3159,7 @@ FILESERVICES_API bool FILESERVICES_CALL fsGetNoCommonDefault()
         Owned<IPropertyTree> env = getEnvironmentTree(daliEnv);
 
         if (env.get())
-            noCommonDef = env->getPropBool("@noCommon", noCommonDef);
+            noCommonDef = env->getPropBool("Software/Globals@noCommon", noCommonDef);
```
ghalliday, line 3162: @AttilaVamos did you test this?
@jakesmith Does this work?  I thought it needed to be "Software/Globals/@noCommon"

ghalliday (changes requested):

plugins/fileservices/fileservices.cpp
jakesmith: should be "Software/Globals/@noCommon"

@AttilaVamos - it does suggest the previous PR was not adequately tested? (i.e. need to make sure an environment setting change does actually get picked up and used)

jakesmith: Also, I suspect this (and similar easy to spot issues in previous PR) would be spotted by you, if you scanned the changes after you issued the PR, i.e the PR issuer should always review their own PR 1st to spot issues.
As per the devdoc/CodeReview.md document:
```
* Check for silly mistakes\
  Scan the code after creating the PR to check there are no tracing/other debugging artifacts or typos.  It doesn't take long, and often catches trivial issues.  It can avoid the need for a cycle of code-review/fixes.
```

jakesmith (changes requested):
@AttilaVamos - see comments

plugins/fileservices/fileservices.cpp
AttilaVamos: Yes, I tested it manually with change the value of the  "noCommon" attribute in envirionment.xml, then restart HPCC and the changed value appeared in DFU Server log.

ghalliday:
@jakesmith see code in getTailIdLength().  I think it may work even though I didn't expect it to.

plugins/fileservices/fileservices.cpp
jakesmith: ok, yes I meant to say simplify by removing unecessary Owned object, didn't mean to change path used.

AttilaVamos:
@ghalliday @jakesmith  I opened a 8.10.0 targeted PR for this fix: https://github.com/hpcc-systems/HPCC-Platform/pull/16603

jakesmith:
> @jakesmith see code in getTailIdLength(). I think it may work even though I didn't expect it to.

hmmm, yes, I think it may, but should not.
That should be fixed (so it doesn't work). I'll open a separate JIRA (when track is working again)

AttilaVamos:
@jakesmith so I should change the 'noCommon' to 'nocommon' and wait until your fix will be merged?

jakesmith:
> @ghalliday @jakesmith I opened a 8.10.0 targeted PR for this fix: #16603

@AttilaVamos  better to retarget this one PR than open a fresh one.

jakesmith:
> @jakesmith so I should change the 'noCommon' to 'nocommon' and wait until your fix will be merged?

my fix is unrelated. It is to stop blah@something being treated as blah/@something.

Afaik "noCommon" (cased) is correct, that is what dfuserver is expecting, the only place afaik that it is forced to lower case and picked up separately is from the dfu command line (where it comes in a uncased IProperties tree), and uses that option to set the spray request field.

AttilaVamos:
How can I change the base back from master to 8.10.0 without messing-up?

jakesmith:
> How can I change the base back from master to 8.10.0 without messing-up?

You have to rebase your branch and re-push.

e.g. 
git rebase --onto upstream/candidate-8.10.x <SHA>

where SHA is e.g. HEAD^^ if your branch has 2 commits ..

jakesmith (approved):
@AttilaVamos - looks correct, please squash

AttilaVamos:
@jakesmith @ghalliday squashed.

ghalliday (approved):

PR #16608, opened 2022-10-04
HPCC-28348 Optimize IF(cond) when cond already true/false

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/ecl/hqlcpp/hqlcpp.cpp
+++ b/ecl/hqlcpp/hqlcpp.cpp
@@ -7726,36 +7726,59 @@ void HqlCppTranslator::doBuildAssignIf(BuildCtx & ctx, const CHqlBoundTarget & t
         return;
     }
 
+    IHqlExpression * condExpr = expr->queryChild(0);
     IHqlExpression * trueExpr = expr->queryChild(1);
     IHqlExpression * falseExpr = expr->queryChild(2);
 
-    BuildCtx subctx(ctx);
-    CHqlBoundExpr cond;
-    buildCachedExpr(subctx, expr->queryChild(0), cond);
+    HqlExprAssociation * boundCond = ctx.queryMatchExpr(condExpr);
+    if (boundCond && boundCond->queryExpr()->queryValue())
+    {
+        if (matchesConstValue(boundCond->queryExpr(), true))
+            buildExprAssign(ctx, target, trueExpr);
+        else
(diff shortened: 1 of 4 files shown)


ghalliday:
Quite a few significant improvements.  E.g. bug86867 reduced by 80%.  A couple of very minor increases I still need to investgate, but would merge as is if I don't discover anything.

richardkchapman:
Surprised that this has ot already been removed by constant folder...

ghalliday:
It isn't removed by the constant folder because it is often non-obvious.  For instance

```
x := IF(cond, a, b)
y := PROJECT(x, transform());
z := IF(cond, y, z);
```

In the context of evaluating z, this can be optimized to

```
x := a;
y := PROJECT(x, transform());
z := IF(cond, y, z);
```

because x is only evaluated on the true branch.  However if you apply that optimization generally, and y is used from somewhere else you will end up evaluating both  PROJECT(a, transform()) and PROJECT(IF(cond, a,b), transform) - increasing the amount of work.
One of the optimizers tries to take that into account, but it is very hard to accurate count for child queries, so it tends to be concervative.

ghalliday:
Wrong jira used in commit message - should be 28352.  I will fix.

It isn't removed by the constant folder because it is often non-obvious.  For instance


```
x := IF(cond, a, b)
y := PROJECT(x, transform());
z := IF(cond, y, z);
```

In the context of evaluating z, this can be optimized to

```
x := a;
y := PROJECT(x, transform());
z := IF(cond, y, z);
```

because x is only evaluated on the true branch.  However if you apply that optimization generally, and y is used from somewhere else you will end up evaluating both  PROJECT(a, transform()) and PROJECT(IF(cond, a,b), transform) - increasing the amount of work.
One of the optimizers tries to take that into account, but it is very hard to accurate count for child queries, so it tends to be conservative.

richardkchapman (approved):

ghalliday:
@richardkchapman rerun the regression suite.  no queries got larger.   I have another similar change which I will push as an extra commit.

richardkchapman (approved):

PR #16613, opened 2022-10-05
HPCC-28368 Roxie may be blocked by non-SSL probe connections to SSL port

richardkchapman:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/roxie/ccd/ccdprotocol.cpp
+++ b/roxie/ccd/ccdprotocol.cpp
@@ -128,6 +128,8 @@ public:
         started.wait();
     }
 
+    virtual ISocket *createSecureSocket (ISocket *base) const = 0;
+
     virtual bool stop()
     {
         if (running)
@@ -225,6 +227,7 @@ class ProtocolSocketListener : public ProtocolListener
     StringAttr keyFile;
     StringAttr passPhrase;
     Owned<ISecureSocketContext> secureContext;
+    bool isSSL = false;
 
 public:
     ProtocolSocketListener(IHpccProtocolMsgSink *_sink, unsigned _port, unsigned _listenQueue, const char *_protocol, const char *_certFile, const char *_keyFile, const char *_passPhrase)
@@ -237,6 +240,11 @@ public:
         certFile.set(_certFile);
         keyFile.set(_keyFile);
         passPhrase.set(_passPhrase);
+        isSSL = streq(protocol.str(), "ssl");
+#ifdef _USE_OPENSSL
+        if (isSSL)
+            secureContext.setown(createSecureSocketContextEx(certFile.get(), keyFile.get(), passPhrase.get(), ServerSocket));
+#endif
     }
 
     IHpccProtocolMsgSink *queryMsgSink()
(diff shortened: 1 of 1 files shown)


roxie/ccd/ccdprotocol.cpp
```diff
@@ -1523,12 +1533,13 @@ class RoxieSocketWorker : public ProtocolQueryWorker
...
     virtual void init(void *_r) override
     {
-        client.setown(new CSafeSocket((ISocket *) _r));
+        rawClient.setown((ISocket *) _r);
         ProtocolQueryWorker::init(_r);
     }
 
     virtual void threadmain() override
     {
+        client.setown(new CSafeSocket(rawClient.getClear()));
```
mckellyln, line 1542: Still do not think this runs the [potentially blocking] secure_accept() in the pooled thread ?
Perhaps:
client.setown(new CSafeSocket(createSecureSocket(rawClient.getClear());
or something like that ?  Handling exceptions etc.

roxie/ccd/ccdprotocol.cpp
```diff
@@ -286,50 +331,14 @@ class ProtocolSocketListener : public ProtocolListener
         while (running)
         {
             Owned<ISocket> client = socket->accept(true);
-            Owned<ISecureSocket> ssock;
             if (client)
             {
-                if (streq(protocol.str(), "ssl"))
+                if (isSSL)
```
mckellyln, line 336: I think we can remove this if (isSSL) block

richardkchapman: Just checking you were awake... or proving that I was not, or something...

Pushed another fix

mckellyln:
Looks good.  Let me test it and then I'll approve.

roxie/ccd/ccdprotocol.cpp
```diff
@@ -1523,12 +1522,13 @@ class RoxieSocketWorker : public ProtocolQueryWorker
...
     virtual void init(void *_r) override
     {
-        client.setown(new CSafeSocket((ISocket *) _r));
+        rawClient.setown((ISocket *) _r);
         ProtocolQueryWorker::init(_r);
     }
 
     virtual void threadmain() override
     {
+        client.setown(new CSafeSocket(listener->createSecureSocket(rawClient.getClear())));
```
mckellyln, line 1531: I think createSecureSocket() does not throw, should we check here if client is nullptr and avoid the doMain() call ?

mckellyln (commented):
I've tested this with a port checker that holds on to the port - and it works great.
Just one last comment about if there is ever a problem in the secure_accept() code path - do we throw/skip calling doMain() ?

roxie/ccd/ccdprotocol.cpp
ghalliday: Yes.  client will be non null, but the result of createSecureSocket may be null.

mckellyln: Owned<ISocket> clisock = listener->createSecureSocket(rawClient.getClear());
        if (clisock)
        {
            client.setown(new CSafeSocket(clisock.getClear()));
            doMain("");
        }
        else
            rawClient->shutdownNoThrow();

mckellyln: Shutdown rawClient to help other side see/feel this ?

mckellyln (commented):
Just one last comment about if we should shutdown rawClient if there is a failure ?

mckellyln (commented):
I don't think my idea works, sorry.
We should revert this last commit.
Perhaps there is some other way to shutdown socket upon failure, but not sure how just yet.
But would new CSafeSocket(nullptr) always return non-NULL ?

ghalliday (changes requested):

ghalliday (changes requested):
Still another issue

roxie/ccd/ccdprotocol.cpp
```diff
@@ -1523,13 +1522,24 @@ class RoxieSocketWorker : public ProtocolQueryWorker
...
     {
-        client.setown(new CSafeSocket((ISocket *) _r));
+        rawClient.setown((ISocket *) _r);
         ProtocolQueryWorker::init(_r);
     }
 
     virtual void threadmain() override
     {
-        doMain("");
+        ISocket *secure = listener->createSecureSocket(rawClient);
```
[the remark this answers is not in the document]
mckellyln: Yes, rawClient.getLink() perhaps.

mckellyln (approved):
Looks good now.

mckellyln (approved):
Good to add back in compile check.
Approved.

ghalliday (approved):
please squash

PR #16631, opened 2022-10-13
HPCC-28110 Fix issues reading/copying multi cluster file

jakesmith:
Reading a foreign file that exists on multiple clusters, caused problems at the client side.
The deserialized groups referencing the foreign locations, are instatiated in the local environment as 'anon' groups. Then, when the deserialized IFileDescriptor info is used to create a IDistributedFile and later used again to serialize, the 'anon' groups are seen as a local match and used in place of the original groups, and as a side-effect also replace the remote base directories with the local env. default directories, appending 'unknown' to the end of them.

This in turn causes the adjustClusterDir/replaceClusterDir functions, to misbehave and fail to substitue in the preferred cluster.
The consequence is that the prefered cluster is used, but with the wrong base directory, and the physical file parts are not found.

Fix by ensuring the original deserialized foreign group is preserved in the IClusterInfo's.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dali/base/dafdesc.cpp
+++ b/dali/base/dafdesc.cpp
@@ -390,6 +390,7 @@ struct CClusterInfo: implements IClusterInfo, public CInterface
     Linked<IGroup> group;
     StringAttr name; // group name
     ClusterPartDiskMapSpec mspec;
+    bool foreignGroup = false;
     void checkClusterName(INamedGroupStore *resolver)
     {
         // check name matches group
@@ -461,9 +462,14 @@ public:
             return;
         name.set(pt->queryProp("@name"));
         mspec.fromProp(pt);
+        if (0 != (flags & IFDSF_FOREIGN_GROUP))
+            foreignGroup = true;
+        else if (pt->getPropBool("@foreign"))
+            foreignGroup = true;
+
         if ((((flags&IFDSF_EXCLUDE_GROUPS)==0)||name.isEmpty())&&pt->hasProp("Group"))
             group.setown(createIGroup(pt->queryProp("Group")));
-        if (!name.isEmpty()&&!group.get()&&resolver)
+        if (!name.isEmpty()&&!group.get()&&resolver&&!foreignGroup)
         {
             StringBuffer defaultDir;
             GroupType groupType;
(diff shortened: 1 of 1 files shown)


dali/base/dafdesc.cpp
```diff
@@ -390,6 +390,7 @@ struct CClusterInfo: implements IClusterInfo, public CInterface
     Linked<IGroup> group;
     StringAttr name; // group name
     ClusterPartDiskMapSpec mspec;
+    bool foreignGroup = false;
```
ghalliday, line 393: How does foreignGroup ever get set?  It looks like it is only set when deserializing from a property tree.  Is that initially in extracting the information from the foreign dali when IFDSF_FOREIGN_GROUP is set?
When it is coming from a serialized foreign dali it then has the foreign flag set?

jakesmith: yes, when it gets a foreign file, it gets a property tree, and calls deserializeFileDescriptorTree with flags=IFDSF_FOREIGN_GROUP, which constructs CClusterInfo via a property tree, passing through the IFDSF_FOREIGN_GROUP.
That ctor, sets foreignGroup and it is preserved through serialization (when CDistributedFile::getFileDescriptor serializes) through @foreign in serializeTree.

>When it is coming from a serialized foreign dali it then has the foreign flag set?

so it's not coming from the remote side, the:
```
    Owned<IFileDescriptor> fdesc = deserializeFileDescriptorTree(tree,&queryNamedGroupStore(),IFDSF_FOREIGN_GROUP);
```

in CDistributedFileDirectory::getFile( on a foreign will propagate it through during deserialization.

AttilaVamos:
Tested on my environment, where the source is a 2 nodes platform, one node is a Thor (IP: x.x.x.58) and the second is a Roxie cluster (IP: x.x.x.54). The target system is a one node cluster (IP: x.x.x.52). Logically the same configuration as where the problem occurred (Fido prod and Fido Azure)  

After creating a file into both clusters, I remote copied that into the target system (in that case the preferred/closest cluster was Roxie, second in the file MetaInfo.) and the process finished without any problem.

ghalliday (approved):

PR #16650, opened 2022-10-24
Log refactor: step 1 - remove detail setting

richardkchapman:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/thorlcr/activities/choosesets/thchoosesetsslave.cpp
+++ b/thorlcr/activities/choosesets/thchoosesetsslave.cpp
@@ -116,24 +116,24 @@ class ChooseSetsActivity : public BaseChooseSetsActivity
         if (!receiveMsg(msg, queryJobChannel().queryMyRank()-1, mpTag))
             return;
         memcpy(tallies, msg.readDirect(numSets*sizeof(unsigned)), numSets*sizeof(unsigned));
-        if (!REJECTLOG(MCthorDetailedDebugInfo))
+        if (container.doDetailedTracing())
         {
             StringBuffer s;
             unsigned idx=0;
             for (; idx<numSets; idx++)
                 s.append("[").append(tallies[idx]).append("]");
-            ::ActPrintLog(this, thorDetailedLogLevel, "CHOOSESETS: Incoming count = %s", s.str());
+            ::ActPrintLog(this, TraceFlags::Detailed, "CHOOSESETS: Incoming count = %s", s.str());
         }
     }
     void sendTallies() // NB: not called on last node.
     {
-        if (!REJECTLOG(MCthorDetailedDebugInfo))
+        if (container.doDetailedTracing())
         {
             StringBuffer s;
             unsigned idx=0;
             for (; idx<numSets; idx++)
                 s.append("[").append(tallies[idx]).append("]");
-            ::ActPrintLog(this, thorDetailedLogLevel, "CHOOSESETS: Outgoing count = %s", s.str());
+            ::ActPrintLog(this, TraceFlags::Detailed, "CHOOSESETS: Outgoing count = %s", s.str());
         }
         CMessageBuffer msg;
         msg.append(numSets * sizeof(unsigned), tallies); 
@@ -367,7 +367,7 @@ class ChooseSetsLastActivity : public ChooseSetsPlusActivity
         for (unsigned idx=0; idx < numSets; idx++)
         {
             rowcount_t firstToCopy = 0;
-            ::ActPrintLog(this, thorDetailedLogLevel, "CHOOSESETSLAST: %d P(%" RCPF "d) C(%" RCPF "d) L(%" I64F "d) T(%" RCPF "d)", idx, priorCounts[idx], counts[idx], limits[idx], totalCounts[idx]);
+            ::ActPrintLog(this, TraceFlags::Detailed, "CHOOSESETSLAST: %d P(%" RCPF "d) C(%" RCPF "d) L(%" I64F "d) T(%" RCPF "d)", idx, priorCounts[idx], counts[idx], limits[idx], totalCounts[idx]);
             if (((rowcount_t)limits[idx]) < totalCounts[idx])
                 firstToCopy = totalCounts[idx] - (rowcount_t)limits[idx];
             if (priorCounts[idx] + counts[idx] > firstToCopy)
@@ -378,7 +378,7 @@ class ChooseSetsLastActivity : public ChooseSetsPlusActivity
                     numToSkip[idx] = (unsigned)(firstToCopy - priorCounts[idx]);
                 numToReturn[idx] = (unsigned)(priorCounts[idx] + counts[idx] - firstToCopy);
                 skipAll = false;
-                ::ActPrintLog(this, thorDetailedLogLevel, "CHOOSESETSLAST: Selection %d.  Range(%d,%" RCPF "d)", idx, numToSkip[idx], counts[idx]);
+                ::ActPrintLog(this, TraceFlags::Detailed, "CHOOSESETSLAST: Selection %d.  Range(%d,%" RCPF "d)", idx, numToSkip[idx], counts[idx]);
             }
         }
         if (skipAll)
--- a/thorlcr/graph/thgraph.cpp
+++ b/thorlcr/graph/thgraph.cpp
@@ -386,11 +386,23 @@ CGraphElementBase::CGraphElementBase(CGraphBase &_owner, IPropertyTree &_xgmml, 
     alreadyUpdated = false;
     whichBranch = (unsigned)-1;
     log = true;
+    VStringBuffer traceOptName("log%s", activityKindStr(kind));
+    traceOptName.stripChar(' ').toLowerCase();
+    traceFlags = (TraceFlags) queryXGMML().getPropInt("hint[@name=\"traceFlags\"]/@value", queryJob().getOptInt(traceOptName, (int) TraceFlags::Standard));;
     sentActInitData.setown(createThreadSafeBitSet());
     maxCores = queryXGMML().getPropInt("hint[@name=\"max_cores\"]/@value", 0);
     if (0 == maxCores)
         maxCores = queryJob().queryMaxDefaultActivityCores();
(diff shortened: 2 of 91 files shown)


thorlcr/graph/thgraph.cpp
```diff
@@ -3221,6 +3224,9 @@ CActivityBase::CActivityBase(CGraphElementBase *_container, const StatisticsMapp
     defaultRoxieMemHeapFlags = (roxiemem::RoxieHeapFlags)container.getOptInt("heapflags", defaultHeapFlags);
     if (container.queryJob().queryUsePackedAllocators())
         defaultRoxieMemHeapFlags = (roxiemem::RoxieHeapFlags)(defaultRoxieMemHeapFlags | roxiemem::RHFpacked);
+    if (traceOptName)
+        traceActivity = getOptBool(traceOptName, false);
+    container.setDetailedTracing(traceActivity);
```
jakesmith, line 3229: I think this should be other way around, i.e. the container probes the option, and this (activity object) either uses the container.traceActivity flag, or copies it locally.

(see other comment re. traceOptName to container)

thorlcr/activities/choosesets/thchoosesetsslave.cpp
```diff
@@ -30,7 +30,7 @@ class BaseChooseSetsActivity : public CSlaveActivity
     unsigned *tallies;
 
 public:
-    BaseChooseSetsActivity(CGraphElementBase *_container) : CSlaveActivity(_container)
+    BaseChooseSetsActivity(CGraphElementBase *_container) : CSlaveActivity(_container, basicActivityStatistics, "traceChooseSets")
```
jakesmith, line 33: See other comment in CActivityBase ctor.

Suggestion: the flag is set in the container object (CGraphElementBase).
The option name could be based on getActivityText (by default), or a space stripped version of it perhaps, it's probably close enough to what we want in most cases, but could have a separate helper for exceptions.

thorlcr/activities/choosesets/thchoosesetsslave.cpp
```diff
@@ -367,7 +367,7 @@ class ChooseSetsLastActivity : public ChooseSetsPlusActivity
         for (unsigned idx=0; idx < numSets; idx++)
         {
             rowcount_t firstToCopy = 0;
-            ::ActPrintLog(this, thorDetailedLogLevel, "CHOOSESETSLAST: %d P(%" RCPF "d) C(%" RCPF "d) L(%" I64F "d) T(%" RCPF "d)", idx, priorCounts[idx], counts[idx], limits[idx], totalCounts[idx]);
+            ::ActPrintTrace(this, "CHOOSESETSLAST: %d P(%" RCPF "d) C(%" RCPF "d) L(%" I64F "d) T(%" RCPF "d)", idx, priorCounts[idx], counts[idx], limits[idx], totalCounts[idx]);
```
jakesmith, line 370: question: the intention is to change all other ActPrintLog* functions to ActPrintTrace I assume?

thorlcr/activities/choosesets/thchoosesetsslave.cpp
```diff
@@ -378,7 +378,7 @@ class ChooseSetsLastActivity : public ChooseSetsPlusActivity
                     numToSkip[idx] = (unsigned)(firstToCopy - priorCounts[idx]);
                 numToReturn[idx] = (unsigned)(priorCounts[idx] + counts[idx] - firstToCopy);
                 skipAll = false;
-                ::ActPrintLog(this, thorDetailedLogLevel, "CHOOSESETSLAST: Selection %d.  Range(%d,%" RCPF "d)", idx, numToSkip[idx], counts[idx]);
+                ::ActPrintTrace(this, "CHOOSESETSLAST: Selection %d.  Range(%d,%" RCPF "d)", idx, numToSkip[idx], counts[idx]);
```
jakesmith, line 381: unrelated/not new: "CHOOSETSLAST" should be removed from the message, since ActPrintTrace will always trace the activity name.

thorlcr/graph/thgraph.hpp
```diff
@@ -475,8 +476,10 @@ class graph_decl CGraphElementBase : public CInterface, implements IInterface
     unsigned getOutputs() const { return outputs.ordinality(); }
     bool isSource() const { return isActivitySource(kind); }
     bool isSink() const { return sink; }
-    inline bool doLogging() const { return log; }
+    inline bool doLogging() const { return log || detailedTracing; }
     inline void setLogging(bool _log) { log = _log; }
+    inline bool doDetailedTracing() const { return detailedTracing; }
+    inline void setDetailedTracing(bool _detailedTracing) { detailedTracing = _detailedTracing; }
```
jakesmith, line 482: could we make this an unsigned so it can later be expanded to be a feature set at the activity level, e.g. for sort, to trace partitioning only, or merge only, etc. ?

thorlcr/graph/thgraph.hpp
```diff
@@ -475,8 +476,10 @@ class graph_decl CGraphElementBase : public CInterface, implements IInterface
     unsigned getOutputs() const { return outputs.ordinality(); }
     bool isSource() const { return isActivitySource(kind); }
     bool isSink() const { return sink; }
-    inline bool doLogging() const { return log; }
+    inline bool doLogging() const { return log || detailedTracing; }
```
jakesmith, line 479: I think we should drop of doLogging/setLogging and replace with doDetailedLogging(),
but maybe in next PR..

thorlcr/activities/choosesets/thchoosesetsslave.cpp
richardkchapman: No - trace implies "Only if asked for" while log is "always" (we may want better names). The variety of ActPrintLog that took a detail level always used the higher level...

thorlcr/graph/thgraph.hpp
richardkchapman: They are slightly different (see the distinction between logging and tracing, above).

jakesmith (commented):
@richardkchapman - a couple of questions/comments

system/jlib/jtrace.hpp
```diff
@@ -0,0 +1,66 @@
...
+    flag8 = 0x200,
+    flag9 = 0x400,
+    flag10 = 0x800,
+    flag11 = 0x1000,
+    flag12 = 0x2000,
+    flag13 = 0x4000,
+    flag14 = 0x8000,
+};
+BITMASK_ENUM(TraceFlags);
+
```
ghalliday, line 48: Some comments.

I think the idea broadly works, but I would change it in the following ways:

- include all the flags for all the components in this file.
- use a different name for the jlib flags (and any other very widely-shared library e.g. jhtree) - so that they can be added and flags renumbered without causing grief.

Introduce a helper function to make it easy to define and process the options for a component.  Something along the lines of....
```

struct TraceOption { const char * name; unsigned value; };
#define OPT(NAME) { "@" # NAME, TraceFlag::NAME }
TraceFlag getTraceFlags(IConstWorkunit * wu, const std::initializer_list<TraceOption> & y, TraceFlag dft);
TraceFlag getTraceFlags(IPropertyTree * globals, const std::initializer_list<TraceOption> & y, TraceFlag dft);
-----
constexpr std::initializer_list<TraceOption> roxieOptions{ OPT(traceX), OPT(traceY)) };
roxieTraceFlags = getTraceFlags(globals, roxieOptions, TraceFlag::none);
roxieTraceFlags = getTraceFlags(wu, roxieOptions, roxieTraceFlags);
```

There would cover the vast majority of cases.  There are two cases that may need different processing, where the category may be more fine grained.
- tracing on per activity type (in thor/roxie?)
- tracing on a per transform type (in code generator)

I'm not sure if that granularity is need in thor.  It would be useful in the code generator, but it looks like there are n the code generator I'm not sure how many different types if transform there are 107 different transformations so they wouldn't fit into that scheme.  Probably something similar using an array of booleans would work. It can be within the code generator - but any other component with similar issues should use the same approach, so the helper functions should be global.

ghalliday (changes requested):
@richardkchapman Very useful for clarifying what the issues are, but I think this would be too big to merge.  Better to make the changes to allow the switch - e.g. adding jtrace and helper functions.  Then switch different components over, and finally remove the detail level.

Most of my comments are attached as a comment to trace.hpp.

system/jlib/jtrace.hpp
```diff
@@ -0,0 +1,66 @@
...
+    See the License for the specific language governing permissions and
+    limitations under the License.
+############################################################################## */
+
+
+
+#ifndef JTRACE_HPP
+#define JTRACE_HPP
+
+enum class TraceFlags : unsigned
```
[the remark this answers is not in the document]
richardkchapman: There's a difference between "Normally output" and "Normally output if a particular feature is being traced". Standard level means "A normal level of tracing given you specifically requested tracing of this feature", detailed means "more than that" and max means "give me it all" - but all within the context of "if you asked to trace this feature, which would normally not be the case....

common/thorhelper/roxiehelper.hpp
```diff
@@ -29,6 +29,26 @@
 
 //========================================================================================= 
 
+// Feature trace flags for hthor, thor and Roxie engines
+
+// Common to several engines
+constexpr TraceFlags traceHttp = TraceFlags::flag1;
+
+// Specific to Roxie
```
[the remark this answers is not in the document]
richardkchapman: I put them in a separate file temporarily to reduce rebuilds while developing... I agree they should all be jtrace.hpp

roxie/udplib/udplib.cmake
```diff
@@ -40,6 +40,11 @@ include_directories (
          ./../../system/include 
          ./../../system/jlib 
          ./../../roxie/ccd
+         ./../../common/thorhelper
```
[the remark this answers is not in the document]
richardkchapman: Temporary...

PR #16652, opened 2022-10-25
HPCC-28410 Fix the code for filtering out duplicated LDAP BaseDNs

wangkx:
The BaseDNs are stored in the ws_access as a member variable. The BaseDNs are read from settings into the variable when the variable is used at the first time. There is the code to filter out some duplicated items in the setting. The code may not work when multiple threads access the code. In this PR, a CriticalSection is added so that only one thread can access the code at the same time.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/services/ws_access/ws_accessService.cpp
+++ b/esp/services/ws_access/ws_accessService.cpp
@@ -26,7 +26,7 @@
 
 #include <set>
 
-#define MSG_SEC_MANAGER_IS_NULL "Security manager is not found. Please check if the system authentication is set up correctly"
+#define MSG_SEC_MANAGER_IS_NULL "Security manager is not found, or is not LDAP. Please check if the system authentication is set up correctly."
 #define MSG_SEC_MANAGER_ISNT_LDAP "LDAP Security manager is required for this feature. Please enable LDAP in the system configuration"
 
 #define FILE_SCOPE_URL "FileScopeAccess"
@@ -37,11 +37,14 @@
 #define MAX_RESOURCES_DISPLAY 3000
 static const long MAXXLSTRANSFER = 5000000;
 
-void checkUser(IEspContext& context, const char* rtype = NULL, const char* rtitle = NULL, unsigned int SecAccessFlags = SecAccess_Full)
+void Cws_accessEx::checkUser(IEspContext& context, CLdapSecManager* secmgr, const char* rtype, const char* rtitle, unsigned int SecAccessFlags)
 {
-    CLdapSecManager* secmgr = dynamic_cast<CLdapSecManager*>(context.querySecManager());
-    if(secmgr == NULL)
-        throw MakeStringException(ECLWATCH_INVALID_SEC_MANAGER, MSG_SEC_MANAGER_IS_NULL);
+    if (secmgr == nullptr)
+    {
+        secmgr = queryLDAPSecurityManager(context);
+        if (secmgr == nullptr)
+            throw makeStringException(ECLWATCH_INVALID_SEC_MANAGER, MSG_SEC_MANAGER_IS_NULL);
+    }
 
     if (rtype && rtitle && strieq(rtype, FILE_SCOPE_RTYPE) && strieq(rtitle, FILE_SCOPE_RTITLE))
     {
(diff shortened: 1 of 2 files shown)


RussWhitehead (commented):
Do you need to add "setBasedns(context)" to Cws_accessEx::onResources, Cws_accessEx::onResourceQuery, etc?

wangkx:
@RussWhitehead looks like that there is only one place where the setBasedns() should be added. It has been called in all of other places (including Cws_accessEx::onResources and Cws_accessEx::onResourceQuery where the setBasedns() is called through the getBasednReq()). Please review.

RussWhitehead (approved):
Looks good

wangkx:
@richardkchapman or @ghalliday please merge.

esp/services/ws_access/ws_accessService.cpp
```diff
@@ -241,9 +240,12 @@ void Cws_accessEx::getBasednReq(IEspContext &context, const char* name, const ch
 void Cws_accessEx::setBasedns(IEspContext &context)
 {
     CLdapSecManager* secmgr = (CLdapSecManager*)(context.querySecManager());
+    if (secmgr == nullptr)
```
afishbeck, line 243: I know the line is not new, but since querySecManager() may return a different type of secmgr than LDAP I think it should be a dynamic_cast.  An old fashion cast or a static_cast may not detect that the manager is not compatible with ldap.
CLdapSecManager* secmgr = (CLdapSecManager*)(context.querySecManager());

CLdapSecManager* secmgr = dynamic_cast<CLdapSecManager*>(context.querySecManager());

RussWhitehead: @afishbeck Or better yet, call the new ISecObject::checkFeatures to check the level of Security Manager support (avoiding the peculiar dynamic_cast model).  But this should probably all be refactored in a single, separate PR

wangkx: Will change to: CLdapSecManager* secmgr = queryLDAPSecurityManager(context);
The same change has to be in many other places.

RussWhitehead: @wangkx I believe this will result in the message "Security manager is not found. Please check if the system authentication is set up correctly"  if there is a security manager, but it is not LDAP.  This is wrong, perhaps an improved error message stating something like "Security manager is not found, or is not LDAP. Please check if the system authentication is set up correctly"

wangkx: I may also combine the code into the checkUser().

wangkx:

ghalliday:
@afishbeck still waiting for your review...

esp/services/ws_access/ws_accessService.cpp
```diff
@@ -37,11 +37,14 @@
...
+void Cws_accessEx::checkUser(IEspContext& context, CLdapSecManager* secmgr, const char* rtype, const char* rtitle, unsigned int SecAccessFlags)
 {
-    CLdapSecManager* secmgr = dynamic_cast<CLdapSecManager*>(context.querySecManager());
-    if(secmgr == NULL)
-        throw MakeStringException(ECLWATCH_INVALID_SEC_MANAGER, MSG_SEC_MANAGER_IS_NULL);
+    if (secmgr == nullptr)
+    {
+        secmgr = queryLDAPSecurityManager(context);
+        if (secmgr == nullptr)
+            throw makeStringException(ECLWATCH_INVALID_SEC_MANAGER, MSG_SEC_MANAGER_IS_NULL);
```
afishbeck, line 46: If you add a flag to queryLDAPSecurityManager(IEspContext &context, bool required) which tells it to throw the exception if it results in nullptr then you can probably eliminate all of these duplicate throw lines.

Could be another JIRA though.

afishbeck: I see you got rid of most of them in this code already using queryLDAPSecurityManagerAndCheckUser()... you can decide if moving it up into queryLDAPSecurityManager(IEspContext &context, bool required) will eliminate even more, but again that's for another JIRA if you decide you want to do it.

afishbeck (approved):
@ghalliday one comment for a possible future change, but this PR looks good.

esp/services/ws_access/ws_accessService.cpp
wangkx: The JIRA is created: https://track.hpccsystems.com/browse/HPCC-28570.

PR #16659, opened 2022-10-26
HPCC-28432 Fix DfuPlusExec unable to call mtls eclservices

afishbeck:
Adds new internal only (undocumented) option "mtls-secret" to dfuplus. "mtls-secret" pics up the secret that sets the client cert generally used for connecting to cluster visibility services like eclservices.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/plugins/fileservices/fileservices.cpp
+++ b/plugins/fileservices/fileservices.cpp
@@ -2855,6 +2855,22 @@ static bool build_dfuplus_globals(int argc, const char *argv[], IProperties * gl
         if (strchr(argv[i],'='))
             globals->loadProp(argv[i]);
 
+    // if server is empty we can look up the settings for the default eclservices (eclwatch if eclservices is not found)
+    if (isEmptyString(globals->queryProp("server")))
+    {
+        const char *url = fsGetEspURL(nullptr, nullptr);
+        if (!isEmptyString(url))
+        {
+            if (strncmp("mtls:", url, 5)==0)
+            {
+                if (!globals->hasProp("mtls-secret"))
+                    globals->setProp("mtls-secret", "local");
+                url+=5;
+            }
+            globals->setProp("server", url);
+        }
+    }
+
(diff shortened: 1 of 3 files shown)


plugins/fileservices/fileservices.cpp
```diff
@@ -2855,6 +2855,23 @@ static bool build_dfuplus_globals(int argc, const char *argv[], IProperties * gl
         if (strchr(argv[i],'='))
             globals->loadProp(argv[i]);
 
+    // if server matches the exact string "eclservices" we can look up the settings for that service
+    const char *server = globals->queryProp("server");
+    if (!isEmptyString(server) && streq(server, "eclservices"))
```
jakesmith, line 2860: question: perhaps instead, this should be the default behaviour if no "server" is provided?
i.e. auto lookup the local service url and local secret, i.e. eclservices is implicit.

jakesmith (approved):
@afishbeck - looks good approved, but 1 question/suggestion.

plugins/fileservices/fileservices.cpp
afishbeck: Yes, agreed.  The thought had crossed my mind.  I think since the implicit server value would be handled in fileservices.DfuPlusExec() and not the command line dfuplus executable it will be fine.  I'll change it.

afishbeck: @jakesmith I've updated as discussed, can you check?

ghalliday (approved):
Looks sensible.  Will merge if @jakesmith is happy with it.

jakesmith (approved):
@afishbeck - looks good, please squash

PR #16666, opened 2022-11-01
HPCC-27585 Containerized builds allow for ECL library code signing

Michael-Gardner:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [x] Security
  - [ ] Thread-safety
  - [x] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/.github/workflows/build-and-publish.yml
+++ b/.github/workflows/build-and-publish.yml
@@ -28,6 +28,10 @@ jobs:
           latest: 1   # this should only be set on the current minor branch
           github_actor: ${{ secrets.LNB_ACTOR }}
           github_token: ${{ secrets.LNB_TOKEN }}
+          sign_modules: 1
+          signing_secret: ${{ secrets.SIGNING_SECRET }}
+          signing_keyid: ${{ secrets.SIGNING_KEYID }}
+          signing_passphrase: ${{ secrets.SIGNING_PASSPHRASE }}
 
   ml-builds:
     needs: build
@@ -70,3 +74,7 @@ jobs:
         build_ln: 1
         github_actor: ${{ secrets.LNB_ACTOR }}
         github_token: ${{ secrets.LNB_TOKEN }}
+        sign_modules: 1
+        signing_secret: ${{ secrets.SIGNING_SECRET }}
+        signing_keyid: ${{ secrets.SIGNING_KEYID }}
+        signing_passphrase: ${{ secrets.SIGNING_PASSPHRASE }}
--- a/dockerfiles/buildall-common.sh
+++ b/dockerfiles/buildall-common.sh
@@ -30,6 +30,11 @@ BUILD_TYPE=                                     # Set to Debug for a debug build
 DOCKER_REPO=hpccsystems
 DEST_DOCKER_REGISTRY=docker.io
 USE_CPPUNIT=1
+SIGN_MODULES=OFF
+SIGNING_SECRET=
+SIGNING_KEYID=HPCCSystems
+SIGNING_PASSPHRASE=
+KEY_COMMAND=
 
 # These values are set in a GitHub workflow build
 
@@ -38,6 +43,8 @@ USE_CPPUNIT=1
 [[ -n ${INPUT_GITHUB_TOKEN} ]] && GITHUB_TOKEN=${INPUT_GITHUB_TOKEN}
 [[ -n ${INPUT_BUILD_VER} ]] && BUILD_TAG=${INPUT_BUILD_VER}
 [[ -n ${INPUT_DOCKER_REPO} ]] && DOCKER_REPO=${INPUT_DOCKER_REPO}
+[[ -n ${INPUT_SIGN_MODULES} ]] && SIGN_MODULES=${INPUT_SIGN_MODULES}
+[[ -n ${INPUT_SIGNING_KEYID} ]] && SIGNING_KEYID=${INPUT_SIGNING_KEYID}
 
 DEST_DOCKER_REPO=${DOCKER_REPO}
 [[ -n ${INPUT_LN_DOCKER_REPO} ]] && DEST_DOCKER_REPO=${INPUT_LN_DOCKER_REPO}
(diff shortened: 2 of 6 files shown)


ghalliday (approved):
Looks good.  @jakesmith please check, and also what changes do we need to make to the actions so that our images builds start signing the library?

dockerfiles/buildall-common.sh
```diff
@@ -87,7 +112,8 @@ build_image() {
   [[ -z ${buildTag} ]] && buildTag=$BUILD_TAG
 
   if [ "$rebuild" = "1" ] || ! docker pull ${DEST_DOCKER_REGISTRY}/${DOCKER_REPO}/${name}:${label} ; then
-    docker image build -t ${DEST_DOCKER_REGISTRY}/${DEST_DOCKER_REPO}/${name}:${label} \
+    DOCKER_BUILDKIT=1 docker image build \
```
jakesmith, line 115: question: may be desirable anyway, but curious whether DOCKER_BUILDKIT=1 was required for this change?

dockerfiles/platform-build/Dockerfile
```diff
@@ -65,13 +65,28 @@ RUN mono `./vcpkg fetch nuget | tail -n 1` \
     -username "${GITHUB_ACTOR}" \
     -password "${GITHUB_TOKEN}"
 
+
 WORKDIR /hpcc-dev
 RUN mkdir build
 WORKDIR /hpcc-dev/build
 
+ARG SIGN_MODULES=OFF
+RUN --mount=type=secret,id=key,uid=10000 if [[ "${SIGN_MODULES}" != "OFF" ]] ; \
```
jakesmith, line 74: The uid and gid are confirable (via helm values), so uid is not necessarily 10000.
I'm not sure of the sifnificance/nee for it to be 10000 here though?
Could/should this mount/command run as user root instead?

dockerfiles/platform-build/README.md
```diff
@@ -0,0 +1,32 @@
+# Platform Build dockerfile
+
+
+## ECL code signing within docker containers
+
+ ### github action values
```
jakesmith, line 6: trivial: leading space unintentional?

jakesmith:
>also what changes do we need to make to the actions so that our images builds start signing the library?

Does .github/workflows/build-and-publish.yml need updating to pass 'sign_modules', 'signing_secret', 'signing_keyid' and 'signing_passphrase' onto the dockerfiles/action.yml ?  ( i.e. here: https://github.com/hpcc-systems/HPCC-Platform/blob/a6c6369d0360aa214a114a1ef83c81c5fbe56db0/.github/workflows/build-and-publish.yml#L25 )

dockerfiles/buildall-common.sh
Michael-Gardner: Buildkit isn't turned on by default at the moment.  On tests with github actions, without explicitly declaring the use of docker's buildkit, it was using the default builder that doesn't have support for mounting secrets into image layers.  This line just ensures it's turned on for our build so we have access to those newer features.

dockerfiles/platform-build/README.md
Michael-Gardner: Yes unintentional, fixing.

dockerfiles/platform-build/Dockerfile
Michael-Gardner: The 10000 uid is hardcoded on line 33 for our docker builds.  I don't believe this 'hpcc' user is the same one that gets configured within our helm code. Mounting it as root causes issues because I want the key to be visible in the gpg keychain for the hpcc user for when we run our cmake commands.

Michael-Gardner:
@jakesmith Yes those values will need to be added to the gpg secrets for the repository as well as the build-and-publish.yml workflow.  I tested it myself using a stripped down & modified version of build-and-publish on my own repo.  Ensure that the exported secret key uses the --armor flag though. The standard binary output was causing issues in my testing when I was loading it as a secret on my repository.

stuartort:
@jakesmith - if you are ok with comments by @Michael-Gardner, can this be merged?

jakesmith:
> and then added to the build-and-publish.yml workflow.

Is there any reason not to update the action now?
I presume that if the secrets are not defined in the github repo, then it should behave the smae as it does now, i.e. INPUT_SIGNING_SECRET etc. will be empty ?

dockerfiles/platform-build/Dockerfile
jakesmith: ok yes, this is the build image, the platform-core image uid and gid are configurable.

jakesmith:
@Michael-Gardner - 1 remaining question: https://github.com/hpcc-systems/HPCC-Platform/pull/16666#issuecomment-1309255526

Btw, this PR fell off my radar, because it is not on my PR watch list unless the reviewer status has been refreshed after replies.

Michael-Gardner:
@jakesmith I added the build-and-publish.yml changes to this PR as suggested.

They've been added to the build and ln-builds job only. The ml-builds job pulls from platform-core as it's base so it doesn't need to explicitly do any signing (as per my understanding).

.github/workflows/build-and-publish.yml
```diff
@@ -28,6 +28,10 @@ jobs:
           latest: 1   # this should only be set on the current minor branch
           github_actor: ${{ secrets.LNB_ACTOR }}
           github_token: ${{ secrets.LNB_TOKEN }}
+          sign_modules: 1
```
jakesmith, line 31: trivial: The README.md (in this PR) says:
```
The following values should be added to github action workflows that wish to
utilize the signing behavior.

> sign_modules: ON
```

this should probably match and use "ON" rather than 1.

.github/workflows/build-and-publish.yml
```diff
@@ -70,3 +74,7 @@ jobs:
         build_ln: 1
         github_actor: ${{ secrets.LNB_ACTOR }}
         github_token: ${{ secrets.LNB_TOKEN }}
+        sign_modules: 1
```
jakesmith, line 77: trivial: ( as per previous comment )

jakesmith:
> @jakesmith I added the build-and-publish.yml changes to this PR as suggested.
> 
> They've been added to the build and ln-builds job only. The ml-builds job pulls from platform-core as it's base so it doesn't need to explicitly do any signing (as per my understanding).

@Michael-Gardner - please do remember to press the refresh button on PR after updating a PR for re-review, without doing that, PR's do not appear on the reviewers' github pending PR list.
Also, generally it is better to leave changes as separate commits, and only squash when the PR review process is fully finished.
(It makes it a a lot easier to review what's actually changed between reviews)
There's a fuller description of these points and others here : https://github.com/hpcc-systems/HPCC-Platform/blob/master/devdoc/CodeReviews.md

jakesmith (approved):
@Michael-Gardner - looks good, 1 new trivial comment, but approved, can be merged as is

PR #16685, opened 2022-11-09
HPCC-28544 DFUServer crashes when source file is not distributed file.

AttilaVamos:
Check distributedSource/distributedTarget value before use.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

## Testing:

Tested manually.

### DIFF
--- a/dali/ft/filecopy.cpp
+++ b/dali/ft/filecopy.cpp
@@ -1522,16 +1522,28 @@ IAPICopyClient * FileSprayer::getAPICopyClient()
     if (needCalcCRC && !sources.item(0).hasCRC)
         return nullptr;
 
+    if (!distributedSource)
+        return nullptr;
+
     StringBuffer sourceClusterName;
     distributedSource->getClusterName(0, sourceClusterName);
     Owned<IStoragePlane> sourcePlane = getDataStoragePlane(sourceClusterName.str(), false);
+    if (!sourcePlane)
+        return nullptr;
+
     Owned<IStorageApiInfo> sourceAPIInfo = sourcePlane->getStorageApiInfo();
     if (!sourceAPIInfo)
+        return nullptr;
+
+    if (!distributedTarget)
         return nullptr;
 
     StringBuffer targetClusterName;
     distributedTarget->getClusterName(0, targetClusterName);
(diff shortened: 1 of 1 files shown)


jakesmith:
Not sure this is the right fix, but could be a short term workaround.

If the source or the target (set via setSource/setTarget) was a IFileDescriptor not a IDistributedFile, it implies it was not a published logical file, e.g. because it's a file on a LZ. ( it looks like it's also true via other routes, e.g. IDistributeFileSystem::replicate ).

I think the copy API can still be used in some of these cases, it would mean getAPICopyClient would have to determine the plane(s) involved via other means (e.g. extracting and setting as members during setSource/setTarget)

I'll defer to @shamser though.

dali/ft/filecopy.cpp
```diff
@@ -1522,6 +1522,9 @@ IAPICopyClient * FileSprayer::getAPICopyClient()
     if (needCalcCRC && !sources.item(0).hasCRC)
         return nullptr;
 
+    if (!distributedSource)
+        return nullptr;
+
```
ghalliday, line 1527: I think this change is good.  There should also be checks that sourcePlane and targetPlane are not null before they are derefenced.

ghalliday:
I also agree with Jake, that it should be possible to use the file copy for spray/despray, but a quick fix is needed.  I think this should probably go into an rc2 since it is a significant regression.

AttilaVamos:
I agree with both of you, this is a quick fix only and we need a better approach.

jakesmith (approved):
Could mege as is, but it is probably worth adding the sourcePlane and targetPlane checks Gavin pointed out in this PR too.

dali/ft/filecopy.cpp
AttilaVamos: Two more checks added

ghalliday:
@AttilaVamos I think you need both checks.  Please can you add them back and repush.

ghalliday (approved):

AttilaVamos:
@ghalliday squashed.

PR #16692, opened 2022-11-10
HPCC-28534 Add disk fetch index stats

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:


system/jhtree/jhtree.cpp
```diff
@@ -1113,13 +1117,14 @@ CDiskKeyIndex::CDiskKeyIndex(unsigned _iD, IFileIO *_io, const char *_name, bool
...
-CJHTreeNode *CDiskKeyIndex::loadNode(CJHTreeNode * optNode, offset_t pos)
+CJHTreeNode *CDiskKeyIndex::loadNode(cycle_t * fetchCycles, CJHTreeNode * optNode, offset_t pos)
 {
     nodesLoaded++;
     unsigned nodeSize = keyHdr->getNodeSize();
     MemoryAttr ma;
     char *nodeData = (char *) ma.allocate(nodeSize);
-    MTIME_SECTION(queryActiveTimer(), "JHTREE read node");
+
+    CCycleTimer fetchTimer(fetchCycles != nullptr);
```
ghalliday, line 1127: I'm not sure if I should keep this call to MTIME_SECTION.  I think the stats coming back are likely to be much more useful.  Thoughts?

mckellyln: I say keep the MTIME_SECTION(queryActiveTimer(), "JHTREE read node");

mckellyln: After more discussion I think removing some MTIME_SECTION() calls would be good, since they have a CritSec and a find().

mckellyln (approved):
Looks good.

system/jhtree/jhtree.cpp
```diff
@@ -62,10 +62,13 @@
 #include "rtlrecord.hpp"
 #include "rtldynfield.hpp"
 
+constexpr __uint64 defaultFetchThresholdNs = 50000; // Assume anything < 50us comes from the page cache, everything above probably went to disk
```
ghalliday, line 65: I think this is probably too high, and should probably change to 20ns

richardkchapman (approved):

ghalliday:
I've reduced the threshold to 20ns (my timings seem to be ~5ns), and also separately recorded the time when it is above the threshold.  Is that overkill?  if the page cache timing is low enough then dividing the leaf read by leaf fetches may be good enough, and avoid accumulating yet another set of stats.  On reflection I may roll that one back (but keep the reduced threshold).  Any comments/thoughts?

mckellyln (approved):
I think its fine.
If you wanted or were concerned about adding extra timing to happen always, we could disable the timing with a specific setting for the threshold (such as 0 or -1)

PR #16697, opened 2022-11-11
HPCC-28072 Add "DFS" option to Dali Admin

kunalaswani:
DaliAdmin implementation of DFSCheck, DFSExists, DFSLS

Signed-off-by: Kunal Aswani <Kunal.Aswani@lexisnexisrisk.com>

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:


esp/src/src-react/components/DFSExists.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+}) => {
+
+    const [columns, setColumns] = React.useState<IColumn[]>([]);
+    const [items, setItems] = React.useState<any[]>([]);
+    const [fileName, setFileName] = React.useState<string>("");
+
+    const onSubmit = React.useCallback(() => {
+        myDaliService.DFSExists({ FileName: fileName }).then(response => {
+            const data = csvParse(Result);
```
jeclrsg, line 28: I missed this in the previous PR, but that should be response.Result

jeclrsg (changes requested):
@kunalaswani There's one issue in the DFSExists that was causing the npm build to fail

esp/src/src-react/components/DFSExists.tsx
kunalaswani: corrected for response, good catch.

esp/src/src-react/components/DFSExists.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+}) => {
+
+    const [columns, setColumns] = React.useState<IColumn[]>([]);
+    const [items, setItems] = React.useState<any[]>([]);
+    const [fileName, setFileName] = React.useState<string>("");
+
+    const onSubmit = React.useCallback(() => {
+        myDaliService.DFSExists({ FileName: fileName }).then(response => {
+            const data = csvParse(response.Result);
```
jeclrsg, line 28: I didn't look at the API earlier, but the npm build step is failing now because `response` in this case is a boolean and not an object, and so doesn't have a `Result`. Also, since it just returns a true or false, this callback would be different than the other components - there won't be any data to set, and the HTML of the component would be different as well. I'm not sure how though.

jeclrsg (changes requested):
@kunalaswani Noticed the npm build check was failing. Seems like that DFSExists component will be different than the others because the DaliService.DFSCheck call doesn't return any CSV to parse, etc

esp/src/src-react/components/DFSCheck.tsx
```diff
@@ -0,0 +1,64 @@
...
+                    minWidth: 100
+                };
+            }));
+            setItems(data);
+        }).catch(err => logger.error(err));
+    }, [dfsCheck]);
+
+    return <HolyGrail
+        header={<span><TableGroup fields={{
+            "DFSCheck": { label: nlsHPCC.DFSCheck, type: "checkbox", value: true ? true : false },
```
jeclrsg, line 44: should be `... type: "checkbox", value: dfsCheck },`

esp/src/src-react/components/DFSLS.tsx
```diff
@@ -0,0 +1,86 @@
...
+                return {
+                    key: col,
+                    name: col,
+                    fieldName: col,
+                    minWidth: 100
+                };
+            }));
+            setItems(data);
+        }).catch(err => logger.error(err));
+    }, [name, pathAndNameOnly, includeSubFileInfo, recursively]);
```
jeclrsg, line 43: name here doesn't match the variable definition on line 25, should be Name?

esp/src/src-react/components/DFSLS.tsx
```diff
@@ -0,0 +1,86 @@
...
+            "PathAndNameOnly": { label: nlsHPCC.PathAndNameOnly, type: "checkbox", value: true ? true : false },
+            "IncludeSubFileInfo": { label: nlsHPCC.IncludeSubFileInfo, type: "checkbox", value: true ? true : false },
+            "Recursively": { label: nlsHPCC.Recursively, type: "checkbox", value: true ? true : false },
+
+        }} onChange={(id, value) => {
+            switch (id) {
+                case "name":
+                    setName(value);
+                    break;
+                case "pathAndNameOnly":
```
jeclrsg, line 57: These cases should match the ids in the TableGroup fields object above. So "PathAndNameOnly", etc.

esp/src/src-react/components/DFSLS.tsx
```diff
@@ -0,0 +1,86 @@
...
+                };
+            }));
+            setItems(data);
+        }).catch(err => logger.error(err));
+    }, [name, pathAndNameOnly, includeSubFileInfo, recursively]);
+
+    return <HolyGrail
+        header={<span><TableGroup fields={{
+            "Name": { label: nlsHPCC.Name, type: "string", value: Name },
+            "PathAndNameOnly": { label: nlsHPCC.PathAndNameOnly, type: "checkbox", value: true ? true : false },
```
jeclrsg, line 48: Same as in DFSCheck, these values should use the state variables defined above as the value, so `... type: "checkbox", value: pathAndNameOnly },` etc

jeclrsg (changes requested):
@kunal Things that were causing problems with those checkboxes

jeclrsg (changes requested):
@kunalaswani Github flagged some lint issues that cause the CI to fail, but otherwise looked good

kunalaswani:
@jeclrsg i believe the lint issue is resolved now

esp/src/src-react/components/DFSCheck.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { DefaultButton, DetailsList, DetailsListLayoutMode, IColumn } from "@fluentui/react";
+import { SizeMe } from "react-sizeme";
+import { csvParse } from "d3-dsv";
+import { DaliService } from "@hpcc-js/comms";
+import { scopedLogger } from "@hpcc-js/util";
+import { TableGroup } from "./forms/Groups";
+import nlsHPCC from "src/nlsHPCC";
+import { HolyGrail } from "../layouts/HolyGrail";
+
+
```
jeclrsg, line 11: @kunalaswani I don't recall if this was what the GH actions was complaining about last time, but you'll need to delete the extra blank lines. You can also check for this kind of stuff by executing `npm run lint`. I probably forget to do it before opening a PR almost half the time.

jeclrsg (changes requested):

esp/src/src-react/components/DFSCheck.tsx
kunalaswani: I'll keep that in mind, thank you, i thought i had removed those on the last update, sorry.

jeclrsg (approved):

ghalliday:
@GordonSmith I'll wait for your agreement (you are welcome to merge).  Only thing I spotted was that the new modules were called DFSExists rather than GetDFSExists - I'm not sure if you care.

GordonSmith:
@ghalliday DFSExists is more consistent with what users are already used to.

PR #16710, opened 2022-11-15
HPCC-28553 Check dropzone scope access in ESP FileSpray

wangkx:
1. Add getDropZoneScopePermissions() to CDistributedFileDirectory.
2. Add getDropZoneScopePermissions() to the ESP SMC lib.
3. Check dropzone scope access using the getDropZoneScopePermissions().

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dali/base/dadfs.cpp
+++ b/dali/base/dadfs.cpp
@@ -1136,6 +1136,7 @@ public:
     SecAccessFlags getFilePermissions(const char *lname,IUserDescriptor *user,unsigned auditflags);
     SecAccessFlags getNodePermissions(const IpAddress &ip,IUserDescriptor *user,unsigned auditflags);
     SecAccessFlags getFDescPermissions(IFileDescriptor *,IUserDescriptor *user,unsigned auditflags=0);
+    SecAccessFlags getDropZoneScopePermissions(const char *dropZoneName,const char *dropZonePath,IUserDescriptor *user,unsigned auditflags=0);
     void setDefaultUser(IUserDescriptor *user);
     IUserDescriptor* queryDefaultUser();
 
@@ -11774,6 +11775,15 @@ SecAccessFlags CDistributedFileDirectory::getFDescPermissions(IFileDescriptor *f
         }
     }
     return retPerms;
+}
+
+SecAccessFlags CDistributedFileDirectory::getDropZoneScopePermissions(const char *dropZoneName,const char *dropZonePath,IUserDescriptor *user,unsigned auditflags)
+{
+    CDfsLogicalFileName dlfn;
+    dlfn.setPlaneExternal(dropZoneName,dropZonePath);
+    StringBuffer scopes;
+    dlfn.getScopes(scopes);
+    return getScopePermissions(scopes,user,auditflags);
 }
 
 void CDistributedFileDirectory::setDefaultUser(IUserDescriptor *user)
--- a/dali/base/dadfs.hpp
+++ b/dali/base/dadfs.hpp
@@ -646,6 +646,7 @@ interface IDistributedFileDirectory: extends IInterface
     virtual IUserDescriptor* queryDefaultUser()=0;
     virtual SecAccessFlags getNodePermissions(const IpAddress &ip,IUserDescriptor *user,unsigned auditflags=0)=0;
     virtual SecAccessFlags getFDescPermissions(IFileDescriptor *,IUserDescriptor *user,unsigned auditflags=0)=0;
+    virtual SecAccessFlags getDropZoneScopePermissions(const char *dropZoneName,const char *dropZonePath,IUserDescriptor *user,unsigned auditflags=0)=0;
 
     virtual DistributedFileCompareResult fileCompare(const char *lfn1,const char *lfn2,DistributedFileCompareMode mode,StringBuffer &errstr,IUserDescriptor *user)=0;
     virtual bool filePhysicalVerify(const char *lfn1,IUserDescriptor *user,bool includecrc,StringBuffer &errstr)=0;
--- a/dali/base/dautils.cpp
+++ b/dali/base/dautils.cpp
@@ -46,18 +46,20 @@
 #define SDS_CONNECT_TIMEOUT  (1000*60*60*2)     // better than infinite
 #define MIN_REDIRECTION_LOAD_INTERVAL 1000
 
+static IPropertyTree *getPlaneHostGroup(IPropertyTree *plane)
+{
+    if (plane->hasProp("@hostGroup"))
+        return getHostGroup(plane->queryProp("@hostGroup"), true);
+    else if (plane->hasProp("hosts"))
+        return LINK(plane); // plane itself holds 'hosts'
+    return nullptr;
+}
 
 bool isHostInPlane(IPropertyTree *plane, const char *host, bool ipMatch)
 {
-    Owned<IPropertyTree> planeGroup;
-    if (plane->hasProp("@hostGroup"))
-        planeGroup.setown(getHostGroup(plane->queryProp("@hostGroup"), true));
-    else
-    {
-        if (!plane->hasProp("hosts"))
-            return false;
-        planeGroup.set(plane); // plane itself holds 'hosts'
-    }
+    Owned<IPropertyTree> planeGroup = getPlaneHostGroup(plane);
+    if (!planeGroup)
+        return false;
     Owned<IPropertyTreeIterator> hostsIter = planeGroup->getElements("hosts");
     SocketEndpoint hostEp;
     if (ipMatch)
@@ -79,12 +81,8 @@ bool isHostInPlane(IPropertyTree *plane, const char *host, bool ipMatch)
 
 bool getPlaneHost(StringBuffer &host, IPropertyTree *plane, unsigned which)
 {
-    Owned<IPropertyTree> hostGroup;
-    if (plane->hasProp("@hostGroup"))
-        hostGroup.setown(getHostGroup(plane->queryProp("@hostGroup"), true));
-    else if (plane->hasProp("hosts"))
-        hostGroup.set(plane); // the plane holds the "hosts"
-    else
+    Owned<IPropertyTree> hostGroup = getPlaneHostGroup(plane);
+    if (!hostGroup)
         return false;
 
     if (which >= hostGroup->getCount("hosts"))
@@ -92,6 +90,17 @@ bool getPlaneHost(StringBuffer &host, IPropertyTree *plane, unsigned which)
     VStringBuffer xpath("hosts[%u]", which+1); // which is 0 based
     host.append(hostGroup->queryProp(xpath));
     return true;
+}
+
+void getPlaneHosts(StringArray &hosts, IPropertyTree *plane)
+{
+    Owned<IPropertyTree> hostGroup = getPlaneHostGroup(plane);
+    if (hostGroup)
+    {
+        Owned<IPropertyTreeIterator> hostsIter = hostGroup->getElements("hosts");
+        ForEach (*hostsIter)
+            hosts.append(hostsIter->query().queryProp(nullptr));
+    }
 }
 
 constexpr const char * lz_plane_path = "storage/planes[@category='lz']";
--- a/dali/base/dautils.hpp
+++ b/dali/base/dautils.hpp
@@ -545,6 +545,7 @@ extern da_decl IPropertyTree * getDropZonePlane(const char * name);
 extern da_decl IPropertyTree * findDropZonePlane(const char * path, const char * host, bool ipMatch);
 extern da_decl bool isHostInPlane(IPropertyTree *plane, const char *host, bool ipMatch);
 extern da_decl bool getPlaneHost(StringBuffer &host, IPropertyTree *plane, unsigned which);
+extern da_decl void getPlaneHosts(StringArray &hosts, IPropertyTree *plane);
 extern da_decl void setPageCacheTimeoutMilliSeconds(unsigned timeoutSeconds);
 extern da_decl void setMaxPageCacheItems(unsigned _maxPageCacheItems);
 extern da_decl IRemoteConnection* connectXPathOrFile(const char* path, bool safe, StringBuffer& xpath);
--- a/esp/services/ws_fs/ws_fsBinding.cpp
+++ b/esp/services/ws_fs/ws_fsBinding.cpp
@@ -393,12 +393,17 @@ int CFileSpraySoapBindingEx::downloadFile(IEspContext &context, CHttpRequest* re
         if (!context.validateFeatureAccess(FILE_SPRAY_URL, SecAccess_Full, false))
             throw MakeStringException(ECLWATCH_FILE_SPRAY_ACCESS_DENIED, "Failed to download file. Permission denied.");
 
-        StringBuffer netAddressStr, osStr, pathStr, nameStr;
+        StringBuffer netAddressStr, osStr, pathStr, nameStr, dropZoneName;
         request->getParameter("NetAddress", netAddressStr);
         request->getParameter("OS", osStr);
         request->getParameter("Path", pathStr);
         request->getParameter("Name", nameStr);
+        request->getParameter("DropZoneName", dropZoneName);
 
+        SecAccessFlags permission = getDropZoneScopePermissions(context, dropZoneName, pathStr, netAddressStr);
+        if (permission < SecAccess_Read)
+            throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "Access DropZone Scope %s %s %s not allowed for user %s (permission:%s). Read Access Required.",
+                dropZoneName.str(), netAddressStr.str(), pathStr.str(), context.queryUserId(), getSecAccessFlagName(permission));
 #if 0
         StringArray files;
         IProperties* params = request->queryParameters();
@@ -479,11 +484,16 @@ int CFileSpraySoapBindingEx::downloadFile(IEspContext &context, CHttpRequest* re
 
 int CFileSpraySoapBindingEx::onStartUpload(IEspContext& ctx, CHttpRequest* request, CHttpResponse* response, const char* serv, const char* method)
 {
-    StringBuffer netAddress, path;
+    StringBuffer netAddress, path, dropZoneName;
     request->getParameter("NetAddress", netAddress);
     request->getParameter("Path", path);
+    request->getParameter("DropZoneName", dropZoneName);
     if (!validateDropZonePath(nullptr, netAddress, path)) //The path should be the absolute path for the dropzone.
         throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "Invalid Landing Zone path %s", path.str());
+    SecAccessFlags permission = getDropZoneScopePermissions(ctx, dropZoneName, path, netAddress);
+    if (permission < SecAccess_Full)
+        throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "Access DropZone Scope %s %s %s not allowed for user %s (permission:%s). Full Access Required.",
+            dropZoneName.str(), netAddress.str(), path.str(), ctx.queryUserId(), getSecAccessFlagName(permission));
 
     return EspHttpBinding::onStartUpload(ctx, request, response, serv, method);
 }
--- a/esp/services/ws_fs/ws_fsService.cpp
+++ b/esp/services/ws_fs/ws_fsService.cpp
@@ -2338,100 +2338,6 @@ void CFileSprayEx::getDropZoneInfoByDestPlane(double clientVersion, const char* 
     getDropZoneHost(destPlane, dropZone, hostip);
 }
 
-void CFileSprayEx::getDropZoneInfoByIP(double clientVersion, const char* ip, const char* destFileIn, StringBuffer& destFileOut, StringBuffer& umask)
-{
-#ifndef _CONTAINERIZED
-    if (destFileIn && *destFileIn)
-        destFileOut.set(destFileIn);
-
-    if (!ip || !*ip)
-        throw MakeStringExceptionDirect(ECLWATCH_INVALID_IP, "Network address must be specified for a drop zone!");
-
-    Owned<IEnvironmentFactory> factory = getEnvironmentFactory(true);
-    Owned<IConstEnvironment> constEnv = factory->openEnvironment();
-
-    StringBuffer destFile;
-    if (isAbsolutePath(destFileIn))
-    {
-        destFile.set(destFileIn);
-        Owned<IConstDropZoneInfo> dropZone = constEnv->getDropZoneByAddressPath(ip, destFile.str());
-        if (!dropZone)
-        {
-            if (constEnv->isDropZoneRestrictionEnabled())
-                throw MakeStringException(ECLWATCH_DROP_ZONE_NOT_FOUND, "No drop zone configured for '%s' and '%s'. Check your system drop zone configuration.", ip, destFile.str());
-            else
-            {
-                LOG(MCdebugInfo, unknownJob, "No drop zone configured for '%s' and '%s'. Check your system drop zone configuration.", ip, destFile.str());
-                return;
-            }
-        }
-
-
-        SCMStringBuffer directory, maskBuf;
-        dropZone->getDirectory(directory);
-        destFileOut.set(destFile.str());
-        dropZone->getUMask(maskBuf);
-        if (maskBuf.length())
-            umask.set(maskBuf.str());
-
-        return;
-    }
-
-    Owned<IConstDropZoneInfoIterator> dropZoneItr = constEnv->getDropZoneIteratorByAddress(ip);
-    if (dropZoneItr->count() < 1)
-    {
-        if (constEnv->isDropZoneRestrictionEnabled())
-            throw MakeStringException(ECLWATCH_DROP_ZONE_NOT_FOUND, "Drop zone not found for network address '%s'. Check your system drop zone configuration.", ip);
-        else
-        {
-            LOG(MCdebugInfo, unknownJob, "Drop zone not found for network address '%s'. Check your system drop zone configuration.", ip);
-            return;
-        }
-    }
-
-    bool dzFound = false;
-    ForEach(*dropZoneItr)
-    {
-        IConstDropZoneInfo& dropZoneInfo = dropZoneItr->query();
-
-        SCMStringBuffer dropZoneDirectory, dropZoneUMask;
-        dropZoneInfo.getDirectory(dropZoneDirectory);
-        dropZoneInfo.getUMask(dropZoneUMask);
-        if (!dropZoneDirectory.length())
-            continue;
-
-        if (!dzFound)
-        {
-            dzFound = true;
-            destFileOut.set(dropZoneDirectory.str());
-            addPathSepChar(destFileOut);
-            destFileOut.append(destFileIn);
-            if (dropZoneUMask.length())
-                umask.set(dropZoneUMask.str());
-        }
-        else
-        {
-            if (constEnv->isDropZoneRestrictionEnabled())
-                throw MakeStringException(ECLWATCH_INVALID_INPUT, "> 1 drop zones found for network address '%s'.", ip);
-            else
-            {
-                LOG(MCdebugInfo, unknownJob, "> 1 drop zones found for network address '%s'.", ip);
-                return;
-            }
-        }
-    }
-    if (!dzFound)
-    {
-        if (constEnv->isDropZoneRestrictionEnabled())
-            throw MakeStringException(ECLWATCH_DROP_ZONE_NOT_FOUND, "No valid drop zone found for network address '%s'. Check your system drop zone configuration.", ip);
-        else
-            LOG(MCdebugInfo, unknownJob, "No valid drop zone found for network address '%s'. Check your system drop zone configuration.", ip);
-    }
-#else
-    throw makeStringException(-1, "Internal error: CFileSprayEx::getDropZoneInfoByIP should not be called in containerized environment");
-#endif
-}
-
 static StringBuffer & expandLogicalAsPhysical(StringBuffer & target, const char * name, const char * separator)
 {
     const char * cur = name;
@@ -2472,30 +2378,17 @@ bool CFileSprayEx::onDespray(IEspContext &context, IEspDespray &req, IEspDespray
         MemoryBuffer& dstxml = (MemoryBuffer&)req.getDstxml();
         if(dstxml.length() == 0)
         {
-#ifdef _CONTAINERIZED
             if (isEmptyString(destPlane))
                 destPlane = req.getDestGroup();  // allow eclwatch to continue providing storage plane as 'destgroup' field
             if (isEmptyString(destPlane))
             {
                 if (destip.isEmpty())
-                    throw MakeStringException(ECLWATCH_INVALID_INPUT, "Neither destination storage plane or destination IP specified.");
-                Owned<IPropertyTreeIterator> planesIter = getDropZonePlanesIterator();
-                ForEach(*planesIter)
-                {
-                    IPropertyTree &lzPlane = planesIter->query();
-                    if (isHostInPlane(&lzPlane, destip, true))
-                    {
-                        destPlane = lzPlane.queryProp("@name");
-                        break;
-                    }
-                }
-                if (isEmptyString(destPlane))
-                    throw makeStringException(ECLWATCH_INVALID_INPUT, "Destination IP does not match a hosts based storage plane.");
+                    throw makeStringException(ECLWATCH_INVALID_INPUT, "Neither destination storage plane or destination IP specified.");
+                Owned<IPropertyTree> plane = findDropZonePlane(destPath, destip, true);
+                if (!plane)
+                    throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "DropZone Plane not found for host %s path %s.", destip.str(), destPath.str());
+                destPlane = plane->queryProp("@name");
             }
-#else
-            if (isEmptyString(destPlane) && destip.isEmpty())
-                throw MakeStringException(ECLWATCH_INVALID_INPUT, "Destination network IP/storage plane not specified.");
-#endif
 
             //If the destination filename is not provided, calculate a relative filename from the logical filename
             if(!destfile || !*destfile)
@@ -2526,10 +2419,7 @@ bool CFileSprayEx::onDespray(IEspContext &context, IEspDespray &req, IEspDespray
         if(dstxml.length() == 0)
         {
             StringBuffer destfileWithPath, umask;
-            if (!isEmptyString(destPlane))
-                getDropZoneInfoByDestPlane(version, destPlane, destfile, destfileWithPath, umask, destip);
-            else
-                getDropZoneInfoByIP(version, destip, destfile, destfileWithPath, umask);
+            getDropZoneInfoByDestPlane(version, destPlane, destfile, destfileWithPath, umask, destip);
 
             RemoteFilename rfn;
             SocketEndpoint ep(destip.str());
@@ -2922,9 +2812,10 @@ bool CFileSprayEx::onFileList(IEspContext &context, IEspFileListRequest &req, IE
             throw MakeStringException(ECLWATCH_INVALID_INPUT, "Path not specified.");
 
         double version = context.getClientVersion();
+        const char* dropZoneName = req.getDropZoneName();
         const char* netaddr = req.getNetaddr();
-        if (!netaddr || !*netaddr)
-            throw MakeStringException(ECLWATCH_INVALID_INPUT, "Network address not specified.");
+        if (isEmptyString(dropZoneName) && isEmptyString(netaddr))
+            throw makeStringException(ECLWATCH_INVALID_INPUT, "DropZoneName or Netaddr must be specified.");
         const char* fileNameMask = req.getMask();
         bool directoryOnly = req.getDirectoryOnly();
         PROGLOG("FileList:  Netaddr %s, Path %s", netaddr, path);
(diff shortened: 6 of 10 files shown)


dali/base/dadfs.hpp
```diff
@@ -646,6 +646,7 @@ interface IDistributedFileDirectory: extends IInterface
     virtual IUserDescriptor* queryDefaultUser()=0;
     virtual SecAccessFlags getNodePermissions(const IpAddress &ip,IUserDescriptor *user,unsigned auditflags=0)=0;
     virtual SecAccessFlags getFDescPermissions(IFileDescriptor *,IUserDescriptor *user,unsigned auditflags=0)=0;
+    virtual SecAccessFlags getDropZoneScopePermissions(const char *dropZoneNameOrIP,const char *dropZonePath,bool isPlane,IUserDescriptor *user,unsigned auditflags=0)=0;
```
RussWhitehead, line 649: Inconsistency in naming of isPlane in interface, isExternalPlane in implementation

esp/services/ws_fs/ws_fsBinding.cpp
```diff
@@ -393,12 +393,15 @@ int CFileSpraySoapBindingEx::downloadFile(IEspContext &context, CHttpRequest* re
...
-        StringBuffer netAddressStr, osStr, pathStr, nameStr;
+        StringBuffer netAddressStr, osStr, pathStr, nameStr, dropZoneName;
         request->getParameter("NetAddress", netAddressStr);
         request->getParameter("OS", osStr);
         request->getParameter("Path", pathStr);
         request->getParameter("Name", nameStr);
+        request->getParameter("DropZoneName", dropZoneName);
 
+        if (getDropZoneScopePermissions(context, dropZoneName, netAddressStr, pathStr) < SecAccess_Read)
+            throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "Access %s:%s not allowed.", netAddressStr.str(), pathStr.str());
```
RussWhitehead, line 404: To help user's debug their permission problems without having to get development involved, can we log more information about the denial?  Such as the LDAP feature name we are checking, or the required access ("Read Access Required") ?

RussWhitehead: I'm not sure SecAccessFlags are meant to be treated as a scalar. Perhaps consider only testing the SecAccess_Read bit, such as 

if (getDropZoneScopePermissions(context, dropZoneName, netAddressStr, pathStr) & SecAccess_Read)

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -2972,6 +2972,11 @@ bool CFileSprayEx::onFileList(IEspContext &context, IEspFileListRequest &req, IE
         if (!validateDropZonePath(nullptr, netaddr, sPath) && !validateConfigurationDirectory(nullptr, "log", nullptr, nullptr, sPath)) //The path should be the absolute path for the dropzone or log file.
             throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "Invalid file path %s", sPath.str());
 
+        const char* dropZoneName = req.getDropZoneName();
+        if (getDropZoneScopePermissions(context, dropZoneName, netaddr, sPath) < SecAccess_Read)
```
RussWhitehead, line 2976: I believe this only requires SecAccess_Access permission (get file attributes)?

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -3103,8 +3105,11 @@ void CFileSprayEx::addDropZoneFile(IEspContext& context, IDirectoryIterator* di,
     files.append(*aFile.getLink());
 }
 
-void CFileSprayEx::searchDropZoneFiles(IEspContext& context, const char* server, const char* dir, const char* nameFilter, IArrayOf<IEspPhysicalFileStruct>& files, unsigned& filesFound)
+void CFileSprayEx::searchDropZoneFiles(IEspContext& context, const char* dropZoneName, const char* server, const char* dir, const char* nameFilter, IArrayOf<IEspPhysicalFileStruct>& files, unsigned& filesFound)
 {
+    if (getDropZoneScopePermissions(context, dropZoneName, server, dir) < SecAccess_Read)
```
RussWhitehead, line 3110: I believe this only requires SecAccess_Access ?

esp/smc/SMCLib/TpContainer.cpp
```diff
@@ -802,3 +802,13 @@ StringArray & getRoxieDirectAccessPlanes(StringArray & planes, StringBuffer &def
 
     return planes;
 }
+
+extern TPWRAPPER_API SecAccessFlags getDropZoneScopePermissions(IEspContext & context, const char * dropZoneName, const char * dropZoneServer, const char * dropZonePath)
+{
+    if (isEmptyString(dropZoneName) || isEmptyString(dropZonePath))
```
RussWhitehead, line 808: Inconsistency between the condition test || and the error message ("Both cannot be empty").  I think you need && instead

esp/smc/SMCLib/TpWrapper.cpp
```diff
@@ -2135,3 +2135,13 @@ StringArray & getRoxieDirectAccessPlanes(StringArray & planes, StringBuffer &def
         planes.append(defaultPlane);
     return planes;
 }
+
+extern TPWRAPPER_API SecAccessFlags getDropZoneScopePermissions(IEspContext & context, const char * dropZoneName, const char * dropZoneServer, const char * dropZonePath)
+{
+    if (isEmptyString(dropZoneServer) || isEmptyString(dropZonePath))
```
RussWhitehead, line 2141: Inconsistency between the condition test || and the error message ("Both cannot be empty"). I think you need && instead

RussWhitehead (changes requested):

dali/base/dadfs.hpp
wangkx: It may be more accurate using the isPlane.

esp/services/ws_fs/ws_fsService.cpp
wangkx: The logic here is the same as listing logical files (SecAccess_Read). The SecAccess_Read is also checked in line 2941.

esp/services/ws_fs/ws_fsService.cpp
wangkx: Same as the line 2976

esp/services/ws_fs/ws_fsBinding.cpp
wangkx: > To help user's debug their permission problems without having to get development involved, can we log more information about the denial? Such as the LDAP feature name we are checking, or the required access ("Read Access Required") ?

I will add more information to the error message.

> I'm not sure SecAccessFlags are meant to be treated as a scalar. Perhaps consider only testing the SecAccess_Read bit

We treat the SecAccessFlags as a scalar in many other places: checkWuScopeSecAccess(), SecHandler::authorizeSecReqFeatures(), etc.

esp/smc/SMCLib/TpContainer.cpp
wangkx: The exception will be thrown if one or both of them are empty.

esp/smc/SMCLib/TpWrapper.cpp
wangkx: The exception will be thrown if one or both of them are empty.

RussWhitehead (changes requested):

dali/base/dautils.cpp
```diff
@@ -65,6 +65,29 @@ IPropertyTree * getDropZonePlane(const char * name)
...
+        if (isPathSepChar(*s))
+        {
+            s++;
+            if (*s)//Skip Trailing Path Sep Char
+                scope.append("::");
+        }
+        else
+        {
+            if ((*s=='^') || isupper(*s))
+                scope.append('^'); //Do we need to convert the upper cases here?
```
jakesmith, line 83: not sure what this comment means? you do need to convert, but you are on next line

dali/base/dautils.hpp
```diff
@@ -541,6 +541,7 @@ extern da_decl ILockInfoCollection *deserializeLockInfoCollection(MemoryBuffer &
 
 extern da_decl IPropertyTreeIterator * getDropZonePlanesIterator(const char * name=nullptr);
 extern da_decl IPropertyTree * getDropZonePlane(const char * name);
+extern da_decl StringBuffer & makeDropZoneScope(StringBuffer & scope, bool isExternalPlane, const char * dropZone, const char * path);
```
jakesmith, line 544: agree with your comment elsewhere, it would more consistent to use flip this and rename it to isPlane

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -2972,6 +2972,11 @@ bool CFileSprayEx::onFileList(IEspContext &context, IEspFileListRequest &req, IE
         if (!validateDropZonePath(nullptr, netaddr, sPath) && !validateConfigurationDirectory(nullptr, "log", nullptr, nullptr, sPath)) //The path should be the absolute path for the dropzone or log file.
             throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "Invalid file path %s", sPath.str());
 
+        const char* dropZoneName = req.getDropZoneName();
+        if (getDropZoneScopePermissions(context, dropZoneName, netaddr, sPath) < SecAccess_Read)
+            throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "Access DropZone Scope %s:%s not allowed. Read Access Required.", netaddr, sPath.str());
+
+
```
jakesmith, line 2979: trivial: 1 too many new newlines.

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -3121,11 +3126,33 @@ void CFileSprayEx::searchDropZoneFiles(IEspContext& context, const char* server,
...
+        const char* ptr = fname.str();
+        const char* pName = strrchr(ptr, pathSep);
+        if (pName)
+        {
+            sPath.append(pName - ptr, ptr);
+            fname.set(pName + 1);
+
+            StringBuffer fullPath(dir);
+            addPathSepChar(fullPath).append(sPath);
+            if (getDropZoneScopePermissions(context, dropZoneName, server, fullPath) < SecAccess_Read)
```
jakesmith, line 3147: I suspect I'm misreading.. but line 3133 has already checked scope perms if it's a directory,

this (lines 3137 to 3149) seems to be checking if di->getName is a file or directory within a subdir (by seeing if it has a path sep. in it), then building up the path up to the parent of that file, and checking that dir's perms.

But if it's a subdir (not a file within a subdir) - then wouldn't 3133 have already checked it?

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -3504,6 +3542,9 @@ bool CFileSprayEx::onDeleteDropZoneFiles(IEspContext &context, IEspDeleteDropZon
         if (!checkDropZoneIPAndPath(version, dzName, netAddress, path.str()))
             throw MakeStringException(ECLWATCH_DROP_ZONE_NOT_FOUND, "Dropzone is not found in the environment settings.");
 
+        if (getDropZoneScopePermissions(context, dzName, netAddress, path) < SecAccess_Full)
```
jakesmith, line 3545: minor: this should prob. go after the  if(f->isDirectory()!=fileBool::foundYes) check (on line 3555)

jakesmith (commented):
@wangkx - please see comments

dali/base/dautils.cpp
wangkx: The code does the conversion. I just want to double check (in case I should not convert). I will remove the comment.

esp/services/ws_fs/ws_fsService.cpp
wangkx: Will remove the extra newline.

esp/services/ws_fs/ws_fsService.cpp
wangkx: The directoryFiles() returns the files under the dir. The line 3133 only checks scope perms if the di is a folder. If the di is not a folder (ex. /dir/subfolder1/file1), the code here checks the scope access for the subfolders (ex. /dir/subfolder1).

esp/services/ws_fs/ws_fsService.cpp
wangkx: I thought about that. Should we expose the 'cannot resolve network IP' error or the 'not a directory' error if the user does not have the scope access?

esp/services/ws_fs/ws_fsService.cpp
wangkx: I should add the code to do this only if di is not a folder.

wangkx:
@jakesmith I made the changes based on your comments. Please let me know if I miss anything.

dali/base/dadfs.hpp
jakesmith: I agree

dali/base/dautils.hpp
jakesmith: @wangkx - missed this one? i.e. I think worth changing to isPlane so matches other places (e.g. getDropZoneScopePermissions)

esp/services/ws_fs/ws_fsService.cpp
jakesmith: thinking about it, I think prob. more correct to check perms. 1st, before validating if a dir, or if host resolvable.

esp/smc/SMCLib/TpContainer.cpp
```diff
@@ -802,3 +802,13 @@ StringArray & getRoxieDirectAccessPlanes(StringArray & planes, StringBuffer &def
 
     return planes;
 }
+
+extern TPWRAPPER_API SecAccessFlags getDropZoneScopePermissions(IEspContext & context, const char * dropZoneName, const char * dropZoneServer, const char * dropZonePath)
```
jakesmith, line 806: it would be clearer to remove the out of context dropZoneServer parameter here,
and in the non-containerized version the dropZoneName, and unify into a single parameter.

esp/smc/SMCLib/TpWrapper.cpp
```diff
@@ -2135,3 +2135,13 @@ StringArray & getRoxieDirectAccessPlanes(StringArray & planes, StringBuffer &def
         planes.append(defaultPlane);
     return planes;
 }
+
+extern TPWRAPPER_API SecAccessFlags getDropZoneScopePermissions(IEspContext & context, const char * dropZoneName, const char * dropZoneServer, const char * dropZonePath)
```
jakesmith, line 2139: see commet in TpContainer.cpp re. removing unusued param

esp/services/ws_fs/ws_fsBinding.cpp
```diff
@@ -393,12 +393,15 @@ int CFileSpraySoapBindingEx::downloadFile(IEspContext &context, CHttpRequest* re
...
 
-        StringBuffer netAddressStr, osStr, pathStr, nameStr;
+        StringBuffer netAddressStr, osStr, pathStr, nameStr, dropZoneName;
         request->getParameter("NetAddress", netAddressStr);
         request->getParameter("OS", osStr);
         request->getParameter("Path", pathStr);
         request->getParameter("Name", nameStr);
+        request->getParameter("DropZoneName", dropZoneName);
 
+        if (getDropZoneScopePermissions(context, dropZoneName, netAddressStr, pathStr) < SecAccess_Read)
```
jakesmith, line 403: see comments elsewhere. Only 1 of either dropZoneName or netAddressStr is valid depending on whehter containerized or not, it would be clearer for the impls. that use 1 to have 1 param, and for the caller to check and use the correct one.

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -3180,7 +3210,7 @@ bool CFileSprayEx::onDropZoneFileSearch(IEspContext &context, IEspDropZoneFileSe
                 IConstTpMachine& tpMachine = tpMachines.item(ii);
                 if (isEmptyString(dropZoneServerReq) || matchNetAddressRequest(dropZoneServerReq, isIPAddressReq, tpMachine))
                 {
-                    searchDropZoneFiles(context, tpMachine.getNetaddress(), dropZone.getPath(), nameFilter, files, filesFound);
+                    searchDropZoneFiles(context, dropZoneName, tpMachine.getNetaddress(), dropZone.getPath(), nameFilter, files, filesFound);
```
jakesmith, line 3213: similar to comments elsewhere, do both dropZoneName and netAddress being set make sense?
If not - it would make things clearer - if only 1 was passed based on mode.

jakesmith (commented):
@wangkx - please see comments.

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -3121,11 +3125,37 @@ void CFileSprayEx::searchDropZoneFiles(IEspContext& context, const char* server,
...
+
+        StringBuffer sPath;
+        const char* ptr = fname.str();
+        const char* pName = strrchr(ptr, pathSep);
+        if (pName)
+        {
+            sPath.append(pName - ptr, ptr);
+            fname.set(pName + 1);
+
+            if (!isDir)
```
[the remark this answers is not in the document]
wangkx: The getDropZoneFiles and onFileList only retrieve one layer of the folders/files.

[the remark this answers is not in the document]
wangkx: In the latest revision, I added my own recursion into subdirs for searchDropZoneFiles.

esp/services/ws_fs/ws_fsService.hpp
```diff
@@ -152,8 +152,8 @@ class CFileSprayEx : public CFileSpray
     void getDropZoneInfoByIP(double clientVersion, const char* destIP, const char* destFile, StringBuffer& path, StringBuffer& mask);
     void getDropZoneInfoByDestPlane(double clientVersion, const char* destGroup, const char* destFileIn, StringBuffer& destFileOut, StringBuffer& umask, StringBuffer & hostip);
     bool checkDropZoneIPAndPath(double clientVersion, const char* dropZone, const char* netAddr, const char* path);
-    void addDropZoneFile(IEspContext& context, IDirectoryIterator* di, const char* name, const char pathSep, const char* server, IArrayOf<IEspPhysicalFileStruct>&files);
-    void searchDropZoneFiles(IEspContext& context, const char* server, const char* dir, const char* nameFilter, IArrayOf<IEspPhysicalFileStruct>& files, unsigned& filesFound);
+    void addDropZoneFile(IEspContext& context, IDirectoryIterator* di, const char* name, const char* path, const char* server, IArrayOf<IEspPhysicalFileStruct>&files);
+    void searchDropZoneFiles(IEspContext& context, const char* dropZone, const char* server, const char* dir, const char* nameFilter, IArrayOf<IEspPhysicalFileStruct>& files, unsigned& filesFound);
```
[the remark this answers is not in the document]
wangkx: I may not understand correctly. Is this related to: https://github.com/hpcc-systems/HPCC-Platform/pull/16710#discussion_r1029416424 ?

esp/smc/SMCLib/TpContainer.cpp
wangkx: Is this related to: https://github.com/hpcc-systems/HPCC-Platform/pull/16710#discussion_r1029416424 ?

esp/smc/SMCLib/TpWrapper.cpp
wangkx: Is this related to: https://github.com/hpcc-systems/HPCC-Platform/pull/16710#discussion_r1029416424 ?

esp/services/ws_fs/ws_fsBinding.cpp
wangkx: Is this related to: https://github.com/hpcc-systems/HPCC-Platform/pull/16710#discussion_r1029416424 ?

esp/smc/SMCLib/TpContainer.cpp
jakesmith: Yes

esp/smc/SMCLib/TpWrapper.cpp
jakesmith: Yes

esp/services/ws_fs/ws_fsBinding.cpp
jakesmith: Yes

esp/services/ws_fs/ws_fsService.cpp
jakesmith: yes, but I would prefer code to be kept as similar as possible everywhere, and it is generally better unless time critical to use isContainerized() , so both code paths get compiled and tested.
Code clarity is the most import aspect here/reducing duplicated code etc.

I'd code something like:
```
const char *dropZoneId = isContainerized() ? dropZoneName : tpMachine.getNetaddress();
searchDropZoneFiles(context, dropZoneId, dropZone.getPath(), nameFilter, files, filesFound);
```

jakesmith (commented):
@wangkx - looks close

esp/services/ws_fs/ws_fsService.cpp
wangkx: Thanks for the clarification. I am changing the code related to the getDropZoneScopePermissions(). For searchDropZoneFiles(), I may still need to pass both dropZoneName and netAddress. The dropZoneName is for scope check. The netAddress is for retrieving file information using the SocketEndpoint.

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -3117,46 +3118,32 @@ void CFileSprayEx::searchDropZoneFiles(IEspContext& context, const char* dropZon
         throw MakeStringException(ECLWATCH_INVALID_DIRECTORY, "%s is not a directory.", dir);
 
     const char pathSep = getPathSepChar(dir);
-    Owned<IDirectoryIterator> di = f->directoryFiles(nameFilter, true, true);
+    Owned<IDirectoryIterator> di = f->directoryFiles(nameFilter, false, true);
     ForEach(*di)
     {
         StringBuffer fname;
         di->getName(fname);
         if (!fname.length())
```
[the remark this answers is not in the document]
wangkx: This is a check just in case. I can remove it.

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -3113,20 +3118,32 @@ void CFileSprayEx::searchDropZoneFiles(IEspContext& context, const char* server,
         throw MakeStringException(ECLWATCH_INVALID_DIRECTORY, "%s is not a directory.", dir);
 
     const char pathSep = getPathSepChar(dir);
```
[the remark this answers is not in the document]
wangkx: Will do

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -3567,6 +3597,93 @@ bool CFileSprayEx::onDeleteDropZoneFiles(IEspContext &context, IEspDeleteDropZon
...
+        accessReqStr.set("Write Access");
+        break;
+    case SecAccess_Read:
+        accessReqStr.set("Read Access");
+        break;
+    case SecAccess_Access:
+        accessReqStr.set("Access");
+        break;
+    default:
+        accessReqStr.set("Full Access");
```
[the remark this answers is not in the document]
wangkx: I may move it to system\security\shared\seclib.hpp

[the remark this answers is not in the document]
wangkx: In fact, I can use the getSecAccessFlagName() inside the system\security\shared\seclib.hpp.

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -3103,8 +3104,12 @@ void CFileSprayEx::addDropZoneFile(IEspContext& context, IDirectoryIterator* di,
...
-void CFileSprayEx::searchDropZoneFiles(IEspContext& context, const char* server, const char* dir, const char* nameFilter, IArrayOf<IEspPhysicalFileStruct>& files, unsigned& filesFound)
+bool CFileSprayEx::searchDropZoneFiles(IEspContext& context, const char* dropZoneName, const char* server,
+    const char* dir, const char* relDir, const char* nameFilter, IArrayOf<IEspPhysicalFileStruct>& files, unsigned& filesFound)
 {
+    if (getDropZoneScopePermissions(context, dropZoneName, server, dir) < SecAccess_Read)
+        return false;
+
     RemoteFilename rfn;
     SocketEndpoint ep(server);
     rfn.setPath(ep, dir);
```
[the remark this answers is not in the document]
wangkx: In bare metal environment, one dropzone may have multiple servers. Can the same happen in a K8s environment?

esp/services/ws_fs/ws_fsService.cpp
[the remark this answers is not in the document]
jakesmith: as far as I know it's not possible (if it were empty, it's an error elsewhere).
Checking here, makes it look like it is legitimate for it to be empty in some scenarios, better to remove for clarity.

esp/services/ws_fs/ws_fsService.cpp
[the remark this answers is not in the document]
jakesmith: Good question, I'm not sure it makes sense.

In bare metal I think it's just a way of grouping different dropzones with the same configuration.
In cloud these 'out of cloud' dropzones would I think all be be configured separately.

As I say, let's leave handling these for next JIRA/PR (worth opening a JIRA now though).

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -3611,6 +3597,93 @@ bool CFileSprayEx::onDeleteDropZoneFiles(IEspContext &context, IEspDeleteDropZon
     return true;
 }
 
+void CFileSprayEx::checkDropZoneFileScopeAccess(IEspContext &context, const char *dropZoneName, const char *netAddress,
```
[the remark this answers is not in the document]
jakesmith: ok

esp/services/ws_fs/ws_fsService.cpp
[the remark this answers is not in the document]
wangkx: https://track.hpccsystems.com/browse/HPCC-28634

esp/services/ws_fs/ws_fsService.cpp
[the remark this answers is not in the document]
wangkx: The if statement has been removed.

esp/services/ws_fs/ws_fsBinding.cpp
```diff
@@ -393,12 +393,16 @@ int CFileSpraySoapBindingEx::downloadFile(IEspContext &context, CHttpRequest* re
...
 
-        StringBuffer netAddressStr, osStr, pathStr, nameStr;
+        StringBuffer netAddressStr, osStr, pathStr, nameStr, dropZoneName;
         request->getParameter("NetAddress", netAddressStr);
         request->getParameter("OS", osStr);
         request->getParameter("Path", pathStr);
         request->getParameter("Name", nameStr);
+        request->getParameter("DropZoneName", dropZoneName);
 
+        const char *dropZoneId = isContainerized() ? dropZoneName.str() : netAddressStr.str();
```
jakesmith, line 403: on reflection, the aim is to support planes in bare-metal too, so I think it would be better to not make this conditional on isContainerized().

And, in bare-metal a named dropzone will have been automatically created from the environment.

I think this and similar code should use the dropZoneName in the request, and if a legacy IP/hostname is supplied, convert it (look it up) to it's drop zone name.
There is similar code that does this for despray here : https://github.com/hpcc-systems/HPCC-Platform/blob/2de4ad0b7f8fec292513bf067b9edb35e400e306/esp/services/ws_fs/ws_fsService.cpp#L2499

We should move that code to a helper function (dautils) e.g. to findHostedPlane(const char *host) 
Then code here and in other similar places can look at the supplied dropzonename and/or host/netaddr and lookup the dropzone name.

Then all the calls to getDropZoneScopePerm* can unify into only accepting a dropzone name.

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -2972,6 +2972,10 @@ bool CFileSprayEx::onFileList(IEspContext &context, IEspFileListRequest &req, IE
         if (!validateDropZonePath(nullptr, netaddr, sPath) && !validateConfigurationDirectory(nullptr, "log", nullptr, nullptr, sPath)) //The path should be the absolute path for the dropzone or log file.
             throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "Invalid file path %s", sPath.str());
 
+        const char *dropZoneId = isContainerized() ? req.getDropZoneName() : netaddr;
```
jakesmith, line 2975: see prev. comments re. looking up dropzone name.

But I see this ::onFileList method insists on a netaddr : 
```
        const char* netaddr = req.getNetaddr();
        if (!netaddr || !*netaddr)
            throw MakeStringException(ECLWATCH_INVALID_INPUT, "Network address not specified.");
```

that looks wrong ?

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -3567,6 +3596,78 @@ bool CFileSprayEx::onDeleteDropZoneFiles(IEspContext &context, IEspDeleteDropZon
     return true;
 }
 
+void CFileSprayEx::checkDropZoneFileScopeAccess(IEspContext &context, const char *dropZoneName, const char *netAddress,
+    const char *dropZonePath, const StringArray &dropZoneFiles, SecAccessFlags accessReq)
+{
+    const char *accessReqName = getSecAccessFlagName(accessReq);
+    const char *dropZoneId = isContainerized() ? dropZoneName : netAddress;
```
jakesmith, line 3603: see prev. comments, I think this and other places should unify on dropZoneName only.

dali/base/dautils.cpp
```diff
@@ -65,6 +65,29 @@ IPropertyTree * getDropZonePlane(const char * name)
     return getGlobalConfigSP()->getPropTree(xpath);
 }
 
+StringBuffer & makeDropZoneScope(StringBuffer & scope, bool isPlane, const char *dropZone, const char *path)
```
jakesmith, line 68: this is _very_ similar to CDfsLogicalFilename::setExternal
and it supports validating the drop zone is a valid plane etc.

It would be better to alter CDfsLogicalFilename to support this.
i.e. introduce a setPlaneExternal(const char *planeName, const char *path);
which is almost same as the existing setExternal, can use common code.

dali/base/dadfs.cpp
```diff
@@ -11759,6 +11760,13 @@ SecAccessFlags CDistributedFileDirectory::getFDescPermissions(IFileDescriptor *f
     return retPerms;
 }
 
+SecAccessFlags CDistributedFileDirectory::getDropZoneScopePermissions(const char *dropZoneNameOrIP,const char *dropZonePath,bool isPlane,IUserDescriptor *user,unsigned auditflags)
+{
+    StringBuffer scope;
+    makeDropZoneScope(scope,isPlane,dropZoneNameOrIP,dropZonePath);
```
jakesmith, line 11766: see comments on impl. re. using CDfsLogicalFileName,
e.g. this becomes:
```
CDfsLogicalFileName lfn;
lfn.setPlaneExternal(dropZoneName, dropZonePath);
return getScopePermissions(lfn.get(),user,auditflags);
```

jakesmith (commented):
@wangkx - please see comments

It would be worth making 2 of these changes as separate small PR's:
1) Modify CDfsLogicalFileName to cope with plane setting.
2) Add a findHostedPlane function to map from a host/ip to a known dropzone name.

esp/services/ws_fs/ws_fsService.cpp
wangkx: In bare-metal, one dropzone may have multiple dropzone servers. The ::onFileList method is designed to retrieve dropzone files in one dropzone server.

esp/services/ws_fs/ws_fsBinding.cpp
wangkx: Do we also need a findDropZoneName(const char *host) for non-plane dropzones?

wangkx: For a non-plane dropzone, is it possible to allow a user to access one dropzone server, but, not access another dropzone server?

dali/base/dadfs.cpp
wangkx:
```
    CDfsLogicalFileName lfn;
    if (isPlane)
        lfn.setPlaneExternal(dropZoneNameOrIP, dropZonePath);
    else
        lfn.setExternal(dropZoneNameOrIP, dropZonePath);
    return getScopePermissions(lfn.get(),user,auditflags);
```

esp/services/ws_fs/ws_fsBinding.cpp
jakesmith: they could.. if we've constructed a external scope with host/ip - they could LDAP scope those if they wanted to.

esp/services/ws_fs/ws_fsService.cpp
jakesmith: ok.. but what about onFileList in cloud ?

dali/base/dadfs.cpp
jakesmith: yes something like that, I think it will further simplify when the dropzone plane name is resolved in the case when an host/ip is supplied ( as per prev. comments re. planes are auto-generated on baremetal for drop zones )

esp/services/ws_fs/ws_fsService.cpp
wangkx: In the existing commits of this PR, a DropZoneName is added into the request. This check should be removed since, in cloud, we may find out the host based on the DropZoneName.

wangkx:
> Add a findHostedPlane function to map from a host/ip to a known dropzone name.
@jakesmith If the host is 'localhost' (see https://github.com/hpcc-systems/HPCC-Platform/blob/fac22a93301d232d77cb191c11869123a6ec9e31/esp/smc/SMCLib/TpContainer.cpp#L114), could the host be mapped to a plane name? If not, should the function name be changed to findHostedDropZonePlane() -- find out a dropzone plane without "hosts"?

esp/services/ws_fs/ws_fsService.cpp
jakesmith: in cloud it should always be a dropzone name that is supplied, never a host/ip.

jakesmith:
> > Add a findHostedPlane function to map from a host/ip to a known dropzone name.
> 
> @jakesmith If the host is 'localhost' (see
> 
> https://github.com/hpcc-systems/HPCC-Platform/blob/fac22a93301d232d77cb191c11869123a6ec9e31/esp/smc/SMCLib/TpContainer.cpp#L114
> 
> ), could the host be mapped to a plane name? If not, should the function name be changed to findHostedDropZonePlane() -- find out a dropzone plane with/without "hosts"?


I'm not sure how/when gatherDropZoneMachines is used, it looks like it's for the other way around, i.e. it's handed a plane, and it's returning a list of hosts from it..

At least in cloud world, the user/UI normally not see the underlying hosts/ip, only the dropzone name.
In bare-metal - I think it should really be the same, but there's the caveat as you say of a single dropzone with multiple hosts.. I wonder whethe we should try to drop support for that.
Is there any benefit vs 2 dropzones each with 1 host?

wangkx:
The gatherDropZoneMachines is called when ECLWatch displays the dropzone-hosts-files tree through the WsTopology.TpDropZoneQuery.
A single dropzone with multiple hosts support was added to ECLWatch by https://track.hpccsystems.com/browse/HPCC-17058. Even before that, ECWatch exposes hosts/ip more than a name. In bare-metal, each dropzone host has a name attribute. But, the name is not forced to be unique.

jakesmith:
> But, the name is not forced to be unique.

it should be.

esp/services/ws_fs/ws_fsService.cpp
wangkx: > This check should be removed since, in cloud, we may find out the host based on the DropZoneName.
@jakesmith I may take it back. In the existing design, the ::onFileList method is designed to retrieve dropzone files in one dropzone host. One DropZone may have >1 hosts. If the netaddr is not specified, we do not know which host's dropzone files should be reported, even in cloud. I am not sure, here, we should report the dropzone files in all of the hosts. To report the dropzone files in all of the hosts, a host layer may add on the top of the files. That will break the existing ECLWatch (the Landing Zone page). ECLWatch has to parse the data, find the files for each host, and pre-populate the data for each host. Or ECLWatch may remove the host layer (not display the files under a host, but, display the files under a dropzone).

wangkx: I guess I can report the dropzone files in all of the hosts if no host is specified.

jakesmith (commented):
@wangkx - passing this back to you. Looks like it's work in progress/needs rebasing etc.

wangkx:
Yes, I am still working on it. Will rebase.

wangkx:
@jakesmith please review

jakesmith (changes requested):
@wangkx - looks close, please see comments.

wangkx:
@jakesmith please review the last 2 commits.

dali/base/dautils.cpp
```diff
@@ -46,18 +46,21 @@
...
     if (plane->hasProp("@hostGroup"))
         planeGroup.setown(getHostGroup(plane->queryProp("@hostGroup"), true));
-    else
-    {
-        if (!plane->hasProp("hosts"))
-            return false;
+    else if (plane->hasProp("hosts"))
         planeGroup.set(plane); // plane itself holds 'hosts'
-    }
+    return planeGroup.getClear();
```
jakesmith, line 56: trivial: setting into 'planeGroup' doesn't really make code clearer, would be more concise to code as:

```
    if (plane->hasProp("@hostGroup"))
        return getHostGroup(plane->queryProp("@hostGroup"), true);
    else if (plane->hasProp("hosts"))
        return LINK(plane); // plane itself holds 'hosts'
```

jakesmith (commented):
@wangkx - looks good, 1 trivial comment, please squash.

wangkx:
@jakesmith I made the changes based on your last review. Please review.

jakesmith (approved):
@wangkx - can't see any other issues. 

@RussWhitehead - please check.

esp/services/ws_fs/ws_fsBinding.cpp
```diff
@@ -393,12 +393,15 @@ int CFileSpraySoapBindingEx::downloadFile(IEspContext &context, CHttpRequest* re
...
 
-        StringBuffer netAddressStr, osStr, pathStr, nameStr;
+        StringBuffer netAddressStr, osStr, pathStr, nameStr, dropZoneName;
         request->getParameter("NetAddress", netAddressStr);
         request->getParameter("OS", osStr);
         request->getParameter("Path", pathStr);
         request->getParameter("Name", nameStr);
+        request->getParameter("DropZoneName", dropZoneName);
 
+        if (getDropZoneScopePermissions(context, dropZoneName, pathStr, netAddressStr) < SecAccess_Read)
```
RussWhitehead, line 403: To help Ops diagnose problems without our involvement, would it help to include the user name, and the actual LDAP  permission setting?

esp/services/ws_fs/ws_fsBinding.cpp
```diff
@@ -479,11 +482,14 @@ int CFileSpraySoapBindingEx::downloadFile(IEspContext &context, CHttpRequest* re
...
 {
-    StringBuffer netAddress, path;
+    StringBuffer netAddress, path, dropZoneName;
     request->getParameter("NetAddress", netAddress);
     request->getParameter("Path", path);
+    request->getParameter("DropZoneName", dropZoneName);
     if (!validateDropZonePath(nullptr, netAddress, path)) //The path should be the absolute path for the dropzone.
         throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "Invalid Landing Zone path %s", path.str());
+    if (getDropZoneScopePermissions(ctx, dropZoneName, path, netAddress) < SecAccess_Full)
+        throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "Access DropZone Scope %s %s %s not allowed. Full Access Required.", dropZoneName.str(), netAddress.str(), path.str());
```
RussWhitehead, line 492: See above comment line 403

wangkx:
@RussWhitehead the user name and allowed permission have been added to the error messages. Please review.

RussWhitehead (approved):
@wangkx  Looks good

wangkx:

PR #16714, opened 2022-11-17
HPCC-28506 Report helm warnings/errors from esp config in wudetails for global scope

shamser:
## Type of change:
- [ ] This change is a bug fix (non-breaking change which fixes an issue).
- [x] This change is a new feature (non-breaking change which adds functionality).
- [ ] This change improves the code (refactor or other change that does not change the functionality)
- [ ] This change fixes warnings (the fix does not alter the functionality or the generated code)
- [ ] This change is a breaking change (fix or feature that will cause existing behavior to change).
- [ ] This change alters the query API (existing queries will have to be recompiled)

## Checklist:
- [x] My code follows the code style of this project.
  - [ ] My code does not create any new warnings from compiler, build system, or lint.
- [x] The commit message is properly formatted and free of typos.
  - [x] The commit message title makes sense in a changelog, by itself.
  - [x] The commit is signed.
- [x] My change requires a change to the documentation.
  - [ ] I have updated the documentation accordingly, or...
  - [ ] I have created a JIRA ticket to update the documentation.
  - [ ] Any new interfaces or exported functions are appropriately commented.
- [x] I have read the CONTRIBUTORS document.
- [x] The change has been fully tested:
  - [ ] I have added tests to cover my changes.
  - [ ] All new and existing tests passed.
  - [ ] I have checked that this change does not introduce memory leaks.
  - [ ] I have used Valgrind or similar tools to check for potential issues.
- [ ] I have given due consideration to all of the following potential concerns:
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization
  - [ ] Existing deployed queries will not be broken
  - [ ] This change fixes the problem, not just the symptom
  - [ ] The target branch of this pull request is appropriate for such a change.
- [ ] There are no similar instances of the same problem that should be addressed
  - [ ] I have addressed them here
  - [ ] I have raised JIRA issues to address them separately
- [ ] This is a user interface / front-end modification
  - [ ] I have tested my changes in multiple modern browsers
  - [ ] The component(s) render as expected

## Smoketest:
- [ ] Send notifications about my Pull Request position in Smoketest queue.
- [x] Test my draft Pull Request.

PR #16716, opened 2022-11-18
HPCC-28493 Add existing helm warnings to global system errors

shamser:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/helm/hpcc/templates/NOTES.txt
+++ b/helm/hpcc/templates/NOTES.txt
@@ -1,33 +1,8 @@
 {{- $defaultImage :=  include "hpcc.imageName" (dict "root" $ "me" dict ) | quote }}
 {{- $defaultVersion :=  .Values.global.image.version | default .Chart.Version }}
 Thank you for installing the HPCC chart version {{ .Chart.Version }} using image {{ $defaultImage }}
+{{- include "hpcc.printWarnings" (dict "root" $) | indent 4 }}
 
-{{ if (ne $defaultVersion .Chart.Version) }}
- {{- printf "**** WARNING: The expected image version for this helm chart is %s ****" .Chart.Version }}
-{{- end }}
-{{/* Gather a list of ephemeral and persistant planes */}}
-{{- $storage := (.Values.storage | default dict) -}}
-{{- $match := dict "ephemeral" (list) "persistant" (list) -}}
-{{- $planes := ($storage.planes | default list) -}}
-{{- $searchLabels := list "data" "dali" "sasha" "dll" "lz" -}}
-{{- range $plane := $planes -}}
- {{- $labels := $plane.labels | default (list "data") -}}
- {{- $tag := (hasKey $plane "storageClass" | ternary "ephemeral" "persistant") }}
- {{- range $label := $labels -}}
-  {{- if has $label $searchLabels -}}
-   {{- $prev := get $match $tag -}}
-   {{- $_ := set $match $tag (append $prev $plane.name) -}}
-  {{- end }}
- {{- end }}
-{{- end }}
-{{- /* Warn if any of the planes are ephemeral */ -}}
-{{- if (ne (len $match.ephemeral) 0) }}
- {{- printf "**** WARNING: The configuration contains ephemeral planes: %s ****" (print (uniq $match.ephemeral)) }}
-{{- end }}
-{{/* Check for a mixture of ephemeral and persistent planes - ignore spill - generally indicates a problem.  Could use required to force a failure  */}}
-{{- if and (ne (len $match.ephemeral) 0) (ne (len $match.persistant) 0) }}
- {{- printf "**** ERROR: The configuration contains a mixture of ephemeral %s and persistent %s planes ****" (print (uniq $match.ephemeral)) (print (uniq $match.persistant)) }}
-{{ end }}
 This chart has defined the following HPCC components:
 {{- $names := dict -}}
 {{- range $sectionName, $section := .Values -}}
--- a/helm/hpcc/templates/_helpers.tpl
+++ b/helm/hpcc/templates/_helpers.tpl
@@ -2058,3 +2058,31 @@ spec:
 {{- end -}}
 {{- end -}}
 
+{{- define "hpcc.generateWarnings" -}}
+{{/*
+Generate a list of warnings
+Pass in dict with root
+*/}}
+{{- $ctx := dict "warnings" list "root" .root -}}
+{{- include "hpcc.getWarnings" $ctx -}}
+{{- if $ctx.warnings }}
+warnings:
+ {{- range $warning := $ctx.warnings }}
+- msg: {{ $warning.msg | quote }}
+  severity: {{ $warning.severity | quote }}
+  source: {{ $warning.source | quote }}
+ {{- end -}}
+{{- end -}}
+{{- end -}}
+
+{{- define "hpcc.printWarnings" -}}
+{{/*
+Print warnings
+Pass in dict with root
+*/}}
+{{- $ctx := dict "warnings" list "root" .root -}}
+{{- include "hpcc.getWarnings" $ctx -}}
+{{- range $warning := $ctx.warnings }}
+ {{ printf "**** %s: %s ****" (upper $warning.severity) $warning.msg }}
+{{- end -}}
+{{- end -}}
--- a/helm/hpcc/templates/_warnings.tpl
+++ b/helm/hpcc/templates/_warnings.tpl
@@ -1,0 +1,55 @@
+{{/*
+
+##############################################################################
+
+    HPCC SYSTEMS software Copyright (C) 2022 HPCC Systems®.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+##############################################################################
+
+*/}}
+{{/*
+Generate a list of warnings
+Pass in dict with root and warnings
+*/}}
+{{- define "hpcc.getWarnings" -}}
+ {{- $ctx := . -}}
+ {{- $defaultVersion := .root.Values.global.image.version | default .root.Chart.Version -}}
+ {{- if (ne $defaultVersion .root.Chart.Version) -}}
+  {{- $warning := dict "source" "helm" "severity" "warning" -}}
+  {{- $_ := set $warning "msg" (printf "The expected image version for this helm chart is %s" .root.Chart.Version) -}}
+  {{- $_ := set $ctx "warnings" (append $ctx.warnings $warning) -}}
+ {{- end -}}
+ {{- /* Gather a list of ephemeral and persistant planes */ -}}
+ {{- $storage := (.root.Values.storage | default dict) -}}
+ {{- $match := dict "ephemeral" (list) "persistant" (list) -}}
+ {{- $planes := ($storage.planes | default list) -}}
+ {{- $searchLabels := list "data" "dali" "sasha" "dll" "lz" -}}
+ {{- range $plane := $planes -}}
+  {{- $labels := $plane.labels | default (list "data") -}}
+  {{- $tag := (hasKey $plane "storageClass" | ternary "ephemeral" "persistant") -}}
+  {{- range $label := $labels -}}
+   {{- if has $label $searchLabels -}}
+    {{- $prev := get $match $tag -}}
+    {{- $_ := set $match $tag (append $prev $plane.name) -}}
+   {{- end -}}
+  {{- end -}}
+ {{- end -}}
+ {{- /* Warn if any of the planes are ephemeral */ -}}
+ {{- if (ne (len $match.ephemeral) 0) -}}
+  {{ $warning := dict "source" "helm" "severity" "warning" -}}
+  {{- $_ := set $warning "msg" (printf "The configuration contains ephemeral planes: %s" (print (uniq $match.ephemeral)) )  -}}
+  {{- $_ := set $ctx "warnings" (append $ctx.warnings $warning) -}}
+ {{- end -}}
+{{- end -}}
--- a/helm/hpcc/templates/esp.yaml
+++ b/helm/hpcc/templates/esp.yaml
@@ -75,6 +75,9 @@ data:
       dfuQueues:
 {{- include "hpcc.generateConfigDfuQueues" . | indent 6}}
 {{ include "hpcc.generateVaultConfig" . | indent 6 }}
+{{- if eq .me.application "eclwatch" }}
+ {{- include "hpcc.generateWarnings" (dict "root" .root) | nindent 6 }}
+{{- end }}
     global:
 {{ include "hpcc.generateGlobalConfigMap" .root | indent 6 }}
 {{- end -}}
@@ -83,8 +86,6 @@ data:
 {{- if not .disabled -}}
 {{- $env := concat ($.Values.global.env | default list) (.env | default list) -}}
 {{- $application := .application | default "eclwatch" -}}
-
-
 {{- $authtype := .auth | default "none" -}}
 {{- $noAuthSecretsCategories := ternary (list "storage" "esp" "codeSign" "codeVerify")  (list "storage" "esp") (eq $application "eclwatch") -}}
 {{- $secretsCategories := ternary (append $noAuthSecretsCategories "authn") $noAuthSecretsCategories  (eq $authtype "ldap") -}}


shamser:
This PR writes helm warnings to esp config.  (https://track.hpccsystems.com/browse/HPCC-28506 make these warnings available through wudetails)

helm/hpcc/templates/_helpers.tpl
```diff
@@ -2058,3 +2058,27 @@ spec:
...
+{{- define "hpcc.generateWarnings" -}}
+{{/*
+Generate a list of warnings
+*/}}
+{{ $warnings := include "hpcc.getWarnings" .root | fromJsonArray }}
+{{ if $warnings }}
+warnings:
+ {{ range $warning := $warnings }}
+  - msg: {{ $warning.msg | quote }}
+    severity: warning
```
ghalliday, line 2070: Should use the severity from the $warning

helm/hpcc/templates/_helpers.tpl
```diff
@@ -2058,3 +2058,27 @@ spec:
...
 
+{{- define "hpcc.generateWarnings" -}}
+{{/*
+Generate a list of warnings
+*/}}
+{{ $warnings := include "hpcc.getWarnings" .root | fromJsonArray }}
+{{ if $warnings }}
+warnings:
+ {{ range $warning := $warnings }}
+  - msg: {{ $warning.msg | quote }}
```
ghalliday, line 2069: indent: indent looks wrong.  Will this cause problems processing the file?

shamser: Strangely, it did work.  But I'll fix the format.

shamser:
> @shamser @ghalliday - I'll admit haven't starred v. closely, but this PR looks like a subset of #16714 ?

This PR of the changes made in #16714 . Once 16714 it merged, I'll rebase to candidate-8.10.x.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -2058,3 +2058,29 @@ spec:
...
+{{- end -}}
+
+{{- define "hpcc.printWarnings" -}}
+{{/*
+Print warnings
+*/}}
+{{- $ctx := dict "warnings" list | merge .root -}}
+{{- include "hpcc.getWarnings" $ctx -}}
+{{- range $warning := $ctx.warnings }}
+  {{ printf "**** %s: %s ****" (upper $warning.severity) $warning.msg }}
```
jakesmith, line 2084: formatting: indentation should be 1 space

helm/hpcc/templates/_warnings.tpl
```diff
@@ -0,0 +1,59 @@
...
+    limitations under the License.
+
+##############################################################################
+
+*/}}
+{{/*
+Generate a list of warnings
+*/}}
+{{- define "hpcc.getWarnings" -}}
+  {{- $warnings := list -}}
```
jakesmith, line 28: There are some existing inconsistencies, but we agreed to use 1 character indentation in helm charts

helm/hpcc/templates/_warnings.tpl
```diff
@@ -0,0 +1,59 @@
...
+  {{- $root := . -}}
+  {{- $defaultImage :=  include "hpcc.imageName" (dict "root" $ "me" dict ) | quote -}}
+  {{- $defaultVersion := .Values.global.image.version | default .Chart.Version -}}
+  {{ if (ne $defaultVersion .Chart.Version) -}}
+    {{ $warning := dict "source" "helm" "severity" "warning" -}}
```
jakesmith, line 33: trivial: might not make a material difference, but as this helper func. is not outputting anything, should consistently chomp leading and trailing whitespace on ech template line, i.e. above 2 lines should have {-

helm/hpcc/templates/_warnings.tpl
```diff
@@ -0,0 +1,59 @@
...
+##############################################################################
+
+*/}}
+{{/*
+Generate a list of warnings
+*/}}
+{{- define "hpcc.getWarnings" -}}
+  {{- $warnings := list -}}
+  {{- $root := . -}}
+  {{- $defaultImage :=  include "hpcc.imageName" (dict "root" $ "me" dict ) | quote -}}
```
jakesmith, line 30: trivial/formatting: double space after :=

helm/hpcc/templates/esp.yaml
```diff
@@ -75,6 +75,9 @@ data:
       dfuQueues:
 {{- include "hpcc.generateConfigDfuQueues" . | indent 6}}
 {{ include "hpcc.generateVaultConfig" . | indent 6 }}
+{{- if eq .me.application "eclwatch" }}
+{{ include "hpcc.generateWarnings" . | indent 6 }}
```
jakesmith, line 79: trivial: indent by 1 space for clarity

helm/hpcc/templates/_warnings.tpl
```diff
@@ -0,0 +1,59 @@
...
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+##############################################################################
+
+*/}}
+{{/*
+Generate a list of warnings
```
jakesmith, line 25: trivial: generally helper function have a line "Pass in ..." documenting what the function expects as input, e.g.:
```
{{/*
A template to generate a PVC
Pass in dict with root, me, name, and optional path
*/}}
```

jakesmith (commented):
@shamser - some trivial issues.

jakesmith (approved):
@shamser - looks good

helm/hpcc/templates/NOTES.txt
```diff
@@ -1,10 +1,8 @@
 {{- $defaultImage :=  include "hpcc.imageName" (dict "root" $ "me" dict ) | quote }}
 {{- $defaultVersion :=  .Values.global.image.version | default .Chart.Version }}
 Thank you for installing the HPCC chart version {{ .Chart.Version }} using image {{ $defaultImage }}
+{{ include "hpcc.printWarnings" (dict "root" .) | indent 4 -}}
```
ghalliday, line 4: Can lines 6-20 also be deleted?  I think they are only used to update $match which was used for reporting warnings.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -2058,3 +2058,31 @@ spec:
 {{- end -}}
 {{- end -}}
 
+{{- define "hpcc.generateWarnings" -}}
+{{/*
+Generate a list of warnings
+Pass in dict with root
```
ghalliday, line 2064: Not clear what root is for.  To allow extra items to be passed in?  Otherwise I don't think it is used (or I don't understand) - .root is not used within the function.

ghalliday (commented):
@shamser looks good.  A couple of final questions.

helm/hpcc/templates/NOTES.txt
shamser: You're right. I'll remove these lines.

helm/hpcc/templates/_helpers.tpl
shamser: It is passed down to getWarnings() and it is needed to generate the warnings.

ghalliday: Why?  $ctx is passed, and warnings is filled in.  But what does merging .root do?  trying follow the logic.

shamser: _warnings.tpl (line 31) includes helper function hpcc.imageName.  And hpcc.imageName uses .root.Values.global.image.name.  If I don't pass root, I get the following error: 

Error: template: ...error calling include: template: hpcc/templates/_helpers.tpl:1825:25: executing "hpcc.filterConfig" at <include .configMapHelper .>: error calling include: template: hpcc/templates/esp.yaml:79:3: executing "hpcc.espConfigMap" at <include "hpcc.generateWarnings" .>: error calling include: template: hpcc/templates/_helpers.tpl:2067:4: executing "hpcc.generateWarnings" at <include "hpcc.getWarnings" $ctx>: error calling include: template: hpcc/templates/_warnings.tpl:31:22: executing "hpcc.getWarnings" at <include "hpcc.imageName" (dict "root" $ "me" dict)>: error calling include: template: hpcc/templates/_helpers.tpl:698:8: executing "hpcc.imageName" at <.root.Values.global.image.root>: nil pointer evaluating interface {}.global

helm/hpcc/templates/_warnings.tpl
```diff
@@ -0,0 +1,60 @@
...
+Pass in dict with root with warnings
+*/}}
+{{- define "hpcc.getWarnings" -}}
+ {{- $warnings := list -}}
+ {{- $root := . -}}
+ {{- $defaultImage := include "hpcc.imageName" (dict "root" $ "me" dict ) | quote -}}
```
ghalliday, line 31: $defaultImage isn't actually used by the following code.

helm/hpcc/templates/_helpers.tpl
jakesmith: Agree, it should pass in the root dict as "root":
```
{{- $ctx := dict "warnings" list "root" .root -}}
```

and then reference "root" in getWarnings.
But that means changing all lines that reference . to .root. e.g.:
```
 {{- $defaultVersion := .Values.global.image.version | default .Chart.Version -}}
```
should change to:
```
 {{- $defaultVersion := .Values.global.image.version | default .root.Chart.Version -}}
```

Also, the last line where it's setting "warnings" into the returned ctx, that should be changed to:
```
 {{- $_ := set . "warnings" $warnings -}}
```

since the passed in ctx is . 

and remove the:
```
 {{- $root := . -}}
```

helm/hpcc/templates/_warnings.tpl
```diff
@@ -0,0 +1,60 @@
...
+ {{- $defaultVersion := .Values.global.image.version | default .Chart.Version -}}
```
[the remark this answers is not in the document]
jakesmith: yes, .root though, the $root = .root isn't necessary / makes less clear

helm/hpcc/templates/_helpers.tpl
ghalliday: This still isn't right.  Now you are passing in . as the parameter, so there is no .root.

I think you need to
i) go back to passing a dictionary as a parmeter
ii) use $ctx := dict "warnings" list "root" .root
iii) delete the definition of $defaultImage in the warning code
iv) Use .root.x where you previously have .x e.g. .root.Chart.version, .root.Values.storage

I still don't quite understand what $ means @jakesmith can you clarify when that changes?

jakesmith: $ changes within an include template function to the dictionary passed to it.
So by convention we pass in the root context ($) as "root" and use it within our template functions

helm/hpcc/templates/esp.yaml
```diff
@@ -75,6 +75,9 @@ data:
       dfuQueues:
 {{- include "hpcc.generateConfigDfuQueues" . | indent 6}}
 {{ include "hpcc.generateVaultConfig" . | indent 6 }}
+{{ if eq .me.application "eclwatch" }}
```
jakesmith, line 78: minor: this will 'leak' a newline, should I think be (but should be checked):
```
{{- if eq .me.application "eclwatch" }}
```
similarly with the end

helm template whitespace chomping is horrid!

helm/hpcc/templates/_warnings.tpl
```diff
@@ -0,0 +1,58 @@
...
+    limitations under the License.
+
+##############################################################################
+
+*/}}
+{{/*
+Generate a list of warnings
+Pass in dict with root with warnings
+*/}}
+{{ define "hpcc.getWarnings" }}
```
jakesmith, line 28: minor: why was the whitespace chomping removed here (in commit 5) ?

All our helper functions have whitespace chomping here to avoid introducing new lines each time they are called

helm/hpcc/templates/_warnings.tpl
```diff
@@ -0,0 +1,58 @@
...
+ {{- $warnings := list -}}
+ {{- $defaultVersion := .root.Values.global.image.version | default .root.Chart.Version -}}
+ {{- if (ne $defaultVersion .root.Chart.Version) -}}
+  {{- $warning := dict "source" "helm" "severity" "warning" -}}
+  {{- $_ := set $warning "msg" (printf "The expected image version for this helm chart is %s" .root.Chart.Version) -}}
+  {{- $warnings = append $warnings $warning -}}
+ {{- end -}}
+ {{/* Gather a list of ephemeral and persistant planes */}}
```
jakesmith, line 36: minor: should be {{- /* ... */ -}}

helm/hpcc/templates/_warnings.tpl
```diff
@@ -0,0 +1,58 @@
...
+    {{- $prev := get $match $tag -}}
+    {{- $_ := set $match $tag (append $prev $plane.name) -}}
+   {{- end -}}
+  {{- end -}}
+ {{- end -}}
+ {{- /* Warn if any of the planes are ephemeral */ -}}
+ {{- if (ne (len $match.ephemeral) 0) -}}
+  {{ $warning := dict "source" "helm" "severity" "warning" -}}
+  {{- $_ := set $warning "msg" (printf "The configuration contains ephemeral planes: %s" (print (uniq $match.ephemeral))) -}}
+  {{- $warnings = append $warnings $warning -}}
```
jakesmith, line 55: hm, I'm guessing this works as is (?), but should be := 
(same on line 34)

helm/hpcc/templates/_warnings.tpl
```diff
@@ -0,0 +1,58 @@
...
+ {{- end -}}
+ {{- $_ := set . "warnings" $warnings -}}
+{{ end -}}
```
jakesmith, line 58: minor: would expect this to be {{- end -}}

If there are whitespace problems after making this change (and adding chomping to the define), then needs fixing where called.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -2058,3 +2058,31 @@ spec:
 {{- end -}}
 {{- end -}}
 
+{{ define "hpcc.generateWarnings" }}
```
jakesmith, line 2061: minor: why was the whitespace chomping removed here (in commit 5) ?

All our helper functions have whitespace chomping here to avoid introducing new lines each time they are called

helm/hpcc/templates/_helpers.tpl
```diff
@@ -2058,3 +2058,31 @@ spec:
...
+warnings:
+ {{- range $warning := $ctx.warnings }}
+- msg: {{ $warning.msg | quote }}
+  severity: {{ $warning.severity | quote }}
+  source: {{ $warning.source | quote }}
+ {{- end -}}
+{{- end -}}
+{{- end }}
+
+{{ define "hpcc.printWarnings" }}
```
jakesmith, line 2078: minor: why was the whitespace chomping removed here (in commit 5) ?

All our helper functions have whitespace chomping here to avoid introducing new lines each time they are called

jakesmith (commented):
@shamser - looks good, 1 issue re. "=" vs ":=" and some minor white space issues.

helm/hpcc/templates/esp.yaml
shamser: I removed it because the formatting became messed up - it was reporting that it couldn't convert to json.  But I;ve just adding it back in and I can't see the original problem anymore.

helm/hpcc/templates/_warnings.tpl
shamser: Again, I removed when I was trying to get it to work.  I'll add back in.

helm/hpcc/templates/_warnings.tpl
shamser: It does work with just = but I'll change

jakesmith:
@shamser - I expect to see some warnings when I ran ./startall.sh -t , because branch is dirty and so version will not match the helm version, and without any other config all planes will be ephemeral,
but I don't see any generated warnings (in esp eclwatch config)

helm/hpcc/templates/esp.yaml
jakesmith: as I say, getting whitespace chomping right is horrible. You have to be really careful, not to remove a newline where you need it etc.

helm/hpcc/templates/esp.yaml
```diff
@@ -75,6 +75,9 @@ data:
       dfuQueues:
 {{- include "hpcc.generateConfigDfuQueues" . | indent 6}}
 {{ include "hpcc.generateVaultConfig" . | indent 6 }}
+{{- if eq .me.application "eclwatch" }}
+ {{ include "hpcc.generateWarnings" (dict "root" .root) | indent 6 }}
```
jakesmith, line 79: this might cause problems, it is indented by 1 space, and then by 6,.
I suggest:
```
{{- if eq .me.application "eclwatch" }}
 {{- include "hpcc.generateWarnings" (dict "root" .root) | nindent 6 }}
{{- end }}
```

jakesmith (commented):
@shamser - spotted a few issues that may be causing the lack of generated warnings I'm seeing.

shamser:
@jakesmith I made the suggested just now but then I get different errors.  The latest commit works.

helm/hpcc/templates/_warnings.tpl
```diff
@@ -25,13 +25,13 @@
...
 */}}
-{{- define "hpcc.getWarnings" -}}
+{{ define "hpcc.getWarnings" }}
  {{- $warnings := list -}}
  {{- $defaultVersion := .root.Values.global.image.version | default .root.Chart.Version -}}
  {{- if (ne $defaultVersion .root.Chart.Version) -}}
   {{- $warning := dict "source" "helm" "severity" "warning" -}}
   {{- $_ := set $warning "msg" (printf "The expected image version for this helm chart is %s" .root.Chart.Version) -}}
-  {{- $warnings := append $warnings $warning -}}
+  {{- $warnings = append $warnings $warning -}}
```
jakesmith, line 34: I'm not keen on this. It's undocumented behaviour afaics (and the sole use of this).
As elsewhere, if we need to set something inside a scope, we use set it into a dictionary with $_ : = set ...

shamser: I'll try to get it to work with :=

helm/hpcc/templates/_helpers.tpl
jakesmith: @shamser - fixing this (adding the chomping back) caused problems?

My guess is that it's because on line 2068:
```
{{- if $ctx.warnings -}}
```

that shouldn't chomp the newline at the end. It should be:
```
{{- if $ctx.warnings }}
```

I suspect spurious newline(s) introduced on the {{ define .. }} were working around the chomp here.

jakesmith: Tested. I made the change above, and added back the chomping to the generateWarnings, printWarnings and getWarning define's and it generate correctly.

helm/hpcc/templates/_warnings.tpl
```diff
@@ -0,0 +1,57 @@
...
+Generate a list of warnings
+Pass in dict with root with warnings
+*/}}
+{{ define "hpcc.getWarnings" }}
+ {{- $warnings := list -}}
+ {{- $defaultVersion := .root.Values.global.image.version | default .root.Chart.Version -}}
+ {{- if (ne $defaultVersion .root.Chart.Version) -}}
+  {{- $warning := dict "source" "helm" "severity" "warning" -}}
+  {{- $_ := set $warning "msg" (printf "The expected image version for this helm chart is %s" .root.Chart.Version) -}}
+  {{- $_ := set . "warnings" (append .warnings $warning) -}}
```
ghalliday, line 34: As Jake suggested elsewhere it is better to have
`{{- $ctx := . }}
`at the head of the function, and then set $ctx "warnings
The reason is, that if this was copied and executed within a range it will update the wrong value.

helm/hpcc/templates/_warnings.tpl
```diff
@@ -0,0 +1,57 @@
+{{/*
+
+---  DO NOT EDIT THIS FILE ---
```
ghalliday, line 3: Why not?!

helm/hpcc/templates/esp.yaml
```diff
@@ -75,6 +75,9 @@ data:
       dfuQueues:
 {{- include "hpcc.generateConfigDfuQueues" . | indent 6}}
 {{ include "hpcc.generateVaultConfig" . | indent 6 }}
+{{- if eq .me.application "eclwatch" }}
+ {{- include "hpcc.generateWarnings" (dict "root" .root) | indent 6 }}
```
ghalliday, line 79: You need to change indent to nindent (newline + indent)

helm/hpcc/templates/_warnings.tpl
jakesmith: I'd prefer that (as per my suggession) for same reason - "." is 'ok' here, but it is contextual - so not as clear as using an explicit ctx.

helm/hpcc/templates/_warnings.tpl
shamser: Not sure but it's copied from the _helpers.tpl.

jakesmith (commented):
>https://github.com/hpcc-systems/HPCC-Platform/pull/16716#issuecomment-1325132511

Did you see my other comment (https://github.com/hpcc-systems/HPCC-Platform/pull/16716#discussion_r1030461370) re. fix to if chomping that will cause problems? (but are probabl getting away with because of the lack of chomping on the helper define lines)


@shamser - this one not yet fixed.

jakesmith (commented):
Having to add another comment to fix review status

jakesmith (changes requested):
Not sure where the approve status came from, or how to change it without adding another comment with request changes..

jakesmith (commented):
I think looks good now.
Unless @ghalliday objects, please squash

helm/hpcc/templates/_warnings.tpl
```diff
@@ -0,0 +1,56 @@
...
+##############################################################################
+
+*/}}
+{{/*
+Generate a list of warnings
+Pass in dict with root with warnings
+*/}}
+{{- define "hpcc.getWarnings" -}}
+ {{- $ctx := . -}}
+ {{- $warnings := list -}}
```
jakesmith, line 28: minor: unused

helm/hpcc/templates/_warnings.tpl
```diff
@@ -0,0 +1,56 @@
...
+ {{- $defaultVersion := .root.Values.global.image.version | default .root.Chart.Version -}}
+ {{- if (ne $defaultVersion .root.Chart.Version) -}}
+  {{- $warning := dict "source" "helm" "severity" "warning" -}}
+  {{- $_ := set $warning "msg" (printf "The expected image version for this helm chart is %s" .root.Chart.Version) -}}
+  {{- $_ := set $ctx "warnings" (append .warnings $warning) -}}
```
jakesmith, line 33: although "." is ctx in this context - it is clearer if it explicitly references it:
```
```

helm/hpcc/templates/_warnings.tpl
```diff
@@ -0,0 +1,56 @@
...
+    {{- $prev := get $match $tag -}}
+    {{- $_ := set $match $tag (append $prev $plane.name) -}}
+   {{- end -}}
+  {{- end -}}
+ {{- end -}}
+ {{- /* Warn if any of the planes are ephemeral */ -}}
+ {{- if (ne (len $match.ephemeral) 0) -}}
+  {{ $warning := dict "source" "helm" "severity" "warning" -}}
+  {{- $_ := set $warning "msg" (printf "The configuration contains ephemeral planes: %s" (print (uniq $match.ephemeral)) )  -}}
+  {{- $_ := set $ctx "warnings" (append .warnings $warning) -}}
```
jakesmith, line 54: $ctx.warnings

jakesmith (commented):
@shamser - final couple of snagging issue.

helm/hpcc/templates/_warnings.tpl
ghalliday, line 28: trivial: unused variable

helm/hpcc/templates/_warnings.tpl
```diff
@@ -0,0 +1,56 @@
...
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+##############################################################################
+
+*/}}
+{{/*
+Generate a list of warnings
+Pass in dict with root with warnings
```
ghalliday, line 24: root and warnings

ghalliday (commented):
@shamser two trivial final fixes.

helm/hpcc/templates/_warnings.tpl
ghalliday: well spotted

jakesmith (approved):
@shamser - looks good.

ghalliday (approved):
@shamser please squash and I will merge

ghalliday (approved):

PR #16719, opened 2022-11-18
HPCC-28608 LDAP Scope Delete on child scopes fails without message

RussWhitehead:
Throw error if deleting a scope fails, providing the user with a meaningful error message. Also slipped in additional logging of LDAP Config

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/services/ws_access/ws_accessService.cpp
+++ b/esp/services/ws_access/ws_accessService.cpp
@@ -2070,6 +2070,7 @@ bool Cws_accessEx::onResourceDelete(IEspContext &context, IEspResourceDeleteRequ
             resp.setPrefix(req.getPrefix());
         }
         SecResourceType rtype = str2type(basednReq->getRtype());
+        Owned<IMultiException> me = MakeMultiException("onResourceDelete");
         try
         {
             for(unsigned i = 0; i < names.length(); i++)
@@ -2087,23 +2088,26 @@ bool Cws_accessEx::onResourceDelete(IEspContext &context, IEspResourceDeleteRequ
                 if(prefix && *prefix)
                     namebuf.insert(0, prefix);
 
-                secmgr->deleteResource(rtype, namebuf.str(), basednReq->getBasedn(), context.querySecureContext());
+                try
+                {
+                    secmgr->deleteResource(rtype, namebuf.str(), basednReq->getBasedn(), context.querySecureContext());
+                }
+                catch(IException* e)
+                {
+                    me->append(*e);
+                }
             }
         }
-        catch(IException* e)
-        {
-            StringBuffer emsg;
-            e->errorMessage(emsg);
-            resp.setRetcode(e->errorCode());
-            resp.setRetmsg(emsg.str());
-            return false;
-        }
+
         catch(...)
         {
             resp.setRetcode(-1);
             resp.setRetmsg("Unknown error");
             return false;
         }
+
+        if(me->ordinality())
+            throw me.getLink();
 
         resp.setRetcode(0);
     }
--- a/system/security/LdapSecurity/ldapconnection.cpp
+++ b/system/security/LdapSecurity/ldapconnection.cpp
@@ -290,6 +290,9 @@ private:
     void debugPrintout()
     {
         DBGLOG("@@LDAP CONFIG--------------");
+        StringBuffer sb;
+        getLdapHost(sb);
+        DBGLOG("LDAP host = %s", sb.str());
         DBGLOG("m_adminGroupDN = %s", m_adminGroupDN.str());
         DBGLOG("m_protocol = %s", m_protocol.str());
         DBGLOG("m_basedn = %s", m_basedn.str());
@@ -4565,8 +4568,7 @@ public:
 
         if ( rc != LDAP_SUCCESS )
         {
-            DBGLOG("error deleting %s: %s", dn.str(), ldap_err2string(rc));
-            //throw MakeStringException(-1, "error deleting %s: %s", dn.str(), ldap_err2string(rc));
+            throw MakeStringException(-1, "Error deleting %s: %s", dn.str(), ldap_err2string(rc));
         }
         
     }


wangkx (commented):
Is this the correct way to notify the caller about the error? When the caller requests to delete multiple multiple resources or delete a user with multiple resources, should we stop the whole process in the middle? Or we may collect the error(s) into a buffer and throw them at the end of the process?

esp/services/ws_access/ws_accessService.cpp
```diff
@@ -2087,24 +2089,37 @@ bool Cws_accessEx::onResourceDelete(IEspContext &context, IEspResourceDeleteRequ
                 if(prefix && *prefix)
                     namebuf.insert(0, prefix);
 
-                secmgr->deleteResource(rtype, namebuf.str(), basednReq->getBasedn(), context.querySecureContext());
+                try
+                {
+                    secmgr->deleteResource(rtype, namebuf.str(), basednReq->getBasedn(), context.querySecureContext());
+                }
+
```
wangkx, line 2096: Minor: The newline should be removed.

wangkx (approved):
@RussWhitehead looks fine now.

RussWhitehead:
@richardkchapman @ghalliday Please merge

esp/services/ws_access/ws_accessService.cpp
```diff
@@ -2087,24 +2089,36 @@ bool Cws_accessEx::onResourceDelete(IEspContext &context, IEspResourceDeleteRequ
...
         {
             resp.setRetcode(-1);
             resp.setRetmsg("Unknown error");
             return false;
         }
 
+        if (!errMsgs.isEmpty())
+        {
+            resp.setRetcode(iLastRC);
+            errMsgs.append("\n");
```
ghalliday, line 2117: @afishbeck  Is this the correct way to return an exception?  From what I can see the callers catch exceptions that are thrown, and then serialize it in different ways depending on the request format.

RussWhitehead: @afishbeck @ghalliday The reason it was done this way is that I don't want to abort the operation if any single "delete" fails, instead I want to proceed with the remaining ones and report back all the ones that failed

ghalliday: yes, that makes sense.  The question was more whether this code here, once the failures have been collected should be thrown as an exception.  (It is a question from ignorance since I don't know what the general pattern is for these services.)

RussWhitehead: @ghalliday @afishbeck It used to throw, so I took your recommendation and, if there are errors,  throw an exception after all deletes are attempted

ghalliday (approved):
Looks good.  Will merge once @afishbeck approves

esp/services/ws_access/ws_accessService.cpp
afishbeck: @RussWhitehead @ghalliday sorry it just occurred to me that the exception being thrown when there are multiple is an IMultiException.  In the same place where you catch the exception and append to your error string you should be appending to an IMultiException.  Then throwing that at the end.

The exceptions returned in the response is an array.

esp/services/ws_access/ws_accessService.cpp
```diff
@@ -2087,24 +2089,36 @@ bool Cws_accessEx::onResourceDelete(IEspContext &context, IEspResourceDeleteRequ
...
+                try
+                {
+                    secmgr->deleteResource(rtype, namebuf.str(), basednReq->getBasedn(), context.querySecureContext());
+                }
+                catch(IException* e)
+                {
+                    StringBuffer emsg;
+                    e->errorMessage(emsg);
+                    errMsgs.appendf("\n%d %s", e->errorCode(), emsg.str());//accumulate error messages/return codes
+                    iLastRC = e->errorCode();
```
afishbeck, line 2101: I would also consider when there are more than one exceptions having a new return code that means "multiple errors".  Then the caller will know they can iterate through them all.

afishbeck (changes requested):

afishbeck (changes requested):
Some suggested changes regarding throwing multiple exceptions.

esp/services/ws_access/ws_accessService.cpp
afishbeck: Here's a fairly simple example:

https://github.com/hpcc-systems/HPCC-Platform/blob/master/esp/services/ws_workunits/ws_workunitsService.cpp#L882

esp/services/ws_access/ws_accessService.cpp
```diff
@@ -2087,24 +2088,27 @@ bool Cws_accessEx::onResourceDelete(IEspContext &context, IEspResourceDeleteRequ
...
+                    me->append(*e);
+                }
             }
         }
-        catch(IException* e)
-        {
-            StringBuffer emsg;
-            e->errorMessage(emsg);
-            resp.setRetcode(e->errorCode());
-            resp.setRetmsg(emsg.str());
```
afishbeck, line 2098: The change isn't strictly backward compatible if a client is coded to look for this retcode and msg, but discussed with Russ and he thinks it's ok.  This function could return exceptions before, just not in this case, so clients should look for them anyway.

afishbeck (approved):
Extra newline, and discussion about backward compatibility but overall looks fine.

RussWhitehead:
@ghalliday Please merge

esp/services/ws_access/ws_accessService.cpp
```diff
@@ -2087,24 +2088,27 @@ bool Cws_accessEx::onResourceDelete(IEspContext &context, IEspResourceDeleteRequ
...
+
         catch(...)
         {
             resp.setRetcode(-1);
             resp.setRetmsg("Unknown error");
             return false;
         }
 
+        if(me->ordinality())
+            throw me.getLink();
```
ghalliday, line 2110: trivial:  throw me.getClear() is slightly more efficient.  Does not matter in this context.

ghalliday (approved):

PR #16721, opened 2022-11-21
HPCC-28618 Refactor jhtree to avoid fetching entire row when only key fields needed

richardkchapman:
Refactor jhtree code to only use keyed portion of buffer when searching, and fetch full record only when required for delivery, in preparation for potentially compressing the keyed and unkeyed portions of the record separately.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Cloud-compatibility
  - [x] Premature optimization

## Smoketest:

## Testing:

Regression suite

### DIFF
--- a/system/jhtree/ctfile.cpp
+++ b/system/jhtree/ctfile.cpp
@@ -228,9 +228,7 @@ CNodeBase::CNodeBase()
 {
     keyHdr = NULL;
     fpos = 0;
-    keyLen = 0;
     keyType = 0;
-    keyCompareLen = 0;
     isVariable = false;
 }
 
@@ -238,8 +236,6 @@ void CNodeBase::load(CKeyHdr *_keyHdr, offset_t _fpos)
 {
     _keyHdr->Link();
     keyHdr = _keyHdr;
-    keyLen = keyHdr->getMaxKeyLength();
-    keyCompareLen = keyHdr->getNodeKeyLength();
     keyType = keyHdr->getKeyType();
     memset(&hdr, 0, sizeof(hdr));
     fpos = _fpos;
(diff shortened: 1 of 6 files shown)


system/jhtree/ctfile.cpp
```diff
@@ -810,6 +821,33 @@ bool CJHTreeNode::contains(const char *src) const
...
 
+void CJHTreeNode::dump(FILE *out, int length, unsigned rowCount, bool raw) const
+{
+    if (!raw)
+        fprintf(out, "Node dump: fpos(%" I64F "d) type %s\n", getFpos(), getNodeTypeName());
+    if (rowCount==0 || rowCount > getNumKeys())
+        rowCount = getNumKeys();
+    for (unsigned int i=0; i<rowCount; i++)
+    {
+        char *dst = (char *) alloca(getKeyLen()+50);
```
ghalliday, line 832: +50?  
Also can alloca in a loop be dangerous?  Safer to dynamically allocate e.g. with MemoryAttr or similar.

system/jhtree/ctfile.hpp
```diff
@@ -229,13 +225,16 @@ class jhtree_decl CNodeBase : public CWritableKeyNode
...
 {
 protected:
+    size32_t keyLen = 0;
+    size32_t keyCompareLen = 0;
     size32_t keyRecLen;
     char *keyBuf;
 
     void unpack(const void *node, bool needCopy);
     unsigned __int64 firstSequence;
     size32_t expandedSize;
```
ghalliday, line 235: unrelated.  Moving this after keyRecLen would save 8 bytes packing.

ghalliday (commented):
@richardkchapman didn't spot much on a fisrst scan.  
Do you need two separate buffers though, or can you use a single buffer with the flag to indicate it is full?
Then you can call a new function appendPayloadValue(<idx>, buffer), and avoid re-copying the keyed portion.  For some indexes that are (keyed=>filepos) that would avoid the bulk of the work.  It would also avoid extra work in a new format where keyed and non keyed are compressed separately.

richardkchapman:
I could probably combine the buffers. I didn't in the first pass because it made it easier to be sure I was not reading/writing off the end of the keyed bit (using valgrind), but also because I'm not sure whether the "full" buffer is really needed at all, or whether I will eventually be able to move the data directly into the user-supplied destination. At present it seems the only real reason for the "full" buffer is to have something that has moved the fpos to the end.

ghalliday (approved):

PR #16725, opened 2022-11-22
HPCC-28594 Publish extended statistics for indexes in getFileDetails()

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:


ghalliday:
@jakesmith please can you check the dfu logic (including updating attributes).
@mckellyln does this provide the information that we need to know to understand how indexes will work.  (I could easily return nodeSize and keySize which might be useful.)
@GordonSmith any comments on the format.  See the linked issue for ideas on displaying in eclwatch

ghalliday:
(Linked eclwatch issue is HPCC-28624)

common/thorhelper/thorfile.cpp
```diff
@@ -100,3 +101,257 @@ IHThorDiskReadArg * createWorkUnitReadArg(const char * filename, IHThorWorkunitR
...
     return new DiskWorkUnitReadArg(filename, wuRead);
 }
+
+
+
+//-----------------------------------------------------------------------------
+
+#define MAX_FILE_READ_FAIL_COUNT 3
+
+IKeyIndex *openKeyFile(IDistributedFilePart & keyFile)
```
ghalliday, line 111: was already duplicated, moved to a common location (and used again)

common/thorhelper/thorfile.cpp
```diff
@@ -100,3 +101,257 @@ IHThorDiskReadArg * createWorkUnitReadArg(const char * filename, IHThorWorkunitR
...
+            size_t nodeSize = index->getNodeSize();
+
+            //FUTURE: When file information is returned from an esp service, will this be updated?
+            file->lockProperties(defaultTimeout);
+            file->queryAttributes().setPropInt("@nodeSize", nodeSize);
+            file->queryAttributes().setPropInt("@keySize", keySize);
+            file->unlockProperties();
+        }
+
+        if (!attrs.hasProp("@numLeafNodes"))
```
jakesmith, line 210: question: should somewhere be setting this yet (e.g. in the case of no blob nodes) ?

jakesmith (commented):
@ghalliday - looks good, 1 question

common/thorhelper/thorfile.cpp
ghalliday: It is set by hthor and roxie when the index is built - there is a jira to add it to thor.  I think it is better not to set it here - it can deduce most of it from offsetBranches and it is useful to know it is an approximation.

common/thorhelper/thorfile.cpp
```diff
@@ -100,3 +101,257 @@ IHThorDiskReadArg * createWorkUnitReadArg(const char * filename, IHThorWorkunitR
...
+    offset_t uncompressedSize = attrs.getPropInt64("@size");
+    //size only corresponds to the uncompressed size if both size and compressedSize are filled in.
+    if ((uncompressedSize != 0) && (compressedSize != 0))
+        result.sizeOriginalData = uncompressedSize;
+
+    //The following will depend on the compression format - e.g. if compressed searching is implemented
+    result.sizeMemoryBranches = result.sizeOriginalBranches;
+
+    //Could calculate from the size of the raw record (with blobids) multiplied by the count.
+    if ((result.numBlobNodes == 0) && result.sizeOriginalData)
```
ghalliday, line 293: @jakesmith a question.  Currently the value filled in in @size is the size of the rows sent to the builder, but this doesn't include the size of any blobs.  That means this calculation is wrong.  There are two potential fixes:

* Always go through the 1st branch - since blobs aren't included they don't affect the in memory leaf size
* Add the size of any blobs to the @size reported for a file.

Making the second change means that the compression ratio will be more accurate.  We could also record @sizeOriginalBlobs in the meta, but is that overkill?

jakesmith: I can't think of any good reason not to include the blob size in @size too, and to record sizeOriginalBlobs, which I think could be useful info. too - other than effort.

GordonSmith (approved):

ghalliday:
@jakesmith I have pushed a couple of changes
i) to correct the in-memory estimate to take account of blob sizes which will be recorded in a separate PR
ii) to avoid forcing gathering the index information by default - I was concerned it might have an impact on some 400 way indexes. etc.

common/thorhelper/thorfile.cpp
```diff
@@ -289,11 +289,18 @@ static void gatherDerivedIndexInformation(DerivedIndexInformation & result, IDis
     //The following will depend on the compression format - e.g. if compressed searching is implemented
     result.sizeMemoryBranches = result.sizeOriginalBranches;
 
-    //Could calculate from the size of the raw record (with blobids) multiplied by the count.
-    if ((result.numBlobNodes == 0) && result.sizeOriginalData)
-        result.sizeMemoryLeafs = result.sizeOriginalData;
-    else if (result.sizeOriginalData)
-        result.sizeMemoryLeafs = (offset_t)((double)result.numLeafNodes * result.sizeOriginalData)/(result.numLeafNodes + result.numBranchNodes);
+    //NOTE: sizeOriginalData now includes the blob sizes that areremoved before passing to the builder
```
jakesmith, line 292: typo: areremoved

jakesmith (approved):
@ghalliday - 1 trivial comment typo', otherwise looks good.

common/thorhelper/thorfile.cpp
```diff
@@ -289,11 +289,18 @@ static void gatherDerivedIndexInformation(DerivedIndexInformation & result, IDis
...
+    //      if the original blob size is recorded then use it, otherwise estimate it
+    if (result.sizeOriginalData)
+    {
+        offset_t originalBlobSize = attrs.getPropInt64("@originalBlobSize");
+        if (result.numBlobNodes == 0)
+            result.sizeMemoryLeafs = result.sizeOriginalData;
+        else if (originalBlobSize)
+            result.sizeMemoryLeafs = result.sizeOriginalData - originalBlobSize;
```
richardkchapman, line 300: Not new, but shouldn't it be sizeMemoryLeaves ?

ghalliday: It should be, most places I tried to avoid the plural because it looks a bit ugly.  I'll change though.

system/jhtree/jhtree.cpp
```diff
@@ -1322,6 +1322,25 @@ offset_t CKeyIndex::queryMetadataHead()
...
+
+    Linked<CJHTreeNode> cur = rootNode;
+    for (unsigned nextBranch = 2; ; nextBranch++)
+    {
+        offset_t branchOffset = cur->getFPosAt(0);
+        if (nextBranch == branchDepth)
+            return branchOffset;
+        IContextLogger * ctx = nullptr;
+        cur.setown(getNode(branchOffset, NodeBranch, ctx));
+    }
```
mckellyln, line 1341: Could we ever fall off the end here ?

common/thorhelper/thorfile.cpp
```diff
@@ -100,3 +101,264 @@ IHThorDiskReadArg * createWorkUnitReadArg(const char * filename, IHThorWorkunitR
...
+
+    //These may or may not be known
+    if (other.sizeOriginalData)
+        result.sizeOriginalData += other.sizeOriginalData;
+    else
+        result.sizeOriginalData = 0;
+    if (other.sizeMemoryLeaves)
+        result.sizeMemoryLeaves += other.sizeMemoryLeaves;
+    else
+        result.sizeMemoryLeaves = 0;
```
mckellyln, line 363: why zero these out when other is zero ?

mckellyln (approved):
Looks good, but a lot for me to consider / understand.
Just had some basic questions.

common/thorhelper/thorfile.cpp
ghalliday: because the value is unknown so the aggregated value can not be used for a calculation

PR #16734, opened 2022-11-29
HPCC-28644 Refactor in preparation for new index formats

ghalliday:
Various changes to the structure of jhtree classes to allow future expansion for new node types and compression options.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/jhtree/ctfile.cpp
+++ b/system/jhtree/ctfile.cpp
@@ -24,12 +24,17 @@
 #endif
 
 #include "jmisc.hpp"
+#include "jset.hpp"
 #include "hlzw.h"
 
 #include "ctfile.hpp"
 #include "jstats.h"
 
-void SwapBigEndian(KeyHdr &hdr)
+//#define TIME_NODE_SEARCH            // Useful for performance testing, but generally should be off
+
+//---------------------------------------------------------------------------------------------------------------------
+
+void SwapBigEndian(KeyHdr &hdr, bool read)
 {
     _WINREV(hdr.phyrec);
     _WINREV(hdr.delstk);
@@ -78,11 +83,15 @@ void SwapBigEndian(KeyHdr &hdr)
     _WINREV(hdr.fileSize);
     _WINREV(hdr.nodeKeyLength);
     _WINREV(hdr.version);
+    _WINREV(hdr.keyFlags);
     _WINREV(hdr.blobHead);
     _WINREV(hdr.metadataHead);
     _WINREV(hdr.bloomHead);
     _WINREV(hdr.partitionFieldMask);
     _WINREV(hdr.firstLeaf);
+    //Clone the old flags byte into the new flags word for backwards compatibility
+    if (read)
+        hdr.keyFlags |= (byte)hdr.oldktype;
 }
 
 inline void SwapBigEndian(NodeHdr &hdr)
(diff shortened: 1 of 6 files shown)


system/jlib/jlzw.cpp
```diff
@@ -315,6 +315,20 @@ void CLZWCompressor::open(void *buf,size32_t max)
 
 
 size32_t CLZWCompressor::write(const void *buf,size32_t buflen)
+{
+    return doWrite((size32_t)-1, buf, buflen);
+}
+
+size32_t CLZWCompressor::limitWrite(size32_t limit, const void *buf,size32_t buflen)
```
ghalliday, line 322: This may not be the best way to implement this.  Better may be to dynamically call something to adjust the limit before calling write.  That would avoid the complication of maxlen potentially changing in doWrite.  Something to discuss and revisit.

system/jhtree/ctfile.hpp
```diff
@@ -92,7 +98,7 @@ struct __declspec(novtable) jhtree_decl KeyHdr
     __int64 fileSize; /* fileSize - was once used in the bias calculation e0x */
     short nodeKeyLength; /* key length in intermediate level nodes e8x */
     unsigned short version; /* build version - to be updated if key format changes    eax*/
-    short unused[2]; /* unused ecx */
+    unsigned keyFlags; /* extended version of ktype ecx */
```
richardkchapman, line 101: This structure probably should be defined with pack(1). Without it, I had to think to convince myself that this was ok (also we should really be using types with determinate sizes, rather than unsigned whose size is implementation-defined).

system/jhtree/ctfile.hpp
```diff
@@ -278,7 +309,7 @@ class CJHVarTreeNode : public CJHTreeNode
     virtual int compareValueAt(const char *src, unsigned int index) const;
 };
 
-class CJHRowCompressedNode : public CJHTreeNode
+class CJHRowCompressedNode : public CJHExpandedTreeNode
```
richardkchapman, line 312: I thought rowCompressed could be read without expanding

ghalliday: Yes it can.  For the moment I would leave it like this, but it probably would make sense to refactor it later.

system/jhtree/ctfile.hpp
```diff
@@ -162,7 +169,7 @@ class jhtree_decl CKeyHdr : public CWritableKeyNode
     {
         return (hdr.nodeKeyLength != -1);
     }
-    inline char getKeyType() { return hdr.ktype; }
+    inline unsigned getKeyType() { return hdr.keyFlags; }
```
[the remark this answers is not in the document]
ghalliday: It doesn't because SwapBigEndian ensures it is filled in.  I could leave the old bits where they are and only store new bits in the new flags.  e.g. always OR them together in this function.

system/jhtree/ctfile.cpp
```diff
@@ -78,11 +83,15 @@ void SwapBigEndian(KeyHdr &hdr)
...
     _WINREV(hdr.version);
+    _WINREV(hdr.keyFlags);
     _WINREV(hdr.blobHead);
     _WINREV(hdr.metadataHead);
     _WINREV(hdr.bloomHead);
     _WINREV(hdr.partitionFieldMask);
     _WINREV(hdr.firstLeaf);
+    //Clone the old flags byte into the new flags word for backwards compatibility
+    if (read)
+        hdr.keyFlags |= (byte)hdr.oldktype;
```
richardkchapman, line 94: Not convinced this works - won't it be 0xffff in an old key.

Also rather unexpected to find this code in a function called swapBigEndian... 

A comment on the field definition that it gets updated on load to account foc compatibility issues would be helpful

system/jhtree/ctfile.cpp
```diff
@@ -24,12 +24,17 @@
...
 
 #include "ctfile.hpp"
 #include "jstats.h"
 
-void SwapBigEndian(KeyHdr &hdr)
+//#define TIME_NODE_SEARCH            // Useful for performance testing, but generally should be off
+
+//---------------------------------------------------------------------------------------------------------------------
+
+void SwapBigEndian(KeyHdr &hdr, bool read)
```
richardkchapman, line 37: i don't like having swapBigEndian also do compatibility patching. Better to leave if without the parameter and do the compatibility patch elsewhere (or don't do it at all - not really sure why it is needed, given we could just add new node types)

richardkchapman (changes requested):
Some concerns about compatibility with old keys

ghalliday:
Superseded by Richard's refactoring.

PR #16735, opened 2022-11-29
HPCC-28615 Helm warning: cost values misconfigured

shamser:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/helm/hpcc/templates/_helpers.tpl
+++ b/helm/hpcc/templates/_helpers.tpl
@@ -2058,11 +2058,11 @@ spec:
 {{- end -}}
 {{- end -}}
 
-{{- define "hpcc.generateWarnings" -}}
 {{/*
 Generate a list of warnings
 Pass in dict with root
 */}}
+{{- define "hpcc.generateWarnings" -}}
 {{- $ctx := dict "warnings" list "root" .root -}}
 {{- include "hpcc.getWarnings" $ctx -}}
 {{- if $ctx.warnings }}
@@ -2075,14 +2075,26 @@ warnings:
 {{- end -}}
 {{- end -}}
 
-{{- define "hpcc.printWarnings" -}}
 {{/*
 Print warnings
 Pass in dict with root
 */}}
+{{- define "hpcc.printWarnings" -}}
 {{- $ctx := dict "warnings" list "root" .root -}}
 {{- include "hpcc.getWarnings" $ctx -}}
 {{- range $warning := $ctx.warnings }}
  {{ printf "**** %s: %s ****" (upper $warning.severity) $warning.msg }}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Returns true if resources exist
+Pass in dict with resources object
+*/}}
+{{- define "hpcc.hasResources" -}}
+{{- if .resources -}}
+ {{- if and .resources.cpu .resources.memory -}}
+  true
+ {{- end -}}
 {{- end -}}
 {{- end -}}
--- a/helm/hpcc/templates/_warnings.tpl
+++ b/helm/hpcc/templates/_warnings.tpl
@@ -52,4 +52,78 @@ Pass in dict with root and warnings
   {{- $_ := set $warning "msg" (printf "The configuration contains ephemeral planes: %s" (print (uniq $match.ephemeral)) )  -}}
   {{- $_ := set $ctx "warnings" (append $ctx.warnings $warning) -}}
  {{- end -}}
+ {{- /* Warn if resources not defined for any cost component or default cpuRate being used */ -}}
+ {{- if eq .root.Values.global.cost.perCpu 0.0565000000001 -}}
+  {{- $_ := set $ctx "usingDefaultCpuCost" "true" -}}
+ {{- end -}}
+ {{- $_ := set $ctx "missingResources" list -}}
+ {{- $_ := set $ctx "defaultCpuRateComponents" list -}}
+ {{- range $component := .root.Values.eclagent -}}
+  {{- if not $component.disabled -}}
+   {{- $hasResources := include "hpcc.hasResources" (dict "resources" $component.resources) -}}
+   {{- if not $hasResources -}}
+    {{- $_ := set $ctx "missingResources" (append $ctx.missingResources $component.name) -}}
+   {{- end -}}
+   {{- if and $ctx.usingDefaultCpuCost (not $component.cost) -}}
+    {{- $_ := set $ctx "defaultCpuRateComponents" (append $ctx.defaultCpuRateComponents $component.name) -}}
+   {{- end -}}
+  {{- end -}}
+ {{- end -}}
+ {{- range $component := .root.Values.eclccserver -}}
+  {{- if not $component.disabled -}}
+   {{- $hasResources := include "hpcc.hasResources" (dict "resources" $component.resources) -}}
+   {{- if not $hasResources -}}
+    {{- $_ := set $ctx "missingResources" (append $ctx.missingResources $component.name) -}}
+   {{- end -}}
+   {{- if and $ctx.usingDefaultCpuCost (not $component.cost) -}}
+    {{- $_ := set $ctx "defaultCpuRateComponents" (append $ctx.defaultCpuRateComponents $component.name) -}}
+   {{- end -}}
+  {{- end -}}
+ {{- end -}}
+ {{- range $component := .root.Values.thor -}}
+  {{- if not $component.disabled -}}
+   {{- $hasResources1 := include "hpcc.hasResources" (dict "resources" $component.managerResources) -}}
+   {{- $hasResources2 := include "hpcc.hasResources" (dict "resources" $component.workerResources) -}}
+   {{- $hasResources3 := include "hpcc.hasResources" (dict "resources" $component.eclAgentResources) -}}
+   {{- if or (not $hasResources1) (or (not $hasResources2 ) (not $hasResources3)) -}}
+    {{- $_ := set $ctx "missingResources" (append $ctx.missingResources $component.name) -}}
+   {{- end -}}
+   {{- if and $ctx.usingDefaultCpuCost (not $component.cost) -}}
+    {{- $_ := set $ctx "defaultCpuRateComponents" (append $ctx.defaultCpuRateComponents $component.name) -}}
+   {{- end -}}
+  {{- end -}}
+ {{- end -}}
+ {{- if $ctx.missingResources -}}
+  {{- $warning := dict "source" "helm" "severity" "warning" -}}
+  {{- $_ := set $warning "msg" (printf "Cost calculation requires resources to be provided for %s: %s" ((len $ctx.missingResources)| plural "component" "components") ($ctx.missingResources|toStrings)) -}}
+  {{- $_ := set $ctx "warnings" (append $ctx.warnings $warning) -}}
+ {{- end -}}
+ {{- if $ctx.defaultCpuRateComponents -}}
+  {{ $warning := dict "source" "helm" "severity" "warning" -}}
+  {{- $_ := set $warning "msg" (printf "Default cpu cost rate is being used for %s: %s" ((len $ctx.defaultCpuRateComponents)| plural "component" "components") ($ctx.defaultCpuRateComponents|toStrings)) -}}
+  {{- $_ := set $ctx "warnings" (append $ctx.warnings $warning) -}}
+ {{- end -}}
+ {{- /* Warn if any storage planes uses defaults cost values */ -}}
+ {{- with (.root.Values.global.cost) -}}
+  {{- if and (eq .storageWrites 0.0500000000001) (and (eq .storageAtRest 0.0208000000001) (eq .storageReads 0.00400000000001)) -}}
+   {{- $_ := set $ctx "usingDefaultStorageCosts" "true" -}}
+  {{- end -}}
+ {{- end -}}
+ {{- if $ctx.usingDefaultStorageCosts -}}
+  {{- $_ := set $ctx "planesWithDefaultCosts" list -}}
+  {{- range $storagePlane := .root.Values.storage.planes -}}
+   {{- if not $storagePlane.disabled -}}
+    {{- if not $storagePlane.cost -}}
+     {{- $_ := set $ctx "planesWithDefaultCosts" (append $ctx.planesWithDefaultCosts $storagePlane.name) -}}
+    {{- else if or (not $storagePlane.cost.storageAtRest) (or (not $storagePlane.cost.storageReads) (not $storagePlane.cost.storageWrites)) -}}
+     {{- $_ := set $ctx "planesWithDefaultCosts" (append $ctx.planesWithDefaultCosts $storagePlane.name) -}}
+    {{- end -}}
+   {{- end -}}
+  {{- end -}}
+ {{- end -}}
+ {{- if $ctx.planesWithDefaultCosts -}}
+   {{- $warning := dict "source" "helm" "severity" "warning" -}}
+   {{- $_ := set $warning "msg" (printf "Default cost parameters are being used for the storage %s: %s" ((len $ctx.planesWithDefaultCosts)|plural "plane" "planes") ($ctx.planesWithDefaultCosts|toStrings)) -}}
+   {{- $_ := set $ctx "warnings" (append $ctx.warnings $warning) -}}
+ {{- end -}}
 {{- end -}}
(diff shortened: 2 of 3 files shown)


helm/hpcc/templates/_helpers.tpl
```diff
@@ -2086,3 +2086,17 @@ Pass in dict with root
  {{ printf "**** %s: %s ****" (upper $warning.severity) $warning.msg }}
 {{- end -}}
 {{- end -}}
+
+{{- define "hpcc.checkResources" -}}
+{{/*
+Returns true if resources provided, otherwise return false
```
ghalliday, line 2092: This isn't really true.  It returns blank if true, otherwise false.  More normal is to return "true" if true, and blank if false.  (E.g. see "hpcc.hasPlaneForCategory"
Then the result can be tested with

   {{- $hasResources := include "hpcc.checkResources" (dict "resources" $component.resources) -}}
   {{- if not $hasResources -}}

helm/hpcc/templates/_warnings.tpl
```diff
@@ -52,4 +52,79 @@ Pass in dict with root and warnings
...
+ {{- end -}}
+ {{- $_ := set $ctx "missingResources" list -}}
+ {{- $_ := set $ctx "defaultCpuRateComponents" list -}}
+ {{- range $component := .root.Values.eclagent -}}
+  {{- if and (not $component.disabled) (eq $component.type "hthor" ) -}}
+   {{- $hasResources := include "hpcc.checkResources" (dict "resources" $component.resources) -}}
+   {{- if eq $hasResources "false" -}}
+    {{- $_ := set $ctx "missingResources" (append $ctx.missingResources $component.name) -}}
+   {{- end -}}
+   {{- if not $component.cost -}}
```
ghalliday, line 67: Should this only be reported if not usingDefaultCpuCost?

helm/hpcc/templates/_warnings.tpl
```diff
@@ -52,4 +52,79 @@ Pass in dict with root and warnings
...
  {{- end -}}
+ {{- /* Warn if resources not defined for any cost component or default cpuRate being used */ -}}
+ {{- if eq .root.Values.global.cost.perCpu 0.0565000000001 -}}
+  {{- $_ := set $ctx "usingDefaultCpuCost" "true" -}}
+ {{- end -}}
+ {{- $_ := set $ctx "missingResources" list -}}
+ {{- $_ := set $ctx "defaultCpuRateComponents" list -}}
+ {{- range $component := .root.Values.eclagent -}}
+  {{- if and (not $component.disabled) (eq $component.type "hthor" ) -}}
+   {{- $hasResources := include "hpcc.checkResources" (dict "resources" $component.resources) -}}
```
ghalliday, line 63: Could pass $component instead of dict "resources" $component.resources.  But this is consistent with the more complex cases, so I think it make sense.

helm/hpcc/templates/_warnings.tpl
```diff
@@ -52,4 +52,79 @@ Pass in dict with root and warnings
...
+  {{- $_ := set $warning "msg" (printf "Default cpu cost rate is being used for %s: %s" ((len $ctx.defaultCpuRateComponents)| plural "Component" "Components") ($ctx.defaultCpuRateComponents|toStrings)) -}}
+  {{- $_ := set $ctx "warnings" (append $ctx.warnings $warning) -}}
+ {{- end -}}
+ {{- /* Warn if any storage planes uses defaults cost values */ -}}
+ {{- with (.root.Values.global.cost) -}}
+  {{- if and (eq .storageWrites 0.0500000000001) (and (eq .storageAtRest 0.0208000000001) (eq .storageReads 0.00400000000001)) -}}
+   {{- $_ := set $ctx "usingDefaultStorageCosts" "true" -}}
+  {{- end -}}
+ {{- end -}}
+ {{- $usingDefaultCost := $ctx.usingDefaultStorageCosts |default "false" -}}
```
ghalliday, line 112: Is this necessary.  Could the following test be if $ctx.usingDefautlStorageCosts?

helm/hpcc/templates/_warnings.tpl
```diff
@@ -52,4 +52,79 @@ Pass in dict with root and warnings
...
+  {{- end -}}
+ {{- end -}}
+ {{- if $ctx.missingResources -}}
+  {{ $warning := dict "source" "helm" "severity" "warning" -}}
+  {{- $_ := set $warning "msg" (printf "Cost calculation requires resources to be provided for %s: %s" ((len $ctx.missingResources)| plural "component" "components") ($ctx.missingResources|toStrings)) -}}
+  {{- $_ := set $ctx "warnings" (append $ctx.warnings $warning) -}}
+ {{- end -}}
+ {{- if $ctx.defaultCpuRateComponents -}}
+  {{ $warning := dict "source" "helm" "severity" "warning" -}}
+  {{- $_ := set $warning "msg" (printf "Default cpu cost rate is being used for %s: %s" ((len $ctx.defaultCpuRateComponents)| plural "Component" "Components") ($ctx.defaultCpuRateComponents|toStrings)) -}}
```
ghalliday, line 103: ultra picky: Component(s) are capitalized, but plane(s) below starts with lower case.

helm/hpcc/templates/_warnings.tpl
```diff
@@ -52,4 +52,79 @@ Pass in dict with root and warnings
...
+   {{- $hasResources2 := include "hpcc.checkResources" (dict "resources" $component.workerResources) -}}
+   {{- $hasResources3 := include "hpcc.checkResources" (dict "resources" $component.eclAgentResources) -}}
+   {{- if or (eq $hasResources1 "false") (or (eq $hasResources1 "false") (eq $hasResources1 "false")) -}}
+    {{- $_ := set $ctx "missingResources" (append $ctx.missingResources $component.name) -}}
+   {{- end -}}
+   {{- if not $component.cost -}}
+    {{- $_ := set $ctx "defaultCpuRateComponents" (append $ctx.defaultCpuRateComponents $component.name) -}}
+   {{- end -}}
+  {{- end -}}
+ {{- end -}}
```
ghalliday, line 95: also applies to dali, roxie etc.  Is the plan to add those in a later PR?

ghalliday (commented):
Looks like a good addition.  A few comments

helm/hpcc/templates/_warnings.tpl
shamser: Yes, I did it like this because there are components like Thor that have multiple resource objects.

helm/hpcc/templates/_warnings.tpl
shamser: Yes, they'll be checked in HPCC-28614.  This one is only to check components that affects WU costs.

helm/hpcc/templates/_warnings.tpl
shamser: It does add it to the list but it never generates a warning if not usingDefaultCpuCost.  I could check usingDefaultCpuCost here instead of in line 101.  What do you think?

helm/hpcc/templates/_warnings.tpl
shamser: No, it's not required. Not sure why I did that.

helm/hpcc/templates/_warnings.tpl
```diff
@@ -52,4 +52,78 @@ Pass in dict with root and warnings
...
   {{- $_ := set $ctx "warnings" (append $ctx.warnings $warning) -}}
  {{- end -}}
+ {{- /* Warn if resources not defined for any cost component or default cpuRate being used */ -}}
+ {{- if eq .root.Values.global.cost.perCpu 0.0565000000001 -}}
+  {{- $_ := set $ctx "usingDefaultCpuCost" "true" -}}
+ {{- end -}}
+ {{- $_ := set $ctx "missingResources" list -}}
+ {{- $_ := set $ctx "defaultCpuRateComponents" list -}}
+ {{- range $component := .root.Values.eclagent -}}
+  {{- if and (not $component.disabled) (eq $component.type "hthor" ) -}}
```
ghalliday, line 62: Why the check on $component.type?  I think the alternative value is "roxie" which should also check resources.  Is that for a follow on PR?

helm/hpcc/templates/_warnings.tpl
```diff
@@ -52,4 +52,78 @@ Pass in dict with root and warnings
...
+   {{- $hasResources := include "hpcc.hasResources" (dict "resources" $component.resources) -}}
+   {{- if not $hasResources -}}
+    {{- $_ := set $ctx "missingResources" (append $ctx.missingResources $component.name) -}}
+   {{- end -}}
+   {{- if and (not $ctx.usingDefaultCpuCost) (not $component.cost) -}}
```
ghalliday, line 67: First test is the wrong way around...

helm/hpcc/templates/_warnings.tpl
```diff
@@ -52,4 +52,78 @@ Pass in dict with root and warnings
...
+   {{- end -}}
+  {{- end -}}
+ {{- end -}}
+ {{- range $component := .root.Values.eclccserver -}}
+  {{- if not $component.disabled -}}
+   {{- $hasResources := include "hpcc.hasResources" (dict "resources" $component.resources) -}}
+   {{- if not $hasResources -}}
+    {{- $_ := set $ctx "missingResources" (append $ctx.missingResources $component.name) -}}
+   {{- end -}}
+   {{- if and (not $ctx.usingDefaultCpuCost) (not $component.cost) -}}
```
ghalliday, line 78: first test inverted (a result of refactoring I think).  Also thor below

helm/hpcc/templates/_warnings.tpl
```diff
@@ -52,4 +52,78 @@ Pass in dict with root and warnings
...
+   {{- if or (not $hasResources1) (or (not $hasResources2 ) (not $hasResources3)) -}}
+    {{- $_ := set $ctx "missingResources" (append $ctx.missingResources $component.name) -}}
+   {{- end -}}
+   {{- if and (not $ctx.usingDefaultCpuCost) (not $component.cost) -}}
+    {{- $_ := set $ctx "defaultCpuRateComponents" (append $ctx.defaultCpuRateComponents $component.name) -}}
+   {{- end -}}
+  {{- end -}}
+ {{- end -}}
+ {{- if $ctx.missingResources -}}
+  {{ $warning := dict "source" "helm" "severity" "warning" -}}
```
ghalliday, line 97: consistency: missing {{-

ghalliday (commented):
@shamser one problem (x3) and a couple of comments

helm/hpcc/templates/_warnings.tpl
shamser: The reason was that the "roxie" type isn't used in the cost calculation at the moment.  However, I'm going to remove this check as it's unnecessary complication.

ghalliday (approved):
@shamser please squash.  I'll ask Jake to sanity check once squashed.

ghalliday (approved):

ghalliday:
Jake, please sanity check

helm/hpcc/templates/_helpers.tpl
```diff
@@ -2086,3 +2086,15 @@ Pass in dict with root
  {{ printf "**** %s: %s ****" (upper $warning.severity) $warning.msg }}
 {{- end -}}
 {{- end -}}
+
+{{- define "hpcc.hasResources" -}}
+{{/*
```
jakesmith, line 2091: trivial: normally these comments are declared before the function.

it doesn't matter apart from consistency.
But means white space chomping may be more important, i.e. shoul be {{- /* .... */ -}}

helm/hpcc/templates/_warnings.tpl
```diff
@@ -52,4 +52,78 @@ Pass in dict with root and warnings
...
  {{- end -}}
+ {{- /* Warn if resources not defined for any cost component or default cpuRate being used */ -}}
+ {{- if eq .root.Values.global.cost.perCpu 0.0565000000001 -}}
+  {{- $_ := set $ctx "usingDefaultCpuCost" "true" -}}
+ {{- end -}}
+ {{- $_ := set $ctx "missingResources" list -}}
+ {{- $_ := set $ctx "defaultCpuRateComponents" list -}}
+ {{- range $component := .root.Values.eclagent -}}
+  {{- if not $component.disabled -}}
+   {{- $hasResources := include "hpcc.hasResources" (dict "resources" $component.resources) -}}
```
jakesmith, line 63: question: does $component.resources cause a template problem, if the "resources" key is missing? i.e. does it need to be  something like:
```
{{- $hasResources := include "hpcc.hasResources" (dict "resources" ($component.resources | default dict)) -}}
```

(and same for $component.resource references below)

helm/hpcc/templates/_warnings.tpl
```diff
@@ -52,4 +52,78 @@ Pass in dict with root and warnings
...
+    {{- $_ := set $ctx "missingResources" (append $ctx.missingResources $component.name) -}}
+   {{- end -}}
+   {{- if and $ctx.usingDefaultCpuCost (not $component.cost) -}}
+    {{- $_ := set $ctx "defaultCpuRateComponents" (append $ctx.defaultCpuRateComponents $component.name) -}}
+   {{- end -}}
+  {{- end -}}
+ {{- end -}}
+ {{- range $component := .root.Values.thor -}}
+  {{- if not $component.disabled -}}
+   {{- $hasResources1 := include "hpcc.hasResources" (dict "resources" $component.managerResources) -}}
```
jakesmith, line 85: question: same as above, but I'm guessing it doesn't cause problem ?

helm/hpcc/templates/_warnings.tpl
```diff
@@ -52,4 +52,78 @@ Pass in dict with root and warnings
...
+    {{- $_ := set $ctx "missingResources" (append $ctx.missingResources $component.name) -}}
+   {{- end -}}
+   {{- if and $ctx.usingDefaultCpuCost (not $component.cost) -}}
+    {{- $_ := set $ctx "defaultCpuRateComponents" (append $ctx.defaultCpuRateComponents $component.name) -}}
+   {{- end -}}
+  {{- end -}}
+ {{- end -}}
+ {{- if $ctx.missingResources -}}
+  {{- $warning := dict "source" "helm" "severity" "warning" -}}
+  {{- $_ := set $warning "msg" (printf "Cost calculation requires resources to be provided for %s: %s" ((len $ctx.missingResources)| plural "component" "components") ($ctx.missingResources|toStrings)) -}}
```
jakesmith, line 98: ok for now, but I we might want to make the warning more general..
it's a bad (or toy) config in general if resource haven't been defined

jakesmith (commented):
A couple of comments.

From a readibility point of view, getWanings is getting big.
It would be more digestable if it was broken up into separate functions, e.g. could put this new resource checking code it it's own function.

helm/hpcc/templates/_warnings.tpl
shamser: It doesn't cause problems if the resources are missing. I tested this and it works correctly, if resources are missing.

helm/hpcc/templates/_warnings.tpl
shamser: There is a another jira (draft PR: https://github.com/hpcc-systems/HPCC-Platform/pull/16738) to report if resources are missing.   This jira concerns costs only.

shamser:
> re digestable if it was broken up into separate functions, e.g. could put this new resource checking code it it's own function.

The latest cost checking code is a little larger.  However, they are separated into sections using comments and I'm not sure that it is unwieldy.    I'd like to leave it as is for now.  And if it does become too cumbersome, it is something we can do later.  Do you feel strongly that it should be split into functions at this time?

jakesmith:
> > re digestable if it was broken up into separate functions, e.g. could put this new resource checking code it it's own function.
> 
> The latest cost checking code is a little larger. However, they are separated into sections using comments and I'm not sure that it is unwieldy. I'd like to leave it as is for now. And if it does become too cumbersome, it is something we can do later. Do you feel strongly that it should be split into functions at this time? @jakesmith

it's fine as is, I don't have strong feelings about. I'd say the next amendment/addition would be a good time to break it up

helm/hpcc/templates/_warnings.tpl
jakesmith: ok, when missing (as per our default values.yaml), I wonder what $component.resources resolves to.

jakesmith (approved):
Looks good.

ghalliday:
@shamser please squash

ghalliday (approved):

PR #16742, opened 2022-11-30
HPCC-28207 File services failing to find host in drop zone plane

jakesmith:
The implementation of getDropZoneHost was wrong, it was expecting the plane to have a 'hosts' list, whereas it should have a @hostGroup that points to a list of hosts.
I think at one early stage planes did have a hosts list not a reference, but it was changed, so this code was probably never updated to reflect that.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dali/base/dautils.cpp
+++ b/dali/base/dautils.cpp
@@ -47,6 +47,53 @@
 #define MIN_REDIRECTION_LOAD_INTERVAL 1000
 
 
+bool isHostInPlane(IPropertyTree *plane, const char *host, bool ipMatch)
+{
+    Owned<IPropertyTree> planeGroup;
+    if (plane->hasProp("@hostGroup"))
+        planeGroup.setown(getHostGroup(plane->queryProp("@hostGroup"), true));
+    else
+    {
+        if (!plane->hasProp("hosts"))
+            return false;
+        planeGroup.set(plane); // plane itself holds 'hosts'
+    }
+    Owned<IPropertyTreeIterator> hostsIter = planeGroup->getElements("hosts");
+    SocketEndpoint hostEp;
+    if (ipMatch)
+        hostEp.set(host);
+    ForEach (*hostsIter)
+    {
+        const char *planeHost = hostsIter->query().queryProp(nullptr);
+        if (ipMatch)
+        {
+            SocketEndpoint planeHostEp(planeHost);
+            if (planeHostEp.ipequals(hostEp))
+                return true;
+        }
+        else if (streq(planeHost, host))
+            return true;
+    }
+    return false;
+}
+
+bool getPlaneHost(StringBuffer &host, IPropertyTree *plane, unsigned which)
+{
+    Owned<IPropertyTree> hostGroup;
+    if (plane->hasProp("@hostGroup"))
+        hostGroup.setown(getHostGroup(plane->queryProp("@hostGroup"), true));
+    else if (plane->hasProp("hosts"))
+        hostGroup.set(plane); // the plane holds the "hosts"
+    else
+        return false;
+
+    if (which >= hostGroup->getCount("hosts"))
+        throw makeStringException(0, "getPlaneHost: index out of range");
+    VStringBuffer xpath("hosts[%u]", which+1); // which is 0 based
+    host.append(hostGroup->queryProp(xpath));
+    return true;
+}
+
 constexpr const char * lz_plane_path = "storage/planes[@category='lz']";
 
 IPropertyTreeIterator * getDropZonePlanesIterator(const char * name)
--- a/esp/services/ws_fs/ws_fsService.cpp
+++ b/esp/services/ws_fs/ws_fsService.cpp
@@ -1833,29 +1833,12 @@ bool CFileSprayEx::onGetDFUExceptions(IEspContext &context, IEspGetDFUExceptions
 
 static StringBuffer &getDropZoneHost(const char *planeName, IPropertyTree *plane, StringBuffer &host)
 {
-    if (plane->hasProp("hosts"))
+    if (!host.isEmpty())
     {
-        if (!host.isEmpty())
-        {
-            // Already have host. Just need to check that the ip is valid for storage plane.
-            bool ipAddressMatches = false;
-            Owned<IPropertyTreeIterator> hostIter = plane->getElements("hosts");
-            ForEach (*hostIter)
-            {
-                const char *knownIP = hostIter->query().queryProp(nullptr);
-                if (streq(knownIP, host))
-                {
-                    ipAddressMatches=true;
-                    break;
-                }
-            }
-            if (!ipAddressMatches)
-                throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "host %s is not valid storage plane %s", host.str(), planeName);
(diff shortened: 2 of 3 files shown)


esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -1833,13 +1833,14 @@ bool CFileSprayEx::onGetDFUExceptions(IEspContext &context, IEspGetDFUExceptions
 
 static StringBuffer &getDropZoneHost(const char *planeName, IPropertyTree *plane, StringBuffer &host)
 {
-    if (plane->hasProp("hosts"))
+    if (plane->hasProp("@hostGroup"))
```
ghalliday, line 1836: I think we support both hosts and @hostGroup.  On baremetal I think @hostGroup is expanded into hosts.  Probably should cope with both.

ghalliday (commented):
@jakesmith one comment/question

esp/services/ws_fs/ws_fsService.cpp
jakesmith: >On baremetal I think @hostGroup is expanded into hosts. 

it wasn't which is why it caused the problem to be highlighted.

I think it's the other way around, in bare-metal afaics, it's always a reference to a @hostGroup at the moment, as per constructed from env. via generateHosts and normalizeHostGroups.

In cloud, the schema allows for both.

So I think in bare-metal (at least at moment) it's always a @hostGroup ref., and in cloud it could be either.

But yes, that implies this code must support both too.
I'll change it.

AttilaVamos (commented):
@jakesmith @ghalliday 
Should we have all Spray related tests in 3 versions, like:
- Containerized
- Bare metal with planes
- Bare metal with IPs (deprecated?)

jakesmith:
> @jakesmith @ghalliday Should we have all Spray related tests in 3 versions, like:
> 
> * Containerized
> * Bare metal with planes
> * Bare metal with IPs (deprecated?)

Containerized will need to be tested in a k8s env., I'm not sure if there's anything we can do in bare-metal to test it.
The existing spray tests should work in both bare-metal and containerized, if they don't, I suspect the tests need tweaking.
I'm not sure how many spray tests there are, but it thinkit would be overkill for all of them to test a 'planes' test vs a legacy dropzone/ip/host test.

esp/services/ws_fs/ws_fsService.cpp
jakesmith: @ghalliday - please see reworked code/added helpers.

dali/base/dautils.cpp
```diff
@@ -47,6 +47,59 @@
 #define MIN_REDIRECTION_LOAD_INTERVAL 1000
 
 
+bool isHostInPlane(IPropertyTree *plane, const char *host, bool ipMatch)
+{
+    Owned<IPropertyTreeIterator> hostsIter;
+    if (plane->hasProp("@hostGroup"))
+    {
+        Owned<IPropertyTree> hostGroup = getHostGroup(plane->queryProp("@hostGroup"), true);
+        hostsIter.setown(hostGroup->getElements("*"));
```
ghalliday, line 56: Should be

`hostsIter.setown(hostGroup->getElements("hosts"));
`

dali/base/dautils.cpp
```diff
@@ -47,6 +47,59 @@
...
+    return false;
+}
+
+bool getPlaneHost(StringBuffer &host, IPropertyTree *plane, unsigned which)
+{
+    VStringBuffer xpath("hosts[%u]", which+1); // which is 0 based
+    if (plane->hasProp("@hostGroup"))
+    {
+        Owned<IPropertyTree> hostGroup = getHostGroup(plane->queryProp("@hostGroup"), true);
+        if (which >= hostGroup->numChildren())
```
ghalliday, line 88: should also be checking getCount("hosts").

esp/services/ws_fs/ws_fsService.cpp
```diff
@@ -1833,29 +1833,12 @@ bool CFileSprayEx::onGetDFUExceptions(IEspContext &context, IEspGetDFUExceptions
...
-            if (!ipAddressMatches)
-                throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "host %s is not valid storage plane %s", host.str(), planeName);
-        }
-        else
-            host.set(plane->queryProp("hosts[1]"));
+        if (!isHostInPlane(plane, host, true))
+            throw makeStringExceptionV(ECLWATCH_INVALID_INPUT, "host %s is not valid storage plane %s", host.str(), planeName);
     }
-    else
+    if (!getPlaneHost(host, plane, 0))
```
ghalliday, line 1841: should be else...

ghalliday (commented):
@jakesmith a couple of issues.

dali/base/dautils.cpp
jakesmith: can change it, this is not time critical, but thinking was that numChildren is faster, but I'll change it as more readable if same.

esp/services/ws_fs/ws_fsService.cpp
jakesmith: whoops, yes.

dali/base/dautils.cpp
jakesmith: presumably equivalent, because only "hosts" there, but I'll chane for clarity.

jakesmith:

dali/base/dautils.cpp
```diff
@@ -47,6 +47,59 @@
 #define MIN_REDIRECTION_LOAD_INTERVAL 1000
 
 
+bool isHostInPlane(IPropertyTree *plane, const char *host, bool ipMatch)
+{
```
AttilaVamos, line 51: I am confused with ipMatch parameter. It seems always true. Why is it needed?

jakesmith: it's debatable whether you want to resolve all host names to see if a match, or if in other situations it should be a text only match, e.g. resolving to IP's involves a DNS lookup and might fail/stall.
The option is there for the alternative.

AttilaVamos: Ok

AttilaVamos (approved):
Changes tested on Ubuntu/CentOS and seems ok.

dali/base/dautils.cpp
ghalliday: @jakesmith this hasn't been changed yet.

jakesmith: @ghalliday - see new commit (what I should have released last time)

ghalliday (approved):
@jakesmith please squash

jakesmith:
@ghalliday - squashed.

ghalliday:
@AttilaVamos that seems reasonable to me.  It is a bug fix, and I think it is safe.  @jakesmith what do you think?

jakesmith:
@ghalliday @AttilaVamos - retargeted to 8.10.x

PR #16744, opened 2022-12-01
HPCC-28661 Improve stripped build package naming

GordonSmith:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:


ghalliday:
@jakesmith @richardkchapman  what do you think about this?  Will it be confusing for debug builds?  I assume we also need to default our builds to stripping symbols so that the correct rpm is built.  Is that going to cause us problems?
I think this naming convention is a good idea for docker images.  Not so sure about rpms, but it does make sense to be consistent.

GordonSmith:
It is worth noting that we are currently stripping symbols for all "Release" builds...

jakesmith:
It'll be okay for developers IMO, they will continue to be built without stripping, albeit with a package name change.
Release, as GS says, already stripped, and the default, so that package name will remain the same.
Agree it would also make sense for docker images to follow the same convention

CMakeLists.txt
```diff
@@ -304,6 +304,10 @@ if(TOP_LEVEL_PROJECT)
         set(CPACK_STRIP_FILES TRUE)
     endif()
 
+    if (NOT CPACK_STRIP_FILES) 
+        set(packageStrippedLabel "_with_symbols")
```
GordonSmith, line 308: Rename to `_withsymbols`

GordonSmith:
@ghalliday @jakesmith rebased onto newbuild and renamed to "_withsymbols" (ensure each `_` denotes a different thing)

ghalliday (approved):

ghalliday:
@jakesmith @cloLN care you happy with this change?  I think it means all our rpms will change names.  What issues will that cause?

jakesmith (approved):
@GordonSmith - looks fine by me.

PR #16746, opened 2022-12-01
HPCC-28665 Fix lost stats. issue merging inactive/active stats.

jakesmith:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jstats.cpp
+++ b/system/jlib/jstats.cpp
@@ -2354,6 +2354,24 @@ void CRuntimeStatistic::merge(unsigned __int64 otherValue, StatsMergeAction merg
 
 //--------------------------------------------------------------------------------------------------------------------
 
+CRuntimeStatisticCollection::CRuntimeStatisticCollection(const CRuntimeStatisticCollection & _other) : mapping(_other.mapping)
+#ifdef _DEBUG
+, ignoreUnknown(_other.ignoreUnknown)
+#endif
+{
+    unsigned num = mapping.numStatistics();
+    values = new CRuntimeStatistic[num+1];
+    for (unsigned i=0; i <= num; i++)
+        values[i].set(_other.values[i].get());
+
+    CNestedRuntimeStatisticMap *otherNested = _other.queryNested();
+    if (otherNested)
+    {
+        nested = createNested();
+        queryNested()->set(*otherNested, 0);
+    }
+}
+
 CRuntimeStatisticCollection::~CRuntimeStatisticCollection()
 {
     delete [] values;
--- a/thorlcr/activities/indexread/thindexreadslave.cpp
+++ b/thorlcr/activities/indexread/thindexreadslave.cpp
@@ -76,6 +76,7 @@ protected:
     rowcount_t rowLimit = RCMAX;
     bool useRemoteStreaming = false;
     Owned<IFileIO> lazyIFileIO;
+    mutable CriticalSection ioStatsCS;
     unsigned fileTableStart = NotFound;
 
     template<class StatProvider>
@@ -376,9 +377,10 @@ public:
     }
     void updateStats()
     {
+        // NB: updateStats() should always be called whilst ioStatsCS is held.
         if (lazyIFileIO)
         {
-            mergeStats(stats, lazyIFileIO);
+            mergeStats(inactiveStats, lazyIFileIO);
             if (currentPart<partDescs.ordinality())
                 mergeFileStats(&partDescs.item(currentPart), lazyIFileIO);
         }
@@ -387,9 +389,12 @@ public:
     {
         if (currentManager)
         {
-            updateStats();
             resetManager(currentManager);
             currentManager = nullptr;
+
+            CriticalBlock b(ioStatsCS);
+            updateStats();
+            lazyIFileIO.clear();
         }
         IKeyManager *keyManager = nullptr;
         currentInput.setown(getNextInput(keyManager, currentPart, true));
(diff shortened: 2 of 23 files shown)


jakesmith:
@ghalliday - please review, I've still to do 1 or 2 tests, e.g. of loop global sort that uses inactiveStats

thorlcr/graph/thgraph.hpp
```diff
@@ -1092,7 +1092,7 @@ class graph_decl CActivityBase : implements CInterfaceOf<IThorRowInterfaces>, im
     bool timeActivities; // purely for access efficiency
     bool receiving, cancelledReceive, initialized, reInit;
     Owned<IThorGraphResults> ownedResults; // NB: probably only to be used by loop results
-    CRuntimeStatisticCollection stats;
+    CRuntimeStatisticCollection stats; // NB: these are current live act. stats - merged with inactive via gatherActiveStats
```
ghalliday, line 1095: I think it should be possible/desirable to delete this, and only have inactive.

thorlcr/graph/thgraphslave.cpp
```diff
@@ -571,30 +571,20 @@ void CSlaveActivity::serializeStats(MemoryBuffer &mb)
     // and updates the activity's 'stats'.
     // The callback fetches the current state of the activity's stats, called within a critical section ('statsCs'),
     // which the activity should use to protect any objects it uses whilst stats are being collected.
+
+    CRuntimeStatisticCollection serializedStats(stats.queryMapping());
```
ghalliday, line 575: I think better and simpler would be

CRuntimeStatisticCollection serializedStats(inactiveStats);

remove the someInactiveStats variable.

The copy constructor for CRuntimeStatisticCollection should be fixed to copy nested stats (see set())

ghalliday (commented):
@jakesmith I know you hadn't tagged me (check boxes not completed), but a couple of comments..  Looks much simpler.

thorlcr/graph/thgraphslave.cpp
ghalliday: The copy constructor effectively does a memcpy.  It doesn't map via the mapping:

```
        for (unsigned i=0; i <= num; i++)
            values[i].set(_other.values[i].get());
```

thorlcr/graph/thgraph.hpp
jakesmith: I agree.. I took that approach initially.
There are quite a lot of updates to 'stats' at the moment, but I think they could all be to inactiveStats, since they are overwritten each time they are updated.

But I backed out of that change, because of the way 'stats' is passed into the code context and used by registerTimer.
I suspect the code context object should instead have it's own CRuntimeStatisticCollection , and during activity.serializeStats for it's stats to be requested and merged

thorlcr/graph/thgraphslave.cpp
jakesmith: @ghalliday - I've pushed a commit that unconditionally copies via ctor.. but let me know what u think re. efficiency loss.

system/jlib/jstats.cpp
```diff
@@ -2354,6 +2354,24 @@ void CRuntimeStatistic::merge(unsigned __int64 otherValue, StatsMergeAction merg
...
+    unsigned num = mapping.numStatistics();
+    values = new CRuntimeStatistic[num+1];
+    for (unsigned i=0; i <= num; i++)
+        values[i].set(_other.values[i].get());
+
+    CNestedRuntimeStatisticMap *otherNested = _other.queryNested();
+    if (otherNested)
+    {
+        nested = createNested();
+        queryNested()->set(*otherNested, 0);
```
jakesmith, line 2371: wondered if for code simplicity, it wouldn't have just been better for ctor to call set()..

thorlcr/graph/thgraphslave.cpp
jakesmith: > The copy constructor effectively does a memcpy. It doesn't map via the mapping:
> 
> ```
>         for (unsigned i=0; i <= num; i++)
>             values[i].set(_other.values[i].get());
> ```

ok yes, and I guess a extra condition on nesting is nothing

thorlcr/graph/thgraph.hpp
jakesmith: @ghalliday - pushed another commit (3rd) - that removes the dependency on the activities 'stats' from the code context, and changes it so it has it's own collection which is gathered on serialization.
Please take a look/comment.

jakesmith:
@ghalliday - simplified/consolidated all other stats to use 'inactiveStats' and removed 'stats' as an activity member. Please see 4th commit.

ghalliday (approved):
@jakesmith looks good.  One minor comment, please squash

I have a slight question whether this will impact performance if there were very large numbers of child activities.  I suspect it won't but may be worth keeping an eye out for any strange behaviour.

thorlcr/activities/indexread/thindexreadslave.cpp
```diff
@@ -378,7 +379,7 @@ class CIndexReadSlaveBase : public CSlaveActivity
     {
         if (lazyIFileIO)
         {
-            mergeStats(stats, lazyIFileIO);
+            mergeStats(inactiveStats, lazyIFileIO);
```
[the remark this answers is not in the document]
jakesmith: it is protected by ioStatsCS - in the sense that this method (updateStats()) is always called within a ioStatsCS crit block.
I'll add a comment to say that it must be.

jakesmith:
@ghalliday - added comment, and squashed.

PR #16749, opened 2022-12-01
HPCC-28667 Add setPlaneExternal() to CDfsLogicalFilename

wangkx:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dali/base/dautils.cpp
+++ b/dali/base/dautils.cpp
@@ -1003,19 +1003,8 @@ IPropertyTree *CDfsLogicalFileName::createSuperTree() const
     return ret;
 }
 
-
-void CDfsLogicalFileName::setExternal(const char *location,const char *path)
+static void convertPosixPathToLfn(StringBuffer &str,const char *path)
 {
-    if (!path||!*path)
-        return;
-    if (isPathSepChar(path[0])&&(path[0]==path[1])) {
-        RemoteFilename rfn;
-        rfn.setRemotePath(path);
-        setExternal(rfn);  // overrides ip
-        return;
-    }
-    StringBuffer str(EXTERNAL_SCOPE "::");
-    str.append(location);
     if ((path[1]==':')&&(path[2]=='\\')) {
         str.append("::").append(path[0]).append('$');
         path+=2;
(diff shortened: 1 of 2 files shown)


dali/base/dautils.cpp
```diff
@@ -1003,6 +1003,15 @@ IPropertyTree *CDfsLogicalFileName::createSuperTree() const
     return ret;
 }
 
+void CDfsLogicalFileName::setPlaneExternal(const char *plane,const char *path)
+{
+    Owned<IPropertyTree> dropZone = getDropZonePlane(plane);
```
jakesmith, line 1008: this should not be drop zone specific.. CDfsLogicalFileName setPlaneExternal makes sense for other planes too.
The rejection or a non LZ plane should live elsewhere.

It's debatable whether it should force the check here at all (even for any plane), since you may want to use this utility helper to construct logical names even if the planes don't exist.

dali/base/dautils.cpp
```diff
@@ -1014,8 +1023,19 @@ void CDfsLogicalFileName::setExternal(const char *location,const char *path)
         setExternal(rfn);  // overrides ip
         return;
     }
-    StringBuffer str(EXTERNAL_SCOPE "::");
-    str.append(location);
+
+    StringBuffer str;
+    if (planeName.isEmpty())
```
jakesmith, line 1028: it would be better not to inro. a new member, and/or alter setExternal to cope with both.
I think setPlaneExternal should look like this:
```
void CDfsLogicalFileName::setPlaneExternal(const char *planeName,const char *path)
{
    if (isEmptyString(path))
        return;
    if (isPathSepChar(path[0])&&(path[0]==path[1]))
        throw .... // not expected..
    StringBuffer str(PLANE_SCOPE "::");
    str.append(planeName);
    // call common function that setExternal calls too
}


```
since //ip/ for the path for setPlaneExternal

The common part (that is currentlly after     str.append(location); ), can be extract to a static func. that both setExternal and setPlaneExternal calls.

jakesmith (changes requested):
@wangkx - please see comments

wangkx:
@jakesmith I made the changes based on your comments. Please let me know if I miss anything. I rebased the 2 commits into one because I think it is easier for you to review.

dali/base/dautils.cpp
```diff
@@ -1016,6 +1026,11 @@ void CDfsLogicalFileName::setExternal(const char *location,const char *path)
     }
     StringBuffer str(EXTERNAL_SCOPE "::");
     str.append(location);
+    appendExternalStr(str,path);
+}
+
+void CDfsLogicalFileName::appendExternalStr(StringBuffer &str,const char *path)
```
jakesmith, line 1032: minor: as this method doesn't depend on anything in the object instance, it would make it clearer if it was declared as a static, and only passed the result back. Then setExternal and setPlaneExternal can call set() with that result.

"convertPosixPathToLfn" might be more intuitive.

jakesmith (approved):
@wangkx - looks fine, but a couple of suggestions.

wangkx:
@ghalliday I made the changed based on Jake's suggestions. Please review and merge.

ghalliday (approved):

PR #16753, opened 2022-12-01
HPCC-26018 Add XPath extension functions for text manipulation

timothyklemm:
- Convert an XML node set into XML text.
- Compress and decompress text.
- Apply Base64 character encoding and decoding.
- Apply XML character encoding and decoding.
- Apply deprecated encryption and decryption.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/esdllib/esdl_xpath_extensions_libxml.cpp
+++ b/esp/esdllib/esdl_xpath_extensions_libxml.cpp
@@ -472,6 +472,247 @@ static void strTokenizeFunction(xmlXPathParserContextPtr ctxt, int nargs)
     xmlFree(delimiters);
 }
 
+static void deprecatedEncryptString(xmlXPathParserContextPtr ctxt, int nargs)
+{
+    if (nargs != 1)
+    {
+        xmlXPathSetArityError(ctxt);
+        return;
+    }
+
+    xmlChar *toEncrypt = xmlXPathPopString(ctxt);
+    if (xmlXPathCheckError(ctxt)) //includes null check
+        return;
+
+    StringBuffer encrypted;
+    try
+    {
+        encrypt(encrypted, (const char*)toEncrypt);
+    }
+    catch (IException* e)
+    {
+        e->Release();
+        xmlXPathSetError(ctxt, XPATH_EXPR_ERROR);
+    }
+    catch (...)
+    {
+        xmlXPathSetError(ctxt, XPATH_EXPR_ERROR);
+    }
+    if (!xmlXPathCheckError(ctxt))
+        xmlXPathReturnString(ctxt, xmlStrdup((const xmlChar *)encrypted.str()));
+    xmlFree(toEncrypt);
+}
+
+static void deprecatedDecryptString(xmlXPathParserContextPtr ctxt, int nargs)
+{
+    if (nargs != 1)
+    {
+        xmlXPathSetArityError(ctxt);
+        return;
+    }
+
+    xmlChar *toDecrypt = xmlXPathPopString(ctxt);
+    if (xmlXPathCheckError(ctxt)) //includes null check
+        return;
+
+    StringBuffer decrypted;
+    try
+    {
+        decrypt(decrypted, (const char*)toDecrypt);
+    }
+    catch (IException* e)
+    {
+        e->Release();
+        xmlXPathSetError(ctxt, XPATH_EXPR_ERROR);
+    }
+    catch (...)
+    {
+        xmlXPathSetError(ctxt, XPATH_EXPR_ERROR);
+    }
+    if (!xmlXPathCheckError(ctxt))
+        xmlXPathReturnString(ctxt, xmlStrdup((const xmlChar *)decrypted.str()));
+    xmlFree(toDecrypt);
+}
+
+static void encodeBase64String(xmlXPathParserContextPtr ctxt, int nargs)
+{
+    if (0 == nargs || nargs > 2)
+    {
+        xmlXPathSetArityError(ctxt);
+        return;
+    }
+
+    xmlChar *toEncode = xmlXPathPopString(ctxt);
+    if (xmlXPathCheckError(ctxt)) //includes null check
+        return;
+    bool lineBreaks = false;
+    if (2 == nargs)
+    {
+        lineBreaks = xmlXPathPopBoolean(ctxt);
+        if (xmlXPathCheckError(ctxt))
+        {
+            xmlFree(toEncode);
+            return;
+        }
+    }
+
+    StringBuffer encoded;
+    JBASE64_Encode((const char*)toEncode, long(strlen((const char*)toEncode)), encoded, lineBreaks);
+    xmlXPathReturnString(ctxt, xmlStrdup((const xmlChar *)encoded.str()));
+    xmlFree(toEncode);
+}
+
+static void decodeBase64String(xmlXPathParserContextPtr ctxt, int nargs)
+{
+    if (nargs != 1)
+    {
+        xmlXPathSetArityError(ctxt);
+        return;
+    }
+
+    xmlChar *toDecode = xmlXPathPopString(ctxt);
+    if (xmlXPathCheckError(ctxt)) //includes null check
+        return;
+
+    StringBuffer decoded;
+    JBASE64_Decode((const char*)toDecode, decoded);
+    xmlXPathReturnString(ctxt, xmlStrdup((const xmlChar *)decoded.str()));
+    xmlFree(toDecode);
+}
+
+static void escapeXmlCharacters(xmlXPathParserContextPtr ctxt, int nargs)
+{
+    if (nargs != 1)
+    {
+        xmlXPathSetArityError(ctxt);
+        return;
+    }
+
+    xmlChar *toEncode = xmlXPathPopString(ctxt);
+    if (xmlXPathCheckError(ctxt)) //includes null check
+        return;
+
+    StringBuffer encoded;
+    encodeXML((const char*)toEncode, encoded);
+    xmlXPathReturnString(ctxt, xmlStrdup((const xmlChar *)encoded.str()));
+    xmlFree(toEncode);
+}
+
+static void unescapeXmlCharacters(xmlXPathParserContextPtr ctxt, int nargs)
+{
+    if (nargs != 1)
+    {
+        xmlXPathSetArityError(ctxt);
+        return;
+    }
+
+    xmlChar *toDecode = xmlXPathPopString(ctxt);
+    if (xmlXPathCheckError(ctxt)) //includes null check
+        return;
+
+    StringBuffer decoded;
+    decodeXML((const char*)toDecode, decoded);
+    xmlXPathReturnString(ctxt, xmlStrdup((const xmlChar *)decoded.str()));
+    xmlFree(toDecode);
+}
+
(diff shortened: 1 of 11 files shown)


esp/bindings/http/platform/httpbinding.cpp
```diff
@@ -878,8 +879,13 @@ bool EspHttpBinding::doAuth(IEspContext* ctx)
 {
     if(m_authtype.length() == 0 || stricmp(m_authtype.str(), "Basic") == 0)
     {
-        CumulativeTimer* timer = ctx->queryTraceSummaryCumulativeTimer(LogNormal, "custom_fields.basicAuthTime", TXSUMMARY_GRP_ENTERPRISE);
-        CumulativeTimer::Scope authScope(timer);
+        static const char* timerName = "custom_fields.basicAuthTime";
+        CumulativeTimer* timer = ctx->queryTraceSummaryCumulativeTimer(LogNormal, timerName, TXSUMMARY_GRP_ENTERPRISE);
```
afishbeck, line 883: These PRs are going to need rebasing in the right order as the others are merged.

esp/esdllib/esdl_script.cpp
```diff
@@ -471,9 +472,38 @@ class CEsdlTransformOperationMySqlCall : public CEsdlTransformOperationBase
 
         m_server.setown(compileOptionalXpath(stag.getValue("server")));
         m_user.setown(compileOptionalXpath(stag.getValue("user")));
-        m_password.setown(compileOptionalXpath(stag.getValue("password")));
         m_database.setown(compileOptionalXpath(stag.getValue("database")));
 
+        /*
```
afishbeck, line 477: I don't think you really needed to overlay these PRs like this.

esp/esdllib/esdl_xpath_extensions_libxml.cpp
```diff
@@ -472,6 +474,326 @@ static void strTokenizeFunction(xmlXPathParserContextPtr ctxt, int nargs)
...
+
+    StringBuffer decrypted;
+    try
+    {
+        decrypt(decrypted, (const char*)toDecrypt);
+    }
+    catch (IException* e)
+    {
+        e->Release();
+        xmlXPathSetError(ctxt, XPATH_INVALID_OPERAND);
```
afishbeck, line 619: Not sure this would be an operand error.  Perhaps just go with the vague XPATH_UNKNOWN_FUNC_ERROR?

esp/esdllib/esdl_xpath_extensions_libxml.cpp
```diff
@@ -472,6 +474,326 @@ static void strTokenizeFunction(xmlXPathParserContextPtr ctxt, int nargs)
...
+
+    StringBuffer encrypted;
+    try
+    {
+        encrypt(encrypted, (const char*)toEncrypt);
+    }
+    catch (IException* e)
+    {
+        e->Release();
+        xmlXPathSetError(ctxt, XPATH_INVALID_OPERAND);
```
afishbeck, line 588: See XPATH_UNKNOWN_FUNC_ERROR below.

esp/esdllib/esdl_xpath_extensions_libxml.cpp
```diff
@@ -472,6 +474,326 @@ static void strTokenizeFunction(xmlXPathParserContextPtr ctxt, int nargs)
...
+        xmlXPathSetArityError(ctxt);
+        return;
+    }
+
+    xmlChar *toDecode = xmlXPathPopString(ctxt);
+    if (xmlXPathCheckError(ctxt)) //includes null check
+        return;
+
+    StringBuffer decoded;
+    JBASE64_Decode((const char*)toDecode, decoded);
```
afishbeck, line 671: In both cases do you mean the data has to be text only?  You're making a huge assumption here that this isn't true binary data.  I'm not sure doing this as an xpath function is a good idea.  If we do this we need to understand why the script needs access to the binary data, and how they would actually manipulate it.  I could see for example sending an image to an http call decoded, but that wouldn't work through this mechanism.  Unless you have some good use cases for a pure text version, and can put some safeguards around this that it really will be text, base64 may be another thing we have to separate out of this PR.

esp/esdllib/esdl_xpath_extensions_libxml.cpp
```diff
@@ -472,6 +474,326 @@ static void strTokenizeFunction(xmlXPathParserContextPtr ctxt, int nargs)
...
+        lineBreaks = xmlXPathPopBoolean(ctxt);
+        if (xmlXPathCheckError(ctxt))
+        {
+            xmlFree(toEncode);
+            return;
+        }
+    }
+
+    StringBuffer encoded;
+    JBASE64_Encode((const char*)toEncode, long(strlen((const char*)toEncode)), encoded, lineBreaks);
```
afishbeck, line 653: Even if we wait for decode (see below) you could make a case for still being able to base64 encode text.  Not sure it's totally useful, but it's safe.

esp/esdllib/esdl_xpath_extensions_libxml.cpp
```diff
@@ -472,6 +474,326 @@ static void strTokenizeFunction(xmlXPathParserContextPtr ctxt, int nargs)
...
+    StringBuffer encoded;
+    JBASE64_Encode((void*)compressed.bufferBase(), compressed.length(), encoded, false);
+    xmlXPathReturnString(ctxt, xmlStrdup((const xmlChar *)encoded.str()));
+
+    xmlFree(toCompress);
+}
+
+static void decompressString(xmlXPathParserContextPtr ctxt, int nargs)
+{
+    if (nargs != 1)
```
afishbeck, line 736: Have you tested what happens if either the base64 or compressed data isn't valid?

afishbeck (commented):
@timothyklemm some comments.

esp/bindings/http/platform/httpbinding.cpp
timothyklemm: Yes. The first two PRs share a change, and are logically stacked based on complexity (first in the stack is the first to complete the review). This one was stacked out of convenience, before it's priority was changed. The TxSummary change is more important, so my priority is for it to be merged first.

esp/esdllib/esdl_xpath_extensions_libxml.cpp
timothyklemm: My thought was that "invalid operand" was closest to "invalid parameter", which seems applicable since the parameter value is not valid for the intended use. "Unknown function" sounds more like the function name itself is invalid, instead of the parameters passed, which is more confusing to me than vague.

esp/esdllib/esdl_xpath_extensions_libxml.cpp
timothyklemm: I had not. I'm not concerned about bad encoding because even a good encoding can yield invalid compressed data. I'll catch exceptions from the decompress.

esp/esdllib/esdl_xpath_extensions_libxml.cpp
timothyklemm: I am making that assumption. The return types are limited, with only one ("external") that could possibly contain true binary data. We have no support for that type.

The function name is also significant. The presence of "String" in the name can refer to both the input parameter type and the result type. We can't return true binary in a string without some sort of encoding or escaping of bytes.

If we want to (and can) support true binary, I would consider a new function like `decodeBase64Binary`. I'm not sure it's worth holding this out because somebody could try to decode binary data and gets garbled text instead.

esp/esdllib/esdl_xpath_extensions_libxml.cpp
timothyklemm: Looking at the enumeration again, perhaps `XPATH_EXPR_ERROR` provides an appropriate level of vagueness if it just means something in XPath is wrong.

afishbeck: I see what you mean about the likely cause being the content of the parameter.. but we don't really know that.

But there are no great choices.  They made explicit errors for their needs and didn't leave anything similar to an unknown error.

In libxml2 the cases of XPATH_INVALID_OPERAND error were because the operand couldn't be retrieved.

We could consider using XPTR_RESOURCE_ERROR as a vague error.  I don't even see it used in libxml2.  That would be thinking of our extension as a resource I guess.  But it's also a stretch.  So no good choices basically means it doesn't matter as long as we don't imply something way off that libxml2 may act on.

esp/esdllib/esdl_xpath_extensions_libxml.cpp
afishbeck: Right.  Thought about this some more and I'm sure scapps has cases where it base64s text content so this is fine.  We need use cases for binary before designing anything.

esp/esdllib/esdl_xpath_extensions_libxml.cpp
afishbeck: Did you mean you are going to add a try to this PR?

esp/esdllib/esdl_xpath_extensions_libxml.cpp
afishbeck: Sure XPATH_EXPR_ERROR would be fine.

esp/esdllib/esdl_xpath_extensions_libxml.cpp
timothyklemm: Yes. The decompress call does throw. When I passed "this is bad data" to the XPath function the decode "completed" and the decompress threw an exception.

timothyklemm:
@afishbeck and @asselitx, I've addressed Tony's comments. Including by rebasing directly onto 8.10.x.

afishbeck (approved):

esp/esdllib/esdl_xpath_extensions_libxml.cpp
```diff
@@ -472,6 +472,247 @@ static void strTokenizeFunction(xmlXPathParserContextPtr ctxt, int nargs)
...
+    StringBuffer xml;
+    for (int i = 0; i < toSerialize->nodeNr; i++)
+    {
+        xmlNodePtr node = toSerialize->nodeTab[i];
+        if (!node)
+            continue;
+        xmlOutputBufferPtr xmlOut = xmlAllocOutputBuffer(nullptr);
+        xmlNodeDumpOutput(xmlOut, node->doc, node, 0, 1, nullptr);
+        xmlOutputBufferFlush(xmlOut);
+        xmlBufPtr buf = (xmlOut->conv != nullptr) ? xmlOut->conv : xmlOut->buffer;
```
asselitx, line 706: I'm cleanup not required for `xmlBufPtr buf` ?

asselitx (approved):

esp/esdllib/esdl_xpath_extensions_libxml.cpp
timothyklemm: It should not be necessary. It is part of `xmlOut`, which is cleaned by `xmlOutputBufferClose`.

timothyklemm:
@ghalliday please merge

PR #16759, opened 2022-12-02
HPCC-28682 Read larger chunks of data in Roxie

richardkchapman:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Cloud-compatibility
  - [x] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jfile.cpp
+++ b/system/jlib/jfile.cpp
@@ -7484,3 +7484,95 @@ IAPICopyClient * createApiCopyClient(IStorageApiInfo * source, IStorageApiInfo *
     }
     return nullptr;
 }
+
+class CBlockFileIO : public CInterfaceOf<IFileIO>
+{
+private:
+    Owned <IFileIO> io;
+    offset_t fileSize = 0;
+    static size32_t blockSize;
+    static thread_local byte *buffer;
+    static thread_local offset_t lastReadPos;
+    static thread_local size32_t lastReadLen;
+    static thread_local CBlockFileIO *lastReadCaller;
+    static bool onThreadTerm(bool isPooled);
+public:
+    CBlockFileIO(IFileIO *_io) : io(_io), fileSize(_io->size()) {}
+    ~CBlockFileIO();
+    virtual size32_t read(offset_t pos, size32_t len, void * data) override;
+    virtual offset_t size() override { return fileSize; }
+    virtual size32_t write(offset_t pos, size32_t len, const void * data) override { throwUnexpected(); }
+    virtual offset_t appendFile(IFile *file,offset_t pos=0,offset_t len=(offset_t)-1) override { throwUnexpected(); }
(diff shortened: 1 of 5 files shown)


system/jlib/jfile.cpp
```diff
@@ -7484,3 +7484,95 @@ IAPICopyClient * createApiCopyClient(IStorageApiInfo * source, IStorageApiInfo *
...
+            lastReadPos = readPos;
+            lastReadLen = readLen;
+            lastReadCaller = this;
+        }
+        size32_t copyNow;
+        if (pos+len <= readPos+readLen)
+            copyNow = len;
+        else
+            copyNow = readPos+readLen-pos;
+        memcpy(data, buffer + pos-readPos, copyNow);
```
ghalliday, line 7562: data + totalCopied 

or something similar.

system/jlib/jfile.cpp
```diff
@@ -7484,3 +7484,95 @@ IAPICopyClient * createApiCopyClient(IStorageApiInfo * source, IStorageApiInfo *
...
+    virtual offset_t appendFile(IFile *file,offset_t pos=0,offset_t len=(offset_t)-1) override { throwUnexpected(); }
+    virtual void setSize(offset_t size) override { throwUnexpected(); }
+    virtual void flush() override { throwUnexpected(); }
+    virtual void close() override { io->close(); }
+    virtual unsigned __int64 getStatistic(StatisticKind kind) override { return io->getStatistic(kind); }
+
+    inline static void setBlockSize(size32_t _blockSize) { blockSize = _blockSize; }
+};
+
+size32_t CBlockFileIO::blockSize = 64*1024;
```
ghalliday, line 7514: Might be better if these were static variables rather than static members of the class (different visibility and possibly less efficient access.).  May be identical though..

ghalliday (commented):
A couple of comments from a quick scan

ghalliday:
I am not sure how useful this is going to be if we don't also insert those nodes into the table at the same time.  
A read size of 64K would have little extra overhead, but is less likely to be reused.  It might help with sequential reads through leaf nodes.  Not sure about branch nodes because it will tend to walk down the branches rather than moving to the next branch node.  A size of 1MB is likely to significantly increase the latency.
We should add some stats to enable us to determine how successful the optimization is.

richardkchapman:
We can't insert nodes into the table until we have the ability to read without decompressing.

Idea isn't necessarily to speed things up, but rather to reduce costs (by reading any additional data that we can without being charged, in the hope that we may use it).

Without knowing what the latency of a 1mb read vs an 8k read vs a 64k read, it's hard to comment further.

richardkchapman:
Superseded

PR #16764, opened 2022-12-05
HPCC-28675 Add findDropZonePlane()

wangkx:
Map from a host/ip to a known dropzone plane name.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/dali/base/dautils.cpp
+++ b/dali/base/dautils.cpp
@@ -103,6 +103,7 @@ IPropertyTreeIterator * getDropZonePlanesIterator(const char * name)
         xpath.appendf("[@name='%s']", name);
     return getGlobalConfigSP()->getElements(xpath);
 }
+
 IPropertyTree * getDropZonePlane(const char * name)
 {
     if (isEmptyString(name))
@@ -110,6 +111,29 @@ IPropertyTree * getDropZonePlane(const char * name)
     StringBuffer xpath(lz_plane_path);
     xpath.appendf("[@name='%s']", name);
     return getGlobalConfigSP()->getPropTree(xpath);
+}
+
+IPropertyTree * findDropZonePlane(const char * path, const char * host, bool ipMatch)
+{
+    if (streq(host, "localhost"))
+        host = nullptr;
+    Owned<IPropertyTreeIterator> iter = getDropZonePlanesIterator();
+    ForEach(*iter)
+    {
+        IPropertyTree & plane = iter->query();
+        if (host)
+        {
+            if (!isHostInPlane(&plane, host, ipMatch))
+                continue;
+        }
+        else if (plane.hasProp("@hostGroup") || plane.hasProp("hosts"))
+            continue;
+
+        //Match path
+        if (startsWith(path, plane.queryProp("@prefix")))
+            return LINK(&plane);
+    }
+    return nullptr;
 }
 
 extern da_decl const char *queryDfsXmlBranchName(DfsXmlBranchKind kind)
--- a/dali/base/dautils.hpp
+++ b/dali/base/dautils.hpp
@@ -542,6 +542,7 @@ extern da_decl ILockInfoCollection *deserializeLockInfoCollection(MemoryBuffer &
 
 extern da_decl IPropertyTreeIterator * getDropZonePlanesIterator(const char * name=nullptr);
 extern da_decl IPropertyTree * getDropZonePlane(const char * name);
+extern da_decl IPropertyTree * findDropZonePlane(const char * path, const char * host, bool ipMatch);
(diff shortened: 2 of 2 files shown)


jakesmith (commented):
@wangkx - please revisit this once https://github.com/hpcc-systems/HPCC-Platform/pull/16742 is merged.

jakesmith:
> @wangkx - please revisit this once #16742 is merged.

@wangkx - https://github.com/hpcc-systems/HPCC-Platform/pull/16742 has now been merged.

wangkx:
I see the isHostInPlane() in https://github.com/hpcc-systems/HPCC-Platform/pull/16742. Not sure it can be used here. If it returns false, how do I know whether it is because the plane has no "@hostGroup"/"hosts" (I should check the given host with the "localhost") or the given host does not match with any host in the plane (I should not check the given host with the "localhost")?  Any other changes in https://github.com/hpcc-systems/HPCC-Platform/pull/16742 should be used here?

jakesmith:
> @jakesmith I see the isHostInPlane() in #16742. Not sure it can be used here. If it returns false, how do I know whether it is because the plane has no "@hostGroup"/"hosts" (I should check the given host with the "localhost") or the given host does not match with any host in the plane (I should not check the given host with the "localhost")? Any other changes in #16742 should be used here?

I don't really get the "localhost" case.. nothing should be using these functions with a localhost.
It sounds like "localhost" is being used as a stand-in for cases where planes don't have hosts in other parts of the code, but they should not be doing any host matching in that case.

For cases when a specific host has been passed in, then it should not ever be localhost, and this function can call isHostInPlane on each plane to see if it matches.

jakesmith (commented):
@wangkx - see reply.

wangkx:
@jakesmith please review my latest commit. I may use the following code to find a plane based on the host.
```
        if (!findHostedPlane(host, true, hostedPlane) && strieq(host, "localhost"))
            findPlaneNoHost("lz", nullptr, hostedPlane);
```

jakesmith:
> @jakesmith please review my latest commit. I may use the following code to find a plane based on the host.
> 
> ```
>         if (!findHostedPlane(host, true, hostedPlane) && strieq(host, "localhost"))
>             findPlaneNoHost("lz", nullptr, hostedPlane);
> ```

Discussed offline.

Going forward, anything that is making a request to interact with a dropzone, should supply the dropzone name, not just the hostname (and path).
Those changes have largely already been made in other PR's some merged as I understand,
however, to cater for legacy client requests, we need to find the dropzone based on the hostname *and* path.
NB: in a cloud system with multiple dropzones, it is quite likely they will be non-hosted and mounted, so hostname is irrelevant, the only distiguishing factor is the path prefix.

So I think:
a) if dropzone name is supplied - use it directly, no lookup of dropzone involved.

b) if not, you need to find a matching dropzone, based on the host and prefix. (there should only be 1 unique match)
A request with "localhost" should be treated as unhosted, mounted off-node storage (the norm. in cloud case). 

I would add a:
```
IPropertyTree *findPlane(const char *path, const char *host, bool ipMatch)
```

treat incoming requests for "localhost" as null, and search all non hosted planes for a match.

jakesmith (commented):
@wangkx - see comment following offline discussion.

wangkx:
@jakesmith I noticed that the getDropZonePlanesIterator() is used inside the code of the findPlane() you drafted. Since the code can only be applied for dropzone, I am wondering if the name should be changed to findDropZonePlane(). Or, I may add a getPlanesIterator() as I did in https://github.com/hpcc-systems/HPCC-Platform/pull/16764/commits/3e5a2f7f634701a5fa66496393b9780a355973df and call it the findPlane(). Please advise.

dali/base/dautils.cpp
```diff
@@ -111,6 +111,28 @@ IPropertyTree * getDropZonePlane(const char * name)
     xpath.appendf("[@name='%s']", name);
     return getGlobalConfigSP()->getPropTree(xpath);
 }
+IPropertyTree * findDropZonePlane(const char * path, const char * host, bool ipMatch)
```
jakesmith, line 114: trivial/not new: unusual not to have new line between implementation methods in a cpp.
Can you add and to ones that proceed this new method?

jakesmith (approved):
@wangkx  - looks good.
You may want to address trivial comment, if so, please change and squash.

wangkx:
@ghalliday Please review and merge.

ghalliday (approved):

PR #16773, opened 2022-12-08
HPCC-26831 Support vault secret access from ECL in bare metal

afishbeck:
After this change, to configure vaults for a baremetal enviroment add vault information such as:

<vaults>
  <ecl name="my-ecl-vault" url="http://localhost:8200/v1/secret/data/ecl/${secret}" kind="kv-v2" client-secret="vaultSecret"></ecl>
  <eclUser name="my-eclUser-vault" url="http://localhost:8200/v1/secret/data/eclUser/${secret}" kind="kv-v2" client-secret="vaultSecret"></eclUser>
</vaults>

To the environment.xml file at the path: /Environment/Software/vaults

Notice in the vault config xml, the mention of "client-secret". That specifies where the token to access vault itself is stored. In the above example client-secret="vaultSecret" so the hpcc will look for the vault token in:

/opt/HPCCSystems/secrets/system/vaultSecret/token

With the token in that file, make sure there is no whitespace, newlines, etc.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/initfiles/componentfiles/configxml/RoxieTopology.xsl
+++ b/initfiles/componentfiles/configxml/RoxieTopology.xsl
@@ -121,6 +121,7 @@
                 </xsl:variable>
                 <xsl:value-of select="translate($path3, $oldPathSeparator, $newPathSeparator)"/>
             </xsl:attribute>
+            <xsl:copy-of select="/Environment/Software/vaults"/>
             <xsl:copy-of select="/Environment/Software/Directories"/>
             <!--
             # Generated for configuration info. accessed by getGlobalConfig()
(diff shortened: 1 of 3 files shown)


afishbeck:
@ghalliday this change was previously merged into 8.8.x and 8.10.x.  They would like it in 8.6.x.  Even though the goal is to get off 8.6.x, they would like to start using ECL secrets in bare-metal more quickly than upgrades from 8.6.x are feasible.

ghalliday (approved):

PR #16781, opened 2022-12-14
HPCC-28714 Add hashicorp vault support for namespaces

afishbeck:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/helm/hpcc/templates/_helpers.tpl
+++ b/helm/hpcc/templates/_helpers.tpl
@@ -569,6 +569,9 @@ vaults:
   {{- range $vault := . }}
     - name: {{ $vault.name }}
       kind: {{ $vault.kind }}
+    {{- if $vault.namespace }}
+      namespace: {{ $vault.namespace }}
+    {{- end }}
       url: {{ $vault.url }}
     {{- if index $vault "client-secret" }}
       client-secret: {{ index $vault "client-secret" }}
--- a/system/jlib/jsecrets.cpp
+++ b/system/jlib/jsecrets.cpp
@@ -236,6 +236,7 @@ private:
 
     StringBuffer schemeHostPort;
     StringBuffer path;
+    StringBuffer vaultNamespace;
     StringBuffer username;
     StringBuffer password;
     StringAttr name;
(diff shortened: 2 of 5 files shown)


afishbeck:
@ghalliday this is the change needed for our corporate vault support.  The change allows us to access enterprise vaults segmented into namespaces.

helm/hpcc/templates/_helpers.tpl
```diff
@@ -569,6 +569,7 @@ vaults:
   {{- range $vault := . }}
     - name: {{ $vault.name }}
       kind: {{ $vault.kind }}
+      namespace: {{ $vault.namespace }}
```
ghalliday, line 572: Doesn't this need to be inside a check that namespace is set.

```
{{- if $value.namespace }}
```

ghalliday (commented):
@afishbeck one question

helm/hpcc/templates/_helpers.tpl
afishbeck: Yes, missed that, thanks.

afishbeck: @ghalliday Updated to only add namespace if one is provided.

afishbeck:
@ghalliday updated with suggested change.

system/jlib/jsecrets.cpp
```diff
@@ -259,6 +260,10 @@ class CVault
         name.set(vault->queryProp("@name"));
         kind = getSecretType(vault->queryProp("@kind"));
 
+        vaultNamespace.set(vault->queryProp("@namespace"));
+        if (vaultNamespace.length())
+            addPathSepChar(vaultNamespace, '/');
```
ghalliday, line 265: @afishbeck is the trailing / correct?  All the examples I have seen do not have a trailing / e.g.

```
--header "X-Vault-Namespace: some/path
X-Vault-Namespace: root
```

afishbeck: It could be optional.  Vault cli has an option to generate a CURL example for each command (-output-curl-string).  The CURL example used a forward slash at the end of the namespace.  We tested via curl with the forward slash.  I don't remember if we also tried without it to see if it would work, but I decided for now at least to error on the side of matching what vault cli suggested.
I can't test namespaces on my non-enterprise vault but I could work with Fernando again to see if we can remove the slash.  Or I could open a JIRA to try removing the slash next time we are working with the enterprise vault.

PR #16783, opened 2022-12-14
HPCC-28619 CSV Heading spills to next row

AttilaVamos:
Fix the 4096-byte, hardwired CSV record size problem in Hthor CSV read.

The fix similar to the solution what is used in Thor.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/ecl/hthor/hthor.cpp
+++ b/ecl/hthor/hthor.cpp
@@ -9476,7 +9476,6 @@ void CHThorCsvReadActivity::gatherInfo(IFileDescriptor * fd)
     ICsvParameters * csvInfo = helper.queryCsvParameters();
 
     headerLines = csvInfo->queryHeaderLen();
-    maxDiskSize = csvInfo->queryMaxSize();
     limit = helper.getRowLimit();
     if (helper.getFlags() & TDRlimitskips)
         limit = (unsigned __int64) -1;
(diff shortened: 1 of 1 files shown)


ghalliday (commented):
It looks like Thor doesn't take any notice of the csv option for the maxlength.  For consistency it is probably worth changing so maxRowSize should be the max of the debug option, the csv option and the dfs maxRecordSize,  and use it in both peek calls.

10MB default is rather large and potentially inefficient depending on how the buffering is done.  See comment below...

ecl/hthor/hthor.cpp
```diff
@@ -9553,7 +9558,9 @@ bool CHThorCsvReadActivity::openNext()
         while (lines-- && !inputstream->eos())
         {
             size32_t numAvailable;
-            const void * next = inputstream->peek(maxDiskSize, numAvailable);
+            // Probably the first fixing point. This changes use same value as Thor does
+            // to skip CSV header. 
+            const void * next = inputstream->peek(maxRowSize, numAvailable);
             inputstream->skip(csvSplitter.splitLine(numAvailable, (const byte *)next));
```
ghalliday, line 9564: This is not the right call to make:

This should use the same scheme as thor calling splitLine(steam, max) which dynamically increases the size requetested until the max is reached. Rather than peek on the maximum line length.

ghalliday (changes requested):
Failed to add final comment before submitting.

ghalliday (approved):
Looks good.  Please squash.

ghalliday:
@AttilaVamos also needs to be a non-draft PR and have the checkboxes filled in.

AttilaVamos:
@ghalliday squashed and check boxes are filled. Please merge.

jakesmith (approved):
@AttilaVamos - looks good.

PR #16791, opened 2022-12-16
HPCC-28476 Refactors Jlog Message categories

rpastrana:
- Renames log categories to match audience and class
- Updates disaster category target audience to operator

Signed-off-by: R Pastrana <915949+rpastrana@users.noreply.github.com>

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/jhtree/keydiff.cpp
+++ b/system/jhtree/keydiff.cpp
@@ -710,13 +710,13 @@ public:
     void addDiffSize(size32_t sz) { diffSize += sz; }
     void log() const
     {
-        LOG(MCstats, "Matching rows: %u", stats[CKeyDiff::CMD_MATCH-1]);
-        LOG(MCstats, "Rows close to previous old row: %u", stats[CKeyDiff::CMD_DIFF_OLD_PREV-1]);
-        LOG(MCstats, "Rows close to current old row: %u", stats[CKeyDiff::CMD_DIFF_OLD_CURR-1]);
-        LOG(MCstats, "Rows close to previous new row: %u", stats[CKeyDiff::CMD_DIFF_NEW_PREV-1]);
+        LOG(MCoperatorProgress, "Matching rows: %u", stats[CKeyDiff::CMD_MATCH-1]);
+        LOG(MCoperatorProgress, "Rows close to previous old row: %u", stats[CKeyDiff::CMD_DIFF_OLD_PREV-1]);
+        LOG(MCoperatorProgress, "Rows close to current old row: %u", stats[CKeyDiff::CMD_DIFF_OLD_CURR-1]);
+        LOG(MCoperatorProgress, "Rows close to previous new row: %u", stats[CKeyDiff::CMD_DIFF_NEW_PREV-1]);
         unsigned diffNum = stats[CKeyDiff::CMD_DIFF_OLD_PREV-1] + stats[CKeyDiff::CMD_DIFF_OLD_CURR-1] + stats[CKeyDiff::CMD_DIFF_NEW_PREV-1];
         if(diffNum > 0)
-            LOG(MCstats, "Average diff size: %u", ((diffSize + diffNum/2) / diffNum));
+            LOG(MCoperatorProgress, "Average diff size: %u", ((diffSize + diffNum/2) / diffNum));
     }
 
 private:
@@ -1097,8 +1097,8 @@ public:
 
     virtual void logStats() const
     {
-        LOG(MCstats, "Rows in old index: %u", oldInput.queryCount());
-        LOG(MCstats, "Rows in new index: %u", newInput.queryCount());
+        LOG(MCoperatorProgress, "Rows in old index: %u", oldInput.queryCount());
+        LOG(MCoperatorProgress, "Rows in new index: %u", newInput.queryCount());
         keydiff.logStats();
     }
 
--- a/system/jlib/jlog.hpp
+++ b/system/jlib/jlog.hpp
@@ -81,7 +81,7 @@ typedef enum
  *    DBGLOG([LogMsgCode,] format,..)          - uses MCdebugInfo                       *
  *                                                                                      *
  * 2) For fatal errors or unrecoverable errors:                                         *
- *    DISLOG([LogMsgCode,] format,..)          - uses MCdisaster                        *
+ *    DISLOG([LogMsgCode,] format,..)          - uses MCoperatorDisaster                        *
  *                                                                                      *
  * 3) For warning messages:                                                             *
  *    (i) Messages for End-users (including ECL coders) should use:                     *
@@ -91,7 +91,7 @@ typedef enum
  *        OWARNLOG([LogMsgCode,] format,..)    - uses MCoperatorWarning                 *
  *                                                                                      *
  *    (iii) Messages for platform developers:                                           *
- *        IWARNLOG([LogMsgCode,] format,..)    - uses MCinternalWarning                 *
+ *        IWARNLOG([LogMsgCode,] format,..)    - uses MCdebugWarning                 *
  *                                                                                      *
  * 4) For error messages:                                                               *
  *    (i) Messages for End-users (including ECL coders) should use:                     *
@@ -101,10 +101,10 @@ typedef enum
  *        OERRLOG([LogMsgCode,] format,..)     - uses MCoperatorError                   *
  *                                                                                      *
  *    (iii) Messages for platform developers:                                           *
- *        IERRLOG([LogMsgCode,] format,..)     - uses MCinternalError                   *
+ *        IERRLOG([LogMsgCode,] format,..)     - uses MCdebugError                   *
  *                                                                                      *
  *    (iv) Messages for audit:                                                          *
- *        AERRLOG([LogMsgCode,] format,..)     - uses MCinternalError                   *
+ *        AERRLOG([LogMsgCode,] format,..)     - uses MCdebugError                   *
  *                                                                                      *
  *                                                                                      *
  * 5) For progress messages:                                                            *
@@ -797,14 +797,14 @@ extern jlib_decl ILogMsgHandler * attachLogMsgMonitorFromPTree(IPropertyTree * t
 extern jlib_decl void attachManyLogMsgMonitorsFromPTree(IPropertyTree * tree);            // Takes tree containing many <monitor> elements
 
 // Standard categories and unknown jobInfo
-constexpr LogMsgCategory MCdisaster(MSGAUD_all, MSGCLS_disaster, FatalMsgThreshold);
+constexpr LogMsgCategory MCoperatorDisaster(MSGAUD_operator, MSGCLS_disaster, FatalMsgThreshold);
 constexpr LogMsgCategory MCuserError(MSGAUD_user, MSGCLS_error, ErrMsgThreshold);
 constexpr LogMsgCategory MCoperatorError(MSGAUD_operator, MSGCLS_error, ErrMsgThreshold);
-constexpr LogMsgCategory MCinternalError(MSGAUD_programmer, MSGCLS_error, ErrMsgThreshold);
+constexpr LogMsgCategory MCdebugError(MSGAUD_programmer, MSGCLS_error, ErrMsgThreshold);
 constexpr LogMsgCategory MCauditError(MSGAUD_audit, MSGCLS_error, ErrMsgThreshold);
 constexpr LogMsgCategory MCuserWarning(MSGAUD_user, MSGCLS_warning, WarnMsgThreshold);
 constexpr LogMsgCategory MCoperatorWarning(MSGAUD_operator, MSGCLS_warning, WarnMsgThreshold);
-constexpr LogMsgCategory MCinternalWarning(MSGAUD_programmer, MSGCLS_warning, WarnMsgThreshold);
+constexpr LogMsgCategory MCdebugWarning(MSGAUD_programmer, MSGCLS_warning, WarnMsgThreshold);
 constexpr LogMsgCategory MCauditWarning(MSGAUD_audit, MSGCLS_warning, WarnMsgThreshold);
 constexpr LogMsgCategory MCuserProgress(MSGAUD_user, MSGCLS_progress, ProgressMsgThreshold);
 constexpr LogMsgCategory MCoperatorProgress(MSGAUD_operator, MSGCLS_progress, ProgressMsgThreshold);
@@ -812,10 +812,8 @@ constexpr LogMsgCategory MCdebugProgress(MSGAUD_programmer, MSGCLS_progress, Deb
 constexpr LogMsgCategory MCuserInfo(MSGAUD_user, MSGCLS_information, InfoMsgThreshold);
 constexpr LogMsgCategory MCdebugInfo(MSGAUD_programmer, MSGCLS_information, DebugMsgThreshold);
 constexpr LogMsgCategory MCauditInfo(MSGAUD_audit, MSGCLS_information, AudMsgThreshold);
-constexpr LogMsgCategory MCstats(MSGAUD_operator, MSGCLS_progress, ProgressMsgThreshold);
 constexpr LogMsgCategory MCoperatorInfo(MSGAUD_operator, MSGCLS_information, InfoMsgThreshold);
-constexpr LogMsgCategory MCmetrics(MSGAUD_operator, MSGCLS_metric, ErrMsgThreshold);
-constexpr LogMsgCategory MCExtraneousInfo(MSGAUD_programmer, MSGCLS_information, ExtraneousMsgThreshold);
+constexpr LogMsgCategory MCoperatorMetric(MSGAUD_operator, MSGCLS_metric, ErrMsgThreshold);
 
 /*
  * Function to determine log level (detail) for exceptions, based on log message class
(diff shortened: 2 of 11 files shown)


rpastrana:
@ghalliday please focus on the changes here:
https://github.com/hpcc-systems/HPCC-Platform/blob/425d083310a2bd1be4dfadb769506ac1da120274/system/jlib/jlog.hpp#L800

The rest of the changes are side effects of the refactoring of pre-existing references.

system/jlib/jlog.hpp
```diff
@@ -845,9 +845,10 @@ inline LogMsgCategory MCexception(IException * e, LogMsgClass cls = MSGCLS_error
     return LogMsgCategory((e)->errorAudience(),cls, mapClassToDefaultDetailLevel(cls));
 }
 
-#define MCerror MCuserError
-#define MCwarning MCuserWarning
-#define MCprogress MCuserProgress
+#define MCstats LogCategory_OperatorWarning
```
ghalliday, line 848: Not really an equivalent category, and we shouldn't be  adding more #defines - should delete them if anyything because they don't match the pattern of (audience, category)

ghalliday (changes requested):

common/dllserver/dllserver.cpp
```diff
@@ -632,7 +632,7 @@ ILoadedDllEntry * DllServerBase::doLoadDll(const char * name, DllLocationType ty
     location->getDllFilename(rfile);
     StringBuffer x;
     rfile.getPath(x);
-    LOG(MCdebugInfo, unknownJob, "Loading dll (%s) from location %s", name, x.str());
+    LOG(LogCategory_ProgrammerInfo, unknownJob, "Loading dll (%s) from location %s", name, x.str());
```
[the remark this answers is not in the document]
rpastrana: Why change? For accurate self documentation. The label prefix should properly reflect the MC prefix doesn't map
What is the guideline for "too long" and why choose brevity over readability/maintainability?
Too long to type? All IDEs I'm aware of provide intellisense type features which prevent the developer from typing out the entire label, but even so, a few extra chars is a sensible sacrifice for the sake of maintainability and search-ability.
LogCategory is too long to read? I didn't consider that a hindrance, how much weight do we want to apply to that point?
Are there any extra changes because of the new prefix? How many new changes are we willing to accept for the sake of readability?

common/environment/environment.cpp
```diff
@@ -1878,15 +1878,15 @@ bool CLocalEnvironment::getRunInfo(IStringVal & path, IStringVal & dir, const ch
         Owned<IConstMachineInfo> machine = getMachineByAddress(machineaddr);
         if (!machine)
         {
-            LOG(MCdebugInfo, unknownJob, "Unable to find machine for %s", machineaddr);
+            LOG(LogCategory_ProgrammerInfo, unknownJob, "Unable to find machine for %s", machineaddr);
```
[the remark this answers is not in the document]
rpastrana: I think the former is unmistakable and if fact easier to read. Prog could also mean progress.

common/dllserver/dllserver.cpp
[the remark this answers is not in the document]
ghalliday: My concern is with the readability of the source code, particularly with the cognitive effort needed to understand it.  Try an experiment scanning some source code.

LOG(MCdebugInfo) can be read in two glances with very little delay.  It is slightly more work than DBGLOG(), but I don't lose the flow.

LOG(LogCategory_ProgrammerInfo requires 3 or 4.  With (when I read it) a significant pause on "LOG(LogCategory" before I get to the information that is interesting.  The pause is long enough to interrupt my flow in understanding.

LOG(MCprogrammerInfo) v LOG(MCdebugInfo)   is similar, not quite as intrusive.  Because Programmer is a long word I can't read that it is programmer and info in the same glance.  

Most of the time picky details of naming isn't significant enough to comment on.  But with something that is used everywhere, especially when we are planning to replace DBGLOG with LOG(...) it really needs to be right.  And by right I mean as easy to read as possible.  Consistency is important for maintainability and search-ability, spelling things out in full isn't.

common/environment/environment.cpp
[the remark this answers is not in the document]
ghalliday: I agree prog is a bad idea because it could mean progress.

common/dllserver/dllserver.cpp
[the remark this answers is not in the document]
rpastrana: The proposed changes emphasize and improve consistency, search-ability, IntelliSense code completion, and self documentation based on explicit naming convention - all for the sake of maintainability and usability.

It seems that "spelling things out in full isn't [important]." is the actual issue and the argument I see is that reading "LogCategory_ProgrammerInfo" vs "MCdebugInfo" requires 2 to 3 extra glances...

Not sure how that was quantified, but makes one wonder if speed-scanning is being prioritized above other important aspects of maintainability such as self-documentation (which actually can speed up code understanding significantly)? 

Anyway, however you prioritize these details, if they're going to affect code reviews and productivity, I suggest they're defined/quantified and published to avoid these types of conversations.

I concede the naming convention is long, perhaps pedantic, and will condense to single-glance len.

What do you prefer:
LogCatProgmrInfo
LogCatProgInfo
LogCatDbgInfo
MCDebugInfo
LCDebugInfo
LCProgInfo




. If the extra milliseconds/seconds it might require a developer to fully grasp a variable meaning is more important than variable meaning clarity/self documentation, that's fine, but let's make it a well established best-practice for HPCC development in order to avoid these types of miscommunications.



For this particular case, I still don't see how the MC prefix helps,  the average developer, and I would prefer something like , etc.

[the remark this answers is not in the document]
ghalliday: The priority is ease of reading.  That isn't necessarily the same as being fully descriptive.   I want to be able to look at a line of code an understand it with the least effort.  Naming variables/functions is sometimes really hard.

If something is too verbose without providing any extra information then it slows down your comprehension.  When you read text you read it in chunks, so reading

LOG(LogCategory_....

That would be the first chunk - but the LogCategory tells me nothing - since I know every log call is followed by a category.

LOG(debugInfo,

is a similar length, but tells me much more information.  It allows me to concentrate on the rest of the line - the message, which contains the details that need to be understand.

In this context I would prefer no prefix (as above) - since it is always a category, but that would pollute the namespace too much, hence a short 2 character prefix.  LC might be better than MC but they are fairly similar and any improvement doesn't justify the bulk change or need for programmers to change.

This is a slightly extreme case - in that the message categories only occur in a single context.

In this case I would vote for MCdebugInfo.  (We tend to use a lower case letter after an upper case prefix.  See examples in eclhelper.hpp, jfile.hpp etc.)

system/jlib/jlog.hpp
```diff
@@ -797,25 +797,24 @@ extern jlib_decl ILogMsgHandler * attachLogMsgMonitorFromPTree(IPropertyTree * t
...
 constexpr LogMsgCategory MCauditWarning(MSGAUD_audit, MSGCLS_warning, WarnMsgThreshold);
 constexpr LogMsgCategory MCuserProgress(MSGAUD_user, MSGCLS_progress, ProgressMsgThreshold);
 constexpr LogMsgCategory MCoperatorProgress(MSGAUD_operator, MSGCLS_progress, ProgressMsgThreshold);
-constexpr LogMsgCategory MCdebugProgress(MSGAUD_programmer, MSGCLS_progress, DebugMsgThreshold);
+constexpr LogMsgCategory MCprogrammerProgress(MSGAUD_programmer, MSGCLS_progress, DebugMsgThreshold);
 constexpr LogMsgCategory MCuserInfo(MSGAUD_user, MSGCLS_information, InfoMsgThreshold);
-constexpr LogMsgCategory MCdebugInfo(MSGAUD_programmer, MSGCLS_information, DebugMsgThreshold);
+constexpr LogMsgCategory MCprogrammerInfo(MSGAUD_programmer, MSGCLS_information, DebugMsgThreshold);
 constexpr LogMsgCategory MCauditInfo(MSGAUD_audit, MSGCLS_information, AudMsgThreshold);
 constexpr LogMsgCategory MCstats(MSGAUD_operator, MSGCLS_progress, ProgressMsgThreshold);
```
ghalliday, line 815: Hasn't been updated to MCoperatorProgress

system/jlib/jlog.hpp
```diff
@@ -797,25 +797,24 @@ extern jlib_decl ILogMsgHandler * attachLogMsgMonitorFromPTree(IPropertyTree * t
 extern jlib_decl void attachManyLogMsgMonitorsFromPTree(IPropertyTree * tree);            // Takes tree containing many <monitor> elements
 
 // Standard categories and unknown jobInfo
-constexpr LogMsgCategory MCdisaster(MSGAUD_all, MSGCLS_disaster, FatalMsgThreshold);
+constexpr LogMsgCategory MCoperatorDisaster(MSGAUD_operator, MSGCLS_disaster, FatalMsgThreshold);
 constexpr LogMsgCategory MCuserError(MSGAUD_user, MSGCLS_error, ErrMsgThreshold);
-constexpr LogMsgCategory MCoperatorError(MSGAUD_operator, MSGCLS_error, ErrMsgThreshold);
-constexpr LogMsgCategory MCinternalError(MSGAUD_programmer, MSGCLS_error, ErrMsgThreshold);
+constexpr LogMsgCategory MCOperatorError(MSGAUD_operator, MSGCLS_error, ErrMsgThreshold);
```
ghalliday, line 802: Why the leading uppercase O?

ghalliday (changes requested):
I don't understand this change.  The last comment I see in the discussion is:

> LOG(MCdebugInfo, unknownJob, "Unable to find machine for %s", machineaddr);
> ...
> if these I think I prefer the first even though it doesn't match the audience. 
> 
> I think the former is unmistakable and in fact easier to read. Prog could also mean progress.
> 

This PR then changes MCdebugInfo to MCprogrammerInfo which goes against that.  The scale of those changes also makes it very hard to spot any other changes.

rpastrana:
> I don't understand this change. The last comment I see in the discussion is:
> 
> > LOG(MCdebugInfo, unknownJob, "Unable to find machine for %s", machineaddr);
> > ...
> > if these I think I prefer the first even though it doesn't match the audience.
> > I think the former is unmistakable and in fact easier to read. Prog could also mean progress.
> 
> This PR then changes MCdebugInfo to MCprogrammerInfo which goes against that. The scale of those changes also makes it very hard to spot any other changes.

you stated:
>Also not convinced about Programmer v Prog.

To which I replied:
>I think the former is unmistakable and in fact easier to read. Prog could also mean progress.

The former being Programmer - which is easier to read, and cannot be mistaken as progress.

To which you agreed:
>I agree prog is a bad idea because it could mean progress.

If you want it changed again to MCdebugInfo, so be it. 

However in my very humble opinion, the scale of changes should hold very little priority vs. the effectiveness of a change. Unless that's another priority you would like to impose, and in that case it would be very helpful to define the priorities and publish them, again, to avoid these types of situations.

system/jlib/jlog.hpp
rpastrana: clearly an unintentional side effect of reverting from a more standard format to the preferred format, will change.

system/jlib/jlog.hpp
rpastrana: Line 810 already declares MCoperatorProgress, that's why I originally kept MCstats as an alias.
If we oppose the use of aliases, Line815 should be removed, and all current references to MCstats would target the variable on line 810.

rpastrana:
@ghalliday will split large refactoring commit into 2 next time.
Latest commit removes the MCStats alias.

system/jlib/jmisc.cpp
```diff
@@ -252,7 +252,7 @@ jlib_decl void PrintMemoryStatusLog()
 #ifdef _WIN32
     MEMORYSTATUS mS;
     GlobalMemoryStatus(&mS);
-    LOG(MCdebugInfo, unknownJob, "Available Physical Memory = %dK", (unsigned)(mS.dwAvailPhys/1024));
+    LOG(LogCategory_ProgrammerInfo, unknownJob, "Available Physical Memory = %dK", (unsigned)(mS.dwAvailPhys/1024));
```
ghalliday, line 255: These need updating/reverting  (see failure in github actions)

thorlcr/master/thmastermain.cpp
```diff
@@ -955,8 +955,8 @@ int main( int argc, const char *argv[]  )
         thorJob.setJobID(thorJobId);
         setDefaultJobId(thorJobId);
         StringBuffer thorEpStr;
-        LOG(MCdebugProgress, thorJob, "ThorMaster version %d.%d, Started on %s", THOR_VERSION_MAJOR,THOR_VERSION_MINOR,thorEp.getUrlStr(thorEpStr).str());
-        LOG(MCdebugProgress, thorJob, "Thor name = %s, queue = %s, nodeGroup = %s",thorname,queueName.str(),nodeGroup.str());
+        LOG(LogCategory_ProgrammerProgress, thorJob, "ThorMaster version %d.%d, Started on %s", THOR_VERSION_MAJOR,THOR_VERSION_MINOR,thorEp.getUrlStr(thorEpStr).str());
```
ghalliday, line 958: Needs to revert.

thorlcr/slave/thslavemain.cpp
```diff
@@ -411,7 +411,7 @@ int main( int argc, const char *argv[]  )
             getMPServer()->setOpt(mpsopt_channelreopen, "true");
 #ifdef USE_MP_LOG
         startLogMsgParentReceiver();
-        LOG(MCdebugProgress, thorJob, "MPServer started on port %d", getFixedPort(TPORT_mp));
+        LOG(MCdebugInfo, thorJob, "MPServer started on port %d", getFixedPort(TPORT_mp));
```
ghalliday, line 414: I think this category was correct.  I think progress should  correspond to logging where the timing is important (an Event), rather than whether something has happened is important.

thorlcr/slave/thslavemain.cpp
```diff
@@ -498,9 +498,9 @@ int main( int argc, const char *argv[]  )
             ULARGE_INTEGER free;
             if (GetDiskFreeSpaceEx("c:\\",&userfree,&total,&free)&&total.QuadPart) {
                 unsigned pc = (unsigned)(free.QuadPart*100/total.QuadPart);
-                LOG(MCdebugProgress, thorJob, "Total disk space = %" I64F "d k", total.QuadPart/1000);
-                LOG(MCdebugProgress, thorJob, "Free  disk space = %" I64F "d k", free.QuadPart/1000);
-                LOG(MCdebugProgress, thorJob, "%d%% disk free\n",pc);
+                LOG(MCdebugInfo, thorJob, "Total disk space = %" I64F "d k", total.QuadPart/1000);
```
ghalliday, line 501: I'm happy with these changing...

thorlcr/thorutil/thbuf.cpp
```diff
@@ -778,7 +778,7 @@ class CSharedWriteAheadBase : public CSimpleInterface, implements ISharedSmartBu
         if (readersWaiting)
         {
 #ifdef TRACE_WRITEAHEAD
-            ActPrintLogEx(&activity->queryContainer(), thorlog_all, MCdebugProgress, "signalReaders: %d", readersWaiting);
+            ActPrintLogEx(&activity->queryContainer(), thorlog_all, LogCategory_ProgrammerProgress, "signalReaders: %d", readersWaiting);
```
ghalliday, line 781: More changes that need to be reverted.

ghalliday (changes requested):
@rpastrana looking better.  The removal of internal looks like a helpful change - I would not have guessed the meaning.  A few bits left over from previous changes to correct.

system/jhtree/keydiff.cpp
```diff
@@ -710,13 +710,13 @@ class KeyDiffStats
...
     void log() const
     {
-        LOG(MCstats, "Matching rows: %u", stats[CKeyDiff::CMD_MATCH-1]);
-        LOG(MCstats, "Rows close to previous old row: %u", stats[CKeyDiff::CMD_DIFF_OLD_PREV-1]);
-        LOG(MCstats, "Rows close to current old row: %u", stats[CKeyDiff::CMD_DIFF_OLD_CURR-1]);
-        LOG(MCstats, "Rows close to previous new row: %u", stats[CKeyDiff::CMD_DIFF_NEW_PREV-1]);
+        LOG(MCoperatorWarning, "Matching rows: %u", stats[CKeyDiff::CMD_MATCH-1]);
+        LOG(MCoperatorWarning, "Rows close to previous old row: %u", stats[CKeyDiff::CMD_DIFF_OLD_PREV-1]);
+        LOG(MCoperatorWarning, "Rows close to current old row: %u", stats[CKeyDiff::CMD_DIFF_OLD_CURR-1]);
+        LOG(MCoperatorProgress, "Rows close to previous new row: %u", stats[CKeyDiff::CMD_DIFF_NEW_PREV-1]);
```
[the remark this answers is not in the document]
rpastrana: MCStats was replaced by MCoperatorProgress, assuming that's what you meant (not MCoperatorInfo). Let me know, if you wanted to change it to Info.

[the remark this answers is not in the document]
ghalliday: It is fine to leave it, but I in this case MCoperatorInfo would be the correct category.  An example where using an alias for a category confuses rather than helps.

ghalliday (approved):
@rpastrana looks good.  Please squash.

rpastrana:
@ghalliday ready for merge

PR #16796, opened 2022-12-20
HPCC-28651 Disable multicast by default in bare-metal

ghalliday:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/deployment/envgen2/EnvGen.cpp
+++ b/deployment/envgen2/EnvGen.cpp
@@ -989,7 +989,7 @@ void CEnvGen::cloudConfiguration(IPropertyTree * config, const char* cloud)
    if (!stricmp(cloud, "aws"))
    {
        createUpdateTask("modify", config, "sw:espsmc@enableSystemUseRewrite=true|false");
-       createUpdateTask("modify", config, "sw:roxie@roxieMulticastEnabled=false");
+       createUpdateTask("modify", config, "sw:roxie@roxieIsMulticastEnabled=false");
    }
 }
 
--- a/initfiles/componentfiles/configschema/xsd/roxie.xsd
+++ b/initfiles/componentfiles/configschema/xsd/roxie.xsd
@@ -299,8 +299,8 @@
                 </xs:attributeGroup>
 
                 <xs:attributeGroup name="udp" hpcc:groupByName="UDP">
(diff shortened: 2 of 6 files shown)


ghalliday:
@richardkchapman @mckellyln pushed to provoke a discussion....

mckellyln (approved):
Looks good.

Just an fyi, there are a few places such as:
in ccdmain:
```
            IWARNLOG("multicastTTL value (%d) invalid, must be >=0, resetting to %u", ttlTmp, multicastTTL);
```
and in ccdqueue is:
```
            DBGLOG("Roxie: multicastTTL: %u", multicastTTL);
```
these should not log or not log "multicast" at least if it is disabled.

richardkchapman:
This change will have the documented effect, but what the impact will be on bare metal performance is hard to predict. I suspect most bare metal systems will actually be unaffected unless you rename the option, since the setting is likely burned in to existing environment.xml files

richardkchapman (approved):

richardkchapman:
This change will have the documented effect, but what the impact will be on bare metal performance is hard to predict. I suspect most bare metal systems will actually be unaffected unless you rename the option, since the setting is likely burned in to existing environment.xml files

richardkchapman:
Also should be noted that multicast isn't just used for the outbound message from server to agent - it's also used for IBYTI and potentially sniffer flow control (though I think the latter may have been removed)

ghalliday:
@richardkchapman I think I did rename it.   I may need to check all the xsl files etc copy it across correctly

ghalliday:
@richardkchapman pushed a commit to update the other configuration files.  The main question is do we think this is a good idea?

richardkchapman:
Changing bare metal to be closer to cloud means the cloud setup is less of an unknown, but also means that we risk breaking our "fallback" bare metal config.

In this instance operation with multicast off is well-enough known that I don't think it's a big deal either way. On that basis I'd be inclined not to bother, but it's up to you.

ghalliday:
Needs testing and probably shouldn't change the option name.  Closing for now.

PR #16800, opened 2023-01-03
HPCC-28685 Thor, spill/temp usage status/aggregation

shamser:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/common/workunit/workunit.cpp
+++ b/common/workunit/workunit.cpp
@@ -2695,6 +2695,39 @@ cost_type aggregateDiskAccessCost(const IConstWorkUnit * wu, const char *scope)
     }
     return totalCost;
 }
+
+void gatherSpillSize(const IConstWorkUnit * wu, const char *scope, stat_type & totalSizeSpill, stat_type & peakSizeSpill)
+{
+    WuScopeFilter filter;
+    if (!isEmptyString(scope))
+        filter.addScope(scope);
+    else
+        filter.addScope("");
+    filter.setIncludeNesting(2);
+    filter.addSource("global");
+    filter.addSource("statistics");
+    filter.addOutputStatistic(StSizeSpillFile);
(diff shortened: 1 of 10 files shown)


ghalliday (changes requested):
@shamser a few comments on the code, but I haven't finished because I don't think the logic is correct (although I may have mis-interpreted).

It seems to be taking the peak spill size as the highest individual spill size within a subgraph.  That isn't correct - because multiple spills may have been active at the the same time.

Also, the important spill size also includes the size of the inter-subgraph spills.  You can't calculate the peak by summing the individual stats - because the inter subgraph spill files are deleted once they are no longer needed.

I think it needs implementing something along the following lines:
i) maintain a current active-spill-size in the master.
ii) When a subgraph completes, total spill size = size of all spills in a subgraph + active-spill-size.  
Publish that as the high water mark (StSizeTotalSpillFile) of the spill usage (not the peak).
iii) active spill size is increased by any new inter-graph **output** spills and decreased by any spills that are no longer needed and have been removed.  (Thor almost certainly has a structure for maintaining this information already).

Peak is the max value of StSizeTotalSpillFile for any nested SubGraph, which could be published on a graph as a peak (not sure if it should use a different stat or not, probably ok as is).  It doesn't make any sense to publish StSizeSpillFile on a graph.  It would possibly make sense to publish on a subgraph as a sum of all the spills for that sub-graph.

ghalliday:
@jakesmith see comments, added you as a second reviewer.

shamser:
Created new sub-tasks created for this jira: 1) inter-subgraphs spills 2) activity related spills stats

PR #16804, opened 2023-01-04
HPCC-28616 Always acknowledge from worker to agent to spot lost packets

richardkchapman:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Cloud-compatibility
  - [x] Premature optimization

## Smoketest:

## Testing:
Added tracing to observe behaviour, including behaviour when initial ack is lost

### DIFF
--- a/roxie/ccd/ccdqueue.cpp
+++ b/roxie/ccd/ccdqueue.cpp
@@ -503,6 +503,8 @@ protected:
     unsigned smartStepInfoLength = 0;
     const byte *contextData = nullptr;
     unsigned contextLength = 0;
+    std::atomic<unsigned> timeFirstSent = 0;
+    std::atomic<bool> acknowledged = false;
     
 public:
     IMPLEMENT_IINTERFACE;
@@ -539,6 +541,7 @@ public:
         assertex(lengthRemaining >= 0);
         contextData = finger;
         contextLength = lengthRemaining;
+        noteTimeSent();
     }
 
     virtual RoxiePacketHeader &queryHeader() const
@@ -635,6 +638,29 @@ public:
             mb.append(length, data);
         }
         return createSerializedRoxiePacket(mb);
+    }
+
+    virtual void noteTimeSent() override
+    {
+        acknowledged = false;
+        timeFirstSent = msTick();
+        if (!timeFirstSent)
+            timeFirstSent = 1;
+    }
+
+    virtual void setAcknowledged() override
+    {
+        acknowledged = true;
+    }
+
+    virtual bool isAcknowledged() const override
+    {
+        return acknowledged;
+    }
+
+    virtual bool resendNeeded(unsigned timeout, unsigned now) const override
+    {
+        return timeFirstSent && !acknowledged && now-timeFirstSent > timeout;
     }
 };
 
@@ -2022,6 +2048,7 @@ public:
                 Owned<ISerializedRoxieQueryPacket> serialized = packet->serialize();
                 if (!channelWrite(serialized->queryHeader(), true))
                     DBGLOG("Roxie packet write wrote too little");
+                packet->noteTimeSent();
                 packetsSent++;
             }
             catch (StoppedException *E)
@@ -2412,6 +2439,7 @@ public:
                 StringBuffer s;
                 throw MakeStringException(ROXIE_PACKET_ERROR, "Maximum packet length %d exceeded sending packet %s", maxPacketSize, header.toString(s).str());
             }
+            x->noteTimeSent();
             Owned <ISerializedRoxieQueryPacket> serialized = x->serialize();
             if (!channelWrite(serialized->queryHeader(), true))
                 logctx.CTXLOG("Roxie packet write wrote too little");
@@ -2624,7 +2652,7 @@ public:
                 // Turn broadcast packet (channel 0), as early as possible, into non-0 channel packets.
                 // So retries and other communication with Roxie server (which uses non-0 channel numbers) will not cause double work or confusion.
                 // Unfortunately this is bad news for dropping packets
-            // In SUBCHANNELS_IN_HEADER mode this translation has been done on server before sending, except for some control messages like PING or UNLOAD
+                // In SUBCHANNELS_IN_HEADER mode this translation has been done on server before sending, except for some control messages like PING or UNLOAD
 
                 Owned<const ITopologyServer> topology = getTopology();
                 const std::vector<unsigned> channels = topology->queryChannels();
@@ -2667,6 +2695,22 @@ public:
 #endif
                 Owned<ISerializedRoxieQueryPacket> packet = createSerializedRoxiePacket(mb);
                 unsigned retries = header.thisChannelRetries(mySubchannel);
+                if (acknowledgeAllRequests)
+                {
+#ifdef DEBUG
+                    if (testAgentFailure & 0x1 && !retries)
+                        return;
+#endif
+                    AgentContextLogger logctx(packet);
+                    if (doTrace(traceRoxiePackets))
+                    {
+                        StringBuffer buf;
+                        DBGLOG("Sending ROXIE_ALIVE to acknowledge request %s", header.toString(buf).str());
+                    }
+                    RoxiePacketHeader newHeader(header, ROXIE_ALIVE, mySubchannel);
+                    Owned<IMessagePacker> output = ROQ->createOutputStream(newHeader, true, logctx);
+                    output->flush();
+                }
                 if (retries)
                 {
                     // MORE - is this fast enough? By the time I am seeing retries I may already be under load. Could move onto a separate thread
@@ -2676,7 +2720,7 @@ public:
 
                     AgentContextLogger logctx(packet);
                     // Send back an out-of-band immediately, to let Roxie server know that channel is still active
-                    if (!(testAgentFailure & 0x800))
+                    if (!(testAgentFailure & 0x800) && !acknowledgeAllRequests)
                     {
                         RoxiePacketHeader newHeader(header, ROXIE_ALIVE, mySubchannel);
                         Owned<IMessagePacker> output = ROQ->createOutputStream(newHeader, true, logctx);
--- a/roxie/ccd/ccdserver.cpp
+++ b/roxie/ccd/ccdserver.cpp
@@ -4070,10 +4070,20 @@ class CRemoteResultAdaptor : implements IEngineRowStream, implements IFinalRoxie
         }
     }
 
-    void retryPending()
+    void retryPending(unsigned timeout)
     {
-        CriticalBlock b(pendingCrit);
         checkDelayed();
+        unsigned now = 0;
+        if (timeout)
+        {
+            if (doTrace(traceRoxiePackets))
+                DBGLOG("Checking %d pending packets for ack status", pending.ordinality());
+            now = msTick();
+            if (now-lastRetryCheck < timeout/4)
+                return;
+            lastRetryCheck = now;
+        }
+        CriticalBlock b(pendingCrit);
         ForEachItemIn(idx, pending)
         {
             IRoxieServerQueryPacket &p = pending.item(idx);
@@ -4082,6 +4092,12 @@ class CRemoteResultAdaptor : implements IEngineRowStream, implements IFinalRoxie
                 IRoxieQueryPacket *i = p.queryPacket();
                 if (i)
                 {
+                    if (timeout)
+                    {
+                        if (!i->resendNeeded(timeout, now))
+                            continue;
+                        activity.queryLogCtx().CTXLOG("Input has not been acknowledged for %u ms - retry required?", timeout);
+                    }
                     if (!i->queryHeader().retry())
                     {
                         StringBuffer s;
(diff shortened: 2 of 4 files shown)


roxie/ccd/ccdqueue.cpp
```diff
@@ -636,6 +639,29 @@ class CRoxieQueryPacket : public CRoxieQueryPacketBase, implements IRoxieQueryPa
         }
         return createSerializedRoxiePacket(mb);
     }
+
+    virtual void noteTimeSent() override
+    {
+        acknowledged = false;
+        timeFirstSent = msTick();
+        if (!timeFirstSent)
```
mckellyln, line 647: could msTick() ever return 0 ?

roxie/ccd/ccdserver.cpp
```diff
@@ -4082,6 +4084,12 @@ class CRemoteResultAdaptor : implements IEngineRowStream, implements IFinalRoxie
                 IRoxieQueryPacket *i = p.queryPacket();
                 if (i)
                 {
+                    if (timeout)
+                    {
+                        if (!i->resendNeeded(timeout))
+                            continue;
+                        activity.queryLogCtx().CTXLOG("Input has not been acknowledged for %u ms - retry required?", timeout);
```
mckellyln, line 4091: could this ever get logged too often ?

roxie/ccd/ccdserver.cpp
```diff
@@ -5210,11 +5219,15 @@ class CRemoteResultAdaptor : implements IEngineRowStream, implements IFinalRoxie
...
+                if (acknowledgeAllRequests)
+                {
+                    retryPending(checkInterval);
+                }
+                else if (!anyActivity && !localAgent && (timeNow-lastActivity >= checkInterval))
                 {
                     lastActivity = timeNow;
-                    activity.queryLogCtx().CTXLOG("Input has stalled for %u ms - retry required?", timeout);
-                    retryPending();
+                    activity.queryLogCtx().CTXLOG("Input has stalled for %u ms - retry required?", checkInterval);
```
mckellyln, line 5229: could this get logged too often ?  (now that checkInterval is 1/10th of timeout)

mckellyln (commented):
Looks very good, just a few comments.

roxie/ccd/ccdserver.cpp
richardkchapman: If a network is very sick, or a timeout is too low... A general mechanism for spotting excessive logging and deduping might be good, or we can roll our own for each instance I suppose...

roxie/ccd/ccdqueue.cpp
mckellyln: ok, could wrap ...

roxie/ccd/ccdqueue.cpp
```diff
@@ -636,6 +639,29 @@ class CRoxieQueryPacket : public CRoxieQueryPacketBase, implements IRoxieQueryPa
...
+    }
+
+    virtual bool isAcknowledged() const override
+    {
+        return acknowledged;
+    }
+
+    virtual bool resendNeeded(unsigned timeout) const override
+    {
+        return timeFirstSent && !acknowledged && msTick()-timeFirstSent > timeout;
```
ghalliday, line 663: Would it be be better to pass in an expiry time, so that msTick() is called once in the caller, rather than potentially multiple times.  A trade off in likelihoods.  Are there likely to be a reasonable number of packets that have been sent but timeout has not happened?

It could be optimized by passing in a timeout class that called and cached msTick() on demand if required.  I doubt that is worth it though.

roxie/ccd/ccdqueue.cpp
```diff
@@ -503,6 +503,8 @@ class CRoxieQueryPacket : public CRoxieQueryPacketBase, implements IRoxieQueryPa
     unsigned smartStepInfoLength = 0;
     const byte *contextData = nullptr;
     unsigned contextLength = 0;
+    unsigned timeFirstSent = 0;
```
ghalliday, line 506: I suspect both of these should be atomic.

roxie/ccd/ccdserver.cpp
```diff
@@ -5210,11 +5219,15 @@ class CRemoteResultAdaptor : implements IEngineRowStream, implements IFinalRoxie
             else
             {
                 unsigned timeNow = msTick();
-                if (!anyActivity && !localAgent && (timeNow-lastActivity >= timeout))
+                if (acknowledgeAllRequests)
```
ghalliday, line 5222: Should this also be nested in an if (localAgent)?

roxie/ccd/ccdserver.cpp
```diff
@@ -5210,11 +5219,15 @@ class CRemoteResultAdaptor : implements IEngineRowStream, implements IFinalRoxie
...
+                {
+                    retryPending(checkInterval);
+                }
+                else if (!anyActivity && !localAgent && (timeNow-lastActivity >= checkInterval))
                 {
                     lastActivity = timeNow;
-                    activity.queryLogCtx().CTXLOG("Input has stalled for %u ms - retry required?", timeout);
-                    retryPending();
+                    activity.queryLogCtx().CTXLOG("Input has stalled for %u ms - retry required?", checkInterval);
+                    retryPending(0);
```
ghalliday, line 5230: does this mean that retries will be send 10x quicker than before?

Is the following the difference?
if acknowledgeAllRequests is true:
  if there has been no acknowledgement for a packet within the checkInterval then it will be re-sent.

if acknowledgeAllRequests is false:
  if there is no activity in the checkInterval then all packets will request to be resent.

If so it is probably worth documenting somewhere.

What is the difference if acknowledgeAll is set

ghalliday (commented):
@richardkchapman a few questions/comments

roxie/ccd/ccdqueue.cpp
```diff
@@ -539,6 +541,7 @@ class CRoxieQueryPacket : public CRoxieQueryPacketBase, implements IRoxieQueryPa
         assertex(lengthRemaining >= 0);
         contextData = finger;
         contextLength = lengthRemaining;
+        noteTimeSent();
```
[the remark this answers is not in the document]
richardkchapman: Yes, that is the intention. There are some complications with channel 0 (which is added to pending but not sent as such) that is made a lot easier with this call.

roxie/ccd/ccdserver.cpp
mckellyln: Yes 10x quicker retries.   I think as it is now - if the query timeout is less than the HIGH/LOW timeout then no retries at all.
And we still have the bug where queries copied always get LOW priority.

roxie/ccd/ccdserver.cpp
richardkchapman: Probably wouldn't hurt... I will push a fix

roxie/ccd/ccdserver.cpp
richardkchapman: The 10x faster could be viewed as a separate though related fix - it's what the changes in HPCC-25452 were supposed to do, but failed...

acknowledgeAllRequests enables the new behaviour of resending any request that has not seen an ack, rather than requiring ALL requests to stall for the timeout period before any get resent. It's not an option I intend to document though - the option is there to enable us to test and observe the difference (and to disable should it cause unanticipated problems) rather than for end-users to decide which they prefer...

roxie/ccd/ccdserver.cpp
```diff
@@ -5207,14 +5216,18 @@ class CRemoteResultAdaptor : implements IEngineRowStream, implements IFinalRoxie
                     }
                 }
             }
-            else
+            else if (!localAgent)
```
mckellyln, line 5219: Why have an else and possibly do the retry only in the else clause - why not always check at a regular interval (implies not calling msTick() so often but some sort of timer/poll) and then if there are lots of other msgs/activity retries are still checked at a regular interval ?

mckellyln (commented):
I just have one more comment/question about the larger for(;;) loop structure and when we check for retries.

roxie/ccd/ccdserver.cpp
richardkchapman: In the acknowledgeAll case, I do... Leaving the old code closer to how it used to behave simply because it's only there as a fall-back in case of unanticipated issues in the new...

mckellyln: Maybe I am reading the code/loop incorrectly and if so I apologize, but I think I see -
```
for(;;)
{
    mr = ...
    if (mr)
    {
       ...
    }
    else if (!localAgent)
    {
        if(acknowledgeAllRequests)
            retryPending(checkInterval)
    }
}
```
and perhaps its just not an issue, but if handing msgs often (mr != nullptr) then the else that calls retryPending() is not called at some minimum guaranteed interval ?

mckellyln: The
if (mr)
code path can return if there are continuations?, so if we do make any changes to this loop, perhaps the
retryPending(checkInterval) 
call should be made before the if(mr) check ?
Anyway let me know if this sort of change is worth any more consideration.

ghalliday (commented):
@richardkchapman
I think Mark's comment is valid.  The code

```
                if (acknowledgeAllRequests)
                    retryPending(checkInterval);
```
would be better processed whether or not a message was received, ie outside the if(mr) else.  Otherwise, when the system is heavily loaded the retry messages may not be sent.

I think it would also be worth changing the code slightly so that "timeNow = msTick()" is unconditional before updating lastActivity, and that value is then passed into retryPending() to prevent re-evaluation for each packet.  

It might be worth combining with the new retry code with the checkDelayed() function - but I haven't looked at the code to see if that is at all sensible.

richardkchapman:
I had made a change as Mark suggested, but looks like I forgot to push it.

Pushed now, along with another one to try to reduce the calls to msTick()

roxie/ccd/ccdserver.cpp
```diff
@@ -5207,16 +5226,20 @@ class CRemoteResultAdaptor : implements IEngineRowStream, implements IFinalRoxie
...
+                if (timeNow-lastActivity >= checkInterval)
                 {
                     lastActivity = timeNow;
-                    activity.queryLogCtx().CTXLOG("Input has stalled for %u ms - retry required?", timeout);
-                    retryPending();
+                    activity.queryLogCtx().CTXLOG("Input has stalled for %u ms - retry required?", checkInterval);
+                    retryPending(0);
                 }
             }
+            if (acknowledgeAllRequests && !localAgent)
```
mckellyln, line 5239: Could this if block go before the if(mr) (at line 4945) ?

mckellyln (commented):
Looks very good, I just have one ques if it would be good to move retryPending() to before the if(mr) ?

roxie/ccd/ccdserver.cpp
richardkchapman: It could, but I'm not sure it should. Would mean for example that you would do a resend before discovering that the message you just received was the ack you were waiting for...

mckellyln: ok, good point.  I was concerned with the case where the if (mr) block returns and then retryPending() is not checked/called.
But I dont know how much of an issue that really is or if we could just also check retryPending() before the return.
Or just leave it and merge as is.

richardkchapman: Perhaps I should move it to the top of the loop then, before calling getNextResult()

mckellyln (approved):
Looks good to merge.

ghalliday (approved):

roxie/ccd/ccdserver.cpp
```diff
@@ -4082,6 +4092,12 @@ class CRemoteResultAdaptor : implements IEngineRowStream, implements IFinalRoxie
                 IRoxieQueryPacket *i = p.queryPacket();
                 if (i)
                 {
+                    if (timeout)
+                    {
+                        if (!i->resendNeeded(timeout, now))
+                            continue;
+                        activity.queryLogCtx().CTXLOG("Input has not been acknowledged for %u ms - retry required?", timeout);
```
mckellyln, line 4099: I think we could put a traceLevel check on this logging so it could be controlled.

roxie/ccd/ccdserver.cpp
```diff
@@ -5207,14 +5231,14 @@ class CRemoteResultAdaptor : implements IEngineRowStream, implements IFinalRoxie
...
+            else if (!anyActivity && !localAgent && !acknowledgeAllRequests)
             {
                 unsigned timeNow = msTick();
-                if (!anyActivity && !localAgent && (timeNow-lastActivity >= timeout))
+                if (timeNow-lastActivity >= checkInterval)
                 {
                     lastActivity = timeNow;
-                    activity.queryLogCtx().CTXLOG("Input has stalled for %u ms - retry required?", timeout);
-                    retryPending();
+                    activity.queryLogCtx().CTXLOG("Input has stalled for %u ms - retry required?", checkInterval);
```
mckellyln, line 5240: Same here - a traceLevel check might be a good idea ?

PR #16818, opened 2023-01-06
HPCC-28454 Add IFile rename retry logic

jakesmith:
Allow IFile::rename to retry on failure.
This is mainly to cope/diagnose an issue with failing renames on Azure file systems.
By default the retry count will be 0, i.e. behave as before, except in Thor where default will be 10.
Configurable per job, at a Thor level, or globally.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/helm/hpcc/templates/_helpers.tpl
+++ b/helm/hpcc/templates/_helpers.tpl
@@ -247,6 +247,10 @@ cost:
 logAccess:
 {{ toYaml .Values.global.logAccess | indent 2 }}
 {{- end }}
+{{- if .Values.global.expert }}
+expert:
+{{ toYaml .Values.global.expert | indent 2 }}
+{{- end }}
 {{- end -}}
 
 {{/*
--- a/initfiles/componentfiles/configxml/RoxieTopology.xsl
+++ b/initfiles/componentfiles/configxml/RoxieTopology.xsl
@@ -130,6 +130,9 @@
              <storage>
               <xsl:copy-of select="/Environment/Software/RemoteStorage/*"/>
              </storage>
+             <expert>
+              <xsl:copy-of select="/Environment/Software/Global/@*"/>
+             </expert>
             </global>
             <xsl:copy-of select="/Environment/Hardware/NAS"/>
             <xsl:for-each select="ACL">
--- a/system/jlib/jfile.cpp
+++ b/system/jlib/jfile.cpp
@@ -760,6 +760,79 @@ bool CFile::remove()
 #endif
 }
 
+static std::atomic<unsigned> renameRetries{(unsigned)-1}; // initially uninitialized, set on 1st use (in rename)
+
+void setRenameRetries(unsigned numRetries)
+{
+    renameRetries = numRetries;
+}
+
+static unsigned getRenameRetries()
+{
+    unsigned retries;
+    // NB: potentially could be >1 thread here, but that's ok.
+    try
+    {
+        retries = getGlobalConfigSP()->getPropInt("expert/@numRenameRetries");
+    }
+    catch (IException *e) // handle cases where config. not available
+    {
+        EXCLOG(e, "doRename");
+        e->Release();
+        retries = 0;
+    }
+    dbgassertex((unsigned)-1 != retries);
+    setRenameRetries(retries);
+    return retries;
+}
+
+static void doRename(const char *src, const char *dst, const char *callerName)
+{
+    unsigned retriesLeft = renameRetries;
+    unsigned retry = 0;
+    while (true)
+    {
+        if (-1 != ::rename(src, dst))
+            return; // success
+
+        if ((unsigned)-1 == retriesLeft)
+            retriesLeft = getRenameRetries();
+
+        int err = errno; // preserve to use after tracing if exhausted retry attempts
+
+#ifndef _WIN32 // NB: this is primary here to help track down the issue reported in HPCC-28454
+        // why did rename fail? - check that src and path of dst exists (and/or is accessible), and trace it's gid and uid, size and is_dir status.
+        struct stat srcInfo;
+        if (0 != stat(src, &srcInfo))
+            throw makeErrnoExceptionV("%s-doRename - stat SRC(%s)", callerName, src);
+        CDateTime modTime;
+        timetToIDateTime(&modTime, srcInfo.st_mtime);
+        StringBuffer srcModTime;
+        modTime.getString(srcModTime);
+
+        StringBuffer dstPath;
+        splitDirTail(dst, dstPath);
+        struct stat dstInfo;
+        if (0 != stat(dstPath, &dstInfo))
+            throw makeErrnoExceptionV("%s-doRename - stat DST-PATH(%s)", callerName, dstPath.str());
+        timetToIDateTime(&modTime, dstInfo.st_mtime);
+        StringBuffer dstModTime;
+        modTime.getString(dstModTime);
+        DBGLOG("%s-doRename(%s, %s) FAILED (attempt %u)", callerName, src, dst, retry+1);
+        DBGLOG("       src stat[uid=%u, gid=%u, size=%" I64F "u, dir=%s, modtime=%s]", (unsigned)srcInfo.st_uid, (unsigned)srcInfo.st_gid, (offset_t)srcInfo.st_size, boolToStr(S_ISDIR(srcInfo.st_mode)), srcModTime.str());
+        DBGLOG("  dst-path stat[uid=%u, gid=%u, size=%" I64F "u, dir=%s, modtime=%s]", (unsigned)dstInfo.st_uid, (unsigned)dstInfo.st_gid, (offset_t)dstInfo.st_size, boolToStr(S_ISDIR(dstInfo.st_mode)), dstModTime.str());
+#endif
+
+        if (0 == retriesLeft)
+            throw makeErrnoExceptionV(err, "%s(%s, %s)", callerName, src, dst);
+
+        MilliSleep(retry>5 ? 5000 : (100 << retry)); // start at 100ms and double each retry, but cap at 5 seconds
+
+        --retriesLeft;
+        ++retry;
+    }
+}
+
 void CFile::rename(const char *newname)
 {
     // now hopefully newname is just file tail 
@@ -779,8 +852,7 @@ void CFile::rename(const char *newname)
         if (rfn.isLocal())
             dst = rfn.getLocalPath(path.clear()).str();
     }
-    if (-1 == ::rename(filename, dst)) 
-        throw makeErrnoExceptionV("CFile::rename(%s, %s)", filename.get(),dst);
+    doRename(filename, dst, "CFile::rename");
     filename.set(path);
 }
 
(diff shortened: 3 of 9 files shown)


system/jlib/jfile.cpp
```diff
@@ -760,6 +760,69 @@ bool CFile::remove()
...
+    renameRetries = numRetries;
+}
+
+static void doRename(const char *src, const char *dst, const char *errMsg)
+{
+    unsigned retriesLeft = renameRetries;
+    unsigned retry = 0;
+    while (true)
+    {
+        if ((unsigned)-1 == retriesLeft)
```
ghalliday, line 776: Would be a bit cleaner if this was in a getRenameRetries() function which was called to initialise retriesLeft at the start of the function.

system/jlib/jfile.cpp
```diff
@@ -760,6 +760,69 @@ bool CFile::remove()
...
+        if (0 != stat(dstPath, &dstInfo))
+            throw makeErrnoExceptionV("%s-doRename - stat DST-PATH(%s)", errMsg, dstPath.str());
+        timetToIDateTime(&modTime,  dstInfo.st_mtime);
+        StringBuffer dstModTime;
+        modTime.getString(dstModTime);
+        DBGLOG("%s-doRename(%s, %s) FAILED (attempt %u)", errMsg, src, dst, retry+1);
+        DBGLOG("       src stat[uid=%u, gid=%u, size=%" I64F "u, dir=%s, modtime=%s]", (unsigned)srcInfo.st_uid, (unsigned)srcInfo.st_gid, (offset_t)srcInfo.st_size, boolToStr(S_ISDIR(srcInfo.st_mode)), srcModTime.str());
+        DBGLOG("  dst-path stat[uid=%u, gid=%u, size=%" I64F "u, dir=%s, modtime=%s]", (unsigned)dstInfo.st_uid, (unsigned)dstInfo.st_gid, (offset_t)dstInfo.st_size, boolToStr(S_ISDIR(dstInfo.st_mode)), dstModTime.str());
+#endif
+        MilliSleep(100 * 2^retry); // start at 100ms and double each retry
```
ghalliday, line 819: This will xor with retry!  I think what you mean is

`100 << retry`

system/jlib/jfile.cpp
```diff
@@ -760,6 +760,69 @@ bool CFile::remove()
...
+            setRenameRetries(retriesLeft);
+        }
+        if (-1 != ::rename(src, dst))
+            return; // success
+
+        if (0 == retriesLeft)
+            throw makeErrnoExceptionV("%s(%s, %s)", errMsg, src, dst);
+
+#ifndef _WIN32 // NB: this is primary here to help track down the issue report in HPCC-28454
+        struct stat srcInfo;
```
ghalliday, line 799: Could do with a comment describing what it is doing..

helm/hpcc/templates/_helpers.tpl
```diff
@@ -247,6 +247,10 @@ cost:
 logAccess:
 {{ toYaml .Values.global.logAccess | indent 2 }}
 {{- end }}
+{{- if .Values.global.misc }}
```
ghalliday, line 250: misc doesn't seem the most descriptive name, and it is currently used for something unrelated.  I'm not sure that would be better.  "global.options", "global.config", "storage.options"?  The latter may cause problems for bare-metal.
Not sure I have a strong preference, but worth thinking a bit more about - especially if more options might be added later.

system/jlib/jfile.cpp
```diff
@@ -760,6 +760,69 @@ bool CFile::remove()
...
+                e->Release();
+                retriesLeft = 0;
+            }
+            dbgassertex((unsigned)-1 != retriesLeft);
+            setRenameRetries(retriesLeft);
+        }
+        if (-1 != ::rename(src, dst))
+            return; // success
+
+        if (0 == retriesLeft)
```
ghalliday, line 795: Would it be useful to add the tracing even if retries are 0, e.g. for components that haven't changed the option.

system/jlib/jfile.cpp
```diff
@@ -760,6 +760,69 @@ bool CFile::remove()
...
+        timetToIDateTime(&modTime, srcInfo.st_mtime);
+        StringBuffer srcModTime;
+        modTime.getString(srcModTime);
+
+        StringBuffer dstPath;
+        splitDirTail(dst, dstPath);
+        struct stat dstInfo;
+        if (0 != stat(dstPath, &dstInfo))
+            throw makeErrnoExceptionV("%s-doRename - stat DST-PATH(%s)", errMsg, dstPath.str());
+        timetToIDateTime(&modTime,  dstInfo.st_mtime);
```
ghalliday, line 812: trivial: extra space.

system/jlib/jfile.cpp
```diff
@@ -760,6 +760,69 @@ bool CFile::remove()
...
 }
 
+static std::atomic<unsigned> renameRetries{(unsigned)-1}; // initially uninitialized, set on 1st use (in rename)
+
+void setRenameRetries(unsigned numRetries)
+{
+    renameRetries = numRetries;
+}
+
+static void doRename(const char *src, const char *dst, const char *errMsg)
```
ghalliday, line 770: picky: errMsg isn't a great variable name.  maybe callerName or similar.

ghalliday (changes requested):
@jakesmith a few comments about naming/structure, one real issue.

system/jlib/jfile.cpp
jakesmith: whoops, yes, will change.

system/jlib/jfile.cpp
jakesmith: could do. I deliberatly kept it to default 0 and no tracing etc. in normal case, because I was slightly concerned there was exisitng code that might be using rename in a try and see type manner.

That was probably being overly cautious.

I can change it so all components trace.

system/jlib/jfile.cpp
jakesmith: will remove.

system/jlib/jfile.cpp
jakesmith: will change.

system/jlib/jfile.cpp
jakesmith: will add a comment.

system/jlib/jfile.cpp
jakesmith: the rename should have been above this code (I move it in some testing and forgot to move it back),
such that it only when through the renameRetries initialization / read config code on 1st rename failure.
I didn't want it go through config lookup in the std. (successful) case.

I'll change.

jakesmith:
@ghalliday - please see changes.

initfiles/componentfiles/configxml/RoxieTopology.xsl
```diff
@@ -130,6 +130,9 @@
              <storage>
               <xsl:copy-of select="/Environment/Software/RemoteStorage/*"/>
              </storage>
+             <misc>
```
ghalliday, line 133: needs to be expert tag + other files.

ghalliday (commented):
@jakesmith

system/jlib/jfile.cpp
```diff
@@ -760,6 +760,79 @@ bool CFile::remove()
...
+        modTime.getString(dstModTime);
+        DBGLOG("%s-doRename(%s, %s) FAILED (attempt %u)", callerName, src, dst, retry+1);
+        DBGLOG("       src stat[uid=%u, gid=%u, size=%" I64F "u, dir=%s, modtime=%s]", (unsigned)srcInfo.st_uid, (unsigned)srcInfo.st_gid, (offset_t)srcInfo.st_size, boolToStr(S_ISDIR(srcInfo.st_mode)), srcModTime.str());
+        DBGLOG("  dst-path stat[uid=%u, gid=%u, size=%" I64F "u, dir=%s, modtime=%s]", (unsigned)dstInfo.st_uid, (unsigned)dstInfo.st_gid, (offset_t)dstInfo.st_size, boolToStr(S_ISDIR(dstInfo.st_mode)), dstModTime.str());
+#endif
+
+        if (0 == retriesLeft)
+            throw makeErrnoExceptionV(err, "%s(%s, %s)", callerName, src, dst);
+
+        MilliSleep(100 << retry); // start at 100ms and double each retry
```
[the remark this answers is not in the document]
jakesmith: Not sure either. I hope this will be all be for short term setting to help diagnose what the actual underlying issue is, and we can either remove it completely, or modify it later when we know more.
Perhaps I should just cap the limit to a few seconds if retries is set high..

initfiles/componentfiles/configxml/RoxieTopology.xsl
jakesmith: hm, should not have missed that.

jakesmith:

ghalliday (approved):

ghalliday:
@GordonSmith do you know why various tests are failing not being able to pull data?  Will this be better with newbuild?

e.g. 
https://github.com/hpcc-systems/HPCC-Platform/actions/runs/3887283391/jobs/6633511454

system/jlib/jfile.cpp
```diff
@@ -760,6 +760,79 @@ bool CFile::remove()
...
+
+        if ((unsigned)-1 == retriesLeft)
+            retriesLeft = getRenameRetries();
+
+        int err = errno; // preserve to use after tracing if exhausted retry attempts
+
+#ifndef _WIN32 // NB: this is primary here to help track down the issue reported in HPCC-28454
+        // why did rename fail? - check that src and path of dst exists (and/or is accessible), and trace it's gid and uid, size and is_dir status.
+        struct stat srcInfo;
+        if (0 != stat(src, &srcInfo))
```
mckellyln, line 806: If the rename() fails because the file does not exist yet (I know that's just a guess at what might be happening) then perhaps we should not throw upon stat() failure ?  Perhaps just report stat() failure ??

thorlcr/graph/thgraph.cpp
```diff
@@ -2874,6 +2874,11 @@ void CJobBase::startJob()
         else
             IWARNLOG("Failed to capture process stacks: %s", output.str());
     }
+
+    constexpr unsigned defaultNumRenameRetries = 10;
+    unsigned numRenameRetries = getOptInt64("numRenameRetries", getGlobalConfigSP()->getPropInt("expert/@numRenameRetries", defaultNumRenameRetries));
+    if (numRenameRetries)
```
mckellyln, line 2880: If you wanted to disable the retries logic, could zero be an ok value ?

jakesmith: If it were 0 in the global configuration, it wouldn't get it here, but it would pick it up/set it the 1st time rename is used in jfile. But agree it's misleading, and means you can't set it to 0 as a workunit or Thor setting.

system/jlib/jfile.cpp
jakesmith: on reflection I agree - the theory that this issue is being caused because the file is still just one of the theories, but it is probably the most likely. As it stands it will prove that, but yes it would be better to avoid the throw here if it doesn't and still allow it to retry. I'll change it.

PR #16823, opened 2023-01-10
HPCC-28326 Allow connections to multiple servers

jackdelv:
Signed-off-by: jackdelv <jackdelv39@gmail.com>

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [x] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/plugins/mongodb/mongodbembed.hpp
+++ b/plugins/mongodb/mongodbembed.hpp
@@ -42,6 +42,7 @@
 #include "tokenserialization.hpp"
 #include "rtlfield.hpp"
 #include "roxiemem.hpp"
+#include <mutex>
 
 namespace mongodbembed
 {
@@ -196,11 +197,14 @@ namespace mongodbembed
          * 
          * @param database MongoDB database to connect to.
          * @param collection MongoDB collection to connect to.
+         * @param _connectionString Connection string for creating the mongocxx::uri.
+         * @param _batchSize The number of documents MongoDB should return per batch.
          */
-        MongoDBQuery(const char *database, const char *collection, std::int32_t _batchSize) 
+        MongoDBQuery(const char *database, const char *collection, const char *_connectionString, std::int32_t _batchSize) 
         {
             databaseName = database;
             collectionName = collection;
+            connectionString = _connectionString;
             batchSize = _batchSize;
         }
 
@@ -227,6 +231,7 @@ namespace mongodbembed
             embeddedScript = end + 1; // Save embedded script as what is after the parenthesis
             if(embeddedScript.find(";") == std::string::npos) failx("Syntax Error: missing semicolon");
             cursor = embeddedScript.c_str();
+            queryString.set(embeddedScript.c_str());
         }
 
         /**
@@ -299,6 +304,27 @@ namespace mongodbembed
             return batchSize;
         }
 
+        /**
+         * @brief Returns a const char pointer to the connection string for
+         * creating the uri.
+         * 
+         * @return The connection string as a const char pointer.
+         */
+        const char * queryConnectionString()
+        {
+            return connectionString.str();
+        }
+
+        /**
+         * @brief Returns the query string for building the key to the hash.
+         * 
+         * @return const char* Query string for hashing.
+         */
+        const char * queryQueryString()
+        {
+            return queryString.str();
+        }
+
     protected:
         std::string databaseName;                      //! Local copy of database name.
         std::string collectionName;                    //! Local copy of collection name.
@@ -308,6 +334,8 @@ namespace mongodbembed
         const char* cursor = nullptr;                  //! Pointer for keeping track of parsing the embedded script.
         StringArray result_rows;                       //! Local copy of result rows.
         std::int32_t batchSize;                        //! Batch Size for result rows.
+        StringBuffer connectionString;                 //! Pointer to connection string for hashing and creating the uri.
+        StringBuffer queryString;                      //! Pointer to query string for hashing.
     };
 
     /**
@@ -327,7 +355,7 @@ namespace mongodbembed
         Linked<IEngineRowAllocator> m_resultAllocator;      //!< Pointer to allocator used when building result rows.
         bool m_shouldRead;                                  //!< If true, we should continue trying to read more messages.
         __int64 m_currentRow;                               //!< Current result row.
-        std::shared_ptr<MongoDBQuery> m_query;                              //!< Pointer to MongoDBQuery object.
+        std::shared_ptr<MongoDBQuery> m_query;              //!< Pointer to MongoDBQuery object.
     };
 
     /**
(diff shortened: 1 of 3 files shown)


plugins/mongodb/mongodbembed.hpp
```diff
@@ -353,39 +383,60 @@ namespace mongodbembed
...
+         * 
+         * @param queryString A const char * holding the query string for hashing.
+         */
+        void create_connection(const char *connectionString, const char * queryString) 
+        {
+            std::shared_ptr<mongocxx::client> client_ptr = std::make_shared<mongocxx::client>(mongocxx::client{mongocxx::uri{connectionString}});
+
+            // Use a hash of the connection string as the key to finding
+            // any connection objects
+            std::string hashString = std::string(connectionString) + std::string(queryString);
```
ghalliday, line 411: minor:  You can avoid the temporary string by calling hash on each in turn.  (More significant in lookup)

hash64_t key = rtlHash64VStr(queryString, rtlHash64VStr(connectionString,0));

plugins/mongodb/mongodbembed.hpp
```diff
@@ -353,39 +383,60 @@ namespace mongodbembed
...
+            hash64_t key = rtlHash64VStr(hashString.c_str(), 0); 
+
+            clientConnections[key] = client_ptr;
```
ghalliday, line 414: Is this thread-safe?  I suspect reading and writing to clientConnections needs protecting with a critical section.

ghalliday (commented):
A couple of comments.  I'll wait for Rodrigo to review before a full look.

plugins/mongodb/mongodbembed.hpp
jackdelv: Changed

plugins/mongodb/mongodbembed.hpp
jackdelv: Ok I did some digging, and the couchbase plugin definitely uses critical sections whenever it modifies the connection pool. I will get working on that.

jackdelv:
@ghalliday I added a critical section to the create_connection function. It should be thread safe reading from the connection map because all the queries have unique connections. Back to you.

plugins/mongodb/mongodbembed.hpp
ghalliday: I don't think the critical section is quite correct.  There are two different questions

i) is the mongodb library thread safe when creating objects.
I assume that is true since most modern code takes this into account.

ii) Are the data structures in this file thread safe.
It is possible for multiple plugin functions to be called at the same time on different threads.  The clientConnections objects is potentially modified by 2 threads at the same time, or read at the same time that it is being written.

To protect against this you need a CriticalBlock around the code that reads from the map as well as the code that modifies it.  You should also minimize the length of time the critical lock is held (otherwise other threads can unnecessarily be held up).  If the mongodb library is thread safe then the criticalblock should be moved to just before the update of clientConnections.

jackdelv: In the documentation they don't mention anything specific to creating client objects but from looking at the source code for the client constructor it doesn't seem to use any shared resources and creates everything from the uri passed in. I believe you are correct in that it the creation of the client is thread safe, but if you would like to look at the mongocxx code I can show you where that is. I added a CriticalBlock to where the client gets read from the map and removed the client creation from the other CriticalBlock.

plugins/mongodb/mongodbembed.hpp
```diff
@@ -227,6 +231,7 @@ namespace mongodbembed
             embeddedScript = end + 1; // Save embedded script as what is after the parenthesis
             if(embeddedScript.find(";") == std::string::npos) failx("Syntax Error: missing semicolon");
             cursor = embeddedScript.c_str();
+            queryString.append(embeddedScript.c_str());
```
rpastrana, line 234: why don't we perform a queryString.set rather than append here?

plugins/mongodb/mongodbembed.hpp
```diff
@@ -299,6 +304,27 @@ namespace mongodbembed
             return batchSize;
         }
 
+        /**
+         * @brief Returns a const char pointer to the connection string for
+         * creating the uri.
+         * 
+         * @return The connection string as a const char pointer.
+         */
+        const char * uri()
```
rpastrana, line 313: Unless we're forced to use this name, let's change to queryConnectionString()

rpastrana (commented):
@jackdelv looks good overall, left a few questions/comments

plugins/mongodb/mongodbembed.hpp
jackdelv: That was an oversight. I was not familiar with the StringBuffer methods. I have changed it to use set().

plugins/mongodb/mongodbembed.hpp
```diff
@@ -299,6 +304,27 @@ namespace mongodbembed
...
+        {
+            return connectionString.str();
+        }
+
+        /**
+         * @brief Returns the query string for building the key to the hash.
+         * 
+         * @return const char* Query string for hashing.
+         */
+        const char * queryStr()
```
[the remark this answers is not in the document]
jackdelv: I have made the change.

plugins/mongodb/mongodbembed.hpp
```diff
@@ -353,39 +384,67 @@ namespace mongodbembed
...
+         * 
+         * @param connectionString The connection string for constructing the client object.
+         * 
+         * @param queryString A const char * holding the query string for hashing.
+         */
+        void create_connection(const char *connectionString, const char * queryString) 
+        {
+            // Use a hash of the connection string as the key to finding
+            // any connection objects
+            hash64_t key = rtlHash64VStr(queryString, rtlHash64VStr(connectionString, 0));
```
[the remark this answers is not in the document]
jackdelv: Yes, they are both guaranteed to be valid. queryString gets set in compileEmbeddedScript and checks for an empty query, and connectionString gets set in MongoDBEmbedFunctionContext::MongoDBEmbedFunctionContext() and fails if the necessary port and server options are not specified.

plugins/mongodb/mongodbembed.hpp
```diff
@@ -353,39 +384,67 @@ namespace mongodbembed
...
-         * @return connection returns mongocxx::pool::entry
+         * @param queryString A const char * holding the query string for hashing.
+         * 
+         * @return A shared pointer to the mongocxx:client object for connecting to the database.
          */
-        connection get_connection() 
+        std::shared_ptr<mongocxx::client> get_connection(const char *connectionString, const char * queryString)
         {
-            return _pool->acquire();
+            // Get key for client object
```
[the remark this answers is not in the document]
jackdelv: They are guaranteed to by non-empty and the queryString gets explicitly checked in compileEmbeddedScript(). If the connectionString is empty that is because the proper arguments were supplied by the user and will fail with details about what is missing.

plugins/mongodb/mongodbembed.hpp
```diff
@@ -353,39 +384,67 @@ namespace mongodbembed
...
+                return clientConnections[key];
+            }
         }
 
     private:
         MongoDBConnection() = default;
 
-        std::unique_ptr<mongocxx::instance> _instance = nullptr;
-        std::unique_ptr<mongocxx::pool> _pool = nullptr;
+        std::unique_ptr<mongocxx::instance> _instance = nullptr;        //!< Unique pointer to the single mongocxx::instance for the program.
```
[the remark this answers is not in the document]
jackdelv: There is an underscore because in MongoDBConnection::configure the parameter to the function is also an instance and to avoid conflicts I chose to prefix an underscore to the private member. This can be changed to something more helpful, but I wasn't sure what to change it to.

jackdelv:
Ok, I have addressed all of your comments @rpastrana. Back to you.

rpastrana:
@jackdelv it looks like something went wrong in the last commit, you might need a rebase

jackdelv:
@rpastrana Yes, I rebased my branch to candidate-9.0.x, but forgot to change the base of the pull request to match. It should be alright now.

plugins/mongodb/mongodbembed.hpp
rpastrana: Not a big deal, in fact I think it's fairly common in our code base to use the append method as long as we can ensure the string is empty, just wanted to make sure we were doing what we intended to do.

plugins/mongodb/mongodbembed.hpp
```diff
@@ -353,39 +384,67 @@ namespace mongodbembed
...
-        using connection = mongocxx::pool::entry;
+        /**
+         * @brief Creates a client object using the specified connections string. The client is
+         * added to the map of active connections.
+         * 
+         * @param connectionString The connection string for constructing the client object.
+         * 
+         * @param queryString A const char * holding the query string for hashing.
+         */
+        void create_connection(const char *connectionString, const char * queryString)
```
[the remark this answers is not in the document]
rpastrana: I meant the actual conn string format, a lot of times the caller will pass in incomplete/unexpected format and not really know why their connections fail.

plugins/mongodb/mongodbembed.hpp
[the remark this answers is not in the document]
rpastrana: typically I'd prefix the member w/ a 'm_' but apparently the platform is going away from that nomenclature. I'd add the _ to the temporary function param

rpastrana (approved):

plugins/mongodb/mongodbembed.hpp
```diff
@@ -299,6 +304,27 @@ namespace mongodbembed
             return batchSize;
         }
 
+        /**
+         * @brief Returns a const char pointer to the connection string for
+         * creating the uri.
+         * 
+         * @return The connection string as a const char pointer.
+         */
+        const char * queryConnectionString()
```
ghalliday, line 313: style:  no need to change, but this could be a const function.

plugins/mongodb/mongodbembed.hpp
```diff
@@ -338,54 +366,90 @@ namespace mongodbembed
...
+        std::shared_ptr<mongocxx::client> get_connection(const char *connectionString, const char * queryString)
         {
-            return _pool->acquire();
+            // Get key for client object 
+            hash64_t key = rtlHash64VStr(queryString, rtlHash64VStr(connectionString, 0));
+
+            {
+                CriticalBlock block(connLock);
+
+                return clientConnections[key];
```
ghalliday, line 443: note: This may not be obvious, but this will insert a null entry into the map.  In your case that is unlikely to matter, and the code is simpler as it stands.  In other cases you may not want that overhead.

ghalliday (approved):
@jackdelv please can you squash the commits.  A couple of comments, mainly for education.

PR #16845, opened 2023-01-16
HPCC-28756 New spill stats for inter-subgraph spills

shamser:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/common/workunit/workunit.cpp
+++ b/common/workunit/workunit.cpp
@@ -2695,6 +2695,51 @@ cost_type aggregateDiskAccessCost(const IConstWorkUnit * wu, const char *scope)
     }
     return totalCost;
 }
+
+// totalSizeSpill and peakSizeSpill should be initialized to zero
+void gatherSpillSize(const IConstWorkUnit * wu, const char *scope, stat_type & totalSizeSpill, stat_type & peakSizeSpill)
+{
+    WuScopeFilter filter;
+    if (!isEmptyString(scope))
+        filter.addScope(scope);
+    else
+        filter.addScope("");
+    filter.setIncludeNesting(2);
+    filter.addSource("global");
+    filter.addOutputStatistic(StSizeSpillFile);
+    filter.addOutputStatistic(StSizePeakSpillFile);
+    filter.finishedFilter();
+    Owned<IConstWUScopeIterator> it = &wu->getScopeIterator(filter);
+    for (it->first(); it->isValid(); )
+    {
+        stat_type value = 0;
+        if (it->getStat(StSizeSpillFile, value))
+        {
+            totalSizeSpill += value;
+            if (it->getStat(StSizePeakSpillFile, value))
+            {
+                if (value>peakSizeSpill)
+                    peakSizeSpill = value;
+            }
+            it->nextSibling();
+        }
+        else
+        {
+            it->next();
+        }
+    }
+}
+
+void updateSpillSize(IWorkUnit * wu, const char * scope, StatisticScopeType scopeType)
+{
+    stat_type totalSizeSpill = 0, peakSizeSpill = 0;
+    gatherSpillSize(wu, scope, totalSizeSpill, peakSizeSpill);
+    if (totalSizeSpill)
+    {
+        wu->setStatistic(queryStatisticsComponentType(), queryStatisticsComponentName(), scopeType, scope, StSizePeakSpillFile, nullptr, peakSizeSpill, 1, 0, StatsMergeReplace);
+        wu->setStatistic(queryStatisticsComponentType(), queryStatisticsComponentName(), scopeType, scope, StSizeSpillFile, nullptr, totalSizeSpill, 1, 0, StatsMergeReplace);
+    }
+}
 //---------------------------------------------------------------------------------------------------------------------
 
 
--- a/ecl/eclagent/eclagent.cpp
+++ b/ecl/eclagent/eclagent.cpp
@@ -1980,6 +1980,8 @@ void EclAgent::doProcess()
         const cost_type diskAccessCost = aggregateDiskAccessCost(w, nullptr);
         if (diskAccessCost)
             w->setStatistic(queryStatisticsComponentType(), queryStatisticsComponentName(), SSTglobal, "", StCostFileAccess, NULL, diskAccessCost, 1, 0, StatsMergeReplace);
+        updateSpillSize(w, nullptr, SSTglobal);
+
         addTimings(w);
 
         switch (w->getState())
@@ -2528,6 +2530,7 @@ void EclAgentWorkflowMachine::noteTiming(unsigned wfid, timestamp_type startTime
     const cost_type diskAccessCost = aggregateDiskAccessCost(wu, scope);
     if (diskAccessCost)
         wu->setStatistic(queryStatisticsComponentType(), queryStatisticsComponentName(), SSTworkflow, scope, StCostFileAccess, NULL, diskAccessCost, 1, 0, StatsMergeReplace);
+    updateSpillSize(wu, scope, SSTworkflow);
 }
 
 void EclAgentWorkflowMachine::doExecutePersistItem(IRuntimeWorkflowItem & item)
--- a/thorlcr/activities/thdiskbaseslave.cpp
+++ b/thorlcr/activities/thdiskbaseslave.cpp
@@ -594,7 +594,7 @@ void CDiskWriteSlaveActivityBase::process()
     StringBuffer tmpStr;
     fName.set(getPartFilename(*partDesc, 0, tmpStr).str());
     if (diskHelperBase->getFlags() & TDXtemporary && !container.queryJob().queryUseCheckpoints())
-        container.queryTempHandler()->registerFile(fName, container.queryOwner().queryGraphId(), usageCount, true);
+        tmpUsage = container.queryTempHandler()->registerFile(fName, container.queryOwner().queryGraphId(), usageCount, true);
     try
     {
         ActPrintLog("handling fname : %s", fName.get());
@@ -671,6 +671,8 @@ void CDiskWriteSlaveActivityBase::processDone(MemoryBuffer &mb)
     modifiedTime.getTime(hour, min, sec, nanosec);
     modifiedTime.setTime(hour, min, sec, 0);
     modifiedTime.serialize(mb);
+    if (tmpUsage)
+        tmpUsage->setSize(sz);
 }
 
 /////////////
--- a/thorlcr/graph/thgraph.cpp
+++ b/thorlcr/graph/thgraph.cpp
@@ -2285,14 +2285,16 @@ IThorGraphResults *CGraphBase::createThorGraphResults(unsigned num)
 
 ////
 
-void CGraphTempHandler::registerFile(const char *name, graph_id graphId, unsigned usageCount, bool temp, WUFileKind fileKind, StringArray *clusters)
+CFileUsageEntry * CGraphTempHandler::registerFile(const char *name, graph_id graphId, unsigned usageCount, bool temp, WUFileKind fileKind, StringArray *clusters)
 {
     assertex(temp);
     LOG(MCdebugProgress, thorJob, "registerTmpFile name=%s, usageCount=%d", name, usageCount);
     CriticalBlock b(crit);
     if (tmpFiles.find(name))
         throw MakeThorException(TE_FileAlreadyUsedAsTempFile, "File already used as temp file (%s)", name);
-    tmpFiles.replace(* new CFileUsageEntry(name, graphId, fileKind, usageCount));
+    CFileUsageEntry *fileEntry = new CFileUsageEntry(name, graphId, fileKind, usageCount);
+    tmpFiles.replace(*fileEntry);
+    return fileEntry;
 }
 
 void CGraphTempHandler::deregisterFile(const char *name, bool kept)
@@ -2341,6 +2343,22 @@ void CGraphTempHandler::clearTemps()
     tmpFiles.kill();
 }
 
+void CGraphTempHandler::serializeUsageStats(MemoryBuffer &mb, graph_id gid)
+{
+    CriticalBlock b(crit);
+    Owned<IFileUsageIterator> iter = getIterator();
+    offset_t activeSpillSize = 0;
+    offset_t graphSpillSize = 0;
+    ForEach(*iter)
+    {
+        CFileUsageEntry &entry = iter->query();
+        if (entry.queryGraphId() == gid)
+            graphSpillSize += entry.getSize();
+        activeSpillSize += entry.getSize();
+    }
+    mb.append(graphSpillSize);
+    mb.append(activeSpillSize);
+}
 /////
 
 class CGraphExecutor;
--- a/thorlcr/graph/thgraph.hpp
+++ b/thorlcr/graph/thgraph.hpp
@@ -199,6 +199,7 @@ class CFileUsageEntry : public CInterface
     unsigned usage;
     graph_id graphId;
     WUFileKind fileKind;
+    offset_t size = 0;
 public:
     CFileUsageEntry(const char *_name, graph_id _graphId, WUFileKind _fileKind, unsigned _usage) :name(_name), graphId(_graphId), fileKind(_fileKind), usage(_usage) { }
     unsigned queryUsage() const { return usage; }
@@ -208,16 +209,24 @@ public:
     void decUsage() { --usage; }
 
     const char *queryFindString() const { return name; }
+    inline void setSize(offset_t _size) { size = _size; }
+    inline offset_t getSize() const { return size; }
 };
 
 typedef IIteratorOf<CFileUsageEntry> IFileUsageIterator;
 
 interface IGraphTempHandler : extends IInterface
 {
-    virtual void registerFile(const char *name, graph_id graphId, unsigned usageCount, bool temp, WUFileKind fileKind=WUFileStandard, StringArray *clusters=NULL) = 0;
+    virtual CFileUsageEntry * registerFile(const char *name, graph_id graphId, unsigned usageCount, bool temp, WUFileKind fileKind=WUFileStandard, StringArray *clusters=NULL) = 0;
     virtual void deregisterFile(const char *name, bool kept=false) = 0;
     virtual void clearTemps() = 0;
     virtual IFileUsageIterator *getIterator() = 0;
+    virtual void serializeUsageStats(MemoryBuffer &mb, graph_id gid) = 0;
+    static void serializeNullUsageStats(MemoryBuffer &mb)
+    {
+        mb.append((offset_t)0);
+        mb.append((offset_t)0);
+    }
 };
 
 class CGraphDependency : public CInterface
@@ -564,10 +573,10 @@ public:
     }
     virtual bool removeTemp(const char *name) = 0;
 // IGraphTempHandler
-    virtual void registerFile(const char *name, graph_id graphId, unsigned usageCount, bool temp, WUFileKind fileKind, StringArray *clusters);
-    virtual void deregisterFile(const char *name, bool kept=false);
-    virtual void clearTemps();
-    virtual IFileUsageIterator *getIterator()
+    virtual CFileUsageEntry * registerFile(const char *name, graph_id graphId, unsigned usageCount, bool temp, WUFileKind fileKind, StringArray *clusters) override;
+    virtual void deregisterFile(const char *name, bool kept=false) override;
+    virtual void clearTemps() override;
+    virtual IFileUsageIterator *getIterator() override
     {
         class CIterator : implements IFileUsageIterator, public CInterface
         {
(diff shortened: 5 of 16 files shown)


thorlcr/activities/thdiskbaseslave.cpp
```diff
@@ -671,6 +671,8 @@ void CDiskWriteSlaveActivityBase::processDone(MemoryBuffer &mb)
     modifiedTime.getTime(hour, min, sec, nanosec);
     modifiedTime.setTime(hour, min, sec, 0);
     modifiedTime.serialize(mb);
+    if (diskHelperBase->getFlags() & TDXtemporary && !container.queryJob().queryUseCheckpoints())
```
ghalliday, line 674: precedence is much clearer with brackets around the (a & b) && !

thorlcr/graph/thgraphslave.cpp
```diff
@@ -1258,6 +1258,23 @@ bool CSlaveGraph::serializeStats(MemoryBuffer &mb)
...
+    IGraphTempHandler *tempHandler = queryJob().queryTempHandler(false);
+    if (tempHandler)
+    {
+        offset_t activeSpillSize = 0;
+        graph_id gid = queryGraphId();
+        Owned<IFileUsageIterator> iter = tempHandler->getIterator();
+        ForEach(*iter)
+        {
+            CFileUsageEntry &entry = iter->query();
+            if (entry.queryGraphId()==gid)
```
ghalliday, line 1272: Is this a graphid or a subgraph id.  I'm not sure why this is filtering....

thorlcr/graph/thgraphslave.cpp
```diff
@@ -1258,6 +1258,23 @@ bool CSlaveGraph::serializeStats(MemoryBuffer &mb)
...
+                activeSpillSize += entry.getSize();
+        }
+        if (activeSpillSize)
+            stats.setStatistic(StSizeSpillFile, activeSpillSize);
```
ghalliday, line 1276: Not quite sure if this is the spills for this subgraph, or the total active.  I think you are calculating the 1st, but I reporting it as the second.   I may be wrong though!

ghalliday (commented):
@shamser a couple of questions.

thorlcr/graph/thgraphslave.cpp
shamser: I seems that ::queryGraphId() should be return graphId but it returns subgraphId.

The filtering is to avoid double counting open spill files.  Each subgraph reports the spill size for files it has created.

jakesmith: CGraphBase objects (and it's derivatives) always correspond to subgraphs.
A [outer] graph is submitted to thor as a job, those graphs correspond to Thor CJobBase objects.
The [outer]graph name is retried from queryJob()->queryGraphName()

jakesmith: Afaics, this isn't going to work as planned.
When this method (CSlaveGraph::serliazeStats) is called, it's going to be either during, or at the end of it's life.
i.e. registered spills that were created by this subgraph will not be considered after this subgraph is finished.

For example, take 3 subgraphs, in series, where sg1 spills 's1', sg2 consumes 's1' and creates a new spill ('s2'), which sg3 consumes.
The sg1 will serialize it's spill aggregate, the master will call the job::addGraphSpill and add to total and peak.
The 2nd cosumes 's1' (deletes it), before producing 's2'.
sg2 serializes it's spill aggregate, the master addGraphSpill adds to total (ok as total of everything spilt ever), but then considers that total as the new peak. But it's not as this subgraph's spill is now the only spill in play.

i.e. there's no calculation of the current usage which goes down as well as up afaics.

The spill total related to the current subgraph could be useful as well (for diving down into what contributed to the peak), but the overall spill usage usage/peak needs to be the aggregate of all current spills registered in the handler, which will fluctuate up and down.

jakesmith: You could continue to aggregate per subgraph and publish as a separate stat., though if not needed now, I'd leave for another PR, since you can already dive into a subgraph and see this detail manually by looking at the spill activities (there's usually only 1 or 2).

For the total currently in use (registered) *for the current graph (thor job)*, I think you need to check that you're in the root subgraph (owner == null), and total all registred in the job temp handler and serialize that.
That is the normal case (the job temp handler holds all the spills), however there is another case, where there's a contained group of child graphs in a resultsGraph, it will have it's own temp handler (this is so it clears up when goes out of scope).
These will need totalling and serializing too.

So I think you'll need something like (in CSlaveGraph::serializeStats):
```
   if (!owner)
      queryJob().queryTempHandler()->serializeUsageStats(stats);
   else
   {
        IGraphTempHandler *tempHandler = queryTempHandler(false);
        if (tempHandler) // graphResults temp handler
            tempHandler->serizlieUsageStats(stats);
    }
```

(NB: in theory I think there could be grandchild resultgraph temp handlers).

These current totals will need totalling up on the master and seeing if > peak.
This will be the peak for graph, not the job overall.

e.g. graph1 may be the 3 subgraph sequence example in my previous comment and had a peak due to 's2', but the next thor job (graph2), may exceed this.
So need to track workunit peak, read it, and increase it at the end of the job if exceeded too.

jakesmith (changes requested):

ecl/eclagent/eclagent.cpp
```diff
@@ -2528,6 +2536,14 @@ void EclAgentWorkflowMachine::noteTiming(unsigned wfid, timestamp_type startTime
     const cost_type diskAccessCost = aggregateDiskAccessCost(wu, scope);
     if (diskAccessCost)
         wu->setStatistic(queryStatisticsComponentType(), queryStatisticsComponentName(), SSTworkflow, scope, StCostFileAccess, NULL, diskAccessCost, 1, 0, StatsMergeReplace);
+    stat_type totalSizeSpill = 0, peakSizeSpill = 0;
+    gatherSpillSize(wu, scope, totalSizeSpill, peakSizeSpill);
```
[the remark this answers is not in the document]
jakesmith: I think you may need to keep track of the current job wide peak too (in eclagent) and feed it into gatherSpillSize instead of 0, 0

jakesmith (commented):
@shamser - please see comments/questions.

thorlcr/activities/thdiskbaseslave.cpp
```diff
@@ -671,6 +671,8 @@ void CDiskWriteSlaveActivityBase::processDone(MemoryBuffer &mb)
...
+        container.queryTempHandler()->setSize(fName, sz);
```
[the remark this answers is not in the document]
shamser: I was looking at modifying registerFile to return CFileUsageEntry.   CMasterGraphTempHandler::registerFile  doesn't always create a CFileUsageEntry.  Should CMasterGraphTempHandler::registerFile return a nullptr if calls wu->addFile() rather than CGraphTempHandler::registerFile?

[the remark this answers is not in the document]
jakesmith: yes, it will only create/add a CFileUsageEntry if it's not a published file.
I'd change it to return null where needed.

ecl/eclagent/eclagent.cpp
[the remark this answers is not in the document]
shamser: gatherSpillSize on line 2540 should gather the total and peaks for this workflow.  These are then written to the workflow scope, so these should be the totals and peaks for the work flow.

There is another gatherSpillSize (line 1984) where it sums up the totals and the peaks across all the workflows.  These are written to the global scope.  The job-wide spill total and peak should written to the global scope (by summing all totals from the workflows and getting the peaks from the workflows).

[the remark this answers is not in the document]
jakesmith: ok, yes, that makes sense.
I wonder if should update the global total as a running progress though, as jobs can be long running, and so the global total won't be seen until the job completes (or fails).
That could be added as a separate PR/improvement though.

jakesmith (commented):
@shamser - looks good, but I think registerFile linking/ownership of object/return should be clarified.

jakesmith (approved):
@shamser - looks good, please squash.

jakesmith (approved):
@ghalliday - please check and merge.

ecl/eclagent/eclagent.cpp
```diff
@@ -2528,6 +2536,13 @@ void EclAgentWorkflowMachine::noteTiming(unsigned wfid, timestamp_type startTime
     const cost_type diskAccessCost = aggregateDiskAccessCost(wu, scope);
     if (diskAccessCost)
         wu->setStatistic(queryStatisticsComponentType(), queryStatisticsComponentName(), SSTworkflow, scope, StCostFileAccess, NULL, diskAccessCost, 1, 0, StatsMergeReplace);
+    stat_type totalSizeSpill = 0, peakSizeSpill = 0;
```
ghalliday, line 2539: Are these timings recorded if the workunit fails (e.g. because out of spill file)?  I suspect not.

shamser: I don't think they are.  Can this be done in a separate Jira?

thorlcr/graph/thgraphslave.cpp
```diff
@@ -1330,6 +1331,15 @@ void CSlaveGraph::serializeDone(MemoryBuffer &mb)
...
     }
     mb.writeDirect(cPos, sizeof(count), &count);
+
+    if (!owner)
+        queryJob().queryTempHandler()->serializeUsageStats(mb, gid);
+    else
+    {
+        IGraphTempHandler *tempHandler = queryTempHandler(false);
+        if (tempHandler)
+            tempHandler->serializeUsageStats(mb, gid);
```
ghalliday, line 1341: What happens if there is no tempHandler?  No stats will be serialized, and the master will try to read data that isn't there?

ghalliday (changes requested):
@shamser I think there are a couple of potential problems.

thorlcr/graph/thgraphslave.cpp
jakesmith: good point, missed that. Need to serialize dummy values.

thorlcr/activities/thdiskbaseslave.cpp
```diff
@@ -594,7 +594,7 @@ void CDiskWriteSlaveActivityBase::process()
     StringBuffer tmpStr;
     fName.set(getPartFilename(*partDesc, 0, tmpStr).str());
     if (diskHelperBase->getFlags() & TDXtemporary && !container.queryJob().queryUseCheckpoints())
-        container.queryTempHandler()->registerFile(fName, container.queryOwner().queryGraphId(), usageCount, true);
+        tmpUsage.setown(container.queryTempHandler()->registerFile(fName, container.queryOwner().queryGraphId(), usageCount, true));
```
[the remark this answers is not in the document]
jakesmith: agree, and missed this, and also initially did think it should just return an unlinked pointer, but erred on the side of caution of linking it here, forgeting about the other calls to registerFile which are not owning and would therefore leak.

I should be guaranteed to still be in scope when it gets to use this pointer in the same activity, the temp handler is owner at a higher scope.

So I would change it to return an unlinked pointer and store it in a plain member pointer instead of an Owned.

jakesmith:
>@jakesmith I don't think I agree with this (although I didn't see the original code) - see comments above.

Agree with your other comments, see my reply.

ecl/eclagent/eclagent.cpp
```diff
@@ -1980,6 +1980,14 @@ void EclAgent::doProcess()
         const cost_type diskAccessCost = aggregateDiskAccessCost(w, nullptr);
         if (diskAccessCost)
             w->setStatistic(queryStatisticsComponentType(), queryStatisticsComponentName(), SSTglobal, "", StCostFileAccess, NULL, diskAccessCost, 1, 0, StatsMergeReplace);
+        stat_type totalSizeSpill = 0, peakSizeSpill = 0;
```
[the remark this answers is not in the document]
jakesmith: >This means the global values will not be updated until the workunit completes. It is likely to be useful if this was updated while it runs

Made same point in previous comment here : https://github.com/hpcc-systems/HPCC-Platform/pull/16845#discussion_r1088167357

But thought it could be tackled as a separate/next JIRA/PR.

thorlcr/graph/thgraph.cpp
```diff
@@ -2285,14 +2285,16 @@ IThorGraphResults *CGraphBase::createThorGraphResults(unsigned num)
...
+CFileUsageEntry * CGraphTempHandler::registerFile(const char *name, graph_id graphId, unsigned usageCount, bool temp, WUFileKind fileKind, StringArray *clusters)
 {
     assertex(temp);
     LOG(MCdebugProgress, thorJob, "registerTmpFile name=%s, usageCount=%d", name, usageCount);
     CriticalBlock b(crit);
     if (tmpFiles.find(name))
         throw MakeThorException(TE_FileAlreadyUsedAsTempFile, "File already used as temp file (%s)", name);
-    tmpFiles.replace(* new CFileUsageEntry(name, graphId, fileKind, usageCount));
+    Owned<CFileUsageEntry> fileEntry(new CFileUsageEntry(name, graphId, fileKind, usageCount));
+    tmpFiles.replace(*fileEntry);
```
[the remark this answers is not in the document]
jakesmith: Agree with your other comments, see my reply.

ecl/eclagent/eclagent.cpp
jakesmith: it looks like the way exceptions can be thrown in perform() means noteTiming will be skipped - therefore the workflow item's noteTiming values will be lost.
But the total aggregate (- the missing failing workflow item ) will be recorded afaics.

NB: it's not just these new spill sizes which will be lost, it's also Started,  Elapsed, Cost .. that won't be recorded for the failing workflow item afaics.

ecl/eclagent/eclagent.cpp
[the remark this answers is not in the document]
shamser: I'll work on this in a separate Jira.

jakesmith (approved):
@shamser - looks good, please squash

jakesmith:
@shamser - could you also retarget this at master

ecl/eclagent/eclagent.cpp
jakesmith: @shamser - I think a JIRA was opened for this(?) Could you link it into HPCC-28756?

jakesmith (approved):
@shamser - looks good.

shamser:
@ghalliday Please can you check.

ecl/eclagent/eclagent.cpp
```diff
@@ -2528,6 +2536,13 @@ void EclAgentWorkflowMachine::noteTiming(unsigned wfid, timestamp_type startTime
     const cost_type diskAccessCost = aggregateDiskAccessCost(wu, scope);
     if (diskAccessCost)
         wu->setStatistic(queryStatisticsComponentType(), queryStatisticsComponentName(), SSTworkflow, scope, StCostFileAccess, NULL, diskAccessCost, 1, 0, StatsMergeReplace);
+    stat_type totalSizeSpill = 0, peakSizeSpill = 0;
+    gatherSpillSize(wu, scope, totalSizeSpill, peakSizeSpill);
+    if (totalSizeSpill)
+    {
+        wu->setStatistic(queryStatisticsComponentType(), queryStatisticsComponentName(), SSTworkflow, scope, StSizePeakSpillFile, nullptr, peakSizeSpill, 1, 0, StatsMergeReplace);
+        wu->setStatistic(queryStatisticsComponentType(), queryStatisticsComponentName(), SSTworkflow, scope, StSizeSpillFile, nullptr, totalSizeSpill, 1, 0, StatsMergeReplace);
+    }
```
ghalliday, line 2545: This code is almost identical to the global code above, and could be combined into a function.  This could then be also called from roxie - which needs similar code if it is acting as the eclagent.

common/workunit/workunit.cpp
```diff
@@ -2695,6 +2695,39 @@ cost_type aggregateDiskAccessCost(const IConstWorkUnit * wu, const char *scope)
     }
     return totalCost;
 }
+
+void gatherSpillSize(const IConstWorkUnit * wu, const char *scope, stat_type & totalSizeSpill, stat_type & peakSizeSpill)
```
ghalliday, line 2699: Should probably have a comment that total and peak should be initialised before calling, since that is slightly unusual.

thorlcr/graph/thgraph.hpp
```diff
@@ -208,16 +209,24 @@ class CFileUsageEntry : public CInterface
...
 {
-    virtual void registerFile(const char *name, graph_id graphId, unsigned usageCount, bool temp, WUFileKind fileKind=WUFileStandard, StringArray *clusters=NULL) = 0;
+    virtual CFileUsageEntry * registerFile(const char *name, graph_id graphId, unsigned usageCount, bool temp, WUFileKind fileKind=WUFileStandard, StringArray *clusters=NULL) = 0;
     virtual void deregisterFile(const char *name, bool kept=false) = 0;
     virtual void clearTemps() = 0;
     virtual IFileUsageIterator *getIterator() = 0;
+    virtual void serializeUsageStats(MemoryBuffer &mb, graph_id gid) = 0;
+    static void serializeNullUsageStats(MemoryBuffer &mb)
+    {
+        mb.append(0);
```
ghalliday, line 227: I don't think this is correct.  It will append an int (4 bytes), instead of an offset_t (8 bytes)

thorlcr/graph/thgraphmaster.cpp
```diff
@@ -2720,6 +2721,11 @@ void CMasterGraph::handleSlaveDone(unsigned node, MemoryBuffer &mb)
         sdMb.setBuffer(len, (void *)d);
         act->slaveDone(node, sdMb);
     }
+    offset_t activeSpillSize, nodeGraphSpill;
+    mb.read(nodeGraphSpill);
+    mb.read(activeSpillSize);
+    totalActiveSpillSize += activeSpillSize;
```
ghalliday, line 2727: Where do these get reset - e.g., if this is a spill within a LOOP will it incorrectly add up the spill file sizes for the all the  iterations.  @jakesmith not sure where that reset code would go, or 100% sure how spills in loops are handled.

thorlcr/graph/thgraphmaster.ipp
```diff
@@ -197,7 +201,17 @@ public:
         CriticalBlock b(wuDirty);
         dirty = true;
     }
-    
+// Track spills
+    CriticalSection critSpillTrack;
```
ghalliday, line 205: Are these updated from multiple threads?  If so using atomic (RelaxedAtomic) is probably cleaner, although not necessarily more efficient, than a critical section - especially since they are accessed without a critical section.  (RelaxedAtomic::store_max).

thorlcr/graph/thgraphmaster.cpp
```diff
@@ -2799,6 +2805,7 @@ void CMasterGraph::getFinalProgress()
             }
         }
     }
+    jobM->updateActiveSpillSize(graphSpillSize, totalActiveSpillSize);
```
ghalliday, line 2808: @jakesmith Will this be called more than once for a loop subgraph?  If so, will that cause potential problems.

ghalliday (changes requested):
@shamser looks close.  One problem, and a few comments to clean up (none of which are needed).
Please can you also make sure that there are issues/subtasks linked to HPCC-28685 for i) adding the same code to roxie ii) updating spill files as it goes along/when a job fails.

thorlcr/graph/thgraph.hpp
jakesmith: you're right, should be changed to something like:
```
mb.append((offset_t)0);
mb.append((offset_t)0);
```

thorlcr/graph/thgraphmaster.cpp
jakesmith: I think 'totalActiveSpillSize' needs resetting in CMasterGraph::preStart.
It will be recalculated each LOOP iteration, i.e. the current active total.

Would be a good to test that it does what's expected though in a LOOP that has spills within it.

thorlcr/graph/thgraphmaster.cpp
jakesmith: Yes, and related to https://github.com/hpcc-systems/HPCC-Platform/pull/16845#discussion_r1108208706,

Both 'graphSpillSize' and 'totalActiveSpillSize' should be reset at start of graph, to cater for the case where the graph is re-executed in a LOOP/GRAPH pattern.

jakesmith (commented):
@shamser - see additional comments relating to the issues Gavin spotted.

jakesmith (changes requested):
-

jakesmith (approved):
@shamser looks good

ghalliday (approved):
@shamser looks good. Please squash and I'll merge

shamser:
@ghalliday

PR #16846, opened 2023-01-16
HPCC-28785 Suppress autoCopyBackup in containerized mode

jakesmith:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/thorlcr/activities/thactivityutil.cpp
+++ b/thorlcr/activities/thactivityutil.cpp
@@ -835,7 +835,7 @@ StringBuffer &locateFilePartPath(CActivityBase *activity, const char *logicalFil
 {
     unsigned location;
     OwnedIFile ifile;
-    if (globals->getPropBool("@autoCopyBackup", true)?ensurePrimary(activity, partDesc, ifile, location, filePath):getBestFilePart(activity, partDesc, ifile, location, filePath, activity))
+    if (globals->getPropBool("@autoCopyBackup", !isContainerized())?ensurePrimary(activity, partDesc, ifile, location, filePath):getBestFilePart(activity, partDesc, ifile, location, filePath, activity))
         ActPrintLog(activity, "reading physical file '%s' (logical file = %s)", filePath.str(), logicalFilename);
     else
     {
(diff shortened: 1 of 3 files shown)


thorlcr/master/thmastermain.cpp
```diff
@@ -997,6 +997,13 @@ int main( int argc, const char *argv[]  )
             if (workerJobInstalled)
                 doWorkerRegistration = true;
         }
+
+        // autoCopyBackup will copy *any* other copy to the primary (1st) location,
+        // In containerized mode we are unlikely to ever want this behaviour,
+        // and specifically in the case where preferred planes have prioritized a copy
+        // which isn't present, we do not want to auto-copy a non-prioritized copy.
+        if (!globals->hasProp("@autoCopyBackup"))
+            globals->setPropBool("@autoCopyBackup", false);
```
ghalliday, line 1006: Change is ok, but is it better than passing !isContainerized() as the default to the getPropBool() calls?
Also, will this be wiped out if the configuration is updated by a change to the helm file?

ghalliday (commented):
@jakesmith a couple of questions.

thorlcr/master/thmastermain.cpp
jakesmith: I agree that would be cleaner. I'll change it.

>Also, will this be wiped out if the configuration is updated by a change to the helm file?

No, but it made me think a bit..

Thor does not monitor its configuration for changes (it calls loadConfiguration with monitor=false).
That's because it is assumed that it is launched per job and settings shouldn't change within a job's lifetime.

However, for a multiLinger setup, Thor instances run multiple jobs of course.
Assuming we want to [continue to] ensure config properties are _not_ auto-updated during the lifetime of a single given job, which I think we do, but allow updates for new jobs being submitted, we would need to change the mechanism so that it could auto-refresh if idle and it knew it wasn't mid-job, or make the multi linger instances become associated with a workunit and wind-up when the job was finished (this is the way it operates when multiJobLinger is off).

A separate issue, I will open a JIRA and describe more fully.

jakesmith:
@ghalliday - made change

ghalliday (approved):

PR #16891, opened 2023-01-24
HPCC-28554 Check dropzone scope access in ESP FileIO

wangkx:
The existing CWsFileIOEx::CheckServerAccess() has 2 functions: 1. validate the path request; 2. find out more information about the targeted dropzone: host and root path. In this PR, the CWsFileIOEx::CheckServerAccess() is replaced by 2 new methods: 1. ensureDropZoneFileAccess(), including: validate the path request and check DropZone scope permission; 2. getDropZoneInfo(): simplify the code for the information about the targeted dropzone.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/services/ws_fileio/ws_fileioservice.cpp
+++ b/esp/services/ws_fileio/ws_fileioservice.cpp
@@ -20,6 +20,7 @@
 
 #include "jiface.hpp"
 #include "environment.hpp"
+#include "dautils.hpp"
 #include "ws_fileioservice.hpp"
 #ifdef _WIN32
 #include "windows.h"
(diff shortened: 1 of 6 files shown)


jakesmith:
@wangkx - can you update the JIRA description to describe what the problem this aims to solve and how it will be acheived ?

jakesmith (commented):
>can you update the JIRA description to describe what the problem this aims to solve and how it will be acheived ?

@wangkx - passing back to you for now, re. adding detail to the JIRA

wangkx:
@jakesmith I may not be able to change the JIRA description (after a JIRA is created). But, I just added a comment there. Please let me know I need to change more.

jakesmith (changes requested):
@wangkx - I think there's a problem (I think pre-existing) in the way that setPlaneExternal is being used (it should always be passed a relative path), see comments.

Also, I'm not convinced the 2 new methods (getDropZoneInfo and ensureDropZoneFileAccess) are helpful, rather than combining into 1.
They are both always called and they both require quite a bit of the same info.
It looks like it would be cleaner to have a single method, e.g. 'validateDropZoneAccess' which threw exceptions consistenly (it seems some paths throw exceptions and others return false and a response?).
I think it would also make more sense to return (fill) a CDfsLogicalFileName with the validated dropzone+relative path.
That way the caller can use it directly by e.g. calling CDfsLogicalFileName::getExternalFilename.

So the calling could could look something like this in the 3 cases:
```
    CDfsLogicalFileName dlfn;
    validateDropZoneAccess(context, server, req.getDestNetAddress(), destRelativePath, SecAccess_Write, dlfn); // dlfn is populated as return value

    RemoteFilename rfn;
    dlfn.getExternalFilename(rfn);
    Owned<IFile> file = createIFile(rfn);
    ....
```

It would also help I suspect if there was a public getScopePermissions that took a CDfsLogicalFileName (that could be used in validateDropZoneAccess). There's already a static version that takes a string scope, so this would be a simpler wrapper, and could be reused in other existing places that gets the scope 1st, e.g. CDistributedFileDirectory::getFilePermissions

jakesmith:
@wangkx - any particular reason this is targetting 8.10.x vs 8.12.x?

wangkx:
After review, I can target to 8.12.x. I use 8.10.x because: 1. the merged similar changes in ESP FileSpary are in 8.10.x; 2. avoid some build error (for testing) in different environments (Docker Desktop, ubunru 20.04, etc). Looks like the 8.12.x is not stable to be built.

jakesmith:
>In other ESP services (WsWorkunits, WsDFU, etc), the code almost always throws exceptions when an error occurs. But, in this ESP service, the code tries to report the results within the response (success or some (input or execution) error). In the code I am adding in this PR, I only throw exceptions if it is a security related error. Should I change the code to always throws exceptions, including the code within the 3 methods, but not related to this PR? Not set the result in response or only set the result when success?


I'm not sure, what is the impact/consequences, of always throwing an exception on error?
Will the exception be seen by the client(s) in much the same way as the response would be ?

wangkx:
This PR is similar to https://github.com/hpcc-systems/HPCC-Platform/pull/16710 which has been rolled back from 8.10.x. I am waiting for the decision about which build should be targeted to. After the target is decided, I need to rebase my changes and test.

wangkx:
@jakesmith I made the changes. Please check if I missed any.

esp/services/ws_fileio/ws_fileioservice.cpp
```diff
@@ -34,128 +35,56 @@ void CWsFileIOEx::init(IPropertyTree *cfg, const char *process, const char *serv
...
+    if (!dropZone) //The targetDZNameOrHost could be a dropzone host.
     {
-        if (stricmp(targetDZNameOrAddress, "localhost")==0)
-            targetDZNameOrAddress = ".";
-
-        Owned<IConstDropZoneInfoIterator> dropZoneItr = env->getDropZoneIteratorByAddress(targetDZNameOrAddress);
-        ForEach(*dropZoneItr)
+        Owned<IPropertyTreeIterator> planes = getDropZonePlanesIterator(nullptr);
+        ForEach(*planes)
         {
```
jakesmith, line 49: why isn't this using findDropZonePlane ?

jakesmith (commented):
@wangkx - 1 question.

esp/services/ws_fileio/ws_fileioservice.cpp
wangkx: The path used for findDropZonePlane() is an absolute path. The path here is a relative path (from the DestRelativePath).

wangkx:
@jakesmith please see my answer.

esp/services/ws_fileio/ws_fileioservice.cpp
jakesmith: ok, but it is very similar code..
It would be better if findDropZonePlane was reused, and this code simplified.
I suggest changing findDropZonePlane, so that it allows path to be null, and returns a host only match.

Also, isIPAddress(targetDZNameOrHost) is currently called in the loop, the above change would avoid that,
but more importantly, I'm not sure you should set ipMatch=false if 'targetDZNameOrHost' is an IP..
That would mean that if a host in a dropzone resolved to the IP - it would not match it.

jakesmith (commented):
@wangkx - please see reply.

jakesmith (commented):
@wangkx - looks good, please squash.

jakesmith (approved):
@wangkx - looks good

RussWhitehead (approved):

wangkx:

ghalliday (approved):

PR #16892, opened 2023-01-24
HPCC-28612 Helm warning: TLS not enabled

shamser:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/helm/hpcc/templates/_warnings.tpl
+++ b/helm/hpcc/templates/_warnings.tpl
@@ -139,4 +139,25 @@ Pass in dict with root and warnings
   {{- $_ := set $warning "msg" (printf "Insecure feature enabled in ecl: %s " $ctx.insecureEclFeature) -}}
   {{- $_ := set $ctx "warnings" (append $ctx.warnings $warning) -}}
  {{- end -}}
+ {{- /* Warn if TLS not enabled */ -}}
+ {{- $_ := set $ctx "TLSdisabled" list -}}
+ {{- range $espservice := .root.Values.esp -}}
+  {{- if not $espservice.tls -}}
+   {{- if (and ($ctx.root.Values.certificates|default false) $ctx.root.Values.certificates.enabled) -}}
+    {{- $externalCert := (ne (include "hpcc.isVisibilityPublic" (dict "root" $ctx.root "visibility" $espservice.service.visibility)) "") -}}
+    {{- $externalIssuerKeyName := ternary "remote" "public" (eq "true" ( include "hpcc.usesRemoteClientCertificates" $espservice )) -}}
+    {{- $issuerKeyName := ternary $externalIssuerKeyName "local" $externalCert -}}
+    {{- if ne (include "hpcc.isIssuerEnabled" (dict "root" $ctx.root "issuerKeyName" "local")) "true" -}}
+     {{- $_ := set $ctx "TLSdisabled" (append $ctx.TLSdisabled $espservice.name) -}}
+    {{- end -}}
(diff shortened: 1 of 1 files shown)


helm/hpcc/templates/_warnings.tpl
```diff
@@ -124,4 +124,16 @@ Pass in dict with root and warnings
   {{- $_ := set $warning "msg" (printf "Default cpu cost rate is being used for %s: %s" ((len $ctx.defaultCpuRateComponents)| plural "component" "components") ($ctx.defaultCpuRateComponents|toStrings)) -}}
   {{- $_ := set $ctx "warnings" (append $ctx.warnings $warning) -}}
  {{- end -}}
+ {{- /* Warn if TSL not enabled */ -}}
```
afishbeck, line 127: typo TSL => TLS

helm/hpcc/templates/_warnings.tpl
```diff
@@ -124,4 +124,16 @@ Pass in dict with root and warnings
...
+ {{- $_ := set $ctx "TLSdisabled" list -}}
+ {{- range $espservice := .root.Values.esp -}}
+  {{- if not $espservice.tls -}}
```
afishbeck, line 130: I think it's a bit more complicated:

A. If the tls tag is there then this check is fine as is.
B. if the tls tag is missing then whether tls is enabled depends on other config:
    1. if certmanager is enabled, and the issuer this service uses is enabled then then tls is enabled and you shouldn't warn.
    2. otherwise tls is false and you should warn.

I'm sure you could simplify this or common up the code more but something like:

`{{- if (hasKey $espservice "tls" ) -}}
 {{- if not $espservice.tls -}}
  {{- $_ := set $ctx "TLSdisabled" (append $ctx.TLSdisabled $espservice.name) -}}
 {{- end -}}
{{- else if (and .root.Values.certificates .root.Values.certificates.enabled) -}}
 {{- $externalCert := (ne (include "hpcc.isVisibilityPublic" (dict "root" .root "visibility" $espservice.service.visibility)) "") -}}
 {{- $externalIssuerKeyName := ternary "remote" "public" (eq "true" ( include "hpcc.usesRemoteClientCertificates" $espservice )) -}}
 {{- $issuerKeyName := ternary $externalIssuerKeyName "local" $externalCert -}}
 {{- if ne (include "hpcc.isIssuerEnabled" (dict "root" .root "issuerKeyName" "local")) "true" -}}
  {{- $_ := set $ctx "TLSdisabled" (append $ctx.TLSdisabled $espservice.name) -}}
 {{- end -}}
{{- else -}} 
 {{- $_ := set $ctx "TLSdisabled" (append $ctx.TLSdisabled $espservice.name) -}}
{{- end -}}`

And if you think disabled issuers should be checked in a section specific to certmanager, that could be good as well, and the middle part of my code sample could be made a no-op.

shamser: Thank-you... it was  very useful explanation.

afishbeck: @shamser Can I assume this has been tested both with and without certificates enabled?

afishbeck (approved):
Looks good, I also tested it locally and it worked.

ghalliday:
Closed and reopened to rerun tests

PR #16894, opened 2023-01-24
HPCC-28478 Lower default containerized log level

rpastrana:
- Lowers default containerized log level to 60 (info)

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/helm/examples/logging/README.md
+++ b/helm/examples/logging/README.md
@@ -87,7 +87,7 @@ The available target categories include disaster(DIS), error(ERR), warning(WRN),
     helm install myhpcc ./hpcc --set dali[0].logging.classes="ALL-PRO" --set dali[0].name="mydali"
 
 ### Log Detail Level Configuration
-Log output verbosity can be adjusted from "critical messages only" (1) up to "report all messages" (100). The default log level is rather high (80) and should be adjusted accordingly.
+Log output verbosity can be adjusted from "critical messages only" (1) up to "report all messages" (100). The default log level is rather high (60) and should be adjusted accordingly.
 
     For example, verbosity should be medium for all components:
     helm install myhpcc ./hpcc --set global.logging.detail="50"
--- a/helm/hpcc/values.yaml
+++ b/helm/hpcc/values.yaml
@@ -23,7 +23,7 @@ global:
 
   # logging sets the default logging information for all components. Can be overridden locally
   logging:
-    detail: 80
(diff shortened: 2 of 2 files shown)


rpastrana:
@ghalliday there's a few ways to accomplish this depending on how intrusive we care to be, immediate impact and how much we want to avoid the log detail

- set logging.detail down to 60 from 80 (filters out programmer info, but relies on detail/level)
- Adjust DefaultDetail #define accordingly in code (can be overwritten in config, but again relies on log detail/level )
- set logging.classes: "ALL-INF" and logging.audiences: "ALL-PRO" (filters out user/operator info, not exactly what is being called for)
- Adjust defaultLogMsgFilter accordingly in code (can be overwritten in config, will affect bare metal)

Is one of these what you had in mind for this task?

ghalliday:
The plan is to completely remove detail from the system, so filtering on the detail will not help.   (E.g. see https://github.com/hpcc-systems/HPCC-Platform/pull/16650 which is currently stalled.  Before that is merged the few cases that do rely on the detail level need to switch to using feature flags.)

I suspect the idea in mind when the jira was created would be to set default log msg filter to match audiences "MSGAUD_all & ~MSGAUD_programmer".  The call from eclcc would need to change to stay the same as it currently is i.e. include programmer.

We also might need to do a careful search of all debugging lines to check the audience before this is enabled - otherwise key tracing information will not be available in the logs.

rpastrana:
Are we ok losing the ability to control log output on a escalating manner? Without the log level feature, we'll be able to filter out entire audiences, and or entire log types.

For example, if we filter out audience programmer and class info we will lose all programmer errors/warnings, and also user/operator info...

rpastrana:
If the end-goal is to implement a industry standard log control mechanism, we should control log output based on graduated levels such as:    Emergency, Alert, Critical, Error, Warning, Notice, Informational, Debug (https://sematext.com/blog/logging-levels/), each of those essentially map to various thresholds of the existing log level/detail.
The log message audience/class can remain more of an annotation for each log message.

rpastrana:
@ghalliday I still contend this would have lowered costs without significant debug information sacrifice.

PR #16897, opened 2023-01-25
HPCC-28687 Add metric naming restrictions

kenrowland:
Added restrictions to metric framework to only accept valid names

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [x] Scalability
  - [x] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/espcommon/espcommon.cpp
+++ b/esp/espcommon/espcommon.cpp
@@ -52,7 +52,17 @@ static std::vector<__uint64> getExecutionProfileBuckets(const char *profileOptio
     return buckets;
 }
 
-// Registers a profiling metric for an ESP service method
+// Generate metric name for the indicated service method and register a profiling metric
+ESPCOMMON_API std::shared_ptr<hpccMetrics::ScaledHistogramMetric> registerServiceMethodProfilingMetric(
+        const char *processName, const char *serviceName, const char *methodName,
+        const char *desc, const char *profilingOptions)
+{
+    std::string metricName(processName);
+    metricName.append(".").append(serviceName).append(".").append(methodName);
+    return registerProfilingMetric(metricName.c_str(), desc, profilingOptions);
+}
+
+// Registers a profiling metric
 ESPCOMMON_API std::shared_ptr<hpccMetrics::ScaledHistogramMetric> registerProfilingMetric(const char *histogramMetricName, const char *desc, const char *profilingOptions)
 {
     return hpccMetrics::registerCyclesToNsScaledHistogramMetric(histogramMetricName, desc, getExecutionProfileBuckets(profilingOptions));
--- a/system/jlib/jmetrics.cpp
+++ b/system/jlib/jmetrics.cpp
@@ -13,6 +13,7 @@
 
 #include "jmetrics.hpp"
 #include "jlog.hpp"
+#include <regex>
 
 using namespace hpccMetrics;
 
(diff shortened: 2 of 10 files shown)


kenrowland:
@afishbeck  You will notice a change to httpbinding. This makes sure the process name is actually set. During debug, I found that the constructor would exit prior to storing the name. The name is used in the generated code's binding constructor as a piece of the matric name.

Also, there is a change to some dali code to change the names of the current metrics created to meet the naming restrictions.

system/jlib/jmetrics.cpp
```diff
@@ -138,35 +139,49 @@ bool MetricsManager::addMetric(const std::shared_ptr<IMetric> &pMetric)
...
     }
     else
     {
-        // If there is a match only report an error if the metric has not been destroyed in the meantime
-        auto match = it->second.lock();
-        if (match)
-        {
 #ifdef _DEBUG
-            throw MakeStringException(MSGAUD_operator, "addMetric - Attempted to add duplicate named metric with name '%s'", name.c_str());
+        throw MakeStringException(MSGAUD_operator, "addMetric - Attempted to add metric with invalid name ('%s')", name.c_str());
```
afishbeck, line 181: Hmmm, I don't usually throw only in debug.  I'm often trying to reproduce behavior found in non-debug builds while stepping through or increasing logging by running debug.  This could get really confusing.  Or if I didn't know how to configure something to get past something like this I would have to modify the code to keep debugging.  We do sometimes have asserts though that only happen in debug.  Might want to see what Gavin thinks.

system/jlib/jmetrics.cpp
```diff
@@ -138,35 +139,49 @@ bool MetricsManager::addMetric(const std::shared_ptr<IMetric> &pMetric)
     bool rc = false;
     std::unique_lock<std::mutex> lock(metricVectorMutex);
     std::string name = pMetric->queryName();
-    auto metaData = pMetric->queryMetaData();
-    for (auto &metaDataIt: metaData)
-    {
-        name.append(".").append(metaDataIt.value);
-    }
+    std::regex pattern("^[A-Za-z][A-Za-z0-9.]*[^.]$");
```
afishbeck, line 143: Is the last character too permissive?

testing/unittests/metrics/MetricFrameworkTests.cpp
```diff
@@ -79,10 +80,51 @@ class MetricFrameworkTests : public CppUnit::TestFixture
...
 
+    void Test_valid_and_invalid_metric_names()
+    {
+        std::vector<std::string> invalid_names = {
+            {".startswithperiod"},
+            {"endswithperiod."},
+            {"9startswithnumber"},
+            {"has_underscore"},
+            {"has&^%$#@!-~`+=|[]{}?/<>,_specialcharacters"}
+        };
```
afishbeck, line 91: I think "has.a.&" would currently pass?  One special character at the end of the string?

tools/hidl/hidlcomp.cpp
```diff
@@ -5757,26 +5792,59 @@ void EspServInfo::write_esp_binding()
...
     {
-        if (mthi->isExecutionProfilingEnabled())
-        {
-            outf("\t%s = registerProfilingMetric(\"%s\", \"\", \"%s\");\n", mthi->getExecutionProfilingMetricName(), mthi->getName(), mthi->getExecutionProfilingOptions().c_str());
+        outf("#ifdef ESP_SERVICE_%s\n", name_);
+        outf("\nvoid C%sSoapBinding::init_metrics()\n", name_);
+        outs("{\n");
+        outs("\tstd::string processName(queryProcessName());\n");  // process/app name
+        outs("\tstd::string metricName;\n");
+        outs("\tmetricName.reserve(100);\n");
```
afishbeck, line 5806: I don't think this reserve buys you much?  get_metric_name()  does multiple appends to a locally declared string and then returns it by value... the return might be optimized by the compiler?   But I think if you did the reserve and then passed the string by reference into get_metric_name() then it might be more efficient doing those multiple appends directly into this variable with the length reserved?

tools/hidl/hidlcomp.cpp
```diff
@@ -5757,26 +5792,59 @@ void EspServInfo::write_esp_binding()
...
+
+    // only output the method to build the metric name if enabled for the service
+    if (executionProfilingEnabled)
+    {
+        // Retrieve Get metric name with illegal characters removed
+        outf("#ifdef ESP_SERVICE_%s\n", name_);
+        outf("\nstd::string C%sSoapBinding::get_metric_name(const char *processName, const char *serviceName, const char *methodName)\n",
+             name_);
+        outs("{\n");
+        //outs("\tstatic std::string illegalChars = \"_\"\n;");
```
afishbeck, line 5830: Can you remove the commented out lines?

system/jlib/jmetrics.cpp
kenrowland: Only throwing in debug was a design choice made by Gavin and myself. The reason is we wanted the use to know during development that they have provided an invalid metric and to halt system execution. However, in a production environment, we did not want to halt.

tools/hidl/hidlcomp.cpp
kenrowland: Correct. Before ensuring all generated code was ifdef'd out, I had an empty method, so I needed to return something, hence the local variable. Now that no code is created if not instrumented, I changed it to pass a reference as indicated.

tools/hidl/hidlcomp.cpp
kenrowland: removed.

system/jlib/jmetrics.cpp
afishbeck: I think the changed line is good.

afishbeck (commented):
@kenrowland some new comments / possible change.

afishbeck (approved):
Looks much cleaner this way.

kenrowland:
@ghalliday or @richardkchapman Please merge

ghalliday (commented):
@kenrowland one minor issue

kenrowland:
@ghalliday This should be ready for merge.

esp/espcommon/espcommon.cpp
```diff
@@ -52,8 +52,12 @@ static std::vector<__uint64> getExecutionProfileBuckets(const char *profileOptio
     return buckets;
 }
 
-// Registers a profiling metric for an ESP service method
-ESPCOMMON_API std::shared_ptr<hpccMetrics::ScaledHistogramMetric> registerProfilingMetric(const char *histogramMetricName, const char *desc, const char *profilingOptions)
+// Generate metric name and register it for the indicated service method
```
ghalliday, line 55: Note: You will probably require the original function later on.  This should probably be an additional function which calls the old one.

system/jlib/jmetrics.cpp
```diff
@@ -138,35 +139,49 @@ bool MetricsManager::addMetric(const std::shared_ptr<IMetric> &pMetric)
     bool rc = false;
     std::unique_lock<std::mutex> lock(metricVectorMutex);
     std::string name = pMetric->queryName();
-    auto metaData = pMetric->queryMetaData();
-    for (auto &metaDataIt: metaData)
-    {
-        name.append(".").append(metaDataIt.value);
-    }
+    std::regex pattern("^[A-Za-z][A-Za-z0-9.]*[A-Za-z0-9]$");
```
ghalliday, line 142: trivial: This should be a member of MetricManager, so that the pattern object can be created once (which can be expensive), and reused by all calls.  If this is going to be called for large numbers it may become significant.

ghalliday (commented):
@kenrowland a few trivial issues from a final check.  The regex moving within the class is may be significant enough to change.  I can merge as-is if you prefer.

kenrowland:
@ghalliday I went ahead and addressed your comments instead of waiting for another Jira. This code should be as clean as possible from the start. Please take look.

ghalliday (approved):
@kenrowland please squash

kenrowland:
@ghalliday sorry for delay, was on PTO. Ready to be merged.

PR #16912, opened 2023-02-01
HPCC-28868 Recode CHttpRequest::parseQueryString to avoid a temporary…

ghalliday:
… buffer

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jstring.cpp
+++ b/system/jlib/jstring.cpp
@@ -1595,6 +1595,28 @@ StringBuffer &appendDecodedURL(StringBuffer &s, const char *url)
     return s;
 }
 
+StringBuffer &appendDecodedURL(StringBuffer &s, size_t len, const char *url)
+{
+    size_t i = 0;
+    while (i < len)
+    {
+        char c = url[i++];
+        if (c == '+')
+            c = ' ';
+        else if (c == '%' && i + 1 < len)
+        {
+            if (isxdigit(url[i]) && isxdigit(url[i+1]))
+            {
+                c = translateHex(url[i], url[i+1]);
+                i+=2;
+            }
+        }
+        s.append(c);
+    }
+    return s;
+}
+
+
 static StringBuffer & appendStringExpandControl(StringBuffer &out, unsigned len, const char * src, bool addBreak, bool isCpp, bool isUtf8)
 {
     const int minBreakPos = 0;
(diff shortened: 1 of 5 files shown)


ghalliday:
@afishbeck I have not tested this change.  How speed critical is this piece of code?  I made the minimal changes to remove the strcpy, but more could be done if performance is important - e.g., reuse the stringbuffers and recode the loop slightly.

system/jlib/jstring.cpp
```diff
@@ -1595,6 +1595,28 @@ StringBuffer &appendDecodedURL(StringBuffer &s, const char *url)
     return s;
 }
 
+StringBuffer &appendDecodedURL(StringBuffer &s, size_t len, const char *url)
+{
+    size_t i = 0;
```
richardkchapman, line 1600: Should this function do something different if it encounters a null char?

ghalliday: If anything it should throw an exception - I'm not sure how a url encoded string can include a null.  I think it is probably enough to pass it through - it doesn't validate any other characters are valid.

afishbeck:
> @afishbeck I have not tested this change. How speed critical is this piece of code? I made the minimal changes to remove the strcpy, but more could be done if performance is important - e.g., reuse the stringbuffers and recode the loop slightly.

This code is called whenever an HTTP request has parameters after the '?' and for an HTTP POST with form encoded content (submitting a web form, or some REST calls).

Most calls that happen in volume would be SOAP or JSON and would have short strings after the '?' if any.

EclWatch does make rest calls in the FORM POST format.. but those usually happen because of user action, rather than large scale.

That doesn't stop someone from making any ESP call using a queryString or FORM POST which are forms of REST calls.  It's just less likely to be used at scale than SOAP or JSON.

That being said, optimizing this further would be great, just not sure of the priority based on what I said above.

afishbeck (commented):
@ghalliday your change looks good barring further optimization.  I was going to test but then got trapped in build error hell and I'll need to climb out of that tomorrow.  If you'd like I can test once I'm able.

ghalliday:
@afishbeck thanks.  Testing would be good - I'm in no rush to merge it.

afishbeck (approved):
@ghalliday I've tested this commit on a different branch to avoid my build issues.  It worked fine and the change looks good.  There is plenty of room for optimizations and commoning up code, but not needed for this PR.

PR #16936, opened 2023-02-07
HPCC-28154 TCP keepalive for all sockets from globals/expert config

mckellyln:
## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [x] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

## Testing:

manual testing

### DIFF
--- a/system/jlib/jsocket.cpp
+++ b/system/jlib/jsocket.cpp
@@ -368,6 +368,73 @@ enum SOCKETMODE { sm_tcp_server, sm_tcp, sm_udp_server, sm_udp, sm_multicast_ser
 # endif
 #endif
 
+static CriticalSection queryKACS;
+
+enum UseKA { UNINIT, DISABLED, ENABLED };
+static std::atomic<UseKA> doKeepAlive { UNINIT };
+static int keepAliveTime = -1;
+static int keepAliveInterval = -1;
+static int keepAliveProbes = -1;
+
+/*
+<Software>
+  <Globals>
+    <keepalive time="200" interval="75" probes="9"/>
+  </Globals>
+
+global:
+  expert:
+    keepalive:
+      time: 200
+      interval: 75
+      probes: 9
+*/
+
+static bool queryKeepAlive(int &time, int &intvl, int &probes)
+{
+    UseKA state = doKeepAlive.load();
+    if (state == UNINIT)
+    {
+        CriticalBlock block(queryKACS);
+        state = doKeepAlive.load();
+        if (state == UNINIT)
+        {
+#ifdef _CONTAINERIZED
+            Owned<IPropertyTree> expert = getGlobalConfigSP()->getPropTree("expert");
+#else
+            // MCK - without this many components will not have a global prop
+            IPropertyTree *expert = nullptr;
+            Owned<IPropertyTree> envtree = getHPCCEnvironment();
+            if (envtree)
+                expert = envtree->queryPropTree("Software/Globals");
+#endif
+            state = DISABLED;
+            if (expert)
+            {
+                IPropertyTree *keepalive = expert->queryPropTree("keepalive");
+                if (keepalive)
+                {
+                    keepAliveTime = keepalive->getPropInt("@time", keepAliveTime);
+                    keepAliveInterval = keepalive->getPropInt("@interval", keepAliveInterval);
+                    keepAliveProbes = keepalive->getPropInt("@probes", keepAliveProbes);
+                    state = ENABLED;
+                }
+            }
+            doKeepAlive = state;
+        }
+    }
+
+    if (state == ENABLED)
+    {
+        time = keepAliveTime;
+        intvl = keepAliveInterval;
+        probes = keepAliveProbes;
+        return true;
+    }
+    else
+        return false;
+}
+
 class CSocket: public ISocket, public CInterface
 {
 public:
@@ -505,6 +572,10 @@ private:
         else
             return 0;
     }
+
+    void checkCfgKeepAlive();
+    void setKeepAlive(bool set, int time, int intval, int probes);
+
 };
 
 CriticalSection CSocket::crit;
@@ -884,6 +955,9 @@ int CSocket::pre_connect (bool block)
         int err = SOCKETERRNO();
         THROWJSOCKEXCEPTION(err);
     }
+
+    checkCfgKeepAlive();
+
     STATS.activesockets++;
     int err = 0;
     set_nonblock(!block);
@@ -946,6 +1020,9 @@ void CSocket::open(int listen_queue_size,bool reuseports)
     if (sock == INVALID_SOCKET) {
         THROWJSOCKEXCEPTION(SOCKETERRNO());
     }
+
+    checkCfgKeepAlive();
+
     STATS.activesockets++;
 
 #ifdef SOCKTRACE
@@ -1098,12 +1175,56 @@ void CSocket::set_linger(int lingertime)
     }
 }
 
-void CSocket::set_keep_alive(bool set)
+void CSocket::setKeepAlive(bool set, int time, int intvl, int probes)
 {
     int on=set?1:0;
-    if (setsockopt(sock, SOL_SOCKET, SO_KEEPALIVE, (char*)&on, sizeof(on)) != 0) {
-        IWARNLOG("KeepAlive not set");
+    if (setsockopt(sock, SOL_SOCKET, SO_KEEPALIVE, (char*)&on, sizeof(on)) != 0)
+    {
+        OWARNLOG("KeepAlive not set");
+        return;
     }
+
+    if (!on)
+        return;
+
+    int srtn, optval;
+    socklen_t optlen = sizeof(optval);
+
+    if (time >= 0)
+    {
+        optval = time;
+        srtn = setsockopt(sock, IPPROTO_TCP, TCP_KEEPIDLE, &optval, optlen);
+        if (srtn != 0)
+            OWARNLOG("KeepAlive time not set");
+    }
+
+    if (intvl >= 0)
+    {
+        optval = intvl;
+        srtn = setsockopt(sock, IPPROTO_TCP, TCP_KEEPINTVL, &optval, optlen);
+        if (srtn != 0)
+            OWARNLOG("KeepAlive probes not set");
+    }
+
+    if (probes >= 0)
+    {
+        optval = probes;
+        srtn = setsockopt(sock, IPPROTO_TCP, TCP_KEEPCNT, &optval, optlen);
+        if (srtn != 0)
+            OWARNLOG("KeepAlive probes not set");
+    }
+}
+
+void CSocket::checkCfgKeepAlive()
+{
+    int time, intvl, probes;
+    if (queryKeepAlive(time, intvl, probes))
+        setKeepAlive(true, time, intvl, probes);
+}
+
+void CSocket::set_keep_alive(bool set)
+{
+    setKeepAlive(set, -1, -1, -1);
 }
 
 


mckellyln:
Without the bare-metal code in the ifdef most components do not seem to have the global config pointer and will not pick up the settings in the env.xml.  DaServer, saserver, thormaster, thorslave, dafilesrv, esp, eclscheduler all show as not finding the expert section in the config with a call to getGlobalConfigSP()->getPropTree("expert");

In the _CONTAINERIZED build, toposerver is the only component that does not pick up the global config.

mckellyln:
How should we add/describe/document the settings ?
```
<Software>
  <Globals>
    <expert>
      <keepalive time="200" interval="75" probes="9"/>
    </expert>
  </Globals>
```
and in helm values.yaml:
```
global:
  expert:
    keepalive:
      time: 200
      interval: 75
      probes: 9
```

system/jlib/jsocket.cpp
```diff
@@ -1098,12 +1173,63 @@ void CSocket::set_linger(int lingertime)
...
+        setKeepAlive(true, time, intvl, probes);
+}
+
+void CSocket::set_keep_alive(bool set)
+{
+    if (set)
+    {
+        int time, intvl, probes;
+        (void)queryKeepAlive(time, intvl, probes);
+        setKeepAlive(true, time, intvl, probes);
```
jakesmith, line 1229: Previously ISocket::set_keep_alive would only toggle the SO_KEEPALIVE option, now via setKeepAlive it will also make the system calls to set TCP_KEEPIDLE, TCP_KEEPINTVL and TCP_KEEPCNT , is that necessary/wanted?
If the global/export/keepalive settings are absent, it means it's going to force the keepalive idle/interval/probes to 7200,75,9- i.e. overridding the system defaults which may not be 7200,75.9

Why not keep as before and turn on keep alive only here (SO_KEEPALIVE)?
if global/expert/keepalive is configured, checkCfgKeepAlive() will set them before the socket is connect/pre_connect'd.

jakesmith: Could acheive this (previous semantics of set_keep_alive) by coding it as:
```
void CSocket::set_keep_alive(bool set)
{
    setKeepAlive(set, 0, 0, 0);
}
```

and having setKeepAlive bail out before setting the idle/interval/probes if e.g. time==0

system/jlib/jsocket.cpp
```diff
@@ -1098,12 +1173,63 @@ void CSocket::set_linger(int lingertime)
...
+    if (queryKeepAlive(time, intvl, probes))
+        setKeepAlive(true, time, intvl, probes);
+}
+
+void CSocket::set_keep_alive(bool set)
+{
+    if (set)
+    {
+        int time, intvl, probes;
+        (void)queryKeepAlive(time, intvl, probes);
```
jakesmith, line 1228: The semantics of (void)queryKeepAlive isn't that intuitive here - it fills out reference params with hard-wired defaults if it returns false.

It would be better I think if this method wasn't used here (see comments below).
And queryKeepAlive() didn't fill in the reference params if returned false, because elsewhere where used, the result is only used if queryKeepAlive returns true.

system/jlib/jsocket.cpp
```diff
@@ -368,6 +368,71 @@ enum SOCKETMODE { sm_tcp_server, sm_tcp, sm_udp_server, sm_udp, sm_multicast_ser
...
+
+global:
+  expert:
+    keepalive:
+      time: 200
+      interval: 75
+      probes: 9
+*/
+
+static bool queryKeepAlive(int &time, int &intvl, int &probes)
```
jakesmith, line 395: pedantic: shouldn't these really be unsigned's, since -ve would be invalid ?

jakesmith (commented):
@mckellyln - please see comments and questions.

system/jlib/jsocket.cpp
mckellyln: I agree, and have changed the code so that you can set keepalive on and then it will not set time, intvl, probes unless you have also set those (non zero).
I assume zero is not a good value, we could make those -1 but I dont think its much of an issue.

system/jlib/jsocket.cpp
```diff
@@ -368,6 +368,71 @@ enum SOCKETMODE { sm_tcp_server, sm_tcp, sm_udp_server, sm_udp, sm_multicast_ser
...
+        if (envtree)
+            expert = envtree->queryPropTree("Software/Globals/expert");
+#endif
+        if (expert)
+        {
+            bool hasKeepAlive = expert->hasProp("keepalive");
+            if (hasKeepAlive)
+            {
+                doKeepAlive = ENABLED;
+                keepAliveTime = expert->getPropInt("keepalive/@time", keepAliveTime);
```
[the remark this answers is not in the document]
mckellyln: That didn't seem to work for me, keepalive does not return as a valid proptree ?

system/jlib/jsocket.cpp
```diff
@@ -368,6 +368,71 @@ enum SOCKETMODE { sm_tcp_server, sm_tcp, sm_udp_server, sm_udp, sm_multicast_ser
...
+
+    if (doKeepAlive == UNINIT)
+    {
+        doKeepAlive = DISABLED;
+
+#ifdef _CONTAINERIZED
+        // NOTE: toposerver does not load globals
+        Owned<IPropertyTree> expert = getGlobalConfigSP()->getPropTree("expert");
+#else
+        // MCK - without this many components will not have a global prop
```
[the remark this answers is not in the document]
mckellyln: Yes - and that PR should change/remove this code.  Just wanted this PR to be effective at changing all sockets.

mckellyln:
@jakesmith made some changes from your review, thank you.  Please take a look at this next iteration when you have time.

system/jlib/jsocket.cpp
```diff
@@ -394,10 +394,10 @@ static int keepAliveProbes = 9;
 
 static bool queryKeepAlive(int &time, int &intvl, int &probes)
 {
-    CriticalBlock block(queryKACS);
-
     if (doKeepAlive == UNINIT)
```
jakesmith, line 397: this change can cause problems (races).
Another thread could call queryKeepAlive whilst a thread is in the CS block and has just set DISABLED for example, meaning the other thread will return false spuriously.
I think it would be ok, if you ensure that doKeepAlive set is at the end of the block, after any initialization is checked/done.

system/jlib/jsocket.cpp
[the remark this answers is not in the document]
jakesmith: it should be directly equivalent.
>keepalive does not return as a valid proptree ?

you mean it returns null, or?

system/jlib/jsocket.cpp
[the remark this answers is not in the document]
jakesmith: that's fine, I'll open another JIRA and reference in this one for the missing 'global' sections in other components.

system/jlib/jsocket.cpp
jakesmith: Right, probably -1 is better.

system/jlib/jsocket.cpp
```diff
@@ -368,6 +368,71 @@ enum SOCKETMODE { sm_tcp_server, sm_tcp, sm_udp_server, sm_udp, sm_multicast_ser
...
+            expert = envtree->queryPropTree("Software/Globals/expert");
+#endif
+        if (expert)
+        {
+            bool hasKeepAlive = expert->hasProp("keepalive");
+            if (hasKeepAlive)
+            {
+                doKeepAlive = ENABLED;
+                keepAliveTime = expert->getPropInt("keepalive/@time", keepAliveTime);
+                keepAliveIntvl = expert->getPropInt("keepalive/@intvl", keepAliveIntvl);
```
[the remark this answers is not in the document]
jakesmith: I think a non-abbreviated descriptive form for the configuration option, e.g. @interval

system/jlib/jsocket.cpp
jakesmith: did you agree with this one?

jakesmith (commented):
@mckellyln - back to you

system/jlib/jsocket.cpp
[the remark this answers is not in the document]
mckellyln: ok, changed.

system/jlib/jsocket.cpp
mckellyln: changed

system/jlib/jsocket.cpp
mckellyln: left as ints so we can easily use -1 as not specified in config.

system/jlib/jsocket.cpp
mckellyln: Yes, thanks, coded it in too much of a hurry, added the extra if check.

system/jlib/jsocket.cpp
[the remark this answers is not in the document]
mckellyln: You are correct, it should be fine.  I double checked and changed the code.

system/jlib/jsocket.cpp
mckellyln: If set_keep_alive(true) is called I think we should use the global keepalive time, interval, probes settings, if present.
So I would like to leave the query in to get those.

mckellyln:
@jakesmith thanks, made the changes, let me know what you think.

system/jlib/jsocket.cpp
jakesmith: my earlier point, was that the custom settings will be applied anyway, without setting here in set_keep_alive,
because checkCfgKeepAlive is called per socket before the in open() or pre_connect() and the custom idle, interval and probes will be applied there.
But maybe I missing something?

system/jlib/jsocket.cpp
jakesmith: @mckellyln - the idea or our convension for the 'Resolved conversation' button, is for it to be used by the reviewer only, so it helps the reviewer to track their comments: 
https://github.com/hpcc-systems/HPCC-Platform/blob/master/devdoc/CodeReviews.md#:~:text=Only%20the%20reviewer%20should%20mark%20issues%20as%20resolved%20using%20the%20Github%20resolve%20conversation%20button.

system/jlib/jsocket.cpp
jakesmith: It's still an issue. e.g.:
Thread 1 (T1), gets to line 415.
Thread 2 (T2), hits line 397, and doKeepAlive is DISABLED , T2 incorrectly returns false.

>if you ensure that doKeepAlive set is at the end of the block, after any initialization is checked/done.

still think you need to do this, ensure it's not changed from UNINIT until after initialization is done (or fails), then other threads will enter the crit block.

system/jlib/jsocket.cpp
```diff
@@ -368,6 +368,77 @@ enum SOCKETMODE { sm_tcp_server, sm_tcp, sm_udp_server, sm_udp, sm_multicast_ser
...
+            Owned<IPropertyTree> envtree = getHPCCEnvironment();
+            if (envtree)
+                expert = envtree->queryPropTree("Software/Globals/expert");
+#endif
+            if (expert)
+            {
+                IPropertyTree *keepalive = expert->queryPropTree("keepalive");
+                if (keepalive)
+                {
+                    doKeepAlive = ENABLED;
```
jakesmith, line 420: also too early, another thread() could enter queryKeepAlive, just after this thread sets doKeepAlive  to ENABLED, and thefore the other thread(s) will skip the block and set the reference params to the yet as unset keepAliveTime, keepAliveInterval and keepAliveProbes

jakesmith (commented):
@mckellyln - please see replies.

system/jlib/jsocket.cpp
mckellyln: ok, I see your point, if already on then set_keep_alive(true) doesn't really need to do anything, or maybe at the minimum do what it always did.

system/jlib/jsocket.cpp
mckellyln: Yes, agree.  Will fix.

mckellyln: I could hold the CS always, or CriticalUnblock ub(queryKACS) after the keepalive check.  Which do you think is better ?

system/jlib/jsocket.cpp
mckellyln: Agreed

jakesmith (commented):
@mckellyln - final comments, but could be merged as is. Please make any final changes and squash.

system/jlib/jsocket.cpp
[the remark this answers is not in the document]
jakesmith: New JIRA : https://track.hpccsystems.com/browse/HPCC-28914

system/jlib/jsocket.cpp
```diff
@@ -368,6 +368,73 @@ enum SOCKETMODE { sm_tcp_server, sm_tcp, sm_udp_server, sm_udp, sm_multicast_ser
...
+static int keepAliveTime = -1;
+static int keepAliveInterval = -1;
+static int keepAliveProbes = -1;
+
+/*
+<Software>
+  <Globals>
+    <expert>
+      <keepalive time="200" interval="75" probes="9"/>
+    </expert>
```
jakesmith, line 384: in fact, \<expert\> is generated based on whatever is in Software/Globals.
It's because there are other sections generated into the bare-metal component configs from elsewhere (e.g. RemoteStorage), so these settings should look like:
```
Software>
  <Globals>
    <keepalive time="200" interval="75" probes="9"/>
  </Globals>
```

system/jlib/jsocket.cpp
```diff
@@ -368,6 +368,73 @@ enum SOCKETMODE { sm_tcp_server, sm_tcp, sm_udp_server, sm_udp, sm_multicast_ser
...
+        doKeepAlive = DISABLED;
+
+#ifdef _CONTAINERIZED
+        Owned<IPropertyTree> expert = getGlobalConfigSP()->getPropTree("expert");
+#else
+        // MCK - without this many components will not have a global prop
+        IPropertyTree *expert = nullptr;
+        Owned<IPropertyTree> envtree = getHPCCEnvironment();
+        if (envtree)
+            expert = envtree->queryPropTree("Software/Globals/expert");
```
jakesmith, line 410: see comment (https://github.com/hpcc-systems/HPCC-Platform/pull/16936/files#r1100062627).
This should be:
```
expert = envtree->queryPropTree("Software/Globals");
```

mckellyln:
@jakesmith went with avoiding the CS for the initial check, didn't seem like the other was was much different than the basic method of always holding CS.
Squashed.
One final pass and then let me know if ok to merge.

jakesmith (commented):
@mckellyln - a couple of the most recent comments missed, re. dropping /expert in bare-metal, passing back to you for now.

system/jlib/jsocket.cpp
mckellyln: updated

jakesmith (approved):
@mckellyln - looks good.

@ghalliday - please merge.

system/jlib/jsocket.cpp
```diff
@@ -1098,14 +1175,67 @@ void CSocket::set_linger(int lingertime)
...
+    int srtn, optval;
+    socklen_t optlen = sizeof(optval);
+
+    if (time >= 0)
+    {
+        optval = time;
+        srtn = setsockopt(sock, IPPROTO_TCP, TCP_KEEPIDLE, &optval, optlen);
+        if (srtn != 0)
+        {
+            IWARNLOG("KeepAlive time not set");
```
ghalliday, line 1199: Should this be OWARNLOG?  Is it something operations should know is failing?

system/jlib/jsocket.cpp
```diff
@@ -1098,14 +1175,67 @@ void CSocket::set_linger(int lingertime)
...
+            return;
```
ghalliday, line 1200: Is it better to return, or try the other options?

ghalliday (commented):
@mckellyln Looks good, a couple of questions.

system/jlib/jsocket.cpp
mckellyln: ok, good point, I will continue on to the other options.

system/jlib/jsocket.cpp
mckellyln: Probably, I will change.

mckellyln:
@ghalliday made those small changes.

ghalliday (approved):
@mckellyln looks good, please squash

system/jlib/jsocket.cpp
```diff
@@ -368,6 +368,73 @@ enum SOCKETMODE { sm_tcp_server, sm_tcp, sm_udp_server, sm_udp, sm_multicast_ser
...
+  expert:
+    keepalive:
+      time: 200
+      interval: 75
+      probes: 9
+*/
+
+static bool queryKeepAlive(int &time, int &intvl, int &probes)
+{
+    UseKA state = doKeepAlive.load();
```
richardkchapman, line 395: Can the standard "singleton" pattern be used?

mckellyln:
Squashed

system/jlib/jsocket.cpp
ghalliday: @richardkchapman Our singleton classes are mainly for pointers, and I think std::once has inefficiencies.  I would be inclined to merge as is.

ghalliday:
@mckellyln how does this work in containerized mode?  Does the schema need extending?

ghalliday:
ok, so the schema does not need to be changed to support the option.  We should have some kind of documentation of the options that can be contained in the expert section - especially if we think they are likely to be needed.

jakesmith:
@mckellyln - can you open a documentation JIRA

> ok, so the schema does not need to be changed to support the option. We should have some kind of documentation of the options that can be contained in the expert section - especially if we think they are likely to be needed.

I've open a JIRA: https://track.hpccsystems.com/browse/HPCC-28925

mckellyln:
Thanks for the Doc JIRA.
Right, no schema change needed.

ghalliday:
@mckellyln should this be merged to 8.12.x instead of 8.10.x, or is there a need for 8.10.x?

mckellyln:
Some groups are using a mix of on-prem and cloud and this PR is mostly to address that situation, when a firewall for traffic to/from AZ closes idle connections.  I thought the best target would then be 8.10.x to be in both 8.10 and 8.12.
Another note is that for 8.12 only, this code could change now that https://track.hpccsystems.com/browse/HPCC-28914 has been merged.  We were planning an addl PR for 8.12.x for that change after this PR was merged.

PR #16940, opened 2023-02-08
HPCC-28802 Install vcpkg third party headers

GordonSmith:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -216,6 +216,10 @@ if(APPLE)
     HPCC_ADD_SUBDIRECTORY(package)
 endif(APPLE)
 
+if (NOT CLIENTTOOLS_ONLY)
+    install(DIRECTORY ${VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}/include DESTINATION ${COMPONENTFILES_DIR}/cl FILES_MATCHING PATTERN "*")
+endif()
+
 ###
 ## CPack install and packaging setup.
 ###


dcamper (approved):
Looks good.

richardkchapman:
Will this install only the "outward facing" headers for these libraries, or are we going to have every header used in any third-party library when it is built from source?

The former is probably ok, the latter feels like it would be a nightmare (both from a namespace pollution and a size perspective).

dcamper:
@richardkchapman The headers need to be complete at compile time, so I would think that all headers for a library would need to be available because the "public" header may include "private" headers from the same build.

This is really a nasty problem, especially when you consider conditional library inclusion -- different platform builds get different subsets of the 3rd party libraries we're building -- and cross-library dependencies.  For instance, libicu may require libfoo.so, but libfoo.so is used on only some of our platform builds for whatever reason, and libfoo.so is provided by the OS everywhere.  Because libfoo is needed by some of our builds, we build it separately so the headers are present, but then we only conditionally install it.  That scenario may not exist in the real world, but it is a logical possibility.

Installing all 3rd party headers actually inverts the problem the Jira defined:  Instead of old (OS-provided) headers and new libraries supplied by us, we will now have a scenario where new headers (provided by us) are used with old (OS-provided) libraries.

I don't have a good solution here.  I just wanted to get some thoughts recorded.

GordonSmith:
> Will this install only the "outward facing" headers for these libraries, or are we going to have every header used in any third-party library when it is built from source?
> 
> The former is probably ok, the latter feels like it would be a nightmare (both from a namespace pollution and a size perspective).

This is the current behaviour (which is sub optimal, but will get better):
1. We list all the third-party dependencies directly referenced by HPCC in vcpkg.json
2. Prior to building HPCC, these dependencies are expanded to include their dependencies (recursively) 
3. This longer list is then downloaded, built and installed to the build folder
4. Then HPCC-Platform is built
5. When creating the install scripts and/or install packages, library dependencies are checked against the vcpkg install and when matched they get added to install package (this is done via LDD or similar).
6. We add ALL the vcpkg installed headers to the package (I had started to see if I could filter this list to match the potentially smaller list of installed third party libs - but ran out of time).

Note:  These headers and libraries only come into play when ECL Workunits are compiled that have embedded c++ that `#include` a library from the vcpkg list.

ghalliday (approved):
@GordonSmith Approving, but created another jira to see if we can reduce the size.

PR #16942, opened 2023-02-08
HPCC-28460 Extraneous error logging from secure socket accept

mckellyln:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

## Testing:

manual testing by port scanning secure listeners

### DIFF
--- a/system/security/securesocket/securesocket.cpp
+++ b/system/security/securesocket/securesocket.cpp
@@ -655,7 +655,7 @@ int CSecureSocket::secure_accept(int logLevel)
         // which can happen with port scan / VIP ...
         // NOTE: ret could also be SSL_ERROR_ZERO_RETURN if client closed
         // gracefully after ssl neg initiated ...
-        if ( (logLevel >= SSLogNormal) || (ret != SSL_ERROR_SYSCALL) )
+        if ( (logLevel > SSLogNormal) || (ret != SSL_ERROR_SYSCALL) )
         {
             char errbuf[512];
             ERR_error_string_n(ERR_get_error(), errbuf, 512);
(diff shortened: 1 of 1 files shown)


mckellyln:
We should really get the logging level detail from each component that ends up making this call, and use that here, and this can become a follow-on JIRA issue to properly propagate the logging level through to secure socket functions.
But in this specific case we know its an accept() failure from load balancer / port scanner so we can use this quick change to eliminate excessive logging.

system/security/securesocket/securesocket.cpp
```diff
@@ -666,8 +666,12 @@ int CSecureSocket::secure_accept(int logLevel)
     else if(err < 0)
     {
         int ret = SSL_get_error(m_ssl, err);
+        unsigned long errnum = ERR_get_error();
+        // See note above, SSL_accept() might also return -1
+        if ( (logLevel < SSLogNormal) && (ret == SSL_ERROR_SYSCALL) && (errnum == 0) )
```
jakesmith, line 671: question: should this be <= SSLogNormal ?

jakesmith: > && (errnum == 0)

If this is intented (return only when ERR_get_error returns no error?), it could with a comment,
or wa it supposed to be || (errnum == 0) ?

jakesmith (commented):
@mckellyln - a couple of questions.

system/security/securesocket/securesocket.cpp
mckellyln: Yes, its exactly like the above case, except when err == -1 we can also get errnum and in this particular situation it is zero.
So its either we add the additional errnum == 0 check or we skip it, its not meant to be an || errnum == 0.
Like above I thought SSLogNormal would still log, but its my opinion.  We could make this and above be only if > SSLogNormal.

mckellyln:
@jakesmith I could change to only log these port check type failed accepts when logLevel > SSLogNormal if you think its better that way.  I think perhaps yes.

system/security/securesocket/securesocket.cpp
jakesmith: ok, it's only that it's non obvious from reading the code (ERR_get_error() from man implies if 0 there is no error).
Perhaps add a small comment to clarify for future readability.

mckellyln: changed check and added comment.

jakesmith (approved):
@mckellyln - looks good, please squash

ghalliday (approved):

PR #16962, opened 2023-02-13
HPCC-28939 Increase timeout duration of component launch in hpcc-init

dcamper:

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/initfiles/bash/etc/init.d/hpcc_common.in
+++ b/initfiles/bash/etc/init.d/hpcc_common.in
@@ -460,6 +460,10 @@ startCmd() {
     local RCSTART=0
     local SENTINEL_CHECK=1
 
+    if [[ ${compType} = "dali" ]]; then
+      WAITTIME=720
+    fi
+
     if [[ ${compType} = "dafilesrv" || ${compType} = "backupnode" ]]; then
       SENTINEL_CHECK=0
     fi
@@ -598,6 +602,9 @@ startCmd() {
                 cleanupRuntimeEnvironment
                 return 1;
             fi
+        fi
+        if ! ((WAITTIME % 60)); then
+            echo "still starting ..."
         fi
         sleep 1
     done


dcamper:
FYI @jakesmith

jakesmith:
@Michael-Gardner @dcamper - 10 minutes might not be enough.. (e.g. it starting up cold with a huge transaction log..)
Perhaps we should ensure Dali does start up before proceeding and trace messages that still starting or similar, as it's critical it is working for almost all other components,

dcamper:
@jakesmith I agree with your comments, but I think it depends on what the purpose (or goal) of this change is.  What I was aiming for was just a "mitigation" rather than an outright solution, reducing the number of times someone would see the TIMEOUT indicator and temporarily-alarming subsequent messages.  This change does not outright remove that behavior, just reduces its occurrence.

If the goal should be to make a clean and accurate representation of what is going on during the launch cycle then we may need quite a bit more scaffolding to monitor Dali.

jakesmith:
> @jakesmith I agree with your comments, but I think it depends on what the purpose (or goal) of this change is. What I was aiming for was just a "mitigation" rather than an outright solution, reducing the number of times someone would see the TIMEOUT indicator and temporarily-alarming subsequent messages. This change does not outright remove that behavior, just reduces its occurrence.
> 
> If the goal should be to make a clean and accurate representation of what is going on during the launch cycle then we may need quite a bit more scaffolding to monitor Dali.

it should probably be both.
This one is okay as a stop-gap, to mitigate the issue for those with large stores (or cold starts with large transaction histories), but we also need a JIRA for a more formal improvement as described above I think.

dcamper:
This is specific to bare-metal environments, correct?  Just trying to determine priorities.

jakesmith:
> This is specific to bare-metal environments, correct? Just trying to determine priorities.

yep.

dcamper:
@Michael-Gardner Jake mentioned that this older PR may still be needed, but made Dali-specific. Please review.  Thanks!

Michael-Gardner (approved):
@dcamper This looks fine for baremetal. Have you tested it with an environment.xml you know will cause a timeout?  If you haven't, I can test it today.  Otherwise changes look good and approved.

ghalliday (commented):
Looks like a good change to me - Michael if you can confirm that would be good - otherwise I will merge anyway.

Michael-Gardner:
@ghalliday It's working fine for me. Lets merge.

ghalliday:
@dcamper please squash and I will merge.

dcamper:
@ghalliday Squashed and ready for merge. Thanks!

PR #16972, opened 2023-02-15
HPCC-28920 Roxie crash in DBGLOG from deblacklist()

richardkchapman:
Additional fix was needed.

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

## Testing:
Reproduced via additional tracing in thorsoapcall (left in but commented out) to make the problem easier to provoke, then running a query in server mode that made a soapcall to a non-existant port.

### DIFF
--- a/system/jlib/jthread.cpp
+++ b/system/jlib/jthread.cpp
@@ -371,6 +371,17 @@ void Thread::init(const char *_name)
     stacksize = 0; // default is EXE default stack size  (set by /STACK)
 }
 
+void Thread::getThreadLoggingInfo()
+{
+    ::getThreadLoggingInfo(logctx, traceFlags);
+}
+
+void Thread::setThreadLoggingInfo(const IContextLogger *_logctx, TraceFlags _traceFlags)
+{
+    logctx = _logctx;
+    traceFlags = _traceFlags;
+}
+
 void Thread::start()
 {
     if (alive) {
@@ -382,7 +393,6 @@ void Thread::start()
         return;
     }
     Link();
-    getThreadLoggingInfo(logctx, traceFlags); // New thread uses context from parent. This may or may not be a good idea by default!
     startRelease();
 }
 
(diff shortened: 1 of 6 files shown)


system/jlib/jthread.cpp
```diff
@@ -874,6 +874,12 @@ class CPooledThreadWrapper: public Thread
     {
         parent.numrunning++;
     }
+    virtual void start() override
```
richardkchapman, line 877: Arguably a better fix is to move the capture of logctx from Thread::start into Thread::Thread constructor

mckellyln (approved):
Looks good.

ghalliday:
Just to clarify - this fixes the problem because a thread may actually be started much later than it is created - in which case it will potentially refer to a context that then gets destroyed?

ghalliday (commented):
Is a similar change required for CThreadedPersistent?

ghalliday:
@richardkchapman

mckellyln:
https://track.hpccsystems.com/browse/HPCC-28955

richardkchapman:
In the particular instance that is crashing, the thread is started to asynchronously clean up (deblacklist a socket), so the thread that is active at the time is very likely to have terminated before the started thread is. The thread pool however is always going to outlive the threads in it.

There was code present to try to ensure that the context was captured at the time that the threadpool was created, but it didnt work properly. This change ensures that the context for threadpools IS captured properly at threadpool creation time, but also means the context for threads is captured at thread creation time rather than thread start time. THis is more likely to be safe, though neither is gruaranteed.

ThreadedPersistent are started at creation time so the change does not affect their behaviour.

mckellyln (approved):
Looks ok to merge

ghalliday (approved):

PR #17003, opened 2023-02-24
HPCC-28770 Add value masking to ESDL script syntax

timothyklemm:
- Rename trace operation to trace-content.
- Support the trace operation name for backward compatibility.
- Define trace-value operation.

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/esdllib/docs/trace-content.md
+++ b/esp/esdllib/docs/trace-content.md
@@ -1,0 +1,30 @@
+
+### trace-content
+```xml
+    <trace-content
+        class="enumerated text literal"
+        content_type="enumerated text literal"
+        label="text literal"
+        select="XPath evaluable to text"
+        skip_mask="XPath evaluable to Boolean"
+        test="XPath evaluable to Boolean"/>
+```
+
+A childless ESDL script operation used to create a trace output entry using script generated content. If a data masking engine is configured, a masker has been selected prior to this operation, and `skip_mask` is not *true*, the generated content will be masked. The masker&apos;s `maskContent` method is invoked to locate and mask any number of values within the selected text.
+
+| Property | Count | Description |
+| :- | :-: | :- |
+| class | 0..1 | Choice of how the message should be classified, one of **error**, **warning**, **information**, or **progress**. Default when omitted or empty is **information**. |
+| content_type | 0..1 | Choice of how a data masking profile may filter the rules applied when masking content. May be empty, which requests not filter. Should be one of the content type identifiers present in data masking profiles, which requests rules be filtered by the identifier.<br/><br/>For example, if your profiles use both *xml* and *json* to describe filters for XML and JSON content, then this would be a choice of empty, *xml*, or *json*. Use of any other value is equivalent to empty. |
+| label | 0..1 | User defined prefix appearing before selected text. |
+| select | 1..1 | XPath expression that evaluates to the scalar value to be recorded. If the XPath evaluates to an XML node set, the scalar begins as the serialization of each XML node in the set. If the XPath evaluates to a value, the scalar begins as the evaluated value. The scalar may be modified by a selected data masking profile context if one is selected. |
+| skip_mask | 0..1 | XPath expression that controls whether generated data bypasses configured data masking. Masking is bypassed only when this expression evaluates to *true*.<br/><br/>Masking content can be a time consuming process. If the selected data is known to contain no values to be masked, the performance impact can be avoided with this property. The property should be omitted or evaluate to *false* if there is any chance the data contains a value requiring masking. |
+| test | 0..1 | XPath expression that controls whether trace content should be generated. |
+
+Trace output will be generated unless `@test` is present and evaluates to *false*. Neither the availability of data masking nor its ability to mask a particular value are prerequisites. Script authors are responsible for limiting the inclusion of restricted values, such as passwords or PII.
+
+Future development will extend the ability to use, manipulate, and inspect masking capabilities:
+
+- [Extend ESDL script syntax to mask data within an XPath](https://track.hpccsystems.com/browse/HPCC-28771)
+- [Extend ESDL script syntax to interrogate data masking capabilities](https://track.hpccsystems.com/browse/HPCC-28772)
+- [Extend ESDL script syntax to control the data masking environment](https://track.hpccsystems.com/browse/HPCC-28770)
--- a/esp/esdllib/docs/trace-value.md
+++ b/esp/esdllib/docs/trace-value.md
@@ -1,0 +1,36 @@
+
+### trace-value
+```xml
+    <trace-value
+        class="enumerated text literal"
+        label="text literal"
+        mask_style="text literal"
+        select="XPath evaluable to text"
+        test="XPath evaluable to Boolean"
+        value_type="text literal"
+        xpath_mask_style="XPath evaluable to text"
+        xpath_value_type="XPath evaluable to text"/>
+```
+
+A childless ESDL script operation used to create a trace output entry using script generated content. If a data masking engine is configured, a masker has been selected prior to this operation, and `skip_mask` is not *true*, the generated content will be masked. If `select` evaluates to either a string value or a single element, the masker&apos;s `maskValue` method is invoked.
+
+> Trace output is not generated and a warning is recorded if `select` resolves to multiple elements. This can only output a single value.
+
+| Property | Count | Description |
+| :- | :-: | :- |
+| class | 0..1 | Choice of how the message should be classified, one of **error**, **warning**, **information**, or **progress**. Default when omitted or empty is **information**. |
+| label | 0..1 | User defined prefix appearing before selected text. |
+| mask_style | 0..1 | Choice of how an indidual data value should be masked. Should be empty or one of the mask style labels defined by the selected value type (see `value_type` or `xpath_value_type` for details on value type selection).<br/><br/>Only accepted when either `value_type` or `xpath_value_type` is given. Ignored when `xpath_mask_style` is given. |
+| select | 1..1 | XPath expression that evaluates to the scalar value to be recorded. If the XPath evaluates to an XML node set, the scalar begins as the serialization of each XML node in the set. If the XPath evaluates to a value, the scalar begins as the evaluated value. The scalar may be modified by a selected data masking profile context if one is selected. |
+| test | 0..1 | XPath expression that controls whether trace content should be generated. |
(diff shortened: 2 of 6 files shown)


esp/esdllib/esdl_script.cpp
```diff
@@ -1671,6 +1671,78 @@ class CEsdlTransformOperationTrace : public CEsdlTransformOperationWithoutChildr
...
+    }
+
+    virtual bool getText(StringBuffer& text, IXpathContext* targetContext, IXpathContext* sourceContext, IDataMaskingProfileContext* masker) override
+    {
+        bool isValue = true;
+        if (!targetContext->selectText(m_select, text, isValue))
+        {
+            if (!isValue)
+            {
+                recordWarning(ESDL_SCRIPT_Warning, "ambiguous value xpath");
```
afishbeck, line 1727: Should the xpath that turned out to be ambiguous be output in the warning message to make it easier to track down?

esp/esdllib/esdl_script.cpp
```diff
@@ -1671,6 +1671,78 @@ class CEsdlTransformOperationTrace : public CEsdlTransformOperationWithoutChildr
...
+                return false;
+            }
+        }
+        if (!masker)
```
afishbeck, line 1731: You do some work up above but then just return here.  Should the masker be checked at the top?

afishbeck (commented):
@timothyklemm some questions.

esp/esdllib/esdl_script.cpp
timothyklemm: No. The work above - getting the base text that will be output by the operation - is always required. The absence of a masker is not an error; it means that the base text will be output unchanged. Then we attempt to apply the mask, which could fail.

esp/esdllib/esdl_script.cpp
timothyklemm: Updated warning messages for both warning scenarios, and added tests to confirm that both scenarios do block trace output.

afishbeck (approved):

esp/esdllib/esdl_script.cpp
```diff
@@ -1615,6 +1671,78 @@ class CEsdlTransformOperationTrace : public CEsdlTransformOperationWithoutChildr
...
+    }
+
+    virtual bool getText(StringBuffer& text, IXpathContext* targetContext, IXpathContext* sourceContext, IDataMaskingProfileContext* masker) override
+    {
+        bool isValue = false;
+        if (!targetContext->selectText(m_select, text, isValue) || !masker)
+            return true;
+        if (!m_skipMask || !sourceContext->evaluateAsBoolean(m_skipMask))
+        {
+            char* buffer = const_cast<char*>(text.str());
```
asselitx, line 1700: I know that this is really the only realistic way to manage this but could we be opening ourselves up to problems by casting away the const? Are we guaranteed that masking will only ever change the contents and not the length?

asselitx (changes requested):

esp/esdllib/docs/trace-content.md
```diff
@@ -0,0 +1,30 @@
...
+    <trace-content
+        class="enumerated text literal"
+        content_type="enumerated text literal"
+        label="text literal"
+        select="XPath evaluable to text"
+        skip_mask="XPath evaluable to Boolean"
+        test="XPath evaluable to Boolean"/>
+```
+
+A childless ESDL script operation used to create a trace output entry using script generated content. If a data masking engine is configured, a masker has been selected prior to this operation, and `skip_mask` is not *true*, the generated content will be masked. The masker&apos;s `maskContent` method is invoked to locate and mask any number of values within the selected text.
```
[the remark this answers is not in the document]
timothyklemm: It can be, depending on the markdown editor in use. I frequently use an online editor that does require it. I suggest tabling this pending completion of https://track.hpccsystems.com/browse/HPCC-29109.

esp/esdllib/docs/trace-value.md
```diff
@@ -0,0 +1,36 @@
+
+### trace-value
+```xml
+    <trace-value
+        class="enumerated text literal"
+        label="text literal"
+        mask_style="text literal"
+        select="XPath evaluable to text"
+        test="XPath evaluable to Boolean"
+        value_type="text_literal"
```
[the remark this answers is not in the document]
timothyklemm: The underscore will be removed. "Standardizing" terminology is part of https://track.hpccsystems.com/browse/HPCC-29109.

esp/esdllib/docs/trace-content.md
```diff
@@ -0,0 +1,30 @@
+
+### trace-content
+```xml
+    <trace-content
+        class="enumerated text literal"
+        content_type="enumerated text literal"
+        label="text literal"
+        select="XPath evaluable to text"
+        skip_mask="XPath evaluable to Boolean"
+        test="XPath evaluable to Boolean"/>
```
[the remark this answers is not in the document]
timothyklemm: https://track.hpccsystems.com/browse/HPCC-29109

esp/esdllib/esdl_script.cpp
timothyklemm: A cast seems to be necessary to avoid making copies of the data. We could define an encapsulation that prevents a plugin from directly manipulating the underlying character buffer, while also avoiding copies. Such an encapsulation would need to reinvent most (all) of the standard library string functionality in ways that don't expose the underlying buffer.

If it's important enough to do, it should probably be done as part of HPCC-27471 since it impacts the plugin interface. @afishbeck, @asselitx, thoughts?

timothyklemm: HPCC-27471 has now been merged...

asselitx: That is a big enough of a task I don't think it would make sense to do on this PR as you noted. I think Tony has a better sense of of how strongly this is discouraged. I do see a few other instances in the codebase.

asselitx (approved):

ghalliday:
@afishbeck please review

esp/esdllib/esdl_script.cpp
afishbeck: @ghalliday any thoughts on casting away const and passing the string to a mask routine that never changes the length, just masks characters?  TraceValue may be dealing with small strings, but TraceContent could potentially be large blobs.  For example some xml content.

system/xmllib/libxml_xpathprocessor.cpp
```diff
@@ -991,8 +991,24 @@ class CLibXpathContext : public CInterfaceOf<IXpathContext>, implements IVariabl
         switch (obj->type)
         {
             case XPATH_NODESET:
-                toXMLNodeSet(obj, content);
-                isValue = false;
+                if (isValue)
```
afishbeck, line 994: I'm not sure if it's a big deal, but I'm not a huge fan of using the same parameter as an input that may or may not be modified to become an output.  It can get a bit confusing.

system/xmllib/libxml_xpathprocessor.cpp
```diff
@@ -991,8 +991,24 @@ class CLibXpathContext : public CInterfaceOf<IXpathContext>, implements IVariabl
...
+                {
+                    if (obj->nodesetval->nodeNr > 1)
+                    {
+                        isValue = false;
+                        result = false;
```
afishbeck, line 999: If result is false you have no idea if isValue=false is relevant.  For example the function may have returned at line 985 above, and you'd not know if isValue was updated or not.

timothyklemm: The original intent of isValue was to distinguish between nodeset and other evaluation results. The thinking at the time was that all values would be passed to maskValue and all node sets would be passed to maskContent. This was wrong, since an XPath to get an element value evaluates to a nodeset. The value is now predominantly an input parameter, remaining a reference for exactly one scenario - a value is expected and multiple nodes are found.

`trace-content` disregards isValue on output.

`trace-value` is only concerned with isValue on output when result is false. If this method returns at line 985, isValue will still be true and `trace-value` will not register a warning for node ambiguity. Returning false at line 1024 will trigger a warning if isValue has changed.

If this method could be counted on to register an appropriate warning, isValue could be changed from reference to value. I don't think it would be appropriate to pass a trace messenger from esdllib to xmllib to get the message we want.

afishbeck: I supose it's fine.  If you get a return of false and isValue is changed you can tell it was a nodeset above 1.  It's the kind of routine the caller has to understand internals of which isn't ideal but it's fine.

system/xmllib/libxml_xpathprocessor.cpp
```diff
@@ -991,8 +991,24 @@ class CLibXpathContext : public CInterfaceOf<IXpathContext>, implements IVariabl
...
+                    else
+                    {
+                        evaluateAsString(obj, content, ccXpath->getXpath());
+                        obj = nullptr; // freed above
+                    }
+                }
+                else
+                {
+                    toXMLNodeSet(obj, content);
+                    isValue = false;
```
afishbeck, line 1010: Isn't isValue already false?

afishbeck (approved):

afishbeck (approved):

asselitx (approved):

esp/esdllib/esdl_script.cpp
timothyklemm: The concern, such as it is, is not introduced by this change. As mentioned in prior comments, it was introduced by HPCC-27471. There are three additional PRs building on this one that are unrelated to this concern.

If this warrants a change, I suggest it be done in a new JIRA focused on the change.

timothyklemm:
@ghalliday please merge. As mentioned in another comment, I believe it would be better to address the open question in a separate issue if needed.

esp/esdllib/esdl_script.cpp
```diff
@@ -1652,26 +1658,20 @@ class CEsdlTransformOperationTrace : public CEsdlTransformOperationWithoutChildr
...
+            if (!m_category)
+                return false;
             if (m_test && !sourceContext->evaluateAsBoolean(m_test))
                 return false;
-            StringBuffer    content;
-            bool            isValue;
-            ITracer& tracer = scriptContext->tracerRef();
-            if (targetContext->selectText(m_select, content, isValue))
+
+            StringBuffer                      text;
```
ghalliday, line 1666: trivial: unusual indent of the identifier.

esp/esdllib/esdl_script.cpp
ghalliday: Probably the cleanest way would be to add a member function to StringBuffer to return a char * e.g. bufferBase() used in MemoryBuffer.

ghalliday (approved):
@timothyklemm one suggestion for removing the const cast, but will merge as-is.

PR #17013, opened 2023-02-27
HPCC-29022 Remove unreachable in esdl_script.cpp

timothyklemm:

## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/esdllib/esdl_script.cpp
+++ b/esp/esdllib/esdl_script.cpp
@@ -737,12 +737,10 @@ public:
         }
 
         int type = 0;
-        while((type = xpp.next()) != XmlPullParser::END_DOCUMENT)
+        while((type = xpp.next()) != XmlPullParser::END_TAG)
         {
-            switch(type)
+            if (XmlPullParser::START_TAG == type)
             {
-                case XmlPullParser::START_TAG:
-                {
                     StartTag stag;
                     xpp.readStartTag(stag);
                     const char *op = stag.getLocalName();
@@ -754,11 +752,6 @@ public:
                         readFullContent(xpp, m_sql);
                     else
                         xpp.skipSubTreeEx();
-                    break;
-                }
-                case XmlPullParser::END_TAG:
-                case XmlPullParser::END_DOCUMENT:
-                    return;
             }
         }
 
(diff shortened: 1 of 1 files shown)


esp/esdllib/esdl_script.cpp
```diff
@@ -1047,7 +1048,6 @@ class CEsdlTransformOperationHttpPostXml : public CEsdlTransformOperationBase
                     break;
                 }
                 case XmlPullParser::END_TAG:
-                case XmlPullParser::END_DOCUMENT:
```
afishbeck, line 1050: @timothyklemm what happens if the xml is badly formed and the document ends without the closing tag being hit?

timothyklemm: Malformed XML results in parser exceptions.

afishbeck: Just want to make sure we don't end up in an endless loop.  I think the reported issue was because END_DOCUMENT was checked in two places and would never get to the second.  If END_DOCUMENT can't happen without an exception that might be ok.  But if we can end up in a loop that's bad.

timothyklemm: In this and all changed cases, a start tag has been read and processing that element is in progress. To see an END_DOCUMENT before the balancing END_TAG implies malformed XML. The parser throws exceptions when malformed markup is parsed. No loops.

afishbeck (approved):

timothyklemm:
@ghalliday please merge.

esp/esdllib/esdl_script.cpp
```diff
@@ -688,8 +688,9 @@ class CEsdlTransformOperationMySqlCall : public CEsdlTransformOperationBase
         }
 
         int type = 0;
-        while((type = xpp.next()) != XmlPullParser::END_DOCUMENT)
```
ghalliday, line 691: This doesn't look right.  Previously END_DOCUMENT would have finished the loop, and continued, now it will return.

ghalliday (changes requested):
I think one change is incorrect

esp/esdllib/esdl_script.cpp
timothyklemm: The pull parser will throw an exception if it reaches the end of a document before an expected end tag. The END_DOCUMENT condition was never correct. The loop should probably exit on END_TAG, instead of END_DOCUMENT, and the inner switch can be simplified to just a test for START_TAG.

timothyklemm: I have changed loop terminating conditions from END_DOCUMENT, which is unexpected, to END_TAG. I have also created https://track.hpccsystems.com/browse/HPCC-29138 to address making the loops more robust in the event that future coding errors make END_DOCUMENT possible (it isn't at this time).

afishbeck (approved):

ghalliday:
@afishbeck please can you check the recent change - since it will now be going through a different path for an END_TAG

afishbeck (approved):

ghalliday (approved):
@timothyklemm please squash

PR #17015, opened 2023-02-27
HPCC-28979 Remove Clarion support from HIDL compiler

kenrowland:
Removed support in HIDL for generating Clarion code

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/esp/scm/ecllib.ecm
+++ b/esp/scm/ecllib.ecm
@@ -74,11 +74,3 @@ SCMInterface IClientInstantEcl(IInterface)
 
 extern "C" ECL_CLIENT_API IClientInstantEcl * createInstantEclClient();
 
-SCMclarion
-  map
-    module('instant ecl client')
-      createInstantEclClient procedure, *cppClientInstantEcl, pascal, raw, name('_createInstantEclClient@0')
-    end
-  end
-SCMend
-
--- a/tools/hidl/hidlgram.y
+++ b/tools/hidl/hidlgram.y
@@ -84,7 +84,6 @@ int  nCommentStartLine = -1;
   SCMEXPORTDEF
   SCMINTERFACE
   SCMENUM
-  SCMCLARION
   SCMEND
   ESPINCLUDE
   ESPSTRUCT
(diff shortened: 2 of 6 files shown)


esp/scm/ecllib.ecm
```diff
@@ -74,11 +74,3 @@ SCMInterface IClientInstantEcl(IInterface)
 
 extern "C" ECL_CLIENT_API IClientInstantEcl * createInstantEclClient();
 
-SCMclarion
```
afishbeck, line 77: As a separate task, you may want to look and see if we still need these files that use SCMInterface.  Even if we still use them it would be interesting to see how they are used and if we can move away from using them.  But again, no bearing on this PR.

afishbeck (approved):

tools/hidl/hidlgram.y
```diff
@@ -804,12 +803,7 @@ Module
 
 
 ModuleStart
- : MODULE ID
- {
-   CurModule = new ModuleInfo($2.getName());
-   LastProc = NULL;
- }
- | SCMINTERFACE ID
+ : SCMINTERFACE ID
```
ghalliday, line 806: @afishbeck any backward compatibility concerns removing module and scmclarion?

ghalliday (commented):
One question for tony, otherwise looiks like good cleanup.

tools/hidl/hidlgram.y
afishbeck: @ghalliday If we are able to build without the *.ecm files that this PR has removed then I don't think those keywords are used anywhere any longer.  It used to build clarion compatible code and HRPC related code for the defined interface.  We used to do quite a few of our binary internal interfaces this way so that we could support those things; all but these few files were converted to be directly writen in c++ a long time ago.

PR #17050, opened 2023-03-04
HPCC-29103 Helm: Improve use of external ClusterIssuers

afishbeck:
## Type of change:

## Checklist:
  - [ ] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/helm/examples/certmanager/README.md
+++ b/helm/examples/certmanager/README.md
@@ -11,14 +11,6 @@ The following steps can be used to set up cert-manager in a kubernetes cluster.
 
 --------------------------------------------------------------------------------------------------------
 
-## Install cert-manager custom resource defintions:
-
-This adds new custom resource types to kubernetes for certificate issuers and certificates.
-
-```
-kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.1.0/cert-manager.crds.yaml
-```
-
 ## Install cert-manager helm chart:
 
 Add Jetstack helm repo:
@@ -27,10 +19,10 @@ Add Jetstack helm repo:
 helm repo add jetstack https://charts.jetstack.io
 ```
 
-Install vault server.
+Install cert-manager.
 
 ```bash
-helm install cert-manager jetstack/cert-manager --version v1.1.0
+helm install cert-manager jetstack/cert-manager --set installCRDs=true --namespace cert-manager --create-namespace
 ```
 
 ## Run from HPCC-Platform/helm directory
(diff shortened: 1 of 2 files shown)


helm/examples/certmanager/README.md
```diff
@@ -11,14 +11,6 @@ The following steps can be used to set up cert-manager in a kubernetes cluster.
 
 --------------------------------------------------------------------------------------------------------
 
-## Install cert-manager custom resource defintions:
```
afishbeck, line 14: Adding --set installCRDs=true to the helm install line below means we don't have to separately apply the CRDs ourselves.

helm/hpcc/templates/issuers.yaml
```diff
@@ -92,9 +92,13 @@ spec:
 {{- end -}}
 
 {{- define "hpcc.addIssuer" -}}
+ {{- $issuerKind := (hasKey .me "kind" | ternary .me.kind "") -}}
  {{- $issuerEnabled := (hasKey .me "enabled" | ternary .me.enabled true) -}}
- {{- if $issuerEnabled }}
-  {{- if (hasKey .me "spec") -}} {{- /* no spec means we are referencing an external issuer and don't need to instantiate our own */ -}}
+ {{- $issuerPredefined := (hasKey .me "predefined" | ternary .me.predefined false) -}}
```
afishbeck, line 97: Adding "predefined: true" as another way of saying the issuer is defined externally only because it seems terraform may not make it easy to set something to null.  Previously setting "spec: null" was the intended way of indicating that it was an external issuer and that we shouldn't generate the k8s issuer resource.

jakesmith (approved):
@afishbeck - looks good, does it need to target 8.10? would 8.12 be more suitable?

helm/examples/certmanager/README.md
```diff
@@ -200,3 +208,16 @@ You should see a result similar to this:
...
+How to set up a cert-manager ClusterIssuer to use something like LetsEncrypt or Zerossl is outside the
+scope of this document, but once you have one set up you can use it to generate your public certificates
+by setting the public issuer values something like this:
+
+certificates:
+  issuers:
+    public:
+      name: zerossl-issuer
+      kind: ClusterIssuer
+      spec: null
```
ghalliday, line 221: Is this needed (especially since kind is ClusterIssuer)?

helm/hpcc/templates/issuers.yaml
ghalliday: Not worth changing, but I think this has the same behaviour as
`{{- $issuerPredefined := .me.predefined  -}}`
when used in the if below.

ghalliday (approved):
A couple of comments, but looks good to merge.  See Jake's question about the target version.

afishbeck:
@ghalliday @jakesmith I was hoping for 8.10.x because this is likely blocking the opinionated terraform from using the features that allow zerossl or letsencrypt for browser facing certificates.  Until that is in place users we get warnings from browsers that they have to dismiss in able to get to EclWatch etc.  And the site will not show as secure even though it is using TLS.

That being said if you think it's too risky for 8.10.x I can retarget.

helm/examples/certmanager/README.md
afishbeck: Yes with this change that's now true, I'll change it.

afishbeck:
@ghalliday removed "spec: null" from the README.  Should be ready to merge.

PR #17052, opened 2023-03-06
HPCC-29110 - Fix uncaught thrown exceptions in metrics

kenrowland:
Added try catch blocks around std::regex usages that throw exceptions

## Type of change:

## Checklist:
  - [x] The commit is signed.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Cloud-compatibility
  - [ ] Premature optimization

## Smoketest:

### DIFF
--- a/system/jlib/jmetrics.cpp
+++ b/system/jlib/jmetrics.cpp
@@ -116,6 +116,20 @@ MetricsManager &hpccMetrics::queryMetricsManager()
 }
 
 
+MetricsManager::MetricsManager()
+{
+    try
+    {
+        nameValidator = "^[A-Za-z][A-Za-z0-9.]*[A-Za-z0-9]$";
+    }
+    catch (std::regex_error &regex_error)
+    {
+        throw makeStringExceptionV(MSGAUD_operator, "Metrics manager initialized name validator regex with invalid expression, code=%d, what=%s",
+                                   regex_error.code(), regex_error.what());
+    }
+}
+
+
 MetricsManager::~MetricsManager()
 {
     for (auto const &sinkIt : sinks)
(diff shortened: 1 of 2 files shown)


system/jlib/jmetrics.cpp
```diff
@@ -116,6 +116,19 @@ MetricsManager &hpccMetrics::queryMetricsManager()
...
 
+MetricsManager::MetricsManager()
+{
+    try
+    {
+        nameValidator = "^[A-Za-z][A-Za-z0-9.]*[A-Za-z0-9]$";
+    }
+    catch (...)
+    {
+        throw MakeStringException(MSGAUD_operator, "Metrics manager initialized name validator regex with invalid expression");
```
jakesmith, line 127: trivial: MakeStringException should be considered deprecated. use makeStringException in this case (no vargs), or makeStringExceptionV if there are vargs.

system/jlib/jmetrics.cpp
```diff
@@ -140,52 +153,66 @@ bool MetricsManager::addMetric(const std::shared_ptr<IMetric> &pMetric)
...
-            }
-            else
-            {
-                rc = true;  // old metric no longer present, so it's considered unique
-            }
-            it->second = pMetric;
         }
     }
-    else
+    catch (...)
```
jakesmith, line 205: minor/question: are there any other exception classes other than std::regex that could be thrown above.

Not keen on catching ... in general, because (for example) it will also catch ISEH_Exception.
But also in this case if it's an IException (like the ones thrown in this func), this code will now lose those IException's. effectively replacing it with this new one.

jakesmith (changes requested):
@kenrowland - see comments

system/jlib/jmetrics.cpp
jakesmith: line 206 is still catching all exceptions, e.g. ISEH_Exceptions and IException's like the 2 this function could throw, and effectively replacing them with the new exception.
Should be changed to catch std::regex_error exception + ensure other original exceptions are reported or rethrown.

jakesmith (changes requested):
@kenrowland - still need to rework the catch all to a) not catch all and b) ensure any original exception thrown within the try/catch is reported not replaced by new exception.

system/jlib/jmetrics.cpp
kenrowland: Changed to catch std::regex error exception. Other exceptions thrown by the metrics code are not explicitly caught, so they are handled by any catch blocks in the caller's code (or program termination if not handled).

jakesmith (commented):
Looks fine, please squash.

kenrowland:
@ghalliday Please merge

jakesmith (approved):

system/jlib/jmetrics.cpp
```diff
@@ -140,52 +154,69 @@ bool MetricsManager::addMetric(const std::shared_ptr<IMetric> &pMetric)
     std::string name = pMetric->queryName();
 
     // Ensure metric name follows naming convention
-    if (std::regex_match(name, nameValidator))
+    try
     {
-        auto metaData = pMetric->queryMetaData();
-        std::unique_lock<std::mutex> lock(metricVectorMutex);
-        for (auto &metaDataIt: metaData)
+        if (std::regex_match(name, nameValidator))
```
ghalliday, line 159: I don't think this function ever throws an exception - so I don't think it needs protecting.

ghalliday (commented):
I'm not convinced by the change, because the code that has been protected with a try will never throw an exception (because the pattern is constant and valid).  It is a case of Coverity generating a false-positive.

I suspect the correct fix - if any is to add 
catch(std::exception & e)
 in all main() functions - e.g. end of daseerver.cpp - so that all std exceptions that are thrown are caught and reported cleanly.

system/jlib/jmetrics.cpp
kenrowland: According the the header in regex.h, it throws a regex_error. I based the need for a try catch on this. If this is in error, I can remove it.

  /**
   * @brief Indicates if there is a match between the regular expression @p e
   * and a string.
   *
   * @param __s     [IN] The string to match.
   * @param __re    [IN] The regular expression.
   * @param __flags [IN] Controls how the regular expression is matched.
   *
   * @retval true  A match exists.
   * @retval false Otherwise.
   *
   * @throws an exception of type regex_error.
   */
  template<typename _Ch_traits, typename _Str_allocator,
	   typename _Ch_type, typename _Rx_traits>
    inline bool
    regex_match(const basic_string<_Ch_type, _Ch_traits, _Str_allocator>& __s,
		const basic_regex<_Ch_type, _Rx_traits>& __re,
		regex_constants::match_flag_type __flags
		= regex_constants::match_default)
    { return regex_match(__s.begin(), __s.end(), __re, __flags); }

ghalliday (approved):
The whole change is probably unnecessary (a theoretical rather than real problem), but no reason not to merge it.

